PR #7489, opened 2023-03-30
[TEST] cache module metadata

zekun000:

### DIFF
--- a/aptos-move/aptos-vm/src/move_vm_ext/vm.rs
+++ b/aptos-move/aptos-vm/src/move_vm_ext/vm.rs
@@ -5,25 +5,34 @@ use crate::{
     move_vm_ext::{MoveResolverExt, SessionExt, SessionId},
     natives::aptos_natives,
 };
-use aptos_framework::natives::{
-    aggregator_natives::NativeAggregatorContext,
-    code::NativeCodeContext,
-    cryptography::{algebra::AlgebraContext, ristretto255_point::NativeRistrettoPointContext},
-    state_storage::NativeStateStorageContext,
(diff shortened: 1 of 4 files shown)


aptos-move/aptos-vm/src/move_vm_ext/vm.rs
```diff
@@ -63,10 +72,26 @@ impl MoveVmExt {
...
+            ret.clone()
+        } else {
+            drop(read);
+            let ret = Arc::new(aptos_framework::get_vm_metadata(
+                &self.inner,
+                module_id.clone(),
+            ));
+            self.cache.write().unwrap().insert(module_id, ret.clone());
+            ret
+        }
```
davidiw, line 92: can we verify that a previous write hasn't already done this work?
seems like a better approach might be to get a write lock, check if it doesn't exist, set it if it doesn't, then perform a read -- of course then you have the yuckiness of handling an impossible get that is conditional though it should have been set...

davidiw (commented):
I guess this addresses the eventual optimization needed for handling repeated queries in the execution side. I wonder what type of cache we should consider. For example, should we just try to cache everything? Do we want an LRU? Do we need to make it capable of handling attacks.

zekun000:
ideally the metadata cache should be binded with the module cache itself, that'd be part of the new loader work though.

aptos-move/aptos-vm/src/move_vm_ext/vm.rs
zekun000: I don't think it'd make big difference if I understand what you suggested here, currently VM is not accessed concurrently

PR #7491, opened 2023-03-30
[Move-prover] Allow using local variables when updating global ghost variables

### LINKED ISSUE #7286: [Bug][Move prover] Update statements in the spec block does not recognize variables local to the function
# 🐛 Bug

When updating a global spec variable in the `update` statement in the function body, the prover does not recognize variables that are local to the function. For instance, when checking the following code:

```
spec module {
    global ghost_total_share: num;
}

fun add_shares(pool: &mut Pool, shareholder: address, new_shares: u128): u128 {
   // ...
   spec {
       update ghost_total_share = ghost_total_share + new_shares;
   };
}
```
The prover will generate the following error:

```
error: undeclared `pool_u64_unbound::current_shares`
    ┌─ /Users/tengzhang/aptos-core/aptos-move/framework/aptos-stdlib/sources/pool_u64_unbound.move:143:64
    │
143 │                 update ghost_total_share = ghost_total_share + current_shares;
    │                                                                                                    ^^^^^^^^^^^^^^

{
  "Error": "Move Prover failed: exiting with model building errors"
}
```

## Expected Behavior

There should be no error when local variables or arguments are used in the `update` statement.

rahxephon89:
### Description

This PR allows using local variables and function arguments in `update` statements.
close #7286

### Test Plan

`global_vars.move` is revised to test using local variables when updating global ghost variables.


third_party/move/move-model/src/ast.rs
```diff
@@ -281,6 +281,8 @@ pub struct Spec {
     pub properties: PropertyBag,
     // If this is a function, specs associated with individual code points.
     pub on_impl: BTreeMap<CodeOffset, Spec>,
+    // The map to store ghost variable update statements inlined in the function body
+    pub update_map: BTreeMap<NodeId, Condition>,
```
wrwg, line 285: This seems to contain a map from the node id of inlined conditions in this function to the actual condition? I do not understand this construct. In order to get something out of this map, you need to have a `cond.node_id()`, and then you get `Some(cond')` back s.t. `cond == cond'`. Or am I missing something? Is this perhaps just better a `updates: BTreeSet<NodeId>`?

rahxephon89: This map is updated in [stackless_bytecode_generator.rs](https://github.com/aptos-labs/aptos-core/blob/2465c0799e1697df63b62f00411dec731f87b3fe/third_party/move/move-prover/bytecode/src/stackless_bytecode_generator.rs#L224), when we see an update of global spec variables stored in `cond`. Later, when we want to populate the field `updates` inside `translator.result` in the function `translate_inline_property`, we need to get [access](https://github.com/aptos-labs/aptos-core/blob/2465c0799e1697df63b62f00411dec731f87b3fe/third_party/move/move-model/src/spec_translator.rs#L286) to `cond`. With this map, we no longer need to [traverse](https://github.com/aptos-labs/aptos-core/blob/0a6bc636567e2e831f2a6f260bb961c36298201e/third_party/move/move-model/src/spec_translator.rs#L286) values stored in `translator.fun_env.get_spec().on_impl.values()` to get `cond`.

wrwg: Oh I see. Wow this was a terrible hack before ... (`cond.exp == prop.clone()`). Now it is slightly better, still somehow the abstractions are wrong. We are getting an exp and via its node id identify its parent condition. This is wild but a change would be a larger refactoring.

wrwg (approved):
Great work

third_party/move/move-model/src/ast.rs
rahxephon89: Thanks for the comment, Wolfgang! We could discuss more on how to refactor it later

meng-xu-cs (approved):
LGTM

third_party/move/move-model/src/ast.rs
meng-xu-cs: Hmm, PR merged. NVM, next time then.

PR #7492, opened 2023-03-30
[docs] Update docs explaining how to use debug statement in Move code

banool:
### Description
It was incorrectly referring to the module as `Debug` instead of `debug` (used to be correct, no longer) as well as saying you need to use the nursery package, which you don't need to do now.

### Test Plan
CI.


clay-aptos (approved):
Thank you, Daniel!

gregnazario (approved):

PR #7501, opened 2023-03-30
[TS SDK] add User-Agent header

0xmaayan:
### Description
Add User-Agent header to rest api calls 
### Test Plan
test are passing
and
```
2023-03-30T23:23:08.657964Z [api-18] DEBUG api/src/log.rs:53 {"elapsed":"835.083µs","method":"GET","path":"/v1/accounts/0x1","remote_addr":"127.0.0.1:65273","status":200,"user_agent":"aptos-ts-sdk / 1.7.2"}
```

### DIFF
--- a/ecosystem/typescript/sdk/src/providers/aptos_client.ts
+++ b/ecosystem/typescript/sdk/src/providers/aptos_client.ts
@@ -86,7 +86,7 @@ export class AptosClient {
     } else {
       conf.WITH_CREDENTIALS = true;
     }
-    conf.HEADERS = { ...conf.HEADERS, "User-Agent": `aptos-ts-sdk / ${VERSION}` };
+    conf.HEADERS = { ...conf.HEADERS, "User-Agent": `aptos-ts-sdk/${VERSION}` };
 
     this.client = new Gen.AptosGeneratedClient(conf);
   }
--- a/ecosystem/typescript/sdk/src/providers/aptos_client.ts
+++ b/ecosystem/typescript/sdk/src/providers/aptos_client.ts
@@ -35,6 +35,7 @@ import {
   AnyNumber,
 } from "../bcs";
 import { Ed25519PublicKey, MultiEd25519PublicKey } from "../aptos_types";
+import { VERSION } from "../version";
 
 export interface OptionalTransactionArgs {
   maxGasAmount?: Uint64;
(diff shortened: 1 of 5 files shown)


ecosystem/typescript/sdk/src/version.ts
```diff
@@ -0,0 +1,2 @@
+// hardcoded for now, we would want to have it injected dynamically
+export const VERSION = "1.7.2";
```
gregnazario, line 2: Please add to release instructions

gregnazario (approved):

ecosystem/typescript/sdk/src/providers/aptos_client.ts
```diff
@@ -85,6 +86,7 @@ export class AptosClient {
     } else {
       conf.WITH_CREDENTIALS = true;
     }
+    conf.HEADERS = { ...conf.HEADERS, "User-Agent": `aptos-ts-sdk / ${VERSION}` };
```
banool, line 89:
```suggestion
    conf.HEADERS = { ...conf.HEADERS, "User-Agent": `aptos-ts-sdk/${VERSION}` };
```

ecosystem/typescript/sdk/src/version.ts
```diff
@@ -0,0 +1,2 @@
+// hardcoded for now, we would want to have it injected dynamically
```
banool, line 1: Can't you do something like this?
```
import {version} from './package.json';
```

banool (changes requested):

ecosystem/typescript/sdk/src/version.ts
0xmaayan: not that easy, package.json is a configuration file and It doesn't contain any executable code that can be imported or used in the application, we need to do something like [this](https://github.com/alchemyplatform/alchemy-sdk-js/blob/main/inject-version.js) or [this](https://github.com/MystenLabs/sui/blob/main/sdk/typescript/genversion.mjs).
Also `import` used in esm modules while sdk supports both cjs and esm

banool: Okay unfortunate, let's roll with this for now.

banool (approved):

PR #6301, opened 2023-01-24
Tell user if their CLI is out of date

banool:
### Description
This PR adds a notification to tell the user if their CLI is out of date. We print to stderr to not mess up the output of the CLI for piping.

### Test Plan
1. Lower the version of the CLI and build it locally.
2. Move it somewhere else, e.g. `/tmp`
3. Run it:
```
$ /tmp/aptos info
===
A new version of the Aptos CLI is available: v1.0.3 -> v1.0.4
Run this command to update: aptos update
===

{
  "Result": {
    "build_branch": "main",
    "build_cargo_version": "cargo 1.66.1 (ad779e08b 2023-01-10)",
    "build_commit_hash": "9728485d20d17647200376d4f534436d5ddff65f",
    "build_is_release_build": "false",
    "build_os": "macos-aarch64",
    "build_pkg_version": "1.0.3",
    "build_profile_name": "debug",
    "build_rust_channel": "1.66.1-aarch64-apple-darwin",
    "build_rust_version": "rustc 1.66.1 (90743e729 2023-01-10)",
    "build_tag": "",
    "build_time": "2023-01-23 15:49:23 +00:00",
    "build_using_tokio_unstable": "true"
  }
}
```

PR #7505, opened 2023-03-31
[cli] Add update check once a day

gregnazario:
### Description
This adds a check at the beginning of the CLI once a day that checks if there's an update.

This does unfortunately add extra loading time on the first run a day.  It can be disabled with a flag, but it probably needs to be documented better.

### Test Plan
I tested this by setting the local version to "1.0.6", and it checked accordingly.

```
./target/debug/aptos -h
==UPDATE NOTICE==
 A new version of the Aptos CLI is out 1.0.9, run `aptos update` to update

aptos 1.0.6
Aptos Labs <opensource@aptoslabs.com>
Command Line Interface (CLI) for developing and interacting with the Aptos blockchain

```

### DIFF
--- a/crates/aptos/src/main.rs
+++ b/crates/aptos/src/main.rs
@@ -54,7 +54,7 @@ pub async fn check_for_update() {
                 .await
                 {
                     if update_info.update_required {
-                        eprintln!("==UPDATE NOTICE==\n A new version of the Aptos CLI is out {}, please run `aptos update` to update\n", update_info.latest_version);
+                        eprintln!("== UPDATE NOTICE ==\nA new version of the Aptos CLI is out {}, please run `aptos update` to update\n", update_info.latest_version);
                     }
                 }
 
--- a/crates/aptos/src/main.rs
+++ b/crates/aptos/src/main.rs
@@ -9,17 +9,24 @@
 #[global_allocator]
 static ALLOC: jemallocator::Jemalloc = jemallocator::Jemalloc;
 
-use aptos::{move_tool, Tool};
+use aptos::{config, move_tool, update::helpers::check_if_update_required, Tool};
 use clap::Parser;
-use std::process::exit;
+use std::{
+    process::exit,
+    time::{Duration, SystemTime},
+};
+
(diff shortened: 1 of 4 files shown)


banool:
I don't know if you saw all the conversation in https://github.com/aptos-labs/aptos-core/pull/6301 but we purposely decided not to do something like this. I tend to agree with our decision, this check should happen in a separate thread and only if it is done when the main command is done should we print it, otherwise we give up.

crates/aptos/src/main.rs
```diff
@@ -29,3 +36,32 @@ async fn main() {
...
+            Duration::from_secs(global_config.last_update_check_time.unwrap_or(0));
+        if let Ok(now) = SystemTime::now().duration_since(SystemTime::UNIX_EPOCH) {
+            if (now.saturating_sub(last_checked_time)) > ONE_DAY {
+                if let Ok(Ok(update_info)) = tokio::task::spawn_blocking(move || {
+                    check_if_update_required("aptos-labs", "aptos-core")
+                })
+                .await
+                {
+                    if update_info.update_required {
+                        eprintln!("==UPDATE NOTICE==\n A new version of the Aptos CLI is out {}, please run `aptos update` to update\n", update_info.latest_version);
```
banool, line 57:
```suggestion
                        eprintln!("== UPDATE NOTICE ==\nA new version of the Aptos CLI is out {}, please run `aptos update` to update\n", update_info.latest_version);
```

gregnazario:
> I don't know if you saw all the conversation in #6301 but we purposely decided not to do something like this. I tend to agree with our decision, this check should happen in a separate thread and only if it is done when the main command is done should we print it, otherwise we give up. Doing that with the underlying library is currently not possible without resorting to jank: [jaemk/self_update#100](https://github.com/jaemk/self_update/issues/100).

I didn't see this conversation earlier. I agree, this should probably be done in a different thread.  And we could do something like, only make these calls on interactions that interact with the network already.

I just have the argument that it decreases communication time, to make this.

The other option, is to have an endpoint with a simple notification system, which would be less overhead than querying every single release

PR #7506, opened 2023-03-31
[docs] Look up email -> username map automatically

banool:
### Description
Previously to make contributors.js work we maintained a map of emails to usernames for cases where the contributor email couldn't be used to figure out the username (this is only possible if the email ends with @users.noreply.github.com). With this PR, in addition to this map (which is unfortunately still partially necessary due to reasons I explain the comments), we use the GitHub GraphQL API to lookup usernames based on the contributor emails.

As you can see, the hardcoded map has been pared down because some emails are returned by the GraphQL query. 

Thanks to this PR you can see in `src/contributors.js` that a lot of usernames that were previously null are now filled in.

### Test Plan
```
pnpm contributors
```

### DIFF
--- a/developer-docs-site/README.md
+++ b/developer-docs-site/README.md
@@ -99,12 +99,15 @@ Fix formatting issues by running:
 pnpm fmt
 ```
 
-## Re-generate contributors
+## Regenerating contributors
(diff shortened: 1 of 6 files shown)


clay-aptos (approved):
Thank you so much, Daniel!

gregnazario (approved):

PR #7518, opened 2023-04-02
[Docs] update instruction for base-gas.md

nkysg:
### Description
update u16, u32, u256 and call per local for base-gas.md
### Test Plan

N/A


davidiw (approved):
thanks

PR #7520, opened 2023-04-02
Add string formatting routines to the aptos stdlib

gerben-stavenga:
### Description

A first pass at format functions for strings

1) 
```
fun to_string<T>(x: &T): String;
fun to_string_with_canonical_address<T>(x: &T): String;
fun to_string_with_integer_types<T>(x: &T): String
fun debug_string<T>(x: &T): String;
```

will convert the value x into a human readable string representation. The primitive types are displayed as expected,
vectors become [1, 2, ...] and structs are displayed as MyType { field1: ... , field2: ... }. The only exceptions are strings and options. Strings are displayed as "my string" and options as Some(...) or None.

2) 
```
fun format1<T0>(fmt: &String, arg1: T0);
fun format2<T0, T1>(fmt: &String, arg1: T0, arg2: T1);
fun format3<T0, T1, T2>(fmt: &String, arg1: T0, arg2: T1, arg3: T2);
fun format4<T0, T1, T2, T3>(fmt: &String, arg1: T0, arg2: T1, arg3: T2, arg4: T4);
```

will format the values in args according to the usual rust format spec "a = {} b = {}". 

### Test Plan
Unit tests


aptos-move/framework/src/natives/string_utils.rs
```diff
@@ -0,0 +1,233 @@
...
+    Ok(())
+}
+
+pub fn native_format(
+    _gas_params: &FormatGasParams,
+    context: &mut SafeNativeContext,
+    ty_args: Vec<Type>,
+    mut arguments: VecDeque<Value>,
+    ) -> SafeNativeResult<SmallVec<[Value; 1]>> {
+    debug_assert!(ty_args.len() == 1);
```
movekevin, line 126: We need to charge gas appropriately for all these new native functions and add those gas params to the gas schedule https://github.com/aptos-labs/aptos-core/blob/main/aptos-move/aptos-gas/src/aptos_framework.rs

aptos-move/framework/src/natives/string_utils.rs
```diff
@@ -0,0 +1,233 @@
...
+    let ty = context.deref().type_to_fully_annotated_layout(&ty_args[0])?.unwrap();
+    let x = safely_pop_arg!(arguments, Reference);
+    let v = x.read_ref().map_err(|e| SafeNativeError::InvariantViolation(e))?;
+    let mut out = String::new();
+    native_format_impl(context, &ty, v, &mut out)?;
+    let move_str = Value::struct_(Struct::pack(vec![Value::vector_u8(out.into_bytes())]));
```
movekevin, line 132: What format does this give?

aptos-move/framework/aptos-stdlib/sources/string_utils.move
```diff
@@ -0,0 +1,125 @@
...
+
+    struct NIL has copy, drop, store {}
+
+    public fun cons<T, N>(car: T, cdr: N): List<T, N> {
+        List { car, cdr }
+    }
+
+    public fun nil(): NIL { NIL {} }
+
+    native fun format<T>(s: &T): String;
```
movekevin, line 20: Can you add documentations regarding what these do? Also shouldn't these be public?

aptos-move/framework/src/natives/string_utils.rs
gerben-stavenga: String

movekevin: Lol but in what format? { field_name: field_value } or an array of field values?

gerben-stavenga: I added some examples in the comments. But basically "StructName { field1: ... , field2: ... }" for structs and [...,...] for vectors and the rest being trivial

gerben-stavenga: I was contemplating 0x1::module::StructName wdyt?

aptos-move/framework/src/natives/string_utils.rs
gerben-stavenga: I added gas

aptos-move/framework/aptos-stdlib/sources/string_utils.move
```diff
@@ -0,0 +1,143 @@
+module aptos_std::string_utils {
```
movekevin, line 1: Can you add more documentation here?

aptos-move/framework/aptos-stdlib/sources/string_utils.move
```diff
@@ -0,0 +1,143 @@
...
+    use std::string::String;
+
+    const EARGS_MISMATCH: u64 = 1;
```
movekevin, line 4: Can you add comments for these errors?

aptos-move/framework/aptos-stdlib/sources/string_utils.move
```diff
@@ -0,0 +1,143 @@
...
+    const EINVALID_FORMAT: u64 = 2;
+
+    struct Cons<T, N> has copy, drop, store {
```
movekevin, line 7: nit: What do cons/car/cdr stand for? Can we use better names lol

aptos-move/aptos-gas/src/aptos_framework.rs
```diff
@@ -119,5 +119,7 @@ crate::natives::define_gas_parameters_for_natives!(GasParameters, "aptos_framewo
     [.object.exists_at.base, { 7.. => "object.exists_at.base" }, 250 * MUL],
     // These are dummy value, they copied from storage gas in aptos-core/aptos-vm/src/aptos_vm_impl.rs
     [.object.exists_at.per_byte_loaded, { 7.. => "object.exists_at.per_byte_loaded" }, 1000],
-    [.object.exists_at.per_item_loaded, { 7.. => "object.exists_at.per_item_loaded" }, 8000]
+    [.object.exists_at.per_item_loaded, { 7.. => "object.exists_at.per_item_loaded" }, 8000],
+    [.string_utils.base, "string_utils.format.base", 300 * MUL],
```
movekevin, line 123: This would need to be 8.. => since gas schedule version 7 is already in testnet. cc @vgao1996

aptos-move/framework/src/natives/string_utils.rs
```diff
@@ -0,0 +1,325 @@
+// Copyright © Aptos Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::{
```
movekevin, line 4: Can we reuse functions from https://github.com/aptos-labs/aptos-core/blob/main/third_party/move/move-stdlib/src/natives/debug.rs#L250?

aptos-move/framework/aptos-stdlib/sources/string_utils.move
gerben-stavenga: Those are the standard names

aptos-move/framework/src/natives/string_utils.rs
gerben-stavenga: I think that has quadratic complexity, and is in behind testing feature flag.

aptos-move/aptos-gas/src/aptos_framework.rs
vgao1996: Yes, please bump LATEST_GAS_FEATURE_VERSION and use 8 here.

aptos-move/framework/src/natives/string_utils.rs
movekevin: cc @alinush What do you think?

aptos-move/framework/src/natives/string_utils.rs
movekevin: 0x1::module::StructName makes more sense yeah

gerben-stavenga: But it's incredible long and ugly most of the time

movekevin: Up to you then. I don't really see a strong use case for printing structs on chain anyway, may be as some kind of logs stored on chain (e.g. part of events)

aptos-move/framework/aptos-stdlib/sources/string_utils.move
```diff
@@ -1,8 +1,8 @@
 module aptos_std::string_utils {
     use std::string::String;
 
-    const EARGS_MISMATCH: u64 = 1;
-    const EINVALID_FORMAT: u64 = 2;
+    const EARGS_MISMATCH: u64 = 1;  // The number of values in the list does not match the number of "{}" in the format string.
```
wrwg, line 4: Please do documentation style (everywhere you add comments).

```
/// This code indicates a problem.
const EPROBLEM_CODE: u64 = 1;
```

In this case, these comments are actually extracted by a tool and put into an error description map which is included in Aptos metadata of a bytecode module. Not sure whether other comment styles would work with this.

aptos-move/framework/aptos-stdlib/sources/string_utils.move
wrwg: They have been introduced in the 60ties for the ubiquitous lists in Lisp. But they have been a bit come out of fashion, and more frequently today, you find `head` and `tail`.

aptos-move/framework/aptos-stdlib/sources/string_utils.move
```diff
@@ -35,8 +35,9 @@ module aptos_std::string_utils {
     public native fun format<T>(s: &T): String;
 
     // Format a list of move values as a human readable string.
-    // eg. format_list(&std::string::utf8(b"a = {} b = {} c = {}"), &cons(1, cons(2, cons(3, nil())))) == "a = 1 b = 2 c = 3"
-    public native fun format_list<T>(fmt: &String, val: &T): String;
+    // eg. format_list(b"a = {} b = {} c = {}", &cons(1, cons(2, cons(3, nil())))) == "a = 1 b = 2 c = 3"
```
wrwg, line 38: `///` please, and when you have inline examples, use markdown with `-quotes for code examples.

aptos-move/framework/aptos-stdlib/sources/string_utils.move
```diff
@@ -35,8 +35,9 @@ module aptos_std::string_utils {
...
+    // fmt must be utf8 encoded and must contain the same number of "{}" as the number of values in the list.
```
wrwg, line 39: s/"/`/

aptos-move/framework/aptos-stdlib/sources/string_utils.move
```diff
@@ -53,31 +54,31 @@ module aptos_std::string_utils {
...
         let l = cons(1, cons(2, cons(my_string, nil())));
-        let s = format_list(&std::string::utf8(b"a = {} b = {} c = {}"), &l);
+        let s = format_list(&b"a = {} b = {} c = {}", &l);
         assert!(s == std::string::utf8(b"a = 1 b = 2 c = \"My string\""), 1);
     }
 
     #[test]
     #[expected_failure(abort_code = EARGS_MISMATCH)]
     fun test_format_list_to_many_vals() {
         let l = cons(1, cons(2, cons(3, cons(4, nil()))));
```
wrwg, line 64: Now we would need an infix operator, than this could be even readable, e.g.: 

```
format(&b"{}/{}" , nominator<<denominator)
```

I definitely would like to have them in the language, and hopefully we have them in a quarter or two.

aptos-move/framework/src/natives/string_utils.rs
wrwg: > Can we reuse functions from https://github.com/aptos-labs/aptos-core/blob/main/third_party/move/move-stdlib/src/natives/debug.rs#L250?

For now please not. Sorry I know this is confusing as we now have duplicates of some native functions. But we decided a while ago to branch the native functions so we can adapt their APIs smoothly for our needs as our VM evolves. All our native code should be in `aptos-move/framework`. `third_party/move` contains reference implementations which we do not touch for now.

We can discuss how we push stuff from aptos-move/framework to third_party, but I would be careful with this, as any code we develop cannot reference back into aptos-core, which restricts it a bit. We need more experience with the current setup before we do this.

aptos-move/framework/src/natives/string_utils.rs
```diff
@@ -0,0 +1,322 @@
...
+pub fn native_format_impl(
+    context: &mut SafeNativeContext,
+    base_gas: InternalGas,
+    ty: &MoveTypeLayout,
+    val: Value,
+    out: &mut String,
+) -> SafeNativeResult<()> {
+    context.charge(base_gas)?;
+    match ty {
+        MoveTypeLayout::Bool => {
```
wrwg, line 65: Yeah, the same trick which @zjma uses for the algebra :-) In the absence of traits which could resolve this switch a compile time, without need of natives, we use the native level to resolve the overloading. 

Again a good argument why we need traits in the language.

aptos-move/framework/src/natives/string_utils.rs
```diff
@@ -0,0 +1,322 @@
...
+
+    let arg_mismatch = 1;
+    let invalid_fmt = 2;
+
+    let val = safely_pop_arg!(arguments, Reference);
+    let mut val = val
+        .read_ref()
+        .map_err(SafeNativeError::InvariantViolation)?;
+
+    let fmt = safely_pop_arg!(arguments, Reference);
```
wrwg, line 206: You can actually coerce this by reference avoiding a copy, as [seen here](https://github.com/aptos-labs/aptos-core/blob/a1ae60836f987cacb6ed3f731e585df7c684c67e/third_party/move/move-stdlib/src/natives/string.rs#L127).

aptos-move/framework/src/natives/string_utils.rs
```diff
@@ -0,0 +1,322 @@
...
+                abort_code: arg_mismatch,
+            })
+        }
+    };
+
+    let mut out = String::new();
+    let mut in_braces = false;
+    let mut in_escape = false;
+    for c in fmt.chars() {
+        if !in_escape && c == '\\' {
```
wrwg, line 242: Nit: this can be nicer done with a match, perhaps.


```
match (in_escape, c) {
  (false, '\\') => { in_escape = true; continue }
  (false, '{') => ..
```

wrwg (commented):
Looks great to me, just a few requests.

In general I think this is a good first step _if_ we eventually have better notation than `cons(...)`, which I think we can have via user infixes, which is a lowing hanging fruit.

Also we really need trades, so we can do all those things more cleanly and via static type resolution on Move level...

wrwg:
> Looks great to me, just a few requests.
> 
> In general I think this is a good first step _if_ we eventually have better notation than `cons(...)`, which I think we can have via user infixes, which is a lowing hanging fruit.
> 
> Also we really need trades, so we can do all those things more cleanly and via static type resolution on Move level...

Traits, no trades ;-)

aptos-move/framework/aptos-stdlib/sources/string_utils.move
gerben-stavenga: are you a programming language barbarian? know thy classics

wrwg: Really, this is out of fashion since the 70ties, where functionally languages like Miranda, SML have their historic roots. Thats all you need to know. Anything without a decent type system can not be taken serious, is pre-classic, and in fact a bit barbarian.

gerben-stavenga: Sorry I meant the facitious response to kevin not knowing the names. I put the names in as wink to the old ages

gerben-stavenga: The only thing is that it's not a list in mirande/haskell sense as the second pair is not List<T> but N. Making it a generic pair with first/second kind of hides the intended list property we want.

gerben-stavenga: I'm wondering if there is more use general use of this then just formatting

movekevin: Lol I guess I'm not a computer scientist

aptos-move/framework/aptos-stdlib/sources/string_utils.move
movekevin: This could technically be done currently as well. We can support something like

format(&b"{}/{}", vector[nominator, denominator]).

Would only work for values of the same type, but that should work for most common cases I think

aptos-move/framework/aptos-stdlib/sources/string_utils.move
wrwg: > The only thing is that it's not a list in mirande/haskell sense as the second pair is not List but N. Making it a generic pair with first/second kind of hides the intended list property we want.

LOL! Indeed it is totally untyped, a hack which allows us to workaround until we have either traits or C++ style overloading.

wrwg: totally untyped ==> totally dynamically typed

third_party/move/move-vm/types/src/values/values_impl.rs
```diff
@@ -2207,8 +2207,7 @@ impl Vector {
         Self::pack(type_param, vec![])
     }
 
-    pub fn unpack(self, type_param: &Type, expected_num: u64) -> PartialVMResult<Vec<Value>> {
-        check_elem_layout(type_param, &self.0)?;
+    pub fn unpack_unchecked(self) -> PartialVMResult<Vec<Value>> {
```
alinush, line 2210: Why is it okay to do this unchecked operation?

aptos-move/framework/aptos-stdlib/sources/string_utils.move
```diff
@@ -0,0 +1,144 @@
...
+
+    #[test]
+    fun test_format() {
+        assert!(format(&1u64) == std::string::utf8(b"1"), 1);
+        assert!(format(&false) == std::string::utf8(b"false"), 2);
+        assert!(format(&1u256) == std::string::utf8(b"1"), 3);
+        assert!(format(&vector[1, 2, 3]) == std::string::utf8(b"[1, 2, 3]"), 4);
+        assert!(format(&cons(std::string::utf8(b"My string"),2)) == std::string::utf8(b"Cons {car: \"My string\", cdr: 2}"), 5);
+        assert!(format(&std::option::none<u64>()) == std::string::utf8(b"None"), 6);
+        assert!(format(&std::option::some(1)) == std::string::utf8(b"Some(1)"), 7);
```
alinush, line 50: Can you try formatting a vector of structs here and, possibly, add more test cases from my [pretty-printing PR](https://github.com/move-language/move/pull/630/files)?

I recall that, in that PR, vectors of structs (and other things) were trickier to handle.

I'm surprised to see you found a way to avoid that trickiness/complexity.

Maybe this means that code can be simplified.

alinush: (See the test cases in [`M.move`](https://github.com/move-language/move/pull/630/files#diff-5c4a86cc4aed70ccf11001b91e3ae3760c2ea6b965964e9413ee5ca23f6e0c77))

third_party/move/move-vm/types/src/values/values_impl.rs
gerben-stavenga: I think the semantics of unpack is kind of ridiculous. We have a function to unpack a move-vector into a vector of move values, that needs us to give the number of elements just for a check. There is no actual way to get the number of elements, so it's all useless

vgao1996: @alinush I think this is fine.

The reason why we only had the checked `unpack` function was because at the time we implemented this, we didn't not want to open up values to clients outside the core VM (native functions included), and the `unpack` function we had was directly mimicking the `vec_unpack` instruction.

Nowadays there's a practical need to use values inside native functions, and converting a vector value into the Rust version is one of such reasonable use cases.

aptos-move/framework/src/natives/string_utils.rs
alinush: I'm surprised that there's a simpler way of doing this. My way seems hacky in comparison.

I don't understand yet how @gerben-stavenga was able to do this differently/more efficiently in this PR. I was working off of some example code that Sam (Mysten) gave me, trying to adapt it to every case.

gerben-stavenga: I figured out what is going on. One cannot use value_to_move_value with fully annotated layout, because that would panic here 

https://github.com/aptos-labs/aptos-core/blob/a1ae60836f987cacb6ed3f731e585df7c684c67e/third_party/move/move-core/types/src/value.rs#L242

This implementation is not using that function and is just traversing the value directly as it is printing. I refactored debug code.

1) debug.move is living in aptos-stdlib (not in move-stdlib) so I moved the native impl to aptos-stdlib natives impls
2) I made the debug native just print string. And use string-utils::debug_string to convert value to string

aptos-move/framework/src/natives/string_utils.rs
gerben-stavenga: I added options to either format Foo or 0x1::debug::Foo

aptos-move/framework/aptos-stdlib/sources/string_utils.move
gerben-stavenga: I moved those test over in a saner way (test in move as opposed to test using move-cli testing stdio output)

gerben-stavenga:
This is now ready for last round of review

aptos-move/framework/src/natives/string_utils.rs
gerben-stavenga: Rust uses {{ and }} for escaping brackets so I'm using the same now

aptos-move/framework/src/natives/string_utils.rs
gerben-stavenga: Done

aptos-move/framework/src/natives/string_utils.rs
```diff
@@ -0,0 +1,463 @@
...
+}
+
+fn native_format_impl(
+    context: &mut FormatContext,
+    layout: &MoveTypeLayout,
+    val: Value,
+    depth: usize,
+    out: &mut String,
+) -> SafeNativeResult<()> {
+    context.context.charge(context.base_gas)?;
```
movekevin, line 129: Do we need to charge gas based on the length of the output string? One potential problem I can think of here is I can parse a value with a very high number of nestings. This can lead to this function doing a lot more work than expected

aptos-move/framework/aptos-stdlib/sources/debug.move
```diff
@@ -1,6 +1,271 @@
 /// Module providing debug functionality.
 module aptos_std::debug {
-    native public fun print<T>(x: &T);
+    use std::string::String;
 
-    native public fun print_stack_trace();
+    public fun print<T>(x: &T) {
+        native_print(format(x));
```
movekevin, line 6: Nice refactoring to have this call the string formatting!

movekevin (approved):
LGTM except for the gas part, which I'll leave to you and @vgao1996 to think about + make sure gas is charged appropriately

aptos-move/framework/aptos-stdlib/sources/debug.move
alinush: Is this backwards compatible?

aptos-move/aptos-gas/src/aptos_framework.rs
```diff
@@ -184,5 +184,9 @@ crate::natives::define_gas_parameters_for_natives!(GasParameters, "aptos_framewo
     [.object.exists_at.base, { 7.. => "object.exists_at.base" }, 250 * MUL],
     // These are dummy value, they copied from storage gas in aptos-core/aptos-vm/src/aptos_vm_impl.rs
     [.object.exists_at.per_byte_loaded, { 7.. => "object.exists_at.per_byte_loaded" }, 1000],
-    [.object.exists_at.per_item_loaded, { 7.. => "object.exists_at.per_item_loaded" }, 8000]
+    [.object.exists_at.per_item_loaded, { 7.. => "object.exists_at.per_item_loaded" }, 8000],
+    [.debug.native_print.base_cost, {8.. => "debug.native_print.base_cost"}, MUL],
```
alinush, line 188: Did you mean to leave this as MUL?

aptos-move/framework/aptos-stdlib/sources/debug.move
gerben-stavenga: I mean it's debug and not prod right?

aptos-move/aptos-gas/src/aptos_framework.rs
alinush: Maybe add 1 * MUL

aptos-move/framework/src/natives/string_utils.rs
gerben-stavenga: Well it's charging gas on all function entries, so nesting and repeated elements start charging gas at each entry of this function, so works automatically

movekevin: True we'll be charging base for each level of nesting. Is this sufficient to avoid execution time attacks though?

aptos-move/framework/aptos-stdlib/sources/debug.move
alinush: But I don't see the `#[test_only]` clauses...

movekevin: Debug module is not deployed to any live networks

movekevin: native public fun print_stack_track => public fun might still be backward compatible since both visibility and function signature didn't change. Removing native might be fine

aptos-move/aptos-gas/src/aptos_framework.rs
vgao1996: Let's just change this to 0 to avoid confusion. 0 is fine since the functions are no-ops in production.

aptos-move/aptos-vm/src/natives.rs
```diff
@@ -50,17 +50,6 @@ pub fn aptos_natives(
...
-            aptos_move_stdlib::natives::nursery_natives(
-                CORE_CODE_ADDRESS,
-                aptos_move_stdlib::natives::NurseryGasParameters::zeros(),
-            )
-            .into_iter()
-            .filter(|(addr, module_name, _, _)| {
-                !(*addr == CORE_CODE_ADDRESS && module_name.as_str() == "event")
-            }),
-        )
         .collect()
```
vgao1996, line 53: This needs to be feature gated.

gerben-stavenga: I added the backup functions in debug.rs

aptos-move/framework/aptos-stdlib/sources/debug.move
gerben-stavenga: I added back the native as backup otherwise the thing could fork / crash.

aptos-move/aptos-gas/src/aptos_framework.rs
gerben-stavenga: Removed it totally

aptos-move/framework/src/natives/string_utils.rs
gerben-stavenga: Except for the special case of printing utf8 string all code path in this function are constant time so we only need to charge a constant amount each time

aptos-move/framework/aptos-stdlib/sources/debug.move
```diff
@@ -1,6 +1,272 @@
 /// Module providing debug functionality.
 module aptos_std::debug {
-    native public fun print<T>(x: &T);
+    use std::string::String;
 
-    native public fun print_stack_trace();
+    public fun print<T>(x: &T) {
```
wrwg, line 5: We probably want to have print2, print3, ... here as well.

wrwg (approved):

aptos-move/framework/aptos-stdlib/sources/debug.move
```diff
@@ -1,6 +1,272 @@
...
+    public fun print_stack_trace() {
+        native_print(native_stack_trace());
+    }
+
+    inline fun format<T>(x: &T): String {
+        aptos_std::string_utils::debug_string(x)
+    }
+
+    native fun native_print(x: String);
+    native fun native_stack_trace(): String;
```
davidiw, line 18: this is a breaking change, we probably should have kept around the original behavior for some period of time and deprecated it...

cc: @gregnazario

aptos-move/framework/aptos-stdlib/sources/string_utils.move
```diff
@@ -0,0 +1,146 @@
...
+    /// The format string is not valid.
+    const EINVALID_FORMAT: u64 = 2;
+
+    /// Format a move value as a human readable string,
+    /// eg. `to_string(&1u64) == "1"`, `to_string(&false) == "false"`, `to_string(&@0x1) == "@0x1"`.
+    /// For vectors and structs the format is similar to rust, eg.
+    /// `to_string(&cons(1,2)) == "Cons { car: 1, cdr: 2 }"` and `to_string(&vector[1, 2, 3]) == "[ 1, 2, 3 ]"`
+    /// For vectors of u8 the output is hex encoded, eg. `to_string(&vector[1u8, 2u8, 3u8]) == "0x010203"`
+    /// For std::string::String the output is the string itself including quotes, eg.
+    /// `to_string(&std::string::utf8(b"My string")) == "\"My string\""`
```
lightmark, line 16: It's so annoying to have the double quotes included. Why do we want to include that? @movekevin

PR #7524, opened 2023-04-03
[CLI] Do not suggest people use a faucet for mainnet accounts

banool:
### Description
The help message was suggesting to use a faucet for mainnet, which is ofc not right.

We could say "by transferring funds from another account / using an exchange / using a wallet on ramp" if we wanted. We could also link to a doc in aptos.dev that explains how to do all this. For now though I've just removed the faucet part.

### Test Plan
👀


clay-aptos:
I am adding @davidiw and @saharct as they worked recently to settle upon this related messaging in:
https://aptos.dev/guides/get-test-funds 

Mainnet - There is no faucet available in mainnet to fund your account, so the method for obtaining the Aptos APT tokens is beyond the scope of this site and entirely up to you. Apart from the difference in the funding process, the rest of your workflow remains the same as in devnet and testnet.

davidiw (approved):
this is good for this context

saharct (approved):

PR #7530, opened 2023-04-03
Ignore quorum store messages when recovery manager is active

vusirikala:
### Description
This PR solves the issue [BLK-462](https://linear.app/aptoslabs/issue/BLK-462/[quorum-store]-error-message-logged-when-using-recovery-manager). The PR introduces a `recovery_mode` flag in the epoch manager struct. The flag is set to true when the recovery manager starts, and is set to false when a new epoch starts. If the `recovery_mode` flag is set to true, the node ignores all the quorum store messages.


consensus/src/epoch_manager.rs
```diff
@@ -947,6 +953,9 @@ impl EpochManager {
         peer_id: AccountAddress,
         event: &UnverifiedEvent,
     ) -> anyhow::Result<()> {
+        if self.recovery_mode {
+            return Ok(());
```
zekun000, line 957: it looks better to me to inline this with the quorum_store_enabled condition like this
```
  if self.quorum_store_enabled || self.recovery_mode
```

consensus/src/epoch_manager.rs
```diff
@@ -426,6 +429,8 @@ impl EpochManager {
     }
 
     async fn initiate_new_epoch(&mut self, proof: EpochChangeProof) -> anyhow::Result<()> {
+        // The recovery manager exits before starting a new epoch. So, we unset the recovery_mode flag here.
+        self.recovery_mode = false;
```
zekun000, line 433: it may be better to make a symmetric assignment in `start_new_epoch`'s two branch

consensus/src/epoch_manager.rs
vusirikala: Done

consensus/src/epoch_manager.rs
vusirikala: Done

consensus/src/epoch_manager.rs
```diff
@@ -778,6 +784,8 @@ impl EpochManager {
     }
 
     async fn start_new_epoch(&mut self, payload: OnChainConfigPayload) {
+        // The recovery manager exits before starting a new epoch. So, we unset the recovery_mode flag here.
+        self.recovery_mode = false;
```
zekun000, line 788: this is not what I meant, I was suggesting that in the branch below LivenessStorageData::FullRecoveryData we set recovery_mode = false and PartialRecoveryData we set it to true.

consensus/src/epoch_manager.rs
```diff
@@ -951,7 +956,7 @@ impl EpochManager {
             UnverifiedEvent::BatchMsg(_)
             | UnverifiedEvent::SignedBatchInfo(_)
             | UnverifiedEvent::ProofOfStoreMsg(_) => {
-                if self.quorum_store_enabled {
+                if self.quorum_store_enabled || self.recovery_mode {
```
bchocho, line 959: Hmm, I think we need to filter out the quorum store events on recovery, just without printing out error messages. Quorum store can't do anything productive right now when in recovery.

bchocho (changes requested):

consensus/src/epoch_manager.rs
vusirikala: Made the change. Now the filter_quorum_store_events method returns Result<bool,Err>. The result indicates whether to filter out the messages or not.

consensus/src/epoch_manager.rs
```diff
@@ -838,46 +843,50 @@ impl EpochManager {
...
+                                        remote_peer = peer_id,
+                                        error = ?e,
+                                        unverified_event = unverified_event
+                                    );
+                                },
+                            }
+                        })
+                        .await;
+                },
+                Ok(false) => {}, // This occurs when the quorum store is not enabled, but the recovery mode is enabled. We filter out the messages, but don't raise any error.
```
bchocho, line 887: Could we make this return `Ok(())` and then for `Ok(true)` just make it a noop in the match, so we don't need to nest?

It's a bit of a nit, but nesting in the `Ok(true)` feels like the message being processed is only for a quorum store message, when in fact it can be a round manager or buffer manager message. Un-nesting will make it feel more like a true filter.

vusirikala: Gotcha. Thanks for the comment. Pushed the changes.

consensus/src/epoch_manager.rs
```diff
@@ -838,8 +843,11 @@ impl EpochManager {
 
         if let Some(unverified_event) = maybe_unverified_event {
             // filter out quorum store messages if quorum store has not been enabled
-            self.filter_quorum_store_events(peer_id, &unverified_event)?;
-
+            match self.filter_quorum_store_events(peer_id, &unverified_event) {
+                Ok(true) => (),
```
bchocho, line 847: nit: should probably be `{}`

bchocho (approved):

consensus/src/epoch_manager.rs
vusirikala: Done.

zekun000 (approved):

PR #7533, opened 2023-04-03
[forge] Mainnet Like Network Simulation

ibalajiarun:
### Description

This PR adds the ability to simulate as close to a real network as possible in Forge. It adds a new `NetEm` config that can support emulating delay, loss, corruption, bandwidth between any two targets within the same test, overcoming existing drawbacks where multiple chaos simulations couldn't be stacked. 

This provides the `MultiRegionNetworkEmulationTest` which enables simulating a multi-region network and optionally add chaos between nodes in the same region group. Also, `CpuChaosTest` enables stressing the CPU to mimic varying CPU cores between validators. 

This depends on an upstream change to chaos-mesh in https://github.com/chaos-mesh/chaos-mesh/pull/4017 which adds bandwidth emulation support to NetEm action type.

**TODOs**
- [ ] This test still needs some tuning to replicate previewnet settings. (Will do as a follow-up)
- [x] I will work with PEs to deploy the chaos-mesh change to other Forge clusters as well. It's currently deployed on the GCP forge-0 cluster only.

### Test Plan

Testing in GCP Forge, where the above chaos-mesh change is applied.


testsuite/testcases/src/real_network_simulation_test.rs
```diff
@@ -0,0 +1,224 @@
...
+macro_rules! FOUR_REGION_LINK_STATS_CSV {
+    () => {
+        "data/four_region_link_stats.csv"
+    };
+}
+
+/// A test to simulate a real (e.g. mainnet, previewnet) network with multiple regions
+/// in different clouds and varying CPU performance. It currently supports only 4
+/// regions, due to ChaosMesh limitations.
+pub struct RealNetworkSimulationTest {}
```
igor-aptos, line 27: I feel like we should either have configs, or separate out different types of simulations this class can do (and then combine using CompositeNetworkLoadTest like we do here for failpoints-based injection here - https://gist.github.com/igor-aptos/74c3637547cdae4e73d80e4e899d08c4 )

In the simplest form, I want to test out this config, with turning off cpu stress and replacing it with ExecutionDelayTest

also the configuration constants should probably be here, and then you can have a default() or MainnetLikeNetworkSimulation configuration

types/src/on_chain_config/consensus_config.rs
```diff
@@ -56,7 +56,7 @@ impl OnChainConsensusConfig {
 
     pub fn quorum_store_enabled(&self) -> bool {
         match &self {
-            OnChainConsensusConfig::V1(_config) => false,
+            OnChainConsensusConfig::V1(_config) => true,
```
igor-aptos, line 59: undo, V2 is default now

igor-aptos (changes requested):

testsuite/testcases/src/real_network_simulation_test.rs
ibalajiarun: I think I did what you asked for. Let me know!

testsuite/testcases/src/modifiers.rs
```diff
@@ -194,3 +197,73 @@ impl Test for NetworkUnreliabilityTest {
...
+    let validator_chunks = chunk_validators(all_validators, num_groups);
+
+    let group_cpu_stresses = validator_chunks
+        .into_iter()
+        .enumerate()
+        .map(|(idx, chunk)| GroupCpuStress {
+            name: format!("group-{}-cpu-stress", idx),
+            target_nodes: chunk,
+            num_workers: (num_groups - idx) as u64,
+            load_per_worker: 100,
```
igor-aptos, line 232: this seems like something that should be in CpuChaosConfig

igor-aptos (approved):
awesome , thanks, looks much cleaner!

testsuite/forge/src/backend/k8s/chaos.rs
```diff
@@ -166,12 +178,92 @@ impl K8sSwarm {
...
+                instance_labels = &source_instance_labels,
+                target_instance_labels = &target_instance_labels,
+                rate = group_netem.rate_in_mbps,
+            ));
+        }
+
+        Ok(network_chaos_specs.join("\n---\n"))
+    }
+
+    fn create_cpu_stress_template(&self, swarm_cpu_stress: &SwarmCpuStress) -> Result<String> {
```
sitalkedia, line 229: Having a comment here (or somewhere else relevant) explaining how the CPU stress work would be useful.

testsuite/testcases/src/multi_region_network_test.rs
```diff
@@ -0,0 +1,314 @@
...
+    delay_correlation_percentage: u64,
+    loss_percentage: u64,
+    loss_correlation_percentage: u64,
+}
+
+impl Default for IntraRegionNetEmConfig {
+    fn default() -> Self {
+        Self {
+            bandwidth_rate_mbps: 10 * 1000, // 10 Gbps
+            delay_latency_ms: 50,
```
sitalkedia, line 160: Curious, how did we come up with these configs and how did we verify they result in a mainnet like setup?

sitalkedia (approved):
LGTM % minor nits.

testsuite/forge/src/backend/k8s/chaos.rs
```diff
@@ -32,6 +32,18 @@ macro_rules! NETWORK_LOSS_CHAOS_TEMPLATE {
     };
 }
 
+macro_rules! NETEM_CHAOS_TEMPLATE {
```
perryjrandall, line 35: why are these macros?

Also is it worth just defining this entire struct in code instead of a yaml file?

You can create an struct or hashmap and serialize it after modifying it?

I really hate the helm level of indirection

perryjrandall (approved):

testsuite/forge/src/backend/k8s/chaos.rs
ibalajiarun: I think this was originally done to enable templating and replace certain placeholders with values from a config struct. I don't see how to do it entirely with a well-defined struct.

testsuite/testcases/src/multi_region_network_test.rs
ibalajiarun: Because we couldn't simulate more than 4 inter-regions, Igor and I decided to mandate a minimum latency between validators in the same region. Igor suggested this number based on previewnet experience, but I think this also somewhat falls short of mainnet-like experience. The idea is to tune this in a follow-up to see how close to mainnet we can get as possible.

PR #7547, opened 2023-04-03
[Storage][Sharding] Wrap DBs in StateMerkleDb.

grao1991:
### Description
No-op refactoring, wrap DBs, and pass StateMerkleDb around as parameters instead of the raw DB object.

### DIFF
--- a/storage/aptosdb/src/lru_node_cache.rs
+++ b/storage/aptosdb/src/lru_node_cache.rs
@@ -6,11 +6,18 @@ use aptos_infallible::Mutex;
 use aptos_jellyfish_merkle::node_type::NodeKey;
 use aptos_types::{nibble::nibble_path::NibblePath, transaction::Version};
 use lru::LruCache;
+use std::fmt;
 
 const NUM_SHARDS: usize = 256;
 
 pub(crate) struct LruNodeCache {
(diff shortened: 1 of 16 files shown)


storage/aptosdb/src/lru_node_cache.rs
```diff
@@ -9,6 +9,7 @@ use lru::LruCache;
 
 const NUM_SHARDS: usize = 256;
 
+#[derive(Debug)]
```
msmouse, line 12: This will print a lot in the log if used.. shall we implement it manually?

storage/aptosdb/src/versioned_node_cache.rs
```diff
@@ -13,6 +13,7 @@ use std::{
 
 type NodeCache = HashMap<NodeKey, Node>;
 
+#[derive(Debug)]
```
msmouse, line 16: ditto

storage/aptosdb/src/lru_node_cache.rs
grao1991: Done

storage/aptosdb/src/versioned_node_cache.rs
grao1991: Done

msmouse (approved):

areshand (approved):

PR #7548, opened 2023-04-03
[db-tool] do module-validation from backup-verify instead of replay-verify

msmouse:
### Description

Avoids rebuilding the tree and writing to the DB, hence costs less / is faster.

### Test Plan

https://github.com/aptos-labs/aptos-core/actions/runs/4601955211/jobs/8130405414


rustielin (approved):
LGTM by test plan. 

The test run against `testnet` on this PR also looks good btw: https://github.com/aptos-labs/aptos-core/actions/runs/4602511956/jobs/8131553014?pr=7548

Will you need to cherry-pick this against any branches btw?

msmouse:
> Will you need to cherry-pick this against any branches btw?

I don't think so, it's just a performance improvement for the purpose of this PR. The workflow change depends on the interface change though.. so when running from other branches we need to user workflows there as well.

areshand (approved):

PR #7552, opened 2023-04-04
[Execution Benchmark] Adding github workflow for sequential / parallel execution performance testing

danielxiangzl:
### Description

To prevent any potential degradation in the execution performance, this PR adds a workflow to automatically run the sequential / parallel execution only performance test for any PR. The test is named `execution-performance`.

### TODOs:
- [x] Tune the threshold numbers.
- [ ] More types of machines to run the test (Intel & AMD).
- [ ] Make the test land blocking?
- [ ] Generate flamegraph if possible.


.github/workflows/executor-tests.yaml
```diff
@@ -0,0 +1,51 @@
...
+on:
+  pull_request:
+    paths:
+      - 'aptos-move/'
+      - 'third_party/move/'
+
+jobs:
+  executor-tests:
+    # The biggest machine, you can try some other like ubuntu-latest
+    runs-on: high-perf-docker
```
perryjrandall, line 11: Add back the "if label" statement

```
    if: |
      contains(github.event.pull_request.labels.*.name, 'CICD:your-label-here')
```

.github/workflows/executor-tests.yaml
```diff
@@ -0,0 +1,51 @@
+name: "sequential-execution-performance"
+on:
+  pull_request:
```
perryjrandall, line 3: Lets add workflow_dispatch here as well, that will allow you to run it from outside a PR as well

perryjrandall:
Generally looking good, one annoying facet of github actions workflows is that its not possible to test them until they've landed in main.... a security feature?

.github/workflows/executor-tests.yaml
```diff
@@ -0,0 +1,51 @@
...
+          fetch-depth: 0 # get all the history because cargo xtest --change-since origin/main requires it.
+      - uses: aptos-labs/aptos-core/.github/actions/rust-setup@main
+        with:
+          GIT_CREDENTIALS: ${{ secrets.GIT_CREDENTIALS }}
+      - uses: taiki-e/install-action@v1.5.6
+        with:
+          tool: nextest
+      # Run your actually test command
+      # REPLACE
+      # If you have anything other than a single command here, please write a python script and call it here
```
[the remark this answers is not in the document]
rustielin: yes I've outlined some best practices for how we're using GHA in our repo here: https://www.notion.so/aptoslabs/Github-Actions-Best-Practices-d7b5df833be54e8bad2e95c8a3f0b988. I'll also add some comments directly to this PR. But yea let's try to keep the logic within the actions themselves quite minimal, and instead call out to a script or tool that lives within the repo.

.github/workflows/executor-tests.yaml
rustielin: I'd do the following here:
* Create a reusable workflow called `workflow-run-executor-benchmark.yaml`, and have it triggerable `workflow_call`
* Then, in a separate workflow (the entrypoint, e.g. `executor-tests.yaml`), trigger the reusable workflow above via workflow call. `executor-tests.yaml` can then be triggered by either `pull_request(target)` (on PR), or `workflow_dispatch` (manually via UI or API/CLI)

For an example, check out how it's implemented for replay-verify:
* Here's the reusable workflow, which contains the meat of the logic for running the job: https://github.com/aptos-labs/aptos-core/blob/main/.github/workflows/workflow-run-replay-verify.yaml
* Here's the workflow that calls it: https://github.com/aptos-labs/aptos-core/blob/main/.github/workflows/replay-verify.yaml

rustielin (dismissed):
Left some comments above on GHA best practices. Testing might be tricky as Perry said, since some CI refers to the `main` branch for safety and consistency reasons. Let me know if you need help with testing your PR

.github/workflows/execution-performance.yaml
```diff
@@ -0,0 +1,10 @@
+name: "execution-performance"
+on:
+  pull_request:
+
+jobs:
+  execution-performance:
+    uses: ./.github/workflows/workflow-run-execution-performance.yaml
```
perryjrandall, line 7: Reference @main

perryjrandall (approved):
Lfg

sitalkedia:
@danielxiangzl - Can you clarify when does this run? Is this a land blocking test?

danielxiangzl:
> @danielxiangzl - Can you clarify when does this run? Is this a land blocking test?

Currently for every PR due to github workflow limitations. We will first land it and later decide whether it should be land blocking, but I personally prefer it to be land blocking for detecting any execution performance degradation.

zekun000 (approved):

PR #7555, opened 2023-04-04
[network] cleanup unused rate limiter

zekun000:
Rate limiter is set to open, so essentially a no-op, but it relies on IpAddr for bucketing which means all the dns addr falls under the same bucket and would contend for the same mutex. This commit removes it from the stack.

### DIFF
--- a/network/src/peer_manager/builder.rs
+++ b/network/src/peer_manager/builder.rs
@@ -5,7 +5,6 @@
 use crate::{
     application::storage::PeersAndMetadata,
     counters,
-    counters::NETWORK_RATE_LIMIT_METRICS,
     noise::{stream::NoiseStream, HandshakeAuthMode},
     peer_manager::{
         conn_notifs_channel, ConnectionRequest, ConnectionRequestSender, PeerManager,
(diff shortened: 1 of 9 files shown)


igor-aptos (approved):

bchocho (approved):
We'll have to remember that IpAddr is not a good key (at least without some kind of DNS lookup)

JoshLind:
> but it relies on IpAddr for bucketing which means all the dns addr falls under the same bucket and would contend for the same mutex. This commit removes it from the stack.

Instead of removing it from the stack like this, is there no way to simply fix the issue that we're still contending for the same mutex even if it's disabled? I worry that we may eventually want to maintain/adopt this functionality in the future?

network/src/peer_manager/builder.rs
```diff
@@ -469,24 +442,3 @@ impl PeerManagerBuilder {
...
-
-/// Builds a token bucket rate limiter with attached metrics
-fn token_bucket_rate_limiter(
-    network_context: &NetworkContext,
-    label: &'static str,
-    input: Option<RateLimitConfig>,
-) -> TokenBucketRateLimiter<IpAddr> {
-    if let Some(config) = input {
-        if config.enabled {
-            return TokenBucketRateLimiter::new(
```
davidiw, line 481: should we get rid of the TBRL?

zekun000: maybe onboarding task? 😄

zekun000:
>is there no way to simply fix the issue that we're contending for the same mutex even if it's disabled?

we can fix it simply but it seems hacky and confusing, I did it for the preview branch by just removing it at the bottle layer.

PR #7556, opened 2023-04-04
[txn-generator] Add transaction that changes a shared resource

igor-aptos:
We saw very different behavior with transactions that require serial execution, so adding such type to transaction emitter.

With this we have very granular control, setting num_modules param allows to specify the parallelism (as each published module is accessing a resource stored on the publisher account.

I am passing an address here, because I couldn't figure out how to refer to address of an account under which module is published, within the move code?


perryjrandall (approved):

testsuite/module-publish/src/packages/simple/sources/Simple.move
```diff
@@ -91,6 +95,22 @@ module 0xABCD::Simple {
...
+        let value = {
+            assert!(exists<Counter>(destination), error::invalid_argument(ECOUNTER_RESOURCE_NOT_PRESENT));
+
+            let counter = borrow_global_mut<Counter>(destination);
+            *(&mut counter.count) = counter.count + COUNTER_STEP;
+            counter.count
+        };
+        if (exists<Counter>(signer::address_of(owner))) {
+            let counter_owner = borrow_global_mut<Counter>(signer::address_of(owner));
+            *(&mut counter_owner.count) = value;
```
movekevin, line 108: counter_owner.count = value

testsuite/smoke-test/src/txn_emitter.rs
```diff
@@ -79,11 +81,12 @@ async fn test_txn_emmitter() {
                 (TransactionType::default_call_different_modules(), 20),
                 (
                     TransactionType::CallCustomModules {
-                        entry_point: EntryPoints::BytesMakeOrChange {
-                            data_length: Some(32),
-                        },
+                        entry_point: EntryPoints::StepDst,
+                        // entry_point: EntryPoints::BytesMakeOrChange {
```
movekevin, line 85: Why not just delete this?

testsuite/module-publish/src/packages/simple/sources/Simple.move
```diff
@@ -91,6 +95,22 @@ module 0xABCD::Simple {
         counter.count;
     }
 
+    public entry fun step_destination(owner: &signer, destination: address) acquires Counter {
```
movekevin, line 98: It'd be closer to Renfield and Metapixel minting if we add a batch index here (have multiple different counters that correspond to different batch indices) and have different transmitter instances use those different indices to simulate the account pool that Renfield and Metapixel have.

testsuite/smoke-test/src/txn_emitter.rs
igor-aptos: this file is generally for easy testing, so having way to uncomment and comment seems fine

testsuite/module-publish/src/packages/simple/sources/Simple.move
igor-aptos: not equivalent, but argument:
  num_modules: 1,
allows us to control parallelism. 

instead of having multiple resources in a single module, we publish the same module multiple times, each with single resource. I think that is fine for now.

considering the results (https://aptos-org.slack.com/archives/C03F71PU3B5/p1680760358119509), this is anyways more efficient than those contracts, so we will be adding token minting and other closer operations as well

movekevin: Using the same module would change the efficiency actually since caching would kick in

igor-aptos: caching is per block. If we have working set of 10 modules, that is going to have the same performance (given thousands of transactions in a block). 
If we set to 1000 modules, then yes, module loading will change things.

igor-aptos: And to clarify, goal of this method is not to be representative of Renfield/Metapixel, but to have a simplest serial workload that we can benchmark and compare to others in isolation

We will have separate transaction types for Renfield/Metapixel-like workload, and should make sure those are representative as much as possible

crates/transaction-generator-lib/src/publishing/module_simple.rs
```diff
@@ -303,18 +306,24 @@ fn set_name(rng: &mut StdRng, module_id: ModuleId) -> TransactionPayload {
...
         bcs::to_bytes(&other).unwrap(),
     ])
 }
 
+fn step_dst(module_id: ModuleId, dst: &AccountAddress) -> TransactionPayload {
+    get_payload(module_id, ident_str!("step_destination").to_owned(), vec![
+        bcs::to_bytes(&dst).unwrap(),
+    ])
+}
+
```
davidiw, line 326: this seems off, we pass in a ref, and then pass in a ref to a ref?

davidiw (approved):

Issue #6295, opened 2023-01-23
[Feature Request] Autodetect AccountId for an inputted PrivateKey


gregnazario:
# 🚀 Feature Request

## Motivation

We can lookup onchain with the public key to see if a key is associated to a rotated public key.  This would be easier than also having to pass in the account address, which is what we should do instead.

## Pitch

Automatically discover account address based on private/public key pair

PR #7560, opened 2023-04-04
[CLI] Update load_account_arg to pull  account address directly if exist

### LINKED ISSUE #6666: [Bug] CLI: Rotating keys breaks named-addresses
# 🐛 Bug

When using `named-addresses` with a profile name, it seems that it passes account address somehow differently which leads to different errors (depending on complexity of a module and places where named addresses are used), when we try to upgrade (publish again) the module after rotating the keys:

```
{
  "Error": "Simulation failed with status: Transaction Executed and Committed with Error MODULE_ADDRESS_DOES_NOT_MATCH_SENDER"
}
```

```
{
  "Error": "Simulation failed with status: Transaction Executed and Committed with Error CONSTRAINT_NOT_SATISFIED"
}
```

## To reproduce

**Code snippet to reproduce**
A minimal example of a module

```
module TestCoinType::test_coin {
    struct TestCoin {}

    fun init_module(sender: &signer) {
        aptos_framework::managed_coin::initialize<TestCoin>(
            sender,
            b"Test Coin",
            b"TSTC",
            6,
            false,
        );
    }
}
```

In `Move.toml` we have (I've also tried different revisions of AptosFramework: `devnet`, `mainnet`, `testnet`)
```
[addresses]
TestCoinType = '_'
```

Then, let's publish first revision of a module

`aptos move publish --named-addresses TestCoinType=default --profile default`

Module got published (Here is an example in explorer https://explorer.aptoslabs.com/account/0xe4dea7dbde2b43630231733579eb03ddfe324c96c151c1fe502ce172bce18ece/transactions)

Then let's rotate the keys with these commands

```
aptos key generate --output-file test
aptos account rotate-key --new-private-key-file=test --profile=default
```

I saved the profile with the name `default2` just in case (it gas the same account address as `default`). 
Than let's try to publish module again

`aptos move publish --named-addresses=default2 --profile default2`

We get an error

```
{
  "Error": "Simulation failed with status: Transaction Executed and Committed with Error MODULE_ADDRESS_DOES_NOT_MATCH_SENDER"
}
```

If we try to publish module and pass the exact account address (not a profile reference):

`aptos move publish --named-addresses TestCoinType=e4dea7dbde2b43630231733579eb03ddfe324c96c151c1fe502ce172bce18ece --profile default2`

The module will be published successfully. 

**Stack trace/error message**
```
{
  "Error": "Simulation failed with status: Transaction Executed and Committed with Error MODULE_ADDRESS_DOES_NOT_MATCH_SENDER"
}
```

## Expected Behavior

`--named-addresses` should point to the profile account address, if it did, we would be able to successfully publish the module after rotating the keys.

## System information

**Please complete the following information:**
- Aptos Core version: I've tried different revisions for `AptosFramework`: `devnet`, `mainnet`, `testnet`. 
- Cli version: 1.0.6
- Rust Version: 1.67.1
- OS: MacOS 13.0

0xjinn:
### Description
We should be pulling the account address directly from the profile specified if exist. Instead of derive the account_address from private key

For more detail of the issue, please refer to issue #6666 

resolve #6666 

### Test Plan
Tested locally to make sure it works even with key rotated. (Followed the same instruction from the issue described, and make sure publish successfully)

```
aptos-debug move publish --named-addresses TestCoinType=default2 --profile default2
BUILDING module_test
package size 970 bytes
Do you want to submit a transaction for a range of [17400 - 26100] Octas at a gas unit price of 100 Octas? [yes/no] >
yes
{
  "Result": {
    "transaction_hash": "0x4231bd0efaae519fa36c65e5e54ecac701e33c14ebfb8832ce6f0b1eede3e725",
    "gas_used": 174,
    "gas_unit_price": 100,
    "sender": "d53a65d05c8ca1160920573c54bb4ebaf06be19d6fcb540c5374a62003e65d79",
    "sequence_number": 4,
    "success": true,
    "timestamp_us": 1680592034050222,
    "version": 2828866,
  }
}
```

Double checked from the explore, and make sure transaction succeed on devnet - https://explorer.aptoslabs.com/txn/2828866/userTxnOverview

### DIFF
--- a/crates/aptos/src/common/types.rs
+++ b/crates/aptos/src/common/types.rs
@@ -994,6 +994,11 @@ pub fn load_account_arg(str: &str) -> Result<AccountAddress, CliError> {
         })
     } else if let Ok(account_address) = AccountAddress::from_str(str) {
         Ok(account_address)
+    } else if let Some(Some(account_address)) =
+        CliConfig::load_profile(Some(str), ConfigSearchMode::CurrentDirAndParents)?
+            .map(|p| p.account)
+    {
+        Ok(account_address)
     } else if let Some(Some(private_key)) =
         CliConfig::load_profile(Some(str), ConfigSearchMode::CurrentDirAndParents)?
             .map(|p| p.private_key)


crates/aptos/src/common/types.rs
```diff
@@ -994,6 +994,11 @@ pub fn load_account_arg(str: &str) -> Result<AccountAddress, CliError> {
         })
     } else if let Ok(account_address) = AccountAddress::from_str(str) {
         Ok(account_address)
+    } else if let Some(Some(account_address)) =
+        CliConfig::load_profile(Some(str), ConfigSearchMode::CurrentDirAndParents)?
+            .map(|p| p.account)
+    {
+        Ok(account_address)
```
gregnazario, line 1001: This works great if it's in the CLI.

I do have one more extension to this, which will resolve another GitHub issue, and get you deeper into the weeds of address rotation: https://github.com/aptos-labs/aptos-core/issues/6295

Take a look at `aptos account lookup-address`

gregnazario (approved):

crates/aptos/src/common/types.rs
gregnazario: Similarly, this would also be good to do account management:
* If my private key exists, and the address that I see locally has a different auth key, I can provide a useful message saying the private key isn't supported for the account, it might be rotated or wrong.
* This kind of support will allow for future management of keys in the CLI

0xjinn: Sounds good!

For account management mentioned here, is this part of the Q2 item "Multisig account /account management"?

PR #7561, opened 2023-04-04
[CLI] Fix e2e tests following change to `--skip-faucet`

banool:
### Description
In the tests we were using the `--skip-faucet` flag, which until https://github.com/aptos-labs/aptos-core/pull/7509 did not actually skip the faucet for a local network. Now that it works as intended, using it in these CLI tests is wrong.

Note: Even with this change, the tests are only fully correct if this PR also lands: https://github.com/aptos-labs/aptos-core/pull/7562.

### Test Plan
First, make sure this PR is stacked on top of https://github.com/aptos-labs/aptos-core/pull/7562.

Build a CLI:
```
cargo build -p aptos
```

Run the e2e tests using that CLI:
```
DOCKER_DEFAULT_PLATFORM=linux/amd64 poetry run python main.py --image-repo-with-project us-west1-docker.pkg.dev/aptos-global/aptos-internal --base-network devnet --test-cli-path ../../../target/debug/aptos
```

See that they all pass now:
```
2023-04-04 12:18:50,149 - INFO - Running aptos CLI local testnet from image: us-west1-docker.pkg.dev/aptos-global/aptos-internal/tools:devnet
2023-04-04 12:19:07,590 - INFO - Running test: test_init
2023-04-04 12:19:09,372 - INFO - Running test: test_account_fund_with_faucet
2023-04-04 12:19:11,159 - INFO - Running test: test_account_create
2023-04-04 12:19:13,104 - INFO - Stopping container: aptos-tools-devnet
2023-04-04 12:19:13,321 - INFO - Stopped container: aptos-tools-devnet
2023-04-04 12:19:13,321 - INFO - These tests passed:
2023-04-04 12:19:13,321 - INFO - test_init
2023-04-04 12:19:13,321 - INFO - test_account_fund_with_faucet
2023-04-04 12:19:13,321 - INFO - test_account_create
2023-04-04 12:19:13,321 - INFO - All tests passed!
```

PR #7562, opened 2023-04-04
[CLI] Wait for the funding transactions in aptos init

banool:
### Description
Previously `aptos init` would say the account was funded successfully without actually waiting for and checking the success of the transactions. This PR makes it wait and check.

### Test Plan
See https://github.com/aptos-labs/aptos-core/pull/7561.

### DIFF
--- a/crates/aptos/src/common/init.rs
+++ b/crates/aptos/src/common/init.rs
@@ -7,7 +7,7 @@ use crate::common::{
         ConfigSearchMode, EncodingOptions, PrivateKeyInputOptions, ProfileConfig, ProfileOptions,
         PromptOptions, RngArgs, DEFAULT_PROFILE,
     },
-    utils::{fund_account, prompt_yes_with_override, read_line},
+    utils::{fund_account, prompt_yes_with_override, read_line, wait_for_transactions},
 };
 use aptos_crypto::{ed25519::Ed25519PrivateKey, PrivateKey, ValidCryptoMaterialStringExt};
 use aptos_rest_client::{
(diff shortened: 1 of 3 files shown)


crates/aptos/src/common/init.rs
```diff
@@ -211,13 +215,28 @@ impl CliCommand<()> for InitTool {
...
+                for hash in hashes {
+                    client
+                        .wait_for_transaction_by_hash(
+                            hash.into(),
+                            sys_time,
+                            Some(Duration::from_secs(60)),
+                            None,
+                        )
+                        .await?;
+                }
```
gregnazario, line 239: Can you just put this whole thing in one function, so the behavior is the same?

banool: Done done.

0xmaayan (approved):
that means I (or sdk) dont need to `waitForTransaction` anymore?

banool:
Nope, this just affects the CLI.

crates/aptos/src/common/utils.rs
```diff
@@ -372,6 +372,29 @@ pub async fn fund_account(
...
+        .duration_since(SystemTime::UNIX_EPOCH)
+        .map_err(|e| CliError::UnexpectedError(e.to_string()))?
+        .as_secs()
+        + 30;
+    for hash in hashes {
+        client
+            .wait_for_transaction_by_hash(
+                hash.into(),
+                sys_time,
+                Some(Duration::from_secs(60)),
```
gregnazario, line 390: For future reference, probably should call this wait for faucet transactions.

gregnazario (approved):

PR #7567, opened 2023-04-04
Change the prometheus bucket sizes for histograms for execution metrics

vusirikala:
### Description
This PR is for issue [BLK-459](https://linear.app/aptoslabs/issue/BLK-459/[consensus]-fix-histogram-buckets-for-prometheus-metrics). Earlier, the buckets for prometheus metrics are not properly configured. Many times, all the data points fall in a couple of buckets, and rest of the buckets are empty. In this PR, I changed the bucket boundaries for prometheus metrics related to execution.


consensus/src/counters.rs
```diff
@@ -684,13 +696,19 @@ pub static BUFFER_MANAGER_RETRY_COUNT: Lazy<IntCounter> = Lazy::new(|| {
     .unwrap()
 });
 
+const PROPSER_ELECTION_DURATION_BUCKETS: [f64; 17] = [
+    0.001, 0.002, 0.003, 0.004, 0.006, 0.008, 0.01, 0.012, 0.014, 0.175, 0.02, 0.025, 0.05, 0.25,
```
ibalajiarun, line 700: typo? 

```suggestion
    0.001, 0.002, 0.003, 0.004, 0.006, 0.008, 0.01, 0.012, 0.014, 0.0175, 0.02, 0.025, 0.05, 0.25,
```

consensus/src/counters.rs
```diff
@@ -507,13 +507,19 @@ pub static NUM_BLOCKS_IN_PIPELINE: Lazy<IntGaugeVec> = Lazy::new(|| {
 //     .unwrap()
 // });
 
+const NUM_CONSENSUS_TRANSACTIONS_BUCKETS: [f64; 20] = [
+    5.0, 10.0, 20.0, 40.0, 75.0, 100.0, 125.0, 150.0, 175.0, 200.0, 250.0, 300.0, 350.0, 400.0,
```
ibalajiarun, line 511: Do we need 250, 350, 450 buckets? Also, with Quorum Store, we might have up to 20k transactions per block, so we need to accommodate for that here. Its very possible we will hit that number often in forge at least.

aptos-move/aptos-vm/src/counters.rs
```diff
@@ -3,20 +3,24 @@
...
 
 use aptos_metrics_core::{
-    exponential_buckets, register_histogram, register_int_counter, register_int_counter_vec,
+    histogram_opts, register_histogram, register_int_counter, register_int_counter_vec,
     register_int_gauge, Histogram, IntCounter, IntCounterVec, IntGauge,
 };
 use once_cell::sync::Lazy;
 
+const BLOCK_EXECUTION_TIME_BUCKETS: [f64; 17] = [
+    0.20, 0.30, 0.40, 0.50, 0.60, 0.70, 0.80, 0.90, 1.0, 1.25, 1.5, 1.75, 2.0, 3.0, 4.0, 5.0, 10.0,
```
ibalajiarun, line 12: May be we dont need the last one. 10 sec per blk seems too long and we do have per-block gas limit change on the way.

```suggestion
    0.20, 0.30, 0.40, 0.50, 0.60, 0.70, 0.80, 0.90, 1.0, 1.25, 1.5, 1.75, 2.0, 3.0, 4.0, 5.0,
```

aptos-move/aptos-vm/src/counters.rs
```diff
@@ -27,15 +31,20 @@ pub static BLOCK_EXECUTOR_CONCURRENCY: Lazy<IntGauge> = Lazy::new(|| {
     .unwrap()
 });
 
+const SIGNATURE_VERIFICATION_TIME_BUCKETS: [f64; 18] = [
+    0.0002, 0.0004, 0.0006, 0.0008, 0.001, 0.00125, 0.0015, 0.002, 0.0025, 0.003, 0.0035, 0.004,
+    0.005, 0.006, 0.008, 0.01, 0.015, 0.02,
+];
+
```
ibalajiarun, line 38: If this isn't becoming a problem, I would leave it as the default. Let's check the grafana links from the e2e test after they are done.

aptos-move/aptos-vm/src/counters.rs
```diff
@@ -78,30 +87,55 @@ pub static SYSTEM_TRANSACTIONS_EXECUTED: Lazy<IntCounter> = Lazy::new(|| {
...
-        "Validation time per user transaction"
-    )
-    .unwrap()
+        "Validation time per user transaction",
+        TRANSACTION_VALIDATION_TIME_BUCKETS.to_vec()
+    );
+    register_histogram!(histogram_opts).unwrap()
 });
 
+const TXN_GAS_USAGE_BUCKETS: [f64; 10] = [2.0, 3.0, 4.0, 5.0, 6.0, 7.0, 8.0, 9.0, 10.0, 11.0];
```
ibalajiarun, line 132: Please check with move team on this. I believe we might need to accommodate higher gas usage. probably @vgao1996

ibalajiarun (commented):
Let's check the grafana dashboards from the e2e test on this PR once they are done, and make sure we dont have any regressions. Manual testing for this one, unfortunately.

aptos-move/aptos-vm/src/counters.rs
ibalajiarun: I think thats because the max bucket is 10 so all the values greater than 10 is counted in that last bucket. See [here](https://aptoslabs.grafana.net/explore?orgId=1&left=%7B%22datasource%22:%22fHo-R604z%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22datasource%22:%7B%22type%22:%22prometheus%22,%22uid%22:%22fHo-R604z%22%7D,%22editorMode%22:%22code%22,%22expr%22:%22rate(aptos_vm_txn_gas_usage_sum%5B$__rate_interval%5D)/rate(aptos_vm_txn_gas_usage_count%5B$__rate_interval%5D)%22,%22legendFormat%22:%22__auto%22,%22range%22:true,%22instant%22:true%7D%5D,%22range%22:%7B%22from%22:%22now-1h%22,%22to%22:%22now%22%7D%7D) for average.

vusirikala: Gotcha. I updated the bucket sizes accordingly.

aptos-move/aptos-vm/src/counters.rs
```diff
@@ -3,20 +3,24 @@
...
         // metric name
         "block_executor_execute_block_seconds",
         // metric description
         "The time spent in seconds for executing a block in block executor",
-        exponential_buckets(/*start=*/ 1e-3, /*factor=*/ 2.0, /*count=*/ 20).unwrap(),
-    )
-    .unwrap()
+        BLOCK_EXECUTION_TIME_BUCKETS.to_vec()
+    );
+    register_histogram!(histogram_opts).unwrap()
```
zekun000, line 23: why do we need this change? let and then unwrap?

consensus/src/counters.rs
```diff
@@ -526,14 +533,20 @@ pub static BLOCK_TRACING: Lazy<HistogramVec> = Lazy::new(|| {
     .unwrap()
 });
 
+const CONSENSUS_WAIT_DURATION_BUCKETS: [f64; 17] = [
+    0.02, 0.04, 0.06, 0.08, 0.10, 0.125, 0.15, 0.175, 0.2, 0.225, 0.25, 0.3, 0.4, 0.6, 0.8, 2.0,
```
zekun000, line 537: is the unit second here? if so this might be too large, most of time we see less than 10ms wait duration (this is mostly time drift between nodes)

aptos-move/aptos-vm/src/counters.rs
```diff
@@ -78,30 +82,59 @@ pub static SYSTEM_TRANSACTIONS_EXECUTED: Lazy<IntCounter> = Lazy::new(|| {
     .unwrap()
 });
 
+const NUM_BLOCK_TRANSACTIONS_BUCKETS: [f64; 30] = [
```
ibalajiarun, line 85: I feel this is too many. May be reduce the number of buckets below 100 txns? And, cap to 30k for now, until we hit 50k? This should get us down to 25 buckets.

ibalajiarun (approved):
LGTM overall. Please address feedback below. Thanks!

aptos-move/aptos-vm/src/counters.rs
JoshLind: +1. My recommendation would be to try and limit each one two 20 buckets. More can be used, but it seems like it would require further justification 😄 Otherwise, displaying these things will become a nightmare 😱

consensus/src/counters.rs
vusirikala: Yes. The unit is seconds. I added more buckets with lower time now. Balaji mentioned that we are currently facing problem with properly quantifying the anomalies, and we don't care about non-anomalies that much. I found some anomalies as much as 0.8 second when I looked into grafana for last 6 months. So, I added more points on the higher end.

aptos-move/aptos-vm/src/counters.rs
vusirikala: I picked up the syntax from some PR. I changed it to make it a single line code.

aptos-move/aptos-vm/src/counters.rs
vusirikala: Reduced the number of buckets now.

PR #7568, opened 2023-04-04
reuse move-stdlib sha2 gas parameters in aptos-stdlib

zjma:
SHA2-256 will be used in the new algebra module in aptos-stdlib, and it would be great to reuse the gas parameters defined in move-stdlib.


aptos-move/aptos-vm/src/natives.rs
```diff
@@ -35,11 +35,14 @@ pub fn aptos_natives(
...
     features: Arc<Features>,
 ) -> NativeFunctionTable {
+    let sha256_gas_params: move_stdlib::natives::hash::Sha2_256GasParameters =
+        gas_params.move_stdlib.hash.sha2_256.clone();
     move_stdlib::natives::all_natives(CORE_CODE_ADDRESS, gas_params.move_stdlib)
         .into_iter()
         .filter(|(_, name, _, _)| name.as_str() != "vector")
         .chain(aptos_framework::natives::all_natives(
             CORE_CODE_ADDRESS,
+            sha256_gas_params,
```
alinush, line 45: You could've passed all the Move gas parameters here because it's likely that, down the road, we will need those other parameters too.

PR #7570, opened 2023-04-04
dev_setup: better mac-arm support, big deps become optional

brianolson:
### Description

Improve Mac-ARM support getting more native packages.
Move PostgreSQL, node.js, pnpm, and solidity to optional installs.

### Test Plan
`cargo test` and `cargo test -- --skip prover` work locally (except for known broken doc tests)


scripts/dev_setup.sh
```diff
@@ -705,7 +727,7 @@ Build tools (since -t or no option was provided):
   * lcov
   * pkg-config
   * libssl-dev
-  * NodeJS / NPM
+#  * NodeJS / NPM
```
JoshLind, line 730: Can we remove this entirely?

scripts/dev_setup.sh
```diff
@@ -813,6 +852,12 @@ while getopts "btoprvydsah:i:n" arg; do
     a)
       INSTALL_API_BUILD_TOOLS="true"
       ;;
+    P)
```
JoshLind, line 855: Annoying that we've run out of lower-case letters, but these should do 😄

JoshLind (commented):
Thanks @brianolson. Looks reasonable to me. Will wait for someone more familiar to comment 😄

scripts/dev_setup.sh
```diff
@@ -932,14 +979,10 @@ if [[ "$INSTALL_BUILD_TOOLS" == "true" ]]; then
...
   install_cargo_nextest
   install_grcov
-  install_postgres
   install_pkg git "$PACKAGE_MANAGER"
   install_lcov "$PACKAGE_MANAGER"
-  install_nodejs "$PACKAGE_MANAGER"
-  install_pnpm "$PACKAGE_MANAGER"
   install_pkg unzip "$PACKAGE_MANAGER"
   install_protoc
-  install_solidity
```
davidiw, line 942: postgres was used for the old indexer
nodejs was likely used for some old legacy
pnpm is only required for sdks
solidity is currently required for testing move, afaiu -- so it is probably required -- question I have is that something we intend to continue to support or should we consider sun setting it, don't know if we'll run into any friction in future development...

davidiw:
https://github.com/aptos-labs/aptos-core/actions/runs/4613807097/jobs/8156172029?pr=7570

you're currently failing because you removed solidity
I would just remove postgres, asking explicitly to install that is weird
I would also remove the ts sdk stuff

brianolson:
I added back solidity by adding it to the dev_setup.sh options used in CI
I also ran `shellcheck` locally and it noted a bunch of dead code. b433b967742b8bd108c09a43368d1b052257ee00 deleted the references to several install_* clauses but not that code, so I dropped it now.

banool (commented):
I looked at all places where we refer to `dev_setup.sh` today and it seems like we won't be missing any flags as a result of this. Saying that, some specific things like the TS SDK or the indexer work out of the box with the current `dev_setup.sh` (which is likely why they don't mention it) and now they won't. In places where we have removed default support we should say something in the README like:

To set up your development environment run this from the root:
```
scripts/dev_setup.sh -Jpfsdfsdfdsfdsf
```

With the appropriate flags.

scripts/dev_setup.sh
brianolson: I added back node.js, pnpm, and solidity in the CI setup scripts so that it passes tests.

brianolson:
I think everything passes that should pass. I'm not sure where to document, "but if you want to do TypeScript, run `scripts/dev_setup.sh -J`; and if you want to do Indexer, you should install PostgreSQL by your own favorite source or by `scripts/dev_setup.sh -P`"

davidiw:
@banool 

My recollection was that in general for those, we have recommendations on explicitly how to setup and don't use `dev_setup.sh`. Personally, I prefer that for lighter packages, we don't mention `dev_setup.sh` because it is (excessively) overweight.

Reading your response, I can't tell if you have an opinion or are just stating objectively a fact.

davidiw:
@brianolson , thanks for doing this!

One thing you can do, is grep through the doc site: https://github.com/aptos-labs/aptos-core/tree/main/developer-docs-site

for dev_setup and verify the context in which it is being used

clay-aptos:
> I looked at all places where we refer to `dev_setup.sh` today and it seems like we won't be missing any flags as a result of this. Saying that, some specific things like the TS SDK or the indexer work out of the box with the current `dev_setup.sh` (which is likely why they don't mention it) and now they won't. In places where we have removed default support we should say something in the README like:
> 
> To set up your development environment run this from the root:
> 
> ```
> scripts/dev_setup.sh -Jpfsdfsdfdsfdsf
> ```
> 
> With the appropriate flags.

@brianolson , can you help with those updates? See:
https://aptos.dev/community/site-updates

If not, please let me know and I will gladly help.

banool:
I don't mind yeah, just stating the way it is. Though if we want to define our own setup for process for these other projects, we should consider removing that stuff from this setup script. For example, removing stuff related to the TS SDK from the script. That can happen in another PR.

brianolson:
@clay-aptos  There isn't one place in the docs that I think we should add "and if you want to develop the TypeScript SDK, run `dev_setup.sh -J`". The pages at `developer-docs-site/docs/sdks/ts-sdk/` talk about using published versions through npm/yarn/pnpm. Maybe anyone working on the TS SDK will already have those tools installed? What I **did** find is that `find developer-docs-site -type f |xargs grep -l dev_setup` finds **11** files, and (outside of the scope of this PR) it would be great to consolidate all those to one "Building Aptos" page and linking to it from the other 10 places. Such a "Building Aptos" page could have the proviso about doing TS SDK work. But all of the other flows currently describing dev_setup.sh have only a narrower goal to run a node or something.

davidiw (approved):

PR #7576, opened 2023-04-05
[TS SDK] add User-Agent header to axios call on IndexerClient

0xjinn:
### Description

Add User-Agent to the axios call on IndexerClient. Similar to PR where we had added to AptosClient - https://github.com/aptos-labs/aptos-core/pull/7501

### Test Plan
1. Spin up the local testnet: cargo run -p aptos -- node run-local-testnet --force-restart --assume-yes --with-faucet
2. pnpm test

```
Test Suites: 20 passed, 20 total
Tests:       187 passed, 187 total
Snapshots:   0 total
Time:        142.57 s
```

3. pnpm run lint


ecosystem/typescript/sdk/src/providers/indexer.ts
```diff
@@ -48,6 +49,10 @@ interface PaginationArgs {
   limit?: number;
 }
 
+const headers = {
+  "User-Agent": `aptos-indexer/${VERSION}`,
```
0xmaayan, line 53: let's use here the same header as in the rest api - `aptos-ts-sdk/${VERSION}`

ecosystem/typescript/sdk/src/tests/e2e/indexer.test.ts
```diff
@@ -191,5 +193,20 @@ describe("Indexer", () => {
       const ledgerInfo = await indexerClient.getIndexerLedgerInfo();
       expect(ledgerInfo.ledger_infos[0].chain_id).toBeGreaterThan(1);
     });
+
+    test("make sure queryIndexer post call with correct User-Agent header", async () => {
+      let mockHeaders: any = {};
+
+      // spy on axios.post and mock the response
+      jest.spyOn(axios, 'post').mockImplementation((url, query, config) => {
```
0xmaayan, line 201: can we test it for when it does a real call instead of mocking it?

davidiw (commented):
nice, can we do the same for the Python sdk :pray:

ecosystem/typescript/sdk/src/tests/e2e/indexer.test.ts
0xjinn: Yep, but I wonder though, if we should do the whole integration test or just checking the header would be sufficient?

0xjinn:
Created an issue to track adding user-agent on Python and Rust SDK - https://github.com/aptos-labs/aptos-core/issues/7581

ecosystem/typescript/sdk/src/tests/e2e/indexer.test.ts
gregnazario: I don't think we have to do active testing.  Please test it once with a real world situation and show it, then mocking should be fine, as long as the headers show the same.

0xjinn: > I don't think we have to do active testing. Please test it once with a real world situation and show it, then mocking should be fine, as long as the headers show the same.

Wanted to make sure I understand this correctly. Basically stay with mocking,  but also manually test it with real call and and show/record the result?

ecosystem/typescript/sdk/src/providers/indexer.ts
```diff
@@ -48,6 +49,10 @@ interface PaginationArgs {
   limit?: number;
 }
 
+const headers = {
+  "User-Agent": `aptos-ts-sdk/${VERSION}`,
```
banool, line 53: Since we're using the same string in both cases, can you make a function in some `utils` file that returns this `aptos-ts-sdk/${VERSION}` string?

0xjinn: Good call!

ecosystem/typescript/sdk/src/providers/Utils.ts
```diff
@@ -0,0 +1,7 @@
+import { VERSION } from "../version";
```
0xmaayan, line 1: can we lowercase the file name? `Utils.ts` => `utils.ts`

0xjinn: Yep! But curious, is lowercase file name, the naming convention on Typescript?

0xmaayan: yep

ecosystem/typescript/sdk/src/tests/e2e/indexer.test.ts
0xjinn: Update - Manually checked and make sure that `User-Agent` is included from the request.

gregnazario (approved):

PR #7578, opened 2023-04-05
[Storage][Sharding] Split tree commit into multiple batches.

grao1991:

### DIFF
--- a/storage/aptosdb/src/state_merkle_db.rs
+++ b/storage/aptosdb/src/state_merkle_db.rs
@@ -2,14 +2,17 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use crate::{
+    db_metadata::{DbMetadataKey, DbMetadataSchema, DbMetadataValue},
     db_options::{gen_state_merkle_cfds, state_merkle_db_column_families},
     lru_node_cache::LruNodeCache,
     metrics::NODE_CACHE_SECONDS,
+    new_sharded_merkle_schema_batch,
     schema::jellyfish_merkle_node::JellyfishMerkleNodeSchema,
     stale_node_index::StaleNodeIndexSchema,
     stale_node_index_cross_epoch::StaleNodeIndexCrossEpochSchema,
+    utils::truncation_helper::{get_state_merkle_commit_progress, truncate_state_merkle_db_shards},
     versioned_node_cache::VersionedNodeCache,
-    NUM_STATE_SHARDS, OTHER_TIMERS_SECONDS,
+    ShardedStateMerkleSchemaBatch, NUM_STATE_SHARDS, OTHER_TIMERS_SECONDS,
 };
 use anyhow::Result;
 use aptos_config::config::{RocksdbConfig, RocksdbConfigs};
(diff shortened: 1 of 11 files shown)


storage/aptosdb/src/state_merkle_db.rs
```diff
@@ -140,6 +154,33 @@ impl StateMerkleDb {
...
+        top_levels_batch: SchemaBatch,
+        sharded_state_merkle_batches: ShardedStateMerkleSchemaBatch,
+    ) -> Result<()> {
+        TREE_COMMIT_POOL.scope(|s| {
+            let mut batches = sharded_state_merkle_batches.into_iter();
+            for shard_id in 0..NUM_STATE_SHARDS {
+                let state_merkle_batch = batches.next().unwrap();
+                s.spawn(move |_| {
+                    self.commit_single_shard(version, shard_id as u8, state_merkle_batch)
+                        .unwrap_or_else(|_| panic!("Failed to commit shard {shard_id}."));
```
msmouse, line 169: nit: equivalent to `expect()`

grao1991: linter complains it because I need format! the error string.

storage/aptosdb/src/state_merkle_db.rs
```diff
@@ -293,7 +308,14 @@ impl StateMerkleDb {
                 .collect::<Vec<_>>()
                 .par_iter()
                 .with_min_len(128)
-                .map(|(node_key, node)| batch.put::<JellyfishMerkleNodeSchema>(node_key, node))
+                .map(|(node_key, node)| {
```
areshand, line 311: what is the 128? might be worth of creating a const for it?

storage/aptosdb/src/state_merkle_db.rs
```diff
@@ -280,7 +294,8 @@ impl StateMerkleDb {
             );
         }
 
-        let batch = SchemaBatch::new();
+        let top_levels_batch = SchemaBatch::new();
```
areshand, line 297: nit: might add some comment on why this is in its own batch

storage/aptosdb/src/state_store/state_snapshot_committer.rs
```diff
@@ -75,7 +75,7 @@ impl StateSnapshotCommitter {
                     let version = delta_to_commit.current_version.expect("Cannot be empty");
                     let base_version = delta_to_commit.base_version;
 
-                    let (batch, root_hash) = self
+                    let (top_levels_batch, sharded_batch, root_hash) = self
```
areshand, line 78: maybe consider having a new struct wrapping top-level batch and sharded_batches, which appear together?

grao1991: It's likely the current implementation is just a temporary solution. I'm considering to make the data sharded at an much earlier stage, to get better parallelism.

storage/aptosdb/src/state_merkle_db.rs
grao1991: Done

storage/aptosdb/src/state_merkle_db.rs
grao1991: Just a random number that is not too small. Didn't spend time on tuning this.

areshand (approved):

PR #7596, opened 2023-04-05
[Configs] Remove unnecessary dependency.

JoshLind:
### Description

This PR removes an unnecessary dependency on `PersistableConfig` from the rosetta smoke tests.

### Test Plan
Existing test infrastructure.

### DIFF
--- a/testsuite/smoke-test/src/rosetta.rs
+++ b/testsuite/smoke-test/src/rosetta.rs
@@ -9,10 +9,7 @@ use aptos::{
     test::{CliTestFramework, INVALID_ACCOUNT},
 };
 use aptos_cached_packages::aptos_stdlib;
-use aptos_config::{
-    config::{ApiConfig, PersistableConfig},
-    utils::get_available_port,
-};
+use aptos_config::{config::ApiConfig, utils::get_available_port};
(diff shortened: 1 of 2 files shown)


testsuite/smoke-test/src/rosetta.rs
```diff
@@ -1339,8 +1336,10 @@ async fn parse_operations(
                         native_coin(),
                         "Balance should be the native coin"
                     );
-                    let delta =
-                        u64::parse(&amount.value).expect("Should be able to parse amount value");
+                    let delta = amount
+                        .value
+                        .parse::<u64>()
+                        .expect("Should be able to parse amount value");
```
gregnazario, line 1342: FYI changes like this should probably be in a separate commit.

gregnazario (approved):

testsuite/smoke-test/src/rosetta.rs
JoshLind: These are the changes 😆 The rosetta smoke tests incorrectly relied on the `parse` method from the `PersistentConfig`, so I've cut that dependency and updated the parsing logic 😄

moonclavedev (approved):

PR #7598, opened 2023-04-05
Correct erroneous module bundle comments

alnoki:
Module bundles are not the only way to publish modules. Entry functions can publish modules too.


alnoki:
@davidiw

davidiw (approved):
these are dead / deprecated functions, we can't remove them because we still have a handful of tests that use them... one weekend in the future... I will have fun

PR #7599, opened 2023-04-05
ignore an example for `cargo test --doc` to pass

brianolson:
### Description

Code example would require unreasonable amount of context to actually work, ignore it so that `cargo test --doc` can pass.

### Test Plan

`cargo test --doc` passes now

### DIFF
--- a/ecosystem/node-checker/src/provider/helpers.rs
+++ b/ecosystem/node-checker/src/provider/helpers.rs
@@ -9,7 +9,7 @@ pub const MISSING_PROVIDER_MESSAGE: &str = "Incomplete request";
 ///
 /// Example invocations:
 ///
-/// ```
+/// ```ignore
 /// # use std::io;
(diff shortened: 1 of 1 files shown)


JoshLind (approved):
Nice, thanks @brianolson! 😄 Seems like a larger rewrite would be needed. Ignore sounds good to me.

grao1991 (approved):

ecosystem/node-checker/src/provider/helpers.rs
```diff
@@ -9,7 +9,7 @@ pub const MISSING_PROVIDER_MESSAGE: &str = "Incomplete request";
 ///
 /// Example invocations:
 ///
-/// ```
+/// ```ignore
```
gregnazario, line 12: text would also work

gregnazario (approved):

PR #7600, opened 2023-04-06
[deletion refund] track storage slot payer

msmouse:
### Description
If enabled, start tracking the transaction sender as the storage slot payer on slot allocation.
Governed by a new feature flag.

-----
it's impossible to change the timestamp in a transaction with a sender
to be tracked as a slot payer currently. (and impossible to test)

keeping the code in the commit history just in case. (separate commit to remove that part)

### Test Plan
unit tests


aptos-move/aptos-vm/src/move_vm_ext/session.rs
```diff
@@ -415,17 +415,31 @@ where
...
-                        MoveStorageOp::New(bytes)
-                    } else {
-                        MoveStorageOp::Modify(bytes)
+                    let move_op = match write_set_mut.get(&state_key) {
+                        None => MoveStorageOp::New(bytes),
+                        Some(direct_write) => match direct_write {
+                            WriteOp::Creation(_) | WriteOp::CreationWithMetadata { .. } => {
+                                MoveStorageOp::New(bytes)
+                            },
+                            // FIXME(alden): is this an error?
```
msmouse, line 424: @georgemitenkov  @gelash 
1. Is it true that when an aggregator is first created, there will be a Create from the table change set and then a modify here from the aggregator change set?
2. is that the only known case to you wrt aggregators that a state key is repeated in the (outer) change set? 

Any other concerns?

aptos-move/aptos-vm/src/move_vm_ext/session.rs
```diff
@@ -257,22 +353,36 @@ impl<'r, 'l> SessionExt<'r, 'l> {
     ) -> Result<ChangeSetExt, VMStatus> {
         let mut write_set_mut = WriteSetMut::new(Vec::new());
         let mut delta_change_set = DeltaChangeSet::empty();
+        let mut new_slot_metadata: Option<StateValueMetadata> = None;
+        if is_storage_slot_metadata_enabled {
+            if let Some(payer) = new_slot_payer {
+                if let Some(current_time) = current_time {
```
grao1991, line 359: Do we really want to silently ignore the cases when payer or current_time is None?

aptos-move/aptos-vm/src/move_vm_ext/session.rs
```diff
@@ -368,3 +458,101 @@ impl<'r, 'l> DerefMut for SessionExt<'r, 'l> {
...
+            .get_state_value(state_key)
+            .map_err(|_| VMStatus::Error(StatusCode::STORAGE_ERROR, None))?;
+        let data = serialize(&value);
+
+        let op = match existing_value_opt {
+            None => {
+                match &self.new_slot_metadata {
+                    // n.b. Aggregator writes historically did not distinguish Create vs Modify.
+                    None => WriteOp::Modification(data),
+                    Some(metadata) => WriteOp::CreationWithMetadata {
```
grao1991, line 544: What will happen if multiple "creates" from different txns that have different metadata?

msmouse: This is after the aggregators are materialized, there shouldn't be multiple creates between txns.

aptos-move/aptos-vm/src/move_vm_ext/vm.rs
```diff
@@ -90,6 +93,7 @@ impl MoveVmExt {
         extensions.add(AlgebraContext::new());
         extensions.add(NativeAggregatorContext::new(txn_hash, remote));
 
+        let sender_opt = session_id.sender();
```
msmouse, line 96: @grao1991  where the payer is got

grao1991 (approved):

aptos-move/aptos-vm/src/move_vm_ext/session.rs
```diff
@@ -180,6 +227,12 @@ impl<'r, 'l> SessionExt<'r, 'l> {
                         .add_resource_op(struct_tag, blob_op)
                         .map_err(|_| common_error.clone())?;
                 } else {
+                    if addr == CORE_CODE_ADDRESS
+                        && CurrentTimeMicroseconds::struct_tag() == struct_tag
+                    {
+                        updated_timestamp = Self::parse_updated_timestamp(&blob_op)
```
zekun000, line 233: I don't think this piece belongs to this function, at least rename the function

aptos-move/aptos-vm/src/move_vm_ext/session.rs
```diff
@@ -257,22 +353,36 @@ impl<'r, 'l> SessionExt<'r, 'l> {
     ) -> Result<ChangeSetExt, VMStatus> {
         let mut write_set_mut = WriteSetMut::new(Vec::new());
         let mut delta_change_set = DeltaChangeSet::empty();
+        let mut new_slot_metadata: Option<StateValueMetadata> = None;
+        if is_storage_slot_metadata_enabled {
+            if let Some(payer) = new_slot_payer {
+                if let Some(current_time) = current_time {
+                    new_slot_metadata = Some(StateValueMetadata::new(payer, 0, current_time));
```
zekun000, line 360: the single payer pays for all slots created in the txn?

aptos-move/aptos-vm/src/move_vm_ext/session.rs
```diff
@@ -368,3 +458,101 @@ impl<'r, 'l> DerefMut for SessionExt<'r, 'l> {
...
+        let write_op = match (existing_value_opt, move_storage_op) {
+            (None, Modify(_) | Delete) => {
+                return Err(VMStatus::Error(
+                    StatusCode::UNKNOWN_INVARIANT_VIOLATION_ERROR,
+                    err_msg("When converting write op: updating non-existent value."),
+                ))
+            },
+            (Some(_), New(_)) => {
+                return Err(VMStatus::Error(
+                    StatusCode::UNKNOWN_INVARIANT_VIOLATION_ERROR,
```
zekun000, line 493: if we want to alert on unknown invariant violation, we can't use it like this. this is possible similarly with parallel execution and duplicate txn

aptos-move/aptos-vm/src/move_vm_ext/session.rs
zekun000: I don't think aggregator creation will be here iirc, aggregator creation should go to normal writeset. i.e. None on existing_value should be impossible?

also the word "materialized" is confusing, I think you mean after aggregators are created.

aptos-move/aptos-vm/src/move_vm_ext/session.rs
```diff
@@ -243,10 +296,53 @@ impl<'r, 'l> SessionExt<'r, 'l> {
...
+        }
+    }
+
+    fn get_current_timestamp(
+        updated_timestamp: Result<Option<CurrentTimeMicroseconds>, ()>,
+        remote: &dyn MoveResolverExt,
+    ) -> Option<CurrentTimeMicroseconds> {
+        match updated_timestamp {
+            Ok(Some(timestamp)) => Some(timestamp),
+            Ok(None) => Some(1010),// CurrentTimeMicroseconds::fetch_config(remote),
```
zekun000, line 336: why all the corpus have current_timestamp_usecs: 0 instead of the hard coded 1010 here?

aptos-move/aptos-vm/src/move_vm_ext/session.rs
zekun000: I thought payer should be new_slot_payer.unwrap_or(txn_sender)?

aptos-move/aptos-vm/src/move_vm_ext/session.rs
msmouse: yes

aptos-move/aptos-vm/src/move_vm_ext/session.rs
msmouse: the last commit is to debug performance regression (which turned out to be within the margin of error), tests were not updated, see the part after "//" for real code.

aptos-move/aptos-vm/src/move_vm_ext/session.rs
msmouse: This function is to replace:
https://github.com/aptos-labs/aptos-core/pull/7600#discussion_r1190338606

And this special treatment fixed tests `None => WriteOp::Modification(data),`

let me remove this and see again.

aptos-move/aptos-vm/src/move_vm_ext/session.rs
```diff
@@ -194,7 +194,7 @@ impl<'r, 'l> SessionExt<'r, 'l> {
     ///   * If group or data does't exist, Unreachable
     ///   * If elements remain, Modify
     ///   * Otherwise delete
-    fn split_and_merge_resource_groups(
+    fn process_resource_group_and_timestamp(
```
zekun000, line 197: need to update comment too

zekun000 (approved):

msmouse:
it's impossible to change the timestamp in a transaction with a sender
to be tracked as a slot payer currently. (and impossible to test)

keeping the code in the commit history just in case.

PR #7601, opened 2023-04-06
consensus and mempool get max_receiving_messages config

brianolson:
### Description

add `max_receiving_messages` config to consensus and mempool sections to set the inbound message queue size.

### Test Plan

N/A ? Aside from a test setting these to 1 and testing that a system becomes terrible, I'm not thinking of a meaningful test.


config/src/config/mempool_config.rs
```diff
@@ -29,6 +29,8 @@ pub struct MempoolConfig {
     pub broadcast_buckets: Vec<u64>,
     pub eager_expire_threshold_ms: Option<u64>,
     pub eager_expire_time_ms: u64,
+    // length of inbound queue of messages
+    pub max_receiving_messages: usize,
```
JoshLind, line 33: nit: can we keep this alphabetical (and the same for the ConsensusConfig)? I know some of them aren't, but I think they're wrong/lazy :P

config/src/config/consensus_config.rs
```diff
@@ -57,6 +57,8 @@ pub struct ConsensusConfig {
     // must match one of the CHAIN_HEALTH_WINDOW_SIZES values.
     pub window_for_chain_health: usize,
     pub chain_health_backoff: Vec<ChainHealthBackoffValues>,
+    // length of inbound queue of messages
+    pub max_receiving_messages: usize,
```
JoshLind, line 61: Nice! Can we name them `max_network_channel_size`, so that they match the one already defined by state sync? https://github.com/aptos-labs/aptos-core/blob/7088dd699018e5e2ffa4f85b3336df69d10941b3/config/src/config/state_sync_config.rs#L116

JoshLind (commented):
Looks good, thanks @brianolson 😄

JoshLind (approved):

JoshLind:
Thanks @brianolson! 😄

igor-aptos (approved):

grao1991 (approved):

zekun000 (approved):

PR #7603, opened 2023-04-06
[framework] improve token mint, burn tracking

davidiw:
* Remove untracked collections from being public usage for now
* Add a mint and burn event for fixed supply
* Add tracking of both the collection id and the token's address to the events
* Add an unlimited supply to replace untracked
* Add support for minting deletable tokens

it is worth noting that untracked needs more thought:
* How to make them deletable as guid isn't parallelizable
* How to add events or create collection_ids

### DIFF
--- a/aptos-move/framework/aptos-framework/sources/object.move
+++ b/aptos-move/framework/aptos-framework/sources/object.move
@@ -152,6 +152,11 @@ module aptos_framework::object {
         Object<T>{ inner: object }
     }
 
+    /// Returns true if there exists an object or the remnants of an object.
+    public fun is_object(object: address): bool {
+        exists<ObjectCore>(object)
+    }
+
     /// Derives an object address from source material: sha3_256([creator address | seed | 0xFE]).
     public fun create_object_address(source: &address, seed: vector<u8>): address {
         let bytes = bcs::to_bytes(source);
--- a/aptos-move/framework/aptos-framework/sources/object.move
+++ b/aptos-move/framework/aptos-framework/sources/object.move
@@ -173,6 +173,14 @@ module aptos_framework::object {
         from_bcs::to_address(hash::sha3_256(bytes))
     }
 
+    /// Derives an object from an Account GUID.
+    public fun create_guid_object_address(source: address, creation_num: u64): address {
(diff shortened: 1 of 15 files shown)


aptos-move/framework/aptos-token-objects/sources/collection.move
```diff
@@ -103,15 +124,20 @@ module aptos_token_objects::collection {
 
     /// Creates an untracked collection, or a collection that supports an arbitrary amount of
     /// tokens. This is useful for mass airdrops that fully leverage Aptos parallelization.
-    public fun create_untracked_collection(
+    /// TODO: Hide this until we bring back meaningful way to enforce burns
+    fun create_untracked_collection(
```
movekevin, line 128: This breaks the hero example. Also why not delete this entirely for now?

movekevin (approved):

aptos-move/framework/cached-packages/build.rs
```diff
@@ -12,6 +12,14 @@ fn main() {
...
         let mut prev_dir = current_dir;
         prev_dir.pop();
+        println!(
+            "cargo:rerun-if-changed={}",
+            prev_dir.join("aptos-token-objects").join("Move.toml").display()
+        );
+        println!(
+            "cargo:rerun-if-changed={}",
+            prev_dir.join("aptos-token-objects").join("sources").display()
+        );
```
gregnazario, line 22: We should probably make this shorter to update in the future

gregnazario (approved):

aptos-move/framework/aptos-framework/sources/object.move
```diff
@@ -168,6 +173,14 @@ module aptos_framework::object {
         from_bcs::to_address(hash::sha3_256(bytes))
     }
 
+    /// Derives an object from an Account (GU)ID
```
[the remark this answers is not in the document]
davidiw: Because it is GUID and ID because of BCS

aptos-move/framework/cached-packages/build.rs
davidiw: should be a trivial macro

0xmaayan (approved):

PR #7607, opened 2023-04-06
Add Move Prover dev environment support to the Windows script

saharct:
This PR adds support for the Move Prover. I successfully tested this by running Move Prover tests across our supported Windows OSs. Shout out to the MoveBit team for also testing this and providing their valuable feedback!


clay-aptos:
Nice, @saharct ! Adding @junkil-park to review, as well. Thanks!

wrwg (approved):
Wow. This is supercool. 

@junkil-park has asked MoveBit but they basically switched months ago to MacOS because they did not had a good devex on Win with the prover. Perhaps this one here makes things better.

I wonder whether the versions of the tools could be somehow shared between this script and dev_setup.sh. This is something for later.

scripts/windows_dev_setup.ps1
```diff
@@ -336,17 +360,133 @@ function existing_package {
...
+    # Download and extract the 64-bit version of Protoc
+    $cvc5_exe = "$env:USERPROFILE\cvc5-$global:cvc5_version\cvc5-Win64.exe"
+    Write-Host $path
+    Invoke-WebRequest -Uri "https://github.com/cvc5/cvc5/releases/download/cvc5-$global:cvc5_version/cvc5-Win64.exe" -OutFile (New-Item -Path "$cvc5_exe" -Force) -ErrorAction SilentlyContinue
+    [Environment]::SetEnvironmentVariable("CVC5_EXE", "$cvc5_exe", "User") 
+    Write-Host "User environment variables set for CVC5"
+  }
+  elseif ($global:architecture -eq "86") {
+    Write-Host "Only 64-bit systems can install CVC5"
+  }
```
gregnazario, line 385: Do you want to check this first that it's not 64 bit then, check the path rooted?

scripts/windows_dev_setup.ps1
```diff
@@ -336,17 +360,133 @@ function existing_package {
...
+    $uri = "z3-$global:z3_version"
+    $z3_zip = "z3-$global:z3_version-x$global:architecture-win.zip"
+    $z3_filepath = "$env:USERPROFILE\$z3_zip"
+    # Download and extract the 64-bit version of Protoc
+    Invoke-WebRequest -Uri "https://github.com/Z3Prover/z3/releases/download/$uri/$z3_zip" -OutFile (New-Item -Path "$z3_filepath" -Force) -ErrorAction SilentlyContinue
+    if ($global:architecture -eq "64") {
+      while ((Get-Item "$z3_filepath").Length -lt 52MB) {
+        Start-Sleep -Seconds 1
+      }
+    }
```
gregnazario, line 416: This is a download and extract.  Maybe we want to print a `.` or something so there's progress as the user is waiting for the download.  A quick progress bar kind of thing.

I also wonder if there's a way to get these file sizes before downloading, since the size could change over time / this could get out of sync

gregnazario (approved):

scripts/windows_dev_setup.ps1
saharct: @gregnazario, PowerShell already has a built-in progress bar which automatically displays for the `Invoke-WebRequest` and `Expand-Archive` commands. It looks like this:
image

I meant to remove the while loops earlier, as they are unnecessary. `Invoke-WebRequest` and `Expand-Archive` work synchronously by default. They wait for the download/extraction to complete, regardless of size, before moving on to the next step.

saharct:
> Wow. This is supercool.
> 
> @junkil-park has asked MoveBit but they basically switched months ago to MacOS because they did not had a good devex on Win with the prover. Perhaps this one here makes things better.
> 
> I wonder whether the versions of the tools could be somehow shared between this script and dev_setup.sh. This is something for later.

Thanks, @wrwg! I really liked your idea about sharing versions between scripts, so I went ahead and made a first pass at implementing such functionality in [this commit](https://github.com/aptos-labs/aptos-core/pull/7607/commits/d69f4b3312f61702cfa2651a77edcd21f4ff9a8e). Please let me know what you think.

Junkil connected me with the MoveBit team. They kindly offered to test the script and provide feedback. They are eager for us to support this, as the manual setup is a significant pain point for using the Move Prover on Windows.

0xOutOfGas:
> 
@saharct We like this feature, it would definitely bring more prover users. Nice Work!  I'm not sure whether the prover is available on mac m1/m2? If not, do you have any plan to support this?

saharct:
Thanks, @0xOutOfGas! I believe so, but @junkil-park can answer this question in more detail.

wrwg:
> Thanks, @0xOutOfGas! I believe so, but @junkil-park can answer this question in more detail.

Many of us are using the prover on M1/M2 machines, but there might be the rosetta emulator evolved, which one use to install explicitly from Apple.

0xOutOfGas:
> 

Very cool!

PR #7608, opened 2023-04-06
[fungible_assets] fix bugs, add tests, rename and change object api

lightmark:
### Description

1. rename wallet to store. rename FungibleAssetSupport to DeriveRefPod.
2. add tests to primary_store. Also fixed a bug there.
3. change create_user_derived_object to be a friend function w/o signer but use address instead.

### Test Plan
ut


aptos-move/move-examples/fungible_asset/sources/managed_fungible_asset.move
```diff
@@ -26,7 +26,7 @@ module fungible_asset::managed_fungible_asset {
     /// Initialize metadata object and store the refs.
     fun init_module(admin: &signer) {
         let constructor_ref = &object::create_named_object(admin, ASSET_SYMBOL);
-        primary_wallet::create_primary_wallet_enabled_fungible_asset(
+        primary_store::create_primary_wallet_enabled_fungible_asset(
```
movekevin, line 29: This looks broken. There's no create_primary_wallet_enabled_fungible_asset anymore

aptos-move/framework/aptos-framework/sources/primary_store.move
```diff
@@ -0,0 +1,156 @@
...
+    struct DeriveRefPod has key {
+        metadata_derive_ref: DeriveRef,
+    }
+
+    /// Creators of fungible assets can call this to enable support for creating primary (deterministic) stores for
+    /// their users.
+    public fun add_derived_ref_pod(
+        constructor_ref: &ConstructorRef,
+    ) {
+        let metadata_obj = &object::generate_signer(constructor_ref);
```
movekevin, line 19: Why did you remove the call to fungible_asset::add_fungibility?

aptos-move/framework/aptos-framework/sources/object.move
```diff
@@ -190,8 +192,7 @@ module aptos_framework::object {
 
     /// Create a new object whose address is derived based on the creator account address and another object.
     /// Derivde objects, similar to named objects, cannot be deleted.
-    public fun create_user_derived_object(creator: &signer, derive_ref: &DeriveRef): ConstructorRef {
-        let creator_address = signer::address_of(creator);
+    public(friend) fun create_user_derived_object(creator_address: address, derive_ref: &DeriveRef): ConstructorRef {
```
movekevin, line 195: Why?

aptos-move/framework/aptos-framework/sources/primary_store.move
```diff
@@ -0,0 +1,156 @@
+/// This defines the module for interacting with primary stores of accounts/objects, which have deterministic addresses
+module aptos_framework::primary_store {
+    use aptos_framework::fungible_asset::{Self, FungibleAsset, FungibleAssetStore};
+    use aptos_framework::object::{Self, Object, ConstructorRef, DeriveRef};
+
+    use std::signer;
+
+    #[resource_group_member(group = aptos_framework::object::ObjectGroup)]
+    /// Resource stored on the fungible asset metadata object to allow creating primary stores for it.
+    struct DeriveRefPod has key {
```
movekevin, line 10: Why is DeriveRefPod a better name?

aptos-move/framework/aptos-framework/sources/primary_store.move
```diff
@@ -0,0 +1,156 @@
+/// This defines the module for interacting with primary stores of accounts/objects, which have deterministic addresses
+module aptos_framework::primary_store {
```
movekevin, line 2: Why is store > wallet?

aptos-move/framework/aptos-framework/sources/primary_store.move
lightmark: change it back.
the assumption changes after I make the derived_object function friend.

aptos-move/framework/aptos-framework/sources/primary_store.move
lightmark: Because it can be used for other purpose. If you agree that function is friend function, I agree to change it back to the original name. Otherwise, it is just a holder of derived ref.

aptos-move/framework/aptos-framework/sources/primary_store.move
lightmark: from different ppl including our metapixel friends. I think store is better than container since they are used to xxx_store already.

aptos-move/move-examples/fungible_asset/sources/managed_fungible_asset.move
lightmark: my bad.

aptos-move/framework/aptos-framework/sources/object.move
lightmark: you don't need a signer now.
Another rationale is offering users to use this derive_ref arbitrarily would potentially make difference resources using the same object. I am not sure this is something we want.
cc @davidiw @CapCap

aptos-move/framework/aptos-framework/sources/primary_store.move
movekevin: @gregnazario Any naming suggestion here? :P

aptos-move/framework/aptos-framework/sources/object.move
```diff
@@ -471,6 +472,14 @@ module aptos_framework::object {
     }
 
     /// Accessors
```
gregnazario, line 474: This should be a // and not a ///

aptos-move/framework/aptos-framework/sources/primary_store.move
gregnazario: I mean I would make it PrimaryWalletCapability, though I think we overloaded that word now...

PrimaryWalletDeriveRef

Tells you what it is, and what it's used for.  Do you suspect it'll be used for anything else in the future?  Then

FungibleAssetDeriveRef

gregnazario (approved):

aptos-move/framework/aptos-framework/sources/fungible_asset.move
```diff
@@ -53,14 +76,14 @@ module aptos_framework::fungible_asset {
...
-    struct FungibleAsset has key {
+    /// The store object that holds fungible assets of a specific type associated with an account.
+    struct FungibleStore has key {
         /// The address of the base metadata object.
         metadata: Object<Metadata>,
         /// The balance of the fungible metadata.
         balance: u64,
         /// Fungible Assets transferring is a common operation, this allows for freezing/unfreezing accounts.
-        allow_ungated_transfer: bool,
+        allow_ungated_balance_transfer: bool,
```
movekevin, line 86: allow_owner_balance_transfer?

aptos-move/framework/aptos-framework/sources/fungible_asset.move
```diff
@@ -21,26 +23,47 @@ module aptos_framework::fungible_asset {
...
+    /// Name of the fungible asset metadata is too long
+    const ENAME_TOO_LONG: u64 = 15;
+    /// Symbol of the fungible asset metadata is too long
+    const ESYMBOL_TOO_LONG: u64 = 16;
+
+    //
+    // Constants
+    //
+
+    const MAX_NAME_LENGTH: u64 = 32;
```
movekevin, line 51: How did you come up with these numbers?

PR #7622, opened 2023-04-06
[Move Examples] Add a simple RNG.

JoshLind:
### Description
This PR adds a very simple RNG to the `move-examples` crate. The RNG is inspired by RANDAO (and other RNGs that leverage commit-reveal schemes for random number generation). The code should hopefully be well-documented and clean enough that it can provide a good starting place for someone looking to explore commit-reveal schemes, basic treasury functionality, phased timeouts, etc. The PR also includes a large suite of unit tests.

**Note: I've prefaced the RNG with a disclaimer that the code is an example only. It is not intended to be used in production and is not the recommended way to generate random numbers.**

### Test Plan
The example includes a suite of new tests (which all pass).
```
...
BUILDING Simple Random Number Generator (RNG)
Running Move unit tests
[ PASS    ] 0xcafe::Rng::test_empty_backup_commitment
[ PASS    ] 0xcafe::Rng::test_empty_commitment
[ PASS    ] 0xcafe::Rng::test_identical_commitments
[ PASS    ] 0xcafe::Rng::test_init_invalid_max_nums
[ PASS    ] 0xcafe::Rng::test_init_invalid_timeout
[ PASS    ] 0xcafe::Rng::test_init_invalid_user_count
[ PASS    ] 0xcafe::Rng::test_init_not_owner
[ PASS    ] 0xcafe::Rng::test_invalid_reveal
[ PASS    ] 0xcafe::Rng::test_invalid_reveal_multiple_users
[ PASS    ] 0xcafe::Rng::test_invalid_reveal_not_enough_stake
[ PASS    ] 0xcafe::Rng::test_not_init_commit
[ PASS    ] 0xcafe::Rng::test_not_init_reveal
[ PASS    ] 0xcafe::Rng::test_not_init_reveal_backup
[ PASS    ] 0xcafe::Rng::test_rng_backup_reveal_without_reveal
[ PASS    ] 0xcafe::Rng::test_rng_commit_phase_timeout
[ PASS    ] 0xcafe::Rng::test_rng_happy_invalid_round_high
[ PASS    ] 0xcafe::Rng::test_rng_happy_invalid_round_zero
[ PASS    ] 0xcafe::Rng::test_rng_happy_paths
[ PASS    ] 0xcafe::Rng::test_rng_happy_paths_sequence
[ PASS    ] 0xcafe::Rng::test_rng_reveal_backup_phase_timeout
[ PASS    ] 0xcafe::Rng::test_rng_reveal_phase_timeout
[ PASS    ] 0xcafe::Rng::test_rng_reveal_phase_timeout_all_backups
[ PASS    ] 0xcafe::Rng::test_rng_reveal_phase_timeout_some_backups
Test result: OK. Total tests: 23; passed: 23; failed: 0
test test_rng ... ok
...
```


davidiw:
> Note: I've prefaced the RNG with a disclaimer that the code is an example only. It is not intended to be used in production and is not the recommended way to generate random numbers.

Considering this, do we really want to add this to the code?

JoshLind:
> Considering this, do we really want to add this to the code?

Yeah, thinking about this again, I think my disclaimer is too strong 😄 Personally, I think this example is useful because it illustrates important blockchain features (i.e., on-chain hashing, commit-reveal, staking and slashing, block timestamps and timeouts, etc.). If I were the reader, I could see value in having access to it, e.g.,: (i) RANDAO is a fairly well-known approach to randomness, so this implementation should provide a good starting point for those who wish to leverage it; and (ii) looking at the other code examples, I'd argue the quality of this is fair, so it's not like this is bad/unsafe code.

My only motivation for the disclaimer is that real randomness is coming and thus, when it does come, it would be preferable to use that instead of this. But, I don't see us being there anytime soon. And, given that there are zero options today, I'd argue that using this example a base for crowd-sourced randomness isn't actually bad. In fact, many folks are already rolling their own randomness (e.g., via block timestamps), which is arguable less secure than this approach. So, all-in-all I'm a little torn.

My suggestion would be to remove the disclaimer altogether (it's obvious from the code that it's an example), and then when real randomness does arrive, we update the example to point to the real randomness and/or remove this entirely? @alinush, thoughts? You were my only motivation for the disclaimer :)

JoshLind:
Closing this out after some discussion. Drand support should land soon 😄

PR #7623, opened 2023-04-06
Add Supporting resources list for Move Prover 

clay-aptos:


saharct (approved):

junkil-park:
Thank you @clay-aptos for starting this! Since it may be better for the supporting resources to be closer to the Move Prover book by adding a new section in it, I will close this PR, and do that in a separate PR: https://github.com/aptos-labs/aptos-core/pull/7851

PR #7634, opened 2023-04-07
[Move-prover][Spec] Revive the pragma verify_duration_estimate 

### LINKED ISSUE #7321: [Feature Request][Move prover] New flag and command line option for suppressing proof targets
# 🚀 Feature Request

## Motivation

Some proof targets can sometimes timeout due to different reasons such as butterfly effects. It is desirable to have an explicit `pragma suppress` that suppresses these targets when necessary and a prover command option `--verify-all` to turn them back on conveniently. With this pragma, targets that cannot be proved will be specified with `pragma verify = false` while ones that may cause timeout and need to be suppressed will be specified with `pragma suppress`.

rahxephon89:
### Description

This PR revives the function-level pragma `verify_duration_estimate`, which indicates an estimate how long verification takes. Verification of this function will be skipped if the its timeout value is smaller than this. 

Close #7321.

### DIFF
--- a/aptos-move/framework/aptos-framework/doc/staking_contract.md
+++ b/aptos-move/framework/aptos-framework/doc/staking_contract.md
@@ -2001,7 +2001,7 @@ Staking_contract exists the stacker/operator pair.
 Staking_contract exists the stacker/operator pair.
 
 
-<pre><code><b>pragma</b> verify = <b>false</b>;
+<pre><code><b>pragma</b> verify_duration_estimate = 120;
 <b>let</b> staking_contracts = <b>global</b>&lt;<a href="staking_contract.md#0x1_staking_contract_Store">Store</a>&gt;(staker).staking_contracts;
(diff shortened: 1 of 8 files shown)


aptos-move/framework/aptos-framework/doc/staking_contract.md
```diff
@@ -2430,118 +2430,4 @@ The guid_creation_num of the ccount resource is up to MAX_U64.
 </code></pre>
 
 
-The Store exists under the staker.
```
wrwg, line 2433: Why is this all deleted?

wrwg (approved):
Great to have this back!

junkil-park (approved):

aptos-move/framework/aptos-framework/doc/staking_contract.md
rahxephon89: It seems to be a bug when generating the doc. After removing the comment added to [staking_contract.spec.move](https://github.com/aptos-labs/aptos-core/pull/7634/files/0c1e998fa527c4edaa80b5e2e80c148bd56b4b25#diff-e9a5d468be1897db0ac84561d1bd4770f4a6a789fc18e59b18218c783057c676), this part is added back

PR #7637, opened 2023-04-07
Add failpoints for network sending (for forge tests)

igor-aptos:
Add network::send::any failpoint, which individually decides (per message, even for broadcast), whether to send it or not.

(otherwise, if broadcast is decided in bulk, small failure %, can make nobody receive a broadcast, which is not realistic)

### DIFF
--- a/network/src/peer_manager/senders.rs
+++ b/network/src/peer_manager/senders.rs
@@ -13,6 +13,7 @@ use crate::{
 use aptos_channels::{self, aptos_channel};
 use aptos_types::{network_address::NetworkAddress, PeerId};
 use bytes::Bytes;
+use fail::fail_point;
 use futures::channel::oneshot;
 use std::time::Duration;
 
@@ -47,6 +48,10 @@ impl PeerManagerRequestSender {
         protocol_id: ProtocolId,
         mdata: Bytes,
     ) -> Result<(), PeerManagerError> {
+        fail_point!("network::send::any", |_| {
+            // to look like it all went well
+            Ok(())
+        });
         self.inner.push(
             (peer_id, protocol_id),
             PeerManagerRequest::SendDirectSend(peer_id, Message { protocol_id, mdata }),
@@ -73,6 +78,11 @@ impl PeerManagerRequestSender {
     ) -> Result<(), PeerManagerError> {
         let msg = Message { protocol_id, mdata };
         for recipient in recipients {
+            // Make failure independent for each destination
+            fail_point!("network::send::any", |_| {
+                // to look like it all went well
+                Ok(())
+            });
             // We return `Err` early here if the send fails. Since sending will
             // only fail if the queue is unexpectedly shutdown (i.e., receiver
             // dropped early), we know that we can't make further progress if
@@ -100,6 +110,10 @@ impl PeerManagerRequestSender {
             res_tx,
             timeout,
         };
+        fail_point!("network::send::any", |_| {
+            std::thread::sleep(timeout);
+            Err(RpcError::TimedOut)
+        });
         self.inner.push(
             (peer_id, protocol_id),
             PeerManagerRequest::SendRpc(peer_id, request),
--- a/testsuite/testcases/src/modifiers.rs
+++ b/testsuite/testcases/src/modifiers.rs
@@ -146,7 +146,7 @@ impl NetworkLoadTest for NetworkUnreliabilityTest {
                 );
                 validator
                     .set_failpoint(
-                        "consensus::send::any".to_string(),
+                        "network::send::any".to_string(),
                         format!("{}%return", drop_percentage),
                     )
                     .await
@@ -171,7 +171,7 @@ impl NetworkLoadTest for NetworkUnreliabilityTest {
         runtime.block_on(async {
             for (name, validator) in validators {
                 validator
-                    .set_failpoint("consensus::send::any".to_string(), "off".to_string())
+                    .set_failpoint("network::send::any".to_string(), "off".to_string())
                     .await
                     .map_err(|e| {
                         anyhow::anyhow!(
--- a/.github/workflows/ts-sdk-e2e-tests.yaml
+++ b/.github/workflows/ts-sdk-e2e-tests.yaml
@@ -89,6 +89,7 @@ jobs:
   # to be land blocking for any PR on the aptos repo. This is why we run those tests
   # separate from test-sdk-confirm-client-generated-publish.
   run-indexer-tests:
+    needs: [permission-check]
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8 # pin@v3
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2270,6 +2270,7 @@ dependencies = [
  "async-trait",
  "bcs 0.1.4",
  "bytes",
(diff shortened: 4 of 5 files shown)


network/src/peer_manager/senders.rs
```diff
@@ -100,6 +110,10 @@ impl PeerManagerRequestSender {
             res_tx,
             timeout,
         };
+        fail_point!("network::send::any", |_| {
+            std::thread::sleep(timeout);
```
JoshLind, line 114: You probably want tokio::time::sleep, no? Otherwise this will block the entire OS thread?

network/src/peer_manager/senders.rs
```diff
@@ -47,6 +48,10 @@ impl PeerManagerRequestSender {
         protocol_id: ProtocolId,
         mdata: Bytes,
     ) -> Result<(), PeerManagerError> {
+        fail_point!("network::send::any", |_| {
```
JoshLind, line 51: Is your goal to add this only for consensus? Or for all components? Right now, this will affect everything, right?
Also, why did you add this to the `PeerManagerRequestSender` and not at the network interface level? e.g., https://github.com/aptos-labs/aptos-core/blob/f005fa18b1095aad2ba4f128e5da88c89e4d164b/network/src/application/interface.rs#L166. 

If you want to add it to everything, the interface feels like a cleaner place?

danielxiangzl: +1. But it looks like there is no easy way to add fail points only for consensus, without changing the existing networking?

network/src/peer_manager/senders.rs
zekun000: it's probably not easy to plug in async function here, but not sure

network/src/peer_manager/senders.rs
zekun000: we can decide based on protocol id but probably not ideal

igor-aptos: I wanted to have an unreliable network - for everything, not just for consensus, so it is a realistic scenario. If RPC being delayed, blocks other RPCs from other protocols, and causes issues, we do want that.

for consensus only, we have consensus::send::any already


For why I placed it there, I wanted to have a separate decision for each destination peer, which is simpler to do in PeerManagerRequestSender, because network interface level does send_to_many in a single call

JoshLind (approved):
LGTM. But I think the sleep should be `tokio::time::sleep`?

network/src/peer_manager/senders.rs
igor-aptos: @JoshLind  - don't know how to have async inside of failpoint lambda?

network/src/peer_manager/senders.rs
sitalkedia: This looks reasonable to me but wondering if we should use this for realistic workload scenario or use chaos monkey's network packet drop support. Later seems more realistic to me.

testsuite/testcases/src/modifiers.rs
```diff
@@ -0,0 +1,202 @@
...
+        }
+        Ok::<(), anyhow::Error>(())
+    })
+}
+
+/// Config for additing variable processing overhead/delay into
+/// execution, to make different nodes have different processing speed.
+pub struct ExecutionDelayConfig {
+    /// Fraction (0.0 - 1.0) of nodes on which any delay will be introduced
+    pub inject_delay_node_fraction: f64,
```
sitalkedia, line 78: nit: better to make it percentage instead of fraction to be consistent

testsuite/testcases/src/modifiers.rs
```diff
@@ -0,0 +1,202 @@
...
+    /// For nodes with delay, what percentage (0-100) of transaction will be delayed.
+    /// (this is needed because delay that can be introduced is integer number of ms)
+    /// Different node speed come from this setting, each node is selected a number
+    /// between 1 and given max.
+    pub inject_delay_max_transaction_percentage: u32,
```
sitalkedia, line 83: Can you clarify why is this needed? Why not varying `inject_delay_per_transaction_ms` per node is not sufficient?

testsuite/testcases/src/modifiers.rs
```diff
@@ -0,0 +1,202 @@
...
+    /// Fixed busy-loop delay applied to each transaction that is delayed,
+    /// before it is executed.
+    pub inject_delay_per_transaction_ms: u32,
```
sitalkedia, line 86: Does this need to be lower granularity than `ms` - 1 ms delay per txn is very high..

network/src/peer_manager/senders.rs
JoshLind: Aah, gotcha. Yeah, not sure if failpoints supports that. In which case, ignore me 😄

davidiw (approved):

alokeutpal (approved):

davidiw:
@gelash and team, it might be worth looking at this PR as it seems to trigger the following:

```[aptos_vm] Transaction breaking invariant violation. txn: Ok([90, 120, 118, 196, 241, 97, 45, 148, 77, 27, 231, 239, 120, 137, 74, 39, 247, 239, 223, 157, 160, 20, 194, 255, 171, 80, 251, 184, 53, 251, 4, 147, 22, 0, 0, 0, 0, 0, 0, 0, 2, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 4, 99, 111, 105, 110, 8, 116, 114, 97, 110, 115, 102, 101, 114, 1, 7, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 10, 97, 112, 116, 111, 115, 95, 99, 111, 105, 110, 9, 65, 112, 116, 111, 115, 67, 111, 105, 110, 0, 2, 32, 2, 91, 61, 229, 231, 160, 12, 101, 58, 43, 216, 182, 67, 254, 191, 211, 20, 10, 72, 144, 30, 164, 88, 182, 85, 198, 34, 203, 205, 57, 140, 197, 8, 1, 0, 0, 0, 0, 0, 0, 0, 128, 132, 30, 0, 0, 0, 0, 0, 100, 0, 0, 0, 0, 0, 0, 0, 209, 248, 160, 100, 0, 0, 0, 0, 4, 0, 32, 85, 192, 214, 194, 59, 148, 20, 126, 199, 128, 102, 216, 101, 129, 85, 169, 220, 148, 40, 65, 46, 37, 206, 201, 171, 236, 98, 158, 221, 72, 159, 178, 64, 6, 247, 151, 45, 178, 93, 15, 232, 58, 2, 51, 81, 27, 179, 164, 172, 69, 47, 234, 246, 13, 120, 28, 21, 220, 10, 231, 219, 149, 60, 78, 6, 179, 53, 214, 44, 25, 126, 103, 210, 9, 12, 174, 42, 74, 139, 53, 210, 96, 197, 116, 134, 145, 82, 209, 158, 52, 88, 76, 94, 251, 93, 208, 9]), status: ERROR { status_code: UNEXPECTED_ERROR_FROM_KNOWN_MOVE_FUNCTION, message: None }```

PR #6761, opened 2023-02-24
Simplifcations in executor_benchmark

igor-aptos:
- added some logging
- added support for --split-stages
- removed BenchmarkTransaction, and instead parsed transaction for fake executor. makes code cleaner, and allows me to much more easily reuse transaction generator from txn emitter in the executor benchmark

PR #7638, opened 2023-04-07
Added support for --split-stages in executor benchmark

igor-aptos:
added support for --split-stages, and preparing to add transaction-generator-lib for generating workload

This replaces PR #6761, without removing BenchmarkTransaction. But still adding support to deserialize transactions for execution, as I'll be passing all transactions from  transaction-generator-lib with empty extra info information.


JoshLind (approved):
Unblocking

msmouse (approved):

PR #7639, opened 2023-04-07
[Quorum Store] BatchGenerator produces multiple bucketed batches

bchocho:
### Description

Create multiple bucketed batches after pulling from BatchGenerator. Each batch is for the bucket indicated by gas_bucket_start. The batch, signing, and broadcast messages all support multiple batches to reduce any additional message overhead compared to creating a single batch.

### Test Plan

Added unit test for buckets. Ran forge three_region_simulation_graceful_overload to see that there are no regressions.


consensus/src/quorum_store/batch_generator.rs
```diff
@@ -129,49 +163,50 @@ impl BatchGenerator {
...
-            .collect();
 
-        let batch = Batch::new(
-            batch_id,
-            pulled_txns,
-            self.epoch,
-            expiry_time,
-            self.my_peer_id,
-        );
+        pulled_txns.sort_by_key(|txn| txn.gas_unit_price());
```
zekun000, line 179: so we don't take seqnum into considerations even from the same pull?

consensus/src/quorum_store/batch_store.rs
```diff
@@ -308,7 +308,17 @@ impl<T: QuorumStoreSender + Clone + Send + Sync + 'static> BatchStore<T> {
         ret
     }
 
-    pub fn persist(&self, persist_request: PersistedValue) -> Option<SignedBatchInfo> {
+    pub fn persist(&self, persist_requests: Vec<PersistedValue>) -> Vec<SignedBatchInfo> {
+        let mut signed_infos = vec![];
+        for persist_request in persist_requests.into_iter() {
+            if let Some(signed_info) = self.persist_inner(persist_request) {
```
zekun000, line 314: do we need to handle partial failures now?

consensus/consensus-types/src/proof_of_store.rs
```diff
@@ -125,6 +128,39 @@ impl BatchInfo {
...
+pub struct SignedBatchInfoMsg {
+    signed_infos: Vec<SignedBatchInfo>,
+}
+
+impl SignedBatchInfoMsg {
+    pub fn new(signed_infos: Vec<SignedBatchInfo>) -> Self {
+        Self { signed_infos }
+    }
+
+    pub fn verify(&self, sender: PeerId, validator: &ValidatorVerifier) -> anyhow::Result<()> {
```
zekun000, line 147: we probably should fail if the vector is empty

consensus/src/quorum_store/batch_generator.rs
bchocho: Yes, in this PR we don't consider the sequence number at all. A future PR (or update on this PR) will consider both the current pull and the excluded txns.

consensus/src/quorum_store/batch_store.rs
bchocho: I think the current behavior, that allows to fail some and succeed others in the same message, should be fine. Anyway what will matter if the batch reached quorum or not.

consensus/src/quorum_store/batch_generator.rs
bchocho: I added it to this PR: 5bcb5784e209da8df83c3408a0c2292a46189472

consensus/src/quorum_store/batch_generator.rs
```diff
@@ -124,54 +211,80 @@ impl BatchGenerator {
...
-            .expect("Could not save to db");
+        // Compute bucketed batches
+        let bucket_compute_start = Instant::now();
 
+        let exclude_txns_with_gas: Vec<_> = self
+            .batches_in_progress
+            .values()
+            .flatten()
+            .cloned()
+            .collect();
```
sasha8, line 232: nit: probably could merge the traversals

consensus/src/quorum_store/batch_generator.rs
```diff
@@ -95,18 +102,98 @@ impl BatchGenerator {
...
+            let num_batch_txns = std::cmp::min(self.config.max_batch_txns, remaining_txns);
+            let batch_txns: Vec<_> = txns.drain(0..num_batch_txns).collect();
+            let batch = self.create_new_batch(batch_txns, expiry_time, bucket_start);
+            batches.push(batch);
+            remaining_txns -= num_batch_txns;
+        }
+    }
+
+    fn build_account_min_gas(
+        exclude_txns: &[TransactionInProgress],
```
sasha8, line 164: Some txns here might be already committed. So min might be wrong due to race. Is that an issue?

consensus/src/quorum_store/batch_generator.rs
```diff
@@ -124,54 +211,80 @@ impl BatchGenerator {
...
-            .collect();
 
-        let batch = Batch::new(
-            batch_id,
-            pulled_txns,
-            self.epoch,
-            expiry_time,
-            self.my_peer_id,
-        );
+        pulled_txns.sort_by_cached_key(|txn| {
```
sasha8, line 237: I am not sure implementation guarantees elements with the same values will not be reordered.

sasha8 (commented):
Some initial comments

consensus/src/quorum_store/batch_store.rs
```diff
@@ -308,7 +308,17 @@ impl<T: QuorumStoreSender + Clone + Send + Sync + 'static> BatchStore<T> {
...
+                signed_infos.push(signed_info);
+            }
+        }
+        signed_infos
+    }
+
+    pub fn persist_inner(&self, persist_request: PersistedValue) -> Option<SignedBatchInfo> {
```
sasha8, line 321: Should not be pub.

consensus/src/quorum_store/proof_coordinator.rs
sasha8, line 1: Possible bug, but unrelated to this PR. @zekun000 have a look at add_signature (line 189). After your changes, anybody can cause init_proof. So I can attack you by keep sending you old batches or even garbage. 
I thought only self could call init_proof, but I do not see it. Am I missing something?

sasha8 (commented):
Two bugs:

consensus/src/quorum_store/batch_generator.rs
```diff
@@ -124,54 +207,81 @@ impl BatchGenerator {
...
-            .save_batch_id(self.epoch, self.batch_id)
-            .expect("Could not save to db");
+        // Compute bucketed batches
+        let bucket_compute_start = Instant::now();
 
+        let exclude_txns_with_gas: Vec<_> = self
+            .batches_in_progress
+            .values()
+            .flatten()
+            .cloned()
```
zekun000, line 227: I don't think you need the clone and collect here? I feel like impl Iterator<Item = &TransactionInProgress> is enough

consensus/src/quorum_store/batch_generator.rs
```diff
@@ -124,54 +207,81 @@ impl BatchGenerator {
...
+                &mut batches,
+                &mut pulled_txns,
+                split_index,
+                expiry_time,
+                bucket_start,
+            );
+        }
+        if !pulled_txns.is_empty() {
+            let last_index = self.config.batch_buckets.len() - 1;
+            let bucket_start = self.config.batch_buckets[last_index];
```
zekun000, line 269: .last().unwrap()?

consensus/src/quorum_store/batch_generator.rs
```diff
@@ -124,54 +207,81 @@ impl BatchGenerator {
...
+        for (&bucket_start, &bucket_end) in buckets.iter().tuple_windows() {
+            if pulled_txns.is_empty() {
+                break;
+            }
 
-        self.batches_in_progress.insert(batch_id, txn_summaries);
-        self.batch_expirations.add_item(batch_id, expiry_time);
+            let split_index = pulled_txns
+                .iter()
+                .find_position(|txn| {
```
zekun000, line 247: this is a linear search not binary search

consensus/src/quorum_store/network_listener.rs
```diff
@@ -50,31 +50,33 @@ impl NetworkListener {
...
+                        let cmd = ProofCoordinatorCommand::AppendSignature(*signed_batch_infos);
                         self.proof_coordinator_tx
                             .send(cmd)
                             .await
                             .expect("Could not send signed_batch_info to proof_coordinator");
                     },
                     VerifiedEvent::BatchMsg(batch_msg) => {
-                        let batch = batch_msg.unpack();
+                        let author = batch_msg.author();
+                        let batches = batch_msg.take();
```
zekun000, line 62: do we want to limit how many of them?

consensus/src/quorum_store/batch_generator.rs
```diff
@@ -160,36 +157,59 @@ impl BatchGenerator {
...
-        }
-        for txn in pulled_txns {
-            let current_min = account_min_gas
-                .entry(txn.sender())
-                .or_insert_with(|| txn.gas_unit_price());
-            if txn.gas_unit_price() < *current_min {
-                *current_min = txn.gas_unit_price()
+
+            let num_txns_in_bucket = match pulled_txns
+                .binary_search_by_key(&(bucket_end, PeerId::ZERO), |txn| {
```
sasha8, line 175: Are you sure this is what you want? From the documentation: "If there are multiple matches, then any one of the matches could be returned". Meaning that txns with the same gas can go to different buckets, right?

consensus/consensus-types/src/common.rs
```diff
@@ -22,18 +22,33 @@ pub type Round = u64;
...
 impl fmt::Display for TransactionSummary {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         write!(f, "{}:{}", self.sender, self.sequence_number,)
     }
 }
 
+#[derive(Clone, Eq, PartialEq, Ord, PartialOrd)]
+pub struct TransactionInProgress {
+    pub summary: TransactionSummary,
+    pub gas_unit_price: u64,
```
sasha8, line 49: Why cannot this go inside TransactionSummary?

consensus/src/quorum_store/utils.rs
```diff
@@ -103,14 +103,14 @@ impl MempoolProxy {
...
-        return_non_full: bool,
-        exclude_txns: Vec<TransactionSummary>,
+        exclude_txns: Vec<TransactionInProgress>,
     ) -> Result<Vec<SignedTransaction>, anyhow::Error> {
         let (callback, callback_rcv) = oneshot::channel();
         let msg = QuorumStoreRequest::GetBatchRequest(
             max_items,
             max_bytes,
-            return_non_full,
+            true,
```
sasha8, line 112: If it is always true can we get rid of the parm?
I think direct_mempool sets it to false, but do we still care about direct mempool? 
Or should we wait until we delete the direct mempool code?

sasha8 (commented):
Ready to approve, just check please the comment about the binary_search_by_key.

consensus/src/quorum_store/batch_generator.rs
bchocho: The key for the binary search includes the PeerId, and the search key includes PeerId::ZERO. So we don't have to worry about multiple matches. (AFAICT PeerId::ZERO is not a valid address; it's used as a placeholder for "null" in a few places in the code.)

consensus/src/quorum_store/utils.rs
bchocho: Yup, direct_mempool sets it to false, so we need the param until we remove direct_mempool.

consensus/consensus-types/src/common.rs
bchocho: In mempool, we use TransactionSummary as the HashMap key (it's type aliased to TxnPointer) and we don't want it to be unique based on different gas_unit_price

consensus/src/quorum_store/network_listener.rs
bchocho: I think this is the last remaining comment that I haven't been able to address. I'll think through if there are other cases too and push a commit tomorrow.

sasha8 (approved):

consensus/src/quorum_store/network_listener.rs
bchocho: Addressed in cb8e3c674d26fcf2b8730ad9ded73bd2cfa470fd

bchocho:
@zekun000 is signature verification (for SignedBatchInfo messages) expensive? We might just need a global hard cap on how many batches can be included in this message, checked before the signature verification. Thoughts?

zekun000:
signature verification is relatively expensive compared to the bytes one multi signature has. although nodes can always split into many messages, putting a limit is more for sanity check.

consensus/src/quorum_store/batch_generator.rs
```diff
@@ -131,6 +131,26 @@ impl BatchGenerator {
...
+        batches: &mut Vec<Batch>,
+        txns: &mut Vec<SignedTransaction>,
+        num_txns: usize,
+        expiry_time: u64,
+        bucket_start: u64,
+    ) {
+        let mut remaining_txns = num_txns;
+        while remaining_txns > 0 {
+            let num_batch_txns = std::cmp::min(self.config.max_batch_txns, remaining_txns);
+            let batch_txns: Vec<_> = txns.drain(0..num_batch_txns).collect();
```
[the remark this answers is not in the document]
zekun000: I thought collect is a linear operation with memory allocations but maybe there's some optimizations. split_off can be combined with mem swap but no strong opinion

consensus/src/quorum_store/batch_generator.rs
zekun000: it means it's not supposed to return Ok(..) right?

consensus/src/quorum_store/batch_generator.rs
```diff
@@ -95,9 +99,119 @@ impl BatchGenerator {
...
+                    (txn.gas_unit_price(), txn.sender())
+                }) {
+                Ok(index) => index,
+                Err(index) => index,
+            };
+            if num_txns_in_bucket == 0 {
+                continue;
+            }
+
+            let batches_space_remaining = self.push_bucket_to_batches(
```
zekun000, line 189: this looks like we're creating low gas price batches first?

zekun000: I'm actually confused, after we removed the gas price normalization, do we still have the monotonic increasing gas guarantee in the pulled txn?

consensus/src/quorum_store/batch_generator.rs
```diff
@@ -161,23 +163,35 @@ impl BatchGenerator {
...
+
     fn bucket_into_batches(
         &mut self,
         pulled_txns: &mut Vec<SignedTransaction>,
         expiry_time: u64,
     ) -> Vec<Batch> {
-        pulled_txns.sort_by_key(|txn| txn.gas_unit_price());
+        // Sort by gas bucket, in descending order. Due to stable sort on existing mempool account
+        // ordering, this will result in (descending(gas), account) order.
+        pulled_txns.sort_by_cached_key(|txn| self.get_descending_bucket_key(txn));
```
zekun000, line 181: why not just sort by u64::max - gas_unit_price?

zekun000 (approved):

consensus/src/quorum_store/batch_generator.rs
bchocho: Makes sense, since this is descending order. 👍 
I think I was playing with ascending order at some point, where the bucketing would have to be used.

zekun000: I think it's even simpler to sort_by(|a, b| b.gas_unit_price().cmp(a.gas_unit_price())?

PR #7640, opened 2023-04-07
Updated the authorization check in `aptos_token::authorized_borrow`

junkil-park:
### Description
To make the authorization check meaningful, `authorized_borrow` needs to take the `&signer` of the creator instead of `address` as an argument. Otherwise, one can always bypass it by calling `authorized_borrow(&token, token::creator(token))`.

The same reason goes to `authorized_borrow_collection`.

### Test Plan
cargo test

### DIFF
--- a/aptos-move/framework/aptos-token-objects/sources/aptos_token.move
+++ b/aptos-move/framework/aptos-token-objects/sources/aptos_token.move
@@ -289,7 +289,7 @@ module aptos_token_objects::aptos_token {
 
     // Token mutators
 
-    inline fun authorized_borrow<T: key>(token: &Object<T>, creator: address): &AptosToken {
+    inline fun authorized_borrow<T: key>(token: &Object<T>, creator: &signer): &AptosToken {
         let token_address = object::object_address(token);
         assert!(
             exists<AptosToken>(token_address),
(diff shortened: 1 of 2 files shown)


aptos-move/framework/aptos-token-objects/sources/aptos_token.move
```diff
@@ -289,22 +289,22 @@ module aptos_token_objects::aptos_token {
 
     // Token mutators
 
-    inline fun authorized_borrow<T: key>(token: &Object<T>, creator: address): &AptosToken {
```
movekevin, line 292: This function is private though so there should be no risk with another module calling it?

junkil-park: Right, there is no security concern here. 

However, `authorized_borrow` should be a nice pattern that many other custom tokens will follow to use potentially (e.g., `authorized_worker_borrow` to authorize the worker account and borrow the reference of token in some token module). I thought it would be great to set a good example in this standard module because other people could refer this pattern.

And, I think the right way to "authorize" is to check `&signer` instead of `address` in principle.

davidiw (approved):
Fine by me, I had contemplated doing this, but honestly, I was lazy to make a decision.

movekevin (approved):

PR #7648, opened 2023-04-07
[simple] add failed round counter

igor-aptos:

### DIFF
--- a/consensus/src/block_storage/block_store.rs
+++ b/consensus/src/block_storage/block_store.rs
@@ -57,6 +57,16 @@ pub fn update_counters_for_committed_blocks(blocks_to_commit: &[Arc<ExecutedBloc
         counters::LAST_COMMITTED_ROUND.set(block.round() as i64);
         counters::LAST_COMMITTED_VERSION.set(block.compute_result().num_leaves() as i64);
 
+        let failed_rounds = block
+            .block()
+            .block_data()
+            .failed_authors()
+            .map(|v| v.len())
+            .unwrap_or(0);
+        if failed_rounds > 0 {
+            counters::COMMITTED_FAILED_ROUNDS_COUNT.inc_by(failed_rounds as u64);
+        }
+
         // Quorum store metrics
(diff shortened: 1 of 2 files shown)


zekun000 (commented):
what's the difference between this and this https://github.com/aptos-labs/aptos-core/blob/d9d6791cfc316043e66db25da399c45ae7344539/consensus/src/round_manager.rs#L290?

igor-aptos:
> what's the difference between this and this?
> 
> https://github.com/aptos-labs/aptos-core/blob/d9d6791cfc316043e66db25da399c45ae7344539/consensus/src/round_manager.rs#L290

@zekun000 Not sure, but some concerns:
- Are we guaranteed to not skip rounds in process_new_round_event? (i.e. to go through all?)
- how is timeout certificate handled?

zekun000:
- we do skip rounds (jump rounds with certificate)
- not sure what do you mean about the timeout certificate handling, we move to TC.round + 1 like the QC

igor-aptos:
Given that this will not count all successful/failed rounds, I still think it is useful to have this?

zekun000 (approved):

JoshLind (approved):

consensus/src/block_storage/block_store.rs
```diff
@@ -56,6 +56,11 @@ pub fn update_counters_for_committed_blocks(blocks_to_commit: &[Arc<ExecutedBloc
         counters::LAST_COMMITTED_ROUND.set(block.round() as i64);
         counters::LAST_COMMITTED_VERSION.set(block.compute_result().num_leaves() as i64);
 
+        let failed_rounds = block.block().block_data().failed_authors().map(|v| v.len()).unwrap_or(0);
```
sitalkedia, line 59: Can you clarify how is it different from `aptos_consensus_timeout_rounds_count`?

igor-aptos: that one skips rounds based on how we receive certificates. you can see the discussion above

sitalkedia (approved):

PR #7650, opened 2023-04-07
[storage] add kv only option to statesnapshot restore

areshand:
### Description
based on https://www.notion.so/aptoslabs/Enhanced-DB-Restore-cf86fbfacbb94eab8e155078f56cec34?pvs=4, we need to be able to recover a snapshot with kv only to speed up

### Test Plan

use existing tests
will also jointly test with kv restore PR later

### DIFF
--- a/storage/aptosdb/src/backup/restore_handler.rs
+++ b/storage/aptosdb/src/backup/restore_handler.rs
@@ -60,6 +60,7 @@ impl RestoreHandler {
         &self,
         version: Version,
         expected_root_hash: HashValue,
+        kv_only: bool,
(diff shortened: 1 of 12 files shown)


msmouse:
* Why can't we use `StateValueRestore` directly?
* what's the plan for `tree only` mode which is required when the second snaphot is used?

areshand:
> * Why can't we use `StateValueRestore` directly?
> * what's the plan for `tree only` mode which is required when the second snaphot is used?

This PR is merged to https://github.com/aptos-labs/aptos-core/pull/7699
close this

PR #7651, opened 2023-04-07
Use transaction-generator-lib in executor benchmark

igor-aptos:
enabling to test on same workloads as Forge

### DIFF
--- a/execution/executor-benchmark/src/db_access.rs
+++ b/execution/executor-benchmark/src/db_access.rs
@@ -1,0 +1,154 @@
+// Copyright © Aptos Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+use anyhow::Result;
+use aptos_state_view::TStateView;
+use aptos_storage_interface::{cached_state_view::CachedStateView, state_view::DbStateView};
+use aptos_types::{
+    access_path::AccessPath, account_address::AccountAddress, state_store::state_key::StateKey,
+};
+use move_core_types::{
+    identifier::Identifier,
+    language_storage::{StructTag, TypeTag},
+};
+use once_cell::sync::Lazy;
+use serde::{de::DeserializeOwned, Deserialize, Serialize};
+use std::str::FromStr;
+
+// Note: in case this changes in the future, it doesn't have to be a constant, and can be read from
+// genesis directly if necessary.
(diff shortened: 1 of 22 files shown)


execution/executor-benchmark/src/gen_executor.rs
```diff
@@ -0,0 +1,60 @@
+// Copyright © Aptos Foundation
+// Parts of the project are originally copyright © Meta Platforms, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+use std::{time::Duration, sync::atomic::AtomicUsize};
+use aptos_types::transaction::{SignedTransaction, Transaction};
+use aptos_types::account_address::AccountAddress;
+use aptos_transaction_generator_lib::TransactionExecutor as GenInitTransactionExecutor;
```
msmouse, line 8: What does  "gen init" mean in this context? 😂

msmouse (approved):
(seems legit)

execution/executor-benchmark/src/db_access.rs
```diff
@@ -0,0 +1,154 @@
...
+    language_storage::{StructTag, TypeTag},
+};
+use once_cell::sync::Lazy;
+use serde::{de::DeserializeOwned, Deserialize, Serialize};
+use std::str::FromStr;
+
+// Note: in case this changes in the future, it doesn't have to be a constant, and can be read from
+// genesis directly if necessary.
+pub static TOTAL_SUPPLY_STATE_KEY: Lazy<StateKey> = Lazy::new(|| {
+    StateKey::table_item(
```
JoshLind, line 21: Scary 😨

JoshLind (approved):
Unblocking

execution/executor-benchmark/src/db_access.rs
igor-aptos: @grao1991's code, just extracted to a separate file here :)

execution/executor-benchmark/src/gen_executor.rs
igor-aptos: TransactionExecutor for transaction generator initialization. I didn't think of naming conflict when I was naming it that, maybe  aptos_transaction_generator_lib::TransactionExecutor should be renamed to something like ClientTransactionExecutor or ReliableTransactionSubmitter

PR #7653, opened 2023-04-08
Make shareholder function handle both beneficiary and shareholder addresses to fix `accumulated_rewards` function

banool:
## Description

In the vesting contract at this address:
```
0x3de72a2e5d6a330b919ee6de9a6f51f6250c32b1385107c7b2db02bc82bfd587
```

You see this entry in the beneficiaries map:
```
{
  key: "0x3168c3c9173a81f6923496649ef69995f79d10dee31cc5580b03057ce3a5b27f"
  value: "0x385c76d283c205a1ac3b117fe059d39ef80c5d67bcb1a3b73a2e220f67130512"
}
```

Where key is the shareholder and value is the beneficiary.

You can see that the `shareholder` function works as intended:
```
$ aptos move view --url https://fullnode.mainnet.aptoslabs.com --function-id 0x1::vesting::shareholder --args address:0x3de72a2e5d6a330b919ee6de9a6f51f6250c32b1385107c7b2db02bc82bfd587 address:0xd36d942f6cc900178f981e5196e4a1e4e9bdc7ed760afd33ca5553a4cad83126
{
  "Result": [
    "0x31711902c817a6b8cd83c999320da2c1e328695cd879f6115995c3fff7a29ac1"
  ]
}
```

So far so good. However, the `accumulated_rewards` function says it takes `shareholder_or_beneficiary`. If you pass in the beneficiary, the result looks good:
```
$ aptos move view --url https://fullnode.mainnet.aptoslabs.com --function-id 0x1::vesting::accumulated_rewards --args address:0x3de72a2e5d6a330b919ee6de9a6f51f6250c32b1385107c7b2db02bc82bfd587 address:0x385c76d283c205a1ac3b117fe059d39ef80c5d67bcb1a3b73a2e220f67130512
{
  "Result": [
    "22171919592"
  ]
}
```

If you pass in the shareholder however, it looks like this:
```
$ aptos move view --url https://fullnode.mainnet.aptoslabs.com --function-id 0x1::vesting::accumulated_rewards --args address:0x3de72a2e5d6a330b919ee6de9a6f51f6250c32b1385107c7b2db02bc82bfd587 address:0x3168c3c9173a81f6923496649ef69995f79d10dee31cc5580b03057ce3a5b27f
{
  "Result": [
    "0"
  ]
}
```

Meaning passing in the shareholder address doesn't actually work like the argument would imply for the `accumulated_rewards` function.

This PR fixes this problem by making the `shareholder` function just return the given address if the address is in the shareholders map as a key.

This is not the only way we could solve this, we could make `accumulated_rewards` figure this out instead. But I feel this is less error prone, and makes it match how the `beneficiary` function works.

## Test Plan
```
cd aptos-move/framework/aptos-framework
aptos move test
```

### DIFF
--- a/aptos-move/framework/aptos-framework/sources/vesting.move
+++ b/aptos-move/framework/aptos-framework/sources/vesting.move
@@ -390,20 +390,24 @@ module aptos_framework::vesting {
 
     #[view]
     /// Return the shareholder address given the beneficiary address in a given vesting contract. If there are multiple
-    /// shareholders with the same beneficiary address, only the first shareholder is returned.
+    /// shareholders with the same beneficiary address, only the first shareholder is returned. If the given beneficiary
+    /// address is actually a shareholder address, just return the address back.
     ///
-    /// This returns 0x0 if no shareholder is found for the given beneficiary.
-    public fun shareholder(vesting_contract_address: address, beneficiary: address): address acquires VestingContract {
+    /// This returns 0x0 if no shareholder is found for the given beneficiary / the address is not a shareholder itself.
+    public fun shareholder(vesting_contract_address: address, shareholder_or_beneficiary: address): address acquires VestingContract {
         assert_active_vesting_contract(vesting_contract_address);
 
         let shareholders = &shareholders(vesting_contract_address);
+        if (vector::contains(shareholders, &shareholder_or_beneficiary)) {
+            return shareholder_or_beneficiary
+        };
         let vesting_contract = borrow_global<VestingContract>(vesting_contract_address);
         let i = 0;
         let len = vector::length(shareholders);
         while (i < len) {
             let shareholder = *vector::borrow(shareholders, i);
             // This will still return the shareholder if shareholder == beneficiary.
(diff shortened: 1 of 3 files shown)


davidiw (approved):

aptos-move/framework/aptos-framework/sources/vesting.move
```diff
@@ -373,6 +373,7 @@ module aptos_framework::vesting {
         assert_active_vesting_contract(vesting_contract_address);
 
         let total_accumulated_rewards = total_accumulated_rewards(vesting_contract_address);
+
```
movekevin, line 376: Why this extra new line?

movekevin (approved):
I think you need to regenerate the move docs as well. I usually just run cargo test in aptos-move/framework/aptos-framework

aptos-move/framework/aptos-framework/sources/vesting.move
banool: No idea, I can undo that.

Issue #7575, opened 2023-04-05
[DevEx RFC][TS-SDK] Add support for compound `TypeTag`


JackyWYX:
## Motivation

Currently, compound / nested type argument is not allowed in Aptos TS-SDK. E.g. 

```
0x5a97986a9d031c4567e15b797be516910cfcb4156312482efc6a19c0a30c948::lp_coin::LP<0xf22bede237a07e121b56d91a491eb7bcdfd1f5907926a9e58338f964a01b17fa::asset::USDC, 0x1::aptos_coin::AptosCoin, 0x190d44266241744264b964a37b8f09863167a12d3e70cda39376cfb4e3561e12::curves::Uncorrelated>
```

The given structure is a LP coin type for Pontem liquidity token. The structure is not able to be called in TS-SDK for coin register / transfer. 

## Proposals

Implement the nested TypeTag for compound structure. 

```
// aptos/src/aptos_types/type_tag.ts
export class StructTag {
    ...
    static fromString(structTag: string): StructTag {
    // Type args are not supported in string literal
    if (structTag.includes("<")) {
      throw new Error("Not implemented");
    }
```

wjw12:
Please prioritize this issue as it's blocking many operations involving LP token types

gregnazario:
So summary of the issue here:
* TS SDK doesn't support nested struct tags with generics e.g. (0x1::coin::Coin<0x1::module::Struct<0x2::coin::MyCoin>>)
* Parsing doesn't currently handle this


Items that need to be done:
* Parse the three pieces of the function e.g. `0x1::coin::Coin` into `0x1, coin, Coin`
* Parse any associated generic types accordingly at a top level e.g. `0x1::coin::Coin<0x2::coin::MyCoin, 0x1::module::Struct<0x2::coin::MyCoin>>` into `0x2::coin::MyCoin, 0x1::module::Struct<0x2::coin::MyCoin>`

borispovod:
👍

0xjinn:
Hi @JackyWYX , may I ask what methods you invoke to create the transaction, or how you create the transaction?

Specifically, curious if you done in this approach, or similar?

```
const rawTxn = await aptosClient.generateTransaction(my_address, {
      function: "0x1::managed_coin::register",
      type_arguments: [`0x1:🪙:Coin<0x1::module::Struct<0x2:🪙:MyCoin>>`],
      arguments: [],
    });
```

PR #7659, opened 2023-04-10
[SDK] Refactor and move the TypeTagParser to type_tag file

0xjinn:
### Description
This is a required refactor PR that move the `TypeTagParser` to the type_tag.ts file (Which I believe should be the correct place for the parser). 

So we can use the TypeTagParser to fix issue #7575  

### Why not just reference it from builder_utils?
Primarily, We cannot import it from the `builder_utils.ts` because of **circular dependency**. `type_tag` <-> `builder_utils` we need to move the parser to type_tag, so it can be used to implement StructTag.fromString function here - https://github.com/aptos-labs/aptos-core/blob/500b349df3b495da388d85d8680494f1a01dde22/ecosystem/typescript/sdk/src/aptos_types/type_tag.ts#L192-L204

### Test Plan
pnpm run lint
pnpm test


0xmaayan:
I am a bit confused, why do you need to move the `TypeTagParser` and can't simply import it from `transaction_builder`? as for where it belongs, I would argue it can live in its own file - `type_tag_parser.ts`

0xjinn:
> I am a bit confused, why do you need to move the `TypeTagParser` and can't simply import it from `transaction_builder`? as for where it belongs, I would argue it can live in its own file - `type_tag_parser.ts`

@0xmaayan Yeah, that was actually my first attempt, but importing it from the builder will result in a circular dependency as I mentioned in the description. 

But yeah, i think moving to its own file works as well!

0xmaayan:
Placing the file in `aptos_types` can lead to a breaking change for anyone importing/using this parser. Now, they should be able to simply do `import {TypeTagParser} from aptos`, but once we move it to `aptos_types` they would need to use it as
```
import {TxnBuilderTypes} from aptos;
TxnBuilderTypes.TypeTagParser
```
We might want to think on placing it in a folder that users can keep importing/using it as they do now.

Also, could you make sure we update imports? for example, I believe `property_map_serde.ts` uses `import { TypeTagParser } from "../transaction_builder";`

gregnazario (approved):
Can you write some positive and negative unit tests for the typetag parser

0xmaayan:
can we
1. create a new file for the parser - `type_tag_parser.ts` that lives in `aptos_types` folder
2. make sure we update imports? for example, I believe `property_map_serde.ts` uses `import { TypeTagParser } from "../transaction_builder";`

0xmaayan (approved):

0xjinn:
> Can you write some positive and negative unit tests for the typetag parser

Will add the unit test in the diff https://github.com/aptos-labs/aptos-core/pull/7660

0xmaayan:
could you also add this change to CHANGELOG.md

0xjinn:
> could you also add this change to CHANGELOG.md

@0xmaayan Ah yes! Will include that in the next PR

PR #7663, opened 2023-04-10
Update 1-create-smart-contract.md

rohan-201:


davidiw (changes requested):
Please feel free to re-open and indicate why this change is being made. Each change conflicts with the intended results -- though maybe "2. Update the "... is better than Add, while keeping the original function name.

PR #7664, opened 2023-04-10
[release-cli] Add Release ARM64 Binary

WGB5445:
major changes：Added an arm64 build option in GitHub Actions and included a Dockerfile.

### Description

Hello everyone, I often use arm64 Linux devices for code development. However, I found that aptos doesn't have compiled arm64 binary release packages. 

Every time aptos releases a new version, I need to compile for over 10 hours on my computer.
 
I hope this PR can help build an arm64 binary.

Possible issues: I created a new Dockerfile because it requires using the docker qemu method. If there is a better way, please let me know. Thank you

### Test Plan

need to run it using GitHub Action.


gregnazario:
I was looking into this, and it likely needs some more work on compilation.  This might be a good project to look into for this https://github.com/cross-rs/cross

PR #7666, opened 2023-04-10
Improved Connection Balance and Committed Votes after Updating HAProxy Configuration

p1xel32:
I wanted to share a recent experience I had with my validator server. Before the last update, my server had a connection parameter distribution of approximately 50/50. However, after the update, the distribution shifted to around 80/20 and remained that way for about two weeks. Normally, the distribution would revert within the first 3-4 days after an update.
After discussing the issue with some fellow community members, I decided to make changes to my haproxy.cfg file. I updated the following parameters:
nbthread 8 
timeout queue 10s 
timeout connect 10s 
timeout server 120s 
timeout client 120s 
And also changed the backend validator-fn configuration( belive that this is the key):
default-server maxconn 128 
To my delight, within minutes, the connection rate adjusted to approximately 40/60. Additionally, my committed_votes metric improved from an average of 60/12h to 70.
@rustielin

### DIFF
--- a/.github/workflows/copy-images-to-dockerhub-release.yaml
+++ b/.github/workflows/copy-images-to-dockerhub-release.yaml
@@ -1,4 +1,4 @@
-name: Copy images to dockerhub devnet/testnet
+name: Copy images to dockerhub on release
 on:
   push:
     branches:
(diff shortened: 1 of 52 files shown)


davidiw (changes requested):
Should this have been created as an issue? There appears to be no code and this is just to inform @rustielin of an optimization.

PR #7667, opened 2023-04-10
Remove unsafe

gerben-stavenga:
### Description

Over the weekend I investigated some SIGSEGV I experienced and was looking at unsafe code in our code base as possible culprits. I encountered this which I initially thought must be wrong (Option<Box<Entry>> is type punned as an usize). It turns out that absolute compiler magic makes Option<Box<Entry>> the same size as usize, but I feel that kind of code is ridiculous.

The problem is that one cannot use [None, NB_BUCKETS] as Bucket == Option<Box<Entry>> is non-copyable, however one can simply work around that by using cont INIT: Bucket = None, and use INIT as initialization value.

This seems cleaner

### DIFF
--- a/third_party/move/move-symbol-pool/src/pool.rs
+++ b/third_party/move/move-symbol-pool/src/pool.rs
@@ -62,8 +62,8 @@ impl Pool {
     /// Allocates a contiguous array of buckets on the heap. As strings are
     /// inserted into the pool, buckets in this array are filled with an entry.
     pub(crate) fn new() -> Self {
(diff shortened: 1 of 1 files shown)


davidiw (commented):
Very clean indeed, I don't suppose you have a test that would demonstrate how to get the sigsegv?

wrwg (approved):

davidiw (approved):

gerben-stavenga:
> 

The sigsegv was caused by stack overflow, (I had raised the stack limit with ulimit -s but that didnt help but wolfgang pointed me to env RUST_MIN_STACK) while investigating the memory I stumbled upon this

PR #7671, opened 2023-04-10
add consensus config: use_compression

brianolson:
### Description

Add `use_compression` (default true) value to consensus config. When false the compressed protocol is disabled.

### Test Plan

Added unit test

### DIFF
--- a/aptos-node/src/network.rs
+++ b/aptos-node/src/network.rs
@@ -48,7 +48,15 @@ struct ApplicationNetworkHandle<T> {
 /// TODO: make this configurable (e.g., for compression)
 /// Returns the network application config for the consensus client and service
 pub fn consensus_network_configuration(node_config: &NodeConfig) -> NetworkApplicationConfig {
-    let direct_send_protocols: Vec<ProtocolId> = DIRECT_SEND.into();
+    let direct_send_protocols: Vec<ProtocolId> = if node_config.consensus.use_compression {
+        DIRECT_SEND.into()
+    } else {
+        DIRECT_SEND
+            .iter()
+            .filter(|&&x| x != ProtocolId::ConsensusDirectSendCompressed)
+            .copied()
+            .collect()
+    };
     let rpc_protocols: Vec<ProtocolId> = RPC.into();
 
     let network_client_config =
@@ -380,5 +388,29 @@ fn create_network_interfaces<
     ApplicationNetworkInterfaces {
         network_client,
         network_service_events,
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[test]
+    fn test_consensus_network_configuration() {
+        let mut nc: NodeConfig = NodeConfig::default_for_validator();
+
+        nc.consensus.use_compression = true;
+        let nac1 = consensus_network_configuration(&nc);
+        assert!(nac1
+            .network_service_config
+            .direct_send_protocols_and_preferences
+            .contains(&ProtocolId::ConsensusDirectSendCompressed));
+
+        nc.consensus.use_compression = false;
+        let nac2 = consensus_network_configuration(&nc);
+        assert!(!nac2
+            .network_service_config
+            .direct_send_protocols_and_preferences
+            .contains(&ProtocolId::ConsensusDirectSendCompressed));
     }
 }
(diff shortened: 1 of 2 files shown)


aptos-node/src/network.rs
```diff
@@ -48,7 +48,15 @@ struct ApplicationNetworkHandle<T> {
 /// TODO: make this configurable (e.g., for compression)
 /// Returns the network application config for the consensus client and service
 pub fn consensus_network_configuration(node_config: &NodeConfig) -> NetworkApplicationConfig {
-    let direct_send_protocols: Vec<ProtocolId> = DIRECT_SEND.into();
+    let direct_send_protocols: Vec<ProtocolId> = if node_config.consensus.use_compression {
```
JoshLind, line 51: Nice! Note, we'll need to do the same for the RPC message set below, i.e., remove `ConsensusRpcCompressed` from the list 😄

config/src/config/consensus_config.rs
```diff
@@ -59,6 +59,7 @@ pub struct ConsensusConfig {
     // must match one of the CHAIN_HEALTH_WINDOW_SIZES values.
     pub window_for_chain_health: usize,
     pub chain_health_backoff: Vec<ChainHealthBackoffValues>,
+    pub use_compression: bool, // compress messages for the network
```
JoshLind, line 62: @zekun000, we discussed this feature a while back. It's probably still useful, right? e.g., for benchmarks, experiments, etc.?

zekun000 (commented):
Instead of this, should we just make the protocol id configurable?

JoshLind:
> Instead of this, should we just make the protocol id configurable?

@zekun000, can you elaborate? I don't follow. Are you suggesting we enable/disable it somewhere else? (e.g., [here](https://github.com/aptos-labs/aptos-core/blob/f005fa18b1095aad2ba4f128e5da88c89e4d164b/network/src/protocols/wire/handshake/v1/mod.rs#L107)). Or did you mean something else?

config/src/config/consensus_config.rs
davidiw: my recommendation is to eliminate the comment and make this flag more explicit `enable_network_compression`

brianolson: there's node.yaml `state_sync.aptos_data_client.use_compression: true`
So that started the precedent of having compression on/off for different sections of the protocol; and state_sync has 3 protocol modes, there just isn't a feature flag to disable the one with compression until now.

zekun000:
I was suggesting instead of using a boolean config value `use_compression`, we can make the protocol id in the configs so that next time we migrate and update the protocol id can be a config change instead of binary update.

I imagine we can take the hard coded list here https://github.com/aptos-labs/aptos-core/blob/main/consensus/src/network_interface.rs#L111 and put them into the configs, we can serialize them with human readable string instead of numbers

brianolson:
So the config line might look like this?

```
protocols: ["ConsensusDirectSendCompressed", "ConsensusDirectSendBcs", "ConsensusDirectSendJson"]
```

JoshLind:
> I was suggesting instead of using a boolean config value use_compression, we can make the protocol id in the configs so that next time we migrate and update the protocol id can be a config change instead of binary update.

Aah, nice! You are a wizard 🧙 That makes sense to me 😄

config/src/config/consensus_config.rs
JoshLind: I think @davidiw's comment is simply that he'd prefer the config variable to named `enable_network_compression` instead of `use_compression`.

In regards to state sync, `use_compression` makes slightly more sense because we're not actually compressing the network bytes, but the entire response before it gets to the network layer (this is slightly more efficient, because we're able to do more aggressive caching, etc.)

JoshLind:
@zekun000: @brianolson raises a good point. We can't make the config crate depend on the network crate to get the definition for the protocol IDs (because it results in circular deps). Taking a quick look, it appears to be too much work to extract the protocol ID types into a new crate.

I propose we take @brianolson's previous approach for now (i.e., just remove the compressed protocol's from the list if compression is disabled), and then in the future (once we've cleaned up the network crate), we can make this fully configurable. What do you think?

PR #7672, opened 2023-04-10
Add CLI support for multisig accounts (v2)

movekevin:
### Description
This adds the CLI commands for interacting with multisig accounts v2. The commands are
1. aptos multisig create --addtional-owners owner1,owner2 --num-signatures-required 2
2. aptos multisig create-transaction --multisig-address <multisig address> --function-id 0x1::multisig_account::add_owners --args address:owner3
3. aptos multisig approve --multisig-address <multisig address> --sequence-number <sequence number of tx to approve>
aptos multisig reject --multisig-address <multisig address> --sequence-number <sequence number of tx to approve>
4. Execute the next tx that has enough approvals: aptos multisig execute --multisig-address <multisig address>
5. Execute (effectively remove) the next tx that has enough rejections: aptos multisig execute-reject --multisig-address <multisig address>

### Test Plan
Manual tests in mainnet

### DIFF
--- a/crates/aptos/src/account/multisig_account.rs
+++ b/crates/aptos/src/account/multisig_account.rs
@@ -1,0 +1,257 @@
+// Copyright © Aptos Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::common::types::{
+    CliCommand, CliTypedResult, EntryFunctionArguments, MultisigAccount, TransactionOptions,
+    TransactionSummary,
+};
+use aptos_cached_packages::aptos_stdlib;
+use aptos_rest_client::{
+    aptos_api_types::{WriteResource, WriteSetChange},
+    Transaction,
+};
+use aptos_types::{
+    account_address::AccountAddress,
+    transaction::{Multisig, MultisigTransactionPayload, TransactionPayload},
+};
+use async_trait::async_trait;
+use bcs::to_bytes;
+use clap::Parser;
+use serde::Serialize;
+
+/// Create a new multisig account (v2) on-chain.
+///
+/// This will create a new multisig account and make the sender one of the owners.
+#[derive(Debug, Parser)]
+pub struct Create {
+    /// Addresses of additional owners for the new multisig, beside the transaction sender.
+    #[clap(long, multiple_values = true, parse(try_from_str=crate::common::types::load_account_arg))]
+    pub(crate) additional_owners: Vec<AccountAddress>,
+    /// The number of signatures (approvals or rejections) required to execute or remove a proposed
+    /// transaction.
+    #[clap(long)]
+    pub(crate) num_signatures_required: u64,
+    #[clap(flatten)]
+    pub(crate) txn_options: TransactionOptions,
+}
+
+/// A shortened create multisig account output
+#[derive(Clone, Debug, Serialize)]
+pub struct CreateSummary {
+    #[serde(flatten)]
+    pub multisig_account: Option<MultisigAccount>,
+    #[serde(flatten)]
+    pub transaction_summary: TransactionSummary,
+}
+
+impl From<Transaction> for CreateSummary {
+    fn from(transaction: Transaction) -> Self {
+        let transaction_summary = TransactionSummary::from(&transaction);
+
+        let mut summary = CreateSummary {
+            transaction_summary,
+            multisig_account: None,
+        };
+
+        if let Transaction::UserTransaction(txn) = transaction {
+            summary.multisig_account = txn.info.changes.iter().find_map(|change| match change {
+                WriteSetChange::WriteResource(WriteResource { address, data, .. }) => {
+                    if data.typ.name.as_str() == "Account"
+                        && *address.inner().to_hex() != *txn.request.sender.inner().to_hex()
+                    {
+                        Some(MultisigAccount {
+                            multisig_address: *address.inner(),
+                        })
+                    } else {
+                        None
+                    }
+                },
+                _ => None,
+            });
+        }
+
+        summary
+    }
+}
+
+#[async_trait]
+impl CliCommand<CreateSummary> for Create {
+    fn command_name(&self) -> &'static str {
+        "CreateMultisig"
+    }
+
+    async fn execute(self) -> CliTypedResult<CreateSummary> {
+        self.txn_options
+            .submit_transaction(aptos_stdlib::multisig_account_create_with_owners(
+                self.additional_owners,
+                self.num_signatures_required,
+                // TODO: Support passing in custom metadata.
+                vec![],
+                vec![],
+            ))
+            .await
+            .map(CreateSummary::from)
+    }
+}
+
+/// Propose a new multisig transaction.
+///
+/// As one of the owners of the multisig, propose a new transaction. This also implicitly approves
+/// the created transaction so it has one approval initially. In order for the transaction to be
+/// executed, it needs as many approvals as the number of signatures required.
+#[derive(Debug, Parser)]
+pub struct CreateTransaction {
+    #[clap(flatten)]
+    pub(crate) multisig_account: MultisigAccount,
+    #[clap(flatten)]
+    pub(crate) txn_options: TransactionOptions,
+    #[clap(flatten)]
+    pub(crate) entry_function_args: EntryFunctionArguments,
+}
+
+#[async_trait]
+impl CliCommand<TransactionSummary> for CreateTransaction {
+    fn command_name(&self) -> &'static str {
+        "CreateTransactionMultisig"
+    }
+
+    async fn execute(self) -> CliTypedResult<TransactionSummary> {
+        let payload = MultisigTransactionPayload::EntryFunction(
+            self.entry_function_args.create_entry_function_payload()?,
+        );
+        self.txn_options
+            .submit_transaction(aptos_stdlib::multisig_account_create_transaction(
+                self.multisig_account.multisig_address,
+                to_bytes(&payload)?,
+            ))
+            .await
+            .map(|inner| inner.into())
+    }
+}
+
+/// Approve a multisig transaction.
+///
+/// As one of the owners of the multisig, approve a transaction proposed for the multisig.
+/// With enough approvals (as many as the number of signatures required), the transaction can be
+/// executed (See Execute).
+#[derive(Debug, Parser)]
+pub struct Approve {
+    #[clap(flatten)]
+    pub(crate) multisig_account: MultisigAccount,
+    /// The sequence number of the multisig transaction to approve. The sequence number increments
+    /// for every new multisig transaction.
+    #[clap(long)]
+    pub(crate) sequence_number: u64,
+    #[clap(flatten)]
+    pub(crate) txn_options: TransactionOptions,
+}
+
+#[async_trait]
+impl CliCommand<TransactionSummary> for Approve {
+    fn command_name(&self) -> &'static str {
+        "ApproveMultisig"
+    }
+
+    async fn execute(self) -> CliTypedResult<TransactionSummary> {
+        self.txn_options
+            .submit_transaction(aptos_stdlib::multisig_account_approve_transaction(
+                self.multisig_account.multisig_address,
+                self.sequence_number,
+            ))
+            .await
+            .map(|inner| inner.into())
+    }
+}
+
+/// Reject a multisig transaction.
+///
+/// As one of the owners of the multisig, reject a transaction proposed for the multisig.
+/// With enough rejections (as many as the number of signatures required), the transaction can be
+/// completely removed (See ExecuteReject).
+#[derive(Debug, Parser)]
+pub struct Reject {
+    #[clap(flatten)]
+    pub(crate) multisig_account: MultisigAccount,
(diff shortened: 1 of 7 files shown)


alnoki:
@movekevin if I'm not mistaken it looks like the create transaction functionality only works for core type arguments (via `MoveType`) and hence will not enable coin transfers that require a coin type argument. Is there any way to enable non-core types?

Also, is there a way we could accept a JSON input file, for example, during transaction creation? This will allow signers to all agree on the same transaction information. Then signers can compute the hash or payload of the JSON file-specified transaction data locally and compare against on-chain. Something like:

```json
{
    "multisig_account_address": "0xace",
    "function": "0x1::coin::transfer",
    "type_arguments": [
        "0x1::aptos_coin::AptosCoin"
    ],
    "arguments": [
        {
            "argument_type": "address",
            "argument_value": "0xbee"
        },
        {
            "argument_type": "u64",
            "argument_value": 100
        }
    ]
}
```

Here, a proposed transaction can be posted on GitHub and everyone can agree on the same truth reference.

Lastly, I notice that the current implementation does not support nested vectors, despite being valid Move types. These are crucial in governance operations where on-chain parameter sets may need to be tuned ([example](https://github.com/econia-labs/econia/blob/337f6824157c9988e2310613b6fc63643801f51e/src/move/econia/sources/incentives.move#L979)).

If JSON data like above were allowed as an input, then arguments could simply be specified for the inner Move types then the `[` and `]` could be used by serde to infer the level of nesting. Something like:


```json
        {
            "argument_type": "u64",
            "argument_value": [
                [100, 200],
                [300, 400]
            ]
        }
``` 
Here the inner type is a `u64`, and there are two vector levels, hence `vector<vector<u64>>`

movekevin:
> @movekevin if I'm not mistaken it looks like the create transaction functionality only works for core type arguments (via `MoveType`) and hence will not enable coin transfers that require a coin type argument. Is there any way to enable non-core types?
> 
> Also, is there a way we could accept a JSON input file, for example, during transaction creation? This will allow signers to all agree on the same transaction information. Then signers can compute the hash or payload of the JSON file-specified transaction data locally and compare against on-chain. Something like:
> 
> ```json
> {
>     "multisig_account_address": "0xace",
>     "function": "0x1::coin::transfer",
>     "type_arguments": [
>         "0x1::aptos_coin::AptosCoin"
>     ],
>     "arguments": [
>         {
>             "argument_type": "address",
>             "argument_value": "0xbee"
>         },
>         {
>             "argument_type": "u64",
>             "argument_value": 100
>         }
>     ]
> }
> ```
> 
> Here, a proposed transaction can be posted on GitHub and everyone can agree on the same truth reference.
> 
> Lastly, I notice that the current implementation does not support nested vectors, despite being valid Move types. These are crucial in governance operations where on-chain parameter sets may need to be tuned ([example](https://github.com/econia-labs/econia/blob/337f6824157c9988e2310613b6fc63643801f51e/src/move/econia/sources/incentives.move#L979)).
> 
> If JSON data like above were allowed as an input, then arguments could simply be specified for the inner Move types then the `[` and `]` could be used by serde to infer the level of nesting. Something like:
> 
> ```json
>         {
>             "argument_type": "u64",
>             "argument_value": [
>                 [100, 200],
>                 [300, 400]
>             ]
>         }
> ```
> 
> Here the inner type is a `u64`, and there are two vector levels, hence `vector<vector<u64>>`

Type args are supported. You just need to pass --type-args in the create-transaction command

movekevin:
> Also, is there a way we could accept a JSON input file, for example, during transaction creation? This will allow signers to all agree on the same transaction information. Then signers can compute the hash or payload of the JSON file-specified transaction data locally and compare against on-chain. Something like:
> 
> ```json
> {
>     "multisig_account_address": "0xace",
>     "function": "0x1::coin::transfer",
>     "type_arguments": [
>         "0x1::aptos_coin::AptosCoin"
>     ],
>     "arguments": [
>         {
>             "argument_type": "address",
>             "argument_value": "0xbee"
>         },
>         {
>             "argument_type": "u64",
>             "argument_value": 100
>         }
>     ]
> }
> ```
> 
> Here, a proposed transaction can be posted on GitHub and everyone can agree on the same truth reference.

You mean having the CLI command take a JSON file as input? It's possible, can you file a github issue for this? I'll talk with @gregnazario later to understand the priority here, as this can be generalized for aptos move run as well

movekevin:
> notice that the current implementation does not support nested vectors, despite being valid Move types. These are crucial in governance operations where on-chain parameter sets may need to be tuned ([example](https://github.com/econia-labs/econia/blob/337f6824157c9988e2310613b6fc63643801f51e/src/move/econia/sources/incentives.move#L979)).
> 
> If JSON data like above were allowed as an input, then arguments could simply be specified for the inner Move types then the `[` and `]` could be used by serde to infer the level of nesting. Something like:
> 
> ```json
>         {
>             "argument_type": "u64",
>             "argument_value": [
>                 [100, 200],
>                 [300, 400]
>             ]
>         }
> ```
> 
> Here the inner type is a `u64`, and there are two vector levels, hence `vector<vector<u64>>`

I think we need general support for vector< vector< > > as well for aptos move run. This might already be on @gregnazario's list though. If not, we can file a Github issue to track

gregnazario:
In response to the what about `vector<vector<>>` question, I believe any more complex inputs into the CLI will have to either be handled by a file input as JSON, or as BCS.  Keeping in mind that these inputs get really complicated at this point

crates/aptos/src/account/multisig_account.rs
```diff
@@ -0,0 +1,244 @@
...
+    ///
+    /// Example: `address:0x1 bool:true u8:0 u256:1234 'vector<u32>:a,b,c,d'`
+    #[clap(long, multiple_values = true)]
+    pub(crate) args: Vec<ArgWithType>,
+
+    /// TypeTag arguments separated by spaces.
+    ///
+    /// Example: `u8 u16 u32 u64 u128 u256 bool address vector signer`
+    #[clap(long, multiple_values = true)]
+    pub(crate) type_args: Vec<MoveType>,
```
gregnazario, line 116: We should make these into a common type so we share it with the run function / not duplicate logic.

crates/aptos/src/account/multisig_account.rs
```diff
@@ -0,0 +1,244 @@
...
+}
+
+#[async_trait]
+impl CliCommand<TransactionSummary> for CreateTransaction {
+    fn command_name(&self) -> &'static str {
+        "CreateTransaction"
```
gregnazario, line 122: Let's end all of these with `Multisig` so we can differentiate the metrics.

crates/aptos/src/account/multisig_account.rs
```diff
@@ -0,0 +1,244 @@
...
+    async fn execute(self) -> CliTypedResult<TransactionSummary> {
+        let payload = MultisigTransactionPayload::EntryFunction(create_entry_function_payload(
+            self.function_id,
+            self.args,
+            self.type_args,
+        )?);
+        self.txn_options
+            .submit_transaction(aptos_stdlib::multisig_account_create_transaction(
+                self.multisig_account,
+                to_bytes(&payload).unwrap(),
```
gregnazario, line 134: you should be able to do `?` to return the error conversion rather than an `unwrap()`

crates/aptos/src/lib.rs
```diff
@@ -43,6 +43,8 @@ pub enum Tool {
     #[clap(subcommand)]
     Move(move_tool::MoveTool),
     #[clap(subcommand)]
+    MultisigAccount(account::MultisigAccountTool),
```
gregnazario, line 46: Let's rename this `Multisig` so the command will be `aptos multisig ...` rather than `aptos multisig-account`

crates/aptos/src/account/multisig_account.rs
```diff
@@ -0,0 +1,244 @@
...
+    }
+}
+
+#[derive(Debug, Parser)]
+pub struct ExecuteReject {
+    #[clap(long, parse(try_from_str=crate::common::types::load_account_arg))]
+    pub(crate) multisig_account: AccountAddress,
+    #[clap(flatten)]
+    pub(crate) txn_options: TransactionOptions,
+}
```
gregnazario, line 228: Each of these commands will need the following:
* A short description
* A longer description
* Documentation for each field
```
/// Rejects a multisig transaction
///
/// As a signer of multisig, rejects a transaction proposed by another multisig
/// signer. ...
#[derive(Debug, Parser)]
pub struct ExecuteReject {
    /// The multisig account address to sign
    ///
    /// This address should not be the same as the account signing
    #[clap(long, parse(try_from_str=crate::common::types::load_account_arg))]
    pub multisig_account: AccountAddress
    ...
}
```

Additionally, we should move the `multisig_account` to a common struct since it's used among all of these so it's documented once.

crates/aptos/src/account/multisig_account.rs
```diff
@@ -0,0 +1,244 @@
...
+        }
+
+        summary
+    }
+}
+
+#[async_trait]
+impl CliCommand<CreateSummary> for Create {
+    fn command_name(&self) -> &'static str {
+        "Create"
```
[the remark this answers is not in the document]
movekevin: Added Multisig suffix to all of these command names

crates/aptos/src/lib.rs
movekevin: Updated

crates/aptos/src/account/multisig_account.rs
movekevin: Updated

alnoki:
@movekevin it looks like the commands in this PR will not profile gas if `TransactionOptions.profile_gas` is marked true.

If I'm not mistaken this means that if a user flags the option to profile, they will have the transaction actually executed on accident. @gregnazario Is this a correct interpretation?

See #7693

alnoki:
> You mean having the CLI command take a JSON file as input? It's possible, can you file a github issue for this? I'll talk with @gregnazario later to understand the priority here, as this can be generalized for aptos move run as well

> In response to the what about `vector<vector<>>` question, I believe any more complex inputs into the CLI will have to either be handled by a file input as JSON, or as BCS. Keeping in mind that these inputs get really complicated at this point

@movekevin @gregnazario addressed in #7709

crates/aptos/src/account/multisig_account.rs
```diff
@@ -0,0 +1,253 @@
...
+///
+/// As one of the owners of the multisig, approve a transaction proposed for the multisig.
+/// With enough approvals (as many as the number of signatures required), the transaction can be
+/// executed (See Execute).
+#[derive(Debug, Parser)]
+pub struct Approve {
+    #[clap(flatten)]
+    pub(crate) multisig_account: MultisigAccount,
+    #[clap(long)]
+    pub(crate) sequence_number: u64,
```
gregnazario, line 142: We should add a doc comment saying this is the `multisig` sequence number, rather than any of the other ones

crates/aptos/src/account/multisig_account.rs
```diff
@@ -0,0 +1,253 @@
...
+///
+/// As one of the owners of the multisig, reject a transaction proposed for the multisig.
+/// With enough rejections (as many as the number of signatures required), the transaction can be
+/// completely removed (See ExecuteReject).
+#[derive(Debug, Parser)]
+pub struct Reject {
+    #[clap(flatten)]
+    pub(crate) multisig_account: MultisigAccount,
+    #[clap(long)]
+    pub(crate) sequence_number: u64,
```
gregnazario, line 174: similarly

gregnazario (approved):
Please rebase

crates/aptos/src/account/multisig_account.rs
movekevin: Added

crates/aptos/src/account/multisig_account.rs
movekevin: Added

crates/aptos/src/common/types.rs
```diff
@@ -1669,3 +1671,61 @@ pub struct RotationProofChallenge {
...
+pub struct EntryFunctionArguments {
+    /// Function name as `<ADDRESS>::<MODULE_ID>::<FUNCTION_NAME>`
+    ///
+    /// Example: `0x842ed41fad9640a2ad08fdd7d3e4f7f505319aac7d67e1c0dd6a7cce8732c7e3::message::set_message`
+    #[clap(long)]
+    pub function_id: MemberId,
+
+    /// Arguments combined with their type separated by spaces.
+    ///
+    /// Supported types [u8, u16, u32, u64, u128, u256, bool, hex, string, address, raw, vector<inner_type>]
```
gerben-stavenga, line 1686: Does this need Object etc.. support

xindingw (commented):
Can users see the command description in doc/cli help?

crates/aptos/src/common/types.rs
movekevin: Yeah. It's likely not as high priority so later :)

gerben-stavenga (approved):
Not super familiar with this code yet but lgtm

PR #7673, opened 2023-04-10
[Move-prover] Fix trace bug

### LINKED ISSUE #7594: [Bug][Move-prover] <internal> location should not be printed out in the trace
# 🐛 Bug

See the [trace](https://github.com/aptos-labs/aptos-core/blob/85f562bbd486c291b7ae24c8507a01fbfdd8b6e6/third_party/move/move-prover/tests/sources/functional/loops_with_memory_ops.exp#L47)

rahxephon89:
### Description

This PR:

- close #7594 by filtering out `<internal>` in the trace; 
- remove duplicated trace items.

### Test Plan

See the changed trace file in tests.


junkil-park (approved):
Thanks! By the way, what are the `<internals>` for in the first place?

wrwg (approved):

rahxephon89:
> Thanks! By the way, what are the `<internals>` for in the first place?

A special constant location representing an opaque location, which does not make sense to appear in the trace.

PR #7675, opened 2023-04-11
[aptos-node] Set node version to 1.4.0

gregnazario:
### Description
This will help people identify releases by just typing `aptos-node -v` or `aptos-node --info`

There is often requested to get the "node version" direct from the binary.  Note that this should not be merged in until the 1.4.0 release

```
$ cargo run -p aptos-node -- --info
{
  "build_branch": "set-node-crate-version",
  "build_cargo_version": "cargo 1.66.1 (ad779e08b 2023-01-10)",
  "build_commit_hash": "011c118ca443627616ca58914e8aca5031d58cd5",
  "build_is_release_build": "false",
  "build_os": "macos-aarch64",
  "build_pkg_version": "1.4.0",
  "build_profile_name": "debug",
  "build_rust_channel": "1.66.1-aarch64-apple-darwin",
  "build_rust_version": "rustc 1.66.1 (90743e729 2023-01-10)",
  "build_tag": "",
  "build_time": "2023-04-10 20:46:27 -07:00",
  "build_using_tokio_unstable": "true"
}

$ cargo run -p aptos-node -- -V
Aptos Node 1.4.0
```


aptos-node/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "aptos-node"
 description = "Aptos node"
-version = "0.1.0"
+version = "1.4.0"
```
JoshLind, line 4: Worth calling out that once we do this, we're promising to maintain correctness going forward, for all releases and hotfixes, right? Are we wanting to make that promise -- I worry that this is going to become another thing that we'll forget to update on releases 😢  We don't think `./aptos-node --info` is enough?

gregnazario: I leave it to @perryjrandall his thoughts on the releases.

I already do this for the CLI and the SDK for packaged releases every time, and it's fairly standard

perryjrandall (approved):
Thank you greg!

davidiw (approved):

PR #7676, opened 2023-04-11
[mempool] don't remove from mempool on sequence number too new

bchocho:
### Description

When execution fails with sequence number too low, this txn shouldn't be removed from mempool. This helps if the txn is too new because of consensus-internal operations, e.g., if a block execution is cut due to gas limit. Worst case, it will be unable to be ordered again until the sequence number is filled.

### Test Plan
Existing tests.

### DIFF
--- a/consensus/src/counters.rs
+++ b/consensus/src/counters.rs
@@ -12,11 +12,11 @@ use aptos_metrics_core::{
 use once_cell::sync::Lazy;
 
 /// Transaction commit was successful
-pub const TXN_COMMIT_SUCCESS: &str = "success";
+pub const TXN_COMMIT_SUCCESS_LABEL: &str = "success";
 /// Transaction commit failed (will not be retried)
-pub const TXN_COMMIT_FAILED: &str = "failed";
+pub const TXN_COMMIT_FAILED_LABEL: &str = "failed";
 /// Transaction commit was unsuccessful, but will be retried
-pub const TXN_COMMIT_RETRY: &str = "retry";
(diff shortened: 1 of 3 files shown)


zekun000 (approved):
we should also update counter here https://github.com/aptos-labs/aptos-core/blob/69dc65745d29b493cff7ce5531165d22705e2f57/consensus/src/block_storage/block_store.rs#L70

to reflect that in retry not failed

consensus/src/counters.rs
```diff
@@ -11,6 +11,10 @@ use aptos_metrics_core::{
 };
 use once_cell::sync::Lazy;
 
+pub const TXN_COMMIT_SUCCESS: &str = "success";
```
JoshLind, line 14: nit: I'd probably add something like `LABEL` or `STATUS` to the name of these constants because when I read them from the other file, I thought they were counters 😄 e.g., `counters::TXN_COMMIT_SUCCESS` reads very much like a counter and not a string.

mempool/src/core_mempool/mempool.rs
```diff
@@ -65,28 +66,44 @@ impl Mempool {
             .commit_transaction(sender, sequence_number);
     }
 
-    pub(crate) fn reject_transaction(
-        &mut self,
+    #[inline]
```
JoshLind, line 69: Silly question: why inline?

JoshLind (approved):

mempool/src/core_mempool/mempool.rs
bchocho: Nice catch. Not sure why I added that 🙈

PR #7683, opened 2023-04-11
Check CoinType is initialized in register

movekevin:
### Description
This prevents accounts from registering arbitrary types that are not valid Coins.

### Test Plan
Unit tests

### DIFF
--- a/aptos-move/framework/aptos-framework/sources/coin.move
+++ b/aptos-move/framework/aptos-framework/sources/coin.move
@@ -536,6 +536,8 @@ module aptos_framework::coin {
     }
 
     public fun register<CoinType>(account: &signer) {
+        assert!(is_coin_initialized<CoinType>(), error::invalid_argument(ECOIN_INFO_NOT_PUBLISHED));
+
         let account_addr = signer::address_of(account);
         // Short-circuit and do nothing if account is already registered for CoinType.
         if (is_account_registered<CoinType>(account_addr)) {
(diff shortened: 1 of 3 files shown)


aptos-move/framework/aptos-framework/sources/coin.move
```diff
@@ -497,6 +497,8 @@ module aptos_framework::coin {
     }
 
     public fun register<CoinType>(account: &signer) {
+        assert!(is_coin_initialized<CoinType>(), error::invalid_argument(ECOIN_INFO_NOT_PUBLISHED));
```
gregnazario, line 500: Doesn't this serialize registers on looking up if the coin info exists?

gregnazario (approved):

lightmark (approved):
It's funny...

aptos-move/framework/aptos-framework/sources/coin.move
davidiw: just to be clear, we're going to add in a new read resource for each time we create an account or initialize a coin

I think that's probably why it was left out

movekevin: This shouldn't a problem now with reduced gas

movekevin: Based on the land-blocking forge test, there doesn't seem to be any execution degradation that affects P2P tx throughput (relatively to the test on another recent PR)

davidiw:
Let's not leave fully accepted PRs open for multiple months. Closing out, but feel free to re-open.

davidiw:
Note, this will require a bit more plumbing to actually make work.

PR #7692, opened 2023-04-11
[State Sync] Add sanity checks to storage requests.

JoshLind:
### Description
This PR adds basic sanity checks to the server-side of the storage service. Specifically, we verify all requests are serviceable before handling them (i.e., the data held by the server is a superset of the data being requested).

Note:
- These checks are best-effort only. There are some request parameters that state sync cannot verify without going directly to storage (e.g., in the case of fast sync indices), so we still need to be prepared to handle bad storage requests on the database side.
- Most of this PR is just new unit tests 😄 

### Test Plan
Existing and new tests.

### DIFF
--- a/state-sync/storage-service/server/src/handler.rs
+++ b/state-sync/storage-service/server/src/handler.rs
@@ -108,23 +108,8 @@ impl<T: StorageReaderInterface> Handler<T> {
             request.get_label(),
         );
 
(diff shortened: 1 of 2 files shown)


msmouse (approved):
seems legit ><

brianolson (approved):

PR #7693, opened 2023-04-12
[CLI] DRY out repetitive pattern for profile/submit txn

alnoki:
@davidiw @gregnazario the proposed PR abstracts a common pattern in the CLI in accordance with [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) design principles.

### DIFF
--- a/crates/aptos/src/common/utils.rs
+++ b/crates/aptos/src/common/utils.rs
@@ -2,7 +2,10 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use crate::{
-    common::types::{account_address_from_public_key, CliError, CliTypedResult, PromptOptions},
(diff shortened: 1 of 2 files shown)


crates/aptos/src/common/utils.rs
```diff
@@ -443,3 +449,20 @@ pub fn start_logger() {
     logger.channel_size(1000).is_async(false).level(Level::Warn);
     logger.build();
 }
+
+/// For transaction payload and options, either get gas profile or submit for execution.
+pub async fn profile_or_submit(
```
davidiw, line 454: Potentially makes sense for this to be a method on TransactionOptons

davidiw (approved):

PR #7694, opened 2023-04-12
[CLI] Extend vanity address support to multisig

alnoki:
@gregnazario @movekevin this PR extends CLI vanity address support to multisig accounts.

Closes #7569 since non-vanity-address functionality from that PR will be covered in #7672.


crates/aptos/src/op/key.rs
```diff
@@ -208,25 +221,34 @@ impl CliCommand<HashMap<&'static str, PathBuf>> for GenerateKey {
...
-                    .to_hex_literal();
-                    // Put account address in PathBuf so it can be displayed from typed CLI result.
-                    result_map.insert("Account Address:", PathBuf::from(account_address));
+                    );
+                    // Store account address in a PathBuf so it can be displayed in CLI result.
+                    result_map.insert(
+                        "Account Address:",
+                        PathBuf::from(account_address.to_hex_literal()),
+                    );
+                    if self.vanity_multisig {
```
davidiw, line 244: why are we adding this if multisig is true?

davidiw (commented):
would be good to see the input / output
besides comment, seems fine!

crates/aptos/src/op/key.rs
movekevin: Yeah the flow for multisig here is quite strange, as it technically doesn't create the multisig account itself. The user would then need to create one later with the right seed (0).

davidiw (approved):

crates/aptos/src/common/utils.rs
```diff
@@ -340,19 +344,22 @@ pub fn generate_vanity_account_ed25519(
...
-    while !account_address_from_public_key(&Ed25519PublicKey::from(&private_key))
-        .short_str_lossless()
-        .starts_with(vanity_prefix_ref)
-    {
-        // Generate a new account.
-        private_key = key_generator.generate_ed25519_private_key();
+    let mut key_generator = KeyGen::from_os_rng(); // Get random key generator.
+    loop {
+        // Generate new keys until finding a match against the vanity prefix.
+        let private_key = key_generator.generate_ed25519_private_key();
```
movekevin, line 350: Is this susceptible at all to people getting the same key when asking for the same prefix? This of course highly depend on the seed used in the generation process but I wonder if this search adds bias here that can increase the probability

alnoki: The random key generator used here is the same as that used for standard key generation.

Hence the probability of a "collision" is the same as for using the CLI for key generation without a prefix.

alnoki:
@movekevin @gregnazario is there anything else needed right now to land this?

movekevin (approved):

alnoki:
@davidiw it looks like this PR has all required approving reviews, are you ready to land it?

PR #7696, opened 2023-04-12
[framework] Create release config for 1.4

runtian-zhou:
### Description

Updated the yaml file to track release for 1.4. A few modifications were made in the PR:
1. The validate-proposal command no longer mint to the validator. Tests now need to manually mint to validators before executing the proposals. This allows the tooling to better mimic what the release process will look like in real life.
2. Added the full 1.4 release configuration. Tentatively we will have five different goverance proposals to vote. We are still waiting on finalizing the fifth one to enable transaction reshuffling cc @zekun000 
3. Change the release tooling to be able to handle multiple proposals in one release conifg.
4.  Add the ability to include a customize script for the release builder

### Test Plan

Framework upgrade test

### DIFF
--- a/aptos-move/aptos-release-builder/data/release.yaml
+++ b/aptos-move/aptos-release-builder/data/release.yaml
@@ -1,44 +1,77 @@
 ---
-testnet: true
 remote_endpoint: ~
-is_multi_step: true
-update_sequence:
-  - Framework:
-      bytecode_version: 6
-      git_hash: ~
-  - DefaultGas
-  - FeatureFlag:
-      enabled:
-        - code_dependency_check
-        - treat_friend_as_private
-        - sha512_and_ripe_md160_natives
-        - aptos_std_chain_id_natives
-        - v_m_binary_format_v6
-        - multi_ed25519_pk_validate_v2_natives
-        - blake2b256_native
-        - resource_groups
-        - multisig_accounts
-        - delegation_pools
-        - cryptography_algebra_natives
-        - bls12381_structures
-        - ed25519_pubkey_validate_return_false_wrong_length
-        - struct_constructors
-      disabled: []
-  - Consensus:
-      V1:
-        decoupled_execution: true
-        back_pressure_limit: 10
-        exclude_round: 40
-        proposer_election_type:
-          leader_reputation:
-            proposer_and_voter_v2:
-              active_weight: 1000
-              inactive_weight: 10
-              failed_weight: 1
-              failure_threshold_percent: 10
-              proposer_window_num_validators_multiplier: 10
-              voter_window_num_validators_multiplier: 1
-              weight_by_voting_power: true
-              use_history_from_previous_epoch_max_count: 5
-        max_failed_authors_to_store: 10
-
+proposals:
+  - name: step_1
+    metadata:
+      title: "Enable Struct Constructor Feature"
+      description: "TBD"
+      source_code_url: "TBD"
+      discussion_url: "TBD"
+    execution_mode: MultiStep
+    update_sequence:
+    - FeatureFlag:
+        enabled:
+          - struct_constructors
+        disabled: []
+  - name: step_2
+    metadata:
+      title: "Update to latest framework"
+      description: "TBD"
+      source_code_url: "TBD"
+      discussion_url: "TBD"
+    execution_mode: MultiStep
+    update_sequence:
+      - DefaultGas
+      - Framework:
+          bytecode_version: 6
+          git_hash: ~
+  - name: step_3
+    metadata:
+      title: "Enable bls12381"
+      description: "TBD"
+      source_code_url: "TBD"
+      discussion_url: "TBD"
+    execution_mode: MultiStep
+    update_sequence:
+    - FeatureFlag:
+        enabled:
+          - cryptography_algebra_natives
+          - bls12381_structures
+        disabled: []
+  - name: step_4
+    metadata:
+      title: "Enable ed25519_pubkey_validate_return_false_wrong_length"
+      description: "TBD"
+      source_code_url: "TBD"
+      discussion_url: "TBD"
+    execution_mode: MultiStep
+    update_sequence:
+    - FeatureFlag:
+        enabled:
+          - ed25519_pubkey_validate_return_false_wrong_length
+        disabled: []
+  - name: step_5
+    metadata:
+      title: "Enable Quorum Store"
+      description: "TBD"
+      source_code_url: "TBD"
+      discussion_url: "TBD"
+    execution_mode: MultiStep
+    update_sequence:
+    - Consensus:
+        V2:
+          decoupled_execution: true
+          back_pressure_limit: 10
+          exclude_round: 40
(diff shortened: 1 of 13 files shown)


davidiw (commented):
is our plan moving forward to start checking this into the code? why not keep release.yaml?

aptos-move/aptos-release-builder/data/release-v1.4.yaml
```diff
@@ -0,0 +1,23 @@
...
+testnet: true
+remote_endpoint: ~
+is_multi_step: true
+update_sequence:
+  - Framework:
+      bytecode_version: 6
+      git_hash: ~
+  - FeatureFlag:
+      enabled:
+        - code_dependency_check
```
movekevin, line 11: Why do we list all of these features out? We only need

ed25519_pubkey_validate_return_false_wrong_length
struct_constructors

aptos-move/aptos-release-builder/data/release-v1.4.yaml
```diff
@@ -0,0 +1,23 @@
+---
+testnet: true
+remote_endpoint: ~
+is_multi_step: true
+update_sequence:
```
movekevin, line 5: Also need the new consensus config for quorum store (if blockchain side is ready to launch) and the feature flags for tx shuffling, and any other missing items on the release sheet

movekevin:
> is our plan moving forward to start checking this into the code? why not keep release.yaml?

The problem with this current approach is we need to remember to move new release steps into a new file (e.g. v1.5) after a branch cut. The release tooling generates all steps from the yaml file regardless of if those steps have already been executed

aptos-move/aptos-release-builder/data/release-v1.4.yaml
runtian-zhou: Can you help confirm what we need for quorum store? cc @igor-aptos

aptos-move/aptos-release-builder/data/release-v1.4.yaml
sherry-x: did we change to create new yaml file for every releases? I thought we'd keep one release.yaml file and always use it as source of truth.

aptos-move/aptos-release-builder/src/lib.rs
```diff
@@ -7,7 +7,8 @@ pub mod validate;
 pub use components::{ReleaseConfig, ReleaseEntry};
 use once_cell::sync::Lazy;
 
-const RELEASE_CONFIG: &str = include_str!("../data/release.yaml");
+// Update me after branch cut.
+const RELEASE_CONFIG: &str = include_str!("../data/release-v1.4.yaml");
```
sherry-x, line 11: this is not great that we need to update the file name after the branch cut. maybe, what we can do is still use release.yaml as the latest release file, and move the old stuff to release-v1.x.yaml for tracking purpose for now?

aptos-move/aptos-release-builder/data/release-v1.4.yaml
```diff
@@ -6,19 +6,8 @@ update_sequence:
...
-        - code_dependency_check
-        - treat_friend_as_private
-        - sha512_and_ripe_md160_natives
-        - aptos_std_chain_id_natives
-        - v_m_binary_format_v6
-        - multi_ed25519_pk_validate_v2_natives
-        - blake2b256_native
-        - resource_groups
-        - multisig_accounts
-        - delegation_pools
```
zekun000, line 21: hmm, this looks suspicious, is it because those feature flags being removed i.e. default enabled?

aptos-move/aptos-release-builder/src/lib.rs
perryjrandall: To second this, its much easier to automate clearing and rotating the old version (changing yaml files)

Updating a const in Rust file is pretty fragile and would be less certain / automatable than rotating a yaml file

Thinking on branch cut, we would just need to rotate / clear the yaml on main, instead of changing the rust as well

aptos-move/aptos-release-builder/src/components/mod.rs
```diff
@@ -44,16 +58,21 @@ pub enum ReleaseEntry {
...
     pub fn generate_release_script(
         &self,
         client: Option<&Client>,
         result: &mut Vec<(String, String)>,
-        is_testnet: bool,
-        is_multi_step: bool,
+        execution_mode: ExecutionMode,
     ) -> Result<()> {
+        let (is_testnet, is_multi_step) = match execution_mode {
+            ExecutionMode::MultiStep => (false, true),
```
perryjrandall, line 72: Can we deprecate or make automatic multi vs single step?

Can the tool just know whether to execute multi or single

perryjrandall (approved):

aptos-move/aptos-release-builder/data/release.yaml
```diff
@@ -1,44 +1,50 @@
...
+  - name: v1.4.1
+    description: Update to latest framework
+    execution_mode: MultiStep
+    update_sequence:
+      - DefaultGas
+      - Framework:
+          bytecode_version: 6
+          git_hash: ~
+  - name: v1.4.2
+    description: Enable ed25519_pubkey_validate_return_false_wrong_length
```
sherry-x, line 21: Can we check if those 3 flags are 3 different AIP or all in one AIP? we should have 1 proposal per AIP

aptos-move/aptos-release-builder/src/components/mod.rs
runtian-zhou: I would like to keep them manual for now just to see how we are planning to use this feature in the longer run. It seems that we always need multi-step for framework upgrade but for others, if there's only one entry we can use single step.

aptos-move/aptos-release-builder/data/release.yaml
```diff
@@ -1,44 +1,50 @@
...
-              proposer_window_num_validators_multiplier: 10
-              voter_window_num_validators_multiplier: 1
-              weight_by_voting_power: true
-              use_history_from_previous_epoch_max_count: 5
-        max_failed_authors_to_store: 10
-
+proposals:
+  - name: v1.4.0
+    description: Enable Struct Constructor Feature
+    execution_mode: SingleStep
```
movekevin, line 6: Let's make everything multistep so it uses the right create proposal function even if the proposal only has a single step. This will allow us to delete the single vs multiple step distinction everywhere in the code

sherry-x (approved):

aptos-move/aptos-release-builder/data/release.yaml
```diff
@@ -1,44 +1,77 @@
...
+        disabled: []
+  - name: step_2
+    metadata:
+      title: "Update to latest framework"
+      description: "TBD"
+      source_code_url: "TBD"
+      discussion_url: "TBD"
+    execution_mode: MultiStep
+    update_sequence:
+      - DefaultGas
```
movekevin, line 24: Shouldn't gas schedule happen after the framework upgrade? cc @vgao1996

PR #7699, opened 2023-04-12
[storage] Restore state kv in historical range

areshand:
### Requirements

To restore a database, provide a start version and end version with:

- All data restored, including state key values, write sets, state Merkle tree at the end version, and ledger info from genesis to generate a database with full data of a certain time frame.
- Fast, different from replaying transactions, decouple state key value commit from in-memory state Merkle tree updating.

### Approach

Enhance the existing DB bootstrap to support both the latest DB restore and archive DB restore.

The user provides the arguments in the CLI with (ledger_history_start_version, target_version), specifying the range of the historical database they want to restore.

Depending on the snapshot version's order with user provided arguments, two scenarios may occur:

**Scenario 1: [snapshot_a, lhs, snapshot_b, target]**

1. Calculate `snapshot_a`, restore the snapshot (only KVs are required) at `snapshot_a` to create a baseline.
2. Restore transactions in range [snapshot_a, snapshot_b].
    1. directly commit the write set as key value to KV CF
    2. snapshot_a and snapshot_b are inclusive since we need to restore txn Info at these two versions. 
3. Restore the snapshot at `snapshot_b` with both kv and tree.
4. Replay transactions from `snapshot_b` + 1 to `target`. 

**Scenario 2: [snapshot_a, lhs, target]**

1. Calculate `snapshot_a` and `snapshot_b` In this scenario, snapshot_a == snapshot_b
2. Directly replay transactions from `snapshot_a + 1` till `target`.

Lastly, if the user just wants to bootstrap the database to the latest version, lhs can be set to the same target version. We will just replay from the latest snapshot version to the latest target version

**Enable `TransactionRestoreBatchController`(TxnController) to restore with a start version**

We want to replay transactions between [snapshot_b, target]. However, the TxnController loads all transactions in a backup chunk file, which can contain more transactions than we need. These transaction are saved by default causing waste of computation. TxnController only support `target_version` to specify the range end. We need a `start_version` parameter to specify the exact range of Txns to avoid redundant computation

**Enable `StateSnapshotRestoreController` to restore KV only**

To restore `snapshot_a`, we do not need to restore the state tree since the tree will eventually be restored at `snapshot_b`. We can bypass the tree restore in the SnapshotController by disabling the thread for restoring the state tree using a flag.

Design doc here https://www.notion.so/aptoslabs/Enhanced-DB-Restore-cf86fbfacbb94eab8e155078f56cec34?pvs=4

### Test Plan
add a new unit test showing the KV restored, Tree restored only at the 2nd snapshot.

### DIFF
--- a/storage/aptosdb/src/state_restore/mod.rs
+++ b/storage/aptosdb/src/state_restore/mod.rs
@@ -14,7 +14,7 @@ use aptos_types::{
     transaction::Version,
 };
 use serde::{Deserialize, Serialize};
-use std::{collections::HashMap, hash::Hash, sync::Arc};
+use std::{collections::HashMap, hash::Hash, str::FromStr, sync::Arc};
 
 #[cfg(test)]
 mod restore_test;
@@ -49,6 +49,38 @@ pub trait StateValueWriter<K, V>: Send + Sync {
     fn get_progress(&self, version: Version) -> Result<Option<StateSnapshotProgress>>;
 }
 
+#[derive(Clone, Copy, Deserialize, Serialize)]
+pub enum StateSnapshotRestoreMode {
+    /// Restore both KV and Tree by default
+    Default,
+    /// Only restore the state KV
+    KvOnly,
+    /// Only restore the state tree
+    TreeOnly,
+}
+
+impl Default for StateSnapshotRestoreMode {
+    fn default() -> Self {
+        Self::Default
+    }
+}
+
+impl FromStr for StateSnapshotRestoreMode {
+    type Err = anyhow::Error;
+
+    fn from_str(s: &str) -> Result<Self> {
+        match s {
+            "default" => Ok(Self::Default),
+            "kv_only" => Ok(Self::KvOnly),
+            "tree_only" => Ok(Self::TreeOnly),
+            _ => Err(anyhow::anyhow!(
+                "Invalid state snapshot restore mode: {}",
+                s
+            )),
+        }
+    }
+}
+
 struct StateValueRestore<K, V> {
     version: Version,
     db: Arc<dyn StateValueWriter<K, V>>,
@@ -59,7 +91,11 @@ impl<K: Key + CryptoHash + Eq + Hash, V: Value> StateValueRestore<K, V> {
         Self { version, db }
     }
 
-    pub fn add_chunk(&mut self, mut chunk: Vec<(K, V)>) -> Result<()> {
+    pub fn add_chunk(
+        &mut self,
+        mut chunk: Vec<(K, V)>,
+        restore_mode: StateSnapshotRestoreMode,
+    ) -> Result<()> {
         // load progress
         let progress_opt = self.db.get_progress(self.version)?;
 
@@ -81,14 +117,19 @@ impl<K: Key + CryptoHash + Eq + Hash, V: Value> StateValueRestore<K, V> {
         let mut usage = progress_opt.map_or(StateStorageUsage::zero(), |p| p.usage);
         let (last_key, _last_value) = chunk.last().unwrap();
         let last_key_hash = CryptoHash::hash(last_key);
-        let kv_batch = chunk
-            .into_iter()
-            .map(|(k, v)| {
-                usage.add_item(k.key_size() + v.value_size());
-                ((k, self.version), Some(v))
-            })
-            .collect();
 
+        // In case of TreeOnly Restore, we only restore the usage of KV without actually writing KV into DB
+        for (k, v) in chunk.iter() {
+            usage.add_item(k.key_size() + v.value_size());
+        }
+
+        let kv_batch: StateValueBatch<K, Option<V>> = match restore_mode {
+            StateSnapshotRestoreMode::TreeOnly => StateValueBatch::new(),
+            _ => chunk
+                .into_iter()
+                .map(|(k, v)| ((k, self.version), Some(v)))
+                .collect(),
+        };
         self.db.write_kv_batch(
             self.version,
             &kv_batch,
@@ -115,6 +156,7 @@ impl<K: Key + CryptoHash + Eq + Hash, V: Value> StateValueRestore<K, V> {
 pub struct StateSnapshotRestore<K, V> {
     tree_restore: Arc<Mutex<Option<JellyfishMerkleRestore<K>>>>,
     kv_restore: Arc<Mutex<Option<StateValueRestore<K, V>>>>,
+    restore_mode: StateSnapshotRestoreMode,
 }
 
 impl<K: Key + CryptoHash + Hash + Eq, V: Value> StateSnapshotRestore<K, V> {
@@ -124,6 +166,7 @@ impl<K: Key + CryptoHash + Hash + Eq, V: Value> StateSnapshotRestore<K, V> {
         version: Version,
         expected_root_hash: HashValue,
         async_commit: bool,
+        restore_mode: StateSnapshotRestoreMode,
     ) -> Result<Self> {
         Ok(Self {
             tree_restore: Arc::new(Mutex::new(Some(JellyfishMerkleRestore::new(
@@ -136,6 +179,7 @@ impl<K: Key + CryptoHash + Hash + Eq, V: Value> StateSnapshotRestore<K, V> {
                 Arc::clone(value_store),
                 version,
             )))),
+            restore_mode,
         })
     }
 
@@ -144,6 +188,7 @@ impl<K: Key + CryptoHash + Hash + Eq, V: Value> StateSnapshotRestore<K, V> {
         value_store: &Arc<S>,
         version: Version,
         expected_root_hash: HashValue,
+        restore_mode: StateSnapshotRestoreMode,
     ) -> Result<Self> {
         Ok(Self {
             tree_restore: Arc::new(Mutex::new(Some(JellyfishMerkleRestore::new_overwrite(
@@ -155,6 +200,7 @@ impl<K: Key + CryptoHash + Hash + Eq, V: Value> StateSnapshotRestore<K, V> {
                 Arc::clone(value_store),
                 version,
             )))),
+            restore_mode,
         })
     }
 
@@ -191,45 +237,76 @@ impl<K: Key + CryptoHash + Hash + Eq, V: Value> StateSnapshotReceiver<K, V>
     for StateSnapshotRestore<K, V>
 {
     fn add_chunk(&mut self, chunk: Vec<(K, V)>, proof: SparseMerkleRangeProof) -> Result<()> {
-        let _timer = OTHER_TIMERS_SECONDS
-            .with_label_values(&["state_snapshot_add_chunk"])
-            .start_timer();
+        let kv_fn = || {
+            let _timer = OTHER_TIMERS_SECONDS
+                .with_label_values(&["state_value_add_chunk"])
+                .start_timer();
+            self.kv_restore
+                .lock()
+                .as_mut()
+                .unwrap()
+                .add_chunk(chunk.clone(), self.restore_mode)
+        };
+
+        let tree_fn = || {
+            let _timer = OTHER_TIMERS_SECONDS
+                .with_label_values(&["jmt_add_chunk"])
+                .start_timer();
+            self.tree_restore
+                .lock()
+                .as_mut()
+                .unwrap()
+                .add_chunk_impl(chunk.iter().map(|(k, v)| (k, v.hash())).collect(), proof)
+        };
         // Write KV out first because we are likely to resume according to the rightmost key in the
         // tree after crashing.
-        let (r1, r2) = IO_POOL.join(
-            || {
-                let _timer = OTHER_TIMERS_SECONDS
-                    .with_label_values(&["state_value_add_chunk"])
-                    .start_timer();
-                self.kv_restore
-                    .lock()
-                    .as_mut()
-                    .unwrap()
-                    .add_chunk(chunk.clone())
+        match self.restore_mode {
+            StateSnapshotRestoreMode::KvOnly => kv_fn()?,
+            StateSnapshotRestoreMode::TreeOnly => {
+                // We run kv_fn with TreeOnly to restore the usage of DB
+                let (r1, r2) = IO_POOL.join(kv_fn, tree_fn);
+                r1?;
+                r2?;
             },
-            || {
-                let _timer = OTHER_TIMERS_SECONDS
-                    .with_label_values(&["jmt_add_chunk"])
-                    .start_timer();
-                self.tree_restore
-                    .lock()
-                    .as_mut()
-                    .unwrap()
-                    .add_chunk_impl(chunk.iter().map(|(k, v)| (k, v.hash())).collect(), proof)
+            StateSnapshotRestoreMode::Default => {
+                let (r1, r2) = IO_POOL.join(kv_fn, tree_fn);
+                r1?;
+                r2?;
             },
-        );
-        r1?;
-        r2?;
+        }
+
         Ok(())
     }
 
     fn finish(self) -> Result<()> {
-        self.kv_restore.lock().take().unwrap().finish()?;
-        self.tree_restore.lock().take().unwrap().finish_impl()
+        match self.restore_mode {
+            StateSnapshotRestoreMode::KvOnly => self.kv_restore.lock().take().unwrap().finish()?,
+            StateSnapshotRestoreMode::TreeOnly => {
+                // for tree only mode, we also need to write the usage to DB
+                self.kv_restore.lock().take().unwrap().finish()?;
+                self.tree_restore.lock().take().unwrap().finish_impl()?
+            },
+            StateSnapshotRestoreMode::Default => {
+                self.kv_restore.lock().take().unwrap().finish()?;
+                self.tree_restore.lock().take().unwrap().finish_impl()?;
+            },
+        }
+        Ok(())
     }
 
     fn finish_box(self: Box<Self>) -> Result<()> {
-        self.kv_restore.lock().take().unwrap().finish()?;
-        self.tree_restore.lock().take().unwrap().finish_impl()
+        match self.restore_mode {
+            StateSnapshotRestoreMode::KvOnly => self.kv_restore.lock().take().unwrap().finish()?,
+            StateSnapshotRestoreMode::TreeOnly => {
+                // for tree only mode, we also need to write the usage to DB
+                self.kv_restore.lock().take().unwrap().finish()?;
+                self.tree_restore.lock().take().unwrap().finish_impl()?
+            },
+            StateSnapshotRestoreMode::Default => {
+                self.kv_restore.lock().take().unwrap().finish()?;
+                self.tree_restore.lock().take().unwrap().finish_impl()?;
+            },
+        }
+        Ok(())
     }
 }
--- a/storage/aptosdb/src/state_restore/mod.rs
+++ b/storage/aptosdb/src/state_restore/mod.rs
@@ -84,18 +84,24 @@ impl FromStr for StateSnapshotRestoreMode {
 struct StateValueRestore<K, V> {
     version: Version,
     db: Arc<dyn StateValueWriter<K, V>>,
+    restore_mode: StateSnapshotRestoreMode,
 }
 
 impl<K: Key + CryptoHash + Eq + Hash, V: Value> StateValueRestore<K, V> {
-    pub fn new<D: 'static + StateValueWriter<K, V>>(db: Arc<D>, version: Version) -> Self {
-        Self { version, db }
+    pub fn new<D: 'static + StateValueWriter<K, V>>(
+        db: Arc<D>,
+        version: Version,
+        restore_mode_opt: Option<StateSnapshotRestoreMode>,
+    ) -> Self {
+        let restore_mode = restore_mode_opt.unwrap_or_default();
+        Self {
+            version,
+            db,
+            restore_mode,
+        }
     }
 
-    pub fn add_chunk(
-        &mut self,
-        mut chunk: Vec<(K, V)>,
-        restore_mode: StateSnapshotRestoreMode,
-    ) -> Result<()> {
+    pub fn add_chunk(&mut self, mut chunk: Vec<(K, V)>) -> Result<()> {
         // load progress
         let progress_opt = self.db.get_progress(self.version)?;
 
@@ -123,7 +129,7 @@ impl<K: Key + CryptoHash + Eq + Hash, V: Value> StateValueRestore<K, V> {
             usage.add_item(k.key_size() + v.value_size());
         }
 
-        let kv_batch: StateValueBatch<K, Option<V>> = match restore_mode {
+        let kv_batch: StateValueBatch<K, Option<V>> = match self.restore_mode {
             StateSnapshotRestoreMode::TreeOnly => StateValueBatch::new(),
             _ => chunk
                 .into_iter()
@@ -178,6 +184,7 @@ impl<K: Key + CryptoHash + Hash + Eq, V: Value> StateSnapshotRestore<K, V> {
             kv_restore: Arc::new(Mutex::new(Some(StateValueRestore::new(
                 Arc::clone(value_store),
                 version,
+                Some(restore_mode),
             )))),
             restore_mode,
         })
@@ -199,6 +206,7 @@ impl<K: Key + CryptoHash + Hash + Eq, V: Value> StateSnapshotRestore<K, V> {
             kv_restore: Arc::new(Mutex::new(Some(StateValueRestore::new(
                 Arc::clone(value_store),
                 version,
+                Some(restore_mode),
             )))),
             restore_mode,
         })
@@ -245,7 +253,7 @@ impl<K: Key + CryptoHash + Hash + Eq, V: Value> StateSnapshotReceiver<K, V>
                 .lock()
                 .as_mut()
                 .unwrap()
-                .add_chunk(chunk.clone(), self.restore_mode)
+                .add_chunk(chunk.clone())
         };
 
         let tree_fn = || {
@@ -262,13 +270,8 @@ impl<K: Key + CryptoHash + Hash + Eq, V: Value> StateSnapshotReceiver<K, V>
         // tree after crashing.
         match self.restore_mode {
             StateSnapshotRestoreMode::KvOnly => kv_fn()?,
-            StateSnapshotRestoreMode::TreeOnly => {
+            StateSnapshotRestoreMode::TreeOnly | StateSnapshotRestoreMode::Default => {
                 // We run kv_fn with TreeOnly to restore the usage of DB
-                let (r1, r2) = IO_POOL.join(kv_fn, tree_fn);
-                r1?;
-                r2?;
-            },
-            StateSnapshotRestoreMode::Default => {
                 let (r1, r2) = IO_POOL.join(kv_fn, tree_fn);
                 r1?;
                 r2?;
@@ -281,14 +284,10 @@ impl<K: Key + CryptoHash + Hash + Eq, V: Value> StateSnapshotReceiver<K, V>
     fn finish(self) -> Result<()> {
         match self.restore_mode {
             StateSnapshotRestoreMode::KvOnly => self.kv_restore.lock().take().unwrap().finish()?,
-            StateSnapshotRestoreMode::TreeOnly => {
+            StateSnapshotRestoreMode::TreeOnly | StateSnapshotRestoreMode::Default => {
                 // for tree only mode, we also need to write the usage to DB
                 self.kv_restore.lock().take().unwrap().finish()?;
                 self.tree_restore.lock().take().unwrap().finish_impl()?
-            },
-            StateSnapshotRestoreMode::Default => {
-                self.kv_restore.lock().take().unwrap().finish()?;
-                self.tree_restore.lock().take().unwrap().finish_impl()?;
             },
         }
         Ok(())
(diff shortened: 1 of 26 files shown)


storage/aptosdb/src/state_store/mod.rs
```diff
@@ -569,7 +572,50 @@ impl StateStore {
...
+            .start_timer();
+        // obtain the current state
+        let current_state = self.buffered_state.lock().current_state().current.clone();
+        // get a cached state view
+        let state_view = CachedStateView::new(
+            StateViewId::Miscellaneous,
+            self.state_db.clone(),
+            first_version, // ensure the version is state_view is before the first_verson to write
+            current_state,
+            Arc::new(SyncProofFetcher::new(self.state_db.clone())),
```
msmouse: `BufferedState` and `CachedStateView` probably aren't the correct data structures to use here.. they are tools around the SMT. 

Could use things like this https://github.com/aptos-labs/aptos-core/blob/091538ffca2c55e267b08768fbf864f6ce3faacf/storage/storage-interface/src/state_view.rs#L61

execution/executor-types/src/in_memory_state_calculator.rs
```diff
@@ -199,6 +199,7 @@ impl InMemoryStateCalculator {
             .iter()
             .map(|(key, value)| (key.hash(), value.as_ref()))
             .collect();
+        // TODO(bowu): this should only be a subtree of SMT? consider delay SMT update to accelerate restoring
```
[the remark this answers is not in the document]
areshand: sorry, this is in old version of PR. removed this in the latest version. There are multiple commits. will squash them once we finish the review.

storage/backup/backup-cli/src/coordinators/restore.rs
```diff
@@ -124,74 +130,168 @@ impl RestoreCoordinator {
...
+                version = version,
+                "Found in progress state snapshot restore",
+            );
+            ensure!(snapshot_before_lhs.version == version, "Inconsistent in-progress state snapshot version, please delete the data folder and try again.");
+        }
+        info!(target_version = target_version, "Restore target decided.");
+        let transaction_backups = metadata_view
+            .select_transaction_backups(snapshot_before_lhs.version, target_version)?;
+        let epoch_ending_backups = metadata_view.select_epoch_ending_backups(target_version)?;
+        // This is used to t
```
msmouse, line 171: incomplete comment

storage/aptosdb/src/state_store/mod.rs
```diff
@@ -570,6 +574,33 @@ impl StateStore {
...
+            .iter()
+            .map(|ws| {
+                ws.iter()
+                    .map(|(key, value)| (key.clone(), value.as_state_value()))
+                    .collect::<HashMap<StateKey, Option<StateValue>>>()
+            })
+            .collect();
+        let value_state_sets: Vec<&HashMap<StateKey, Option<StateValue>>> =
+            value_state_sets_raw.iter().collect();
+        self.put_state_value(value_state_sets, first_version, sharded_state_kv_batches)?;
```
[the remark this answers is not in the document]
areshand: added. previously under the impression that this information is not needed

storage/backup/backup-cli/src/backup_types/transaction/restore.rs
```diff
@@ -259,9 +263,11 @@ impl TransactionRestoreBatchController {
         }
 
         let mut loaded_chunk_stream = self.loaded_chunk_stream();
-        let first_version = self
-            .confirm_or_save_frozen_subtrees(&mut loaded_chunk_stream)
-            .await?;
+        // first_version determines the start version for saving txns
+        let first_version = self.first_version.unwrap_or(
+            self.confirm_or_save_frozen_subtrees(&mut loaded_chunk_stream)
```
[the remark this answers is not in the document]
areshand: If the first_version is None, we are still gone restore txn from the first version of the chunk.  Ex, restore DB to latest version while saving txn ws from all loaded chunks.

storage/backup/backup-cli/src/coordinators/restore.rs
```diff
@@ -124,74 +130,168 @@ impl RestoreCoordinator {
...
+            .max_transaction_version()?
+            .ok_or_else(|| anyhow!("No transaction backup found."))?;
+        let lhs = self.ledger_history_start_version();
+        let snapshot_before_lhs =
+            metadata_view.select_state_snapshot(std::cmp::min(lhs, max_txn_ver))?;
+
+        let snapshot_before_target = metadata_view
+            .select_state_snapshot(std::cmp::min(self.target_version(), max_txn_ver))?;
+        ensure!(
+            snapshot_before_lhs.is_some() && snapshot_before_target.is_some(),
```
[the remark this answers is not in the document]
areshand: when lhs == 0 or target ==0, snapshot_before_lhs and snapshot_before_target should be genesis snapshot instead of none?

storage/backup/backup-cli/src/coordinators/restore.rs
```diff
@@ -124,74 +130,168 @@ impl RestoreCoordinator {
             return Ok(());
```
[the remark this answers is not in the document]
areshand: deleted

storage/backup/backup-cli/src/coordinators/restore.rs
```diff
@@ -124,74 +130,168 @@ impl RestoreCoordinator {
...
+            snapshot_before_target.version,
+            snapshot_before_lhs.version
+        );
+
+        if let Some(version) = self.global_opt.run_mode.get_in_progress_state_snapshot()? {
+            info!(
+                version = version,
+                "Found in progress state snapshot restore",
+            );
+            ensure!(snapshot_before_lhs.version == version, "Inconsistent in-progress state snapshot version, please delete the data folder and try again.");
```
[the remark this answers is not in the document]
areshand: add resume from any failure point in middle

storage/aptosdb/src/state_restore/mod.rs
```diff
@@ -191,45 +245,63 @@ impl<K: Key + CryptoHash + Hash + Eq, V: Value> StateSnapshotReceiver<K, V>
...
-                    .start_timer();
-                self.tree_restore
-                    .lock()
-                    .as_mut()
-                    .unwrap()
-                    .add_chunk_impl(chunk.iter().map(|(k, v)| (k, v.hash())).collect(), proof)
+        match self.restore_mode {
+            StateSnapshotRestoreMode::KvOnly => kv_fn()?,
+            StateSnapshotRestoreMode::TreeOnly | StateSnapshotRestoreMode::Default => {
+                // We run kv_fn with TreeOnly to restore the usage of DB
```
msmouse, line 274: when in tree mode, the usage should've been done by previous operations already, right?

storage/aptosdb/src/state_restore/mod.rs
```diff
@@ -191,45 +245,63 @@ impl<K: Key + CryptoHash + Hash + Eq, V: Value> StateSnapshotReceiver<K, V>
...
         Ok(())
     }
 
     fn finish(self) -> Result<()> {
-        self.kv_restore.lock().take().unwrap().finish()?;
-        self.tree_restore.lock().take().unwrap().finish_impl()
+        match self.restore_mode {
+            StateSnapshotRestoreMode::KvOnly => self.kv_restore.lock().take().unwrap().finish()?,
+            StateSnapshotRestoreMode::TreeOnly | StateSnapshotRestoreMode::Default => {
+                // for tree only mode, we also need to write the usage to DB
```
msmouse, line 288: ditto

storage/aptosdb/src/state_restore/mod.rs
```diff
@@ -191,45 +245,63 @@ impl<K: Key + CryptoHash + Hash + Eq, V: Value> StateSnapshotReceiver<K, V>
...
+        Ok(())
     }
 
     fn finish_box(self: Box<Self>) -> Result<()> {
-        self.kv_restore.lock().take().unwrap().finish()?;
-        self.tree_restore.lock().take().unwrap().finish_impl()
+        match self.restore_mode {
+            StateSnapshotRestoreMode::KvOnly => self.kv_restore.lock().take().unwrap().finish()?,
+            StateSnapshotRestoreMode::TreeOnly | StateSnapshotRestoreMode::Default => {
+                // for tree only mode, we also need to write the usage to DB
```
msmouse, line 300: ditto

storage/aptosdb/src/state_store/mod.rs
```diff
@@ -595,6 +636,15 @@ impl StateStore {
             .with_label_values(&["add_state_kv_batch"])
             .start_timer();
 
+        self.put_state_value(value_state_sets, first_version, sharded_state_kv_batches)
+    }
+
+    pub fn put_state_value(
```
msmouse, line 642: unnecessary refactor? or rename: "values"

storage/backup/backup-cli/src/backup_types/transaction/restore.rs
```diff
@@ -216,6 +217,7 @@ pub struct TransactionRestoreBatchController {
     epoch_history: Option<Arc<EpochHistory>>,
     verify_execution_mode: VerifyExecutionMode,
     output_transaction_analysis: Option<PathBuf>,
+    first_version: Option<Version>,
```
msmouse, line 220: how about moving it up before `replay_from_version` (use ide refactor tool)

storage/backup/backup-cli/src/backup_types/transaction/restore.rs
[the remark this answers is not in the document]
msmouse: I see, I misread, I thought you always recover the frozen subtrees according to the chunk.first_version, but infact you skip it when there's self.first_versino.

But when first version is not None, we should also `confirm_or_save_frozen_subtrees`? otherwise the txn accumulator is not valid.

storage/backup/backup-cli/src/backup_types/transaction/restore.rs
```diff
@@ -474,7 +492,7 @@ impl TransactionRestoreBatchController {
         let replay_start = Instant::now();
         let db = DbReaderWriter::from_arc(Arc::clone(&restore_handler.aptosdb));
         let chunk_replayer = Arc::new(ChunkExecutor::<AptosVM>::new(db));
-
+        // Use stream to update the kv and SMT buffered state separately with a long buffer.
```
msmouse, line 495: I don't quite get what you are trying to say here..

storage/backup/backup-cli/src/coordinators/restore.rs
```diff
@@ -109,89 +106,204 @@ impl RestoreCoordinator {
...
+            .get_state_leaf_count(snapshot_before_target.version)
+            > 0;
+
+        info!(
+            lhs = lhs,
+            target_version = target_version,
+            "Starting restore DB from version {} to {}, snapshot_before_lhs: {}, snapshot_before_target: {}, tree restore in progress: {}, expected_version: {} \n\
+            Note: we only guarantee resume from previous in-progress restore. If you want to restore a new DB, please either specify a new target db dir or delete previous in-progress DB in the target db dir.
+            ",
+            lhs,
```
msmouse, line 155: hmm, it's mixture of  structural logging ("lhs = lhs") and embedded variables.. shall we use only one style? (In this particular case I thing a big sentence is easier to read)

storage/backup/backup-cli/src/coordinators/restore.rs
```diff
@@ -109,89 +106,204 @@ impl RestoreCoordinator {
...
+            snapshot_before_lhs.version <= snapshot_before_target.version,
+            "snapshot_before_target {} should be larger than or equal to snapshot_before_lhs {}",
+            snapshot_before_target.version,
+            snapshot_before_lhs.version
+        );
+
+        // Two flags for resuming from a previous in-progress restore
+        // Expected version can be used to tell where to resume when applying writesets or replaying txns
+        // Tree restore in progress can be used to tell the status of 2nd snapshot restore
+        let expected_version = self
```
msmouse, line 139: nit: let's call it db_next_version

storage/backup/backup-cli/src/coordinators/restore.rs
```diff
@@ -109,89 +106,204 @@ impl RestoreCoordinator {
...
+        );
+        let transaction_backups = metadata_view
+            .select_transaction_backups(snapshot_before_lhs.version, target_version)?;
+        let epoch_ending_backups = metadata_view.select_epoch_ending_backups(target_version)?;
+        let mut expected_txn_history_so_far = None;
+
+        // Restore the the state kv between lhs and rs
+        // Ensure the expected_version is smaller than target_version in case we want to resume from in-progress restore
+        if snapshot_before_lhs.version < snapshot_before_target.version
+            && expected_version <= snapshot_before_target.version
```
msmouse, line 170: I think as long as there's already ledger history in the db, we don't need to recover the kv-only snapshot, instead, we just  go on recovering the ledger.  The only bad case I can think of is that the db_next_version is very much earlier than start_version, we can just print it out.

Move the humongous log line (your execution plan) after you know exactly what you want to do: what's the first version you are actually gonna save, which exact snapshots you are gonna use, etc.

storage/aptosdb/src/state_store/mod.rs
```diff
@@ -570,6 +574,43 @@ impl StateStore {
...
+    /// Put the write sets on top of current state
+    pub fn put_write_sets(
+        &self,
+        write_sets: Vec<WriteSet>,
+        first_version: Version,
+        batch: &SchemaBatch,
+        sharded_state_kv_batches: &ShardedStateKvSchemaBatch,
+    ) -> Result<()> {
+        let _timer = OTHER_TIMERS_SECONDS
+            .with_label_values(&["update_with_writesets"])
```
grao1991, line 586: nit: Just call it `put_write_sets`?

grao1991 (approved):

storage/aptosdb/src/state_restore/mod.rs
areshand: no, I didn't update the usage when applying the writeset and only update the usage in batch in snapshot restore. Previously, we discussed it requires reading the current usage of the item corresponding to the WS if we want to update the usage while applying WS.

storage/backup/backup-cli/src/backup_types/transaction/restore.rs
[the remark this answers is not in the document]
areshand: If the first version is not None, we don't want to do anything since this is in the 2nd phase restoring after the 2nd snapshot when the initial TXN ACC was already saved.

storage/backup/backup-cli/src/coordinators/restore.rs
```diff
@@ -88,110 +89,223 @@ impl RestoreCoordinator {
...
+        let max_txn_ver = metadata_view
+            .max_transaction_version()?
+            .ok_or_else(|| anyhow!("No transaction backup found."))?;
+        let lhs = self.ledger_history_start_version();
+        let latest_snapshot_version = self
+            .global_opt
+            .run_mode
+            .get_in_progress_state_snapshot()?
+            .unwrap_or_else(|| std::cmp::min(lhs, max_txn_ver));
+        let latest_snapshot = metadata_view.select_state_snapshot(latest_snapshot_version)?;
```
msmouse, line 125: I would do 
``` rust
let kv_snapshot = if db_next_version.is_none() {
match  self.global_opt.run_mode.get_in_progress_state_snapshot() {
	Some(in_progress_kv_restore) => {
	    // ensure the snapshot still exists in the backup
        },
        None => {
            // get the lastest before lhs 
        }
}
} else {
    None
}

```

storage/backup/backup-cli/src/coordinators/restore.rs
```diff
@@ -88,110 +89,223 @@ impl RestoreCoordinator {
...
+
+        let snapshot_before_target = metadata_view
+            .select_state_snapshot(std::cmp::min(self.target_version(), max_txn_ver))?;
```
msmouse, line 128: To see if there's tree at all, call `let latest_tree_version = db.get_state_snapshot_before(Version::max())`, basically any tree node. And then you call `db.get_state_snapshot_before(latest_tree_version)`, if there is any, the snapshot has been done, else (if there's only one tree version) it might need resuming.

storage/backup/backup-cli/src/coordinators/restore.rs
```diff
@@ -88,110 +89,238 @@ impl RestoreCoordinator {
...
+        let max_txn_ver = metadata_view
+            .max_transaction_version()?
+            .ok_or_else(|| anyhow!("No transaction backup found."))?;
+        let lhs = self.ledger_history_start_version();
+
+        let latest_tree_version = self
+            .global_opt
+            .run_mode
+            .get_state_snapshot_before(Version::MAX);
+        let tree_completed = {
```
msmouse, line 125: nuh.. this doesn't mean the tree is complete.. just there is a tree. (see my previous comment -- you need to call the function again to see if there is only a (partial) tree at that version or there's another tree version before it -- in the latter case, we know trees are complete, while in the former case, the tree can be partial, and that's where you need to try to resume)

areshand:
@msmouse please take a look. Tested with testnet and we can break and resume from any point now.

storage/backup/backup-cli/src/coordinators/restore.rs
```diff
@@ -88,110 +89,284 @@ impl RestoreCoordinator {
...
+                    metadata_view.select_state_snapshot(std::cmp::min(lhs, max_txn_ver))?
+                },
+            }
+        } else {
+            None
+        };
+
+        let tree_snapshot = metadata_view
+            .select_state_snapshot(std::cmp::min(self.target_version(), max_txn_ver))?
+            .expect("Cannot find tree snapshot before target version");
```
msmouse, line 162: when tree is partial (exists but not complete), we need to get the exact snapshot to resume, similar to what you did for the kv snapshot

storage/backup/backup-cli/src/coordinators/restore.rs
```diff
@@ -88,110 +89,284 @@ impl RestoreCoordinator {
...
+                        restore_mode: StateSnapshotRestoreMode::KvOnly,
+                    },
+                    self.global_opt.clone(),
+                    Arc::clone(&self.storage),
+                    epoch_history.clone(),
+                )
+                .run()
+                .await?;
+            }
+            // Save the txn between the start version of first txn chunk and the first kv snapshot
```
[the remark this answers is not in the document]
areshand: This range we cannot restore kv since we haven't restore kv snapshot and we cannot get a baseline usage to write the KV

storage/backup/backup-cli/src/coordinators/restore.rs
```diff
@@ -88,110 +89,308 @@ impl RestoreCoordinator {
...
+                .select_state_snapshot(std::cmp::min(self.target_version(), max_txn_ver))?
+                .expect("Cannot find tree snapshot before target version")
+        };
+
+        let two_phase_restore = if let Some(kv_snapshot) = kv_snapshot.as_ref() {
+            // if we have a kv snapshot, we need to restore the state between lhs and rs
+            kv_snapshot.version < tree_snapshot.version
+        } else {
+            // if we don't have a kv snapshot, we need to restore the state between db_next_version and rs
+            db_next_version < tree_snapshot.version && db_next_version > 0
```
msmouse, line 188: why it's not "two phase" when db_next_version == 0 and tree_snapshot.version > 0?

storage/backup/backup-cli/src/coordinators/restore.rs
```diff
@@ -88,110 +89,308 @@ impl RestoreCoordinator {
...
+                latest_tree_version
+            );
+            snapshot.unwrap()
+        } else {
+            metadata_view
+                .select_state_snapshot(std::cmp::min(self.target_version(), max_txn_ver))?
+                .expect("Cannot find tree snapshot before target version")
+        };
+
+        let two_phase_restore = if let Some(kv_snapshot) = kv_snapshot.as_ref() {
```
msmouse, line 183: let's call it `do_phase_1`

msmouse (commented):
let's go through it in person..

storage/backup/backup-cli/src/coordinators/restore.rs
```diff
@@ -88,110 +89,308 @@ impl RestoreCoordinator {
...
+        };
+
+        let mut db_next_version = self
             .global_opt
             .run_mode
             .get_next_expected_transaction_version()?;
-        if next_txn_version != 0 {
-            // DB is already in workable state
+
+        let kv_snapshot = if db_next_version == 0 {
```
areshand, line 149: in progress version is none and db_next_version to decide if kv_snapshot is some

storage/backup/backup-cli/src/coordinators/restore.rs
```diff
@@ -88,110 +89,308 @@ impl RestoreCoordinator {
...
+        } else {
+            db_next_version
+        };
+        let transaction_backups =
+            metadata_view.select_transaction_backups(txn_start_version, target_version)?;
+        let epoch_ending_backups = metadata_view.select_epoch_ending_backups(target_version)?;
+
+        // Restore the the state kv between lhs and rs
+        if two_phase_restore {
+            let start_version = if let Some(ref kv_snapshot) = kv_snapshot {
```
areshand, line 201: start_version is duplicated to txn_start_version

storage/backup/backup-cli/src/coordinators/restore.rs
```diff
@@ -88,110 +89,308 @@ impl RestoreCoordinator {
...
-                next_txn_version = next_txn_version,
-                "DB is ready to accept transactions, start the node to catch up with the chain. \
-                If the node is unable to catch up because the DB is too old, delete the data folder \
-                and bootstrap again.",
+                "Start restoring DB from version {} to tree snapshot version {}",
+                start_version, tree_snapshot.version,
             );
-            return Ok(());
-        }
+            let epoch_handles = epoch_ending_backups
```
areshand, line 210: remove this epoch restore

storage/backup/backup-cli/src/coordinators/restore.rs
```diff
@@ -88,110 +89,308 @@ impl RestoreCoordinator {
...
+                    .await?,
+                ))
+            } else {
+                None
+            };
 
-        let state_snapshot_backup =
-            if let Some(version) = self.global_opt.run_mode.get_in_progress_state_snapshot()? {
+            // phase1.a: save the txn between the start version of first txn chunk and the first kv snapshot
+            let kv_snapshot_version = self
```
areshand, line 230: resue previous kv_snapshot_version

storage/backup/backup-cli/src/coordinators/restore.rs
```diff
@@ -88,110 +89,308 @@ impl RestoreCoordinator {
...
+                .run_mode
+                .get_in_progress_state_kv_snapshot()?
+                .unwrap_or(
+                    if let Some(kv_snapshot) = kv_snapshot.as_ref() {
+                        kv_snapshot.version
+                    } else {
+                        0
+                    },
+                );
+            if db_next_version < kv_snapshot_version {
```
areshand, line 241: check if kv_snapshot is none

storage/backup/backup-cli/src/coordinators/restore.rs
```diff
@@ -88,110 +89,308 @@ impl RestoreCoordinator {
...
+            // reset the global
+            let mut transaction_restore_opt = self.global_opt.clone();
+            // We should replay kv to include the version of tree snapshot so that we can get correct storage usage at that version
+            // while restore tree only snapshots
+            transaction_restore_opt.target_version = tree_snapshot.version;
+            TransactionRestoreBatchController::new(
+                transaction_restore_opt,
+                Arc::clone(&self.storage),
+                txn_manifests,
+                Some(kv_snapshot_version),
```
areshand, line 305: max(db_next_version, kv_snapshot_version.unwrap_or(0))

storage/backup/backup-cli/src/coordinators/restore.rs
```diff
@@ -88,72 +89,124 @@ impl RestoreCoordinator {
...
-        let version = state_snapshot_backup.version;
-        self.global_opt.target_version = version;
-        let epoch_ending_backups = metadata_view.select_epoch_ending_backups(version)?;
-        let transaction_backups = metadata_view
-            .select_transaction_backups(self.ledger_history_start_version(), version)?;
-        COORDINATOR_TARGET_VERSION.set(version as i64);
-        info!(version = version, "Restore target decided.");
+        let kv_snapshot = match self.global_opt.run_mode.get_in_progress_state_kv_snapshot() {
+            Ok(Some(ver)) => {
+                // already restored the kv snapshot, no need to restore again
```
msmouse, line 153: move under `if`

storage/backup/backup-cli/src/coordinators/restore.rs
```diff
@@ -88,72 +89,124 @@ impl RestoreCoordinator {
...
+            );
+            snapshot.unwrap()
+        } else {
+            metadata_view
+                .select_state_snapshot(std::cmp::min(self.target_version(), max_txn_ver))?
+                .expect("Cannot find tree snapshot before target version")
+        };
 
+        let do_phase_1 = if let Some(kv_snapshot) = kv_snapshot.as_ref() {
+            // if we have a kv snapshot, we need to restore the state between lhs and rs
```
msmouse, line 187: document what happends when both snapshot versions are equal

storage/backup/backup-cli/src/coordinators/restore.rs
```diff
@@ -88,72 +89,124 @@ impl RestoreCoordinator {
...
-        self.global_opt.target_version = version;
-        let epoch_ending_backups = metadata_view.select_epoch_ending_backups(version)?;
-        let transaction_backups = metadata_view
-            .select_transaction_backups(self.ledger_history_start_version(), version)?;
-        COORDINATOR_TARGET_VERSION.set(version as i64);
-        info!(version = version, "Restore target decided.");
+        let kv_snapshot = match self.global_opt.run_mode.get_in_progress_state_kv_snapshot() {
+            Ok(Some(ver)) => {
+                // already restored the kv snapshot, no need to restore again
+                if db_next_version > ver {
```
msmouse, line 154: should be `>=`, right? when `db_next_version == ver`, does it mean the snapshot has done?

storage/backup/backup-cli/src/coordinators/restore.rs
```diff
@@ -88,72 +89,124 @@ impl RestoreCoordinator {
...
+                    let snapshot = metadata_view.select_state_snapshot(ver)?;
+                    ensure!(
+                        snapshot.is_some() && snapshot.as_ref().unwrap().version == ver,
+                        "cannot find in-progress state snapshot {}",
+                        ver
+                    );
+                    snapshot
+                }
+            },
+            Ok(None) | Err(_) => metadata_view
```
msmouse, line 166: shall we assert `db_next_ver == 0` here?

storage/backup/backup-cli/src/coordinators/restore.rs
```diff
@@ -164,34 +217,129 @@ impl RestoreCoordinator {
...
+                    self.global_opt.clone(),
+                    Arc::clone(&self.storage),
+                    epoch_history.clone(),
+                )
+                .run()
+                .await?;
+            }
+
+            // phase 1.b: save the txn between the first txn of the first chunk and the tree snapshot
+            let txn_manifests = transaction_backups
```
msmouse, line 248: add an assert that when `db_next_ver == 0 || txn_manifests.first().first_version <= db_next_ver`, just to be defensive.

storage/backup/backup-cli/src/coordinators/restore.rs
```diff
@@ -164,34 +217,129 @@ impl RestoreCoordinator {
...
+            .run()
+            .await?;
+            // update the expected version for the first phase restore
+            db_next_version = tree_snapshot.version;
+        }
+
+        // Phase 2: restore the full tree snapshot and replay till the target version
+        {
+            let first_version = (db_next_version > 0).then_some(db_next_version);
+            // we don't want to replay txn at exact tree snapshot version since the kv is restored either in phase 1 by snapshot restore in default mode
```
msmouse, line 285: it looks like we need a "or" in this sentence 😂

storage/backup/backup-cli/src/backup_types/transaction/restore.rs
```diff
@@ -259,23 +276,33 @@ impl TransactionRestoreBatchController {
...
+                .await?,
+        );
         if let RestoreRunMode::Restore { restore_handler } = self.global_opt.run_mode.as_ref() {
             ensure!(
                 self.output_transaction_analysis.is_none(),
                 "Bug: requested to output transaction output sizing info in restore mode.",
             );
             AptosVM::set_concurrency_level_once(self.global_opt.replay_concurrency_level);
+
+            let (_, kv_only) = self.replay_from_version.unwrap_or((Version::MAX, false));
```
msmouse, line 293: `map_or` is fine

storage/backup/backup-cli/src/backup_types/transaction/restore.rs
```diff
@@ -382,12 +413,10 @@ impl TransactionRestoreBatchController {
...
         // DB doesn't allow replaying anything before what's in DB already.
-        //
-        // TODO: notice that ideally we detect and avoid calling rh.save_transactions() for txns
-        //       before `first_to_replay` calculated below, but we don't deal with it for now,
-        //       because unlike replaying, that's allowed by the DB. Need to follow up later.
+        // self.replay_from_version is from cli argument. However, in fact, we either not replay or replay
+        // after current DB's version.
         let first_to_replay = max(
-            self.replay_from_version.unwrap_or(Version::MAX),
+            self.replay_from_version.unwrap_or((Version::MAX, false)).0,
```
msmouse, line 419: `map_or`

storage/backup/backup-cli/src/backup_types/transaction/restore.rs
```diff
@@ -462,19 +504,92 @@ impl TransactionRestoreBatchController {
         Ok(first_txn_to_replay.map(|_| txns_to_execute_stream))
     }
 
+    // only apply KV to the DB
+    async fn replay_kv(
```
msmouse, line 508: looks like too much duplication... let's talk

msmouse (approved):

PR #7706, opened 2023-04-12
[Node Configs] Small refactors and clean ups.

JoshLind:
### Description
This PR makes several small refactors and cleanups to the node configs. Specifically, it:
- Removes unused parameters from methods.
- Renames a few methods for consistency and clarity.
- Adds additional comments to the code.
- Unifies a few methods to reduce code duplication and provide better consistency.

Nothing should logically change with this PR. 😄  Also, there's a bunch more cleanups to be done, but I'm delegating that to future PRs.

### Test Plan
Existing test infrastructure.

### DIFF
--- a/config/src/config/persistable_config.rs
+++ b/config/src/config/persistable_config.rs
@@ -10,8 +10,9 @@ use std::{
 };
 
 pub trait PersistableConfig: Serialize + DeserializeOwned {
+    /// Load the config from disk at the given path
     fn load_config<P: AsRef<Path>>(path: P) -> Result<Self, Error> {
-        // Open the file and read it into a string
+        // Open the config file
         let config_path_string = path.as_ref().to_str().unwrap().to_string();
         let mut file = File::open(&path).map_err(|error| {
             Error::Unexpected(format!(
@@ -19,8 +20,10 @@ pub trait PersistableConfig: Serialize + DeserializeOwned {
                 config_path_string, error
             ))
         })?;
-        let mut contents = String::new();
-        file.read_to_string(&mut contents).map_err(|error| {
+
+        // Read the file into a string
+        let mut file_contents = String::new();
+        file.read_to_string(&mut file_contents).map_err(|error| {
             Error::Unexpected(format!(
                 "Failed to read the config file into a string: {:?}. Error: {:?}",
                 config_path_string, error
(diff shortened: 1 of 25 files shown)


config/src/config/persistable_config.rs
```diff
@@ -10,39 +10,48 @@ use std::{
 };
 
 pub trait PersistableConfig: Serialize + DeserializeOwned {
-    fn load_config<P: AsRef<Path>>(path: P) -> Result<Self, Error> {
-        // Open the file and read it into a string
+    /// Load the config from disk at the given path
+    fn load_persisted_config<P: AsRef<Path>>(path: P) -> Result<Self, Error> {
```
davidiw, line 14: I don't see the value in this tweak. If we need to better namespace, because we were loading ... other configs, but I can't imagine loading anything other than a "persisted" config. Because we'd generate an ephemeral config. :shrug:

JoshLind: Yeah, so I'm not super worried about the wording/naming, the issue I'm seeing is that we have a bunch of `load` methods related to the different configs and they aren't super obvious to me 🤦 , e.g.:
- `NodeConfig::load_config()`: calls the persistable config (i.e., load from disk).
- `node_config.load()`: is a wrapper, e.g., that calls `load_config` (above) and `execution_config.load()` (below).
- `execution_config.load()`:  This is a wrapper that loads the execution config from disk and does post-processing.
- `network_config.load()`: This is a function that just does post-processing.
- ... etc.

We have the same issue with all the various `save()` methods 😢 

All I'm wanting here is to somehow identify the method that *actually does the read/write to disk*, so it's obvious that everything else is in memory, hence the explicit `load_persisted_config` and `save_persisted_config`. Does this make sense?

JoshLind: If it helps, I'm actually adding a new `NodeConfigLoader` object that should just encapsulate all of the loading, verification and post-processing of the various configs (so we'll have a uniform and consistent way of handling configs going forward). Right now, I find the flows very difficult to reason about. This is actually my next PR (once this lands) 🤞

davidiw: I guess load_config on a type of persistable config feels redundant... I am absolutely nitpicking.

JoshLind: Lol, okay -- I've reverted this change. 😄 Feels like a very odd hill to die on, so I won't debate the rename and pray that this will unblock things.

davidiw: I have died on smaller and more ridiculous hills.

davidiw (approved):

bchocho (approved):

PR #7712, opened 2023-04-12
use shadow-rs GIT_CLEAN in build info

brianolson:
### Description

Add field `build_clean_checkout` to build info as true/false based on shadow-rs GIT_CLEAN which reports if a git checkout is clean (does not have modifications). This should be `true` for release builds and `false` may indicate a development build or other misc hackery.

### Test Plan

Shown to work locally in `aptos info`


msmouse (approved):

JoshLind (approved):
Awesome, thanks @brianolson 😄

ibalajiarun:
Meaning to add this for a while now. Thanks!

PR #7723, opened 2023-04-12
crypto_algebra and bls12381_structures as default features

zjma:
### Description

Turn on the flags for features crypto_algebra and bls12381_structures by default.

### DIFF
--- a/aptos-move/aptos-release-builder/data/example.yaml
+++ b/aptos-move/aptos-release-builder/data/example.yaml
@@ -19,6 +19,8 @@ update_sequence:
         - resource_groups
         - multisig_accounts
         - delegation_pools
+        - cryptography_algebra_natives
+        - bls12381_structures
         - ed25519_pubkey_validate_return_false_wrong_length
       disabled: []
   - Consensus:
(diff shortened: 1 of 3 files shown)


aptos-move/aptos-release-builder/data/example.yaml
```diff
@@ -19,6 +19,8 @@ update_sequence:
         - resource_groups
         - multisig_accounts
         - delegation_pools
+        - cryptography_algebra_natives
```
movekevin, line 22: You don't need to add them in this file. This is just an example

movekevin (approved):

aptos-move/aptos-release-builder/data/release.yaml
```diff
@@ -19,6 +19,8 @@ update_sequence:
         - resource_groups
         - multisig_accounts
         - delegation_pools
+        - cryptography_algebra_natives
```
movekevin, line 22: This file is getting deleted by @runtian-zhou but whoever makes it second will resolve the conflicts :)

alinush (approved):
LGTM!

PR #7724, opened 2023-04-13
Update aptos init in CLI to lookup onchain address based on public key

0xjinn:
### Description
After user rotated their pri/pub key, and they try to use `aptos init` with the updated rotated private key. The account address is incorrect, because it is derived from pub/pri key pair.

This PR updates it by lookup the account address from Public key instead.

Related feature request #6295 

### Test Plan
1. cargo build -p aptos
2. init with rotated key, and make sure account address is correct
`aptos-debug init --private-key 0x53f667f2b0aa149eb848023c919da16482855878ea9a0ac7cb131f7365612814 --profile rotation-test-3`

### DIFF
--- a/crates/aptos/CHANGELOG.md
+++ b/crates/aptos/CHANGELOG.md
@@ -3,10 +3,10 @@
 All notable changes to the Aptos CLI will be captured in this file. This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html) and the format set out by [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
-* Update `aptos init` to fix an incorrect account address issue, when trying to init with a rotated private key. Right now it does an actual account lookup instead of deriving from public key
 
 ### Fixed
 * If `aptos init` is run with a faucet URL specified (which happens by default when using the local, devnet, or testnet network options) and funding the account fails, the account creation is considered a failure and nothing is persisted. Previously it would report success despite the account not being created on chain.
+* Update `aptos init` to fix an incorrect account address issue, when trying to init with a rotated private key. Right now it does an actual account lookup instead of deriving from public key
 
 ## [1.0.9] - 2023/03/29
 * `aptos move show abi` allows for viewing the ABI of a compiled move package
--- a/crates/aptos/CHANGELOG.md
+++ b/crates/aptos/CHANGELOG.md
@@ -3,6 +3,8 @@
 All notable changes to the Aptos CLI will be captured in this file. This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html) and the format set out by [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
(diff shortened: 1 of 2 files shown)


crates/aptos/src/common/init.rs
```diff
@@ -154,7 +156,17 @@ impl CliCommand<()> for InitTool {
             }
         };
         let public_key = private_key.public_key();
-        let address = account_address_from_public_key(&public_key);
+
+        // lookup the address from onchain instead of deriving it
+        // if this is the rotated key, deriving it will outputs an incorrect address
+        let rest_url = Url::parse(&*profile_config.rest_url.clone().unwrap());
```
gregnazario, line 162: So, "technically" the rest URL is optional, and should probably flow as such:
```
if let Some(rest_url) = &profile_config.rest_url {
  ... Lookup address ...
} else {
  address_from_public_key()
}
```

crates/aptos/src/common/init.rs
```diff
@@ -154,7 +158,26 @@ impl CliCommand<()> for InitTool {
             }
         };
         let public_key = private_key.public_key();
-        let address = account_address_from_public_key(&public_key);
+
+        // lookup the address from onchain instead of deriving it
+        // if this is the rotated key, deriving it will outputs an incorrect address
+        // let rest_url = Url::parse(&*profile_config.rest_url.clone().unwrap());
```
gregnazario, line 164: Missed a commented out line

crates/aptos/CHANGELOG.md
```diff
@@ -3,6 +3,8 @@
 All notable changes to the Aptos CLI will be captured in this file. This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html) and the format set out by [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
+* Update `aptos init` to fix an incorrect account address issue, when trying to init with a rotated private key. Right now it does an actual account lookup instead of deriving from public key
```
gregnazario, line 6: Yeah move this under fixed

gregnazario (approved):

crates/aptos/src/common/init.rs
```diff
@@ -154,7 +158,25 @@ impl CliCommand<()> for InitTool {
...
+                profile_options: Default::default(),
+                rest_options: RestOptions::new(
+                    Option::from(Url::parse(rest_url).expect("Failed to parse url")),
+                    None,
+                ),
+            }
+            .execute()
+            .await?
+        } else {
+            account_address_from_public_key(&public_key)
```
davidiw, line 177: @gregnazario , we also have an on-chain repository of rotated account address -> actual account address: https://github.com/aptos-labs/aptos-core/blob/main/aptos-move/framework/aptos-framework/sources/account.move#L62

it might be worth adding this into the cli as well.

0xjinn: Looking at the `OriginatingAddress` struct. This is for looking up the originating address of the rotated auth key?

I guess we can add something like this in the CLI? -> `aptos account lookup-address --authentication-key`

PR #7725, opened 2023-04-13
[token_v2] make entry functions wrapper of inner functions

lightmark:
### Description

It would be useful to have both inner functions which can be used by other modules.

### Test Plan
n/a

### DIFF
--- a/aptos-move/framework/aptos-token-objects/sources/aptos_token.move
+++ b/aptos-move/framework/aptos-token-objects/sources/aptos_token.move
@@ -139,6 +139,19 @@ module aptos_token_objects::aptos_token {
         property_types: vector<String>,
         property_values: vector<vector<u8>>,
     ) acquires AptosCollection, AptosToken {
+        mint_inner(creator, collection, description, name, uri, property_keys, property_types, property_values);
+    }
+
+    public fun mint_inner(
+        creator: &signer,
+        collection: String,
+        description: String,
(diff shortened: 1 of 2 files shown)


aptos-move/framework/aptos-token-objects/sources/aptos_token.move
```diff
@@ -139,6 +139,19 @@ module aptos_token_objects::aptos_token {
         property_types: vector<String>,
         property_values: vector<vector<u8>>,
     ) acquires AptosCollection, AptosToken {
+        mint_inner(creator, collection, description, name, uri, property_keys, property_types, property_values);
+    }
+
+    public fun mint_inner(
```
movekevin, line 145: This completely changes the intention of AptosToken, which is meant to be not extensible. It's basically the entry level token that cannot be extended. If users want custom token, they should build their own instead of on top of AptosToken.

lightmark: Then I would like to know why we disallow that.

davidiw (changes requested):
Agree, this isn't intended to be extensible code. I'd prefer them to copy paste rather than us have to maintain this as extensible code.

lightmark:
@davidiw @movekevin My point is: we don't provide a no code solution for fungible tokens but only fungible assets and tokens.

PR #7744, opened 2023-04-13
Fix grammar / typos in staking pool operations doc

banool:
### Description
While reading the doc I found some weird phrasing.

In this doc we were using `APT coins` and `APT tokens`. I changed both to just `APT`, but feel free @clay-aptos @saharct to change it to whatever is considered standard in the docs site.

### Test Plan
Docs.


clay-aptos (approved):
Thanks, Daniel!

saharct (approved):

PR #7746, opened 2023-04-13
[Forge] Use poetry in GH workflows

sionescu:
### Description

Use poetry to fetch Forge wrapper deps in GH workflows.
Keep the tasks that runs "pip install" to be removed in the future.


geekflyer (approved):

.github/workflows/forge-gcp.yaml
```diff
@@ -60,13 +60,19 @@ jobs:
 
       - uses: actions/setup-python@13ae5bb136fac2878aff31522b9efb785519f984 # pin@v4
 
-      - name: Install python deps
```
perryjrandall, line 63: This is not backwards compatible.... they both need to exist or be cherry picked in

perryjrandall (changes requested):
These changes (removing pip from workflows) are not backwards compatible

This is probably going to break all branches

Just add poetry for now and point all new things to poetry, then we can separately remove pip and make sure its picked into all branches (after giving time for pull requests to rebase)

Also might wanna give folks a heads up that they potentially need to rebase

.github/workflows/forge-gcp.yaml
sionescu: Do you mean cherry-picked in the release branches ?

sionescu:
@perryjrandall done.

perryjrandall:
If those new shell files (find images.sh determinator.sh) dont exist then they're probably going to break all the old branches they're also not backwards compatible

rustielin (approved):

PR #7749, opened 2023-04-13
Delete the devtools directory

banool:
### Description
The only thing left in this directory was this `license_header.txt` file which is not used anywhere as far as I can tell. @JoshLind to confirm.

### Test Plan
CI.


JoshLind:
I think the pre-commit should fail. Unless something has changed, we should still be verifying that all rust files have the header in them 🤔

Issue #7742, opened 2023-04-13
[Bug] Build E2E Dapp tutorial causes stack trace when adding wallet support


drying:
# 🐛 Bug

In Build E2E Dapp tutorial 3 Add Wallet Support, an error has occurred the "index.tsx" file.

**Stack trace/error message**
```
TS2322: Type 'PetraWallet[]' is not assignable to type 'Wallet<string>[]'.
  Type 'PetraWallet' is not assignable to type 'Wallet<string>'.
    Type 'PetraWallet' is not assignable to type 'AdapterPluginProps<string>'.
      Types of property 'signAndSubmitTransaction' are incompatible.
        Type '(transaction: TransactionPayload$1, options?: any) => Promise<{ hash: string; }>' is not assignable to type '<T extends TransactionPayload, V>(transaction: T, options?: V | undefined) => Promise<{ hash: string; }>'.
          Types of parameters 'transaction' and 'transaction' are incompatible.
            Type 'T' is not assignable to type 'TransactionPayload$1'.
              Type 'TransactionPayload' is not assignable to type 'TransactionPayload$1'.
                Type 'TransactionPayload' is not assignable to type 'TransactionPayload$1'.
                  Type 'T' is not assignable to type 'TransactionPayload_MultisigPayload'.
                    Type 'TransactionPayload' is not assignable to type 'TransactionPayload_MultisigPayload'.
                      Type 'TransactionPayload_EntryFunctionPayload' is not assignable to type 'TransactionPayload_MultisigPayload'.
                        Property 'multisig_address' is missing in type 'TransactionPayload_EntryFunctionPayload' but required in type 'MultisigPayload'.
                          Type 'T' is not assignable to type '{ type: string; }'.
                            Type 'TransactionPayload' is not assignable to type '{ type: string; }'.
                              Property 'type' is missing in type 'TransactionPayload' but required in type '{ type: string; }'.
    13 | root.render(
    14 |   <React.StrictMode>
  > 15 |     <AptosWalletAdapterProvider plugins={wallets} autoConnect={true}>
       |                                 ^^^^^^^
    16 |       <App />
    17 |     </AptosWalletAdapterProvider>
    18 |   </React.StrictMode>

```
## System information

Aptos CLI 1.0.9
Aptos TypeScript SDK 1.8.3
macOS Ventura 13.2.1

0xmaayan:
[...]

PR #7750, opened 2023-04-13
[DOC] Update adapter version to use

0xmaayan:


clay-aptos (approved):
Approving per:
https://github.com/aptos-labs/aptos-core/issues/7742#issuecomment-1507246996

Thanks, Maayan!

saharct (approved):

PR #7751, opened 2023-04-13
Fix lint.py shell invocation

sionescu:
### Description

It was passing an obsolete argument (verbose) to LocalShell constructor.

### Test Plan

Tested manually.

### DIFF
--- a/testsuite/lint.py
+++ b/testsuite/lint.py
@@ -15,7 +15,7 @@ def main() -> None:
 @main.command()
 @click.argument("paths", nargs=-1)
 def helm(paths: Tuple[str]) -> None:
-    shell = LocalShell(True)
+    shell = LocalShell()
 
     error = False
     for path in paths:


areshand (approved):

testsuite/lint.py
```diff
@@ -15,7 +15,7 @@ def main() -> None:
 @main.command()
 @click.argument("paths", nargs=-1)
 def helm(paths: Tuple[str]) -> None:
-    shell = LocalShell(True)
+    shell = LocalShell()
```
perryjrandall, line 18: Please add a call to init_logging with debug level to keep the current level of logging

perryjrandall (approved):

testsuite/lint.py
sionescu: I can do that in another commit to unbreak the tests. I'll need to do some refactoring first.

PR #7752, opened 2023-04-13
[fungible_asset] modify the generated name address seed to avoid conf…

lightmark:
…licts if published

### Description
if user publish this package, using the same named address will conflict with each other.

### Test Plan
n/a

### DIFF
--- a/aptos-move/move-examples/fungible_asset/sources/managed_coin.move
+++ b/aptos-move/move-examples/fungible_asset/sources/managed_coin.move
@@ -13,7 +13,7 @@ module fungible_asset_extension::managed_coin {
     /// Only fungible asset metadata owner can make changes.
     const ENOT_OWNER: u64 = 1;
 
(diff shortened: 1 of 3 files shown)


0xmaayan (approved):
Thanks! Can we also update Move.toml file to have fungible_asset_extensions address to be "_" , that way we can publish it easily with a real address

davidiw (commented):
let's use a fake name here

aptos-move/move-examples/fungible_asset/sources/managed_coin.move
```diff
@@ -29,7 +29,7 @@ module fungible_asset_extension::managed_coin {
         primary_store::create_primary_store_enabled_fungible_asset(
             constructor_ref,
             option::some(option::none()),
-            utf8(b"Aptos Token"), /* name */
+            utf8(b"Libra Coin"), /* name */
```
gregnazario, line 32: lol libra

PR #7144, opened 2023-03-14
[Forge] Multi-region simulation test

ibalajiarun:
### Description

This PR adds support simulating varying bandwidth and latency between nodes in a `MultiRegionMultiCloudSimulation` forge test. While the test is general enough to simulate any number of regions, there are some limitations with the underlying dependency in ChaosMesh and/or Linux Traffic Control that prevents from simulating more than 4 regions currently.

As part of the change, this PR changes how bandwidth chaos is represented. It is now more similar to delay chaos in that it allows for simulation multiple bandwidth configurations between any source and target instances.

**Limitations**
Linux Traffic Control only supports a maximum of 16 qdisc (naively one can think of them as network channels) per interface, but three are reserved for general traffic tiering. Therefore, only 13 is available for simulation. Further, one can create a tree of qdisc inheriting parent qdisc that should theoretically allow simulating latency/bandwidth for at least 13 regions. However, it appears that some qdiscs cannot inherit others in certain situations thus leaving the support to only 4 regions for now.

To support more regions, the change also includes using a single configuration with direction `both` to simulate latency between two regions, instead of having two configurations with direction `from`.

### Test Plan

http://o11y.aptosdev.com/grafana/goto/-FgluSB4z?orgId=1

Repeat this for all network chaos resources:
`kubectl describe networkchaos/aws--sa-east-1-to-gcp--us-central1-delay -n <namespace>`
and make sure the status is `Selected` and `AllInjected`. 

### References
Some references for future work:
- https://tldp.org/HOWTO/Adv-Routing-HOWTO/lartc.qdisc.classful.html
- https://unix.stackexchange.com/questions/654559/can-i-use-multiple-qdiscs-per-device
- https://github.com/torvalds/linux/blob/2faac9a98f010cf5b342fa89ac489c4586364e6e/include/uapi/linux/pkt_sched.h#L145

PR #7757, opened 2023-04-13
[forge] Fix three region simulation

ibalajiarun:
### Description

Three region simulation is no longer injecting latency as expected due to changes introduced in #7144. This PR reverts some of those change to make the three region simulation working. I will investigate the cause and follow-up with a fix later.

### Test Plan

Verified that the fix works by running the three region two traffic test with QS turned on.

[Grafana link](https://aptoslabs.grafana.net/d/SnI98V7Vz/quorum-store?from=1681584425629&to=1681585305517&var-Datasource=VictoriaMetrics%20Global%20(Non-mainnet)&var-BigQuery=Google%20BigQuery&var-metrics_source=vmagent&var-chain_name=forge-big-0&var-cluster=All&var-namespace=forge-e2e-pr-7757&var-kubernetes_pod_name=All&orgId=1&viewPanel=154)

### DIFF
--- a/types/src/on_chain_config/consensus_config.rs
+++ b/types/src/on_chain_config/consensus_config.rs
@@ -70,7 +70,7 @@ impl OnChainConsensusConfig {
 
     pub fn quorum_store_enabled(&self) -> bool {
         match &self {
-            OnChainConsensusConfig::V1(_config) => false,
+            OnChainConsensusConfig::V1(_config) => true,
             OnChainConsensusConfig::V2(_config) => true,
         }
     }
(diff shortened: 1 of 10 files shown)


types/src/on_chain_config/consensus_config.rs
```diff
@@ -70,7 +70,7 @@ impl OnChainConsensusConfig {
 
     pub fn quorum_store_enabled(&self) -> bool {
         match &self {
-            OnChainConsensusConfig::V1(_config) => false,
+            OnChainConsensusConfig::V1(_config) => true,
```
ibalajiarun, line 73: Revert before landing

```suggestion
            OnChainConsensusConfig::V1(_config) => false,
```

bchocho (approved):
LGTM

0xmaayan (approved):

perryjrandall (approved):

PR #7758, opened 2023-04-13
add some inspection service tests

brianolson:
### Description

Add some unit tests to inspection service, testing the http handler without spinning up a node.

### Test Plan

This is tests!


crates/inspection-service/src/inspection_service.rs
```diff
@@ -20,13 +20,13 @@ use std::{
...
 
-// TODO: we need to write tests for this service!
-
 // Useful string constants
 const CONTENT_TYPE_JSON: &str = "application/json";
 const CONTENT_TYPE_TEXT: &str = "text/plain";
-const DISABLED_ENDPOINT_MESSAGE: &str =
-    "This endpoint is disabled! Enable it in the InspectionServiceConfig.";
+const CONFIGURATION_DISABLED_MESSAGE: &str =
+    "This endpoint is disabled! Enable it in node.yaml inspection_service.expose_configuration: true";
```
JoshLind, line 27: nit: not all operators call their config files `node.yaml`. I think the convention is `public_full_node.yaml`, `validator.yaml` and `validator_full_node.yaml`.

But, my feeling would be to avoid the issue entirely and leave it unnamed, e.g., `Enable it in your node config via inspection_service.expose_configuration: true`

crates/inspection-service/src/inspection_service.rs
```diff
@@ -231,3 +231,109 @@ pub fn start_inspection_service(node_config: NodeConfig) {
...
+#[cfg(test)]
+mod test {
+    use super::*;
+    use futures::executor::block_on;
+    use hyper::body;
+    use once_cell::sync::Lazy;
+    use prometheus::{register_int_counter, IntCounter};
+    use std::io::read_to_string;
+
+    const INT_COUNTER_NAME: &str = "INT_COUNTER";
```
JoshLind, line 244: nit: Nice! Maybe add a comment saying that this is a test counter only valid for the life of the tests (to ensure the metrics endpoint works).

crates/inspection-service/src/inspection_service.rs
```diff
@@ -231,3 +231,109 @@ pub fn start_inspection_service(node_config: NodeConfig) {
...
+                .body(Body::from(""))
+                .unwrap(),
+            config,
+        ))
+        .unwrap();
+        assert_eq!(r1.status(), StatusCode::OK);
+        let r1body = block_on(body::to_bytes(r1.body_mut())).unwrap();
+        let r1bs = read_to_string(r1body.as_ref()).unwrap();
+        let linecount = r1bs.lines().count();
+        assert_eq!(3, linecount, "Expected one counter to generate 3 lines, but this could change reasonably if metrics formatting changes");
```
JoshLind, line 336: As you point out, this feels a bit fragile. Perhaps just skip the length assertion?

crates/inspection-service/src/inspection_service.rs
```diff
@@ -231,3 +231,109 @@ pub fn start_inspection_service(node_config: NodeConfig) {
...
+                .method(Method::GET)
+                .body(Body::from(""))
+                .unwrap(),
+            config,
+        ))
+        .unwrap();
+        assert_eq!(r2.status(), StatusCode::OK);
+        let r2body = block_on(body::to_bytes(r2.body_mut())).unwrap();
+        let r2bs = read_to_string(r2body.as_ref()).unwrap();
+        assert!(r2bs.contains("NodeConfig"));
```
JoshLind, line 278: nit: I'd declare these strings as constants in the test module (so it's obvious what we're assuming :D)

crates/inspection-service/src/inspection_service.rs
```diff
@@ -231,3 +231,109 @@ pub fn start_inspection_service(node_config: NodeConfig) {
...
+                .method(Method::GET)
+                .body(Body::from(""))
+                .unwrap(),
+            config,
+        ))
+        .unwrap();
+        assert_eq!(r2.status(), StatusCode::OK);
+        let r2body = block_on(body::to_bytes(r2.body_mut())).unwrap();
+        let r2bs = read_to_string(r2body.as_ref()).unwrap();
+        assert!(r2bs.contains("build_commit_hash"));
```
JoshLind, line 313: nit: I'd declare these strings as constants in the test module (so it's obvious what we're assuming :D)

crates/inspection-service/src/inspection_service.rs
```diff
@@ -231,3 +231,109 @@ pub fn start_inspection_service(node_config: NodeConfig) {
...
+        assert!(r2bs.contains("expose_configuration: true"));
+    }
+
+    #[test]
+    fn test_inspect_system() {
+        let mut config = NodeConfig::default_for_validator();
+        config.inspection_service.expose_system_information = false;
+        let r1f = serve_requests(
+            Request::builder()
+                .uri("http://127.0.0.1:9201/system_information")
```
JoshLind, line 289: nit: I'd introduce file constants for the endpoints and just refer to those in the actual service and the tests (to avoid duplication, etc.), e.g.:

```
// Endpoint constants
const CONFIGURATION_ENDPOINT: &str = "/configuration";
const JSON_METRICS_ENDPOINT: &str = "/json_metrics";

...

// Inspection service
async fn serve_requests(...) {
  ...
  CONFIGURATION_ENDPOINT => { ...}
  JSON_METRICS_ENDPOINT => {...}
  ...
}

// Tests
...
fn test_inspect_configuration() {
  ...
  let request_function = serve_requests(
    Request::builder()
      .uri(LOCALHOST + CONFIGURATION_ENDPOINT)
      ...
  ...
}

```

crates/inspection-service/src/inspection_service.rs
```diff
@@ -231,3 +231,109 @@ pub fn start_inspection_service(node_config: NodeConfig) {
...
+    }
+
+    #[test]
+    fn test_inspect_metrics() {
+        INT_COUNTER.inc();
+
+        let config = NodeConfig::default_for_validator();
+        let mut r1 = block_on(serve_requests(
+            Request::builder()
+                .uri("http://127.0.0.1:9201/metrics")
```
JoshLind, line 325: Can we add tests for the `forge_metrics` and `json_metrics` endpoints, too? 😄

JoshLind (commented):
Nice! 😄 Thanks @brianolson. A couple nits and only two big questions: (i) how do we deal with concurrent test runs that try to bind to the same port?; and (ii) can we add tests for the missing endpoints?

Otherwise, looks solid 🚀

crates/inspection-service/src/inspection_service.rs
brianolson: done

crates/inspection-service/src/inspection_service.rs
brianolson: done

crates/inspection-service/src/inspection_service.rs
brianolson: added a comment to explain them

crates/inspection-service/src/inspection_service.rs
brianolson: added comments explaining each

crates/inspection-service/src/inspection_service.rs
brianolson: yeah, dropped line count assertion, just assert the text content

crates/inspection-service/src/inspection_service.rs
brianolson: commented

crates/inspection-service/src/inspection_service.rs
```diff
@@ -20,13 +20,13 @@ use std::{
...
 
-// TODO: we need to write tests for this service!
-
 // Useful string constants
 const CONTENT_TYPE_JSON: &str = "application/json";
 const CONTENT_TYPE_TEXT: &str = "text/plain";
-const DISABLED_ENDPOINT_MESSAGE: &str =
-    "This endpoint is disabled! Enable it in the InspectionServiceConfig.";
+const CONFIGURATION_DISABLED_MESSAGE: &str =
+    "This endpoint is disabled! Enable it in node config at inspection_service.expose_configuration: true";
```
JoshLind, line 27: nit: "in node" -> "in the node"

crates/inspection-service/src/inspection_service.rs
```diff
@@ -89,13 +89,20 @@ pub fn get_all_metrics() -> HashMap<String, String> {
...
+
 async fn serve_requests(
     req: Request<Body>,
     node_config: NodeConfig,
 ) -> Result<Response<Body>, hyper::Error> {
     // Process the request and get the response components
     let (status_code, body, content_type) = match req.uri().path() {
-        "/configuration" => {
+        CONFIGURATION_PATH => {
+            // /configuration
```
JoshLind, line 105: nit: comments seem kinda redundant, but that's probably subjective 😄

crates/inspection-service/src/inspection_service.rs
```diff
@@ -231,3 +242,113 @@ pub fn start_inspection_service(node_config: NodeConfig) {
...
+        ))
+        .unwrap()
+    }
+
+    #[test]
+    fn test_inspect_configuration() {
+        let mut config = NodeConfig::get_default_validator_config();
+
+        config.inspection_service.expose_configuration = false;
+        let mut r1 = do_test_get(&config, CONFIGURATION_PATH);
```
JoshLind, line 280: nit: this might be subjective, but a lot of our code uses underscore naming conventions, and we generally try to be more explicit than less, e.g.:
- `r1` -> `response_1` (or `resp_1`, etc.)
- `r1_body` -> `resp_1_body`
- `r2bs` -> `resp_2_body_str`

This just makes things more uniform 😄

JoshLind (approved):
Thanks @brianolson, looks reasonable to me! 😄 Just some nits, but nothing big. Will add others for the second stamp.

crates/inspection-service/src/inspection_service.rs
```diff
@@ -89,13 +89,20 @@ pub fn get_all_metrics() -> HashMap<String, String> {
     get_metrics(all_metric_families)
 }
 
+const CONFIGURATION_PATH: &str = "/configuration";
+const METRICS_PATH: &str = "/metrics";
+const SYSTEM_INFORMATION_PATH: &str = "/system_information";
+const JSON_METRICS_PATH: &str = "/json_metrics";
```
JoshLind, line 95: nit: alphabetical? 😄

crates/inspection-service/src/inspection_service.rs
brianolson: yeah, just these variables are only test code and only exist for a moment and they don't really matter, ... </whine>, done

crates/inspection-service/src/inspection_service.rs
brianolson: ok, but I had to alphabetize the match cases where these are used to keep them in the same order

ibalajiarun (approved):

PR #7761, opened 2023-04-14
test view function with struct arguments and generic type parameters

gerben-stavenga:
### Description

Add test for a view function with a generic type parameter taking Object<T>


movekevin (approved):
Nice! Cc @xindingw

movekevin (commented):
Why not add an API test instead to make sure things work e2e?

aptos-move/e2e-move-tests/src/tests/constructor_args.rs
```diff
@@ -21,21 +25,26 @@ const OBJECT_ADDRESS: AccountAddress = AccountAddress::new([
     0x90, 0x71, 0xAA, 0x3F, 0xBD, 0x2A, 0xB9, 0x51, 0x37, 0xF7, 0xCB, 0xAD, 0x13, 0x6F, 0x09, 0x2B,
 ]);
 
-fn success(tests: Vec<(&str, Vec<Vec<u8>>, &str)>) {
-    success_generic(vec![], tests)
+fn module_data() -> StructTag {
```
xindingw, line 28: nit: maybe have a static variable with lazy initialization like [here](https://github.com/aptos-labs/aptos-core/blob/6b610f4fc36098662dd0742ecc16687b8b535bb3/aptos-move/e2e-move-tests/src/tests/stake.rs#L15)

xindingw (approved):

PR #7764, opened 2023-04-14
[cli] bugfix Lookup address on init 1.0.11 release

gregnazario:
### Description
Fixes a bug when `aptos init` for an account doesn't exist will fail immediately.

### Test Plan
Before:
```
./target/debug/aptos init --profile test-init2
Aptos already initialized for profile test-init2, do you want to overwrite the existing config? [yes/no] >
^C
 ✘ greg@jiggy  /opt/git/aptos-core   main  ./target/debug/aptos init --profile test-init3
Configuring for profile test-init3

Enter your private key as a hex literal (0x...) [Current: None | No input: Generate new key (or keep one if present)]

No key given, generating key...
{
  "Error": "API error: API error Error(AccountNotFound): Account not found by Address(0x33c88b2eb0bee2dc483525b5b89a0bf5f0bb710f3f3027baeb3b337eda1fe582) and Ledger version(248990)"
}
```

After:
```
$ ./target/debug/aptos init --profile test-init2
Configuring for profile test-init2

Enter your private key as a hex literal (0x...) [Current: None | No input: Generate new key (or keep one if present)]

No key given, generating key...
Account 21f9bffd0240b91627c3948b69ffdf51030965bcf7e58ce3e4768ee730838074 doesn't exist, creating it and funding it with 100000000 Octas
Account 21f9bffd0240b91627c3948b69ffdf51030965bcf7e58ce3e4768ee730838074 funded successfully

---
Aptos CLI is now set up for account 21f9bffd0240b91627c3948b69ffdf51030965bcf7e58ce3e4768ee730838074 as profile test-init2!  Run `aptos --help` for more information about commands
{
  "Result": "Success"
}

```

Lookup also still works where it gives an error if the account isn't on chain.
```
./target/debug/aptos account lookup-address --profile test-init
{
  "Result": "99e7ea90756cbe5760b0ea88ff3bfc2aab4902edca4972e4e6e05cc9af7ad194"
}

./target/debug/aptos account lookup-address --profile testnet --url https://fullnode.devnet.aptoslabs.com 
{
  "Error": "API error: API error Error(AccountNotFound): Account not found by Address(0x21dbfcef4bafda0a410050852d03e1924edb785e4ab98b19429ae1875e97a591) and Ledger version(252976)"
}
```

### DIFF
--- a/crates/aptos/src/account/key_rotation.rs
+++ b/crates/aptos/src/account/key_rotation.rs
@@ -3,10 +3,10 @@
 
 use crate::common::{
     types::{
-        CliCommand, CliConfig, CliError, CliTypedResult, ConfigSearchMode, EncodingOptions,
-        EncodingType, ExtractPublicKey, ParsePrivateKey, ProfileConfig, ProfileOptions,
-        PublicKeyInputOptions, RestOptions, RotationProofChallenge, TransactionOptions,
-        TransactionSummary,
+        account_address_from_public_key, CliCommand, CliConfig, CliError, CliTypedResult,
+        ConfigSearchMode, EncodingOptions, EncodingType, ExtractPublicKey, ParsePrivateKey,
+        ProfileConfig, ProfileOptions, PublicKeyInputOptions, RestOptions, RotationProofChallenge,
+        TransactionOptions, TransactionSummary,
     },
     utils::{prompt_yes, prompt_yes_with_override, read_line},
 };
@@ -20,10 +20,7 @@ use aptos_rest_client::{
     error::{AptosErrorResponse, RestError},
     Client,
 };
-use aptos_types::{
-    account_address::AccountAddress, account_config::CORE_CODE_ADDRESS,
-    transaction::authenticator::AuthenticationKey,
-};
+use aptos_types::{account_address::AccountAddress, account_config::CORE_CODE_ADDRESS};
 use async_trait::async_trait;
 use clap::Parser;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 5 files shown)


crates/aptos/src/account/key_rotation.rs
```diff
@@ -286,41 +283,66 @@ impl CliCommand<AccountAddress> for LookupAddress {
...
-                // It won't be in the table if it wasn't rotated, then return the derived account address
-                rest_client.get_account_bcs(address_key).await?;
-                Ok(address_key)
-            },
-            Err(err) => Err(err)?,
-        }
+            })) = result
+            {
+                if must_exist {
+                    return Err(result.unwrap_err());
```
banool, line 337: Could you just do `return result`?

crates/aptos/src/common/init.rs
```diff
@@ -159,34 +158,21 @@ impl CliCommand<()> for InitTool {
         };
         let public_key = private_key.public_key();
 
+        let client = aptos_rest_client::Client::new(
+            Url::parse(profile_config.rest_url.as_ref().unwrap())
```
banool, line 162: I don't like this `unwrap` on `rest_url`. If we're assuming that this will always be Some then let's just not make it an Option at all.

crates/aptos/src/common/init.rs
```diff
@@ -159,34 +158,21 @@ impl CliCommand<()> for InitTool {
...
+                .map_err(|err| CliError::UnableToParse("rest_url", err.to_string()))?,
+        );
+
         // lookup the address from onchain instead of deriving it
         // if this is the rotated key, deriving it will outputs an incorrect address
```
banool, line 167: This comment feels out of date since it doesn't cover what happens if the address can't be found on chain.

crates/aptos/src/account/key_rotation.rs
```diff
@@ -286,41 +283,66 @@ impl CliCommand<AccountAddress> for LookupAddress {
...
+                }
+            } else if let Err(err) = result {
+                return Err(err);
+            }
+
+            Ok(address_key)
```
banool, line 343: I find this awkward, can we just use a match statement, I feel it is more readable:
```
match rest_client.get_account_bcs(address_key).await {
    Ok(_) => Ok(address_key),
    Err(RestError::Api(AptosErrorResponse { etc })) => etc,
    Err(e) => Err(e),
}
```

Makes the control flow simpler, no explicit early returns, etc.

banool (changes requested):

crates/aptos/src/account/key_rotation.rs
gregnazario: It's a different result of a response, and not an account address, so it can't

crates/aptos/src/account/key_rotation.rs
gregnazario: I did this awkward flow, because as you can tell, it had duplicated code from this one special case / the result isn't the same

crates/aptos/src/account/key_rotation.rs
banool: Perhaps it'd be better to transform it into an anyhow result using `.context()?` instead.

crates/aptos/src/account/key_rotation.rs
banool: If you transform the result type it should be possible right?

banool: Anyway this is a high pri fix so don't worry about it if it's too tricky, we can come back to it later.

crates/aptos/src/common/init.rs
gregnazario: This is an option, because it's a YAML file that it reads from directly (in serialization).

If it is *not* an option, then it will load poorly from the config and give a very cryptic deserialization error.

Earlier in the code, it enforces that the REST URL is set (it's either preset, or creates one).

It's the unfortunate circumstance of this structure being used in multiple places and not being able to assume that the REST URL is set always because someone could manually modify the file, and therefore mess it all up.

banool (approved):
Send

davidiw (approved):

PR #7769, opened 2023-04-14
[framework] Support partial voting in aptos_governance module

xindingw:
### Test Plan

Unit testing.
Move e2e testing.

### DIFF
--- a/aptos-move/framework/aptos-framework/sources/aptos_governance.move
+++ b/aptos-move/framework/aptos-framework/sources/aptos_governance.move
@@ -7,8 +7,8 @@
  * 1. Proposers can create a proposal by calling AptosGovernance::create_proposal. The proposer's backing stake pool
  * needs to have the minimum proposer stake required. Off-chain components can subscribe to CreateProposalEvent to
  * track proposal creation and proposal ids.
- * 2. Voters can vote on a proposal. Their voting power is derived from the backing stake pool. Each stake pool can
- * only be used to vote on each proposal exactly once.
+ * 2. Voters can vote on a proposal. Their voting power is derived from the backing stake pool. A stake pool can vote
+ * on a proposal multiple times as long as the total voting power of these votes don't exceed its total voting power.
  *
  */
 module aptos_framework::aptos_governance {
@@ -17,6 +17,7 @@ module aptos_framework::aptos_governance {
     use std::signer;
     use std::string::{Self, String, utf8};
     use std::vector;
+    use std::features;
 
     use aptos_std::simple_map::{Self, SimpleMap};
     use aptos_std::table::{Self, Table};
@@ -53,9 +54,17 @@ module aptos_framework::aptos_governance {
     const EMETADATA_HASH_TOO_LONG: u64 = 10;
     /// Account is not authorized to call this function.
     const EUNAUTHORIZED: u64 = 11;
+    /// The stake pool is using voting power more than it has.
+    const EVOTING_POWER_OVERFLOW: u64 = 12;
+    /// Partial voting feature hasn't been properly initialized.
+    const EPARTIAL_VOTING_NOT_INITIALIZED: u64 = 13;
+    /// The proposal in the argument is not a partial voting proposal.
+    const ENOT_PARTIAL_VOTING_PROPOSAL: u64 = 14;
 
     /// This matches the same enum const in voting. We have to duplicate it as Move doesn't have support for enums yet.
     const PROPOSAL_STATE_SUCCEEDED: u64 = 1;
+
+    const MAX_U64: u64 = 18446744073709551615;
 
     /// Proposal metadata attribute keys.
     const METADATA_LOCATION_KEY: vector<u8> = b"metadata_location";
@@ -82,6 +91,11 @@ module aptos_framework::aptos_governance {
     /// Records to track the proposals each stake pool has been used to vote on.
     struct VotingRecords has key {
         votes: Table<RecordKey, bool>
+    }
+
+    /// Records to track the voting power usage of each stake pool on each proposal.
+    struct VotingRecordsV2 has key {
+        votes: Table<RecordKey, u64>
     }
 
     /// Used to track which execution script hashes have been approved by governance.
@@ -196,6 +210,18 @@ module aptos_framework::aptos_governance {
         );
     }
 
+    /// Initializes the state for Aptos Governance partial voting. Can only be called through Aptos governance
+    /// proposals with a signer for the aptos_framework (0x1) account.
+    public fun initialize_partial_voting(
+        aptos_framework: &signer,
+    ) {
+        system_addresses::assert_aptos_framework(aptos_framework);
+
+        move_to(aptos_framework, VotingRecordsV2 {
+            votes: table::new(),
+        });
+    }
+
     #[view]
     public fun get_voting_duration_secs(): u64 acquires GovernanceConfig {
         borrow_global<GovernanceConfig>(@aptos_framework).voting_duration_secs
@@ -209,6 +235,49 @@ module aptos_framework::aptos_governance {
     #[view]
     public fun get_required_proposer_stake(): u64 acquires GovernanceConfig {
         borrow_global<GovernanceConfig>(@aptos_framework).required_proposer_stake
+    }
+
+    #[view]
+    /// Return true if a stake pool has already voted on a proposal before partial governance voting is enabled.
+    public fun has_entirely_voted(stake_pool: address, proposal_id: u64): bool acquires VotingRecords {
+        let record_key = RecordKey {
+            stake_pool,
+            proposal_id,
+        };
+        // If a stake pool has already voted on a proposal before partial governance voting is enabled,
+        // there is a record in VotingRecords.
+        let voting_records = borrow_global<VotingRecords>(@aptos_framework);
+        table::contains(&voting_records.votes, record_key)
+    }
+
+    #[view]
+    /// Return remaining voting power of a stake pool on a proposal.
+    /// Note: a stake pool's voting power on a proposal could increase overtime(e.g. rewards/new stake).
+    public fun get_remaining_voting_power(stake_pool: address, proposal_id: u64): u64 acquires VotingRecords, VotingRecordsV2 {
+        assert_voting_initialization();
+
+        // The voter's stake needs to be locked up at least as long as the proposal's expiration.
+        let proposal_expiration = voting::get_proposal_expiration_secs<GovernanceProposal>(@aptos_framework, proposal_id);
+        let lockup_until = stake::get_lockup_secs(stake_pool);
+        if (proposal_expiration > lockup_until) {
+            return 0
+        };
+
+        // If a stake pool has already voted on a proposal before partial governance voting is enabled, the stake pool
+        // cannot vote on the proposal even after partial governance voting is enabled.
+        if (has_entirely_voted(stake_pool, proposal_id)) {
+            return 0
+        };
+        let record_key = RecordKey {
+            stake_pool,
+            proposal_id,
+        };
+        let used_voting_power = 0u64;
+        if (features::partial_governance_voting_enabled()) {
+            let voting_records_v2 = borrow_global<VotingRecordsV2>(@aptos_framework);
+            used_voting_power = *table::borrow_with_default(&voting_records_v2.votes, record_key, &0);
+        };
+        get_voting_power(stake_pool) - used_voting_power
     }
 
     /// Create a single-step proposal with the backing `stake_pool`.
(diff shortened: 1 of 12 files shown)


aptos-move/framework/aptos-framework/sources/aptos_governance.move
```diff
@@ -84,6 +93,16 @@ module aptos_framework::aptos_governance {
         votes: Table<RecordKey, bool>
     }
 
+    /// Records partial voting proposals.
+    struct PartialVotingProposals has key {
```
movekevin, line 97: This looks only useful during the transitioning period but useless after once all proposals are "partial voting proposals". Is there a way to avoid needing this new struct?

aptos-move/framework/aptos-framework/sources/aptos_governance.move
```diff
@@ -211,6 +245,27 @@ module aptos_framework::aptos_governance {
...
+        if (!exists<PartialVotingProposals>(@aptos_framework)) {
+            return false
+        };
+        let partial_voting_proposals = borrow_global<PartialVotingProposals>(@aptos_framework);
+        table::contains(&partial_voting_proposals.proposals, proposal_id)
+    }
+
+    #[view]
+    /// Return voting power usage of a stake pool on a proposal. Only partial voting proposals could be queried.
+    public fun get_voting_power_usage(stake_pool: address, proposal_id: u64): u64 acquires PartialVotingProposals, VotingRecordsV2 {
```
movekevin, line 259: nit: voting power usage is not a good name for this. Also, should we just have a single view function that returns a tuple of vote (yes or no) and voting power associated with the vote?

aptos-move/framework/aptos-framework/sources/aptos_governance.move
```diff
@@ -294,28 +358,76 @@ module aptos_framework::aptos_governance {
...
+        if (!is_partial_voting_proposal) {
+            assert!(
+                !table::contains(&voting_records.votes, record_key),
+                error::invalid_argument(EALREADY_VOTED));
+            table::add(&mut voting_records.votes, record_key, true);
+        };
 
-        let voting_power = get_voting_power(stake_pool);
+        let staking_pool_voting_power = get_voting_power(stake_pool);
+        if (is_partial_voting_proposal) {
```
movekevin, line 418: This logic can be greatly simplified by always checking voting records (v1) first, then if not present, check voting records v2. Also we can update vote() to dual write to both voting records v1 and v2 to simplify removal in the future (then we can just simply delete voting records v1 write once there are no longer proposals that are on the previous records)

aptos-move/framework/aptos-framework/sources/aptos_governance.move
```diff
@@ -53,10 +54,18 @@ module aptos_framework::aptos_governance {
...
+    const EVOTING_POWER_OVERFLOW: u64 = 12;
+    /// Partial voting feature hasn't been properly initialized.
+    const EPARTIAL_VOTING_NOT_INITIALIZED: u64 = 13;
+    /// The proposal in the argument is not a partial voting proposal.
+    const ENOT_PARTIAL_VOTING_PROPOSAL: u64 = 14;
 
     /// This matches the same enum const in voting. We have to duplicate it as Move doesn't have support for enums yet.
     const PROPOSAL_STATE_SUCCEEDED: u64 = 1;
 
+    const MAX_U64: u64 = 18446744073709551615;
```
movekevin, line 67: I wonder if we can just move this into the math_u64 module and add a max_value function or something. @gerben-stavenga what do you think?

aptos-move/framework/aptos-framework/sources/aptos_governance.move
```diff
@@ -211,6 +237,28 @@ module aptos_framework::aptos_governance {
         borrow_global<GovernanceConfig>(@aptos_framework).required_proposer_stake
     }
 
+    #[view]
+    /// Return remaining voting power of a stake pool on a proposal.
+    public fun get_remaining_voting_power(stake_pool: address, proposal_id: u64): u64 acquires VotingRecords, VotingRecordsV2 {
+        if (features::partial_governance_voting_enabled()) {
```
movekevin, line 243: I think it's cleaner to create a function is_partial_voting_enabled() which checks both the feature flag and that VotingRecordsV2 is initialized

aptos-move/framework/aptos-framework/sources/aptos_governance.move
```diff
@@ -211,6 +237,28 @@ module aptos_framework::aptos_governance {
...
+            assert!(exists<VotingRecordsV2>(@aptos_framework), error::invalid_state(EPARTIAL_VOTING_NOT_INITIALIZED));
+        };
+        let record_key = RecordKey {
+            proposal_id,
+            stake_pool,
+        };
+        let voting_records = borrow_global<VotingRecords>(@aptos_framework);
```
movekevin, line 250: Let's just add a getter function for this as well has_entirely_voted or something like that that returns true if a stake pool has voted as tracked by VotingRecords (v1)

movekevin, line 250: Can you add some comments for the flow here just for posterity?

aptos-move/framework/aptos-framework/sources/aptos_governance.move
```diff
@@ -294,28 +342,43 @@ module aptos_framework::aptos_governance {
...
+    ) acquires ApprovedExecutionHashes, GovernanceEvents, VotingRecords, VotingRecordsV2 {
         let voter_address = signer::address_of(voter);
         assert!(stake::get_delegated_voter(stake_pool) == voter_address, error::invalid_argument(ENOT_DELEGATED_VOTER));
 
-        // Ensure the voter doesn't double vote with the same stake pool.
-        let voting_records = borrow_global_mut<VotingRecords>(@aptos_framework);
-        let record_key = RecordKey {
-            stake_pool,
-            proposal_id,
+        let staking_pool_voting_power = get_remaining_voting_power(stake_pool, proposal_id);
```
movekevin, line 378: This is a bit strange to me 🤔 Why not do the math forward - if current voted power + new vote > total stake pool's voting power?

aptos-move/framework/aptos-framework/sources/aptos_governance.move
```diff
@@ -294,28 +342,43 @@ module aptos_framework::aptos_governance {
...
+        vote_internal(voter, stake_pool, proposal_id, voting_power, should_pass);
+    }
+
+    /// Vote on proposal with `proposal_id` and specified voting_power from `stake_pool`.
+    /// If voting_power is more than all the left voting power of `stake_pool`, use all the left voting power.
+    fun vote_internal(
+        voter: &signer,
+        stake_pool: address,
+        proposal_id: u64,
+        voting_power: u64,
```
movekevin, line 372: What happens when voting_power is MAX_U64 here?

aptos-move/framework/aptos-framework/sources/aptos_governance.move
```diff
@@ -334,6 +397,22 @@ module aptos_framework::aptos_governance {
...
 
+        let record_key = RecordKey {
+            stake_pool,
+            proposal_id,
+        };
+        if (features::partial_governance_voting_enabled()) {
+            let voting_records_v2 = borrow_global_mut<VotingRecordsV2>(@aptos_framework);
+            let used_voting_power = table::borrow_mut_with_default(&mut voting_records_v2.votes, record_key, 0);
+            *used_voting_power = *used_voting_power + voting_power;
+        } else {
```
movekevin, line 408: This logic is a bit hard to follow in the case of a proposal created before partial voting is enabled and with voting extending beyond. IMO, it's likely cleaner to fail early if VotingRecords is already updated and:
1. Update VotingRecords if voting power is MAX_U64
2. Update VotingRecordsV2 if partial voting is enabled, regardless of if voting power is MAX_U64 or not.

movekevin (commented):
I think there might be some quirks in the user flow here we need to think through: partial_vote is incremental. How would a voter updates their total vote? For example:
1. I voted with 100 APT for proposal 10, which is the total stake I have
2. I added more stake and there's still time left to vote. My stake pool now have 20 APT
3. I want to update the vote to be 20 APT, but now I have to do the math and know to do partial_vote(10)

Furthermore, after partial voting is enabled, if someone still uses the old vote() function, their partial vote records will be updated to MAX_U64, which if I understand the code correctly, would prevent any further voting.

aptos-move/framework/aptos-framework/sources/aptos_governance.move
xindingw: Sounds good. Since this refactor could affect many places, I will do it in a later PR. Created an issue: https://github.com/aptos-labs/aptos-core/issues/7894

aptos-move/framework/aptos-framework/sources/aptos_governance.move
xindingw: `voting_power` is bounded by `get_remaining_voting_power` in line 378.

xindingw:
> I think there might be some quirks in the user flow here we need to think through: partial_vote is incremental. How would a voter updates their total vote? For example:

> 1. I voted with 100 APT for proposal 10, which is the total stake I have
> 2. I added more stake and there's still time left to vote. My stake pool now have 20 APT
> 3. I want to update the vote to be 20 APT, but now I have to do the math and know to do partial_vote(10)

What does 10 in `partial_vote(10)` mean here? The user can call `get_remaining_voting_power` to know how much voting power is left and the function will return 20 at 3.

> Furthermore, after partial voting is enabled, if someone still uses the old vote() function, their partial vote records will be updated to MAX_U64, which if I understand the code correctly, would prevent any further voting.

If someone calls the old vote(), only all the current voting power is used.

movekevin:
> > I think there might be some quirks in the user flow here we need to think through: partial_vote is incremental. How would a voter updates their total vote? For example:
> 
> > 1. I voted with 100 APT for proposal 10, which is the total stake I have
> > 2. I added more stake and there's still time left to vote. My stake pool now have 20 APT
> > 3. I want to update the vote to be 20 APT, but now I have to do the math and know to do partial_vote(10)
> 
> What does 10 in `partial_vote(10)` mean here? The user can call `get_remaining_voting_power` to know how much voting power is left and the function will return 20 at 3.
> 
> > Furthermore, after partial voting is enabled, if someone still uses the old vote() function, their partial vote records will be updated to MAX_U64, which if I understand the code correctly, would prevent any further voting.
> 
> If someone calls the old vote(), only all the current voting power is used.

1. The user needing to call get_remaining_vote and also another view function likely to see how much they have voted seems inconvenient. We could introduce a function that allows voting with absolute voting power but I guess that'd introduce complexity if we want to maintain the invariant that voters cannot change their existing vote (e.g. from yes to no).
2. Using all current voting power is fine but not being to add more vote further might not be (since VotingRecords v1 is modified). Not being able to add votes is a very inconsistent/confusing behavior after we enable partial voting.

Ideally, after the update, we should have the following 3 separate voting features:
1. partial_vote for incremental votes (10 more yes votes). This will mainly be used by delegation, liquid staking services, and other pool-based staking protocols.
2. vote() which votes with all current voting power. This should not block adding more votes later
3. vote_total(total_voting_power) to allow voting with an absolute number. This needs more thinking though

movekevin (commented):
Are you planning to update the delegation contract in a separate PR to support voting?

xindingw:
> Are you planning to update the delegation contract in a separate PR to support voting?

Yes, splitting into 2 PRs is cleaner and easier for review. That PR is half way done. Letting a delegator delegate its voting power is a bit tricky. Need some product inputs then it should be quick.

xindingw:
> > > I think there might be some quirks in the user flow here we need to think through: partial_vote is incremental. How would a voter updates their total vote? For example:
> > 
> > 
> > > 1. I voted with 100 APT for proposal 10, which is the total stake I have
> > > 2. I added more stake and there's still time left to vote. My stake pool now have 20 APT
> > > 3. I want to update the vote to be 20 APT, but now I have to do the math and know to do partial_vote(10)
> > 
> > 
> > What does 10 in `partial_vote(10)` mean here? The user can call `get_remaining_voting_power` to know how much voting power is left and the function will return 20 at 3.
> > > Furthermore, after partial voting is enabled, if someone still uses the old vote() function, their partial vote records will be updated to MAX_U64, which if I understand the code correctly, would prevent any further voting.
> > 
> > 
> > If someone calls the old vote(), only all the current voting power is used.
> 
> 1. The user needing to call get_remaining_vote and also another view function likely to see how much they have voted seems inconvenient. We could introduce a function that allows voting with absolute voting power but I guess that'd introduce complexity if we want to maintain the invariant that voters cannot change their existing vote (e.g. from yes to no).
> 2. Using all current voting power is fine but not being to add more vote further might not be (since VotingRecords v1 is modified). Not being able to add votes is a very inconsistent/confusing behavior after we enable partial voting.
> 
> Ideally, after the update, we should have the following 3 separate voting features:
> 
> 1. partial_vote for incremental votes (10 more yes votes). This will mainly be used by delegation, liquid staking services, and other pool-based staking protocols.
> 2. vote() which votes with all current voting power. This should not block adding more votes later

The current `vote` doesn't block adding more votes later. `vote` will only mark all the current voting power(instead of MAX_UINT64) is used in VotingRecordsV2. The only case blocking adding more votes is that a stake pool has already voted before we enable partial voting.

> 3. vote_total(total_voting_power) to allow voting with an absolute number. This needs more thinking though
I think we just need to provide a view function for user to get the amount of used voting power? I didn't add it because `VotingRecords` doesn't track it. We can add it later if necessary.

PR #7781, opened 2023-04-14
[Configs] Fix and update tokio-console features.

JoshLind:
Note: this PR is based on: https://github.com/aptos-labs/aptos-core/pull/7706

### Description
This PR makes a few small fixes/tweaks to our configs and features for tokio-console. Specifically:
- Rename the `aptos-console` feature to `tokio-console` (this seems more descriptive, as all it's doing is enabling tokio console).
- Fix a build error with quorum store (i.e., trying to build with the console feature today results in an error). For example:
```
310 |             let name = format!("batch_coordinator-{}", i).as_str();
    |                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^         - temporary value is freed at the end of this statement
    |                        |
    |                        creates a temporary which is freed while still in use
```
- Rename `console_port` to `tokio_console_port` in the configs (again, this is more descriptive).
- Update the configs such that the `tokio_console_port` is None by default.

### Test Plan
Existing test infrastructure and running `cargo build -p aptos-node --features=tokio-console` now actually builds.

### DIFF
--- a/config/src/config/logger_config.rs
+++ b/config/src/config/logger_config.rs
@@ -17,11 +17,10 @@ pub struct LoggerConfig {
     pub is_async: bool,
     // The default logging level for slog.
     pub level: Level,
-    // tokio-console port
-    pub console_port: Option<u16>,
     pub enable_telemetry_remote_log: bool,
     pub enable_telemetry_flush: bool,
     pub telemetry_level: Level,
+    pub tokio_console_port: Option<u16>,
 }
 
 impl Default for LoggerConfig {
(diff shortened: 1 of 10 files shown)


davidiw (approved):

config/src/config/logger_config.rs
```diff
@@ -31,25 +30,24 @@ impl Default for LoggerConfig {
...
-            // feature is enabled
-            console_port: Some(6669),
             enable_telemetry_remote_log: true,
             enable_telemetry_flush: true,
             telemetry_level: Level::Error,
+
+            // This is the default port used by tokio-console.
+            // Setting this to None will disable tokio-console
+            // even if the "tokio-console" feature is enabled.
+            tokio_console_port: None,
```
ibalajiarun, line 40: Hmm, where do we set this to `Some(..)` now? There is `randomize_ports()` fn below but it doesn't appear to be called.

JoshLind: This should only be set to `Some(...)` if we actually want to enable it, right? Are there any tests that actually use it? If so, that's (great) news to me 😄 My reasoning for `None` was that configs should default to production, and if folks want to test/experiment, they should override them 🤔

ibalajiarun: True, that is reasonable; we want to set a port only when we enable it. But, I was thinking we should set a default port when the feature is actually enabled, so no code is change is necessary to enable tokio-console. Otherwise, it might be too much friction to figure this out for someone.

JoshLind: > But, I was thinking we should set a default port when the feature is actually enabled, so no code is change is necessary to enable tokio-console.

Aah, yeah, that makes sense. FWIW, this is something that should hopefully be enabled soon (i.e., I'm currently working on a way to add finer-grained defaults, such that if a compiler feature is enabled (or a node is a certain type), we can specify these kinds of defaults. It'll hopefully be built on top of this: https://github.com/aptos-labs/aptos-core/pull/7783.

ibalajiarun (approved):
LGTM

PR #7783, opened 2023-04-14
[Configs] Add a new ConfigSanitizer component.

JoshLind:
Note: this PR is based on: https://github.com/aptos-labs/aptos-core/pull/7781

### Description
This PR adds a simple `ConfigSanitizer` component to the node configs. The purpose of the `ConfigSanitizer` is to validate, sanity check and sanitize all configs on node startup, to help ensure configs are well-formed and consistent (e.g., that invariants hold across node types and networks).

To do this, we: (i) load the config from disk; (ii) extract the node role from the config; (iii) extract the chain ID from the genesis blob; and (iv) validate the config according to the node role, chain ID and compilation flags. This allows us to enforce various properties about the configs. Note: we're already doing some of these verifications today, but: (i) we're missing a lot of basic checks to prevent users from shooting themselves in the foot; and (ii) whatever checks we do have are spread throughout the code, making it difficult to reason about. The `ConfigSanitizer` will at least provide us a single source of truth regarding config quality.

The PR offers two commits:
1. Add the new sanitizer trait and implementations.
2. Add unit tests to verify sanitizer correctness.

Note:
- I haven't added all the checks we may want to add (only some obvious ones), but we can extend the checks once this lands.
- My next step is to add support for fine-grained node config defaults.

### Test Plan

### DIFF
--- a/config/src/config/config_sanitizer.rs
+++ b/config/src/config/config_sanitizer.rs
@@ -1,10 +1,13 @@
 // Copyright © Aptos Foundation
 // SPDX-License-Identifier: Apache-2.0
 
-use crate::config::{
-    ApiConfig, BaseConfig, ConsensusConfig, Error, ExecutionConfig, IndexerConfig,
-    IndexerGrpcConfig, InspectionServiceConfig, LoggerConfig, MempoolConfig, NodeConfig,
-    PeerMonitoringServiceConfig, RoleType, StateSyncConfig, StorageConfig,
+use crate::{
+    config::{
+        ApiConfig, BaseConfig, ConsensusConfig, Error, ExecutionConfig, IndexerConfig,
+        IndexerGrpcConfig, InspectionServiceConfig, LoggerConfig, MempoolConfig, NodeConfig,
+        PeerMonitoringServiceConfig, RoleType, StateSyncConfig, StorageConfig,
+    },
+    network_id::NetworkId,
 };
 use aptos_types::chain_id::ChainId;
 use cfg_if::cfg_if;
@@ -190,21 +193,25 @@ fn sanitize_validator_network_config(
     if let Some(validator_network_config) = validator_network {
         let network_id = validator_network_config.network_id;
         if !network_id.is_validator_network() {
+            // TODO: improve the defaults!
+            // We must override the network ID to be a validator
+            // network ID as the config defaults to a public network ID.
+            validator_network_config.network_id = NetworkId::Validator;
+        }
+
+        // Verify that the node is a validator
+        if !node_role.is_validator() {
             return Err(Error::ConfigSanitizerFailed(
                 sanitizer_name,
-                format!(
-                    "Validator network config must have a validator network ID! Found ID: {}",
-                    network_id
-                ),
+                "The validator network config cannot be set for non-validators!".into(),
             ));
         }
 
-        // Verify that mutual authentication is enabled
+        // Ensure that mutual authentication is enabled
         if !validator_network_config.mutual_authentication {
-            return Err(Error::ConfigSanitizerFailed(
-                sanitizer_name,
-                "Mutual authentication must be enabled for the validator network!".into(),
-            ));
+            // TODO: improve the defaults!
+            // We must enable mutual authentication for validators
+            validator_network_config.mutual_authentication = true;
         }
 
         // Prepare the network id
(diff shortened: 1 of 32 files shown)


config/src/config/config_sanitizer.rs
```diff
@@ -0,0 +1,376 @@
...
+    }
+}
+
+impl ConfigSanitizer for NodeConfig {
+    fn sanitize(
+        node_config: &mut NodeConfig,
+        node_role: RoleType,
+        chain_id: ChainId,
+    ) -> Result<(), Error> {
+        // Sanitize the API config
```
davidiw, line 43: let's not add redundant comments loc is not measured :P

davidiw (approved):

config/src/config/config_sanitizer.rs
JoshLind: Lol 😆 The comments were initially there to remind me validate all sub-configs. But, I've removed them now 😄 

> loc is not measured :P
(My secret plan has been thwarted :trollface:)

config/src/config/config_sanitizer.rs
```diff
@@ -0,0 +1,326 @@
...
+    if let Some(failpoints) = failpoints {
+        if failpoints_enabled && failpoints.is_empty() {
+            return Err(Error::ConfigSanitizerFailed(
+                sanitizer_name,
+                "Failpoints are enabled, but the failpoints config is empty?".into(),
+            ));
+        } else if !failpoints_enabled && !failpoints.is_empty() {
+            return Err(Error::ConfigSanitizerFailed(
+                sanitizer_name,
+                "Failpoints are disabled, but the failpoints config is not empty!".into(),
```
brianolson, line 114: Is this too picky? Would someone want to turn it off (temporarily) but leave their config in place?

config/src/config/config_sanitizer.rs
```diff
@@ -0,0 +1,326 @@
...
+    if validator_network.is_none() && node_role.is_validator() {
+        return Err(Error::ConfigSanitizerFailed(
+            sanitizer_name,
+            "Validator network config cannot be empty for validators!".into(),
+        ));
+    }
+
+    // Check the validator network config
+    if let Some(validator_network_config) = validator_network {
+        let network_id = validator_network_config.network_id;
```
brianolson, line 194: check that `node_role.is_validator()` ?

config/src/config/indexer_grpc_config.rs
```diff
@@ -23,3 +32,36 @@ pub struct IndexerGrpcConfig {
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub output_batch_size: Option<u16>,
 }
+
```
brianolson, line 35: why not `impl Default for Indexer GrpcConfig {...}` ? It could set enabled=false and the other values. Then it's more clear what someone gets if they edit the .yaml and set enabled=true

config/src/config/logger_config.rs
```diff
@@ -51,3 +56,66 @@ impl LoggerConfig {
...
+        if is_tokio_console_enabled() && logger_config.tokio_console_port.is_none() {
+            return Err(Error::ConfigSanitizerFailed(
+                sanitizer_name,
+                "The tokio-console feature is enabled but the tokio console port is not set!"
+                    .into(),
+            ));
+        } else if !is_tokio_console_enabled() && logger_config.tokio_console_port.is_some() {
+            return Err(Error::ConfigSanitizerFailed(
+                sanitizer_name,
+                "The tokio-console feature is not enabled but the tokio console port is set!"
```
brianolson, line 80: another maybe-too picky spot? It's not bad to have it disabled but configured in case of turning it back on temporarily for diagnostics?

types/src/chain_id.rs
```diff
@@ -76,6 +76,37 @@ impl FromStr for NamedChain {
 #[derive(Clone, Copy, Deserialize, Eq, Hash, PartialEq, Serialize)]
 pub struct ChainId(u8);
 
+impl ChainId {
+    /// Returns true iff the chain ID matches devnet
+    pub fn is_devnet(&self) -> Result<bool> {
```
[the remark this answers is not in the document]
JoshLind: Personally, I'm not a super big fan of hiding errors. I'd rather the caller know (and explicitly have to deal with it), than just return `false` and allow the code to continue. But, maybe I'm a little extreme 😄

config/src/config/indexer_grpc_config.rs
JoshLind: Yeah, so this is something you can argue with @bowenyang007 and @larry-aptos about 😄 I just moved things around. I agree -- I much prefer defaulting, too.

config/src/config/config_sanitizer.rs
JoshLind: Good question. I think `failpoints` are only enabled in specific types of test infra (e.g., forge), so I can't really imagine anyone wanting to enable the feature locally and not set the failpoints. I'm inclined to be strict, and then loosen it later (if we find a reasonable use-case)?

config/src/config/config_sanitizer.rs
JoshLind: Aah, yeah, makes sense!

config/src/config/logger_config.rs
JoshLind: Potentially. @ibalajiarun actually mentioned something similar too (e.g., see here for my reasoning: https://github.com/aptos-labs/aptos-core/pull/7781). I think the best approach is to probably change the default based on the `tokio-console` feature (i.e., it's it's enabled we set the config to `Some(...)`, otherwise, we set it to `None`). And this kind of fine-grained tuning is coming soon (my next PR?), so hopefully we can get the best of both worlds once this lands.

bchocho (approved):
LGTM!

JoshLind:
Landing this to unblock the quorum store checks.

PR #7788, opened 2023-04-15
[Quorum Store] Include author to transactions in digest

bchocho:
### Description

The author in the payload will prevent a malicious validator from making the same payload as another validator, which has some theoretical attacks given our implementation.

### DIFF
--- a/consensus/src/quorum_store/types.rs
+++ b/consensus/src/quorum_store/types.rs
@@ -156,6 +156,10 @@ impl Batch {
 
     pub fn verify(&self) -> anyhow::Result<()> {
         ensure!(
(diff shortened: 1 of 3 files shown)


consensus/src/quorum_store/types.rs
sasha8, line 1: Having the author be part of the batch payload feels like a hack.
But if @zekun000 will approve I will as well :)

zekun000 (approved):

sasha8 (approved):
This is ugly, but I stamp as I promised :)

sherry-x:
why is this needed? can we have some brief description :)

PR #7789, opened 2023-04-15
[CLI] Fix empty vector, and cleanup vec<u8> parsing

gregnazario:


lightmark:
could we use recursive `[]` to represent vectors? I don't think we should ensure backwards compatibility for CLI if we witness a significant shortcoming of the current input format.

alnoki:
> could we use recursive `[]` to represent vectors? I don't think we should ensure backwards compatibility for CLI if we witness a significant shortcoming of the current input format.

@lightmark @gregnazario #7790 adds recursive `[]`.

Issue #7709, opened 2023-04-12
[Feature Request][CLI] JSON input file support


alnoki:
@movekevin @gregnazario @davidiw

# Feature request summary

Accept JSON input files for CLI entry functions. This will provide:

1. Easier scripted operations for arbitrary on-chain operations.
2. A historical local reference for debugging purposes.
3. A truth reference for multisig owners who want to agree on the same transaction.
4. Argument support currently lacking from the CLI.

# Simple example

A simple coin transfer:

```json
{
    "function": "0x1::coin::transfer",
    "type_arguments": [
        "0x1::aptos_coin::AptosCoin"
    ],
    "arguments": [
        {
            "argument_type": "address",
            "argument_value": "0xbee"
        },
        {
            "argument_type": "u64",
            "argument_value": 100
        }
    ]
}
```

# Nested vectors

Presently, the CLI does not accept nested vector types. This prevents the CLI from facilitating, for example, governance operations that require multi-dimensional parameters sets ([example](https://github.com/econia-labs/econia/blob/337f6824157c9988e2310613b6fc63643801f51e/src/move/econia/sources/incentives.move#L979) requiring `vector<vector<u64>>`).

In a JSON transaction spec, a `vector<vector<u64>>` could be represented via:

```json
{
    "argument_type": "u64",
    "argument_value": [
        [100, 200],
        [300, 400]
    ]
}
```

Here, `argument_type` specifies the inner type, and the [`serde::Value`](https://docs.rs/serde_json/latest/serde_json/value/enum.Value.html) libraries can be used to determing the level of nesting based on the use of `[` and `]` in JSON.

# Byte vectors

As a special case, it is suggested that `vector<u8>` arguments accept JSON syntax in not only the list-style format above, but also in hexstring format. Hence the two would be parsed equivalently:

```json
{
    "argument_type": "u8",
    "argument_value": [255, 0]
}
```

```json
{
    "argument_type": "u8",
    "argument_value": "0xff00"
}
```

This design enables easier readability for byte vectors.

# Multisig operations

The multisig v2 CLI tooling currently under development under #7672 enables proposal/approval/rejection mechanisms, but does not provide straightforward means for owners to verify the transaction they are signing off on.

Hence, it is suggested that the CLI incorporate functionality for generating a multisig transaction metadata file as below:

```json
{
    "multisig_address": "0xace",
    "entry_function": {
        "function": "0x1::coin::transfer",
        "type_arguments": [
[...]

PR #7790, opened 2023-04-15
[CLI] Add nested vector arg support, docs (#7709)

alnoki:
@davidiw @gregnazario @movekevin

Addresses #7709 

# Summary of changes

* Implement recursive vector parsing for `ArgWithType`
* Abstract out `ArgWithTypeVec` field in accordance with [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) design principles
* Add Move example with multiple nested vector permutations
* Add CLI docs for running example Move code

# Vector syntax

Previously, the CLI was restricted to vector arguments with a maximum depth of 1, per the below syntax:

`vector<u8>:2,5,3`

Notably, this schema does *not* support doubly nested vectors of the form `vector<vector<u8>>`, as there is no way to delimit inner vectors.

However, since Move vectors must all use the same inner type, the new schema employs JSON array syntax of the form:

`"u8:[[1, 2], [], [4, 5, 6]]"`

Note here the use of quotes to escape spaces and square braces, as required for typical shell interpreters like `zsh`.

Hence as described in the docs example from this PR, a Move function can be invoked via syntax of the form:

```zsh
aptos move run \
    --function-id <deploy_address>::cli_args::set_vals \
    --private-key-file <keyfile> \
    --args \
        u8:123 \
        "bool:[false, true, false, false]" \
        'address:[["0xace", "0xbee"], ["0xcad"], ["0xdee"], []]'
```

# Additional example

See also testnet transaction [499639189](https://explorer.aptoslabs.com/txn/499639189/payload?network=testnet) for a public entry function invocation with a `vector<vector<u64>>` argument, run from the CLI via:

```zsh
aptos move run \
    --function-id $testnet_address::incentives::update_incentives \
    --private-key-file tesnet_account.key \
    --type-args 0x1::aptos_coin::AptosCoin \
    --args \
        u64:1 \
        u64:1 \
        u64:1 \
        u64:2000 \
        "u64:[[10000, 0, 7], [8333, 1, 6], [7692, 2, 5], [7143, 3, 4], [6667, 4, 3], [6250, 5, 2], [5882, 6, 1]]"
```

# Local checks

The following commands were run in the below directories:

* `aptos-core/`

    ```zsh
    pre-commit run --all-files
    scripts/rust_lint.sh
    ```

* `aptos-core/crates/aptos/`

    ```zsh
    cargo test
    ```

* `aptos-core/developer-docs-site/`

    ```
    pnpm spellcheck
    pnpm lint
    ```

### DIFF
--- a/crates/aptos/src/move_tool/mod.rs
+++ b/crates/aptos/src/move_tool/mod.rs
@@ -14,9 +14,9 @@ use crate::{
     account::derive_resource_account::ResourceAccountSeed,
     common::{
         types::{
-            load_account_arg, CliConfig, CliError, CliTypedResult, ConfigSearchMode,
-            EntryFunctionArguments, MoveManifestAccountWrapper, MovePackageDir, ProfileOptions,
-            PromptOptions, RestOptions, TransactionOptions, TransactionSummary,
+            load_account_arg, ArgWithTypeVec, CliConfig, CliError, CliTypedResult,
+            ConfigSearchMode, EntryFunctionArguments, MoveManifestAccountWrapper, MovePackageDir,
+            ProfileOptions, PromptOptions, RestOptions, TransactionOptions, TransactionSummary,
         },
         utils::{
             check_if_file_exists, create_dir_if_not_exist, dir_default_to_current,
@@ -60,12 +60,11 @@ use move_core_types::{
 use move_package::{source_package::layout::SourcePackageLayout, BuildConfig};
 use move_unit_test::UnitTestingConfig;
 pub use package_hooks::*;
-use serde::Serialize;
+use serde::{Deserialize, Serialize};
 use std::{
     collections::BTreeMap,
     convert::TryFrom,
     fmt::{Display, Formatter},
-    ops::Deref,
     path::{Path, PathBuf},
     str::FromStr,
 };
@@ -1127,7 +1126,7 @@ impl CliCommand<Vec<serde_json::Value>> for ViewFunction {
 
     async fn execute(self) -> CliTypedResult<Vec<serde_json::Value>> {
         let mut args: Vec<serde_json::Value> = vec![];
-        for arg in self.entry_function_args.args {
+        for arg in self.entry_function_args.arg_vec.args {
             args.push(arg.to_json()?);
         }
 
@@ -1151,13 +1150,8 @@ pub struct RunScript {
     pub(crate) txn_options: TransactionOptions,
     #[clap(flatten)]
     pub(crate) compile_proposal_args: CompileScriptFunction,
-    /// Arguments combined with their type separated by spaces.
-    ///
-    /// Supported types [u8, u16, u32, u64, u128, u256, bool, hex, string, address, raw]
-    ///
-    /// Example: `address:0x1 bool:true u8:0 u256:1234`
-    #[clap(long, multiple_values = true)]
-    pub(crate) args: Vec<ArgWithType>,
+    #[clap(flatten)]
+    pub(crate) arg_vec: ArgWithTypeVec,
     /// TypeTag arguments separated by spaces.
     ///
     /// Example: `u8 u16 u32 u64 u128 u256 bool address vector signer`
@@ -1177,7 +1171,7 @@ impl CliCommand<TransactionSummary> for RunScript {
             .compile("RunScript", self.txn_options.prompt_options)?;
 
         let mut args: Vec<TransactionArgument> = vec![];
-        for arg in self.args {
+        for arg in self.arg_vec.args {
             args.push(arg.try_into()?);
         }
 
@@ -1201,7 +1195,6 @@ pub(crate) enum FunctionArgType {
     Address,
     Bool,
     Hex,
-    HexArray,
     String,
     U8,
     U16,
@@ -1210,7 +1203,6 @@ pub(crate) enum FunctionArgType {
     U128,
     U256,
     Raw,
-    Vector(Box<FunctionArgType>),
 }
 
 impl Display for FunctionArgType {
@@ -1219,7 +1211,6 @@ impl Display for FunctionArgType {
             FunctionArgType::Address => write!(f, "address"),
             FunctionArgType::Bool => write!(f, "bool"),
             FunctionArgType::Hex => write!(f, "hex"),
-            FunctionArgType::HexArray => write!(f, "hex_array"),
             FunctionArgType::String => write!(f, "string"),
             FunctionArgType::U8 => write!(f, "u8"),
             FunctionArgType::U16 => write!(f, "u16"),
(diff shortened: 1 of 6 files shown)


crates/aptos/src/move_tool/mod.rs
```diff
@@ -1418,137 +1404,140 @@ impl FromStr for FunctionArgType {
...
+#[derive(Parser)]
+pub struct ArgWithTypeVec {
+    /// Arguments combined with their type separated by spaces.
+    ///
+    /// Supported types [address, bool, hex, string, u8, u16, u32, u64, u128, u256, raw]
+    ///
+    /// Vectors may be specified using JSON array literal syntax (you may need to escape this with
+    /// quotes based on your shell interpreter)
+    ///
+    /// Example: `address:0x1 bool:true u8:0 u256:1234 "bool:[true, false]" 'address:[["0xace", "0xbee"], []]'`
```
gregnazario, line 1436: This is pretty clean, might have to make a version bump and note that it's a breaking change in the CHANGELOG.md

crates/aptos/src/move_tool/mod.rs
```diff
@@ -1418,137 +1404,140 @@ impl FromStr for FunctionArgType {
...
             CliError::UnexpectedError(format!("Failed to parse argument to JSON {}", err))
         })
     }
 }
 
+/// Does not support string arguments that contain the following characters:
+///
+/// * `,`
+/// * `[`
+/// * `]`
```
gregnazario, line 1540: We'll need to add escaping soon

gregnazario (commented):
Pretty nice, I'll have to test it out and all, been very busy.  Noting that this is a breaking change.  @0xjinn can take a look as well

crates/aptos/src/move_tool/mod.rs
alnoki: Agreed.

crates/aptos/src/move_tool/mod.rs
alnoki: Can do, I'm glad to add these unless you would rather handle it.

alnoki:
@gregnazario I've addressed your comments above and resolved merge conflicts, so this should be ready to land!

Let me know if you or @0xjinn have anything to discuss

developer-docs-site/docs/tools/aptos-cli-tool/use-aptos-cli.md
```diff
@@ -719,6 +719,54 @@ The `peer_config.yaml` file will be created in your current working directory, w
 
 Move examples can be found in the [Move section](../../move/move-on-aptos/cli).
 
+You can also pass multiply-nested vector arguments, like in the `cli_args` example from [move-examples](https://github.com/aptos-labs/aptos-core/tree/main/aptos-move/move-examples):
```
gregnazario, line 722: `multi-nested` vs `multiply-nested` sounds like you're multiplying :)

crates/aptos/src/move_tool/mod.rs
```diff
@@ -1223,130 +1214,134 @@ impl Display for FunctionArgType {
...
+    loop {
+        let cur = val & 0x7F;
+        if cur != val {
+            binary.push((cur | 0x80) as u8);
+            val >>= 7;
+        } else {
+            binary.push(cur as u8);
+            break;
+        }
     }
```
gregnazario, line 1344: Is there a reason we're rewriting the BCS encoding logic?

Should be able to just put it in a vector and then BCS encode it.

crates/aptos/src/move_tool/mod.rs
```diff
@@ -1223,130 +1214,134 @@ impl Display for FunctionArgType {
...
+                    // Verify all sub-arguments have same depth.
+                    if let Some(check_depth) = common_sub_arg_depth {
+                        if check_depth != sub_arg_depth {
+                            return Err(CliError::CommandArgumentError(
+                                "Variable vector depth".to_string(),
+                            ));
+                        }
+                    };
+                    common_sub_arg_depth = Some(sub_arg_depth);
+                    bcs.append(&mut sub_arg_bcs); // Append sub-argument BCS.
```
gregnazario, line 1314: My only concern is that we're reimplementing BCS, but probably points out that maybe BCS needs a more interactive streaming approach in Rust like you're doing here.

crates/aptos/src/test/mod.rs
```diff
@@ -591,13 +593,15 @@ impl CliTestFramework {
...
-                    ArgWithType::bytes(vec![]),
-                ],
+                arg_vec: ArgWithTypeVec {
+                    args: vec![
+                        ArgWithType::address(self.account_id(operator_index)),
+                        ArgWithType::address(self.account_id(voter_index)),
+                        ArgWithType::u64(amount),
+                        ArgWithType::u64(commission_percentage),
+                        ArgWithType::bytes(vec![]),
+                    ],
```
gregnazario, line 603: We should probably add a unit test for this function now (I feel like I've avoided it too long), that checks the multi-nested BCS, and ensures it can be deserialized as the type expected.

gregnazario (approved):
I'm going to approve this @0xjinn can you take another look at this.

crates/aptos/src/move_tool/mod.rs
```diff
@@ -1223,130 +1214,134 @@ impl Display for FunctionArgType {
...
-                let raw = hex::decode(arg)
-                    .map_err(|err| CliError::UnableToParse("raw", err.to_string()))?;
-                Ok(raw)
+                hex::decode(arg).map_err(|err| CliError::UnableToParse("raw", err.to_string()))
+            },
+        }
+    }
+
+    /// Recursively parse argument JSON into BCS representation.
+    pub fn parse_arg_json(&self, arg: &serde_json::Value) -> CliTypedResult<ArgWithType> {
```
0xjinn, line 1276: nit: Would be great if we can unit test and safeguard the parser

crates/aptos/src/move_tool/mod.rs
alnoki: > Should be able to just put it in a vector and then BCS encode it.

```rust
            serde_json::Value::Array(_) => {
                // Vector of sub-argument BCS representations.
                let mut sub_args: Vec<Vec<u8>> = vec![]; // <------------------------------------- 1
                let mut common_sub_arg_depth = None;
                // Loop over all of the vector's sub-arguments, which may also be vectors:
                for sub_arg in arg.as_array().unwrap() {
                    let ArgWithType {
                        _ty: _,
                        _vector_depth: sub_arg_depth,
                        arg: sub_arg_bcs,
                    } = self.parse_arg_json(sub_arg)?;
                    // Verify all sub-arguments have same depth.
                    if let Some(check_depth) = common_sub_arg_depth {
                        if check_depth != sub_arg_depth {
                            return Err(CliError::CommandArgumentError(
                                "Variable vector depth".to_string(),
                            ));
                        }
                    };
                    common_sub_arg_depth = Some(sub_arg_depth);
                    // Push back sub-argument BCS.
                    sub_args.push(sub_arg_bcs); // <---------------------------------------------- 2
                }
                // Default sub-argument depth is 0 for when no sub-arguments were looped over.
                Ok(ArgWithType {
                    _ty: self.clone(),
                    _vector_depth: common_sub_arg_depth.unwrap_or(0) + 1,
                    arg: bcs::to_bytes(&sub_args).map_err(|err| CliError::BCS("arg", err))?,
                    // ^ ------------------------------------------------------------------------- 3
                })
            },
            serde_json::Value::Null => {
                Err(CliError::CommandArgumentError("Null argument".to_string()))
            },
            serde_json::Value::Object(_) => Err(CliError::CommandArgumentError(
                "JSON object argument".to_string(),
            )),
```

At first glance it might appear that the above algorithm does what you are asking:

1. Declare a vector of sub-argument BCS encodings (`Vec<Vec<u8>>`)
2. Push all sub-argument BCS encodings onto the vector
3. Convert the entire `Vec<Vec<u8>>` to BCS

However, the problem with this approach is that the BCS encoding logic generates not only a `uleb128` for the top-level vector, but also for each nested vector within. This shouldn't happen, however, because the relevant BCS `Vec<u8>` encoding of each sub-argument has already been calculated (including with the leading `uleb128`) during recursion. The problem here is that when the BCS  encoder parses `Vec<Vec<u8>>` it doesn't treat each subvector `Vec<u8>` as already being in BCS form.

The reason I've rewritten the BCS logic is because it is not publicly available. Note that the same thing is already happening in [`transaction_arg_validation.rs`](https://github.com/aptos-labs/aptos-core/blob/6aed685d6c5288fa039a9ce59908fd7809d861c1/aptos-move/aptos-vm/src/verifier/transaction_arg_validation.rs#L374)

crates/aptos/src/move_tool/mod.rs
alnoki: Agreed. I go into more depth in the above comment https://github.com/aptos-labs/aptos-core/pull/7790#discussion_r1178402832

developer-docs-site/docs/tools/aptos-cli-tool/use-aptos-cli.md
alnoki: Resolved in 329d1e5

crates/aptos/src/test/mod.rs
alnoki: Agreed. I am also working on a PR that will extend this functionality to include JSON input files, and it looks like it'll involve some minor refactors to struct nesting. Once we can get that in perhaps it'll be appropriate to add more unit testing?

crates/aptos/src/move_tool/mod.rs
alnoki: Agreed. As I mention in https://github.com/aptos-labs/aptos-core/pull/7790#discussion_r1178408642 I'm working on a separate PR that will extend support to JSON input files as well. Perhaps once both input types are supported an end-to-end test will be in order?

alnoki:
@gregnazario I believe I've addressed all of your comments/suggestions.

@0xjinn I believe I've addressed your proposition.

Is there anything else needed to get an approving review from @0xjinn and land in `main`?

0xjinn:
I think this looks good on my end. I did manual tested it and ensured the [example transaction](https://explorer.aptoslabs.com/txn/0xef51aff503f9c81f0d033c7c253ce50792a6f996a8018b26494e7e0a9a47d51f/userTxnOverview?network=testnet) works.

I can follow up with PR to update the changelog.

cc: @gregnazario for final look

0xjinn (approved):

davidiw:
claims it needs to be rebased

alnoki:
> claims it needs to be rebased

@davidiw I just merged in `main` without conflicts, perhaps the notice was due to the time since branching?

Either way @gregnazario this should be ready to land

alnoki:
@davidiw @gregnazario I've further squashed the PR so it reflects as a single commit against `main`

There still are no merge conflicts, so I'm not sure if there's anything else I can do to massage the diff history

lightmark:
love this @alnoki

alnoki:
> love this @alnoki

@lightmark see also https://github.com/aptos-labs/aptos-core/pull/8054

PR #7037, opened 2023-03-09
[API] Create counter for request source

banool:
### Description
We want some more insight into where requests to the API are coming from, this counter helps with that.

@0xmaayan this slices up the requests by source, method they're calling, and status they got. Does that sound good, or did you want the data sliced up differently?

### Test Plan
Run local testnet:
```
cargo run -p aptos -- node run-local-testnet --force-restart --assume-yes --with-faucet
```

Make a couple of requests with just curl:
```
curl http://127.0.0.1:8080/v1
curl http://127.0.0.1:8080/v1/accounts/0x1
```
Make a couple of requests with the CLI, to which we just added the User-Agent stuff.
```
aptos account list --url http://127.0.0.1:8080 --account 0x1
```

See the metrics pushed appropriately, with `request-source-client-unknown` for the former and `aptos-cli/1.8.0` for the latter:
```
curl -s http://127.0.0.1:9101/metrics | grep -i request_source
```
```
# HELP aptos_api_request_source_client API requests grouped by source (e.g. which SDK, unknown, etc), operation_id, and status
# TYPE aptos_api_request_source_client counter
aptos_api_request_source_client{operation_id="estimate_gas_price",request_source_client="request-source-client-unknown",status="200"} 1
aptos_api_request_source_client{operation_id="get_account",request_source_client="request-source-client-unknown",status="200"} 1
aptos_api_request_source_client{operation_id="get_account",request_source_client="request-source-client-unknown",status="404"} 1
aptos_api_request_source_client{operation_id="get_account_resources",request_source_client="aptos-cli/1.0.8",status="200"} 1
aptos_api_request_source_client{operation_id="get_ledger_info",request_source_client="request-source-client-unknown",status="200"} 2
aptos_api_request_source_client{operation_id="get_transaction_by_hash",request_source_client="request-source-client-unknown",status="200"} 13
aptos_api_request_source_client{operation_id="submit_transaction",request_source_client="request-source-client-unknown",status="202"} 1
```

PR #7797, opened 2023-04-16
[API] Log aptos client based on a custom header rather than user agent

banool:
### Description
Turns out some browsers don't let you modify the User-Agent string so we need to use our own header to identify the client.

See also the PR that originally added this using the User-Agent string: https://github.com/aptos-labs/aptos-core/pull/7037.

### Test Plan
Run local testnet:
```
cargo run -p aptos -- node run-local-testnet --force-restart --assume-yes --with-faucet
```

Make a couple of requests with just curl:
```
curl http://127.0.0.1:8080/v1
curl http://127.0.0.1:8080/v1/accounts/0x1
```
Make a couple of requests with the header set, to which we just added the User-Agent stuff.
```
curl -H 'X-Aptos-Client: aptos-cli/1.0.11' http://127.0.0.1:8080/v1/accounts/0x1
```

See the metrics pushed appropriately, with `request-source-client-unknown` for the former and `aptos-cli/1.8.0` for the latter:
```
curl -s http://127.0.0.1:9101/metrics | grep -i request_source
```
```
aptos_api_request_source_client{operation_id="estimate_gas_price",request_source_client="unknown",status="200"} 1
aptos_api_request_source_client{operation_id="get_account",request_source_client="aptos-cli/1.0.11",status="200"} 1
aptos_api_request_source_client{operation_id="get_account",request_source_client="unknown",status="200"} 2
aptos_api_request_source_client{operation_id="get_account",request_source_client="unknown",status="404"} 1
aptos_api_request_source_client{operation_id="get_account_resources",request_source_client="unknown",status="200"} 1
aptos_api_request_source_client{operation_id="get_ledger_info",request_source_client="unknown",status="200"} 2
aptos_api_request_source_client{operation_id="get_transaction_by_hash",request_source_client="unknown",status="200"} 37
aptos_api_request_source_client{operation_id="submit_transaction",request_source_client="unknown",status="202"} 3
```

### DIFF
--- a/api/src/log.rs
+++ b/api/src/log.rs
@@ -3,6 +3,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use crate::metrics::{HISTOGRAM, REQUEST_SOURCE_CLIENT, RESPONSE_STATUS};
+use aptos_api_types::X_APTOS_CLIENT;
 use aptos_logger::{
     debug, info,
     prelude::{sample, SampleRate},
@@ -36,6 +37,10 @@ pub async fn middleware_log<E: Endpoint>(next: E, request: Request) -> Result<Re
         user_agent: request
             .headers()
             .get(header::USER_AGENT)
+            .and_then(|v| v.to_str().ok().map(|v| v.to_string())),
+        aptos_client: request
+            .headers()
+            .get(X_APTOS_CLIENT)
             .and_then(|v| v.to_str().ok().map(|v| v.to_string())),
         elapsed: Duration::from_secs(0),
         forwarded: request
@@ -79,7 +84,7 @@ pub async fn middleware_log<E: Endpoint>(next: E, request: Request) -> Result<Re
     // Push a counter based on the request source, sliced up by endpoint + method.
     REQUEST_SOURCE_CLIENT
         .with_label_values(&[
-            determine_request_source_client(&log.user_agent),
+            determine_request_source_client(&log.aptos_client),
             response
                 .data::<OperationId>()
                 .map(|operation_id| operation_id.0)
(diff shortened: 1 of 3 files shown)


api/src/log.rs
```diff
@@ -79,7 +84,7 @@ pub async fn middleware_log<E: Endpoint>(next: E, request: Request) -> Result<Re
     // Push a counter based on the request source, sliced up by endpoint + method.
     REQUEST_SOURCE_CLIENT
         .with_label_values(&[
-            determine_request_source_client(&log.user_agent),
+            determine_request_source_client(&log.aptos_client),
```
davidiw, line 87: are we already using the user agent in other cases? do we need to keep both online for some period of time during the transition?

0xmaayan (approved):

api/types/src/headers.rs
```diff
@@ -17,3 +17,5 @@ pub const X_APTOS_OLDEST_BLOCK_HEIGHT: &str = "X-Aptos-Oldest-Block-Height";
 pub const X_APTOS_LEDGER_TIMESTAMP: &str = "X-Aptos-Ledger-TimestampUsec";
 /// Cursor used for pagination.
 pub const X_APTOS_CURSOR: &str = "X-Aptos-Cursor";
+/// Provided by the client to identify what client it is.
+pub const X_APTOS_CLIENT: &str = "x-aptos-client";
```
banool, line 21: This is lowercase on purpose, using uppercase letters in a header name is actually not standard compliant. We should change the other ones too at some point.

api/src/log.rs
gregnazario: I think it could be useful in general for just viewing clients that aren't created by us.  So, we can revisit this later.

let's also add a way to set it in the CLI / Rust SDK / Python SDK

gregnazario (approved):

PR #7799, opened 2023-04-17
[Peer Monitor] Add build info to the node info requests.

JoshLind:
### Description
This PR updates the peer monitoring service to share the build information for all peers (instead of just the git_hash). This should help improve build visibility in the wild west 😄 

### Test Plan
New and existing tests.

### DIFF
--- a/network/peer-monitoring-service/client/src/peer_states/node_info.rs
+++ b/network/peer-monitoring-service/client/src/peer_states/node_info.rs
@@ -14,7 +14,11 @@ use aptos_peer_monitoring_service_types::{
     response::{NodeInformationResponse, PeerMonitoringServiceResponse},
 };
 use aptos_time_service::TimeService;
-use std::sync::Arc;
+use std::{collections::BTreeMap, sync::Arc};
+
+// The maximum number of entries and string lengths in the build info map
+const MAX_BUILD_INFORMATION_ENTRIES: usize = 100;
+const MAX_BUILD_INFORMATION_STRING_LENGTH: usize = 200;
 
 /// A simple container that holds a single peer's node info
 #[derive(Clone, Debug)]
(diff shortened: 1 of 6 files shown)


network/peer-monitoring-service/client/src/peer_states/node_info.rs
```diff
@@ -96,6 +99,20 @@ impl StateValueInterface for NodeInfoState {
             },
         };
 
+        // Verify the build information is not too large
+        let build_information_length = node_info_response.build_information.len();
```
ibalajiarun, line 103: What if the individual key lengths and values lengths are too large, for instance? that could be an attack vector too, because they are strings?

JoshLind: Yes! Thanks @ibalajiarun! Will fix when I'm back at my laptop!

brianolson: Agree, maybe set a total length for the string+string map? Maybe sum of length of keys and values at around 1 kB?

JoshLind: Cool, so I've made a simple fix for this, we limit both: (i) the number of entries in the map; and (ii) the size of each key and value. If a response violates (i), we reject it (because this must be misbehaviour). If a response violates (ii), we trim and store the strings (because this may actually be an unfortunate case, e.g., bad naming of a branch, etc.) If we don't think this is strong enough, we can always improve it in the future.

Note that even if we didn't patch this right now, we still have a number of layers protecting us:
- The client side of the monitoring service is still disabled, so this isn't a DOS vector today.
- We have hard limits on network messages, so even if someone wanted to try this, it would take a lot of work to DOS the node (and I'm still not 100% sure it'd be doable 😄).

My purpose in wanting to land this ASAP is to stabilize the server side. Client side fixes (such as validations) will come soon 😄 

Let me know what folks think.

JoshLind:
Marking this as draft -- may as well fix all the things while folks are looking at it 😄

JoshLind:
FYI: Closing this out and gonna follow-up with a better approach.

network/peer-monitoring-service/client/src/peer_states/node_info.rs
JoshLind: Makes sense! Will follow up 😄

PR #7806, opened 2023-04-17
[Python SDK] Fix code format and update poetry lock

sionescu:
### Description

The ecosystem-python-lint workflow is failing because the code formatting is incorrect: https://github.com/aptos-labs/aptos-core/actions/runs/4724604620/jobs/8382023711?pr=7746
Make target "fmt" runs isort and black in the wrong order: it should be first isort then black.

Poetry was issuing "Warning: poetry.lock is not consistent with pyproject.toml. You may be getting improper dependencies".

Add script that prints the modified files in case the formatter finds something, then fails the workflow.
Allow invoking lint-test.yaml manually.

Executed:

$ cd ecosystem/python/sdk
$ poetry install
$ poetry lock
$ make fmt


ecosystem/python/sdk/Makefile
```diff
@@ -6,8 +6,8 @@ test:
 
 fmt:
 	find ./examples ./aptos_sdk *.py -type f -name "*.py" | xargs poetry run autoflake -i -r --remove-all-unused-imports --remove-unused-variables --ignore-init-module-imports
-	poetry run black aptos_sdk examples setup.py
-	poetry run isort aptos_sdk examples setup.py
+	- poetry run isort aptos_sdk examples setup.py
+	- poetry run black aptos_sdk examples setup.py
```
davidiw, line 10: why? `-` means ignore failures
I don't care about the ordering.

davidiw (commented):
looks a bit nicer

ecosystem/python/sdk/Makefile
sionescu: Because isort and black enforce different coding styles and they will undo each other's changes. Isort returns 1 if any file was reformatted, indicating an error, but if black reformats the code back after it, it doesn't mean anything. It's a spurious error.
The real error checking is done in lint-test.yaml: if after running all the formatters, there are modified files locally, it's an error, otherwise pass.

davidiw: that's not accurate... if there are syntax issues, isort and black will fail to run and then the linter will pass becaues there are no changes.

sionescu: Right now isort and black are *always* giving false positives, so their use is kind of pointless.

sionescu: Ideally we should be able to ask both tools ignore reformatted files and only throw errors for more serious conditions.

davidiw (approved):

JoshLind (approved):
Unblocking 😄

sionescu:
@JoshLind thanks.

PR #7808, opened 2023-04-17
[Peer Monitor] Add build info to the node info requests and max message sizes.

JoshLind:
### Description
This PR makes several small improvements to the peer monitoring service (each in their own commit):
1. Update the peer monitoring service to share the build information for all peers (instead of just the git_hash). This should help improve build visibility in the wild west 😄 
2. Change the type of the connected peers map from HashMap to BTreeMap (not super necessary, but determinism makes life easier for debugging).
3. Add a message size limit on all peer monitoring messages to prevent malicious peers from forcing us to hold large messages in memory. The message limit is configurable (and defaults to 100KB), which should be more than enough for all messages (including the future messages for discovery).

### Test Plan
New and existing tests.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2323,6 +2323,7 @@ version = "0.1.0"
 dependencies = [
  "aptos-config",
  "aptos-types",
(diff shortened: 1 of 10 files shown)


ibalajiarun (approved):

bchocho (approved):
LGTM

PR #7809, opened 2023-04-17
[docker] Rename experimental builder

ibalajiarun:
### Description

This PR renames the `docker/experimental` directory to `docker/builder` to indicate proper promotion of the new Docker build flow. It also includes a `README.md` file that describes how the builder works at a high level and also instruction on how to extend it to add new images/binaries. 

Also, cleans up the now old builder steps.


sherry-x:
shall we also make performance build the default build profile?

rustielin (approved):

rustielin:
> shall we also make performance build the default build profile?

Not sure about this. Perhaps we can still use the default release profile for PR testing but then use performance to build and test releases? It's still a bit expensive to do performance builds on PRs with our current setup

sionescu (approved):

PR #7817, opened 2023-04-18
small improvements

gerben-stavenga:
### Description

Refactor and remove code. The move-vm suffers from a lot of generics and traits that IMO unnecessarily clutter the code. In this PR I try to make the implementation more concrete by removing dead traits (DataStore), remove unnecessary generics (MoveResolver::Error) and introduce "dyn" instead of "impl" at boundaries where a dynamic function call does not introduce any appreciable overhead. In our case whenever storage needs to be resolved (MoveResolver) the resolver is called through dynamic dispatch instead of inlining. This cuts a dependency and improves factorization
 

### Test Plan
No semantic changes


aptos-move/framework/src/natives/helpers.rs
```diff
@@ -171,30 +171,30 @@ macro_rules! safely_pop_type_arg {
 }
 
 #[allow(unused)]
-pub struct SafeNativeContext<'a, 'b, 'c> {
+pub struct SafeNativeContext<'a, 'b, 'c, 'd> {
```
wrwg, line 174: Curious: why did this actually get more complex, requiring another lifetime?

third_party/move/move-vm/runtime/src/interpreter.rs
```diff
@@ -536,14 +536,14 @@ impl Interpreter {
     }
 
     /// Loads a resource from the data store and return the number of bytes read from the storage.
-    fn load_resource<'b>(
+    fn load_resource<'c>(
         loader: &Loader,
+        data_store: &'c mut TransactionDataCache,
```
wrwg, line 541: Wonder whether this is really an improvement, before the interpreter would work on any impl of a DataStore, now we specialize it to work on the cache directly.

wrwg (commented):
Loving the general direction and removing some of the unnecessary trait complexity, as well as switching to `dyn` for simplicity.

third_party/move/move-vm/runtime/src/interpreter.rs
gerben-stavenga: I do think conceptually this is an improvement, the value cache is important for performance and I think we would want that to be inlined/directly called. For me, intuitively, it seems clear that it's the actual lookup of data from the underlying (untyped) db that is the clear delineation for using dyn. Indeed the MoveResolver has different implementation for testing/prod etc. 

Previously the interpreter was using "impl DataStore" which would specialize for each implementation of TransactionDataCache<S> where S is the MoveResolver used. Because now we use dyn in TransactionDataCache, this has become a non-generic type and removing the DataStore trait allows TransactionDataCache to not carry the loader, which is already present as direct parameters.

wrwg (approved):

movekevin (approved):

third_party/move/extensions/async/move-async-vm/src/async_vm.rs
```diff
@@ -77,12 +77,12 @@ impl AsyncVM {
...
 
     /// Creates a new session.
-    pub fn new_session<'r, 'l, S: MoveResolver>(
+    pub fn new_session<'r, 'l>(
         &'l self,
         for_actor: AccountAddress,
         virtual_time: u128,
-        move_resolver: &'r mut S,
-    ) -> AsyncSession<'r, 'l, S> {
+        move_resolver: &'r mut dyn MoveResolver,
```
zekun000, line 84: not sure if I understand why this can't be impl? how does it reduce the dependency?

gerben-stavenga: The moment you go through a dyn using dynamic dispatch, you lose syntactic dependence on the exact type. When using impl the code still depends on the yet unknown implementation of S: MoveResolver. This means that the compilation happens when the function is instantiated with a concrete type. The actual compilation of the function is postponed till S is known, ie. the compilation has a dependency on knowing S. 

Ideally in the future we would do a similar thing for GasMeter (it requires more thought because you don't want dynamic dispatch for each instruction) in which case the interpreter loop would be a regular function and compilation would happen completely at crate building time and not at monomorphizing time

zekun000: oh i see you're talking about compilation dependencies. yes impl is basically the same as generic static dispatch compared to dyn.

PR #7819, opened 2023-04-18
add a fungible token move-e2e test

lightmark:
### Description
as title

### Test Plan
cargo test


aptos-move/move-examples/fungible_token/sources/managed_fungible_token.move
```diff
@@ -0,0 +1,175 @@
+module fungible_token::managed_fungible_token {
```
movekevin, line 1: Can you add high-level comments explaining what a fungible token is and how it's created? Also descriptions of the different functionalities provided here would be great

types/src/transaction/authenticator.rs
```diff
@@ -360,17 +361,17 @@ impl AccountAuthenticator {
 /// A struct that represents an account authentication key. An account's address is the last 32
 /// bytes of authentication key used to create it
 #[derive(
-    Clone,
```
movekevin, line 363: Where did all these refactorings come from?

movekevin (approved):
Looks good with a few minor comments

bowenyang007 (approved):

PR #7822, opened 2023-04-18
add x-aptos-client header

0xmaayan:
### Description
Adds `x-aptos-client` header to REST calls. 

**Needs to be merged after** https://github.com/aptos-labs/aptos-core/pull/7797
### Test Plan

- tested on Chrome, Safari, Brave and Edge browsers
image
image
image
image

- testes are passing

### DIFF
--- a/ecosystem/typescript/sdk/CHANGELOG.md
+++ b/ecosystem/typescript/sdk/CHANGELOG.md
@@ -6,6 +6,8 @@ All notable changes to the Aptos Node SDK will be captured in this file. This ch
 
 ## Unreleased
 
+- Add `x-aptos-client` header to `AptosClient` requests
+
 ## 1.9.0 (2023-05-17)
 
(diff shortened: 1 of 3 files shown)


0xjinn:
Shall we do a manual test against the aptos explorer?

0xjinn (approved):

gregnazario (approved):

banool:
We cannot release this feature until the API side change (https://github.com/aptos-labs/aptos-core/pull/7797) rolls out to all devnet, testnet, and mainnet fullnodes. Even then we should let it bake for a bit to ensure all community fullnodes have updated. If a client with the change in this (7822) PR tries to talk to an API without the change from #7797, it will fail because the CORS policy does not allow the new header.

gregnazario (changes requested):
Blocked on CORS updates

banool (approved):

PR #7823, opened 2023-04-18
[Quorum Store] Update and sanity check consensus block and quorum store batch sizes

bchocho:
### Description

Sanity check consensus block sizes and quorum store batch sizes, independently and between each other.

Update batch and block sizes to support at least one 1 MiB txn, and to satisfy the sanity checks.


config/src/config/consensus_config.rs
```diff
@@ -176,30 +176,30 @@ impl Default for ConsensusConfig {
...
                     max_sending_block_txns_override: 200,
                     // stop reducing size, so 100k transactions can still go through
-                    max_sending_block_bytes_override: 100 * 1024,
+                    max_sending_block_bytes_override: 1024 * 1024,
                 },
                 ChainHealthBackoffValues {
                     backoff_if_below_participating_voting_power_percentage: 69,
                     // in practice, latencies make it such that 2-4 blocks/s is max,
                     // meaning that most aggressively we limit to ~200-400 TPS
                     max_sending_block_txns_override: 100,
```
igor-aptos, line 201: if batch size of 100 is too small, we can remove this level. What do you think?

(I can add delay to chain health backoff as well, and then we shouldn't need so small blocks)

zekun000 (approved):
the sanitizer looks awesome

config/src/config/consensus_config.rs
bchocho: Bigger is probably better 😄 but do you think it is safe to remove it without additional testing?

igor-aptos (approved):

PR #7826, opened 2023-04-18
Fix several major issues in fungible asset 

lightmark:
### Description
1. Fix merge function
2. Remove the checking of supply underflow and delegate this to aggregator to unblock paralellism.
3. **Only enable adding fungibility to non-deletable objects**.
4. Fix `ungated_balance_transfer_allowed` function that it will not return `true` if non-existent. This is the default behavior for primary_fungible_store but not any store.
5. The second commit rename `asset` to `metadata` in tests.

### Test Plan
ut

### DIFF
--- a/aptos-move/framework/aptos-framework/sources/fungible_asset.move
+++ b/aptos-move/framework/aptos-framework/sources/fungible_asset.move
@@ -2,7 +2,7 @@
 /// metadata object can be any object that equipped with `Metadata` resource.
 module aptos_framework::fungible_asset {
     use aptos_framework::event;
-    use aptos_framework::object::{Self, Object, ConstructorRef, DeleteRef, can_generate_delete_ref};
+    use aptos_framework::object::{Self, Object, ConstructorRef, DeleteRef};
     use aptos_framework::optional_aggregator::{Self, OptionalAggregator};
     use std::string;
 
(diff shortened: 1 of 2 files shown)


movekevin (approved):

aptos-move/framework/aptos-framework/sources/fungible_asset.move
```diff
@@ -2,7 +2,7 @@
 /// metadata object can be any object that equipped with `Metadata` resource.
 module aptos_framework::fungible_asset {
     use aptos_framework::event;
-    use aptos_framework::object::{Self, Object, ConstructorRef, DeleteRef};
+    use aptos_framework::object::{Self, Object, ConstructorRef, DeleteRef, can_generate_delete_ref};
```
davidiw, line 5: please don't import functions 🙏

davidiw (commented):
while this looks technically right, we now need to expose the means to specify that objects aren't deletable at creation time -- otherwise the only types that can be constructed are named objects 😬

aptos-move/framework/aptos-framework/sources/fungible_asset.move
lightmark: my bad

davidiw (approved):

PR #7830, opened 2023-04-18
Remove unnecessary *& syntax

movekevin:
### Description
*& is not necessary for copyable values.

### Test Plan
Code compiles.


gregnazario (approved):
Remember that all of these will need to be in a framework release to ensure the verifiability of the source code.

gerben-stavenga (approved):

PR #7831, opened 2023-04-18
adds weekly Docker image updater

gedigi:
### Description
Adds a script and workflow that checks the latest digest of the Aptos container and updates the Dockerfile as needed. The workflow runs weekly.

### Test Plan


.github/workflows/docker-update-images.yaml
```diff
@@ -0,0 +1,26 @@
+# Continuously build Rosetta
+name: Docker Update Images
+
+on:
```
perryjrandall, line 4: Add a "pull request" that matches this file and the python file path

That way your PR will actually run and test this workflow

.github/workflows/docker-update-images.yaml
```diff
@@ -0,0 +1,26 @@
...
+  schedule:
+    - cron: "0 9 * * 1" # once a week
+
+jobs:
+  update:
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Check and pin updates (if any)
```
perryjrandall, line 13: You might need to add setup python step? Although without dependencies it might just work tm

perryjrandall (approved):
Generally LGTM, please add a test plan

.github/workflows/docker-update-images.yaml
perryjrandall: Also you'll wanna add a workflow_dispatch so that it can be triggered manually as well

sionescu (approved):

PR #7836, opened 2023-04-19
[SDK] Fix MultisigTransactionPayload deserialization

movekevin:
### Description
The serialized multisig sig transaction payload has an extra enum value that indicates which type of payload the multisig tx contains (currently only entry function is supported, but support for script payload can be added in the future). serialize() correctly takes this enum into value but deserialize currently doesn't.

### DIFF
--- a/ecosystem/typescript/sdk/src/aptos_types/transaction.ts
+++ b/ecosystem/typescript/sdk/src/aptos_types/transaction.ts
@@ -225,6 +225,9 @@ export class MultiSigTransactionPayload {
   }
 
   static deserialize(deserializer: Deserializer): MultiSigTransactionPayload {
+    // TODO: Support other types of payload beside EntryFunction.
+    // This is the enum value indicating which type of payload the multisig tx contains.
+    deserializer.deserializeUleb128AsU32();
     return new MultiSigTransactionPayload(EntryFunction.deserialize(deserializer));
   }
 }


ecosystem/typescript/sdk/src/aptos_types/transaction.ts
```diff
@@ -225,6 +225,9 @@ export class MultiSigTransactionPayload {
   }
 
   static deserialize(deserializer: Deserializer): MultiSigTransactionPayload {
+    // TODO: Support other types of payload beside EntryFunction.
+    // This is the enum value indicating which type of payload the multisig tx contains.
+    deserializer.deserializeUleb128AsU32();
     return new MultiSigTransactionPayload(EntryFunction.deserialize(deserializer));
```
gregnazario, line 231: Is there a reason you don't just deserialize the number and then throw an appropriate error saying it's not an entry function?

gregnazario (approved):

0xmaayan (approved):

ecosystem/typescript/sdk/src/aptos_types/transaction.ts
movekevin: We don't support a different payload type anyway so it's not possible to have an unexpected type. If user somehow obtains an invalid hex string from somewhere and tries to deserialize it, that'd fail anyway with deserialization error. Not sure how useful it is to just return an error message based on that single enum bit

PR #7845, opened 2023-04-19
Remove initialize function from execution config

sitalkedia:
### Description

As above to simplify the governance proposal. 

### Test Plan

Existing UTs.

### DIFF
--- a/aptos-move/framework/aptos-framework/sources/configs/execution_config.spec.move
+++ b/aptos-move/framework/aptos-framework/sources/configs/execution_config.spec.move
@@ -4,16 +4,7 @@ spec aptos_framework::execution_config {
         pragma aborts_if_is_strict;
     }
 
-    /// Ensure caller is admin.
-    spec initialize(aptos_framework: &signer, config: vector<u8>) {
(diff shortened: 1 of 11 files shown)


zekun000 (approved):

aptos-move/framework/aptos-framework/sources/configs/execution_config.spec.move
```diff
@@ -4,16 +4,7 @@ spec aptos_framework::execution_config {
         pragma aborts_if_is_strict;
     }
 
-    /// Ensure caller is admin.
-    spec initialize(aptos_framework: &signer, config: vector<u8>) {
```
movekevin, line 8: This only removes the spec. Can you remove that initialize function in execution.move as well?

sitalkedia: Ah, my bad. I had this change locally but didn't include that in the PR.

sitalkedia: Done!

PR #7852, opened 2023-04-19
dag driver shutdown

zjma:

### DIFF
--- a/consensus/src/dag/dag_driver.rs
+++ b/consensus/src/dag/dag_driver.rs
@@ -47,6 +47,7 @@ pub struct DagDriver {
     dag: Dag,
     bullshark: Arc<Mutex<Bullshark>>,
     rb_tx: Sender<ReliableBroadcastCommand>,
+    rb_close_tx: Option<oneshot::Sender<oneshot::Sender<()>>>,
     network_msg_rx: aptos_channel::Receiver<PeerId, VerifiedEvent>,
     time_service: Arc<dyn TimeService>,
 }
@@ -87,7 +88,10 @@ impl DagDriver {
             genesis_block_id,
         )));
 
-        spawn_named!("reliable_broadcast", rb.start(rb_network_msg_rx, rb_rx));
+        let (rb_close_tx, close_rx) = oneshot::channel();
+
+        spawn_named!("reliable_broadcast", rb.
+            start(rb_network_msg_rx, rb_rx, close_rx));
         // spawn_named!("bullshark", bullshark.start(dag_bullshark_rx));
 
         Self {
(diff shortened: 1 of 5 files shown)


consensus/src/epoch_manager.rs
```diff
@@ -759,8 +761,9 @@ impl EpochManager {
             self.time_service.clone(),
             genesis_block_id,
         );
-
-        tokio::spawn(dag_driver.start());
+        let (close_tx, close_rx) = oneshot::channel();
+        self.dag_driver_close_tx = Some(close_tx);
+        tokio::spawn(dag_driver.start(close_rx));
```
sasha8, line 766: We should call it from shutdown_current_processor

consensus/src/dag/dag_driver.rs
sasha8, line 1: We should also shutdown the worker in ReliableBroadcast.

zjma:
This change makes `consensus_fault_tolerance::test_no_failures` failing again....

consensus/src/epoch_manager.rs
zjma: Fixed.

consensus/src/dag/dag_driver.rs
zjma: Fixed.

consensus/src/dag/dag_driver.rs
```diff
@@ -224,10 +232,23 @@ impl DagDriver {
...
+                close_req = close_rx.select_next_some() => {
+                    if let Some(close_tx) = self.rb_close_tx.take() {
+                        let (ack_tx, ack_rx) = oneshot::channel();
+                        close_tx.send(ack_tx).expect("[DagDriver] failed to drop rb");
+                        ack_rx.await.expect("[DagDriver] failed to drop rb");
+                    }
+                    if let Ok(ack_sender) = close_req {
+                        ack_sender.send(()).expect("[DagDriver] Fail to ack shutdown");
+                    }
+                    break;
```
zjma, line 249: synchronized wait for the ack from rb. Is that what we want? @sasha8

consensus/src/dag/dag_driver.rs
```diff
@@ -108,6 +112,7 @@ impl DagDriver {
             ),
             bullshark,
             rb_tx,
+            rb_close_tx: Some(rb_close_tx),
```
sasha8, line 115: Does it have to be Some?

consensus/src/dag/dag_driver.rs
sasha8: Yes.

sasha8 (approved):

consensus/src/dag/dag_driver.rs
zjma: Surely it doesn't. Fixed.

PR #7856, opened 2023-04-19
[bytecode verifier] improved signature checker impl

vgao1996:
This implements a new signature checker that is both faster and stricter (checking even unused type signatures).

This new checker can be enabled by a verifier config option.

### DIFF
--- a/aptos-move/aptos-release-builder/data/release.yaml
+++ b/aptos-move/aptos-release-builder/data/release.yaml
@@ -1,6 +1,18 @@
 ---
 remote_endpoint: ~
 proposals:
+  - name: step_0
+    metadata:
+      title: "Enable the new signature checker"
+      description: "This enables the improved signature checker implementation."
+      source_code_url: "TBD"
+      discussion_url: "https://github.com/aptos-labs/aptos-core/pull/7856"
+    execution_mode: MultiStep
+    update_sequence:
+    - FeatureFlag:
+        enabled:
+          - signature_checker_v2
+        disabled: []
   - name: step_1
     metadata:
       title: "Enable Struct Constructor Feature"
--- a/aptos-move/aptos-release-builder/src/components/feature_flags.rs
+++ b/aptos-move/aptos-release-builder/src/components/feature_flags.rs
@@ -34,6 +34,7 @@ pub enum FeatureFlag {
     StructConstructors,
     PeriodicalRewardRateReduction,
     PartialGovernanceVoting,
+    SignatureCheckerV2,
 }
 
 fn generate_features_blob(writer: &CodeWriter, data: &[u64]) {
@@ -146,6 +147,7 @@ impl From<FeatureFlag> for AptosFeatureFlag {
                 AptosFeatureFlag::PERIODICAL_REWARD_RATE_DECREASE
             },
             FeatureFlag::PartialGovernanceVoting => AptosFeatureFlag::PARTIAL_GOVERNANCE_VOTING,
+            FeatureFlag::SignatureCheckerV2 => AptosFeatureFlag::SIGNATURE_CHECKER_V2,
         }
     }
 }
@@ -183,6 +185,7 @@ impl From<AptosFeatureFlag> for FeatureFlag {
                 FeatureFlag::PeriodicalRewardRateReduction
             },
             AptosFeatureFlag::PARTIAL_GOVERNANCE_VOTING => FeatureFlag::PartialGovernanceVoting,
+            AptosFeatureFlag::SIGNATURE_CHECKER_V2 => FeatureFlag::SignatureCheckerV2,
         }
     }
 }
--- a/third_party/move/move-bytecode-verifier/src/signature_v2.rs
+++ b/third_party/move/move-bytecode-verifier/src/signature_v2.rs
@@ -1,0 +1,1013 @@
+// Copyright (c) The Move Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+use move_binary_format::{
+    binary_views::BinaryIndexedView,
+    errors::{Location, PartialVMError, PartialVMResult, VMResult},
+    file_format::{
+        Ability, AbilitySet, Bytecode, CodeUnit, CompiledModule, CompiledScript,
+        FieldInstantiationIndex, FunctionDefinition, FunctionHandle, FunctionHandleIndex,
+        FunctionInstantiationIndex, SignatureIndex, SignatureToken, StructDefInstantiationIndex,
+        StructDefinition, StructFieldInformation, StructHandle, StructTypeParameter,
+        TypeParameterIndex,
+    },
+    IndexKind,
+};
+use move_core_types::vm_status::StatusCode;
+use std::{
+    cell::RefCell,
+    collections::{btree_map, BTreeMap},
+    fmt::Debug,
+};
+use typed_arena::Arena;
+
+const NUM_ABILITIES: usize = 4;
+const NUM_PARAMS_PER_WORD: usize = 64 / NUM_ABILITIES;
+
+#[derive(Debug)]
+struct BitsetTypeParameterConstraints<const N: usize> {
+    words: [u64; N],
+}
+
+impl<const N: usize> FromIterator<(TypeParameterIndex, AbilitySet)>
+    for BitsetTypeParameterConstraints<N>
+{
+    fn from_iter<T: IntoIterator<Item = (TypeParameterIndex, AbilitySet)>>(iter: T) -> Self {
+        let mut constraints = Self::new();
+        for (idx, abilities) in iter.into_iter() {
+            constraints.insert(idx, abilities);
+        }
+        constraints
+    }
+}
+
+impl<'a, const N: usize> From<&'a [AbilitySet]> for BitsetTypeParameterConstraints<N> {
+    fn from(abilities: &'a [AbilitySet]) -> Self {
+        abilities
+            .iter()
+            .enumerate()
+            .map(|(idx, abilities)| (idx as TypeParameterIndex, *abilities))
+            .collect()
+    }
+}
+
+/// This defines a generic collection of type parameter constraints, which is used
+/// by the signature checking algorithm.
+///
+/// The Bitset-based implementation allow for super-fast merge and subset operations.
+impl<const N: usize> BitsetTypeParameterConstraints<N> {
+    /// Creates an empty set of type parameter constraints.
+    fn new() -> Self {
+        Self { words: [0; N] }
+    }
+
+    /// Inserts an additional constraint to the set.
+    fn insert(&mut self, ty_param_idx: TypeParameterIndex, required_abilities: AbilitySet) {
+        assert!(
+            (ty_param_idx as usize) < N * NUM_PARAMS_PER_WORD,
+            "Type parameter index out of bounds. \
+             The current Bitset implementation is only configured to handle \
+             {} type parameters at max.",
+            N * NUM_PARAMS_PER_WORD
+        );
+
+        if required_abilities == AbilitySet::EMPTY {
+            return;
+        }
+
+        let ty_param_idx = ty_param_idx as usize;
+
+        let word_idx = ty_param_idx / NUM_PARAMS_PER_WORD;
+        let offset_in_word = (ty_param_idx % NUM_PARAMS_PER_WORD) * NUM_ABILITIES;
+
+        self.words[word_idx] |= (required_abilities.into_u8() as u64) << offset_in_word;
+    }
+
+    /// Merges the constraints in another set into this one.
+    fn merge(&mut self, other: &Self) {
+        for i in 0..N {
+            self.words[i] |= other.words[i]
+        }
+    }
+
+    /// Checks if all constraints are satisfied within a given context.
+    ///
+    /// The context is represented as a reference to the associated type `PreferredAbilityContext`,
+    /// which the implementations can choose based on what is optimal.
+    fn check_in_context(&self, context: &Self) -> PartialVMResult<()> {
+        for i in 0..N {
+            if self.words[i] | context.words[i] != context.words[i] {
+                return Err(PartialVMError::new(StatusCode::CONSTRAINT_NOT_SATISFIED));
+            }
+        }
+
+        Ok(())
+    }
+}
+
+/// Checks if the given type is well formed and has the required abilities.
+///
+/// This function does NOT require the context (type parameter abilities) to be specified.
+/// Instead, it will construct a minimal set of ability constraints each type parameter needs to satisfy,
+/// which the caller can then verify in individual contexts.
+/// This allows us to cache the results in the extreme case where a signature is being
+/// referenced many many times.
+///
+/// Time Complexity: `O(size_of_type * num_of_abilities * log(num_of_ty_params))`.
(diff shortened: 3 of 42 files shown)


third_party/move/move-bytecode-verifier/src/signature_v2.rs
```diff
@@ -0,0 +1,1038 @@
...
+    } else if max_num <= NUM_PARAMS_PER_WORD * 2 {
+        verify_script_impl::<BitsetTypeParameterConstraints<2>>(script)
+    } else if max_num <= NUM_PARAMS_PER_WORD * 16 {
+        verify_script_impl::<BitsetTypeParameterConstraints<16>>(script)
+    } else {
+        verify_script_impl::<BTreeTypeParameterConstraints>(script)
+    };
+
+    res.map_err(|e| e.finish(Location::Script))
+}
```
vgao1996, line 1038: @wrwg I added a dispatcher to demonstrate how we can have specialized implementations based on the actual number of type params/args in the program.

However as you have pointed out, this adds additional complexity and we may not need the full generality here. Let's discuss how we can move forward. I can see a few options:
1. Use a single implementation `BitsetTypeParameterConstraints<16>` for all programs. This would allow 256 type parameters, which is the hard limit the current binary format permits. This way we don't have to scan the program for the number of type params/args, but set operations will become slower (especially for normal legit programs)
2. Keep the dispatcher, but get rid of the BTreeMap based implementation. This will also allow us to get rid of the `TypeParameterConstraints` trait. `BitsetTypeParameterConstraints<16>` will be the biggest set we support.
3. Keep the code as is. Theoretically this could make the checker more robust, supporting an arbitrary number of type params/args. However I'm not sure if that brings us any value in practice, especially considering the additional complexity and dead code it adds.

third_party/move/move-bytecode-verifier/src/signature_v2.rs
```diff
@@ -0,0 +1,1037 @@
...
+        self.verify_signature_in_context(&ability_context, code.locals)
+            .map_err(|err| err.at_index(IndexKind::Signature, code.locals.0))?;
+
+        // Local caches to avoid re-verifying identical instantiations within the same context.
+        let mut checked_func_insts = BTreeMap::<FunctionInstantiationIndex, ()>::new();
+        let mut checked_struct_def_insts = BTreeMap::<StructDefInstantiationIndex, ()>::new();
+        let mut checked_struct_def_insts_with_key =
+            BTreeMap::<StructDefInstantiationIndex, ()>::new();
+        let mut checked_vec_insts = BTreeMap::<SignatureIndex, ()>::new();
+        let mut checked_field_insts = BTreeMap::<FieldInstantiationIndex, ()>::new();
```
zekun000, line 768: what's the difference between this function local cache compared to the struct level cache? also why we don't go through all the index and check everything compared going through bytecode and only checked idx that's used

vgao1996: > what's the difference between this function local cache compared to the struct level cache?

The check results stored in the struct level cache is context-less, while the function level one is not.

> also why we don't go through all the index and check everything compared going through bytecode and only checked idx that's used

We do go through everything and check they are well-formed in a context-less fashion. See the code in `verify_module_impl`.

third_party/move/move-bytecode-verifier/src/signature_v2.rs
vgao1996: Ended up getting rid of the TypeParameterConstraints traits and the BTreeMap impl. They won't have any practical use.

third_party/move/move-bytecode-verifier/transactional-tests/tests/signature/reference_as_type_actual_for_bytecode_instruction.exp
```diff
@@ -5,6 +5,6 @@ Error: Unable to publish module '00000000000000000000000000000000000000000000000
     major_status: INVALID_SIGNATURE_TOKEN,
     sub_status: None,
     location: 0x1::M,
-    indices: [(Signature, 2), (FunctionDefinition, 1)],
+    indices: [(FunctionInstantiation, 0)],
```
wrwg, line 8: Curious, why do we have all those baseline changes?

third_party/move/move-bytecode-verifier/transactional-tests/tests/stack_usage_verifier/signer_move_to_extra_arg.exp
```diff
@@ -11,9 +11,9 @@ Error: Unable to publish module '00000000000000000000000000000000000000000000000
 
 task 1 'publish'. lines 15-24:
 Error: Unable to publish module '0000000000000000000000000000000000000000000000000000000000000042::N'. Got VMError: {
-    major_status: POSITIVE_STACK_SIZE_AT_BLOCK_END,
+    major_status: CONSTRAINT_NOT_SATISFIED,
```
wrwg, line 14: Is this a breaking change (covered by the feature flag)?

third_party/move/move-bytecode-verifier/src/signature_v2.rs
```diff
@@ -0,0 +1,984 @@
...
+/// The Bitset-based implementation allow for super-fast merge and subset operations.
+impl<const N: usize> BitsetTypeParameterConstraints<N> {
+    /// Creates an empty set of type parameter constraints.
+    fn new() -> Self {
+        Self { words: [0; N] }
+    }
+
+    /// Inserts an additional constraint to the set.
+    fn insert(&mut self, ty_param_idx: TypeParameterIndex, required_abilities: AbilitySet) {
+        assert!((ty_param_idx as usize) < N * NUM_PARAMS_PER_WORD, "Type parameter index out of bounds. The current Bitset implementation is only configured to handle {} type parameters at max.", N * NUM_PARAMS_PER_WORD);
```
wrwg, line 66: Pl break lines correctly. Surprised auto format didn't cover.

wrwg (approved):

third_party/move/move-bytecode-verifier/transactional-tests/tests/signature/reference_as_type_actual_for_bytecode_instruction.exp
vgao1996: It's because the new and old algorithms report error locations differently.
1. The new algorithm performs context-less checks, so it may not include a function definition index in the error location.
2. The old algorithm would include the signature index of the locals if there is anything wrong in a function definition, even though it may not have anything to do with those locals. The new algorithm gets this fixed.

third_party/move/move-bytecode-verifier/transactional-tests/tests/stack_usage_verifier/signer_move_to_extra_arg.exp
vgao1996: Yes it is a breaking change. The whole pass is now feature-gated.

third_party/move/move-bytecode-verifier/src/signature_v2.rs
```diff
@@ -0,0 +1,984 @@
...
+    /// - No phantom type parameters appear in non-phantom positions.
+    ///
+    /// Time complexity: `O(total_size_of_field_types)`
+    fn verify_struct_def(&self, struct_def: &StructDefinition) -> PartialVMResult<()> {
+        let fields = match &struct_def.field_information {
+            StructFieldInformation::Native => return Ok(()),
+            StructFieldInformation::Declared(fields) => fields,
+        };
+        let struct_handle = self.resolver.struct_handle_at(struct_def.struct_handle);
+        let context: Vec<_> = struct_handle.type_param_constraints().collect();
```
zekun000, line 836: nit: I don't think we need collect here?

third_party/move/move-bytecode-verifier/src/signature_v2.rs
zekun000: I'd prefer to call context-less functions differently than the the verification with context to avoid confusions, maybe something like compute_constraints?

third_party/move/move-bytecode-verifier/src/signature_v2.rs
```diff
@@ -0,0 +1,984 @@
...
+
+    checker.verify_function_handles()?;
+    checker.verify_function_defs()?;
+    checker.verify_struct_defs()?;
+
+    Ok(())
+}
+
+fn verify_script_impl<const N: usize>(script: &CompiledScript) -> PartialVMResult<()> {
+    let constraints = Arena::<BitsetTypeParameterConstraints<N>>::new();
```
zekun000, line 890: nit: this probably should be called arena

third_party/move/move-bytecode-verifier/src/signature_v2.rs
```diff
@@ -0,0 +1,984 @@
...
+
+        self.verify_signature_in_context(&ability_context, code.locals)
+            .map_err(|err| err.at_index(IndexKind::Signature, code.locals.0))?;
+
+        // Local caches to avoid re-verifying identical instantiations within the same context.
+        let mut checked_func_insts = BTreeMap::<FunctionInstantiationIndex, ()>::new();
+        let mut checked_struct_def_insts = BTreeMap::<StructDefInstantiationIndex, ()>::new();
+        let mut checked_struct_def_insts_with_key =
+            BTreeMap::<StructDefInstantiationIndex, ()>::new();
+        let mut checked_vec_insts = BTreeMap::<SignatureIndex, ()>::new();
```
zekun000, line 706: nit: I think what we want is a BtreeSet? also I don't understand much about the index range, but if it's short, maybe a bitmask is more efficient?

zekun000:
also one thing regarding the refcell, the problem is the immutable BinaryIndexedView. I wonder if it's a &'a BinaryIndexedView the borrow checker would be fine. I only tested a small example but it looks working.

```
struct Foo<'a> {
    a: &'a u64,
    b: u64
}

impl<'a> Foo<'a> {
    fn test(a: &u64, b: &mut u64) {

    }
    fn bar(&mut self) {
        Self::test(self.a, &mut self.b);
    }
}
```

third_party/move/move-bytecode-verifier/transactional-tests/tests/stack_usage_verifier/signer_move_to_extra_arg.exp
vgao1996: Actually let me double check. This one is changing from `POSITIVE_STACK_SIZE_AT_BLOCK_END` to `CONSTRAINT_NOT_SATISFIED`, which is not expected. Either the test or the new sig checker needs to be fixed.

third_party/move/move-bytecode-verifier/src/signature_v2.rs
vgao1996: Will do

vgao1996:
@zekun000 No, `the immutable BinaryIndexedView` is not the only problem. The real problem is the borrow checker's inability to understand disjoint mutable borrows. Here's a minimal case:
```
fn compute_1(&mut self) {
    if let Entry::Vacant(entry) = self.cache1.entry(..) {
        // here compute_2 only borrows self.cache2, but the borrow checker cannot get this information from the function's signature, so it has to assume it borrows the whole struct
        let res = self.compute_2(); 
        entry.insert(.., res);
    }
}

fn compute_2(&mut self) -> Res {
    if let Entry::Vacant(entry) = self.cache2.entry(..) {
        entry.insert(.., ..);
    }
}
```

third_party/move/move-bytecode-verifier/src/signature_v2.rs
vgao1996: Good catch. I'll convert it directly into a bitset.

aptos-move/aptos-release-builder/src/components/feature_flags.rs
```diff
@@ -34,6 +34,7 @@ pub enum FeatureFlag {
     StructConstructors,
     PeriodicalRewardRateReduction,
     PartialGovernanceVoting,
+    SignatureCheckerV2,
```
runtian-zhou, line 37: Do you wish to enable this flag for 1.4? If so we probably will need to add the flag to the `release.yaml` that described the flags to enable.

third_party/move/move-bytecode-verifier/transactional-tests/tests/signature/reference_as_type_actual_for_bytecode_instruction.exp
runtian-zhou: Was wondering if we should separate out the two tests: one for the old signature checker and one for the new one. You can pass in the config using arguments from the directive. I did sth similar for the paranoid mode testing for different outputs on when flag is enabled/disabled.

vgao1996: Yeah I thought about something similar. Here are my thoughts
1. It would require some significant changes to the testing framework, in order to support two different implementations of the same bytecode verifier pass, unless you want to duplicate the existing tests
2. It would be a extra burden to maintain two sets of tests, and personally I don't see much value in doing that. The old code is not going to get updates or fixes. One should switch to the new implementation if they care about them.

aptos-move/aptos-release-builder/data/release.yaml
```diff
@@ -23,6 +23,7 @@ update_sequence:
         - bls12381_structures
         - ed25519_pubkey_validate_return_false_wrong_length
         - struct_constructors
+        - signature_checker_v2
```
runtian-zhou, line 26: This seem to come from an old version. Can you rebase?

third_party/move/move-bytecode-verifier/src/verifier.rs
```diff
@@ -189,6 +202,8 @@ impl Default for VerifierConfig {
             // max_per_mod_meter_units: Some(1000 * 8000),
             max_per_fun_meter_units: None,
             max_per_mod_meter_units: None,
+
+            use_signature_checker_v2: Some(true),
```
runtian-zhou, line 206: This should just be a boolean?

third_party/move/move-bytecode-verifier/src/signature_v2.rs
```diff
@@ -0,0 +1,997 @@
...
+/// The Bitset-based implementation allow for super-fast merge and subset operations.
+impl<const N: usize> BitsetTypeParameterConstraints<N> {
+    /// Creates an empty set of type parameter constraints.
+    fn new() -> Self {
+        Self { words: [0; N] }
+    }
+
+    /// Inserts an additional constraint to the set.
+    fn insert(&mut self, ty_param_idx: TypeParameterIndex, required_abilities: AbilitySet) {
+        assert!(
```
runtian-zhou, line 66: Can we assert this statement will never happen?

runtian-zhou (commented):
Overall seems good to me. Was wondering if you could do a test coverage report to see if all lines have been covered so far?

aptos-move/aptos-release-builder/data/release.yaml
vgao1996: Updated

sherry-x (approved):

runtian-zhou (approved):

PR #7858, opened 2023-04-19
[Storage service] Add peer IDs to error logs.

JoshLind:
### Description

This (tiny) PR adds peer network IDs to the error logs in the storage service. This will be helpful to better understand which peers are sending requests that may be invalid and/or fail.

### Test Plan
Existing test infrastructure.


sherry-x (approved):

perryjrandall (approved):
sheepit

PR #7867, opened 2023-04-20
[framework] Refactor reward rate calculation

xindingw:
### Description
**Why this change?**
- `delegation_pool::get_add_stake_fee` doesn't consider reward rate change. This could cause problem when yearly reward rate reduction.

**Behavior change**
- In `stake::on_new_epoch`, Reward rate deduction happens after reward distribution on this epoch instead of before.
- `staking_config::get_reward_rate` won't abort even after `PERIODICAL_REWARD_RATE_DECREASE` is enabled. People can always use this function to get the reward rate of the current epoch.

### Test Plan

Unit testing/ e2e testing


movekevin (approved):
Awesome!

gerben-stavenga (approved):

PR #7871, opened 2023-04-20
Initial docs site cleanup

davidiw:
* Update packages
* Remove what's new
* Fixup first bar so that we drop users off into concepts immediately
* Remove authors for now as the current approach is both fragile and not entirely accurate (moving a file or splitting a file isn't authorship) -- we plan to bring back a list of contributors for the entire doc site instead that can leverage doc folder history
* Remove community contributions -- we'll need to work with the one team that got content in and figure out next steps
* Fix up concepts
* Remove outdated ait3 doc
* Remove unnecessary wallet redirection
* Clean up Started -- focus on actual onboarding tasks
* Merge first time tutorials into a section / chapter of building
* Migrate tokens to be a chapter of building
* Remove some user facing content like wallets and funding accounts
* Merge data and interacting with blockchain
* Remove old redirects
* Remove legacy issues and workarounds (or some variant of an FAQ)
* Remove reference index
* Refactor move documentation -- merge bytecode dependencies into packages, remove my initial move guide, improve the sidebar view


gregnazario (approved):
LGTM so far

.github/workflows/lint-test.yaml
```diff
@@ -75,7 +75,7 @@ jobs:
           node-version-file: .node-version
       - uses: pnpm/action-setup@v2
         with:
-          version: 7.14.2
+          version: 8.2.0
```
banool, line 78: Updating just one project to a different pnpm version will be a pain, particularly since it has a different lock file format. Let's land this first, I update everything to this new pnpm version: https://github.com/aptos-labs/aptos-core/pull/7890.

developer-docs-site/docs/move/book/packages.md
```diff
@@ -290,3 +290,81 @@ a_move_package
 
 See the `move-package` crate for more information on these data structures and
 how to use the Move package system as a Rust library.
+
+## Using Bytecode for Dependencies
+
+Move bytecode can be used as a dependencies when the Move source code for those dependencies are not available locally. To use this feature, you will need co-locate the files in directories at the same level and then specify their paths in the corresponding `Move.toml` files.
```
banool, line 296:
```suggestion
Move bytecode can be used as dependencies when the Move source code for those dependencies are not available locally. To use this feature, you will need co-locate the files in directories at the same level and then specify their paths in the corresponding `Move.toml` files.
```

developer-docs-site/docs/move/book/packages.md
```diff
@@ -290,3 +290,81 @@ a_move_package
...
+
+We use an example to illustrate the dev flow of using this feature. Suppose we want to compile the package `A` . The package layout is:
+```rust
+./A
+├── Move.toml
+├── sources
+  ├ A.move
+```
+
+`A.move` is defined below, depending on the modules `Bar`and `Foo`:
```
banool, line 314:
```suggestion
`A.move` is defined below, depending on the modules `Bar` and `Foo`:
```

developer-docs-site/docs/move/book/packages.md
```diff
@@ -290,3 +290,81 @@ a_move_package
...
+
+```rust
+module A::A {
```
banool, line 317: As a newcomer I found this `A::A` very confusing. What if we did this instead to make it clear which is an address placeholder and which is a module.
```suggestion
module AddressA::ModuleA {
```

developer-docs-site/docs/move/book/packages.md
```diff
@@ -290,3 +290,81 @@ a_move_package
...
+    use C::Foo;
+    public fun foo(): u64 {
+        Bar::foo() + Foo::bar()
+    }
+}
+```
+
+Suppose the source of `Bar` and `Foo` are not available but the corresponding bytecode `Bar.mv` and `Foo.mv` are available locally. To use them as dependencies, we would:
+
+Specify `Move.toml` for `Bar` and `Foo` . Note that named addresses are already instantiated with the actual address in the bytecode. In our example, the actual address for `C` is already bound to `0x3`. As a result, `[addresses]` must be specified `C` as `0x3`, as shown below:
```
banool, line 328:
```suggestion
Specify `Move.toml` for `Bar` and `Foo`. Note that named addresses are already instantiated with the actual address in the bytecode. In our example, the actual address for `C` is already bound to `0x3`. As a result, `[addresses]` must be specified `C` as `0x3`, as shown below:
```

developer-docs-site/docs/move/move-on-aptos/modules-on-aptos.md
```diff
@@ -0,0 +1,14 @@
+---
+title: "Modules on Aptos"
+slug: "modules-on-aptos"
+---
+
+# Modules on Aptos
+
+Aptos allows for permissionless publishing of [Modules](../book/modules-and-scripts.md) within a [packge](../book/packages.md) as well as [upgrading](../book/package-upgrades.md) those that have appropriate compatibility set.
```
banool, line 8:
```suggestion
Aptos allows for permissionless publishing of [modules](../book/modules-and-scripts.md) within a [package](../book/packages.md) as well as [upgrading](../book/package-upgrades.md) those that have an appropriate compatibility policy set.
```

banool (changes requested):
Nothing major, mostly just some typos / punctuation fixes.

movekevin (approved):

developer-docs-site/docs/concepts/base-gas.md
```diff
@@ -1,9 +1,9 @@
 ---
-title: "How Base Gas Works"
+title: "Computing Transaction Gas"
 id: "base-gas"
 ---
 
-# How Base Gas Works
+# Computing Transaction Gas
```
movekevin, line 6: May be for later - we should point to the doc for the gas profiler here. cc @vgao1996

movekevin (commented):
High-level comments:
1. Since this changes the links anyway, I think the next step would be to clean up the doc directory organization altogether. It doesn't match the structure shown in the UI and makes it really hard to find a doc when we need to change something.
2. We're bloating up the Build list a bit by adding separate sections on different tutorials - onboarding vs Tokens. Esp as we'll have more in the future. IMO, we should just separate tutorials out into its own section (Tutorials instead of Build). This would be even better once we have the personas via Developer Hub.

saharct (approved):

davidiw:
>Since this changes the links anyway, I think the next step would be to clean up the doc directory organization altogether. It doesn't match the structure shown in the UI and makes it really hard to find a doc when we need to change something.

Yeah, I'm working on that a bit, it is a lot more consistent now. The first pass was primarily on getting some amount of organization. The second pass can commence soon.

> We're bloating up the Build list a bit by adding separate sections on different tutorials - onboarding vs Tokens. Esp as we'll have more in the future. IMO, we should just separate tutorials out into its own section (Tutorials instead of Build). This would be even better once we have the personas via Developer Hub.

Then I'll make more aggressive approach to bring content together and then we can spread it out again.

PR #7872, opened 2023-04-20
[Execution Benchmark] Check if TPS within range

danielxiangzl:
### Description
Small changes to the execution-only benchmark to check if the TPS is within a range (10%) of the hard-coded threshold.

### DIFF
--- a/testsuite/sequential_execution_performance.py
+++ b/testsuite/sequential_execution_performance.py
@@ -1,39 +1,33 @@
 #!/usr/bin/env python3
 
-# Copyright © Aptos Foundation
-# SPDX-License-Identifier: Apache-2.0
-
-import subprocess
-import re
+import subprocess, re
 
 # Set the tps threshold for block size 1k, 10k and 50k
-THRESHOLD_1k = 3700  # 4800
-THRESHOLD_10k = 4200  # 5400
-THRESHOLD_50k = 4200  # 5600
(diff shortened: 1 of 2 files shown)


perryjrandall (approved):

testsuite/sequential_execution_performance.py
```diff
@@ -31,9 +32,12 @@
...
 # Check if any threshold is not met
-if tps_1k < THRESHOLD_1k or tps_10k < THRESHOLD_10k or tps_50k < THRESHOLD_50k:
-    print("Sequential TPS below the threshold")
+if (
+    abs(tps_1k - THRESHOLD_1k) > THRESHOLD_NOISE * THRESHOLD_1k
+    or abs(tps_10k - THRESHOLD_10k) > THRESHOLD_NOISE * THRESHOLD_10k
+    or abs(tps_50k - THRESHOLD_50k) > THRESHOLD_NOISE * THRESHOLD_50k
+):
+    print(
+        f"Sequential TPS below or above the threshold by {THRESHOLD_NOISE * 100}%! Please optimize the execution performance if below threshold, otherwise please increase the hard-coded threshold if you improved the execution performance. :)"
```
igor-aptos, line 41: instead of if being in the comment, can we print a different message whether we were above or below the threshold?

igor-aptos (approved):

PR #7874, opened 2023-04-20
[Spec] Remove timeout pragma in framework specs

rahxephon89:
### Description

Remove timeout pragma in framework specs


movekevin (approved):
Nice!

junkil-park (approved):

PR #7875, opened 2023-04-20
Update docs site contributors file

banool:
### Description
I just ran this:
```
pnpm contributors
```

Since my changes to make the lookup more automatic (but keeping in mind that we don't have this in CI yet) we need to run this periodically.

### Test Plan
Netlify preview in CI.

### DIFF
--- a/developer-docs-site/src/contributors.json
+++ b/developer-docs-site/src/contributors.json
@@ -104,6 +104,11 @@
       "username": "davidiw"
     },
     {
+      "name": "Brian Olson",
+      "email": "brianolson@users.noreply.github.com",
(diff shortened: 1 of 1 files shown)


davidiw:
Per our recent conversation. Let's strip this out for now and come back with a single page of all contributors -- ideally with some other metrics such as number of PRs, first commit, latest commit -- curious what GH might be able to provide here...

https://github.com/aptos-labs/aptos-core/pull/7871

banool:
I must've not been paying attention during that bit. I think showing contributors per page is a nice feature and encourages contributions. I'll chat with you offline about this.

PR #7881, opened 2023-04-20
[Execution] Enable transaction shuffling in genesis for Forge

sitalkedia:
### Description

Enabling this by default for Forge runs and any other tests.

### DIFF
--- a/consensus/src/epoch_manager.rs
+++ b/consensus/src/epoch_manager.rs
@@ -69,8 +69,8 @@ use aptos_types::{
     epoch_change::EpochChangeProof,
     epoch_state::EpochState,
     on_chain_config::{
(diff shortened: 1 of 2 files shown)


zekun000 (commented):
we probably should change it here instead of the default value https://github.com/aptos-labs/aptos-core/blob/659720530ff2939a37f3c3c799ede6908858c72f/aptos-move/vm-genesis/src/lib.rs#L121

igor-aptos (approved):

igor-aptos:
but let's land after forge tests are cleaned up and work, to see the difference from this

davidiw (approved):

alokeutpal (approved):

PR #7882, opened 2023-04-20
[release-tooling] Support input key as string

runtian-zhou:
### Description

Add support for inputting keys as string for executing governance proposal.

### Test Plan

Manually run the command locally.


aptos-move/aptos-release-builder/src/main.rs
```diff
@@ -27,11 +29,29 @@ pub enum Commands {
...
         release_config: PathBuf,
         #[clap(short, long)]
-        test_dir: PathBuf,
-        #[clap(short, long)]
         endpoint: url::Url,
         #[clap(long)]
         framework_git_rev: Option<String>,
+        #[clap(subcommand)]
+        input_option: InputOptions,
+        #[clap(long)]
```
perryjrandall, line 37: set default to false

runtian-zhou: I think the clap will default to false but lemme add that.

aptos-move/aptos-release-builder/src/main.rs
```diff
@@ -27,11 +29,29 @@ pub enum Commands {
...
+}
+
+#[derive(Subcommand, Debug)]
+pub enum InputOptions {
+    FromDirectory {
+        #[clap(short, long)]
+        test_dir: PathBuf,
+    },
+    FromArgs {
+        #[clap(long)]
```
perryjrandall, line 49: Can you maybe add quick comment / doc for these option?

perryjrandall (approved):
WAGMI

aptos-move/aptos-release-builder/src/main.rs
```diff
@@ -51,21 +71,59 @@ async fn main() -> Result<()> {
...
+                        framework_git_rev: None,
+                        endpoint,
+                    }
+                },
+            };
 
             network_config.framework_git_rev = framework_git_rev;
 
-            aptos_release_builder::validate::validate_config(config, network_config).await
+            if mint_to_validator {
```
perryjrandall, line 118: I would even say, only allow this on local testnet, if someone try this on testnet / mainnet we should fail early and loudly

aptos-move/aptos-release-builder/src/main.rs
```diff
@@ -51,21 +71,59 @@ async fn main() -> Result<()> {
...
+                network_config.mint_to_validator().await?;
+            }
+
+            network_config.set_fast_resolve(30).await?;
```
perryjrandall, line 122: does this actually exist on devnet? (Thinking about how we're gonna execute today)

aptos-move/aptos-release-builder/src/main.rs
```diff
@@ -51,21 +71,59 @@ async fn main() -> Result<()> {
...
+            aptos_release_builder::validate::validate_config(config, network_config.clone())
+                .await?;
+            // Reset resolution time back to normal after resolution
+            network_config.set_fast_resolve(43200).await
```
perryjrandall, line 126: Can you store these constants somewhere

aptos-move/aptos-release-builder/src/main.rs
runtian-zhou: Clap will set the flag to false if the argument is not provided. Lemme add a print statement just to let us know if we minted anything.

aptos-move/aptos-release-builder/src/main.rs
runtian-zhou: This would exist.

aptos-move/aptos-release-builder/src/main.rs
runtian-zhou: Comments added!

aptos-move/aptos-release-builder/src/main.rs
runtian-zhou: Added the logic to verify with testnet/mainnet

aptos-move/aptos-release-builder/src/main.rs
runtian-zhou: Done

aptos-move/aptos-release-builder/src/main.rs
movekevin: Can we make it clear in the comment that this should only be used on a local network?

movekevin (approved):
Nice work!

PR #7888, opened 2023-04-21
Add serialization multi dimensions vector with function

### LINKED ISSUE #6724: [Feature Request] Typescript sdk support bcs serialize vector<vector<T>>
# 🚀 Feature Request

## Motivation

Now , if i defined a vector<vector<T>> args in move program. I can't use bcs native function to serialize vector<vector<T>>  to Sqp<Bytes>.

## Pitch
move :
```
    public entry fun mutil_swap<CoinA, CoinB>(
        pool_addresses: vector<vector<address>>
        amounts: vector<u64>,
    ) 
```
but how can i pass arg ues typescript sdk?

ccbond:
### Description
Typescript sdk support bcs serialize vector<vector<T>>

### Test Plan

- [x] 2D boolean array
- [x] 2D integer array
- [x] 2D string array
- [x] empty 2D array

close #6724

### DIFF
--- a/ecosystem/typescript/sdk/src/bcs/helper.ts
+++ b/ecosystem/typescript/sdk/src/bcs/helper.ts
@@ -48,7 +48,7 @@ export function deserializeVector(deserializer: Deserializer, cls: any): any[] {
 // Recursive function to serialize n-dimensional arrays
 function serializeNestedArrayWithFunc(value: any[], func: string, serializer: Serializer): void {
   // Verify value is null/undefined
-  if (value == null) {
+  if (!Array.isArray(value)) {
     throw new Error("Invalid value for serialization.")
   }
   serializer.serializeU32AsUleb128(value.length)
--- a/ecosystem/typescript/sdk/src/bcs/helper.ts
+++ b/ecosystem/typescript/sdk/src/bcs/helper.ts
@@ -47,13 +47,16 @@ export function deserializeVector(deserializer: Deserializer, cls: any): any[] {
 
 // Recursive function to serialize n-dimensional arrays
 function serializeNestedArrayWithFunc(value: any[], func: string, serializer: Serializer): void {
+  // Verify value is null/undefined
+  if (value == null) {
+    throw new Error("Invalid value for serialization.")
+  }
+  serializer.serializeU32AsUleb128(value.length)
   if (Array.isArray(value[0])) {
-    serializer.serializeU32AsUleb128(value.length)
     value.forEach((innerValue) => {
       serializeNestedArrayWithFunc(innerValue, func, serializer);
     });
   } else {
-    serializer.serializeU32AsUleb128(value.length)
     const f = (serializer as any)[func];
     value.forEach((item) => {
       f.call(serializer, item);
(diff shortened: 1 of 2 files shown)


ecosystem/typescript/sdk/src/bcs/helper.ts
```diff
@@ -45,6 +45,44 @@ export function deserializeVector(deserializer: Deserializer, cls: any): any[] {
   return list;
 }
 
+/**
+ * Serializes a 2D vector with specified item serialization function.
+ * Very dynamic function and bypasses static typechecking.
+ */
+export function serialization2DVectorWithFunc(value: any[][], func: string): Bytes {
```
jjleng, line 52: can you generalize this to make it support vectors of 1D, 2D, 3D...?

ccbond: of course.

ccbond: now use `serializeNDimensionalArrayWithFunc`

ecosystem/typescript/sdk/src/bcs/helper.ts
```diff
@@ -45,6 +45,32 @@ export function deserializeVector(deserializer: Deserializer, cls: any): any[] {
   return list;
 }
 
+// Recursive function to serialize n-dimensional arrays
+function serializeNestedArrayWithFunc(value: any[], func: string, serializer: Serializer): void {
+  if (Array.isArray(value[0])) {
```
jjleng, line 50: It is better to validate and handle the case that `value` is null/undefined as the SDK can be used in scenarios where TS type hints are unavailable.

ecosystem/typescript/sdk/src/bcs/helper.ts
```diff
@@ -45,6 +45,32 @@ export function deserializeVector(deserializer: Deserializer, cls: any): any[] {
...
+    serializer.serializeU32AsUleb128(value.length)
```
jjleng, line 51: `serializer.serializeU32AsUleb128(value.length)` is in both the `if` and the `else` branches. That means we can move it out.

ccbond: I accidentally lost the previous comment. It seems that rebase is not optional.

ecosystem/typescript/sdk/src/bcs/helper.ts
```diff
@@ -45,6 +45,35 @@ export function deserializeVector(deserializer: Deserializer, cls: any): any[] {
   return list;
 }
 
+// Recursive function to serialize n-dimensional arrays
+function serializeNestedArrayWithFunc(value: any[], func: string, serializer: Serializer): void {
+  // Verify value is null/undefined
+  if (value == null) {
```
jjleng, line 51: How about checking `if (!Array.isArray(value))`? This would cover `null`, `undefined` and invalid`value.length`.

davidiw:
Overall seems fine, but why didn't we do this in both directions? Or verify that serialization works correctly via deserialization in the tests? If we can address that, I'll get this unblocked :)

PR #7893, opened 2023-04-21
Improve fake executor to handle coin::transfer and aptos_account::transfer as well

igor-aptos:

### DIFF
--- a/execution/executor-benchmark/src/fake_executor.rs
+++ b/execution/executor-benchmark/src/fake_executor.rs
@@ -17,11 +17,17 @@ use aptos_types::{
     contract_event::ContractEvent,
     event::EventKey,
     transaction::{ExecutionStatus, Transaction, TransactionOutput, TransactionStatus},
+    vm_status::AbortLocation,
     write_set::{WriteOp, WriteSet, WriteSetMut},
 };
-use move_core_types::{language_storage::TypeTag, move_resource::MoveStructType};
+use move_core_types::{
+    identifier::Identifier,
+    language_storage::{ModuleId, TypeTag},
+    move_resource::MoveStructType,
+};
 use once_cell::sync::{Lazy, OnceCell};
 use rayon::{prelude::*, ThreadPool, ThreadPoolBuilder};
+use std::str::FromStr;
 
 pub struct FakeExecutor {}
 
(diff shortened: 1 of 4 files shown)


execution/executor-benchmark/src/fake_executor.rs
```diff
@@ -46,26 +52,13 @@ impl FakeExecutor {
...
-            vec![],
-            /*gas_used=*/ 1,
-            TransactionStatus::Keep(ExecutionStatus::Success),
-        ))
-    }
-
-    fn handle_account_creation(
+    fn handle_account_creation_and_transfer(
         sender_address: AccountAddress,
         new_account_address: AccountAddress,
```
msmouse, line 57: let's call it `receiver_address`? We don't require or assume it to be "new".

execution/executor-benchmark/src/fake_executor.rs
```diff
@@ -142,20 +106,88 @@ impl FakeExecutor {
...
+
+            let mut new_coin_store = {
+                let _timer = TIMER
+                    .with_label_values(&["read_new_coin_store"])
+                    .start_timer();
+                DbAccessUtil::get_coin_store(&new_coin_store_key, state_view)?.unwrap()
+            };
+
+            if initial_balance != 0 {
+                new_coin_store.coin += initial_balance;
```
msmouse, line 140: nit "=" instead of "+=" is cleaner, unless it's deliberate, where a comment is desired.

msmouse (approved):

igor-aptos:
@msmouse addressed your comments, with doing a bit more of a cleanup, and added support for batch_transfer

execution/executor-benchmark/src/fake_executor.rs
```diff
@@ -106,110 +122,207 @@ impl FakeExecutor {
...
+        state_view: &CachedStateView,
+        fail_on_existing: bool,
+        fail_on_missing: bool,
+    ) -> Result<TransactionOutput> {
+        let _timer = TIMER.with_label_values(&["account_creation"]).start_timer();
+
+        let mut output = {
+            let output = Self::withdraw_from_signer(sender_address, transfer_amount, state_view)?;
+            match output {
+                Ok(output) => output,
```
msmouse, line 250: use `Option::or_else()`?

zekun000 (approved):

execution/executor-benchmark/src/fake_executor.rs
igor-aptos: unfortunately, I return from the top level function in the "else", so I cannot do that easily

a little bit too nested with Result<Result<IncrementalOutput, TransactionStatus>>, so having two different Err types basically

PR #7903, opened 2023-04-23
[RFC][CLI] add disassembler to the aptos cli

0xjinn:
### Description
1. create disassembler to aptos CLI
2. The code is basically identical to `/Users/jinhou/Projects/aptos-core/third_party/move/tools/move-disassembler/src/main.rs`
3. Added couple dependencies that were missing (`move_bytecode_source_map`, `move_ir_types`)

### Test Plan

1. aptos move disassemble --help
```
➜  bytecode_modules git:(jin/add_disassembler_to_aptos_cli) ✗ aptos-debug move disassemble --help
aptos-debug-move-disassemble 1.0.11
Disassemble the Move bytecode pointed to

USAGE:
    aptos-debug move disassemble [OPTIONS] --bytecode <BYTECODE_FILE_PATH>

OPTIONS:
    -b, --bytecode <BYTECODE_FILE_PATH>
            The path to the bytecode file to disassemble; let's call it file.mv. We assume that two
            other files reside under the same directory: a source map file.mvsm (possibly) and the
            Move source code file.move

    -c, --move-coverage-path <CODE_COVERAGE_PATH>
            (Optional) Path to a coverage file for the VM in order to print trace information in the
            disassembled output

    -h, --help
            Print help information

    -s, --script
            Treat input file as a script (default is to treat file as a module)

        --skip-basic-blocks
            Do not print the basic blocks of each function

        --skip-code
            Do not print the disassembled bytecodes of each function

        --skip-locals
            Do not print locals of each function

        --skip-private
            Skip printing of private functions

    -V, --version
            Print version information
```

build the module and disassemble bytecode to check
```
➜  hello_blockchain git:(jin/add_disassembler_to_aptos_cli) ✗ aptos-debug move compile --named-addresses hello_blockchain=default
BUILDING Examples
{
  "Result": [
    "d53a65d05c8ca1160920573c54bb4ebaf06be19d6fcb540c5374a62003e65d79::message"
  ]
}
```

how the code looks like in the file
```
// Move bytecode v6
module f65cc7037576d11796a18a8b454a9afbd5d6b407523fed2b47fe1ad0f074764a.message {
use 0000000000000000000000000000000000000000000000000000000000000001::account;
use 0000000000000000000000000000000000000000000000000000000000000001::error;
use 0000000000000000000000000000000000000000000000000000000000000001::event;
use 0000000000000000000000000000000000000000000000000000000000000001::signer;
use 0000000000000000000000000000000000000000000000000000000000000001::string;


struct MessageChangeEvent has drop, store {
        from_message: String,
        to_message: String
}
struct MessageHolder has key {
        message: String,
        message_change_events: EventHandle<MessageChangeEvent>
}

public get_message(Arg0: address): String {
B0:
        0: CopyLoc[0](Arg0: address)
        1: Exists[1](MessageHolder)
        2: BrFalse(4)
```


crates/aptos/src/move_tool/disassembler.rs
```diff
@@ -0,0 +1,130 @@
...
+use move_binary_format::binary_views::BinaryIndexedView;
+use move_binary_format::CompiledModule;
+use move_binary_format::file_format::CompiledScript;
+use move_coverage::coverage_map::CoverageMap;
+use move_disassembler::disassembler::{Disassembler, DisassemblerOptions};
+use clap::Parser;
+use async_trait::async_trait;
+use move_bytecode_source_map::{mapping::SourceMapping, utils::source_map_from_file};
+use move_ir_types::location::Spanned;
+
```
rahxephon89, line 18: Could you add a comment "///Disassemble the Move bytecode pointed to" here?

crates/aptos/src/move_tool/disassembler.rs
```diff
@@ -0,0 +1,131 @@
...
+/// Disassemble the Move bytecode pointed to
+#[derive(Debug, Parser)]
+pub struct AptosDisassembler {
+    /// Skip printing of private functions.
+    #[clap(long = "skip-private")]
```
banool, line 23: I know this is just copy pasta, but it's a good chance for some clean up. With that in mind, let's just do this:
```suggestion
    #[clap(long)]
```
Same for all the other ones where the value for `long` just directly matches the arg name.

crates/aptos/src/move_tool/disassembler.rs
```diff
@@ -0,0 +1,131 @@
...
+use move_coverage::coverage_map::CoverageMap;
+use move_disassembler::disassembler::{Disassembler, DisassemblerOptions};
+use clap::Parser;
+use async_trait::async_trait;
+use move_bytecode_source_map::{mapping::SourceMapping, utils::source_map_from_file};
+use move_ir_types::location::Spanned;
+
+/// Disassemble the Move bytecode pointed to
+#[derive(Debug, Parser)]
+pub struct AptosDisassembler {
```
banool, line 21: Let's just call this `Disassemble`, keeping in line with all the other subcommands.

crates/aptos/src/move_tool/disassembler.rs
```diff
@@ -0,0 +1,131 @@
...
+    /// Do not print locals of each function.
+    #[clap(long = "skip-locals")]
+    pub skip_locals: bool,
+
+    /// Do not print the basic blocks of each function.
+    #[clap(long = "skip-basic-blocks")]
+    pub skip_basic_blocks: bool,
+
+    /// Treat input file as a script (default is to treat file as a module)
+    #[clap(short = 's', long = "script")]
```
banool, line 39: I don't mind but I believe @gregnazario doesn't like short flags, except for obvious ones like `-h` for help 😛

crates/aptos/src/move_tool/disassembler.rs
```diff
@@ -0,0 +1,131 @@
...
+    pub is_script: bool,
+
+    /// The path to the bytecode file to disassemble; let's call it file.mv. We assume that two
+    /// other files reside under the same directory: a source map file.mvsm (possibly) and the Move
+    /// source code file.move.
+    #[clap(short = 'b', long = "bytecode")]
+    pub bytecode_file_path: String,
```
banool, line 46: Use `PathBuf`, not `String`. Also can we just call this `bytecode_path` and get rid of the explicit long name.

crates/aptos/src/move_tool/disassembler.rs
```diff
@@ -0,0 +1,131 @@
...
+
+    /// (Optional) Path to a coverage file for the VM in order to print trace information in the
+    /// disassembled output.
+    #[clap(short = 'c', long = "move-coverage-path")]
+    pub code_coverage_path: Option<String>,
```
banool, line 51: Same thing, use `PathBuf`. Also remove the `long` override here too, it's just confusing when they don't match.

crates/aptos/src/move_tool/disassembler.rs
```diff
@@ -0,0 +1,131 @@
...
+#[async_trait]
+impl CliCommand<String> for AptosDisassembler {
+    fn command_name(&self) -> &'static str {
+        "AptosDisassembler"
+    }
+
+    async fn execute(self) -> CliTypedResult<String> {
+        let move_extension = MOVE_EXTENSION;
+        let mv_bytecode_extension = MOVE_COMPILED_EXTENSION;
+        let source_map_extension = SOURCE_MAP_EXTENSION;
```
banool, line 63: How come this is necessary?

crates/aptos/src/move_tool/disassembler.rs
```diff
@@ -0,0 +1,131 @@
...
+
+        let source_path = Path::new(&self.bytecode_file_path);
+        let extension = source_path
+            .extension()
+            .expect("Missing file extension for bytecode file");
```
banool, line 68: I see a lot of this in the code. Let's replace all of these with `.context()`, like this:
```suggestion
            .context("Missing file extension for bytecode file")?;
```
We should avoid panicking where possible and use Result instead. It's less of a big deal in CLIs but panics are still bad UX, the error messages are gross.

crates/aptos/src/move_tool/disassembler.rs
```diff
@@ -0,0 +1,131 @@
...
+        if extension != mv_bytecode_extension {
+            return Ok(format!(
```
banool, line 70: This shouldn't be `Ok`, this is clearly an error. Let's do `return Err()` here, or just `bail!()` (import it from anyhow: `use anyhow::bail;`.

crates/aptos/src/move_tool/disassembler.rs
```diff
@@ -0,0 +1,131 @@
...
+        let source = fs::read_to_string(&source_path).ok();
+        let source_map = source_map_from_file(
+            &Path::new(&self.bytecode_file_path).with_extension(source_map_extension),
+        );
+
+        let mut disassembler_options = DisassemblerOptions::new();
+        disassembler_options.print_code = !self.skip_code;
+        disassembler_options.only_externally_visible = self.skip_private;
+        disassembler_options.print_basic_blocks = !self.skip_basic_blocks;
+        disassembler_options.print_locals = !self.skip_locals;
```
banool, line 88: The fields of `DisassemblerOptions` are all pub, let's just build the struct normally:
```suggestion
        let disassembler_options = DisassemblerOptions {
            print_code: !self.skip_code,
            only_externally_visible: self.skip_private,
            print_basic_blocks: !self.skip_basic_blocks,
            print_locals: !self.skip_locals,
        };
```

crates/aptos/src/move_tool/disassembler.rs
```diff
@@ -0,0 +1,131 @@
...
+        let script: CompiledScript;
+        let bytecode = if self.is_script {
+            script = CompiledScript::deserialize(&bytecode_bytes)
+                .expect("Script blob can't be deserialized");
+            BinaryIndexedView::Script(&script)
+        } else {
+            module = CompiledModule::deserialize(&bytecode_bytes)
+                .expect("Module blob can't be deserialized");
+            BinaryIndexedView::Module(&module)
+        };
```
banool, line 102: Not sure why we need these two here, let's move them into each relevant block?
```suggestion
        let bytecode = if self.is_script {
            let script: CompiledScript = CompiledScript::deserialize(&bytecode_bytes)
                .expect("Script blob can't be deserialized");
            BinaryIndexedView::Script(&script)
        } else {
            let module: CompiledModule = CompiledModule::deserialize(&bytecode_bytes)
                .expect("Module blob can't be deserialized");
            BinaryIndexedView::Module(&module)
        };
```

crates/aptos/src/move_tool/disassembler.rs
```diff
@@ -0,0 +1,131 @@
...
+            disassembler.add_coverage_map(
+                CoverageMap::from_binary_file(file_path)
+                    .unwrap()
+                    .to_unified_exec_map(),
+            );
+        }
+
+        let disassemble_string = disassembler.disassemble().expect("Unable to dissassemble");
+
+        return Ok(disassemble_string);
```
banool, line 129:
```suggestion
       Ok(disassemble_string)
```

banool (dismissed):
Nice job on porting this over! Now is a good time to fix up some weird parts of this code, so I've left a bunch of comments about that.

crates/aptos/src/move_tool/disassembler.rs
wrwg: Its actually quite a deal in a CLI as well -- never crash on user input, tools doing this are experienced as crappy.

crates/aptos/src/move_tool/disassembler.rs
```diff
@@ -0,0 +1,131 @@
...
+    pub skip_basic_blocks: bool,
+
+    /// Treat input file as a script (default is to treat file as a module)
+    #[clap(short = 's', long = "script")]
+    pub is_script: bool,
+
+    /// The path to the bytecode file to disassemble; let's call it file.mv. We assume that two
+    /// other files reside under the same directory: a source map file.mvsm (possibly) and the Move
+    /// source code file.move.
+    #[clap(short = 'b', long = "bytecode")]
```
wrwg, line 45: The design as discussed on slack (have you seen this?) with @rahxephon89 was different. Instead of the user providing a path to the file (which lives somewhere at internal places), user just says `aptos move disassemble` and then the entire current package content is diassembled to console.

wrwg: > The design as discussed on slack (have you seen this?) with @rahxephon89 was different. Instead of the user providing a path to the file (which lives somewhere at internal places), user just says `aptos move disassemble` and then the entire current package content is diassembled to console.

But its also OK we stay with this design here... Its a bit more low-level but that functionality is anyway not for mainstream use...

banool: That's a good point. This is probably how it should work. This means we'd want to add the standard `--package-dir` flag to point to a different package directory.

If we go this route, we'd probably want to add another flag like `--modules` where you can choose which modules to print, in case you've got a package with heaps of modules in it. So I could use it like `--modules coin account`.

crates/aptos/src/move_tool/disassembler.rs
0xjinn: It looks like the above code will give me the error `borrowed value does not live long enough` on `&script` and `&module`

crates/aptos/src/move_tool/disassembler.rs
rahxephon89: Yeah, I think we can keep it as it is and improve it in another PR.

crates/aptos/src/move_tool/disassembler.rs
banool: Ah interesting, I guess those view types take a reference. Leave it as is then.

crates/aptos/src/move_tool/disassembler.rs
gregnazario: Yeah, I would use the constants directly, rather than copy them into a variable first.

crates/aptos/src/move_tool/disassembler.rs
0xjinn: I actually did try to clone it before passing the reference. Then it gave me `borrowed value does not live long enough`. 

Rust is hard lol

crates/aptos/src/move_tool/disassembler.rs
```diff
@@ -0,0 +1,137 @@
...
+    fn command_name(&self) -> &'static str {
+        "Disassemble"
+    }
+
+    async fn execute(self) -> CliTypedResult<String> {
+        let move_extension = MOVE_EXTENSION;
+        let mv_bytecode_extension = MOVE_COMPILED_EXTENSION;
+        let source_map_extension = SOURCE_MAP_EXTENSION;
+
+        let source_path = Path::new(&self.bytecode_path);
```
gregnazario, line 69: Nit: `self.bytecode_path.as_path()` will give you an `&Path` and then you won't have to add any more `&` before any other usages below

crates/aptos/src/move_tool/disassembler.rs
```diff
@@ -0,0 +1,137 @@
...
+                "Bad source file extension {:?}; expected {}",
+                extension, mv_bytecode_extension
+            )));
+        }
+
+        let bytecode_bytes =
+            fs::read(&self.bytecode_path).context("Unable to read bytecode file")?;
+
+        let source_path = Path::new(&self.bytecode_path).with_extension(move_extension);
+        let source = fs::read_to_string(&source_path).ok();
```
gregnazario, line 84: This will ensure the errors are handled if the source isn't a valid UTF-8 string.
```suggestion
        let source = String::from_utf8(read_from_file(old_config_file.as_path())?)?
```

crates/aptos/src/move_tool/disassembler.rs
```diff
@@ -0,0 +1,137 @@
...
+
+        if let Some(file_path) = &self.code_coverage_path {
+            disassembler.add_coverage_map(
+                CoverageMap::from_binary_file(file_path)
+                    .unwrap()
+                    .to_unified_exec_map(),
+            );
+        }
+
+        let disassemble_string = disassembler.disassemble().expect("Unable to dissassemble");
```
gregnazario, line 134: Similarly.

gregnazario (commented):
Looks great!  Just some high level thoughts:
1. Let's not duplicate constants into local variables, just use them directly, it's easy enough to read.
2. Let's attempt to remove all expects and unwraps and put appropriate CLI errors e.g. `CliError::UnexpectedError("Failed to parse source")`

crates/aptos/src/move_tool/disassembler.rs
```diff
@@ -0,0 +1,137 @@
...
+        );
+
+        let disassembler_options = DisassemblerOptions {
+            print_code: !self.skip_code,
+            only_externally_visible: self.skip_private,
+            print_basic_blocks: !self.skip_basic_blocks,
+            print_locals: !self.skip_locals,
+        };
+
+        // TODO: make source mapping work with the Move source language
```
gregnazario, line 96: just delete this comment

crates/aptos/src/move_tool/disassembler.rs
```diff
@@ -0,0 +1,137 @@
...
+        };
+
+        let mut source_mapping = {
+            if let Ok(s) = source_map {
+                SourceMapping::new(s, bytecode)
+            } else {
+                SourceMapping::new_from_view(bytecode, no_loc)
+                    .context("Unable to build dummy source mapping")?
+            }
+        };
```
gregnazario, line 117: You can pull the block out to the top level

crates/aptos/src/move_tool/disassembler.rs
0xjinn: I think this will return a String, but the original one returns an Optional String?

0xjinn:
Updated to output the result to the `output-dir` instead of print out to console

gregnazario (approved):

lightmark (approved):

PR #7904, opened 2023-04-24
[Config] Add a new config optimizer for fine-grained config tuning.

JoshLind:
Note: most of this PR is new unit tests.

### Description

This PR adds a new config optimizer component that allows fine-grained config modifications based on the node type, chain ID and compiled feature set. The idea is to support finer-grained config defaults in code, so that we can optimize for different environments and no longer need to have overrides in various places (e.g., terraform, docker, etc.).

The config optimizer respects the following order precedence for defining node configs:
1. If a config value has been set in the local config file, that value should be used (and the optimizer should not override it).
2. If a config value has not been set in the local config file, the optimizer may set the value (but, it is not required to do so).
3. Finally, if the config optimizer chooses not to set a value, the default implementation is used.

The PR offers two commits:
1. Add a new `NodeType` enum that allows us to decipher node types based on configs (e.g., validator, VFN and PFN).
2. Add the config optimizer (with unit tests!) and add a few config optimization rules to get us started:
    - Automatically set the `tokio_console_port` in the logger config if the `tokio_console` feature is enabled.
    - Enable all inspection endpoints for non-mainnet nodes (e.g., to aid debugging).
    - Default to `fast sync` for all nodes in testnet (testnet is now long enough that pruning has kicked in).
    - Double the state-sync prefetcher aggression for validator and VFN nodes (to improve syncing speeds).
    - Remove a few manual overrides in the various network config files (e.g., max connections, network ID, authentication type, etc.).

Note:
- The config optimizer runs before the sanitizer (so we can still ensure valid configs are used).
- Once this lands, we can continue to add new config optimization rules, as well as clean up/remove any existing overrides. Ideally, all config values can be defined in code once.

### Test Plan


config/src/config/config_optimizer.rs
```diff
@@ -0,0 +1,219 @@
...
+///
+/// Note: The config optimizer respects the following order precedence when
+/// determining whether or not to optimize a value:
+/// 1. If a config value has been set in the local config file, that value
+///    should be used (and the optimizer should not override it).
+/// 2. If a config value has not been set in the local config file, the
+///    optimizer may set the value (but, it is not required to do so).
+/// 3. Finally, if the config optimizer chooses not to set a value, the default
+///    value is used (as defined in the default implementation).
+pub trait ConfigOptimizer {
```
brianolson, line 29: is there a clear split between 'config optimizer' (sounds like 'makes config better') and 'config sanitizer' (sounds like 'makes config correct') ? should they just be one thing?

brianolson, line 29: is there a clear split between 'config optimizer' (sounds like 'makes config better') and 'config sanitizer' (sounds like 'makes config correct') ? should they just be one thing?

config/src/config/logger_config.rs
```diff
@@ -86,21 +91,96 @@ impl ConfigSanitizer for LoggerConfig {
...
+
+        // Set the tokio console port
+        let mut modified_config = false;
+        if local_logger_config_yaml["tokio_console_port"].is_null() {
+            // If the tokio-console feature is enabled, set the default port.
+            // Otherwise, disable the tokio console port.
+            if is_tokio_console_enabled() {
+                logger_config.tokio_console_port = Some(DEFAULT_TOKIO_CONSOLE_PORT);
+            } else {
+                logger_config.tokio_console_port = None;
```
brianolson, line 112: I think destroying someone's configuration is bad. They set the port, but maybe temporarily turned off the function, but if they come back to turn it on later the port they set will be gone.

config/src/config/network_config.rs
```diff
@@ -43,7 +43,7 @@ pub const CONNECTIVITY_CHECK_INTERVAL_MS: u64 = 5000;
 pub const MAX_CONCURRENT_NETWORK_REQS: usize = 100;
 pub const MAX_CONNECTION_DELAY_MS: u64 = 60_000; /* 1 minute */
 pub const MAX_FULLNODE_OUTBOUND_CONNECTIONS: usize = 4;
-pub const MAX_INBOUND_CONNECTIONS: usize = 30; /* At 5k TPS this could easily hit ~50MiB a second */
+pub const MAX_INBOUND_CONNECTIONS: usize = 100;
```
brianolson, line 46: changing the value seems weirdly out of scope for the rest of the PR

config/src/config/utils.rs
```diff
@@ -34,3 +35,33 @@ impl RootPath {
         }
     }
 }
+
+/// Returns true iff failpoints are enabled
+pub fn are_failpoints_enabled() -> bool {
+    cfg_if! {
```
msmouse, line 41: What's the difference between this and just calling `if cfg!(feature = "failpoints)`, and just calling it on the call site?  -- I feel it's unnecessarily verbose.

config/src/config/persistable_config.rs
```diff
@@ -49,6 +34,29 @@ pub trait PersistableConfig: Serialize + DeserializeOwned {
         Ok(())
     }
 
+    /// Read the config at the given path and return the contents as a string
+    fn read_config_file<P: AsRef<Path>>(path: P) -> Result<String, Error> {
+        // Open the config file
+        let config_path_string = path.as_ref().to_str().unwrap().to_string();
```
brianolson, line 40: https://doc.rust-lang.org/stable/std/fs/fn.read_to_string.html

config/src/config/state_sync_config.rs
```diff
@@ -15,6 +19,10 @@ const MAX_STATE_CHUNK_SIZE: u64 = 4000;
 const MAX_TRANSACTION_CHUNK_SIZE: u64 = 2000;
 const MAX_TRANSACTION_OUTPUT_CHUNK_SIZE: u64 = 1000;
 
+// The maximum number of concurrent requests to send
+const MAX_CONCURRENT_REQUESTS: u64 = 6;
```
brianolson, line 23: copy-pasteo? -> 3?

config/src/config/node_config_loader.rs
```diff
@@ -16,6 +16,45 @@ use aptos_types::{
...
+        let vfn_network_found = node_config
+            .full_node_networks
+            .iter()
+            .any(|network| network.network_id.is_vfn_network());
+        if vfn_network_found {
+            NodeType::ValidatorFullnode
+        } else {
+            NodeType::PublicFullnode
+        }
+    }
```
gregnazario, line 57: This is kinda strange there's now a Config type and an internal type

config/src/config/utils.rs
JoshLind: It's identical 😄 I've just wrapped it in the function `are_failpoints_enabled` so that it can be used across the sanitizer and optimizer (to avoid duplication). Does that make sense?

config/src/config/config_optimizer.rs
```diff
@@ -0,0 +1,219 @@
...
+        if StateSyncConfig::optimize(node_config, local_config_yaml, node_type, chain_id)? {
+            optimizers_with_modifications.push(StateSyncConfig::get_optimizer_name());
+        }
+        if optimize_validator_network_config(node_config, local_config_yaml, node_type, chain_id)? {
+            optimizers_with_modifications.push(VALIDATOR_NETWORK_OPTIMIZER_NAME.to_string());
+        }
+
+        // Log whether or not any config modifications were made
+        if optimizers_with_modifications.is_empty() {
+            println!("No config modifications were made by the config optimizer.");
```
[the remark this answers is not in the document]
JoshLind: Unfortunately, the logger hasn't been initialized at this point (so it's `println!` or nothing). If we don't want it, I can just kill the print statements. Intention was for debugging 😄

config/src/config/config_optimizer.rs
JoshLind: Yeah, they are entirely separate config passes:
- Optimizer defines fine-grained defaults
- Sanitizer ensures properties/invariants hold and will throw errors if they don't.

Technically, the sanitizer should never modify the config (only the optimizer).

JoshLind: See above.

config/src/config/logger_config.rs
JoshLind: The line above completely prevents this:
```
if local_logger_config_yaml["tokio_console_port"].is_null() {
...
}
```

If a user specifies the port in their config, we won't make any modifications. Otherwise, this logic defines which value to use. Does that make sense?

config/src/config/network_config.rs
JoshLind: Aah, it was because these values were being overwritten by various config files (e.g., in the docker and terraform directories). This removes the overrides. Technically, it could be a separate commit (but, it's still part of the goal 😄).

config/src/config/state_sync_config.rs
JoshLind: Aah, no -- I actually wanted to bump it to `6` (from `3`), to match the state value prefetcher. 😄 We used to have it overridden to `6` in our terraforms, so I'm just cleaning up as I go along 🐒

config/src/config/node_config_loader.rs
JoshLind: Yeah, the internal type is technically what we want (e.g., it makes the node configs super clear), but to maintain backward compatibility, I couldn't change the existing config type (i.e., `RoleType`). So, I've just defined the internal locally (and it's never exposed).

If we want to migrate to the external type eventually, we'll need to take a pass over all of our docs and our various template files to add a new `validator_fullnode` entry. But, I don't want to do that here (or now) 😄

config/src/config/persistable_config.rs
JoshLind: Noice! Thanks 😄

brianolson (approved):

config/src/config/config_optimizer.rs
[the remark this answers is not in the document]
JoshLind: FYI: I'm going to just remove the prints for now 😄

config/src/config/node_config_loader.rs
```diff
@@ -14,8 +15,48 @@ use aptos_types::{
     transaction::{Transaction, WriteSetPayload},
     write_set::WriteOp,
 };
+use serde_yaml::Value;
 use std::path::Path;
 
+/// A simple enum to represent the type of a node
+/// as determined from the config file.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum NodeType {
```
ibalajiarun, line 24: We will have one too many `NodeType`s. Should we converge in some way?

https://github.com/aptos-labs/aptos-core/blob/5af60240f125b2596ebd93ceb52fad29f284bbd9/crates/aptos-telemetry-service/src/types/mod.rs#L44

JoshLind: See @gregnazario's comment above. The problem with merging them is that to do it properly, it'll be a breaking change, and we'll need to update the docs and our various config template. I'm not against doing it, but we'll need to do it separately. Note that the new type is internal only to the sanitizer/optimizer for now. Does that make sense?

config/src/config/logger_config.rs
```diff
@@ -86,21 +91,96 @@ impl ConfigSanitizer for LoggerConfig {
...
+    ) -> Result<bool, Error> {
+        let logger_config = &mut node_config.logger;
+        let local_logger_config_yaml = &local_config_yaml["logger"];
+
+        // Set the tokio console port
+        let mut modified_config = false;
+        if local_logger_config_yaml["tokio_console_port"].is_null() {
+            // If the tokio-console feature is enabled, set the default port.
+            // Otherwise, disable the tokio console port.
+            if is_tokio_console_enabled() {
```
sitalkedia, line 109: What happens if the `tokio_console_port` is already specified in the config? We shouldn't be overriding them right?

sitalkedia (approved):
LGTM % minor nits

config/src/config/logger_config.rs
JoshLind: Yeah, exactly. 😄 The line `if local_logger_config_yaml["tokio_console_port"].is_null()` will ensure that the logic only runs if the `tokio_console_port` is not set in the configs.

config/src/config/state_sync_config.rs
```diff
@@ -221,12 +229,302 @@ impl Default for AptosDataClientConfig {
...
+    fn optimize(
+        node_config: &mut NodeConfig,
+        local_config_yaml: &Value,
+        _node_type: NodeType,
+        chain_id: ChainId,
+    ) -> Result<bool, Error> {
+        let state_sync_driver_config = &mut node_config.state_sync.state_sync_driver;
+        let local_driver_config_yaml = &local_config_yaml["state_sync"]["state_sync_driver"];
+
+        // Default to fast sync for all testnet nodes
```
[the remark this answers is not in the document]
JoshLind: Aah, I can update the comment. It's because testnet is long enough now that pruning has kicked in so none of the nodes can access the data anymore (unless you manually add seeds)) 😆

config/src/config/logger_config.rs
sitalkedia: Ignore my previous comment, I see that you are checking for `null` above.

PR #7911, opened 2023-04-24
[gha] Add debug line to check if concurrency is conflictingo

perryjrandall:
I suspect that performance builds on main are being canceled because this complex concurrency group name is not working as intended but it seems very difficult to actually tell without logging it

So this logs that value so that we can see if it does what is intended on PR and main

Test plan: These is not a great way to test this afaik

### DIFF
--- a/.github/workflows/docker-build-test.yaml
+++ b/.github/workflows/docker-build-test.yaml
@@ -98,6 +98,7 @@ jobs:
         run: |
           echo "GIT_SHA: ${{ env.GIT_SHA }}"
           echo "TARGET_CACHE_ID: ${{ env.TARGET_CACHE_ID }}"
+          echo "CONCURRENCY: ${{ github.workflow }}-${{ github.event_name }}-${{ (github.event_name == 'push' || github.event_name == 'workflow_dispatch') && github.sha || github.head_ref || github.ref }}"
     outputs:
       gitSha: ${{ env.GIT_SHA }}
       targetCacheId: ${{ env.TARGET_CACHE_ID }}


.github/workflows/docker-build-test.yaml
```diff
@@ -98,6 +98,7 @@ jobs:
         run: |
           echo "GIT_SHA: ${{ env.GIT_SHA }}"
           echo "TARGET_CACHE_ID: ${{ env.TARGET_CACHE_ID }}"
+          echo "CONCURRENCY: ${{ github.workflow }}-${{ github.event_name }}-${{ (github.event_name == 'push' || github.event_name == 'workflow_dispatch') && github.sha || github.head_ref || github.ref }}"
```
ibalajiarun, line 101: Looking at how GIT_SHA and TARGET_CACHE_ID is defined as env, makes me thinks that we don't have access to `github` here?

perryjrandall: @ibalajiarun those are env because the value is able to be used elsewhere, unfortunately this cannot be env because i dont believe env is accessible from the concurrency block above

See here we access github in an if (it is broadly available) usually we just prefer to maek them env to make possible to use the var in several places *I believe!

Example below
https://github.com/aptos-labs/aptos-core/blob/d5a78c9e3ef6908a9db023394890b067b05e9617/.github/workflows/docker-build-test.yaml#L120

Also, the worst case here is that we either break or print either of which are easy to fix, so can you please accept?  So I can debug this? :)

rustielin (approved):

ibalajiarun (approved):

sionescu (approved):

PR #7916, opened 2023-04-25
Add optivanity tool tip, multisig vanity example

alnoki:
@clay-aptos is there an additional place on the docs site where it may be appropriate to link to the [`optivanity`](https://github.com/econia-labs/optivanity) tool mentioned in this PR?


davidiw (approved):

davidiw:
No issue with this, though if we get competition :grin:

PR #7918, opened 2023-04-25
refactor a bit

gerben-stavenga:
### Description

1) Remove an unnecessary Arc
2) Outline functions from MoveResolverExt trait that are helper and not trait functions
3) Fix what I think is a bug in is_resource_group
4) Remove unnecessary generic type parameters
5) Prevent double wrapping of MoveResolverWithVMMetadata


aptos-move/aptos-vm/src/aptos_vm_impl.rs
```diff
@@ -43,10 +43,9 @@ use std::sync::Arc;
 
 pub const MAXIMUM_APPROVED_TRANSACTION_SIZE: u64 = 1024 * 1024;
 
-#[derive(Clone)]
-/// A wrapper to make VMRuntime standalone and thread safe.
+/// A wrapper to make VMRuntime standalone
 pub struct AptosVMImpl {
-    move_vm: Arc<MoveVmExt>,
+    move_vm: MoveVmExt,
```
zekun000, line 48: cc: @gelash I think we may want to bring the Arc back later when the new loader can be shared across threads

aptos-move/aptos-vm/src/move_vm_ext/resolver.rs
```diff
@@ -66,39 +77,18 @@ pub trait MoveResolverExt:
...
-                    .get(struct_tag.name.as_ident_str().as_str())
-                    .and_then(|attrs| {
-                        attrs
-                            .iter()
-                            .map(|attr| Some(attr.is_resource_group()))
-                            .next()
-                    })
-            })
-            .is_some()
+        (|| {
```
zekun000, line 84: this closure looks weird to me, why do we need it?

aptos-move/aptos-vm/src/move_vm_ext/resolver.rs
```diff
@@ -66,39 +77,18 @@ pub trait MoveResolverExt:
...
-        let metadata = self.get_module_metadata(struct_tag.module_id());
-        metadata
-            .and_then(|metadata| {
-                metadata
-                    .struct_attributes
-                    .get(struct_tag.name.as_ident_str().as_str())
-                    .and_then(|attrs| {
-                        attrs
-                            .iter()
-                            .map(|attr| Some(attr.is_resource_group()))
```
zekun000, line 98: 🤦 I hope resource group is the only allowed attributes on structs

aptos-move/aptos-vm/src/move_vm_ext/resolver.rs
gerben-stavenga: To use ? chaining, instead of the IMO awkward and_then(closure).

aptos-move/aptos-vm/src/move_vm_ext/resolver.rs
gerben-stavenga: I don't know precisely what should happen here. I know that the map above can be removed without semantic changes, as we only check if the iter is empty or not.

The only thing I could imagine is that what meant was to find an is_resource_group attribute. Wdyt?

zekun000: I think the fix is correct but cc: @davidiw

aptos-move/aptos-vm/src/move_vm_ext/resolver.rs
zekun000: i see, and_then is probably the idiomatic way of doing this but I agree it looks ugly

zekun000 (approved):

aptos-move/aptos-vm/src/move_vm_ext/resolver.rs
davidiw: both are pretty uncomfortable... the new way superficially reads cleaner but demonstrates a much higher degree of mastery of the language than the former.

Given that both Zekun and I are unaware of this power, it likely means we're writing code that few others would understand. I tend to err toward writing code that can be commonly understood.

On the flip side, the new method is a lot tighter and maybe we just haven't had any new blood in rust for a long time hence our skills have stagnated.

davidiw (commented):
I'll withhold my approval -- maybe we can do a quick flag in #rust on slack to get a round of feedback on the intuitiveness of this style before moving forward?

aptos-move/aptos-vm/src/move_vm_ext/resolver.rs
gerben-stavenga: @davidiw can you comment on this. As in the current version of the code the map on this line has no semantic effect, which smells like a bug.

davidiw: :grimacing: 

Gah... so what we probably want to do is what you did below... have a find statement...

This only works because we only have a single struct attribute at this point in time...

Alternatively we need a find_map :facepalm:

davidiw (approved):
I don't think we need to hot fix this, since we hardcode allowed attributes in the harness...

aptos-move/aptos-vm/src/move_vm_ext/resolver.rs
gerben-stavenga: yes "find" seemed most natural to me, but wasn't sure.

aptos-move/aptos-vm/src/move_vm_ext/resolver.rs
JoshLind: +1 to chaining (e.g., `and_then`). Even though `and_then` is more verbose, it took me 10x longer to figure out what was going on here 😵‍💫 

> it likely means we're writing code that few others would understand.

Yeah, I like simple code so that simple people (like me) can still have a job 😄

gerben-stavenga: I added aptos_try as a generic utility, in lieu of try language feature in unstable.

I think it's a useful pattern as I've seen this more in the code

aptos-utils/src/lib.rs
```diff
@@ -0,0 +1,11 @@
+// Copyright © Aptos Foundation
+
+/// An internal implementation to imitate the feature of `try` in unstable Rust.
+/// Useful to use '?' chaining on option/result without the need to wrap the expression in a
+/// function.
+#[macro_export]
```
geekflyer, line 6: add a comment stating that this macro is obsolete once:
https://github.com/rust-lang/rust/issues/31436
is resolved/try is stable.

gerben-stavenga: Done

PR #7919, opened 2023-04-25
[docs] Update delegated pool min to 11APT

gregnazario:
### Description
Updated accordingly


movekevin (approved):

jillxuu (approved):

movekevin (changes requested):
Actually its technically 10APT in the contracts (plus a small add fee). Its only 11 apt in the explorer ui

jillxuu:
> Actually its technically 10APT in the contracts (plus a small add fee). Its only 11 apt in the explorer ui

yes exactly. but do we wanna align all kinds of comms to be 11 or we're ok with technical docs saying 10 but UI and other faqs saying 11?

PR #7920, opened 2023-04-25
[Execution][Sharding] Implement InMemoryStateCalculatorV2.

grao1991:
### Description
As the first step of making in memory data structure sharded, this PR is trying to extract a subset of InMemoryStateCalculator logic to support calculation after **block** execution.

Comparing to InMemoryStateCalculator, the main difference is:

- It assumes there is one and only one checkpoint, at the end of the block. Therefore it doesn't need to maintain the delta between the checkpoint and latest.
- It removes the add_transaction, which applies the writeset to the base state cache version by version. Instead, the usage calculation and epoch change handling was separated out, and the write set to state change calculation is done in parallel.

This PR doesn't change the caller side to use the new implementation, thus no production impact.
### Test Plan
Verified in executor benchmark.


zekun000 (approved):

execution/executor/src/components/in_memory_state_calculator_v2.rs
```diff
@@ -0,0 +1,253 @@
...
+        } = base.clone();
+
+        assert!(base_version == current_version);
+        assert!(updates_since_base.is_empty());
+
+        let sharded_state_cache = arr![DashMap::new(); 16];
+
+        // TODO(grao): Move this logic to an earlier stage.
+        let _: Vec<_> = state_cache
+            .into_iter()
```
msmouse, line 104: can't `par_iter()`?

execution/executor/src/components/in_memory_state_calculator_v2.rs
```diff
@@ -0,0 +1,253 @@
...
+
+impl InMemoryStateCalculatorV2 {
+    pub fn new(base: &StateDelta, state_cache: StateCache) -> Self {
+        let StateCache {
+            frozen_base,
+            state_cache,
+            proofs,
+        } = state_cache;
+        // TODO(grao): Only a base tree and version are needed because the current should be the
+        // same as base.
```
msmouse, line 88: I guess I've lost the context.. so why?

execution/executor/src/components/in_memory_state_calculator_v2.rs
```diff
@@ -0,0 +1,253 @@
...
+            _frozen_base: frozen_base,
+            proof_reader: ProofReader::new(proofs),
+
+            sharded_state_cache,
+            next_version: current_version.map_or(0, |v| v + 1),
+            usage: current.usage(),
+
+            checkpoint: base,
+            checkpoint_version: base_version,
+            latest: current.freeze(),
```
msmouse, line 119: you are not gonna make multiple checkpoints (or even update `next_version` or anything incrementally), once these are generated, it's already time to return the result, so these fields shouldn't exist at all?

execution/executor/src/components/in_memory_state_calculator_v2.rs
```diff
@@ -0,0 +1,253 @@
...
+        Vec<Option<HashValue>>,
+        StateDelta,
+        Option<EpochState>,
+    )> {
+        let num_txns = to_keep.len();
+
+        let state_updates_vec = Self::get_state_updates(to_keep);
+
+        let mut state_checkpoint_hashes = Vec::with_capacity(num_txns);
+        for (txn, txn_output) in to_keep.iter().take(num_txns - 1) {
```
msmouse, line 138: potential overflow

execution/executor/src/components/in_memory_state_calculator_v2.rs
```diff
@@ -0,0 +1,253 @@
...
+            assert!(!Self::need_checkpoint(txn, txn_output));
+            state_checkpoint_hashes.push(None);
```
msmouse, line 140: A bit ugly. would leave it along and after you have the snapshot, do `repeat(None, num_txn - 1 ).chain(root_hash).collect()` in one line.

execution/executor/src/components/in_memory_state_calculator_v2.rs
grao1991: Because the assert! at line 97 and 98 will always hold, or am I missing anything?

execution/executor/src/components/in_memory_state_calculator_v2.rs
grao1991: Doesn't support.

execution/executor/src/components/in_memory_state_calculator_v2.rs
grao1991: True, just too lazy to change the parameter and return value to pass them around among functions, use `self` is easy :)

execution/executor/src/components/in_memory_state_calculator_v2.rs
grao1991: Done

execution/executor/src/components/in_memory_state_calculator_v2.rs
grao1991: Do we want to support empty block?

execution/executor/src/components/in_memory_state_calculator_v2.rs
msmouse: well, that'll make it significantly less confusing, which is the purpose of this, so let's be diligent here :p

execution/executor/src/components/in_memory_state_calculator_v2.rs
msmouse: if not, `assert!(!to_keep.is_empty())`

execution/executor/src/components/in_memory_state_calculator_v2.rs
```diff
@@ -0,0 +1,254 @@
...
+        Vec<Option<HashValue>>,
+        StateDelta,
+        Option<EpochState>,
+    )> {
+        let num_txns = to_keep.len();
+
+        let state_updates_vec = Self::get_state_updates(to_keep);
+
+        for (txn, txn_output) in to_keep.iter().take(num_txns - 1) {
+            assert!(!Self::need_checkpoint(txn, txn_output));
```
areshand, line 138: how do we guarantee txn requiring checkpoints are not passed down to here?

execution/executor/src/components/in_memory_state_calculator_v2.rs
```diff
@@ -0,0 +1,254 @@
...
+        }
+
+        let next_epoch_state = if new_epoch {
+            Some(Self::get_epoch_state(&self.sharded_state_cache, &updates)?)
+        } else {
+            None
+        };
+
+        let (txn, txn_output) = &to_keep[num_txns - 1];
+        assert!(Self::need_checkpoint(txn, txn_output));
```
areshand, line 166: can we combine this assert with the one above?

execution/executor/src/components/in_memory_state_calculator_v2.rs
grao1991: Done

execution/executor/src/components/in_memory_state_calculator_v2.rs
grao1991: This is our assumption for block.

execution/executor/src/components/in_memory_state_calculator_v2.rs
grao1991: Done

execution/executor/src/components/in_memory_state_calculator_v2.rs
```diff
@@ -0,0 +1,254 @@
...
+    },
+    transaction::{Transaction, Version},
+    write_set::TransactionWrite,
+};
+use arr_macro::arr;
+use dashmap::DashMap;
+use rayon::prelude::*;
+use std::collections::HashMap;
+
+struct CoreAccountStateView<'a> {
```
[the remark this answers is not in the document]
grao1991: Emmm... I feel the word "code" in the constant name is not precise, so I would rather not use the term here.

execution/executor/src/components/in_memory_state_calculator_v2.rs
```diff
@@ -0,0 +1,254 @@
...
+    base: &'a [DashMap<StateKey, Option<StateValue>>; 16],
```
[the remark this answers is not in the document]
grao1991: Done

areshand (approved):

execution/executor/src/components/in_memory_state_calculator_v2.rs
```diff
@@ -0,0 +1,229 @@
...
+            ensure!(
+                Self::need_checkpoint(txn, txn_output) ^ (i != num_txns - 1),
+                "Checkpoint is allowed iff it's the last txn in the block. index: {i}, is_last: {}, txn: {txn:?}, is_reconfig: {}",
+                i == num_txns - 1,
+                txn_output.is_reconfig()
+            );
+        }
+
+        let StateCache {
+            ///// These don't change during the calculation.
```
msmouse, line 99: this line is irrelevant (used to describe two fields below it, in the declaration)

execution/executor/src/components/in_memory_state_calculator_v2.rs
```diff
@@ -0,0 +1,229 @@
...
+            // This makes sure all in-mem nodes seen while proofs were fetched stays in mem during the
+            // calculation
+            frozen_base: _,
```
msmouse, line 102: Will it actually host the refs if there's no name for it?

execution/executor/src/components/in_memory_state_calculator_v2.rs
grao1991: Removed.

execution/executor/src/components/in_memory_state_calculator_v2.rs
grao1991: Yes

msmouse (approved):

PR #7927, opened 2023-04-26
move test filtering for aptos-framework

zjma:
A way to filter move unit tests and make aptos-framework development slightly easier.

### DIFF
--- a/aptos-move/framework/README.md
+++ b/aptos-move/framework/README.md
@@ -47,3 +47,14 @@ The overall structure of the Aptos Framework is as follows:
 ├── releases                                    # Move release bundles
 └── tests
 ```
+
+## Move unit test selection
+In this package, unit test run triggered by `cargo test <filter>` will only select rust test wrappers
+(for example, those marked as `#[test]` in `tests/move_unit_test.rs`).
+
+To select the actual test cases in Move, use the following command.
+```bash
+MOVE_TEST_FILTER=<move-test-filter> cargo test <the-rest-params>
+```
+A test case is selected if and only if its full name (e.g. `0x1::ristretto255::test_point_neg`)
+contains the value of `MOVE_TEST_FILTER` as a substring.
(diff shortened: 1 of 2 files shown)


zjma:
> That's amazing. Is there a way to do this via a `cargo test` flag instead of an environment variable?

envvar is the best I can come up with so far...

aptos-move/framework/README.md
```diff
@@ -47,3 +47,14 @@ The overall structure of the Aptos Framework is as follows:
...
 ```
+
+## Move unit test selection
+In this package, unit test run triggered by `cargo test <filter>` will only select rust test wrappers
+(for example, those marked as `#[test]` in `tests/move_unit_test.rs`).
+
+To select the actual test cases in Move, use the following command.
+```bash
+MOVE_TEST_FILTER=<move-test-filter> cargo test <the-rest-params>
+```
```
gregnazario, line 58: We should stay away from using environment variables to control configuration, and instead add a flag `--filter` to filter tests

movekevin: +1. I appreciate the attempt here but this is what aptos move test -f filter is designed for

movekevin: I see. Can’t you do cargo run -p aptos — move test …?

alinush: That will take a really long time to recompile the CLI...

davidiw: use path/to/aptos-core/target/debug/aptos move test

alinush: How does that address the long time to recompile after modifying a Rust native?

davidiw: Ahhh... fwiw, you can filter by the framework package.

I do like this direction, just not entirely sure how easy it is to get the flags from test and route them appropriately :grimacing:

PR #7929, opened 2023-04-26
[cli] Add more involved move init command

gregnazario:
### Description
I was fixing some tests, and I realized that this kept pulling from `main`.  I've added it so it now pulls from `testnet` and made a little demo program so it can give some more info there.

Move init now provides a small program that's an example of what you can do along with how to test.  Additionally, shows best practices of naming and error messages

### Test Plan
```
$ ./target/debug/aptos move init --name test_all --package-dir test-all --assume-yes
{
  "Result": "Success"
}

$ aptos move test --package-dir test-all --named-addresses test_all=0x22            
BUILDING test_all
Running Move unit tests
[ PASS    ] 0x22::test_all::test_authorization
[ PASS    ] 0x22::test_all::test_implementation
Test result: OK. Total tests: 2; passed: 2; failed: 0
{
  "Result": "Success"
}

$ aptos move publish --package-dir test-all --named-addresses test_all=devnet --profile devnet
UPDATING GIT DEPENDENCY https://github.com/aptos-labs/aptos-core.git
BUILDING test_all
package size 1378 bytes
Do you want to submit a transaction for a range of [97700 - 146500] Octas at a gas unit price of 100 Octas? [yes/no] >
yes
{
  "Result": {
    "transaction_hash": "0x7abd71cf2abc515e2b1d0d0ecda26e199d75d3a7aba149aa950efcb403cc0f8f",
    "gas_used": 977,
    "gas_unit_price": 100,
    "sender": "b11affd5c514bb969e988710ef57813d9556cc1e3fe6dc9aa6a82b56aee53d98",
    "sequence_number": 3,
    "success": true,
    "timestamp_us": 1682477528360497,
    "version": 193629,
  }
}
```

### DIFF
--- a/crates/aptos/src/move_tool/manifest.rs
+++ b/crates/aptos/src/move_tool/manifest.rs
@@ -25,6 +25,12 @@ pub struct ManifestNamedAddress {
     pub address: Option<AccountAddress>,
 }
 
+impl ManifestNamedAddress {
+    pub fn placeholder() -> Self {
+        ManifestNamedAddress { address: None }
+    }
+}
+
 impl From<Option<AccountAddress>> for ManifestNamedAddress {
     fn from(opt: Option<AccountAddress>) -> Self {
         ManifestNamedAddress { address: opt }
--- a/crates/aptos/src/move_tool/mod.rs
+++ b/crates/aptos/src/move_tool/mod.rs
@@ -59,6 +59,7 @@ use move_core_types::{identifier::Identifier, language_storage::ModuleId, u256::
 use move_package::{source_package::layout::SourcePackageLayout, BuildConfig};
 use move_unit_test::UnitTestingConfig;
 pub use package_hooks::*;
+use regex::Regex;
 use serde::{Deserialize, Serialize};
 use serde_json::json;
 use std::{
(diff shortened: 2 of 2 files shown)


0xjinn (approved):

banool (commented):
Nit, I notice the output is like this:
```
[package]
name = 'my_test'
version = '1.0.0'

[addresses]
my_test = '_'
[dependencies.AptosFramework]
git = 'https://github.com/aptos-labs/aptos-core.git'
rev = 'testnet'
subdir = 'aptos-move/framework/aptos-framework'
```

Can we get a newline after the addresses section?

Another thing, in my opinion using the package name for the named address is confusing. I know I found this quite confusing when I was first learning. Using different values helps delineate. Could we just use `addr`, or like `<package_name>_addr` for the named address?

We don't have to do it here but a cool improvement would be to add more kinds of things to the sample module, like different visibility, view functions, etc. At the least we should try to have a couple of structs, one that is key and one that is store.

gregnazario:
> Nit, I notice the output is like this:
> 
> ```
> [package]
> name = 'my_test'
> version = '1.0.0'
> 
> [addresses]
> my_test = '_'
> [dependencies.AptosFramework]
> git = 'https://github.com/aptos-labs/aptos-core.git'
> rev = 'testnet'
> subdir = 'aptos-move/framework/aptos-framework'
> ```
> 
> Can we get a newline after the addresses section?
> 
> Another thing, in my opinion using the package name for the named address is confusing. I know I found this quite confusing when I was first learning. Using different values helps delineate. Could we just use `addr`, or like `<package_name>_addr` for the named address?
> 
> We don't have to do it here but a cool improvement would be to add more kinds of things to the sample module, like different visibility, view functions, etc. At the least we should try to have a couple of structs, one that is key and one that is store.

The spacing is how the Rust crate does it in a programatic way.  Unsure if there's a way to change that.

We can't use just `addr` because you're likely to get address name conflicts when you inherit crates, so unfortunately we need them named something.  I purposely named it this poor name so people would rename them.

I actually had suggested earlier to add more things to the sample module, though I got some pushback and I agree that adding too much here may just add something that degrades over time

PR #7934, opened 2023-04-26
Add icon_uri to fungible asset and update comments

movekevin:
### Description

This adds a useful field - icon uri to the fungible asset standard. 
1. icon_uri: This is different from the uri in fungible tokens. A fungible asset might not want to be listed as a token but still wants an icon. And token usually has full image not icon.

### Test Plan
Unit tests


lightmark (approved):

junkil-park:
Is this something corresponding to the `icons` in `ERC20TokenMetadata`?

junkil-park (approved):

PR #7938, opened 2023-04-26
Add gzip compression option for indexer grpc

rtso:
### Description
With Typescript indexer grpc client, we observed low throughput in receiving protobuf messages. This change adds the option to send a compressed request and receive compressed responses from the grpc, which hopefully will improve the throughput. 

### Test Plan
1. Run local indexer grpc data service with changes
2. Create Typescript client that's connected to local indexer grpc data service with gzip compression set and confirm the client is able to receive grpc messages. 

Note: I didn't observe any improvement in throughput with compression when testing against my local data service instance, which I suspect is because local instance naturally allows higher throughput. Once this change is deployed, I'd like to test whether there will be an improvement in throughput connecting to deployed instance.


ecosystem/indexer-grpc/indexer-grpc-data-service/src/main.rs
```diff
@@ -78,13 +80,15 @@ fn main() {
     // Add authentication interceptor.
     runtime.spawn(async move {
         let server = RawDataServer::new(config);
-        let svc = aptos_protos::indexer::v1::raw_data_server::RawDataServer::with_interceptor(
+        let svc = ptos_protos::indexer::v1::raw_data_server::RawDataServer::with_interceptor(
```
larry-aptos, line 83: nit: typo

rtso: Fixed

larry-aptos (approved):

larry-aptos (approved):
Nice!

ying-w (approved):

PR #7939, opened 2023-04-26
forge test local-swarm improvements

brianolson:
### Description

write out root key hex to file at swarm_dir/root_key (not just terminal)
and to bcs as `root_key.bin` , and that feeds into other tools like faucet and transaction-emitter.

add `--swarmdir` to forge local clusters, e.g.
```bash
forge test local-swarm --swarmdir /path/to/swarm_dir
```

local cluster build without indexer if we don't need it.
```bash
FORGE_BUILD_WITHOUT_INDEXER=1 forge test local-swarm
```

faucet run-simple `--key-file-path` can decode either bcs or hex.

### Test Plan

running a local cluster with faucet:

```bash
export FORGE_BUILD_WITHOUT_INDEXER=1
target/debug/forge --suite run_forever --num-validators 3 test local-swarm --swarmdir /tmp/swarmdir
target/debug/aptos-faucet-service run-simple --key-file-path /tmp/swarmdir/root_key.bin --node-url "http://"$(python3 testsuite/node_info.py api /tmp/swarmdir/0) --chain-id 4
target/debug/aptos-transaction-emitter emit-tx --mint-file /tmp/swarmdir/root_key.bin --target-tps 100 --targets "http://"$(python3 testsuite/node_info.py api /tmp/swarmdir/0)
```


crates/aptos-faucet/core/src/funder/common.rs
```diff
@@ -96,20 +96,25 @@ impl ApiConnectionConfig {
...
-            .with_context(|| {
+            let key_bytes = std::fs::read(self.key_file_path.as_path()).with_context(|| {
                 format!(
-                    "Failed to deserialize data in key file: {}",
+                    "Failed to read key file: {}",
                     self.key_file_path.to_string_lossy()
                 )
-            })
+            })?;
+            // decode as bcs first, fall back to a file of hex
```
brianolson, line 105: I wrote this code to try both kinds of decoding before I found the code to export the bcs format. This is kinda still useful, should maybe kinda be lifted to common code for reading a key, or should maybe be rolled back as a complication we don't really need.

testsuite/forge/src/backend/local/swarm.rs
```diff
@@ -215,11 +219,22 @@ impl LocalSwarm {
             .collect::<Result<HashMap<_, _>>>()?;
 
         // We print out the root key to make it easy for users to deploy a local faucet
-        let encoded_root_key = hex::encode(root_key.to_bytes());
+        let encoded_root_key = hex::encode_upper(root_key.to_bytes());
```
brianolson, line 222: OR, maybe this should be `EncodingType::Hex.encode_key()`

grao1991 (approved):

sitalkedia (approved):

testsuite/forge/src/backend/local/cargo.rs
```diff
@@ -160,7 +160,12 @@ pub fn git_merge_base<R: AsRef<str>>(rev: R) -> Result<String> {
 pub fn cargo_build_common_args() -> Vec<&'static str> {
     let use_release = use_release();
     let consensus_only = build_consensus_only_node();
-    let mut args = vec!["build", "--features=failpoints,indexer"];
+    let without_indexer = std::env::var("FORGE_BUILD_WITHOUT_INDEXER").is_ok();
```
JoshLind, line 163: nit: can you wrap `std::env::var("FORGE_BUILD_WITHOUT_INDEXER").is_ok()` into a function so it's consistent with `use_release` and `consensus_only`?

testsuite/forge/src/backend/local/mod.rs
```diff
@@ -76,7 +78,7 @@ impl LocalFactory {
             })?;
 
         versions.insert(new_version.version.clone(), new_version);
-        Ok(Self::new(versions))
+        Ok(Self::new(versions, ""))
```
JoshLind, line 81: nit: you could make `swarm_dir` optional if it's not always used?

testsuite/forge/src/backend/local/mod.rs
```diff
@@ -126,6 +128,11 @@ impl LocalFactory {
     where
         R: ::rand::RngCore + ::rand::CryptoRng,
     {
+        let swarmdir = if self.swarm_dir.is_empty() {
+            None
+        } else {
+            Some(PathBuf::from(self.swarm_dir.as_str()))
```
JoshLind, line 134: nit: could you use `map` to avoid this? https://doc.rust-lang.org/std/option/enum.Option.html#method.map

testsuite/forge/src/backend/local/swarm.rs
```diff
@@ -215,11 +219,22 @@ impl LocalSwarm {
...
 
         // We print out the root key to make it easy for users to deploy a local faucet
-        let encoded_root_key = hex::encode(root_key.to_bytes());
+        let encoded_root_key = EncodingType::Hex.encode_key("root_key", &root_key)?;
         info!(
             "The root (or mint) key for the swarm is: 0x{}",
-            encoded_root_key
+            String::from_utf8_lossy(encoded_root_key.as_slice())
         );
+        let root_key_path = dir_actual.as_ref().join("root_key");
```
JoshLind, line 227: nit: could we move these into a helper function? e.g., `write_root_key_to_file()`?

JoshLind (approved):
Thanks @brianolson! Just nits 😄

testsuite/forge/src/backend/local/mod.rs
brianolson: Yeah, it's a little weird, it goes from Option<String> to "" to Option<String> again. And yet I feel like there's a reason why "" makes more sense in the middle data structure.

brianolson: changed to move `Option<String>` around through all the levels and it's better

testsuite/forge/src/backend/local/cargo.rs
brianolson: rearranged, and added comments about how weird it is that compile options for forge's inner `aptos-node` are locked in at compile time for `forge`, except the new option I'm adding which sets compile options for aptos-node when we're building aptos-node.

crates/aptos-faucet/core/src/funder/common.rs
```diff
@@ -96,20 +96,25 @@ impl ApiConnectionConfig {
...
+            let result: Result<Ed25519PrivateKey> =
+                aptos_sdk::bcs::from_bytes(&key_bytes).with_context(|| "bad bcs");
+            match result {
```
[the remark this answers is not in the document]
brianolson: didn't exactly do that, but I think I arrived at an arrangement that's tighter

testsuite/forge/src/backend/local/cargo.rs
```diff
@@ -162,19 +170,20 @@ pub fn git_merge_base<R: AsRef<str>>(rev: R) -> Result<String> {
...
-    if consensus_only {
+    let mut args = if build_aptos_node_without_indexer() {
+        vec!["build", "--features=failpoints"]
+    } else {
+        vec!["build", "--features=failpoints,indexer"]
+    };
+    if build_consensus_only_node() {
         args.push("--features=consensus-only-perf-test");
     }
-    if network_perf_test {
```
brianolson, line 172: @JoshLind I rearranged a thing you just did in merge cleanup, wanted to make sure you still think this looks okay. Note the comments on each build option at the top of this file.

PR #7942, opened 2023-04-26
change the aptos grpc server implementation to `AptosRawDataServer` t…

larry-aptos:
…o avoid confusion


rtso:
Maybe rename it to `RawDataServerWrapper`? Still unclear on the naming of `AptosRawDataServer`.

PR #7943, opened 2023-04-26
Config tuning for local testnet

gregnazario:
### Description
Add documentation to the config, and tune the CLI local testnet config

### Test Plan
I was successful at pulling CPU down to 67-72% from 126% at idle.

### DIFF
--- a/aptos-node/src/lib.rs
+++ b/aptos-node/src/lib.rs
@@ -312,16 +312,47 @@ fn create_single_node_test_config(
 
     // Adjust some fields in the default template to lower the overhead of
     // running on a local machine.
+    node_config
+        .consensus
+        .quorum_store
+        .num_workers_for_remote_batches = 1;
+    node_config.consensus.quorum_store_poll_time_ms = 1000;
+
     node_config.execution.concurrency_level = 1;
     node_config.execution.num_proof_reading_threads = 1;
+    node_config.execution.paranoid_hot_potato_verification = false;
+    node_config.execution.paranoid_type_verification = false;
+    node_config
+        .execution
+        .processed_transactions_detailed_counters = false;
+
     node_config.peer_monitoring_service.max_concurrent_requests = 1;
+    node_config
+        .peer_monitoring_service
+        .enable_peer_monitoring_client = false;
+
     node_config
         .mempool
         .shared_mempool_max_concurrent_inbound_syncs = 1;
+    node_config.mempool.default_failovers = 1;
+    node_config.mempool.max_broadcasts_per_peer = 1;
+
     node_config
         .state_sync
         .state_sync_driver
         .enable_auto_bootstrapping = true;
+    node_config
+        .state_sync
+        .state_sync_driver
+        .max_connection_deadline_secs = 1;
+    node_config
+        .state_sync
+        .state_sync_driver
+        .progress_check_interval_ms = 10_000;
+    node_config
+        .state_sync
+        .data_streaming_service
+        .progress_check_interval_ms = 10_000;
 
     // Configure the validator network
     let validator_network = node_config.validator_network.as_mut().unwrap();
--- a/config/src/config/api_config.rs
+++ b/config/src/config/api_config.rs
@@ -13,43 +13,60 @@ use std::net::SocketAddr;
 #[derive(Clone, Debug, Deserialize, PartialEq, Eq, Serialize)]
 #[serde(default, deny_unknown_fields)]
 pub struct ApiConfig {
+    /// Enables the REST API endpoint
     #[serde(default = "default_enabled")]
     pub enabled: bool,
+    /// Address for the REST API to listen on.  Set to 0.0.0.0:<port> to allow all inbound connections.
     pub address: SocketAddr,
+    /// Path to a local TLS certificate to enable HTTPS
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub tls_cert_path: Option<String>,
+    /// Path to a local TLS key to enable HTTPS
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub tls_key_path: Option<String>,
-    // optional for compatible with old configuration
+    /// A maximum limit to the body of a POST request in bytes
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub content_length_limit: Option<u64>,
+    /// Enables failpoints for error testing
     #[serde(default = "default_disabled")]
     pub failpoints_enabled: bool,
+    /// Enables JSON output of APIs that support it
     #[serde(default = "default_enabled")]
     pub json_output_enabled: bool,
+    /// Enables BCS output of APIs that support it
     #[serde(default = "default_enabled")]
     pub bcs_output_enabled: bool,
+    /// Enables encode submission API
     #[serde(default = "default_enabled")]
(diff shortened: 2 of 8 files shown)


aptos-node/src/lib.rs
```diff
@@ -312,16 +312,40 @@ fn create_single_node_test_config(
...
         .state_sync_driver
         .enable_auto_bootstrapping = true;
+    node_config.state_sync.state_sync_driver.max_consecutive_stream_notifications = 1;
+    node_config.state_sync.state_sync_driver.progress_check_interval_ms = 1000;
+    node_config.state_sync.aptos_data_client.max_num_in_flight_priority_polls = 1;
+    node_config.state_sync.aptos_data_client.max_num_in_flight_regular_polls = 1;
+    node_config.state_sync.data_streaming_service.progress_check_interval_ms = 1000;
+    node_config.state_sync.data_streaming_service.max_concurrent_requests = 1;
+    node_config.state_sync.data_streaming_service.max_concurrent_state_requests = 1;
+    node_config.state_sync.storage_service.max_concurrent_requests = 1;
```
zekun000, line 348: this look like what config optimizer should do? @JoshLind

davidiw (commented):
Most of these could actually be disabled.

JoshLind (commented):
Some high-level questions/concerns:
1. What's the doc goal here? Is this to benefit internal devs, external devs, or users?
   - If it's users (e.g., node operators), I'd prefer we document them elsewhere (e.g., on dedicated component docs, like I have for state sync: https://aptos.dev/guides/state-sync). Users won't read these (and likely shouldn't).
   - If it's internal (or external) devs, I'd argue that many of the comments are redundant (and devs shouldn't expect all code to be annotated -- this seems like an anti-pattern). Likewise, if we're going to do this, why not do it for all configs (instead of just this subset)?
 2. Generally, I'm not a big fan of requiring comments on all struct fields, etc., and it's likely why we've never enforced the "missing docs" [lint](https://doc.rust-lang.org/rustdoc/lints.html#missing_docs), because it tends to result in really meaningless comments that tend to create baggage.

Also, you could probably unblock the optimization part of the PR by splitting the PR into two: (i) single node optimizations; and (ii) doc comments.

Interested to see what others think though... (I know I'm pretty opinionated sometimes 😄)

(Rest of the review is just nitty nits that can be taken or left)

aptos-node/src/lib.rs
JoshLind: Yeah, we could probably set up an explicit optimizer for a single node deployment 😄

aptos-node/src/lib.rs
```diff
@@ -312,16 +312,40 @@ fn create_single_node_test_config(
...
         .mempool
         .shared_mempool_max_concurrent_inbound_syncs = 1;
+    node_config.mempool.default_failovers = 1;
+    node_config.mempool.max_broadcasts_per_peer = 1;
+
     node_config
         .state_sync
         .state_sync_driver
         .enable_auto_bootstrapping = true;
+    node_config.state_sync.state_sync_driver.max_consecutive_stream_notifications = 1;
```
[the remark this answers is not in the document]
gregnazario: I would argue, that there are too many configurations to tell what the configuration could be, so I changed all of them.

This is probably work that's more suited for someone other than me.

config/src/config/api_config.rs
```diff
@@ -13,43 +13,60 @@ use std::net::SocketAddr;
...
     pub max_transactions_page_size: u16,
+    /// Maximum page size for event paginated APIs
     pub max_events_page_size: u16,
+    /// Maximum page size for resource paginated APIs
     pub max_account_resources_page_size: u16,
+    /// Maximum page size for module paginated APIs
     pub max_account_modules_page_size: u16,
-
-    /// Max gas unit for view function.
+    /// Maximum gas unit limit for view functions
```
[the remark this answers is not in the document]
gregnazario: > Some high-level questions/concerns:
> 
> 1. What's the doc goal here? Is this to benefit internal devs, external devs, or users?
>    
>    * If it's users (e.g., node operators), I'd prefer we document them elsewhere (e.g., on dedicated component docs, like I have for state sync: https://aptos.dev/guides/state-sync). Users won't read these (and likely shouldn't).
>    * If it's internal (or external) devs, I'd argue that many of the comments are redundant (and devs shouldn't expect all code to be annotated -- this seems like an anti-pattern). Likewise, if we're going to do this, why not do it for all configs (instead of just this subset)?
> 2. Generally, I'm not a big fan of requiring comments on all struct fields, etc., and it's likely why we've never enforced the "missing docs" [lint](https://doc.rust-lang.org/rustdoc/lints.html#missing_docs), because it tends to result in really meaningless comments that tend to create baggage.
> 
> Also, you could probably unblock the optimization part of the PR by splitting the PR into two: (i) single node optimizations; and (ii) doc comments.
> 
> Interested to see what others think though... (I know I'm pretty opinionated sometimes 😄)
> 
> (Rest of the review is just nitty nits that can be taken or left)

I would argue, that these docs provide value in that someone can easily tell what a configuration is supposed to do.

I spent a lot of time fumbling through this configuration trying to figure out what "X" does.  In all of the state sync docs, there actually are comments on every single field, but aren't made into doc comments.

I'm not viewing that there should be a comment on "every struct", but I would argue something as key as configuration for the node should be easy to view as a reference, given that any docs in aptos.dev will lag behind these updated configurations.

config/src/config/storage_config.rs
```diff
@@ -24,11 +24,17 @@ pub const BUFFERED_STATE_TARGET_ITEMS: usize = 100_000;
 #[derive(Clone, Copy, Debug, Deserialize, PartialEq, Eq, Serialize)]
 #[serde(default, deny_unknown_fields)]
 pub struct RocksdbConfig {
+    /// Maximum number of files open by RocksDB at one time
     pub max_open_files: i32,
+    /// Maximum size of the RocksDB write ahead log (WAL)
     pub max_total_wal_size: u64,
+    /// Maximum number of threads for Rocks DB
```
grao1991, line 31: nit: s/threads/background threads

grao1991 (approved):

brianolson:
confirmed that `aptos node run-local-testnet` runs with much lower CPU, so LGTM

config/src/config/api_config.rs
```diff
@@ -13,43 +13,60 @@ use std::net::SocketAddr;
 #[derive(Clone, Debug, Deserialize, PartialEq, Eq, Serialize)]
 #[serde(default, deny_unknown_fields)]
 pub struct ApiConfig {
+    /// Enables the REST API endpoint
     #[serde(default = "default_enabled")]
     pub enabled: bool,
+    /// Address for the REST API to listen on.  Set to 0.0.0.0:<port> to allow all inbound connections.
```
davidiw, line 19:
```suggestion
    /// Address for the REST API to listen on. Set to 0.0.0.0:<port> to allow all inbound connections.
```

config/src/config/storage_config.rs
davidiw, line 31:
```suggestion
    /// Maximum number of background threads for Rocks DB
```

davidiw (approved):

zekun000 (approved):

PR #7944, opened 2023-04-26
[Docs] Correcting mixed up links and descriptions

Krishnacore:
In many places the networks were mixed up, corrected so that all links lead to the right place and all descriptions match


davidiw:
looks like you resolved this in another pr, thanks!

PR #7946, opened 2023-04-27
fix all the collection_id to index

lightmark:
### Description

As it is called as `index` in token.move but not consistent with collection.move.
This PR has to be cherry-picked to v1.4

### Test Plan
ut


movekevin (approved):

bowenyang007 (approved):
Thank you!

PR #7955, opened 2023-04-27
Fix docker-update-images workflow

gedigi:
### Description

Fixes a bug in which the workflows exits with an error when it doesn't find new updates. This was caused by the python script returning an "error" exit code, which is interpreted by GHA as a failed action. The python script now always exits gracefully, and triggers the next step via its output.

### Test Plan

### DIFF
--- a/.github/workflows/docker-update-images.yaml
+++ b/.github/workflows/docker-update-images.yaml
@@ -14,14 +14,16 @@ on:
 jobs:
   update:
     runs-on: ubuntu-latest
+
     steps:
       - uses: actions/checkout@v3
 
       - name: Check and pin updates (if any)
-        run: python3 scripts/update_docker_images.py
+        id: check_update
+        run: python3 scripts/update_docker_images.py >> $"GITHUB_OUTPUT"
 
       - name: Create Pull Request
-        if: ${{ success() }}
+        if: ${{ steps.check_update.outputs.NEEDS_UPDATE == "True" }}
         uses: peter-evans/create-pull-request@v5
         with:
           token: ${{ secrets.REPO_TOKEN }}
--- a/.github/workflows/docker-update-images.yaml
+++ b/.github/workflows/docker-update-images.yaml
@@ -23,7 +23,7 @@ jobs:
         run: python3 scripts/update_docker_images.py >> $"GITHUB_OUTPUT"
 
       - name: Create Pull Request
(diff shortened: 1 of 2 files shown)


scripts/update_docker_images.py
```diff
@@ -66,10 +66,11 @@ def update() -> int:
...
-        exit_code = 0
+        update_exists = True
 
-    return exit_code
+    return update_exists
 
 
 if __name__ == "__main__":
-    exit(update())
+    print("NEEDS_UPDATE={}".format(update()))
```
rustielin, line 75: nit: instead of printing this, you can probably write directly to `GITHUB_OUTPUT` from within this script: https://github.com/aptos-labs/aptos-core/blob/main/testsuite/determinator.py#L91-L98

rustielin (approved):

.github/workflows/docker-update-images.yaml
```diff
@@ -14,14 +14,16 @@ on:
...
       - uses: actions/checkout@v3
 
       - name: Check and pin updates (if any)
-        run: python3 scripts/update_docker_images.py
+        id: check_update
+        run: python3 scripts/update_docker_images.py >> $"GITHUB_OUTPUT"
 
       - name: Create Pull Request
-        if: ${{ success() }}
+        if: ${{ steps.check_update.outputs.NEEDS_UPDATE == "True" }}
```
rustielin, line 26: This `"True"` results in an error actually: 
image


https://github.com/aptos-labs/aptos-core/actions/runs/4824455865

rustielin (changes requested):

scripts/update_docker_images.py
gedigi: I left this as-is (and fixed the quote in the workflow) because this way you can see directly in the workflow that it evaluates the output of the python script without needing to open it.

sherry-x (approved):

rustielin (approved):

gedigi:
@rustielin I see the image hashes are now in `docker-bake-rust-all.hcl` - should I adapt this python script for that or do you plan to remove versioning and only rely on latest?

rustielin:
@gedigi yea that was part of some recent refactoring on the docker build process. Probably best to adapt this script to update those as well. We won't likely remove our version pins.

PR #7961, opened 2023-04-28
[Indexer] Add delegator pool balances and fix delegator balances

bowenyang007:
### Description
* Added delegator pool balances to track total delegated staking per period
* Added current delegator pool balances to track latest total coin / total shares per period (note that shares are divided by scaling_factor to avoid having ints that are too large for downstream services (e.g. bigquery)
* In delegator balances, rename amount to shares. To get amounts in APT we recommend joining against `current_delegated_staking_pool_balances` and doing the math of `shares / total shares * total coins` (to get latest balance)

### Testing
Number checks out now
image

Also balances look legit
image


jillxuu (approved):

larry-aptos (approved):

perryjrandall (approved):
nice test plan!

PR #7964, opened 2023-04-28
[Forge] Run some of the unstable tests on GCP clusters

sionescu:
### Description

Run part of the forge-unstable workflows on GCP, pinning to cluster "gcp-forge-0".

### Test Plan

Triggered manually: https://github.com/aptos-labs/aptos-core/actions/runs/4876839324.

### DIFF
--- a/.github/workflows/forge-unstable.yaml
+++ b/.github/workflows/forge-unstable.yaml
@@ -30,8 +30,8 @@ env:
   AWS_ACCOUNT_NUM: ${{ secrets.ENV_ECR_AWS_ACCOUNT_NUM }}
   AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
   AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
-  IMAGE_TAG: ${{ inputs.IMAGE_TAG }} # this is only used for workflow_dispatch, otherwise defaults to empty
   AWS_REGION: us-west-2
+  IMAGE_TAG: ${{ inputs.IMAGE_TAG }} # this is only used for workflow_dispatch, otherwise defaults to empty
 
 jobs:
   # This job determines the image tag and branch to test, and passes them to the other jobs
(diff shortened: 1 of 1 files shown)


.github/workflows/forge-unstable.yaml
```diff
@@ -30,8 +30,11 @@ env:
   AWS_ACCOUNT_NUM: ${{ secrets.ENV_ECR_AWS_ACCOUNT_NUM }}
   AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
   AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
-  IMAGE_TAG: ${{ inputs.IMAGE_TAG }} # this is only used for workflow_dispatch, otherwise defaults to empty
   AWS_REGION: us-west-2
+  GCP_WORKLOAD_IDENTITY_PROVIDER: ${{ secrets.GCP_WORKLOAD_IDENTITY_PROVIDER }}
```
rustielin, line 34: Do you need this? 

Looks like GCP_WORKLOAD_IDENTITY_PROVIDER and other gcp auth vars are automatically pulled in in the workflow via secrets context

rustielin (dismissed):

.github/workflows/forge-unstable.yaml
sionescu: Not sure. I'll remove them to see if the workflow works without.

perryjrandall (approved):
instead of pinning cluster, can we just pass "gcp" or cloud flag?

sionescu:
@perryjrandall I'll add a better form of cluster selection in another PR.

.github/workflows/forge-unstable.yaml
sionescu: Done.

sherry-x (approved):
Are we just pining it to a random cluster here? wondering why some are forge 1 vs 0 or if there's any logic for selecting them here

sionescu:
@sherry-x I added an explanation to the PR description (pin the cluster to avoid changing the global Forge config).

rustielin (approved):

PR #7976, opened 2023-04-28
[Peer Monitor] Add performance monitoring support.

JoshLind:
Note: most of this PR is just tests.

### Description

This PR adds simple network performance benchmarking to the peer monitoring service. This allows us to run end-to-end benchmarks to test network stack performance. Right now, the benchmark is simple: the peer monitoring service has all nodes send RPC requests to all other nodes at a configurable frequency (e.g., every second), with a configurable message size of random bytes (e.g., 512 KB). We can then track latency and throughput using the metrics.

Performance monitoring is protected by a feature flag (`network-perf-test`) and the config sanitizer. Nodes without the feature flag enabled will be unable to recognize performance messages and will simply drop them (i.e., similar to a message serialization failure).

The PR offers the following commits (which, I'll just squash when I land):
- Change the peer monitoring interval to use microseconds instead of milliseconds (so we can run the monitor more frequently).
- Change the request tracker to use microseconds instead of milliseconds (again, so we have a finer granularity).
- Add a simple performance monitoring config to the peer monitoring config.
- Add new request and response types for performance monitoring (protected by the feature flag).
- Add server-side support for performance monitoring requests (the server simply acks the message).
- Add client-side support for performance monitoring requests (most of this is new tests and boilerplate).
- Add CI/CD checks for the performance monitor (e.g., smoke and unit tests).

Once this PR lands, I'll need to continue with: (i) supporting direct send messages through the config flags; (ii) updating the node inspection service to display the peer monitoring metadata; (iii) updating the performance monitoring logic to better expose throughput and latency; and (iv) adding a dedicated forge test so that we can run this in a realistic network environment and see how things go.

For those curious, the reason that I've rolled this into the peer monitoring service (and not just add a new dedicated performance monitor) is because the peer monitoring service is perfect for this. All I've needed to do is add a new dedicated request and response type. Everything else we get for free, e.g., metrics, latencies, logging, error support, test framework, etc.

### Test Plan
New and existing unit (and smoke) tests! I've already run this locally and things seem to work reasonable well.

### DIFF
--- a/.github/workflows/lint-test.yaml
+++ b/.github/workflows/lint-test.yaml
@@ -156,6 +156,22 @@ jobs:
         env:
           RUST_MIN_STACK: 4297152
 
+  rust-network-perf-unit-test:
+    runs-on: high-perf-docker
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0 # get all the history because cargo xtest --change-since origin/main requires it.
+      - uses: aptos-labs/aptos-core/.github/actions/rust-setup@main
+        with:
+          GIT_CREDENTIALS: ${{ secrets.GIT_CREDENTIALS }}
+      - uses: taiki-e/install-action@v1.5.6
+        with:
+          tool: nextest
+      - run: | # Test the client and server
+          cargo nextest run --locked -p aptos-peer-monitoring-service-client --no-fail-fast -F network-perf-test
+          cargo nextest run --locked -p aptos-peer-monitoring-service-server --no-fail-fast -F network-perf-test
+
   rust-smoke-test:
     runs-on: high-perf-docker
     steps:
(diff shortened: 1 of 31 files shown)


aptos-node/Cargo.toml
```diff
@@ -82,4 +82,5 @@ consensus-only-perf-test = ["aptos-executor/consensus-only-perf-test", "aptos-me
 default = []
 failpoints = ["fail/failpoints", "aptos-consensus/failpoints", "aptos-executor/failpoints", "aptos-mempool/failpoints", "aptos-api/failpoints"]
 indexer = ["aptos-indexer"]
+network-perf-test = ["aptos-peer-monitoring-service-client/network-perf-test", "aptos-peer-monitoring-service-server/network-perf-test", "aptos-peer-monitoring-service-types/network-perf-test"]
```
ibalajiarun, line 85: basic question: we do expect this feature to be default at some point, right?

.github/workflows/lint-test.yaml
```diff
@@ -214,6 +229,32 @@ jobs:
             !/tmp/.tmp*/**/db/
           retention-days: 14
 
+  rust-network-perf-smoke-test:
```
ibalajiarun, line 232: We should also have a forge test. Problem is we need a separate docker image because of the feature flag. Why do we use a feature over a config flag?

ibalajiarun (approved):
LGTM, some questions.

brianolson (approved):

aptos-node/Cargo.toml
JoshLind: No, not in a production environment. This feature will only be for end-to-end network performance benchmarking (similar to the consensus only tests). One application is to deploy this feature in a chaos-mesh like environment (the one you're working on), so we can see how latencies and packet drops might affect end-to-end throughput. Does that make sense?

.github/workflows/lint-test.yaml
JoshLind: Yeah, I'll add a forge test once this stack lands. Then hopefully we can utilize the network simulation stuff you're working on to see how the networking stack behaves.

> Why do we use a feature over a config flag?

This is mainly for security. With the feature enabled it allows bad peers to send performance messages, which may be huge or frequent. Without the feature, nodes will not accept these types of messages and simply drop them. I wanted to avoid the situation where someone "accidentally" enables this in the config and starts to spam the network.

network/peer-monitoring-service/client/src/logging.rs
```diff
@@ -46,6 +46,9 @@ pub enum LogEntry {
     NodeInfoRequest,
     PeerMonitorLoop,
     SendRequest,
+
+    #[cfg(feature = "network-perf-test")] // Disabled by default
+    PerformanceMonitoringRequest,
```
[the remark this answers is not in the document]
JoshLind: Ughh... I can't think of a use-case. Nodes without support for the feature will simply drop the performance messages because the message enum won't have an entry in them. So, this will essentially be a serialization failure.

.github/workflows/lint-test.yaml
ibalajiarun: Understood from previous answer. I thought this was going to be a default feature at some point. Makes sense now. thanks!

Issue #5349, opened 2022-10-29
[Feature Request] Add Create Resource account and publish package API in TS SDK


dhruvja:
# 🚀 Feature Request

Adding a function in SDK that would create resource account and publish package i.e. call this function in module - https://github.com/aptos-labs/aptos-core/blob/main/aptos-move/framework/aptos-framework/sources/resource_account.move#L121 . Right now we only have a function to publish package, adding this would be a great thing.

## Motivation

- There is a lot of ambiguity in what parameters mean and where to fetch them. The `publish` function in SDK makes it clear and takes care of the serializing part.
- Many people are struggling to call the module function and it would really good to have a function which can do by just passing the important parameters and serializing and other things done under the hood.

## Pitch

Having a function in SDK which takes parameters and does serializing under the hood . And then make a transaction to `0x1::resource_account::create_resource_account_and_publish` . 

I would be willing to open a pull request and add the above solution.

davidiw:
> I would be willing to open a pull request and add the above solution.

Please do :)

Shusanta:
has this been added? @dhruvja @davidiw

PR #7983, opened 2023-04-29
added create resource account and publish package method with a test …

dhruvja:
### Description

This PR is regarding the issue: https://github.com/aptos-labs/aptos-core/issues/5349


ecosystem/typescript/sdk/src/tests/e2e/aptos_client.test.ts
```diff
@@ -533,6 +533,43 @@ test(
...
+          ).toUint8Array(),
+        ),
+      ],
+    );
+
+    await client.waitForTransaction(txnHash);
```
0xmaayan, line 565:
```suggestion
    await client.waitForTransaction(txnHash,{ checkSuccess: true });
```

and then we can remove lines 657-568

ecosystem/typescript/sdk/src/providers/aptos_client.ts
```diff
@@ -770,6 +770,41 @@ export class AptosClient {
...
+   * `packageMetadata` and `modules` can be generated with command
+   * `aptos move compile --save-metadata [ --included-artifacts=<...> ]`. 
+   * @param sender
+   * @param seed seeds for creation of resource address
+   * @param packageMetadata package metadata bytes
+   * @param modules bytecodes of modules
+   * @param extraArgs
+   */
+
+  async createResourceAccountAndpublishPackage(
```
0xmaayan, line 784:
```suggestion
  async createResourceAccountAndPublishPackage(
```

ecosystem/typescript/sdk/src/providers/aptos_client.ts
```diff
@@ -770,6 +770,41 @@ export class AptosClient {
...
 
+  /**
+   * Publishes a move packages by creating a resource account. The package cannot be upgraded since it is deployed by resource account
+   * `packageMetadata` and `modules` can be generated with command
+   * `aptos move compile --save-metadata [ --included-artifacts=<...> ]`. 
+   * @param sender
+   * @param seed seeds for creation of resource address
+   * @param packageMetadata package metadata bytes
+   * @param modules bytecodes of modules
+   * @param extraArgs
```
0xmaayan, line 781:
```suggestion
   * @param extraArgs
   * @returns Transaction hash
```

0xmaayan (commented):
Thank you for looking into it! 
I wonder if it makes more sense to introduce a `ResourceAccount` class with this function, and then we would have this class holds all the `resource_account.move` module [entry functions](https://github.com/aptos-labs/aptos-core/blob/main/aptos-move/framework/aptos-framework/sources/resource_account.move#L84) and the SDK [getResourceAccountAddress](https://github.com/aptos-labs/aptos-core/blob/main/ecosystem/typescript/sdk/src/account/aptos_account.ts#L115) generate method

dhruvja:
> Thank you for looking into it! I wonder if it makes more sense to introduce a `ResourceAccount` class with this function, and then we would have this class holds all the `resource_account.move` module [entry functions](https://github.com/aptos-labs/aptos-core/blob/main/aptos-move/framework/aptos-framework/sources/resource_account.move#L84) and the SDK [getResourceAccountAddress](https://github.com/aptos-labs/aptos-core/blob/main/ecosystem/typescript/sdk/src/account/aptos_account.ts#L115) generate method

Well not a bad idea, but there are hardly any functions entry functions in `resource_account.move` atm.

0xmaayan:
@dhruvja Please fix `Lint+Test` test, than I will approve

davidiw (approved):

ecosystem/typescript/sdk/src/providers/aptos_client.ts
```diff
@@ -770,6 +770,43 @@ export class AptosClient {
...
+   * @returns Transaction hash
+   */
+
```
davidiw, line 785:
```suggestion
```

0xmaayan (approved):

PR #7984, opened 2023-04-29
fixed deployment page with the correct links and descriptions

dhruvja:
### Description

I found that some of the links and data in the deployment page (https://aptos.dev/nodes/deployments/) were in a wrong place. For eg: the devnet faucet URL was in testnet column and so on. This fix has solved those issues on that page.

### <samp>🤖 Generated by Copilot at e8f11b6</samp>

Updated documentation for Testnet and Devnet deployments in `deployments.md`. Reordered columns and revised release and wipe schedules to reflect current practices.

### Test Plan
Please check if the details regarding to the network in their respective columns match their column header.

### DIFF
--- a/developer-docs-site/docs/nodes/deployments.md
+++ b/developer-docs-site/docs/nodes/deployments.md
@@ -6,18 +6,18 @@ hide_table_of_contents: true
 
 # Aptos Deployments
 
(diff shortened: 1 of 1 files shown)


0xmaayan (approved):

davidiw (approved):
thanks

PR #7987, opened 2023-04-30
[python] Add a transaction management layer

davidiw:
The bulk of this work is in adding a transaction management layer with some other fixes along the way.

This provides a framework for managing as many transactions from a
single account at once
* The AccountSequenceNumber allocates up to 100 outstanding sequence numbers to maximize the number of concurrent transactions in the happy path.
* The transaction manager provides async workers that push a transaction from submission through to validating completion
Together they provide the basic harness for scaling transaction
submission on the Aptos blockchain from a single account.

This should be reasonably copyable into other languages and provides a starting point to helping others build their own transaction management planes.


jjleng:
if multiple SDK instances submit txns from the same account, we will see repeated 30s timeouts and seq number resyncs. The sequence number syncs do not always return the true latest sequence number due to full node sync delay, which would make things worse. 

Wonder if we should start building the API with a cached state for accounts and then all clients don't need to worry about how to hack around the mempool capability as this PR did. cc @bowenyang007. Since the current mempool fails to provide a consistent view of sequence numbers and often swallows errors, adding a global cache state will simplify programming model a lot.

ecosystem/python/sdk/aptos_sdk/account_sequence_number.py
```diff
@@ -0,0 +1,178 @@
...
+
+from aptos_sdk.account_address import AccountAddress
+from aptos_sdk.async_client import RestClient
+
+
+class AccountSequenceNumber:
+    """
+    A managed wrapper around sequence numbers that implements the trivial flow control used by the
+    Aptos faucet:
+    * Submit up to 50 transactions per account in parallel with a timeout of 20 seconds
```
gregnazario, line 16: I think the number below now is 100 and not 50.  Similarly the sleep time is .01 and not .1

ecosystem/python/sdk/aptos_sdk/async_client.py
```diff
@@ -408,8 +421,8 @@ async def transfer(
 
         payload = {
             "type": "entry_function_payload",
-            "function": "0x1::coin::transfer",
-            "type_arguments": ["0x1::aptos_coin::AptosCoin"],
```
gregnazario, line 412: :)

ecosystem/python/sdk/aptos_sdk/account_sequence_number.py
```diff
@@ -0,0 +1,178 @@
...
+                    if not block:
+                        return None
+                    await asyncio.sleep(self.sleep_time)
+                    if time.time() - start_time > self.maximum_wait_time:
+                        logging.warn(
+                            f"Waited over 30 seconds for a transaction to commit, resyncing {self.account.address().hex()}"
+                        )
+                        await self.initialize()
+                    else:
+                        await self.__update()
```
gregnazario, line 82: Personally, I prefer comments to tell what we're checking for each of these.

It looks like:
1. Check if the queue is initialized
2. Check if there is over the number in flight
3. Sleep if it's blocking on transactions, otherwise continue to send transactions (need to define `block` in docs)
4. Poll and block to see if the mempool has cleared any transactions
5. If it's clear (or block is false), submit the transaction, and let the next one grab the lock.

ecosystem/python/sdk/aptos_sdk/account_sequence_number.py
```diff
@@ -0,0 +1,178 @@
...
+        self.last_uncommitted_number = self.current_number
+
+    async def synchronize(self):
+        """
+        Poll the network until all submitted transactions have either been committed or until
+        the maximum wait time has elapsed. This will prevent any calls to next_sequence_number
+        until this called has returned.
+        """
+        if self.last_uncommitted_number == self.current_number:
+            return
```
gregnazario, line 101: This is a short cut, but maybe want to provide an input flag to sync regardless (if for some reason you've got something else also handling transactions).

ecosystem/python/sdk/aptos_sdk/account_sequence_number.py
```diff
@@ -0,0 +1,178 @@
...
+
+        await self.lock.acquire()
+        try:
+            await self.__update()
+            start_time = time.time()
+            while self.last_uncommitted_number != self.current_number:
+                print(f"{self.last_uncommitted_number} {self.current_number}")
```
gregnazario, line 108: Probably want to remove, or provide some better context to this print message

ecosystem/python/sdk/aptos_sdk/account_sequence_number.py
```diff
@@ -0,0 +1,178 @@
...
+                if time.time() - start_time > self.maximum_wait_time:
```
gregnazario, line 109: You make this condition a lot, maybe move it into a function?

ecosystem/python/sdk/aptos_sdk/account_sequence_number.py
gregnazario, line 108: This is roughly the same logic as the other loop, probably want to combine

ecosystem/python/sdk/examples/transaction-batching.py
```diff
@@ -2,26 +2,203 @@
...
+        self._txn_generator = TransactionGenerator(self._rest_client, self._recipient)
+        self._txn_worker = TransactionWorker(
+            self._account, self._rest_client, self._txn_generator.next_transaction
+        )
+
+    def run(queue: Pipe, node_url: str, account: Account, recipient: AccountAddress):
+        worker = Worker(queue, node_url, account, recipient)
+        asyncio.run(worker.arun())
+
+    async def arun(self):
```
gregnazario, line 135: nit: prefer `async_run` or `run_async`

ecosystem/python/sdk/examples/transaction-batching.py
```diff
@@ -2,26 +2,203 @@
...
+                    exception,
+                ) = await self._txn_worker.next_processed_transaction()
+                if exception:
+                    print(
+                        f"Account {self._txn_worker.account()}, transaction {sequence_number} submission failed: {exception}"
+                    )
+                else:
+                    txn_hashes.append(txn_hash)
+
+            print(f"Submit txns from {self._account.address()}", flush=True)
```
gregnazario, line 164: I would change this to `Submitted txns` at this point they're already submitted.

gregnazario (approved):
Most of my comments are just around lack of comments.  Otherwise the logic is good

ecosystem/python/sdk/aptos_sdk/account_sequence_number.py
```diff
@@ -0,0 +1,178 @@
...
+
+    def __init__(self, client: RestClient, account: AccountAddress):
+        self.client = client
+        self.account = account
+        self.lock = asyncio.Lock()
+
+        self.last_uncommitted_number = None
+        self.current_number = None
+
+    async def next_sequence_number(self, block: bool = True) -> Optional[int]:
```
igor-aptos, line 54: maybe call this acquire_next_sequence_number to convey the notion that this might block for a long time.

ecosystem/python/sdk/aptos_sdk/transaction_worker.py
```diff
@@ -0,0 +1,225 @@
...
+            while True:
+                sequence_number = (
+                    await self._account_sequence_number.next_sequence_number()
+                )
+                transaction = await self._transaction_generator(
+                    self._account, sequence_number
+                )
+                txn_hash_awaitable = self._rest_client.submit_bcs_transaction(
+                    transaction
+                )
```
igor-aptos, line 74: just a thought (not sure you want to overcomplicate this) - one potential fault-tolerant improvement, is to have AccountSequenceNumber have two counters, issued and submitted, and next_sequence_number increments issued, and after submit_bcs_transaction is called , if transaction was potentially submitted (we received success response, or received no response) we notify AccountSequenceNumber to increment submitted counter
If we received failed response for submission, we notify AccountSequenceNumber to decrement issued number.

and AccountSequenceNumber is changed to never have issued be more than 1 above submitted. 
that would make it such that single failure in submission doesn't look up the whole stream until expiry

ecosystem/python/sdk/aptos_sdk/transaction_worker.py
```diff
@@ -0,0 +1,225 @@
...
+        except Exception as e:
+            # This is insufficient, if we hit this we either need to bail or resolve the potential errors
+            logging.error(e, exc_info=True)
+
+    async def _process_transactions_task(self):
+        try:
+            while True:
+                # Always start waiting for one
+                (
+                    txn_awaitable,
```
igor-aptos, line 89: haven't used SDK before, is this waiting for txn hash (polling the rest endpoint for txn hash) or what is it doing?

looking at ecosystem/python/sdk/aptos_sdk/async_client.py , I see that signature is 
    def submit_bcs_transaction(self, signed_transaction: SignedTransaction) -> str:
so return type is str, not something you can wait on?

ecosystem/python/sdk/aptos_sdk/account_sequence_number.py
davidiw: right, right, I want to also tweak the delay... .01 is probably too aggressive and likely leads to more wasted networking.

ecosystem/python/sdk/aptos_sdk/account_sequence_number.py
```diff
@@ -0,0 +1,178 @@
...
+    * They are initialized to the current on-chain state, so if there are already transactions in
+      flight, they make take some time to reset.
+    * Accounts are automatically initialized if not explicitly
+
+    Notes:
+    * This is co-routine safe, that is many async tasks can be reading from this concurrently.
+    * The synchronize method will create a barrier that prevents additional next_sequence_number
+      calls until it is complete.
+    * This only manages the distribution of sequence numbers it does not help handle transaction
+      failures.
```
banool, line 32: It might be good to include what happens when you submit a transaction and it is instantly, synchronously rejected.

ecosystem/python/sdk/aptos_sdk/account_sequence_number.py
```diff
@@ -0,0 +1,178 @@
...
+    """
+
+    client: RestClient
+    account: AccountAddress
+    lock = asyncio.Lock
```
banool, line 37: Do you mean this?
```suggestion
    lock: asyncio.Lock
```

ecosystem/python/sdk/aptos_sdk/account_sequence_number.py
```diff
@@ -0,0 +1,178 @@
...
+
+    maximum_in_flight: int = 100
+    maximum_wait_time = 30
+    sleep_time = 0.01
```
banool, line 41: How come type hints in some cases but not others?

ecosystem/python/sdk/aptos_sdk/account_sequence_number.py
```diff
@@ -0,0 +1,178 @@
...
+
+        self.last_uncommitted_number = None
+        self.current_number = None
+
+    async def next_sequence_number(self, block: bool = True) -> Optional[int]:
+        """
+        Returns the next sequence number available on this account. This leverages a lock to
+        guarantee first-in, first-out ordering of requests.
+        """
+        await self.lock.acquire()
```
banool, line 59: Nit: I'd recommend using this as a context manager to make it impossible to forget to release.

ecosystem/python/sdk/aptos_sdk/account_sequence_number.py
```diff
@@ -0,0 +1,178 @@
...
+            return
+
+        await self.lock.acquire()
+        try:
+            await self.__update()
+            start_time = time.time()
+            while self.last_uncommitted_number != self.current_number:
+                print(f"{self.last_uncommitted_number} {self.current_number}")
+                if time.time() - start_time > self.maximum_wait_time:
+                    logging.warn(
```
banool, line 110: Could be good to let the user toggle this on or off.

ecosystem/python/sdk/aptos_sdk/account_sequence_number.py
```diff
@@ -0,0 +1,178 @@
...
+                        f"Waited over 30 seconds for a transaction to commit, resyncing {self.account.address}"
```
banool, line 111:
```suggestion
                        f"Waited over {self.maximum_wait_time} seconds for a transaction to commit, resyncing {self.account.address}"
```

ecosystem/python/sdk/aptos_sdk/account_sequence_number.py
```diff
@@ -0,0 +1,178 @@
...
+
+    Notes:
+    * This is co-routine safe, that is many async tasks can be reading from this concurrently.
+    * The synchronize method will create a barrier that prevents additional next_sequence_number
+      calls until it is complete.
+    * This only manages the distribution of sequence numbers it does not help handle transaction
+      failures.
+    """
+
+    client: RestClient
```
banool, line 35: Nit: I notice in the TransactionWorker you prefix all the private fields with `_`, but you don't do it here. Let's go with one or the other.

ecosystem/python/sdk/aptos_sdk/account_sequence_number.py
```diff
@@ -0,0 +1,178 @@
...
+    account: AccountAddress
```
banool, line 36: Is this type correct? I see you're using this as an account, since you're calling `self.account.address()`. I guess this is meant to be just an `Account`?

ecosystem/python/sdk/aptos_sdk/transaction_worker.py
```diff
@@ -0,0 +1,225 @@
...
+        )
+        self._rest_client = rest_client
+        self._transaction_generator = transaction_generator
+
+        self._started = False
+        self._stopped = False
+        self._outstanding_transactions = asyncio.Queue()
+        self._processed_transactions = asyncio.Queue()
+
+    def account(self) -> AccountAddress:
```
banool, line 60:
```suggestion
    def address(self) -> AccountAddress:
```

ecosystem/python/sdk/aptos_sdk/transaction_worker.py
```diff
@@ -0,0 +1,225 @@
...
+
+    async def next_processed_transaction(
+        self,
+    ) -> (int, typing.Optional[str], typing.Optional[Exception]):
+        return await self._processed_transactions.get()
+
+    def stop(self):
+        """Stop the tasks for managing transactions"""
+        if not self._started:
+            raise Exception("Start not yet called")
```
banool, line 128: Nit: RuntimeError or something instead?

ecosystem/python/sdk/aptos_sdk/transaction_worker.py
```diff
@@ -0,0 +1,225 @@
...
+        except asyncio.CancelledError:
+            return
+        except Exception as e:
+            # This is insufficient, if we hit this we either need to bail or resolve the potential errors
+            logging.error(e, exc_info=True)
+
+    async def _process_transactions_task(self):
+        try:
+            while True:
+                # Always start waiting for one
```
banool, line 87: I'm not sure I understand why we do this?

banool (changes requested):

0xmaayan (commented):
does http2 needed to use this layer?

ecosystem/python/sdk/aptos_sdk/transaction_worker.py
0xmaayan, line 89:
```suggestion
                    txn_hash_awaitable,
```

ecosystem/python/sdk/aptos_sdk/account_sequence_number.py
```diff
@@ -0,0 +1,178 @@
...
+    lock = asyncio.Lock
+
+    maximum_in_flight: int = 100
```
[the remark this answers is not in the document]
davidiw: yup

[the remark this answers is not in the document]
davidiw: I need to add a config here in the future....

ecosystem/python/sdk/aptos_sdk/account_sequence_number.py
davidiw: brain problems of the author.

ecosystem/python/sdk/aptos_sdk/transaction_worker.py
```diff
@@ -0,0 +1,225 @@
...
+                        sequence_number,
+                    ) = await self._outstanding_transactions.get()
+                    awaitables.append(txn_awaitable)
+                    sequence_numbers.append(sequence_number)
+                outputs = await asyncio.gather(*awaitables, return_exceptions=True)
+
+                for (output, sequence_number) in zip(outputs, sequence_numbers):
+                    if isinstance(output, BaseException):
+                        await self._processed_transactions.put(
+                            (sequence_number, None, output)
```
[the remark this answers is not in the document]
0xmaayan: oh I read it wrong, `output` **is** the exception

0xmaayan (approved):
beautiful

ecosystem/python/sdk/aptos_sdk/account_sequence_number.py
```diff
@@ -0,0 +1,175 @@
...
+    * Mempool limits the number of transactions per account to 100, hence why we chose 100.
+    """
+
+    _client: RestClient
+    _account: AccountAddress
+    _lock: asyncio.Lock
+
+    _maximum_in_flight: int = 100
+    _maximum_wait_time: int = 30
+    _sleep_time: int = 0.01
```
banool, line 44:
```suggestion
    _sleep_time: float = 0.01
```

banool (approved):
Amazing!!

One small thing to fix before landing so I disabled auto merge.

ecosystem/python/sdk/aptos_sdk/account_sequence_number.py
davidiw: omg

davidiw:
Time to just land this and update in a new pr...

PR #7989, opened 2023-04-30
[docs] add transaction management

davidiw:
this is an advanced builders guide for scaling the submission of transactions to the blockchain through a single on-chain account leveraging lots of worker accounts

note, also created a new link for advanced builders in the doc site and moved the start a local test environment there


developer-docs-site/docs/guides/transaction-management.md
```diff
@@ -0,0 +1,97 @@
...
+
+1. At startup, query the blockchain for the account’s current sequence number.
+2. Support up to 100 transactions in flight at the same time, that is 100 sequence numbers can be allocated without confirming that any have been committed.
+3. If there are 100 transactions in flight, determine the actual committed state by querying the network. This will update the current sequence number.
+4. If there are less than 100 transactions in flight, return to step 2.
+5. Otherwise sleep for .1 seconds and continue to re-evaluate the current on-chain sequence number.
+6. All transactions should have an expiration time. If the expiration time has passed, assume that there has been a failure and reset the sequence number. The trivial case is to only monitor for failures when the maximum number of transactions are in flight and to let other services manages this otherwise.
+
+Note: In general, this is intended to be a happy path framework. That is any failure during transaction submission should delay until all transactions have timed out and should be resubmitted. Handling more complex
```
gregnazario, line 36: Sentence just ends...

gregnazario (approved):

developer-docs-site/docs/guides/transaction-management.md
```diff
@@ -0,0 +1,97 @@
...
+This section covers the major aspects of building a transaction management harness. This consists of the following core components:
+
+- A sequence number generator that allocates and manages available sequence numbers for a single account.
+- A transaction manager that receives payloads from an application or a user, sequence numbers from the sequence number generator, and has access to the account key to combine the three pieces together into a viable signed transaction. It then also takes the responsibility for pushing the transaction to the blockchain.
+- An on-chain worker, leader harness that let’s multiple accounts share the signer of a single shared account.
+
+Currently this framework assumes that the network builds no substantial queue, that is a transaction that is submitted executes and commits with little to no delay. In order to address high demand, this work needs to be extended with the following components:
+
+- Optimizing `base_gas_unit` price to ensure priority transactions can be committed to the blockchain.
+- Further handling of transaction processing rates to ensure that the expiration timer is properly set.
```
igor-aptos, line 16: add:

It is expected a layer on top to be handling retries, for each individual transaction, and creating new (same or different payloads) 

Due to nature of sequence numbers, it is required that each account is managed only by a single machine/instance of transaction manager. Otherwise each instance will be creating transactions with overlapping sequence numbers, and in-memory state of Transaction Manager will become stale as well.

developer-docs-site/docs/guides/transaction-management.md
```diff
@@ -0,0 +1,97 @@
...
+### Managing Sequence Numbers
+
+Each transaction requires a distinct sequence number that is sequential to previously submitted transactions. This can be provided by the following process:
+
+1. At startup, query the blockchain for the account’s current sequence number.
+2. Support up to 100 transactions in flight at the same time, that is 100 sequence numbers can be allocated without confirming that any have been committed.
+3. If there are 100 transactions in flight, determine the actual committed state by querying the network. This will update the current sequence number.
+4. If there are less than 100 transactions in flight, return to step 2.
+5. Otherwise sleep for .1 seconds and continue to re-evaluate the current on-chain sequence number.
+6. All transactions should have an expiration time. If the expiration time has passed, assume that there has been a failure and reset the sequence number. The trivial case is to only monitor for failures when the maximum number of transactions are in flight and to let other services manages this otherwise.
```
igor-aptos, line 34: if you have small throughput, step 3 will never be called, and sequence number rechecked for successful transactions, so all transactions will be treated as failed? or is the order in step 6 the opposite - you first check the sequence number, and then treat some as succeeded and some as expired based on onchain seq_num and onchain_timestamp?

developer-docs-site/docs/guides/transaction-management.md
igor-aptos: failures handled above is logical. 

But this sounds like Transaction Manager will get stuck on a single transaction submission failure, instead of it just not incrementing seq_num in that case?

igor-aptos: ah I see, implementation is done this way, so the comment is correct

developer-docs-site/docs/guides/transaction-management.md
davidiw: So I need to add some more logic here and there, but the gist of it is that the actual manager should provide the feedback or the user can call synchronize

davidiw: so I added a new component that checks the first transaction that was logged and times it out if appropriate

0xmaayan (approved):
looks good!

I am curios how the end user/developer use this layer, would they need to make sure they submit transactions from the same account? as submitting from different accounts would not work, should the layer validates the transactions are from the same account?

davidiw:
> 

Great question. We should definitely evaluate how descriptive we should be here. My implementation basically wraps around a single account, so this issue isn't revealed. We'll likely need to explore more and then discuss integration strategies later.

Issue #7741, opened 2023-04-13
[Feature Request] Add an `upcoming_rewards` function to vesting contract


banool:
Then finally we need an `upcoming_rewards` function or something that is the sum of `accumulated_rewards` (aka `unlockable_rewards`) and `distributable_rewards`.

Related PRs:
- https://github.com/aptos-labs/aptos-core/issues/7739
- https://github.com/aptos-labs/aptos-core/issues/7740
- https://github.com/aptos-labs/aptos-core/issues/7741

Issue #7740, opened 2023-04-13
[Feature Request] Add function called `distributable_rewards` in vesting contract


banool:
We should have a function called `distributable_rewards`, which corresponds to how much will be distributed out when you call distribute.

See also:
- https://github.com/aptos-labs/aptos-core/issues/7739
-

Issue #7739, opened 2023-04-13
[Feature Request] Improve comments of `accumulated_rewards` in vesting contract


banool:
The accumulated_rewards view function doesn’t really do what (I interpret) the comment to mean. It says this:

> Return the accumulated rewards that have not been distributed to the provided shareholder.

I’d assume this means that it tells me what will be given out when I call distribute, but it seems to actually return the amount that hasn’t been unlocked yet. I figured this out because when I call unlock, this value goes to zero.

I’d really consider it `unlockable_rewards`. Unfortunately we can't rename the function now, but we can add comments that clear this up.

Related PRs:
- https://github.com/aptos-labs/aptos-core/issues/7739
- https://github.com/aptos-labs/aptos-core/issues/7740
- https://github.com/aptos-labs/aptos-core/issues/7741

PR #7990, opened 2023-05-01
[framework] Add rewards view functions to vesting

dboures:
### Description
This PR adds `upcoming_rewards` functions as specified in #7741, `distributable_rewards` functions as specified in #7740, and changes the comments of the `accumulated_rewards` functions for greater clarity as specified in #7739

### Test Plan
The `get_stake` and `total_distributable_amount` spec can be tested with `aptos move prove -f staking_contract`. 

I have also added a unit test titled`test_distributable_rewards_with_multiple_stakers` to demonstrate how the `distributable_rewards` functions work

### DIFF
--- a/aptos-move/framework/aptos-framework/sources/staking_contract.move
+++ b/aptos-move/framework/aptos-framework/sources/staking_contract.move
@@ -215,11 +215,47 @@ module aptos_framework::staking_contract {
     #[view]
     /// Return the number of pending distributions (e.g. commission, withdrawals from stakers).
     ///
-    /// This errors out the staking contract with the provided staker and operator doesn't exist.
+    /// This errors out if the staking contract with the provided staker and operator doesn't exist.
     public fun pending_distribution_counts(staker: address, operator: address): u64 acquires Store {
         assert_staking_contract_exists(staker, operator);
         let staking_contracts = &borrow_global<Store>(staker).staking_contracts;
         pool_u64::shareholders_count(&simple_map::borrow(staking_contracts, &operator).distribution_pool)
+    }
+
+    #[view]
+    /// Return the amount that will be split between all stakers the next time distribute() is called.
+    ///
+    /// This errors out if the staking contract with the provided staker and operator doesn't exist.
+    public fun total_distributable_amount(staker: address, operator: address): u64 acquires Store {
+        assert_staking_contract_exists(staker, operator);
+        let store = borrow_global_mut<Store>(staker);
+        let staking_contract = simple_map::borrow(&store.staking_contracts, &operator);
+        let pool_address = staking_contract.pool_address;
+        let (_, inactive, _, pending_inactive) = stake::get_stake(pool_address);
+        inactive + pending_inactive
+    }
+
+    #[view]
+    /// Return the amount that will be disbursed to the staker the next time distribute() is called.
+    ///
+    /// This errors out if the staking contract with the provided staker and operator doesn't exist.
+    public fun distributable_amount(staker: address, operator: address): u64 acquires Store {
+        let total_distributable_amount = total_distributable_amount(staker, copy operator);
+        assert_staking_contract_exists(staker, operator);
+        let store = borrow_global_mut<Store>(staker);
+        let staking_contract = simple_map::borrow(&store.staking_contracts, &operator);
+        let shares = pool_u64::shares(&staking_contract.distribution_pool, staker);     
+        let new_staker_balance =
+            pool_u64::shares_to_amount_with_total_coins(
+                &staking_contract.distribution_pool, 
+                shares, 
+                total_distributable_amount
+            );
+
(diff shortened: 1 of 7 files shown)


aptos-move/framework/aptos-framework/sources/staking_contract.move
```diff
@@ -215,13 +215,24 @@ module aptos_framework::staking_contract {
...
 
+    #[view]
+    /// Return the amount that the staker will receive next time distribute() is called.
+    ///
+    /// This errors out if the staking contract with the provided staker and operator doesn't exist.
+    public fun distributable_amount(staker: address, operator: address): u64 acquires Store {
+        assert_staking_contract_exists(staker, operator);
+        let store = borrow_global_mut<Store>(staker);
+        let staking_contract = simple_map::borrow(&store.staking_contracts, &operator);
+        pool_u64::balance(&staking_contract.distribution_pool, staker)
```
xindingw, line 233: Haven't read through the changes in `vesting.move` but this doesn't look right. `staking_contract.distribution_pool` might not be up-to-date if there is new inactive. So `distribute` always checks if there is new inactive then [update_distribution_pool](https://github.com/aptos-labs/aptos-core/blob/f3db990e9ae1ad80627f64017b018434a3c9bffc/aptos-move/framework/aptos-framework/sources/staking_contract.move#L595).

dboures: Thanks for the pointer! I have updated the code to take this into account now

aptos-move/framework/aptos-framework/sources/staking_contract.move
```diff
@@ -215,13 +215,49 @@ module aptos_framework::staking_contract {
...
+    /// Return the amount that will be split between all stakers the next time distribute() is called.
+    ///
+    /// This errors out if the staking contract with the provided staker and operator doesn't exist.
+    public fun total_distributable_amount(staker: address, operator: address): u64 acquires Store {
+        assert_staking_contract_exists(staker, operator);
+        let store = borrow_global_mut<Store>(staker);
+        let staking_contract = simple_map::borrow(&store.staking_contracts, &operator);
+        let pool_address = staking_contract.pool_address;
+        let (_, inactive, _, pending_inactive) = stake::get_stake(pool_address);
+        inactive + pending_inactive
```
xindingw, line 235: [stake::withdraw_with_cap](https://github.com/aptos-labs/aptos-core/blob/64ac160cc2464f1026509495dc2e67f3bbabbde4/aptos-move/framework/aptos-framework/sources/staking_contract.move#L587) doesn't guarantee that you can withdraw `total_potential_withdrawable`. Check the logic [here](https://github.com/aptos-labs/aptos-core/blob/64ac160cc2464f1026509495dc2e67f3bbabbde4/aptos-move/framework/aptos-framework/sources/stake.move#L904).

aptos-move/framework/aptos-framework/sources/staking_contract.move
```diff
@@ -215,13 +215,49 @@ module aptos_framework::staking_contract {
...
+    /// Return the amount that will be disbursed to the staker the next time distribute() is called.
+    ///
+    /// This errors out if the staking contract with the provided staker and operator doesn't exist.
+    public fun distributable_amount(staker: address, operator: address): u64 acquires Store {
+        let total_distributable_amount = total_distributable_amount(staker, copy operator);
+        assert_staking_contract_exists(staker, operator);
+        let store = borrow_global_mut<Store>(staker);
+        let staking_contract = simple_map::borrow(&store.staking_contracts, &operator);
+        let shares = pool_u64::shares(&staking_contract.distribution_pool, staker);     
+        let new_staker_balance =
```
xindingw, line 248: Can we wrap this as function and use the function in `update_distribution_pool `? It will guarantee that the amount will be same.

aptos-move/framework/aptos-framework/sources/staking_contract.move
```diff
@@ -215,13 +215,49 @@ module aptos_framework::staking_contract {
...
+            pool_u64::shares_to_amount_with_total_coins(
+                &staking_contract.distribution_pool, 
+                shares, 
+                total_distributable_amount
+            );
+
+        let old_staker_balance = pool_u64::balance(&staking_contract.distribution_pool, staker);
+        let commission = 
+            (new_staker_balance - old_staker_balance) * staking_contract.commission_percentage / 100;
+        new_staker_balance - commission
```
xindingw, line 258: We need to fully follow the logic in `update_distribution_pool`. Here we call `pool_u64::amount_to_shares_with_total_coins` which may bring precision error. So the equivalent amount might have a slight difference from `commission`.

xindingw (commented):
Not very familiar with `vesting.move`, need some time to take a look.

davidiw:
As this has been sitting here for a month going to close out to ping author.

PR #7991, opened 2023-05-01
[forge] Remove conflicting feature definition

perryjrandall:
When pulling in a library from aptos-core, if you end up depending on k8s-openapi this will actually break the build with the following message

```
Library crates *must not* enable any features in their direct dependency on k8s-openapi, only in their dev-dependency. The choice of which Kubernetes version to support should be left to the final binary crate, so only binary crates should enable a specific feature. If library crates also enabled features, it could cause multiple features to be enabled simultaneously, which k8s-openapi does not support.
```

So this makes the features specified by the only binary that uses it forge, instead of making that decision for all libraries using it

Test Plan: cargo build + tests on PR pass

Try and pull from another project which has a k8s-openapi dep which previously failed, with this branch, does not fail

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -461,9 +461,7 @@ jemallocator = { version = "0.3.2", features = [
 ] }
 json-patch = "0.2.6"
 jsonwebtoken = "8.1"
-k8s-openapi = { version = "0.11.0", default-features = false, features = [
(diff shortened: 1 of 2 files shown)


perryjrandall:
Abandoning

PR #8000, opened 2023-05-01
[execution] move the reconfig suffix check to buffer manager

zekun000:
This enables us to remove the special case on the block_executor side later


grao1991:
@zekun000 also remove the unnecessary logic from execution side?

PR #8004, opened 2023-05-02
[release] Fix tooling

runtian-zhou:
### Description

Fix the tooling such that it fetch exactly one event instead of 100 events that can lead to potential pruning issues.

### Test Plan
Framework testing would work.

### DIFF
--- a/aptos-move/aptos-release-builder/src/validate.rs
+++ b/aptos-move/aptos-release-builder/src/validate.rs
@@ -216,7 +216,7 @@ impl NetworkConfig {
                 "0x1::aptos_governance::GovernanceEvents",
                 "create_proposal_events",
                 None,
-                Some(100),
+                Some(1),
             )
             .await?
             .into_inner()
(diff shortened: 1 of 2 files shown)


aptos-move/aptos-release-builder/src/validate.rs
```diff
@@ -216,7 +216,7 @@ impl NetworkConfig {
                 "0x1::aptos_governance::GovernanceEvents",
                 "create_proposal_events",
                 None,
-                Some(100),
+                Some(1),
```
perryjrandall, line 219: Do we eventually need to handle pagination here?

perryjrandall (approved):

aptos-move/aptos-release-builder/src/validate.rs
runtian-zhou: I don't think so. We should only need one event here cuz we just created one. One thing to handle tho is if we need any cleanups after some txn failed in the middle.

lightmark (approved):
lgtm

PR #8005, opened 2023-05-02
[forge] upgrade `kube` crate

ibalajiarun:
### Description

This PR upgrades the `kube` crate to `v0.65.0`. I couldn't upgrade to a recent version because `kube` requires a recent version of `zeroize` but there is a minimum dependency requirement on `zeroize` by `aptos-crypto` crate. This was necessary to switch from using `native-tls` to `rustls-tls` feature, because `native-tls` caused an issue on macos when establishing TLS connections via client certificates to recent k8s versions >= v1.25.

### Test Plan

Existing tests


rustielin (approved):

Cargo.toml
```diff
@@ -466,10 +466,8 @@ jemallocator = { version = "0.3.2", features = [
 ] }
 json-patch = "0.2.6"
 jsonwebtoken = "8.1"
-k8s-openapi = { version = "0.11.0", default-features = false, features = [
-    "v1_15",
-] }
-kube = { version = "0.51.0", features = ["jsonpatch"] }
+k8s-openapi = { version = "0.13.1", default-features = false, features = ["v1_22"] }
```
perryjrandall, line 469: So if you try to include this from another binary that has this dep, it will fail since the v1_22 flag cannot be enabled in several places

I tried to fix this to no avail, instead actually I suggest moving the k8s deps to the forge binary Cargo toml

perryjrandall (approved):

Cargo.toml
ibalajiarun: Sounds good, moved these dependencies to forge crate.

PR #8009, opened 2023-05-02
Update staking-pool-operations.md

michelle-aptos:
add section on rewards tracking and move owner operations to this page


davidiw (dismissed):
could we make these commits into a single PR? I'll try to trace this and get it to work right. It might be useful to work with an eng on how we better onboard folks for more complicated workflows like yours?

davidiw:
> could we make these commits into a single PR? I'll try to trace this and get it to work right. It might be useful to work with an eng on how we better onboard folks for more complicated workflows like yours?

I think the answer is to either just continue working in a single commit or you make each PR point to the branch from the next PR.

PR #8010, opened 2023-05-02
Update index.md

michelle-aptos:
add section on stake pool vs. delegation pool


davidiw (changes requested):
delegated staking doesn't exist, the links for concepts were off by one and I fixed that.

PR #8019, opened 2023-05-02
[forge] Support for real multi-region experiments on GCP

ibalajiarun:
### Description

This PR adds support to run Forge tests on the multiregion cluster on GCP. With this change, one should be able to run such experiments on forge from their local machine in forge local mode. GHA integration and k8s forge mode will be in a subsequent PR.

- The genesis helm chart now accepts new helm-values to enable multiregion mode and the cluster names. In this mode, genesis script will generate hostnames that will allow validators between different clusters to communicate with each other using the provided cluster name helm-values.
- The aptos-node helm chart also accepts new helm-values to enable multiregion mode. In this mode, the aptos-node statefulsets and services will include a new labels that will allow karmada to spread them across different clusters. Furthermore, the service resource can be deployed in headless mode, so that the host names are evaluated directly to pod IPs and not service IPs. Only pod IPs are reachable between clusters.

### Test Plan

Test in GCP.


terraform/helm/aptos-node/templates/_helpers.tpl
```diff
@@ -46,6 +46,20 @@ app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
 app.kubernetes.io/managed-by: {{ .Release.Service }}
 {{- end -}}
 
+{{/*
+Multicluster labels
+*/}}
+{{- define "aptos-validator.multiclusterLabels" -}}
+{{- $ctx := index $ 0 -}}
```
rustielin, line 53: maybe include some comments on how this works. Took me a while to figure out the `index $ 0` syntax was for fetching the function params, like doing something like: `"aptos-validator.multiclusterLabels" (tuple $ $i)`

terraform/helm/genesis/values.yaml
```diff
@@ -69,6 +69,10 @@ genesis:
     internal_host_suffix: fullnode-lb
   # -- The local path for move modules in the docker image. Defaults to the aptos-framework in the aptoslabs/tools docker image
   moveModulesDir: /aptos-framework/move/modules
+  multicluster:
```
rustielin, line 72: same as other `values.yaml` -- add comment for experimental, and regenerate the docs

testsuite/forge-cli/src/main.rs
```diff
@@ -1531,6 +1532,50 @@ fn multi_region_multi_cloud_simulation_test(config: ForgeConfig<'static>) -> For
...
+            helm_values["genesis"]["multicluster"]["enabled"] = true.into();
+        }))
+        .with_node_helm_config_fn(Arc::new(|helm_values| {
+            helm_values["multicluster"]["enabled"] = true.into();
+            // Create headless services for validators and fullnodes. 
+            // Note: chaos-mesh will not work with clusterIP services.
+            helm_values["service"]["validator"]["internal"]["type"] = "ClusterIP".into();
+            helm_values["service"]["validator"]["internal"]["headless"] = true.into();
+            helm_values["service"]["fullnode"]["internal"]["type"] = "ClusterIP".into();
+            helm_values["service"]["fullnode"]["internal"]["headless"] = true.into();
```
rustielin, line 1552: nit: probably have some helper to setup the values needed for multiregion, so it can be reused in other test suites?

testsuite/forge-cli/src/main.rs
```diff
@@ -1531,6 +1532,50 @@ fn multi_region_multi_cloud_simulation_test(config: ForgeConfig<'static>) -> For
...
+
+            // Disable pyroscope. Need to ensure secrets are created/propogated before enabling.
+            helm_values["pyroscope"]["enabled"] = false.into();
```
rustielin, line 1555: yea we should either disable or put more effort into bringing back pyroscope

testsuite/forge_wrapper_core/cluster.py
```diff
@@ -146,7 +146,21 @@ def assert_auth(self, shell: Shell) -> None:
...
+            cmd = [
+                "gcloud",
+                "secrets",
+                "versions",
+                "access",
+                "latest",
+                "--secret",
+                "karmada-kubeconfig",
+                "--project",
+                "forge-gcp-multiregion-test",
```
rustielin, line 159: should probably get out of hardcoding this eventually

rustielin (approved):
LGTM overall, just some nits and also generally to try to split out (via comments, docs, etc) the experimental multi-region stuff with the rest of forge

Test plan LGTM, as the land-blocking Forge uses the charts from this repo

terraform/helm/aptos-node/values.yaml
```diff
@@ -14,6 +14,10 @@ numValidators: 1
 # -- Total number of fullnode groups to deploy
 numFullnodeGroups: 1
 
+multicluster:
+  enabled: false 
+  targetClusters: ["cluster1", "cluster2", "cluster3"]
```
sionescu, line 19: Please make this 0, 1 and 2.

terraform/helm/genesis/files/genesis.sh
```diff
@@ -23,6 +23,25 @@ LARGER_STAKE_AMOUNT=${LARGER_STAKE_AMOUNT:-1}
...
+echo $MULTICLUSTER_DOMAIN_SUFFIXES_STRING
+# convert comma separated string to array
+IFS=',' read -r -a MULTICLUSTER_DOMAIN_SUFFIXES <<< "$MULTICLUSTER_DOMAIN_SUFFIXES_STRING"
+
+if ! [[ $(declare -p MULTICLUSTER_DOMAIN_SUFFIXES) =~ "declare -a" ]]; then
+    echo "MULTICLUSTER_DOMAIN_SUFFIXES must be an array"
+    exit 1
+fi
+
+if [ "${ENABLE_MULTICLUSTER_DOMAIN_SUFFIX}" = "true" ]; then
```
sionescu, line 38: Use the built-in "[[" test everywhere.
This should be:
```
  if [[ ${ENABLE_MULTICLUSTER_DOMAIN_SUFFIX} == "true" ]]; then
```

terraform/helm/genesis/files/genesis.sh
```diff
@@ -23,6 +23,25 @@ LARGER_STAKE_AMOUNT=${LARGER_STAKE_AMOUNT:-1}
 # TODO: Fix the usage of this below when not set
 RANDOM_SEED=${RANDOM_SEED:-$RANDOM}
 
+ENABLE_MULTICLUSTER_DOMAIN_SUFFIX=${ENABLE_MULTICLUSTER_DOMAIN_SUFFIX:-false}
+MULTICLUSTER_DOMAIN_SUFFIXES_DEFAULT="forge-multiregion-1,forge-multiregion-2,forge-multiregion-3"
+MULTICLUSTER_DOMAIN_SUFFIXES_STRING=${MULTICLUSTER_DOMAIN_SUFFIXES_STRING:-$MULTICLUSTER_DOMAIN_SUFFIXES_DEFAULT}
```
sionescu, line 28: Try to wrap references with curly braces everywhere, e.g. ${MULTICLUSTER_DOMAIN_SUFFIXES_DEFAULT}.

testsuite/forge_wrapper_core/cluster.py
sionescu: Yes, but the bigger problem is that it requires a different invocation of `gcloud` to setup.

sionescu (approved):

terraform/helm/aptos-node/values.yaml
ibalajiarun: So, 0 is kind of special, it refers to the karmada master cluster in which you cannot schedule any other pods, so 1, 2, 3.

testsuite/forge-cli/src/main.rs
ibalajiarun: I will follow-up in the next PR.

testsuite/forge_wrapper_core/cluster.py
ibalajiarun: Yes, since this is a special cluster, we have to bear with this unfortunately.

PR #8020, opened 2023-05-02
[testnet] release v1.4 with hotfix

lightmark:
### <samp>🤖 Generated by Copilot at 772f577</samp>

Updated the release proposal to apply a hot fix for the framework. The hot fix resolves a bug in `collection.move` that emitted wrong events for token v2.

### Test Plan
cargo run -p aptos-release-builder -- generate-proposals --release-config aptos-move/aptos-release-builder/data/release.yaml --output-dir /tmp/banana

### DIFF
--- a/aptos-move/aptos-release-builder/data/release-v1.4-hotfix.yaml
+++ b/aptos-move/aptos-release-builder/data/release-v1.4-hotfix.yaml
@@ -1,0 +1,14 @@
+---
+remote_endpoint: ~
+proposals:
+  - name: upgrade_of_framework_v1_4_0_with_token_v2_hot_fix
+    metadata:
+      title: "Fix the issue of wrong events emit from collection.move"
+      description: "This is a hot fix of framework v.1.4.0 to emit token v2 BurnEvent"
+      source_code_url: "TBD"
+      discussion_url: "https://github.com/aptos-labs/aptos-core/pull/8011"
+    execution_mode: MultiStep
+    update_sequence:
+      - Framework:
+          bytecode_version: 6
+          git_hash: a4db3a9bbe090946d7cc90f94b56fcee5615622a


aptos-move/aptos-release-builder/data/release.yaml
```diff
@@ -1,63 +1,14 @@
...
     metadata:
-      title: "Enable the new signature checker"
-      description: "This enables the improved signature checker implementation."
+      title: "Fix the issue of wrong events emit from collection.move"
+      description: "This is a hot fix of framework v.1.4.0 to emit token v2 BurnEvent"
       source_code_url: "TBD"
-      discussion_url: "https://github.com/aptos-labs/aptos-core/pull/7856"
-    execution_mode: MultiStep
+      discussion_url: "https://github.com/aptos-labs/aptos-core/pull/8011"
+    execution_mode: SingleStep
```
perryjrandall, line 10: MultiStep, pretty much everything is multistep, maybe @runtian-zhou we can either make this default or error when this is not set properly

lightmark: Yeah, this is something I am not sure. It seems each subdir is a step even for single framework update.

perryjrandall: The name is misleading, we treat almost everything as "multistep" now

runtian-zhou: I would encourage using multistep cuz we will be deprecate single step soon.

runtian-zhou:
I think this makes sense. But my question is we still need the original contents of this file for mainnet. My suggestion is instead of mutating the existing `release.yaml` file, make a copy and point this file on interal-ops git action

sherry-x (changes requested):
we don't need to land this.

perryjrandall (changes requested):
We dont need this

We will just rerun the workflow which will deploy all the changes in release yaml

PR #8032, opened 2023-05-03
[move-examples] Add an example for how to wrap TokenV1 in TokenV2

gregnazario:
### <samp>🤖 Generated by Copilot at e2c2830</samp>

This pull request adds a new wrapped_token example to the aptos-move/move-examples directory. The example demonstrates how to use the Aptos framework to migrate from Token V1 to Token V2 by wrapping Token V1 in a Token V2 resource. The example consists of a Move.toml file and a wrapped_token.move module.

### DIFF
--- a/aptos-move/move-examples/token_migration/sources/token_migration.move
+++ b/aptos-move/move-examples/token_migration/sources/token_migration.move
@@ -1,0 +1,116 @@
+/// This provides an example for migrating an NFT collection and its tokens from the token v1 standard to token v2.
+/// The migrated tokens will contain the original token, so no tokens will be burnt.
+///
+/// WARNING: This example does not apply to fungible or semi-fungible token collections.
+///
+/// For more context on the token v1 standard, see https://aptos.dev/standards/aptos-token.
+/// For token v2, see https://aptos.dev/standards/aptos-token-v2.
+///
+/// The migration flow is as follows:
+/// 1. Create a new collection using the token v2 standard by calling `create_migrated_collection`.
+/// Currently only the creator can call this function.
+/// 2. Wrap each token from the original token v1 collection by calling `migrate_token`. The creator needs to own the
+/// token to be migrated.
+///
+/// Note that this example uses AptosToken which is the convenient no code solution built on top of token v2. Developers
+/// can use their custom implementation if desired.
+module token_migration::token_migration {
+    use aptos_framework::object::{Self, ExtendRef};
+    use aptos_token::token as tokenv1;
+    use aptos_token_objects::aptos_token as tokenv2;
+    use aptos_token_objects::token as base_tokenv2;
+    use aptos_token_objects::royalty as tokenv2_royalty;
+    use std::option::{Self, Option};
+    use std::string::String;
+    use std::signer;
+
+    #[resource_group_member(group = aptos_framework::object::ObjectGroup)]
+    /// Storage state for managing the no-code Token.
+    struct MigratedToken has key {
+        /// Token to wrap
+        token: tokenv1::Token,
+        /// Used to burn.
+        burn_ref: Option<base_tokenv2::BurnRef>,
+        /// Used to control freeze.
+        transfer_ref: Option<object::TransferRef>,
+        /// Used to mutate fields
+        mutator_ref: Option<base_tokenv2::MutatorRef>,
+        /// Used to mutate fields
+        extend_ref: Option<ExtendRef>,
+    }
+
+    /// Create a collection based on the previous collection
+    public entry fun create_migrated_collection(creator: &signer, collection_name: String) {
+        let creator_address = signer::address_of(creator);
+
+        let collection_description = tokenv1::get_collection_description(creator_address, collection_name);
+        let collection_uri = tokenv1::get_collection_uri(creator_address, collection_name);
+        let max_supply = tokenv1::get_collection_maximum(creator_address, collection_name);
+        let collection_mutability_config = tokenv1::get_collection_mutability_config(creator_address, collection_name);
+
+        // For now, allow mutable everything, TODO: Make this configurable / use previous state
+        tokenv2::create_collection(
+            creator,
+            collection_description,
+            max_supply,
+            collection_name,
+            collection_uri,
+            tokenv1::get_collection_mutability_description(&collection_mutability_config),
+            true,
+            tokenv1::get_collection_mutability_uri(&collection_mutability_config),
+            // In v1, token mutability configs is set on the token instead of collection.
+            // This example defaults to true here but the creator can deploy custom code built on token v2 to maintain
+            // the same control setup as token v1 if desired.
+            true, /* mutable_token_description */
+            true, /* mutable_token_name */
+            true, /* mutable_token_properties */
+            true, /* mutable_token_uri */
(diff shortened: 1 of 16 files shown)


aptos-move/move-examples/wrapped_token/sources/wrapped_token.move
```diff
@@ -0,0 +1,110 @@
+/// This provides an example for how to wrap a Token V1 in a Token V2
+module wrapped_token::wrapped_token {
+
+    use std::string::String;
```
movekevin, line 4: nit: Can you order these alphabetically?

aptos-move/move-examples/wrapped_token/sources/wrapped_token.move
```diff
@@ -0,0 +1,110 @@
+/// This provides an example for how to wrap a Token V1 in a Token V2
```
movekevin, line 1: Can you add more info here (ChatGPT?) about what a wrapped TokenV2 means, why do so and the basic user flow?

aptos-move/move-examples/wrapped_token/sources/wrapped_token.move
```diff
@@ -0,0 +1,110 @@
...
+    /// Create a collection based on the previous collection
+    ///
+    /// TODO: what should we do with the original collection?
+    public entry fun create_wrapped_collection(
+        new_creator: &signer,
+        creator_address: address,
+        name: String,
+    ) {
+        // TODO: Relax later?
+        assert!(signer::address_of(new_creator) == creator_address, 12345);
```
movekevin, line 37: nit: Use constant for error code

aptos-move/move-examples/wrapped_token/sources/wrapped_token.move
```diff
@@ -0,0 +1,110 @@
...
+        /// Used to mutate fields
+        extend_ref: Option<ExtendRef>,
+    }
+
+    /// Create a collection based on the previous collection
+    ///
+    /// TODO: what should we do with the original collection?
+    public entry fun create_wrapped_collection(
+        new_creator: &signer,
+        creator_address: address,
```
movekevin, line 33: Might as well skip this param if it's expected to be the same as the signer. This is not subject to compat rule (being an example) so if we want new_creator != creator_address, we can add creator_address param in a later version

aptos-move/move-examples/wrapped_token/sources/wrapped_token.move
```diff
@@ -0,0 +1,110 @@
...
+        let max_supply = aptos_token::token::get_collection_maximum(creator_address, name);
+
+        // For now, allow mutable everything, TODO: Make this configurable / use previous state
+        0x4::aptos_token::create_collection(
+            new_creator,
+            description,
+            max_supply,
+            name,
+            uri,
+            true,
```
movekevin, line 49: These all being true look suspect. Are there any values we can use from v1 collection?

aptos-move/move-examples/wrapped_token/sources/wrapped_token.move
```diff
@@ -0,0 +1,110 @@
...
+            true,
+            true,
+            true,
+            true,
+            true,
+            true,
+            true,
+            true,
+            0,
```
movekevin, line 58: Can we leave a comment here that the token's royalty will be the primary and this (collection-level royalty) is not used in this wrapped version?

aptos-move/move-examples/wrapped_token/sources/wrapped_token.move
```diff
@@ -0,0 +1,110 @@
...
+    public fun wrap(
+        creator: &signer,
+        receiver: address,
+        collection: String,
+        name: String,
+    ) {
+        let creator_address = signer::address_of(creator);
+        let data_id = aptos_token::token::create_token_data_id(creator_address, collection, name);
+        let token_id = aptos_token::token::create_token_id(data_id, 0);
+        // Load token TODO: handle fungible?
```
movekevin, line 73: Let's just throw an error here if the v1 token is fungible to make it clear this example does not support fungible tokens yet

aptos-move/move-examples/wrapped_token/sources/wrapped_token.move
```diff
@@ -0,0 +1,110 @@
...
+            collection,
+            description,
+            name,
+            option::none(),
+            uri
+        );
+        royalty::init(&token_constructor, royalty);
+
+        let signer = object::generate_signer(&token_constructor);
+        let burn_ref = generate_burn_ref(&token_constructor);
```
movekevin, line 95: Do we always need all of these refs? For example burn_ref is not needed if the collection is not burnable by creator I think

davidiw (commented):
should we think about the various options here?

this is a container model, but the burn model is more compact ... right? do we want to support both?

gregnazario:
> should we think about the various options here?
> 
> this is a container model, but the burn model is more compact ... right? do we want to support both?

We should support both.  Keeping in mind that any tokens that don't support burn, will need to use this container model.

movekevin:
> should we think about the various options here?
> 
> this is a container model, but the burn model is more compact ... right? do we want to support both?

This is an example so I think we can just offer both burning + wrapping as 2 different examples. People can then pick whichever one works for them

aptos-move/move-examples/wrapped_token/sources/wrapped_token.move
```diff
@@ -0,0 +1,110 @@
...
+        mutator_ref: Option<0x4::token::MutatorRef>,
+        /// Used to mutate fields
+        extend_ref: Option<ExtendRef>,
+    }
+
+    /// Create a collection based on the previous collection
+    ///
+    /// TODO: what should we do with the original collection?
+    public entry fun create_wrapped_collection(
+        new_creator: &signer,
```
davidiw, line 32: This doesn't feel like the right thing here..
* Make it clear that this is a non-creator generated collection
* Make it so that the creator can have the official tokenv2 collection

aptos-move/move-examples/token_migration/sources/token_migration.move
```diff
@@ -0,0 +1,116 @@
...
+            // Royalty is set on the tokens in v1, so we'll maintain the same setup during migration. This sets
+            // the royalty on collections, so we'll set it to 0.
+            0,
+            1
+        );
+    }
+
+    /// This can only called by the creator, who also needs to own the token being migrated.
+    /// If creators want to let the token owner migrate their own tokens, they would need to deploy custom code.
+    public fun migrate_token(creator: &signer, receiver: address, collection_name: String, token_name: String) {
```
davidiw, line 79: Why does this need to be called by the creator? Feels like it should be callable by anyone that has the token. As an entry fun could be callable by whomever has it in their TokenStore?

davidiw (changes requested):
been a while since I requested changes.... anyway trivial fix

aptos-move/move-examples/token_migration/sources/token_migration.move
gregnazario: So, it looks like the state that this was left in is broken and is half migrated to `AptosToken` which doesn't work with the token level royalties.

This will need to be rewritten over again with the standard `TokenObject`

PR #8034, opened 2023-05-03
Update staking-pool-operations.md

michelle-aptos:
add derive resource account

### DIFF
--- a/developer-docs-site/docs/nodes/validator-node/operator/staking-pool-operations.md
+++ b/developer-docs-site/docs/nodes/validator-node/operator/staking-pool-operations.md
@@ -17,7 +17,11 @@ The current required minimum for staking is 1 million APT.
 
 ## Initializing the stake pool
 
-Make sure that this step is performed by the owner. See [Initialize staking pool](../../validator-node/owner/index.md#initialize-staking-pool) in the owner documentation section.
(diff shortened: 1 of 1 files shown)


developer-docs-site/docs/nodes/validator-node/operator/staking-pool-operations.md
```diff
@@ -17,7 +17,11 @@ The current required minimum for staking is 1 million APT.
 
 ## Initializing the stake pool
 
-Make sure that this step is performed by the owner. See [Initialize staking pool](../../validator-node/owner/index.md#initialize-staking-pool) in the owner documentation section.
+Make sure that this step is performed by the owner. See [Initialize staking pool](../../validator-node/owner/index.md#initialize-staking-pool) in the owner documentation section. 
+
+To get the stake pool address before initializing the stake pool, run this command:
+
+```aptos account derive-resource-account-address --address <owner-address>  --seed ""```
```
davidiw, line 24: I don't think this is correct: https://github.com/aptos-labs/aptos-core/blob/77da3ded71bb9a58e207af4e865aa75bab78dde2/types/src/account_address.rs#L200

I think we'd be better off writing a specific function here as there's no easy way to actually generate that seed.

cc; @gregnazario / @movekevin

davidiw (changes requested):
Lemme know if my understanding is wrong.

PR #8038, opened 2023-05-03
[trivial] Make TransactionFactory fields public

igor-aptos:
When creating multiple transactions, in order to compute needed funds in the account or the cost of transaction, on top of having TransactionFactory, I need to pass gas cost/max gas.

Having access to those fields, makes it easier to manage, I am not sure of any negatives though. I can also provide getters, instead of making fields public

### <samp>🤖 Generated by Copilot at 28a61bd</samp>

Made some fields of `TransactionFactory` public and refactored transaction creation logic. This improves the flexibility and usability of the SDK for creating transactions.

### DIFF
--- a/sdk/src/transaction_builder.rs
+++ b/sdk/src/transaction_builder.rs
@@ -127,6 +127,18 @@ impl TransactionFactory {
         self
     }
 
+    pub fn get_max_gas_amount(&self) -> u64 {
+        self.max_gas_amount
+    }
+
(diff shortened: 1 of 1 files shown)


davidiw (commented):
prefer getters, since we already have setters, but will second stamp anyone more opinionated...

igor-aptos:
changed to getters

sitalkedia (approved):

igor-aptos:
merged into #8039 (which is also accepted), and will land together there.

PR #8044, opened 2023-05-03
Revert "Revert "refactor a bit (#7918)""

gerben-stavenga:
Reverts aptos-labs/aptos-core#8040

The cause was me deleting a feature flag at

https://github.com/aptos-labs/aptos-core/blob/cfecfb386614b708ada45ad81034a49be6fc33cf/aptos-move/aptos-vm/src/aptos_vm_impl.rs#L603

Me reasoning was that get_module_metadata works independently of the feature flag because it just looks up the existence of the V1 key. But the feature flag causes some metadata to be dropped for txn where the feature isn't activated yet but where the metadata is already V1


zekun000 (approved):
good finding

davidiw (approved):

PR #8059, opened 2023-05-04
[aptos-debugger] Fix resource group fetching

runtian-zhou:
### Description

The logic for getting values in resource group was broken in the validator-interface repo: there will be nothing under the struct tag if we are querying by resource group. To work around this temporarily,  I simulated the resource group behavior by fetching all the resources and return the superset blob back to caller. This is fairly ugly workaround which may require us add a better api for resource group in the future

### Test Plan

Manually tested the debugger with a txn that read resource group and make sure it can be executed successfully.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3145,6 +3145,7 @@ dependencies = [
  "aptos-storage-interface",
  "aptos-types",
  "async-trait",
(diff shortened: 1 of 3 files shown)


zekun000 (approved):
hmm, does it actually work?

runtian-zhou:
Yep it worked :D

wrwg (approved):

PR #8066, opened 2023-05-04
Send aptos node metrics to only aptos-telmetry-service

vusirikala:
### Description
This PR is for [BLK-515](https://linear.app/aptoslabs/issue/BLK-515/[telemetry]-deprecate-ga4-for-aptos-node). Currently, we still send aptos node metrics to Google Analytics. This PR deprecates that. After this PR, aptos node metrics will be sent only to aptos-telemetry-service.


ibalajiarun (approved):
LGTM. Thank you!

PR #8067, opened 2023-05-04
[benchmark] Support multiple move modules and add --module-working-set-size flag

igor-aptos:
…

### <samp>🤖 Generated by Copilot at 3cf5a74</samp>

This pull request adds a new feature to the transaction generator and executor libraries and tools, which allows testing custom module transactions with different working set sizes of the modules. It refactors the `TransactionTypeArg` enum to accept a parameter for the number of modules, and adds a new simple package type that has a fixed module layout. It also updates the testsuite and the benchmark tool to support this feature, and renames some variables and functions to improve clarity.


msmouse (approved):
stamp to unblock.

In what sense "multiple move modules"? Multiple under the same account?

msmouse (approved):
stamp to unblock.

In what sense "multiple move modules"? Multiple under the same account?

sitalkedia (approved):
LGTM

igor-aptos:
> stamp to unblock.
> 
> In what sense "multiple move modules"? Multiple under the same account?


@msmouse :

"Support multiple move modules" refers to previously there only being a single move package that was compiled for the benchmark ("simple"), and everything need to be there. With this change you can now have multiple move packages/modules (i.e. folders and files) that will all be compiled for the benchmark, and be usable to generate transactions

"add --module-working-set-size flag" - surfacing the ability to publish the module multiple times, under different accounts, to the CLI (previously it was hardcoded in the TransactionTypeArg, meaning I needed to create things like GlobalResource and TenGlobalResources)

PR #7877, opened 2023-04-20
[move] skip core dump with storage error

zekun000:

PR #8081, opened 2023-05-05
Add stack trace for invariant violation to error message

gerben-stavenga:
### Description

The stacktrace proved useful in the SEV and it might be useful to have it standard in debug mode, for future incidents

### <samp>🤖 Generated by Copilot at 99062dd</samp>

This pull request improves error handling and debugging in `aptos-move`. It refactors the `session.rs` file to use a closure for errors and adds assertions and stack traces for invariant violations in `move-binary-format`.


aptos-move/aptos-vm/src/move_vm_ext/session.rs
```diff
@@ -167,8 +167,10 @@ where
     ) -> VMResult<(MoveChangeSet, MoveChangeSet)> {
         // The use of this implies that we could theoretically call unwrap with no consequences,
         // but using unwrap means the code panics if someone can come up with an attack.
-        let common_error = PartialVMError::new(StatusCode::UNKNOWN_INVARIANT_VIOLATION_ERROR)
-            .finish(Location::Undefined);
+        let common_error = || {
+            PartialVMError::new(StatusCode::UNKNOWN_INVARIANT_VIOLATION_ERROR)
+                .finish(Location::Undefined)
+        };
```
davidiw, line 173: so fancy

third_party/move/move-binary-format/src/errors.rs
```diff
@@ -309,10 +309,21 @@ impl PartialVMError {
     }
 
     pub fn new(major_status: StatusCode) -> Self {
+        debug_assert!(major_status != StatusCode::EXECUTED);
```
davidiw, line 312: we had typically not enabled these with the logic that we don't want different behavior from local and tests, but my memory isn't great either...

davidiw (commented):
why only with debug assertions? is there a way to explicitly tell the logger to include the backtrace? @JoshLind

third_party/move/move-binary-format/src/errors.rs
gerben-stavenga: Typically debug asserts are used to document preconditions, but okay to confirm to a aptos style guide if this is not accepted.

zekun000 (commented):
we probably need to avoid speculatively resulting this error first, otherwise it's going to be expensive

third_party/move/move-binary-format/src/errors.rs
```diff
@@ -329,7 +340,14 @@ impl PartialVMError {
         self
     }
 
-    pub fn with_message(mut self, message: String) -> Self {
+    pub fn with_message(mut self, mut message: String) -> Self {
+        if cfg!(debug_assertions)
+            && self.0.major_status == StatusCode::UNKNOWN_INVARIANT_VIOLATION_ERROR
```
runtian-zhou, line 345: I was wondering if we should just append at all times.

gerben-stavenga: I'm not sure if message goes into the block hash?

gerben-stavenga:
> why only with debug assertions? is there a way to explicitly tell the logger to include the backtrace? @JoshLind



> we probably need to avoid speculatively resulting this error first, otherwise it's going to be expensive

I don't think even in speculative execution we should ever throw invariant error.

zekun000:
> I don't think even in speculative execution we should ever throw invariant error.

Yes, we shouldn't but not sure if we already do. For example I'm changing this from unknown invariant error to storage error https://github.com/aptos-labs/aptos-core/pull/7877

third_party/move/move-binary-format/src/errors.rs
vgao1996: 1. We should append at all times.
2. I don't think this goes into the block hash -- only the status code does. @msmouse is this correct?

third_party/move/move-binary-format/src/errors.rs
```diff
@@ -330,10 +330,19 @@ impl PartialVMError {
     }
 
     pub fn new(major_status: StatusCode) -> Self {
+        debug_assert!(major_status != StatusCode::EXECUTED);
+        let message = if major_status == StatusCode::UNKNOWN_INVARIANT_VIOLATION_ERROR {
+            Some(format!(
+                "UNKNOWN_INVARIANT_VIOLATION_ERROR at {:?}",
+                std::backtrace::Backtrace::force_capture()
```
vgao1996, line 337: Let's probably limit the number of frames we record using https://docs.rs/backtrace/latest/backtrace/

gerben-stavenga: Done

runtian-zhou (approved):

PR #8082, opened 2023-05-05
fix consensus name in docs

sasha8:
### Description

We no longer use HotStuff, it is about time we fix it in the docs.


zekun000 (approved):

consensus/src/lib.rs
```diff
@@ -7,7 +7,7 @@
 //! Consensus for the Aptos Core blockchain
 //!
 //! The consensus protocol implemented is AptosBFT (based on
-//! [HotStuff](https://arxiv.org/pdf/1803.05069.pdf)).
+//! [DiemBFT](https://developers.diem.com/papers/diem-consensus-state-machine-replication-in-the-diem-blockchain/2021-08-17.pdf)).
```
ibalajiarun, line 10: refer to Joleton instead?

ibalajiarun (approved):
but, address the dangling sentence above please.

consensus/README.md
```diff
@@ -24,16 +24,16 @@ In AptosBFT, validators receive transactions from clients and share them with ea
...
+The AptosBFT protocol decomposes into modules for safety (voting and commit rules) and liveness (round_state). This decoupling provides the ability to develop and experiment independently and on different modules in parallel. Due to the simple voting and commit rules, protocol safety is easy to implement and verify. It is straightforward to integrate execution as a part of consensus to avoid forking issues that arise from non-deterministic execution in a leader-based protocol. We did not consider proof-of-work based protocols, such as [Bitcoin](https://bitcoin.org/bitcoin.pdf), due to their poor performance.
 and high energy (and environmental) costs.
```
davidiw, line 32:
```suggestion
The AptosBFT protocol decomposes into modules for safety (voting and commit rules) and liveness (round_state). This decoupling provides the ability to develop and experiment independently and on different modules in parallel. Due to the simple voting and commit rules, protocol safety is easy to implement and verify. It is straightforward to integrate execution as a part of consensus to avoid forking issues that arise from non-deterministic execution in a leader-based protocol. We did not consider proof-of-work based protocols, such as [Bitcoin](https://bitcoin.org/bitcoin.pdf), due to their poor performance and high energy (and environmental) costs.
```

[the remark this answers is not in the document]
sasha8: Thanks

consensus/src/lib.rs
sasha8: That PDF is more accurate to what we implement in case someone will actually read.

consensus/README.md
sasha8: Thanks, fixed already.

gelash (approved):

PR #8083, opened 2023-05-05
[Execution] Add ExecutedBlock struct.

grao1991:
### Description

Split block/chunk logic.


execution/executor-types/src/executed_block.rs
```diff
@@ -0,0 +1,94 @@
...
+                || self.to_commit.last().map_or(true, |txn| matches!(
+                    txn.transaction(),
+                    Transaction::StateCheckpoint(_)
+                )),
+            "Block not ending with a state checkpoint.",
+        );
+        Ok(())
+    }
+
+    pub fn as_state_compute_result(
```
msmouse, line 76: same thing on ExecutedChunk can be removed

msmouse (approved):

execution/executor-types/src/executed_block.rs
grao1991: Done

areshand (approved):

PR #8087, opened 2023-05-06
[doc] Fix the folder layout of the example in the section Using Bytecode for Dependencies

rahxephon89:
### Description

This commit fixes a doc error in the bytecode dependency example and adds a clarification for requiring an empty `sources` folder in the Move package folder.


davidiw (approved):
thanks!

PR #8090, opened 2023-05-08
[framework] Partial governance voting support for delegation pool

xindingw:
### Description
Delegators in a delegation pool can attend Aptos governance. 
For a delegator:
- The delegator could delegate its voting power(the amount equals to its stake) to a voter.
- By default, a delegator's voter is itself.
- If a delegator changes its voter, it won't take effect until the current lockup period ends.

For a voter:

- The voting power of a voter equals to the sum of voting power of all delegators who delegate to the voter.

$$ VotingPower_{voter} = \Sigma_{d \in voter's\ delegators} VotingPower_d$$

- The voter can use its voting power to vote on a proposal.
- If the voting power of a voter meets the minimum threshold of creating a proposal, the voter can create a proposal.

#### Pre-knowledge
- A stake pool can vote on a proposal only when the lockup period of this stake pool ends after the proposal expiration. 
- Delegation pool's lockup period cannot be reset manually. So if the lockup period of a delegation pool ends, all proposals which this delegation pool voted during the lockup period have already expired.

#### High-level ideas
- A delegator's voting power equals to its coins in `active shares pool` + its coins in `pending inactive shares pool`.
- Due to rewards, the amount of coins keeps changing. But the shares of a delegator don't change until the next transaction of this delegator. The amount of coins can be calculated using shares.
- To track a voter's voting power, its total shares in `active shares pool` and  `pending inactive shares pool` need to be tracked. Whenever a delegator's voting power changes, its voter's voting power is also updated.
- Because changing voter takes 1 lockup period to take effects, the total shares of the voter at the next lockup period also need to be tracked.
- Pending inactive shares will become inactive at the next lockup period. So no need to track pending inactive shares at next lockup.

Let's say delegator D's voter is X at this lockup period and its voter will be changed to Y at the next lockup period. To summarize impacts on voter due to A's operations:
| D's change | X's active shares at this lockup | X's pending_active shares at this lockup| Y's active shares at next lockup | Z's active shares at next lockup|
| :--------: | :-------: | :-------: | :-------: | :-------: |
| + A active shares  | +A |  | +A | |
| + A pending inactive shares  |  | +A  |  | |
| - A active shares  |  -A |  | -A | |
| - A pending inactive shares  |  | -A |  | |
| Change voter at next lockup from Y to Z| | | -D's shares | +D's shares |

### <samp>🤖 Generated by Copilot at 4d58991</samp>

This pull request implements partial governance voting, a feature that allows voters to vote with a fraction of their delegated stake, on the Aptos blockchain. It adds and modifies spec functions, transaction payload functions, and entry function variants in the `aptos_governance` and `aptos_framework_sdk_builder` modules. It also adds a new module `aptos_governance` and a new function `increase_lockup` to the `e2e-move-tests` folder, and creates a new module `vote` with several test cases and scripts to test the partial voting feature.

### Test Plan

Unit testing
delegation_pool.move: % Move Coverage: 93.49

### DIFF
--- a/aptos-move/aptos-release-builder/src/components/feature_flags.rs
+++ b/aptos-move/aptos-release-builder/src/components/feature_flags.rs
@@ -9,7 +9,9 @@ use serde::{Deserialize, Serialize};
 
 #[derive(Clone, Deserialize, PartialEq, Eq, Serialize, Debug)]
 pub struct Features {
+    #[serde(default)]
     pub enabled: Vec<FeatureFlag>,
+    #[serde(default)]
     pub disabled: Vec<FeatureFlag>,
 }
 
--- a/aptos-move/aptos-release-builder/src/components/feature_flags.rs
+++ b/aptos-move/aptos-release-builder/src/components/feature_flags.rs
@@ -9,9 +9,7 @@ use serde::{Deserialize, Serialize};
 
 #[derive(Clone, Deserialize, PartialEq, Eq, Serialize, Debug)]
 pub struct Features {
-    #[serde(default)]
     pub enabled: Vec<FeatureFlag>,
-    #[serde(default)]
     pub disabled: Vec<FeatureFlag>,
 }
 
@@ -39,6 +37,7 @@ pub enum FeatureFlag {
     SignatureCheckerV2,
     StorageSlotMetadata,
     ChargeInvariantViolation,
+    DelegationPoolPartialGovernanceVoting,
 }
 
 fn generate_features_blob(writer: &CodeWriter, data: &[u64]) {
@@ -156,6 +155,9 @@ impl From<FeatureFlag> for AptosFeatureFlag {
             FeatureFlag::SignatureCheckerV2 => AptosFeatureFlag::SIGNATURE_CHECKER_V2,
             FeatureFlag::StorageSlotMetadata => AptosFeatureFlag::STORAGE_SLOT_METADATA,
             FeatureFlag::ChargeInvariantViolation => AptosFeatureFlag::CHARGE_INVARIANT_VIOLATION,
+            FeatureFlag::DelegationPoolPartialGovernanceVoting => {
+                AptosFeatureFlag::DELEGATION_POOL_PARTIAL_GOVERNANCE_VOTING
+            },
         }
     }
 }
@@ -196,6 +198,9 @@ impl From<AptosFeatureFlag> for FeatureFlag {
             AptosFeatureFlag::SIGNATURE_CHECKER_V2 => FeatureFlag::SignatureCheckerV2,
             AptosFeatureFlag::STORAGE_SLOT_METADATA => FeatureFlag::StorageSlotMetadata,
             AptosFeatureFlag::CHARGE_INVARIANT_VIOLATION => FeatureFlag::ChargeInvariantViolation,
+            AptosFeatureFlag::DELEGATION_POOL_PARTIAL_GOVERNANCE_VOTING => {
+                FeatureFlag::DelegationPoolPartialGovernanceVoting
+            },
         }
     }
 }
--- a/aptos-move/framework/aptos-framework/sources/aptos_governance.move
+++ b/aptos-move/framework/aptos-framework/sources/aptos_governance.move
@@ -304,6 +304,21 @@ module aptos_framework::aptos_governance {
         metadata_hash: vector<u8>,
         is_multi_step_proposal: bool,
     ) acquires GovernanceConfig, GovernanceEvents {
+        create_proposal_v2_impl(proposer, stake_pool, execution_hash, metadata_location, metadata_hash, is_multi_step_proposal);
+    }
+
+    /// Create a single-step or multi-step proposal with the backing `stake_pool`.
+    /// @param execution_hash Required. This is the hash of the resolution script. When the proposal is resolved,
+    /// only the exact script with matching hash can be successfully executed.
+    /// Return proposal_id when a proposal is sucessfully created.
+    public fun create_proposal_v2_impl(
+        proposer: &signer,
+        stake_pool: address,
+        execution_hash: vector<u8>,
+        metadata_location: vector<u8>,
+        metadata_hash: vector<u8>,
+        is_multi_step_proposal: bool,
+    ): u64 acquires GovernanceConfig, GovernanceEvents {
         let proposer_address = signer::address_of(proposer);
         assert!(stake::get_delegated_voter(stake_pool) == proposer_address, error::invalid_argument(ENOT_DELEGATED_VOTER));
 
@@ -361,6 +376,7 @@ module aptos_framework::aptos_governance {
                 proposal_metadata,
             },
         );
+        proposal_id
     }
 
     /// Vote on proposal with `proposal_id` and all voting power from `stake_pool`.
@@ -1122,6 +1138,16 @@ module aptos_framework::aptos_governance {
 
         let approved_hashes = borrow_global<ApprovedExecutionHashes>(@aptos_framework).hashes;
         assert!(*simple_map::borrow(&approved_hashes, &0) == vector[10u8,], 1);
+    }
+
+    #[test_only]
+    public fun initialize_for_test(
+        aptos_framework: &signer,
+        min_voting_threshold: u128,
+        required_proposer_stake: u64,
+        voting_duration_secs: u64,
+    ) {
+        initialize(aptos_framework, min_voting_threshold, required_proposer_stake, voting_duration_secs);
     }
 
     #[verify_only]
--- a/aptos-move/framework/aptos-framework/sources/aptos_governance.move
+++ b/aptos-move/framework/aptos-framework/sources/aptos_governance.move
@@ -7,8 +7,9 @@
 /// 1. Proposers can create a proposal by calling AptosGovernance::create_proposal. The proposer's backing stake pool
 /// needs to have the minimum proposer stake required. Off-chain components can subscribe to CreateProposalEvent to
 /// track proposal creation and proposal ids.
-/// 2. Voters can vote on a proposal. Their voting power is derived from the backing stake pool. Each stake pool can
-/// only be used to vote on each proposal exactly once.
+/// 2. Voters can vote on a proposal. Their voting power is derived from the backing stake pool. A stake pool can vote
+/// on a proposal multiple times as long as the total voting power of these votes don't exceed its total voting power.
+///
 ///
 module aptos_framework::aptos_governance {
     use std::error;
@@ -16,6 +17,7 @@ module aptos_framework::aptos_governance {
     use std::signer;
     use std::string::{Self, String, utf8};
     use std::vector;
+    use std::features;
 
     use aptos_std::simple_map::{Self, SimpleMap};
     use aptos_std::table::{Self, Table};
@@ -52,9 +54,17 @@ module aptos_framework::aptos_governance {
     const EMETADATA_HASH_TOO_LONG: u64 = 10;
     /// Account is not authorized to call this function.
     const EUNAUTHORIZED: u64 = 11;
+    /// The stake pool is using voting power more than it has.
+    const EVOTING_POWER_OVERFLOW: u64 = 12;
+    /// Partial voting feature hasn't been properly initialized.
+    const EPARTIAL_VOTING_NOT_INITIALIZED: u64 = 13;
+    /// The proposal in the argument is not a partial voting proposal.
+    const ENOT_PARTIAL_VOTING_PROPOSAL: u64 = 14;
 
     /// This matches the same enum const in voting. We have to duplicate it as Move doesn't have support for enums yet.
     const PROPOSAL_STATE_SUCCEEDED: u64 = 1;
+
+    const MAX_U64: u64 = 18446744073709551615;
 
     /// Proposal metadata attribute keys.
     const METADATA_LOCATION_KEY: vector<u8> = b"metadata_location";
@@ -81,6 +91,11 @@ module aptos_framework::aptos_governance {
     /// Records to track the proposals each stake pool has been used to vote on.
     struct VotingRecords has key {
         votes: Table<RecordKey, bool>
+    }
+
+    /// Records to track the voting power usage of each stake pool on each proposal.
+    struct VotingRecordsV2 has key {
+        votes: Table<RecordKey, u64>
     }
 
     /// Used to track which execution script hashes have been approved by governance.
@@ -195,6 +210,18 @@ module aptos_framework::aptos_governance {
         );
     }
 
+    /// Initializes the state for Aptos Governance partial voting. Can only be called through Aptos governance
+    /// proposals with a signer for the aptos_framework (0x1) account.
+    public fun initialize_partial_voting(
+        aptos_framework: &signer,
+    ) {
+        system_addresses::assert_aptos_framework(aptos_framework);
+
+        move_to(aptos_framework, VotingRecordsV2 {
+            votes: table::new(),
+        });
+    }
+
     #[view]
     public fun get_voting_duration_secs(): u64 acquires GovernanceConfig {
         borrow_global<GovernanceConfig>(@aptos_framework).voting_duration_secs
@@ -208,6 +235,49 @@ module aptos_framework::aptos_governance {
     #[view]
     public fun get_required_proposer_stake(): u64 acquires GovernanceConfig {
         borrow_global<GovernanceConfig>(@aptos_framework).required_proposer_stake
+    }
+
+    #[view]
+    /// Return true if a stake pool has already voted on a proposal before partial governance voting is enabled.
+    public fun has_entirely_voted(stake_pool: address, proposal_id: u64): bool acquires VotingRecords {
+        let record_key = RecordKey {
+            stake_pool,
+            proposal_id,
+        };
+        // If a stake pool has already voted on a proposal before partial governance voting is enabled,
+        // there is a record in VotingRecords.
+        let voting_records = borrow_global<VotingRecords>(@aptos_framework);
+        table::contains(&voting_records.votes, record_key)
+    }
+
+    #[view]
+    /// Return remaining voting power of a stake pool on a proposal.
+    /// Note: a stake pool's voting power on a proposal could increase overtime(e.g. rewards/new stake).
+    public fun get_remaining_voting_power(stake_pool: address, proposal_id: u64): u64 acquires VotingRecords, VotingRecordsV2 {
+        assert_voting_initialization();
+
+        // The voter's stake needs to be locked up at least as long as the proposal's expiration.
+        let proposal_expiration = voting::get_proposal_expiration_secs<GovernanceProposal>(@aptos_framework, proposal_id);
+        let lockup_until = stake::get_lockup_secs(stake_pool);
+        if (proposal_expiration > lockup_until) {
+            return 0
+        };
+
+        // If a stake pool has already voted on a proposal before partial governance voting is enabled, the stake pool
+        // cannot vote on the proposal even after partial governance voting is enabled.
+        if (has_entirely_voted(stake_pool, proposal_id)) {
+            return 0
+        };
+        let record_key = RecordKey {
+            stake_pool,
+            proposal_id,
+        };
+        let used_voting_power = 0u64;
+        if (features::partial_governance_voting_enabled()) {
+            let voting_records_v2 = borrow_global<VotingRecordsV2>(@aptos_framework);
+            used_voting_power = *table::borrow_with_default(&voting_records_v2.votes, record_key, &0);
+        };
+        get_voting_power(stake_pool) - used_voting_power
     }
 
     /// Create a single-step proposal with the backing `stake_pool`.
@@ -293,30 +363,40 @@ module aptos_framework::aptos_governance {
         );
     }
 
-    /// Vote on proposal with `proposal_id` and voting power from `stake_pool`.
+    /// Vote on proposal with `proposal_id` and all voting power from `stake_pool`.
     public entry fun vote(
         voter: &signer,
         stake_pool: address,
         proposal_id: u64,
         should_pass: bool,
-    ) acquires ApprovedExecutionHashes, GovernanceEvents, VotingRecords {
+    ) acquires ApprovedExecutionHashes, GovernanceEvents, VotingRecords, VotingRecordsV2 {
+        vote_internal(voter, stake_pool, proposal_id, MAX_U64, should_pass);
+    }
+
+    /// Vote on proposal with `proposal_id` and specified voting power from `stake_pool`.
+    public entry fun partial_vote(
+        voter: &signer,
+        stake_pool: address,
+        proposal_id: u64,
+        voting_power: u64,
+        should_pass: bool,
+    ) acquires ApprovedExecutionHashes, GovernanceEvents, VotingRecords, VotingRecordsV2 {
+        vote_internal(voter, stake_pool, proposal_id, voting_power, should_pass);
+    }
+
+    /// Vote on proposal with `proposal_id` and specified voting_power from `stake_pool`.
+    /// If voting_power is more than all the left voting power of `stake_pool`, use all the left voting power.
+    /// If a stake pool has already voted on a proposal before partial governance voting is enabled, the stake pool
+    /// cannot vote on the proposal even after partial governance voting is enabled.
+    fun vote_internal(
+        voter: &signer,
+        stake_pool: address,
+        proposal_id: u64,
+        voting_power: u64,
+        should_pass: bool,
+    ) acquires ApprovedExecutionHashes, GovernanceEvents, VotingRecords, VotingRecordsV2 {
         let voter_address = signer::address_of(voter);
         assert!(stake::get_delegated_voter(stake_pool) == voter_address, error::invalid_argument(ENOT_DELEGATED_VOTER));
-
-        // Ensure the voter doesn't double vote with the same stake pool.
-        let voting_records = borrow_global_mut<VotingRecords>(@aptos_framework);
-        let record_key = RecordKey {
-            stake_pool,
-            proposal_id,
-        };
-        assert!(
-            !table::contains(&voting_records.votes, record_key),
-            error::invalid_argument(EALREADY_VOTED));
-        table::add(&mut voting_records.votes, record_key, true);
-
-        let voting_power = get_voting_power(stake_pool);
-        // Short-circuit if the voter has no voting power.
-        assert!(voting_power > 0, error::invalid_argument(ENO_VOTING_POWER));
 
         // The voter's stake needs to be locked up at least as long as the proposal's expiration.
         let proposal_expiration = voting::get_proposal_expiration_secs<GovernanceProposal>(@aptos_framework, proposal_id);
@@ -325,6 +405,15 @@ module aptos_framework::aptos_governance {
             error::invalid_argument(EINSUFFICIENT_STAKE_LOCKUP),
         );
 
+        // If a stake pool has already voted on a proposal before partial governance voting is enabled,
+        // `get_remaining_voting_power` returns 0.
+        let staking_pool_voting_power = get_remaining_voting_power(stake_pool, proposal_id);
+        if (staking_pool_voting_power < voting_power) {
+            voting_power = staking_pool_voting_power;
+        };
+        // Short-circuit if the voter has no voting power.
+        assert!(voting_power > 0, error::invalid_argument(ENO_VOTING_POWER));
+
         voting::vote<GovernanceProposal>(
             &governance_proposal::create_empty_proposal(),
             @aptos_framework,
@@ -332,6 +421,23 @@ module aptos_framework::aptos_governance {
             voting_power,
             should_pass,
         );
+
+        let record_key = RecordKey {
+            stake_pool,
+            proposal_id,
+        };
+        if (features::partial_governance_voting_enabled()) {
+            let voting_records_v2 = borrow_global_mut<VotingRecordsV2>(@aptos_framework);
+            let used_voting_power = table::borrow_mut_with_default(&mut voting_records_v2.votes, record_key, 0);
+            // This calculation should never overflow because the used voting cannot exceed the total voting power of this stake pool.
+            *used_voting_power = *used_voting_power + voting_power;
+        } else {
+            let voting_records = borrow_global_mut<VotingRecords>(@aptos_framework);
+            assert!(
+                !table::contains(&voting_records.votes, record_key),
+                error::invalid_argument(EALREADY_VOTED));
+            table::add(&mut voting_records.votes, record_key, true);
+        };
 
         let events = borrow_global_mut<GovernanceEvents>(@aptos_framework);
         event::emit_event<VoteEvent>(
@@ -460,15 +566,23 @@ module aptos_framework::aptos_governance {
         metadata
     }
 
+    fun assert_voting_initialization() {
+        if (features::partial_governance_voting_enabled()) {
+            assert!(exists<VotingRecordsV2>(@aptos_framework), error::invalid_state(EPARTIAL_VOTING_NOT_INITIALIZED));
+        };
+    }
+
     #[test_only]
-    public entry fun create_proposal_for_test(proposer: signer, multi_step:bool) acquires GovernanceConfig, GovernanceEvents {
+    public entry fun create_proposal_for_test(
+        proposer: &signer,
+        multi_step: bool,
+    ) acquires GovernanceConfig, GovernanceEvents {
         let execution_hash = vector::empty<u8>();
         vector::push_back(&mut execution_hash, 1);
-
         if (multi_step) {
             create_proposal_v2(
-                &proposer,
-                signer::address_of(&proposer),
+                proposer,
+                signer::address_of(proposer),
                 execution_hash,
                 b"",
                 b"",
@@ -476,8 +590,8 @@ module aptos_framework::aptos_governance {
             );
         } else {
             create_proposal(
-                &proposer,
-                signer::address_of(&proposer),
+                proposer,
+                signer::address_of(proposer),
                 execution_hash,
                 b"",
                 b"",
@@ -509,17 +623,26 @@ module aptos_framework::aptos_governance {
         no_voter: signer,
         multi_step: bool,
         use_generic_resolve_function: bool,
-    ) acquires ApprovedExecutionHashes, GovernanceConfig, GovernanceEvents, GovernanceResponsbility, VotingRecords {
+    ) acquires ApprovedExecutionHashes, GovernanceConfig, GovernanceEvents, GovernanceResponsbility, VotingRecords, VotingRecordsV2 {
         setup_voting(&aptos_framework, &proposer, &yes_voter, &no_voter);
 
         let execution_hash = vector::empty<u8>();
         vector::push_back(&mut execution_hash, 1);
 
-        create_proposal_for_test(proposer, multi_step);
+        create_proposal_for_test(&proposer, multi_step);
 
         vote(&yes_voter, signer::address_of(&yes_voter), 0, true);
         vote(&no_voter, signer::address_of(&no_voter), 0, false);
 
+        test_resolving_proposal_generic(aptos_framework, use_generic_resolve_function, execution_hash);
+    }
+
+    #[test_only]
+    public entry fun test_resolving_proposal_generic(
+        aptos_framework: signer,
+        use_generic_resolve_function: bool,
+        execution_hash: vector<u8>,
+    ) acquires ApprovedExecutionHashes, GovernanceResponsbility {
         // Once expiration time has passed, the proposal should be considered resolve now as there are more yes votes
         // than no.
         timestamp::update_global_time_for_test(100001000000);
@@ -545,7 +668,7 @@ module aptos_framework::aptos_governance {
         proposer: signer,
         yes_voter: signer,
         no_voter: signer,
-    ) acquires ApprovedExecutionHashes, GovernanceConfig, GovernanceEvents, GovernanceResponsbility, VotingRecords {
+    ) acquires ApprovedExecutionHashes, GovernanceConfig, GovernanceEvents, GovernanceResponsbility, VotingRecords, VotingRecordsV2 {
         test_voting_generic(aptos_framework, proposer, yes_voter, no_voter, false, false);
     }
 
@@ -555,7 +678,7 @@ module aptos_framework::aptos_governance {
         proposer: signer,
         yes_voter: signer,
         no_voter: signer,
-    ) acquires ApprovedExecutionHashes, GovernanceConfig, GovernanceEvents, GovernanceResponsbility, VotingRecords {
+    ) acquires ApprovedExecutionHashes, GovernanceConfig, GovernanceEvents, GovernanceResponsbility, VotingRecords, VotingRecordsV2 {
         test_voting_generic(aptos_framework, proposer, yes_voter, no_voter, true, true);
     }
 
@@ -566,7 +689,7 @@ module aptos_framework::aptos_governance {
         proposer: signer,
         yes_voter: signer,
         no_voter: signer,
-    ) acquires ApprovedExecutionHashes, GovernanceConfig, GovernanceEvents, GovernanceResponsbility, VotingRecords {
+    ) acquires ApprovedExecutionHashes, GovernanceConfig, GovernanceEvents, GovernanceResponsbility, VotingRecords, VotingRecordsV2 {
         test_voting_generic(aptos_framework, proposer, yes_voter, no_voter, true, false);
     }
 
@@ -576,7 +699,7 @@ module aptos_framework::aptos_governance {
         proposer: signer,
         yes_voter: signer,
         no_voter: signer,
-    ) acquires ApprovedExecutionHashes, GovernanceConfig, GovernanceEvents, GovernanceResponsbility, VotingRecords {
+    ) acquires ApprovedExecutionHashes, GovernanceConfig, GovernanceEvents, GovernanceResponsbility, VotingRecords, VotingRecordsV2 {
         test_voting_generic(aptos_framework, proposer, yes_voter, no_voter, false, true);
     }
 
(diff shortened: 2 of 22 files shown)


aptos-move/framework/aptos-framework/sources/delegation_pool.move
```diff
@@ -221,6 +243,48 @@ module aptos_framework::delegation_pool {
...
+        voter: address,
+        proposal_id: u64,
+    }
+
+    struct VotingPowerDelegation has copy, drop, store {
+        // The current voter address of this delegator.
+        voter: address,
+        // The change of voter address won't take effects until the current lockup period ends. Record the new voter
+        // address and update `voter` to `voter_next_lockup` when that lockup period has already ended.
+        voter_next_lockup: address,
```
movekevin, line 256: nit: How about pending_vote?

aptos-move/framework/aptos-framework/sources/delegation_pool.move
```diff
@@ -221,6 +243,48 @@ module aptos_framework::delegation_pool {
...
+        // locked_until_secs of the stake pool when last time this struct is updated.
+        // If <locked_until_secs of the stake pool> != last_locked_until_secs, it means that a lockup period has passed.
+        // This is slightly different from ObservedLockupCycle because ObservedLockupCycle cannot detect if a lockup
+        // period is passed when there is no unlocking during the lockup period .
+        last_locked_until_secs: u64,
```
movekevin, line 261: This would break though if we ever introduce the ability for delegation pools to reset their lockup as although the lockup period might not have passed, the delegation pool's new locked until secs would be higher than what's stored here. Were there any other alternatives that you considered here?

aptos-move/framework/aptos-framework/sources/delegation_pool.move
```diff
@@ -221,6 +243,48 @@ module aptos_framework::delegation_pool {
         distribute_commission_events: EventHandle<DistributeCommissionEvent>,
     }
 
+    struct VotingRecordKey has copy, drop, store {
+        voter: address,
+        proposal_id: u64,
+    }
+
+    struct VotingPowerDelegation has copy, drop, store {
```
movekevin, line 251: Can you add documentation in front of each struct here to explain what data it holds and where it's stored/used?

aptos-move/framework/aptos-framework/sources/delegation_pool.move
```diff
@@ -221,6 +243,48 @@ module aptos_framework::delegation_pool {
...
+        // address and update `voter` to `voter_next_lockup` when that lockup period has already ended.
+        voter_next_lockup: address,
+        // locked_until_secs of the stake pool when last time this struct is updated.
+        // If <locked_until_secs of the stake pool> != last_locked_until_secs, it means that a lockup period has passed.
+        // This is slightly different from ObservedLockupCycle because ObservedLockupCycle cannot detect if a lockup
+        // period is passed when there is no unlocking during the lockup period .
+        last_locked_until_secs: u64,
+    }
+
+    struct VoterVotingPower has copy, drop, store {
```
movekevin, line 264: nit: What do you think about "DelegatedVotes" to be more concise?

aptos-move/framework/aptos-framework/sources/delegation_pool.move
```diff
@@ -221,6 +243,48 @@ module aptos_framework::delegation_pool {
...
+        // Equals to the sum of active shares of delegators who delegate to this voter.
+        active_shares: u128,
+        // Equals to the sum of pending inactive shares of delegators who delegate to this voter.
+        pending_inactive_shares: u128,
+        // Equals to the sum of active shares of delegators who delegate to this voter after the current lockup period.
+        // `active_shares_next_lockup` might be different `active_shares` when some delegators change their voter.
```
movekevin, line 270: Hmm this comment is not very helpful. How about "Total active shares delegated to this voter in the next lockup cycle." I'd add more documentation regarding the fact the a voter's total delegated voting power doesn't change until the start of a new lockup cycle

aptos-move/framework/aptos-framework/sources/delegation_pool.move
```diff
@@ -221,6 +243,48 @@ module aptos_framework::delegation_pool {
...
+        // If <locked_until_secs of the stake pool> != last_locked_until_secs, it means that a lockup period has passed.
+        // This is slightly different from ObservedLockupCycle because ObservedLockupCycle cannot detect if a lockup
+        // period is passed when there is no unlocking during the lockup period .
+        last_locked_until_secs: u64,
+    }
+
+    struct GovernanceRecords has key {
+        // `votes` tracks voting power usage of each voter on each proposal.
+        votes: Table<VotingRecordKey, u64>,
+        voting_power_delegation: Table<address, VotingPowerDelegation>,
```
movekevin, line 282: We also need events for voting power delegation. Otherwise, the voter wouldn't know who delegates to them in which pool (really tough to index and find off chain)

aptos-move/framework/aptos-framework/sources/delegation_pool.move
```diff
@@ -221,6 +243,48 @@ module aptos_framework::delegation_pool {
...
+        voting_power_per_voter: Table<address, VoterVotingPower>,
```
movekevin, line 283: Let's use SmartTable instead of Table everywhere

aptos-move/framework/aptos-framework/sources/delegation_pool.move
```diff
@@ -272,6 +350,12 @@ module aptos_framework::delegation_pool {
         exists<DelegationPool>(addr)
     }
 
+    #[view]
+    /// Return whether a delegation pool has alrady enabled partial govnernance voting.
```
movekevin, line 354: typo: alrady -> already

aptos-move/framework/aptos-framework/sources/delegation_pool.move
```diff
@@ -429,6 +513,51 @@ module aptos_framework::delegation_pool {
             timestamp::now_seconds() >= stake::get_lockup_secs(pool_address)
     }
 
+    #[view]
+    /// Return the total voting power of a delegator in a delegation pool. This funciton synces DelegationPool to the
```
movekevin, line 517: typo: synces => syncs

aptos-move/framework/aptos-framework/sources/delegation_pool.move
```diff
@@ -429,6 +513,51 @@ module aptos_framework::delegation_pool {
...
+        synchronize_delegation_pool(pool_address);
+        let pool = borrow_global<DelegationPool>(pool_address);
+        let governance_records = borrow_global<GovernanceRecords>(pool_address);
+        let (latest_voter_voting_power, _, _) = get_latest_voter_voting_power(pool, governance_records, voter);
+        let delegation_pool = borrow_global<DelegationPool>(pool_address);
+        calculate_total_voting_power(delegation_pool, &latest_voter_voting_power)
+    }
+
+    #[view]
+    /// Return the remaining voting power of a delegator in a delegation pool on a proposal. This funciton synces DelegationPool to the
```
movekevin, line 532: typo: synces => syncs. funciton => function

aptos-move/framework/aptos-framework/sources/delegation_pool.move
```diff
@@ -554,15 +790,165 @@ module aptos_framework::delegation_pool {
...
         ObservedLockupCycle { index }
     }
 
+    /// Given the amounts of shares in `active_shares` pool and `inactive_shares` pool, calculate the total voting
+    /// power, which equals to the sum of the coin amounts.
+    fun calculate_total_voting_power(delegation_pool: &DelegationPool, latest_voter_voting_power: &VoterVotingPower): u64 {
+        let active_amount = pool_u64::shares_to_amount(
+            &delegation_pool.active_shares,
+            latest_voter_voting_power.active_shares);
+        let inactive_amount = pool_u64::shares_to_amount(
```
movekevin, line 929: nit: Name this pending_inactive instead

aptos-move/framework/aptos-framework/sources/delegation_pool.move
```diff
@@ -554,15 +790,165 @@ module aptos_framework::delegation_pool {
...
+            return (VoterVotingPower {
+                active_shares,
+                pending_inactive_shares: inactive_shares,
+                active_shares_next_lockup: active_shares,
+                last_locked_until_secs: locked_until_secs,
+            }, false, false)
+        };
+        let latest_voter_voting_power = *table::borrow(voting_power_per_voter, voter);
+        let needs_updated = false;
+        if (latest_voter_voting_power.last_locked_until_secs < locked_until_secs) {
```
movekevin, line 891: This doesn't take into account the voter's own voting power if they're also a delegator

aptos-move/framework/aptos-framework/sources/delegation_pool.move
```diff
@@ -2709,6 +3271,542 @@ module aptos_framework::delegation_pool {
...
+        assert!(stake::get_delegated_voter(pool_address) == pool_address, 1);
+        assert!(partial_governance_voting_enabled(pool_address), 1);
+
+        // By default, the voter of a delegator is itself.
+        assert!(calculate_and_update_voter_total_voting_power(pool_address, voter1_address) == 0, 1);
+        assert!(calculate_and_update_voter_total_voting_power(pool_address, voter2_address) == 0, 1);
+        assert!(calculate_and_update_voter_total_voting_power(pool_address, delegator1_address) == 10 * ONE_APT, 1);
+        assert!(calculate_and_update_voter_total_voting_power(pool_address, delegator2_address) == 90 * ONE_APT, 1);
+
+        // Delegator1 delegates its voting power to voter1. Delegator2 delegates its voting power to voter2.
```
movekevin, line 3562: Do we need two delegators for this test or can we keep it simple and just have one?

aptos-move/framework/aptos-framework/sources/delegation_pool.move
```diff
@@ -2709,6 +3271,542 @@ module aptos_framework::delegation_pool {
...
+        );
+        aptos_governance::initialize_partial_voting(aptos_framework);
+        features::change_feature_flags(aptos_framework, vector[features::get_partial_governance_voting()], vector[]);
+
+        // 50% commission rate
+        initialize_test_validator_custom(validator, 100 * ONE_APT, true, false, 5000);
+
+        let validator_address = signer::address_of(validator);
+        let pool_address = get_owned_pool_address(validator_address);
+        // Delegation pool is created after partial governance voting feature flag is enabled. So this delegation
```
movekevin, line 3613: Same here. Do we actually need two delegators?

aptos-move/framework/aptos-framework/sources/delegation_pool.move
```diff
@@ -481,6 +610,108 @@ module aptos_framework::delegation_pool {
...
+    ) acquires DelegationPool, GovernanceRecords {
+        assert!(features::partial_governance_voting_enabled(), error::invalid_state(EDISABLED_FUNCTION));
+        assert_delegation_pool_exists(pool_address);
+        let delegation_pool = borrow_global<DelegationPool>(pool_address);
+        let stake_pool_signer = retrieve_stake_pool_owner(delegation_pool);
+        synchronize_delegation_pool(pool_address);
+        // delegated_voter is managed by the stake pool itself.
+        stake::set_delegated_voter(&stake_pool_signer, signer::address_of(&stake_pool_signer));
+
+        if (!exists<GovernanceRecords>(pool_address)) {
```
movekevin, line 633: This doesn't seem necessary?

aptos-move/framework/aptos-framework/sources/delegation_pool.move
xindingw: I suppose you mean "pending_voter"?

aptos-move/framework/aptos-framework/sources/delegation_pool.move
xindingw: Actually here I want to test the case that multiple delegators delegate to the same voter and both of them receive rewards. Added the case.

aptos-move/framework/aptos-framework/sources/delegation_pool.move
xindingw: If we support resetting the lockup period , the time of voter change needs to be postponed to the new `locked_until_secs`. Otherwise let's consider a case: if A delegates to X then changes it to Y, X is A's voter before before the old `locked_until_secs` then Y is A's voter after the old `locked_until_secs`. X could vote on proposal 1 before the old `locked_until_secs` then Y could vote on proposal 1 after the old `locked_until_secs`. A's voting power will be used twice.
To implement this, we will need to record the start and the end of a lockup period. Let's say the start/end is a/b at the moment. If `last_locked_until_secs` is within [a, b], we can know the lockup period for that record hasn't ended yet. 

To maintain [a, b], we need:
- If the current lockup period ends normally and the new `locked_until_secs` is c. The start/end becomes [b, c].
- If the delegation pool resets its lockup period and the new `locked_until_secs` is c. The start/end becomes [a, c].

The solution could be plugged into the module when we want to support resetting. Do you think if we should add [a, b] now?

aptos-move/framework/aptos-framework/sources/delegation_pool.move
xindingw: I see, just let it abort if GovernanceRecords exists. Make sense. Removed this condition.

xindingw:
@movekevin Think more about moving `GovernanceRecords` out of `delegation_pool` module. The main problem is that `active_shares` and `inactive_shares` are needed when populating default values of `DelegatedVotes` but the values are stored inside `DelegationPool`(we cannot call `delegation_pool` module otherwise there is cyclic depedency). There could be 2 ways for the problem:
1. When get `DelegatedVotes` of a voter, always pass its `active_shares` and `inactive_shares` as well. Semantically it's weird and ugly.
2. `DelegatedVotes` doesn't count the voter itself's voting power. Add the voter power back in `delegation_pool` module if a delegator's voter is itself. But this makes `DelegatedVotes` more counterintuitive. 

TBH I think moving `GovernanceRecords` out can reduce the size of `delegation_pool.move`(it has ~4000 lines now 😞  ) but it won't make the logic cleaner/easier to understand because it's so coupled with concepts in `delegation_pool.move`. Personally I prefer leave `GovernanceRecords` there and add more documentation to explain the logic. This is pretty tricky so let me know WDYT.

aptos-move/framework/aptos-framework/sources/delegation_pool.move
movekevin: Yes

aptos-move/framework/aptos-framework/sources/delegation_pool.move
movekevin: That sounds rather complex. Is there no other alternative?

aptos-move/framework/aptos-framework/sources/delegation_pool.move
```diff
@@ -481,6 +626,118 @@ module aptos_framework::delegation_pool {
...
+
+        let voter_address = signer::address_of(voter);
+        let remaining_voting_power = calculate_and_update_remaining_voting_power(pool_address, voter_address, proposal_id);
+        if (voting_power > remaining_voting_power) {
+            voting_power = remaining_voting_power;
+        };
+        assert!(voting_power > 0, error::invalid_argument(ENO_VOTING_POWER));
+
+        let governance_records = borrow_global_mut<GovernanceRecords>(pool_address);
+        let stake_pool_remaining_voting_power = aptos_governance::get_remaining_voting_power(pool_address, proposal_id);
```
movekevin, line 674: I'd move this code from lines 675-682 into a separate (may be inline?) function so it's more readable. Furthermore, this can be removed later in the release after we enable delegation voting

movekevin (changes requested):
Looks correct to me high-level wise, just some minor comments and one important high-level comment: Let's create a separate feature flag for enabling delegation voting

aptos-move/framework/aptos-framework/sources/delegation_pool.move
```diff
@@ -481,6 +626,118 @@ module aptos_framework::delegation_pool {
...
+                voter: voter_address,
+                proposal_id,
+                delegation_pool: pool_address,
+                num_votes: voting_power,
+                should_pass,
+            }
+        );
+    }
+
+    public entry fun create_proposal(
```
movekevin, line 702: Can you add docs for all public functions? May be something concise if there's not a lot to explain (i.e. the function is self-explaining)

aptos-move/framework/aptos-framework/sources/delegation_pool.move
```diff
@@ -554,15 +816,165 @@ module aptos_framework::delegation_pool {
...
+        (vote_delegation, true, needs_update)
+    }
+
+    /// Update VoteDelegation of a delegator to up-to-date then borrow_mut it.
+    fun update_and_borrow_mut_delegator_vote_delegation(
+        pool: &DelegationPool,
+        governance_records :&mut GovernanceRecords,
+        delegator: address
+    ): &mut VoteDelegation {
+        let (latest_vote_delegation, is_existed, needs_update) =
```
movekevin, line 880: This looks funky to me because:
1. get_latest_delegator_vote_delegation is only ever called here in this function
2. get_latest_delegator_vote_delegation borrows multiple, makes a copy of the VotingDelegation struct, and returns it here. The struct is then modified further here

I think you can just merge the logic in get_latest_delegator_vote_delegation here.

aptos-move/framework/aptos-framework/sources/delegation_pool.move
```diff
@@ -554,15 +816,165 @@ module aptos_framework::delegation_pool {
...
+            get_latest_delegator_vote_delegation(pool, governance_records, delegator);
+        let vote_delegation_table = &mut governance_records.vote_delegation;
+        if (!is_existed) {
+            smart_table::add(vote_delegation_table, delegator, latest_vote_delegation);
```
movekevin, line 884: Wouldn't this add an entry for the delegator themselves if they don't delegate to anyone? Not a big deal logic-wise, but this can be storage inefficient as VoteDelegation would exist even if the delegator doesn't delegate to anyone

aptos-move/framework/aptos-framework/sources/delegation_pool.move
```diff
@@ -554,15 +816,165 @@ module aptos_framework::delegation_pool {
...
+        };
+        let vote_delegation = smart_table::borrow_mut(vote_delegation_table, delegator);
+        if (needs_update) {
```
movekevin, line 887: This logic would not be needed if we can merge this function with get_latest_delegator_vote_delegation.

aptos-move/framework/aptos-framework/sources/delegation_pool.move
```diff
@@ -554,15 +816,165 @@ module aptos_framework::delegation_pool {
...
+            *vote_delegation = latest_vote_delegation;
+        };
+        vote_delegation
+    }
+
+    /// Get the latest VoteDelegation of a voter.
+    /// The 2nd bool return value is true when VoteDelegation has already been created before calling this function.
+    /// The 3rd bool return value is true when the latest VoteDelegation is different from the existing VoteDelegation.
+    fun get_latest_vote_delegation(
```
movekevin, line 896: This looks so similar to get_latest_delegator_vote_delegation that I had to check twice that it's not the same function. Can we merge the 2 somehow?

aptos-move/framework/aptos-framework/sources/delegation_pool.move
```diff
@@ -554,15 +816,165 @@ module aptos_framework::delegation_pool {
...
+        };
+        (latest_vote_delegation, true, needs_updated)
+    }
+
+    /// Update VoterVotingPower of a voter to up-to-date then borrow_mut it.
+    fun update_and_borrow_mut_vote_delegation(
+        pool: &DelegationPool,
+        governance_records :&mut GovernanceRecords,
+        voter: address
+    ): &mut DelegatedVotes {
```
movekevin, line 932: This get latest and update logic is a bit janky similar to update_and_borrow_mut_delegator_vote_delegation. Anything we can do here to simplify?

aptos-move/framework/aptos-framework/sources/delegation_pool.move
```diff
@@ -429,6 +529,51 @@ module aptos_framework::delegation_pool {
             timestamp::now_seconds() >= stake::get_lockup_secs(pool_address)
     }
 
+    #[view]
+    /// Return the total voting power of a delegator in a delegation pool. This function syncs DelegationPool to the
+    /// latest state.
+    public fun calculate_and_update_voter_total_voting_power(pool_address: address, voter: address): u64 acquires DelegationPool, GovernanceRecords {
```
movekevin, line 535: It's strange that this is a view function when it can update (has side effects)

aptos-move/framework/aptos-framework/sources/delegation_pool.move
```diff
@@ -429,6 +529,51 @@ module aptos_framework::delegation_pool {
...
+        let governance_records = borrow_global<GovernanceRecords>(pool_address);
+        let (latest_vote_delegation, _, _) = get_latest_vote_delegation(pool, governance_records, voter);
+        let delegation_pool = borrow_global<DelegationPool>(pool_address);
+        calculate_total_voting_power(delegation_pool, &latest_vote_delegation)
+    }
+
+    #[view]
+    /// Return the remaining voting power of a delegator in a delegation pool on a proposal. This function syncs DelegationPool to the
+    /// latest state.
+    public fun calculate_and_update_remaining_voting_power(pool_address: address, voter_address: address, proposal_id: u64): u64 acquires DelegationPool, GovernanceRecords {
```
movekevin, line 550: Same here - view functions that can change state are strange

aptos-move/framework/aptos-framework/sources/delegation_pool.move
```diff
@@ -574,15 +986,74 @@ module aptos_framework::delegation_pool {
...
+        new_voter: address
+    ) acquires DelegationPool, GovernanceRecords {
+        assert_partial_governance_voting_enabled(pool_address);
+
+        // synchronize delegation and stake pools before any user operation
+        synchronize_delegation_pool(pool_address);
+
+        let delegator_address = signer::address_of(delegator);
+        let (active, inactive, pending_inactive) = get_stake(pool_address, delegator_address);
+        assert!(active + inactive + pending_inactive > 0, error::invalid_argument(EDELEGATOR_NOT_EXISTS));
```
alexfilip2, line 1012: is this check required? a delegator that never participated in the delegation pool might be a dApp calling this at its initialization for example

aptos-move/framework/aptos-framework/sources/delegation_pool.move
alexfilip2: As I understand, delegated voting power is delayed for a new voter to the next lockup cycle in order to ensure the proposal expires and double voting cannot happen if original voter already voted.
If lockup time can be refreshed on StakePool then each DelegatedVotes object should somehow detect that and update its `last_locked_until_secs`
However, the observed lockup cycle can be forced to be in sync with the real lockup cycle on StakePool. At `delegate_voting_power`:
1. Sync pool
2. Redeem shares worth 1 octa from caller if the resource account doesn't have any shares in the pending inactive pool
3. Buy shares for resource account for the 1 extracted octa, it will get 1 shares back (10 ^ (scaling factor set on pool_u64) shares really)
4 Resource account cannot reactivate this stake
5 Lockup ends on Stakepool, it inactivates 1 octa which is detected at next synchronization on delegation pool
This ensures OLC advances each lockup when someone changed their delegated voter and this change is guaranteed to be applied. So `last_locked_until_secs` can be replaced by a OLC index instead and not be affected by lockup expiration refreshing.
There will always be one shares pool created to host only 1 octa as buying pending-inactive shares for the resource account triggers the withdrawal of its inactive stake which empties the former shares pool which, in turn, gets deleted.

aptos-move/framework/aptos-framework/sources/delegation_pool.move
```diff
@@ -554,15 +816,165 @@ module aptos_framework::delegation_pool {
...
+                active_shares_next_lockup: active_shares,
+                last_locked_until_secs: locked_until_secs,
+            }, false, false)
+        };
+        let latest_vote_delegation = *smart_table::borrow(voting_power_per_voter, voter);
+        let needs_updated = false;
+        if (latest_vote_delegation.last_locked_until_secs < locked_until_secs) {
+            latest_vote_delegation.active_shares = latest_vote_delegation.active_shares_next_lockup;
+            latest_vote_delegation.pending_inactive_shares = 0;
+            latest_vote_delegation.active_shares_next_lockup = latest_vote_delegation.active_shares;
```
alexfilip2, line 920: I think this line is not needed

aptos-move/framework/aptos-framework/sources/delegation_pool.move
```diff
@@ -862,11 +1415,27 @@ module aptos_framework::delegation_pool {
...
+        // Always update governance records before any change to the shares pool.
+        let pool_address = get_pool_address(pool);
+        // <pending inactive shares> of <shareholder> -= <shares_to_redeem>  ---->
+        // <pending inactive shares> of <current voter of shareholder> -= <shares_to_redeem>
+        // no impact on <pending inactive shares> of <next voter of shareholder>
+        if (partial_governance_voting_enabled(pool_address)) {
+            let governance_records = borrow_global_mut<GovernanceRecords>(pool_address);
+            let current_voter = calculate_and_update_delegator_voter_internal(pool, governance_records, shareholder);
+            let current_vote_delegation = update_and_borrow_mut_vote_delegation(pool, governance_records, current_voter);
+            current_vote_delegation.pending_inactive_shares = current_vote_delegation.pending_inactive_shares - shares_to_redeem;
```
alexfilip2, line 1435: what if redeeming inactive shares, not pending-inactive ones? pending-inactive voting shares should be decreased only if redeeming pending-inactive shares

aptos-move/framework/aptos-framework/sources/delegation_pool.move
```diff
@@ -799,12 +1309,26 @@ module aptos_framework::delegation_pool {
         pool: &mut DelegationPool,
         shareholder: address,
         coins_amount: u64,
-    ): u128 {
-        // cannot buy inactive shares, only pending_inactive at current lockup cycle
-        let new_shares = pool_u64::buy_in(pending_inactive_shares_pool_mut(pool), shareholder, coins_amount);
+    ): u128 acquires GovernanceRecords {
+        let new_shares = pool_u64::amount_to_shares(pending_inactive_shares_pool_mut(pool), coins_amount);
```
alexfilip2, line 1313: `pending_inactive_shares_pool` instead of  `pending_inactive_shares_pool_mut`

aptos-move/framework/aptos-framework/sources/delegation_pool.move
```diff
@@ -784,13 +1255,52 @@ module aptos_framework::delegation_pool {
...
+        let new_shares = pool_u64::amount_to_shares(&pool.active_shares, coins_amount);
+        // No need to buy 0 shares.
+        if (new_shares == 0) { return 0 };
+
+        // Always update governance records before any change to the shares pool.
+        let pool_address = get_pool_address(pool);
+        // <active shares> of <shareholder> += <new_shares> ---->
+        // <active shares> of <current voter of shareholder> += <new_shares>
+        // <active shares> of <next voter of shareholder> += <new_shares>
+        if (partial_governance_voting_enabled(pool_address)) {
```
alexfilip2, line 1281: can this logic be enclosed in a separate function, possibly supporting subtraction too (in case of redeeming active shares)? It could be called here for addition and in `redeem_active_shares` with subtraction of voting shares.

alexfilip2: the same for the buy/redeem of pending-inactive voting shares

aptos-move/framework/aptos-framework/sources/delegation_pool.move
xindingw: @alexfilip2  Hmm I'm not sure I understand. Few questions:
1. What is octa?
2. What does "resource account" refer? NULL_SHAREHOLDER or the stake pool resource accounts?

aptos-move/framework/aptos-framework/sources/delegation_pool.move
xindingw: Make sense. `get_latest_delegator_vote_delegation` was used in a view function so I want to keep it read-only. But the view function has to sync first. Will merge the logic.

aptos-move/framework/aptos-framework/sources/delegation_pool.move
xindingw: Yeah make sense

aptos-move/framework/aptos-framework/sources/delegation_pool.move
xindingw: My concern was that people might create tons of addresses and call this function. But maybe not a vulnerability.

aptos-move/framework/aptos-framework/sources/delegation_pool.move
```diff
@@ -272,6 +366,12 @@ module aptos_framework::delegation_pool {
         exists<DelegationPool>(addr)
     }
 
+    #[view]
+    /// Return whether a delegation pool has already enabled partial govnernance voting.
+    public fun partial_governance_voting_enabled(pool_address: address): bool {
+        exists<GovernanceRecords>(pool_address) && stake::get_delegated_voter(pool_address) == pool_address
```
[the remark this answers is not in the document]
xindingw: In theory no. Just want to be more defensive when checking. The best way is to check this in prover spec but the prover spec is not ready yet.

aptos-move/framework/aptos-framework/sources/delegation_pool.move
alexfilip2: 1. One unit of APT token - 10^-8
2. The resource account storing the DelegationPool, GovernanceRecords etc. It could be the NULL_SHAREHOLDER too, basically any account that cannot be accessed directly and reactivate stake

aptos-move/framework/aptos-framework/sources/delegation_pool.move
xindingw: Good caught! Added a checking on OLC.

aptos-move/framework/aptos-framework/sources/delegation_pool.move
xindingw: Considering the complexity and the recycle logic could be plugged in at any time, maybe we can hold it for now? Added some TODOs.

aptos-move/framework/aptos-framework/sources/delegation_pool.move
xindingw: Think more about it. 1 octa from caller might be problematic. There could be some edge cases:
1. the caller doesn't have any stake.
2. 1 octa might make the caller below `MIN_COINS_ON_SHARES_POOL`.

alexfilip2: after some more thought, I think that 1 octa doesn't have to come from the caller. Step 2 can be replaced by stake::unlock(1) - you get 1 coin from the active shares pool and then move to step 3 to buy pending-inactive shares for the resource account for 1 octa.

alexfilip2: Or: if a delegator wants to change its delegated voter, they should unlock some stake (the minimum possible one). If they reactivate their stake before the lockup expires then OLC won't advance and the new voter remains pending for another lockup cycle at least. This is less UX efficient, but it requires no complex logic for migrating to OLC indices instead of lockup timestamps

xindingw: @alexfilip2 The accounting model has already been very complicated so I prefer not adding more complexity. As long as we change behavior of the staking part, it brings lots of implication and edge cases. e.g. all delegators exit from a delegation pool/ maintaining `MIN_COINS_ON_SHARES_POOL`/ how flags can gate the logic. I think we should keep the governance part as standalone as possible.

aptos-move/aptos-release-builder/src/components/feature_flags.rs
```diff
@@ -9,9 +9,7 @@ use serde::{Deserialize, Serialize};
 
 #[derive(Clone, Deserialize, PartialEq, Eq, Serialize, Debug)]
 pub struct Features {
-    #[serde(default)]
```
movekevin, line 12: Why delete this?

xindingw: Good catch! This should be due to rebase. Fixed.

aptos-move/framework/aptos-framework/sources/aptos_governance.move
```diff
@@ -210,6 +237,49 @@ module aptos_framework::aptos_governance {
...
+    #[view]
+    /// Return remaining voting power of a stake pool on a proposal.
+    /// Note: a stake pool's voting power on a proposal could increase overtime(e.g. rewards/new stake).
+    public fun get_remaining_voting_power(stake_pool: address, proposal_id: u64): u64 acquires VotingRecords, VotingRecordsV2 {
+        assert_voting_initialization();
+
+        // The voter's stake needs to be locked up at least as long as the proposal's expiration.
+        let proposal_expiration = voting::get_proposal_expiration_secs<GovernanceProposal>(@aptos_framework, proposal_id);
+        let lockup_until = stake::get_lockup_secs(stake_pool);
+        if (proposal_expiration > lockup_until) {
```
lightmark: When this can happen? Does this mean the stake_pool can only vote when its lockup_time > proposal_expiration?
n00b: how to make sure nobody can vote for an expired proposal?

aptos-move/framework/aptos-framework/sources/aptos_governance.move
```diff
@@ -325,6 +405,15 @@ module aptos_framework::aptos_governance {
             error::invalid_argument(EINSUFFICIENT_STAKE_LOCKUP),
         );
 
+        // If a stake pool has already voted on a proposal before partial governance voting is enabled,
+        // `get_remaining_voting_power` returns 0.
+        let staking_pool_voting_power = get_remaining_voting_power(stake_pool, proposal_id);
+        if (staking_pool_voting_power < voting_power) {
```
lightmark, line 426: use `max()`

aptos-move/framework/aptos-framework/sources/aptos_governance.move
```diff
@@ -83,6 +93,11 @@ module aptos_framework::aptos_governance {
         votes: Table<RecordKey, bool>
     }
 
+    /// Records to track the voting power usage of each stake pool on each proposal.
+    struct VotingRecordsV2 has key {
+        votes: Table<RecordKey, u64>
```
lightmark, line 96: how do we clean-up records of finished proposals?

aptos-move/framework/aptos-framework/sources/aptos_governance.move
xindingw: Yes a stake pool can only vote([ref](https://github.com/aptos-labs/aptos-core/blob/e5a981cdf48c8bcac65a504b2956230e913517c7/aptos-move/framework/aptos-framework/sources/aptos_governance.move#L322)).

third_party/move/move-prover/bytecode/src/function_target_pipeline.rs
```diff
@@ -225,6 +225,9 @@ impl FunctionTargetsHolder {
         &'env self,
         func_env: &'env FunctionEnv<'env>,
     ) -> Vec<(FunctionVariant, FunctionTarget<'env>)> {
+        if func_env.is_inline() {
```
rahxephon89, line 228: Hi @xindingw, you could remove this. I will fix it in a separate PR.

aptos-move/framework/aptos-framework/sources/delegation_pool.move
```diff
@@ -429,6 +513,51 @@ module aptos_framework::delegation_pool {
...
+    /// latest state.
+    public fun calculate_and_update_voter_total_voting_power(pool_address: address, voter: address): u64 acquires DelegationPool, GovernanceRecords {
+        // This function is meaningful only when the pool has already enabled partial governance voting.
+        assert_partial_governance_voting_enabled(pool_address);
+        // This function is meaningful only when delegation pool is synced.
+        synchronize_delegation_pool(pool_address);
+        let pool = borrow_global<DelegationPool>(pool_address);
+        let governance_records = borrow_global<GovernanceRecords>(pool_address);
+        let (latest_voter_voting_power, _, _) = get_latest_voter_voting_power(pool, governance_records, voter);
+        let delegation_pool = borrow_global<DelegationPool>(pool_address);
```
lightmark, line 542: why borrow it twice?

aptos-move/framework/aptos-framework/sources/transaction_validation.spec.move
```diff
@@ -106,6 +106,7 @@ spec aptos_framework::transaction_validation {
         txn_expiration_time: u64,
         chain_id: u8,
     ) {
+        pragma timeout = 100;
```
lightmark, line 109: ?

aptos-move/framework/aptos-framework/sources/delegation_pool.move
```diff
@@ -481,6 +623,121 @@ module aptos_framework::delegation_pool {
...
+        assert!(features::delegation_pool_partial_governance_voting_enabled(), error::invalid_state(EDISABLED_FUNCTION));
+        assert_delegation_pool_exists(pool_address);
+        // synchronize delegation and stake pools before any user operation.
+        synchronize_delegation_pool(pool_address);
+
+        let delegation_pool = borrow_global<DelegationPool>(pool_address);
+        let stake_pool_signer = retrieve_stake_pool_owner(delegation_pool);
+        // delegated_voter is managed by the stake pool itself, which signer capability is managed by DelegationPool.
+        // So voting power of this stake pool can only be used through this module.
+        stake::set_delegated_voter(&stake_pool_signer, signer::address_of(&stake_pool_signer));
```
lightmark, line 648: I think it should add that the voter address has to be itself because aptos_governance needs it.

aptos-move/framework/aptos-framework/sources/delegation_pool.move
```diff
@@ -554,15 +816,131 @@ module aptos_framework::delegation_pool {
...
+        pool: &DelegationPool,
+        governance_records :&mut GovernanceRecords,
+        delegator: address
+    ): &mut VoteDelegation {
+        let pool_address = get_pool_address(pool);
+        let locked_until_secs = stake::get_lockup_secs(pool_address);
+
+        let vote_delegation_table = &mut governance_records.vote_delegation;
+        // By default, a delegator's delegated voter is itself.
+        // TODO: recycle storage when VoteDelegation equals to default value.
```
lightmark, line 860: how do you recycle?

lightmark (approved):

aptos-move/framework/aptos-framework/sources/aptos_governance.move
```diff
@@ -83,6 +92,11 @@ module aptos_framework::aptos_governance {
...
```
movekevin, line 97: Let's also use SmartTable here instead

xindingw: @rahxephon89 I got these error in specs:

>   error: expression construct not supported in specifications
>       ┌─ /Users/xinding/src/public/aptos-core/aptos-move/framework/aptos-framework/sources/aptos_governance.spec.move:620:13
>       │
>   620 │             *smart_table::borrow(voting_records_v2.votes, record_key)
>       │             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> 
>   error: calling impure function `smart_table::contains` is not allowed
>       ┌─ /Users/xinding/src/public/aptos-core/aptos-move/framework/aptos-framework/sources/aptos_governance.spec.move:619:37
>       │
>   619 │         let used_voting_power = if (smart_table::contains(voting_records_v2.votes, record_key)) {
>       │                                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>       │
>       = impure function `smart_table::contains(&smart_table::SmartTable<#0, #1>, #0): bool`

It seems that I need to write lots of `spec_` functions for smart_table and its dependencies. Any easy way I can do that?

rahxephon89: Unfortunately we do not have intrinsic implementations of smart_table APIs encoded in the prover right now. We will work on them later.

aptos-move/framework/aptos-framework/sources/delegation_pool.move
xindingw: Sry I don't fully understand it. Do you mean adding it into comments? Can you elaborate "aptos_governance needs it"?

aptos-move/framework/aptos-framework/sources/transaction_validation.spec.move
xindingw: Removed. It was added because the spec didn't pass.

aptos-move/framework/aptos-framework/sources/delegation_pool.move
xindingw: When it equals to default value, remove the entry in the table. However doing it inside this function might not be ideal because this function always returns a `&mut`. Lots of storage need to be recycled in this module so I think it could be complicated and we will need a separate PR.

aptos-move/framework/aptos-framework/sources/delegation_pool.move
```diff
@@ -221,6 +243,48 @@ module aptos_framework::delegation_pool {
...
+        // locked_until_secs of the stake pool when last time this struct is updated.
+        // If <locked_until_secs of the stake pool> != last_locked_until_secs, it means that a lockup period has passed.
+        // This is slightly different from ObservedLockupCycle because ObservedLockupCycle cannot detect if a lockup
+        // period is passed when there is no unlocking during the lockup period .
+        last_locked_until_secs: u64,
+    }
+
+    struct GovernanceRecords has key {
+        // `votes` tracks voting power usage of each voter on each proposal.
+        votes: Table<VotingRecordKey, u64>,
```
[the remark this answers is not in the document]
xindingw: Use `SmartTable` instead. In the future we can recycle the storage.

aptos-move/framework/aptos-framework/sources/aptos_governance.move
xindingw: Changed to `SmartTable`! Thanks for Teng's help on specs.

PR #8095, opened 2023-05-08
feat: add is_coin_store_frozen to the coin module

ppoliani:
### Description

There is currently not an easy way to tell if a user has frozen his CoinStore. This is can be a useful information in some scenarios. For example, in an auction based module one might decide to implement a `bid` function that returns coins back to the highest bidder when he is outbided. However, this poses a Denial of Service (DOS) vulnerability as described in the following example:

1. Chunk calls bid sending a small amount
2. Chunk immediately freezes his coin store resource
3. Alice call `bid` with a higher amount, but the call reverts because the bid function cannot return the coins back to Chunk.
4. Chunk unfreezes the CoinStore at the end of the auction and claims the resource 

We can maybe abstract this into a more general `can_deposit` function which will also take into account if the coinstore is initialized in the first place, as well.

### Test Plan
Unit test

### DIFF
--- a/aptos-move/framework/aptos-framework/sources/coin.move
+++ b/aptos-move/framework/aptos-framework/sources/coin.move
@@ -242,6 +242,10 @@ module aptos_framework::coin {
     #[view]
     /// Returns `true` is account_addr has frozen the CoinStore
     public fun is_coin_store_frozen<CoinType>(account_addr: address): bool acquires CoinStore {
+        if(!is_account_registered<CoinType>(account_addr)) {
+          return false
+        };
+
         let coin_store = borrow_global<CoinStore<CoinType>>(account_addr);
         coin_store.frozen
     }
@@ -852,6 +856,9 @@ module aptos_framework::coin {
     #[test(account = @0x1)]
     public fun test_is_coin_store_frozen(account: signer) acquires CoinStore {
         let account_addr = signer::address_of(&account);
+        // An non registered account is has a frozen coin store
+        assert!(!is_coin_store_frozen<FakeMoney>(account_addr), 1);
+
         account::create_account_for_test(account_addr);
         let (burn_cap, freeze_cap, mint_cap) = initialize_and_register_fake_money(&account, 18, true);
 
--- a/aptos-move/framework/aptos-framework/sources/coin.move
+++ b/aptos-move/framework/aptos-framework/sources/coin.move
@@ -240,6 +240,13 @@ module aptos_framework::coin {
     }
 
     #[view]
+    /// Returns `true` is account_addr has frozen the CoinStore
+    public fun is_coin_store_frozen<CoinType>(account_addr: address): bool acquires CoinStore {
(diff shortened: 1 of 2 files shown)


gregnazario (approved):

gregnazario:
This is a good contribution!  Thanks for adding it

movekevin (approved):

aptos-move/framework/aptos-framework/sources/coin.move
```diff
@@ -239,6 +239,13 @@ module aptos_framework::coin {
         exists<CoinInfo<CoinType>>(coin_address<CoinType>())
     }
 
+    #[view]
+    /// Returns `true` is account_addr has frozen the CoinStore
+    public fun is_coin_store_frozen<CoinType>(account_addr: address): bool acquires CoinStore {
+        let coin_store = borrow_global<CoinStore<CoinType>>(account_addr);
```
lightmark, line 245: check `is_account_registered`?

davidiw: Maybe this should be just is frozen?

movekevin: Is_frozen can be confusing because coin::is_frozen gives the impression that the coin is frozen?

ppoliani: I believe `is_coin_store_frozen` name describes in accuracy what the function returns. From the caller moule perspective, this information is needed either when calling `withdraw` or `deposit` because both functions have the following assertion:

```
assert!(
  !coin_store.frozen,
  error::permission_denied(EFROZEN),
);
```

So knowing whether the coin_store is frozen can help caller decide what to do. Or maybe for the same exact reason we can call the function `is_coin_store_active`?

@lightmark I believe you're right, we should probably check `is_account_registered ` instead of aborting (btw how does abort work with #view functions?).

aptos-move/framework/aptos-framework/sources/coin.move
```diff
@@ -842,6 +849,29 @@ module aptos_framework::coin {
         });
     }
 
+    #[test(account = @0x1)]
+    public fun test_is_coin_store_frozen(account: signer) acquires CoinStore {
+        let account_addr = signer::address_of(&account);
+        account::create_account_for_test(account_addr);
+        let (burn_cap, freeze_cap, mint_cap) = initialize_and_register_fake_money(&account, 18, true);
```
brmataptos, line 863: Do you also want a test for the uninitialized account case?

movekevin: Yes please

ppoliani: There is a test for that in the latest commit

movekevin: Awesome!!

alokeutpal (approved):

PR #8100, opened 2023-05-08
Add mempool_commit_ack_timeout_ms to state sync driver config

vusirikala:
### Description
Add `mempool_commit_ack_timeout_ms` to state sync driver config and set its default value to 5000.


config/src/config/state_sync_config.rs
```diff
@@ -113,6 +113,8 @@ pub struct StateSyncDriverConfig {
     pub max_stream_wait_time_ms: u64,
     /// The version lag we'll tolerate before snapshot syncing
     pub num_versions_to_skip_snapshot_sync: u64,
+    /// The time the state sync waits for mempool to repsond to its acks
```
JoshLind, line 116: nit: Can we change this to: `/// The maximum time (ms) allowed for mempool to ack a commit notification`?

config/src/config/state_sync_config.rs
```diff
@@ -113,6 +113,8 @@ pub struct StateSyncDriverConfig {
...
+    pub mempool_commit_ack_timeout_ms: u64,
```
JoshLind, line 117: Can we keep things alphabetical? 😄

state-sync/state-sync-v2/state-sync-driver/src/notification_handlers.rs
```diff
@@ -393,12 +391,17 @@ impl FusedStream for ErrorNotificationListener {
...
 pub struct MempoolNotificationHandler<M> {
     mempool_notification_sender: M,
+    state_sync_driver_config: StateSyncDriverConfig,
 }
 
 impl<M: MempoolNotificationSender> MempoolNotificationHandler<M> {
-    pub fn new(mempool_notification_sender: M) -> Self {
+    pub fn new(
+        mempool_notification_sender: M,
+        state_sync_driver_config: StateSyncDriverConfig,
```
JoshLind, line 400: nit: FWIW, I wouldn't pass the entire config in here because we're only really using one item. I'd rather take the value directly 😄 (e.g., `mempool_commit_ack_timeout_ms`).

But, I guess some could argue this a style choice. So, I'll leave it up to you 😄

JoshLind (commented):
Thanks Satya! Looks solid. Just nits, but otherwise, good to go!

state-sync/state-sync-v2/state-sync-driver/src/notification_handlers.rs
vusirikala: Done.

config/src/config/state_sync_config.rs
vusirikala: Done.

config/src/config/state_sync_config.rs
vusirikala: Done.

JoshLind (approved):
Awesome, thanks @vusirikala 😄

grao1991 (approved):

PR #8103, opened 2023-05-08
[docker] Update docker compose for validator testnet

adschwartz:
The docker compose for the local validator testnet needed an update to get in sync with changes in the code base. 

I've verified that it runs and the faucet transactions are validated by the validator


rustielin (approved):
LGTM, thanks for submitting this fix!

Waiting review by @banool too. I believe the faucet docs/README should be updated as well, since some fields have been renamed/deprecated

PR #8105, opened 2023-05-08
[Block-STM] Support for sharded blockSTM with no conflict transactions

sitalkedia:
### Description

- A very basic block-stm sharding implementation that assumes no conflict transactions. 
- `ShardedBlockExecutor` has been created which manages multiple executor shards and splits an incoming block and send it to the sharded executor to execute. Currently, only uniform sharding is supported.
- `ExecutorShard` runs block-stm locally and executes the set of transactions received from the sharded block executor. 
- Various improvements to block-stm benchmark, which includes running the parallel execution as a standalone mode along with metrics reporting. 
- Please note that these thresholds have changed because of the workload pattern has changed. Earlier, the account working set size was just 10k and we used to regenerate the account for every block. This is not really reflective of the workload pattern we will see. So instead, I have increased the working set size to be 200k and kept the same working set across the blocks, so potentially they might be re-used across blocks.

### Test Plan

Ran the benchmark on `n2d-standard-224 ` machine and I could achieve ~290k TPS for sharded in-memory benchmark. This is as compared with 110k TPS with un-sharded blockSTM. 

Commands ran for reference

```
taskset -c 0-56 cargo run --release -p  aptos-transaction-benchmarks -- parallel-execution --concurrency-level-per-shard 56 --num-executor-shards 1 --num-blocks 20 --num-accounts 200000 --block-size 100000 --num-warmups 5
Avg Parallel TPS = 113753

cargo run --release -p  aptos-transaction-benchmarks -- parallel-execution --concurrency-level-per-shard 1 --num-executor-shards 224 --num-blocks 20 --num-accounts 200000 --block-size 100000 --num-warmups 5
Avg Parallel TPS = 289435
```


aptos-move/aptos-vm-logging/src/lib.rs
```diff
@@ -72,12 +72,7 @@ pub fn speculative_log(level: Level, context: &AdapterLogSchema, message: String
                 alert!("{:?}", e);
             };
         },
-        None => {
-            alert!(
-                "Speculative state not initialized to log message = {}",
-                message
-            );
-        },
+        None => {},
```
sitalkedia, line 75: Remove this error message here as it doesn't play well with multiple blockSTM instances.

testsuite/parallel_execution_performance.py
```diff
@@ -5,30 +5,31 @@
...
-    "1k_8": 15000,
-    "1k_16": 19000,
-    "1k_32": 20000,
-    "10k_8": 27000,
-    "10k_16": 46000,
-    "10k_32": 63000,
-    "50k_8": 29000,
-    "50k_16": 53000,
-    "50k_32": 82000,
+    "1k_8": 11000,
```
sitalkedia, line 12: Please note that these thresholds have changed because of the workload pattern has changed. Earlier, the account working set size was just 10k and we used to regenerate the account for every block. This is not really reflective of the workload pattern we will see. So instead, I have increased the working set size to be 200k and keeping the same working set across the blocks, so potentially they might be re-used across blocks.

aptos-move/aptos-transaction-benchmarks/src/transactions.rs
```diff
@@ -204,56 +209,126 @@ impl TransactionBenchState {
...
-            executor,
-            transactions,
-        }
+        // Insert a blockmetadata transaction at the beginning to better simulate the real life traffic.
+        let validator_set = ValidatorSet::fetch_config(
+            &FakeExecutor::from_head_genesis()
+                .get_state_view()
+                .as_move_resolver(),
+        )
+        .expect("Unable to retrieve the validator set from storage");
```
msmouse, line 255: should cache on the struct?

aptos-move/aptos-transaction-benchmarks/src/transactions.rs
```diff
@@ -204,56 +209,126 @@ impl TransactionBenchState {
...
+        transactions
     }
 
     /// Executes this state in a single block.
-    fn execute(self) {
+    fn execute_sequential(mut self) {
         // The output is ignored here since we're just testing transaction performance, not trying
         // to assert correctness.
-        BlockAptosVM::execute_block(self.transactions, self.executor.get_state_view(), 1)
+        let executor = self.sequential_block_executor.clone();
```
msmouse, line 275: why the clone?

aptos-move/aptos-transaction-benchmarks/src/transactions.rs
```diff
@@ -204,56 +209,126 @@ impl TransactionBenchState {
...
+    fn execute_parallel(mut self) {
         // The output is ignored here since we're just testing transaction performance, not trying
         // to assert correctness.
-        BlockAptosVM::execute_block(
-            self.transactions,
-            self.executor.get_state_view(),
-            num_cpus::get(),
-        )
-        .expect("VM should not fail to start");
+        let executor = self.parallel_block_executor.clone();
```
msmouse, line 285: ditto

aptos-move/aptos-transaction-benchmarks/src/transactions.rs
```diff
@@ -204,56 +209,126 @@ impl TransactionBenchState {
...
+        no_conflict_txns: bool,
     ) -> (usize, usize) {
-        BlockAptosVM::execute_block_benchmark(
-            self.transactions,
-            self.executor.get_state_view(),
-            concurrency_level,
-            run_par,
-            run_seq,
-        )
+        //self.account_universe.reset_pick_s
```
msmouse, line 312: remove?

aptos-move/aptos-vm/src/sharded_block_executor/mod.rs
```diff
@@ -0,0 +1,143 @@
...
+            self.command_txs[i]
+                .send(ExecutorShardCommand::ExecuteBlock(transactions))
+                .unwrap();
+        }
+        // wait for all remote executors to send the result back and append them in order by shard id
+        // maintain a map of shard id to results and aggregate them later
+        let mut results = HashMap::new();
+        trace!("ShardedBlockExecutor Waiting for results");
+        for _ in 0..self.num_executor_shards {
+            let (shard_id, result) = self.result_rx.recv().unwrap();
```
msmouse, line 93: nit: why not hold a `Vec<Receiver>` so you don't need to allocate the hashmap to sort the results

msmouse (approved):

msmouse (approved):

igor-aptos (commented):
Looks pretty good, (and surprisingly clean!), minor nits inline

testsuite/sequential_execution_performance.py
```diff
@@ -4,13 +4,13 @@
 
 # Set the tps threshold for block size 1k, 10k and 50k
 BLOCK_SIZES = ["1k", "10k", "50k"]
```
igor-aptos, line 6: should be pass block sizes in, instead of needing to rely on them being the same here and the defaults in the code?

aptos-move/aptos-vm/src/sharded_block_executor/mod.rs
sitalkedia: Good point - that was pretty stupid of me. Made the code much cleaner as well :)

testsuite/sequential_execution_performance.py
sitalkedia: Not sure If I understand, can you clarify?

sitalkedia:
Addressed most of the comments, PTAL.

testsuite/sequential_execution_performance.py
igor-aptos: I mean this relies on this line above matching and not being changed:
  let block_sizes = opt.block_sizes.unwrap_or_else(|| vec![1000, 10000, 50000]);
  
can we just pass in the --block-sizes param to the command here?

igor-aptos (approved):

testsuite/sequential_execution_performance.py
sitalkedia: Yes, good point. Let me sneak that change in the follow-up change I am working on - that will change this file anyways.

PR #8117, opened 2023-05-09
[indexer-grpc] local filestore implementation

rustielin:
### Description

Set up a `LocalFileStoreOperator` for easier testing, which writes stuff to local filesystem. Separates out `FileStoreOperator` into a trait, which can also be used for mocking down the line to write better unit tests, and opens the door for other FileStores (sinks).

The File Store current implementation only talks to GCS, which is difficult to test without touching GCP. While there are GCS emulators that you can run locally, unfortunately the `cloud-storage` crate that we use does not play nicely with it.

Refactors the config, which now has a top-level `file_store` config key:

```
# for use with local
file_store:
  file_store_type: LocalFileStore
  local_file_store_path: /var/folders/0r/0cqndyk91kb6s_xv23gdvsdw0000gn/T/tmp.RK6NhPJf

# for use with GCS
file_store:
  file_store_type: GcsFileStore
  gcs_file_store_bucket_name: rustie-playground-indexer-grpc-file-store
  
```

### Test Plan

Existing unit tests and new top-level README for testing the whole setup locally.


ecosystem/indexer-grpc/indexer-grpc-utils/src/file_store_operator/mod.rs
```diff
@@ -0,0 +1,157 @@
...
+        Self {
+            chain_id,
+            file_folder_size: BLOB_STORAGE_SIZE,
+            version,
+        }
+    }
+}
+
+#[async_trait::async_trait]
+pub trait FileStoreOperator: Send + Sync {
```
larry-aptos, line 56: neat

larry-aptos (approved):

bowenyang007 (approved):

geekflyer:
> unfortunately the `cloud-storage` crate that we use does not play nicely with it.

why?

rustielin:
@geekflyer The one I found was https://github.com/oittaa/gcp-storage-emulator. The `cloud-storage` crate we're using hardcodes the base URL to `https://storage.googleapis.com/storage/v1` (https://docs.rs/cloud-storage/latest/src/cloud_storage/lib.rs.html#134), whereas with the emulator, we'd need to point it to `http://localhost:9023`.

sionescu (approved):

PR #8119, opened 2023-05-09
Add smoke test for inspection service

vusirikala:
### Description

Add smoke test to check that inspection service is live.


JoshLind (commented):
Nice, thanks @vusirikala 😄  Let's wait for this to land so that you can ping the index page in the test -- it'll be cleaner.

testsuite/smoke-test/src/inspection_service.rs
```diff
@@ -0,0 +1,16 @@
...
+
+use crate::smoke_test_environment::new_local_swarm_with_aptos;
+use aptos_forge::Swarm;
+
+#[tokio::test]
+async fn test_inspection_service_connection() {
+    let mut swarm = new_local_swarm_with_aptos(1).await;
+    let info = swarm.aptos_public_info();
+
+    let resp = reqwest::get(info.inspection_service_url().to_owned())
```
JoshLind, line 12: nit: can we add a comment to this? e.g.,:

```
// Ping the inspection service index page and verify we get a successful response
```

JoshLind (approved):
Awesome! Thanks @vusirikala 🚀

testsuite/smoke-test/src/inspection_service.rs
```diff
@@ -0,0 +1,16 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::smoke_test_environment::new_local_swarm_with_aptos;
+use aptos_forge::Swarm;
+
+#[tokio::test]
+async fn test_inspection_service_connection() {
+    let mut swarm = new_local_swarm_with_aptos(1).await;
+    let info = swarm.aptos_public_info();
+
```
vusirikala, line 11:
```suggestion
// Ping the inspection service index page and verify we get a successful response
```

testsuite/smoke-test/src/inspection_service.rs
```diff
@@ -0,0 +1,16 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::smoke_test_environment::new_local_swarm_with_aptos;
+use aptos_forge::Swarm;
+
+#[tokio::test]
+async fn test_inspection_service_connection() {
+    let mut swarm = new_local_swarm_with_aptos(1).await;
+    let info = swarm.aptos_public_info();
+// Ping the inspection service index page and verify we get a successful response
```
vusirikala, line 11:
```suggestion
    // Ping the inspection service index page and verify we get a successful response
```

testsuite/smoke-test/src/inspection_service.rs
vusirikala: Done.

igor-aptos:
any reason to add it to all utilities, instead of just getting it directly in that test? do we expect it to be widely used across tests? 

should it be optional, or is inspection service required to be enabled?

grao1991 (approved):

PR #8120, opened 2023-05-09
[TS SDK] Get number of delegators

0xmaayan:
### Description

Modify number of delegators Indexer query

### Test Plan
tests are passing


banool (approved):
Seems legit. This is a breaking change, so I guess we need to bump the SDK major version when we eventually release it.

0xjinn (approved):

Issue #7422, opened 2023-03-27
[Prover][Issue] `simple_map` doesn't work as expected.


junkil-park:
## How to reproduce
Run Prover for the following example:

```
    struct Pool {
        shares: SimpleMap<address, u64>,
        shareholders: vector<address>,
    }

    fun f(pool: &mut Pool, shareholder: address) {
        simple_map::remove(&mut pool.shares, &shareholder);
        spec {
            assert !simple_map::spec_contains_key(pool.shares, shareholder);
        };
        let (_, shareholder_index) = vector::index_of(&pool.shareholders, &shareholder);
        vector::remove(&mut pool.shareholders, shareholder_index);
        spec {
            assert !vector::spec_contains(pool.shareholders, shareholder);
        };
    }
    spec f {
        requires len(pool.shareholders) <= 2;
        invariant forall addr: address:
        (simple_map::spec_contains_key(pool.shares, addr) ==> vector::spec_contains(pool.shareholders, addr));

        invariant forall addr: address:
            (vector::spec_contains(pool.shareholders, addr) ==> simple_map::spec_contains_key(pool.shares, addr));

        // `shareholders` does not contain duplicates.
        invariant forall i in 0..len(pool.shareholders), j in 0..len(pool.shareholders):
            pool.shareholders[i] == pool.shareholders[j] ==> i == j;
    }
```

Prover generates the following error:
```
error: post-condition does not hold
   ┌─ /Users/jpark/Temp/move/test/sources/havoced_pool.move:61:9
   │
61 │ ╭         invariant forall addr: address:
62 │ │         (simple_map::spec_contains_key(pool.shares, addr) ==> vector::spec_contains(pool.shareholders, addr));
   │ ╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────^
   │
   = Related Bindings:
   =         pool =
   =           &havoced_pool.Pool{
   =             shares = Table{0x5f90: 0, 0x5f8e: 1, ..},
   =             shareholders = vector{(size): 1, default: 0x1}}
   = Execution Trace:
   =     at /Users/jpark/Temp/move/test/sources/havoced_pool.move:48: f
   =     at /Users/jpark/Temp/move/test/sources/havoced_pool.move:60: f (spec)
   =     at /Users/jpark/Temp/move/test/sources/havoced_pool.move:61: f (spec)
   =     at /Users/jpark/Temp/move/test/sources/havoced_pool.move:64: f (spec)
[...]

PR #8121, opened 2023-05-09
[move-prover] on different theories for vector modeling

meng-xu-cs:
This PR contains two test cases to show how vector theories might have an impact on the prover's ability on proving complicated propositions w.r.t vector, especially with heavy involvement of quantifiers:

- On test case `f1`:
  - `BoogieArray` (the default vector theory) will given a verification error with an inconsistent counterexample while
  - `BoogieArrayIntern` and `SmtSeq` can prove the proposition

- On test case `f2`:
  - `BoogieArray` times out while
  - `BoogieArrayIntern` and `SmtSeq` can prove the proposition

Addressing issue #7422 

### Test Plan
CI


junkil-park (approved):
Thank you! Can we set the vector theory in CLI, like Move CLI or especially Aptos CLI? I can't find the option exposed.

Could we also have some negative test cases which are supposed to fail?

davidiw (approved):

wrwg (approved):

PR #8123, opened 2023-05-09
[Fix] Handle reconfig suffix blocks.

grao1991:
### Description
Should pass empty state updates for reconfig suffix blocks.
Also skip the state commit if the block is empty.

### Test Plan
New UT.


msmouse (approved):
beautiful

zekun000 (approved):

PR #8124, opened 2023-05-09
[TF] Add variable "bootstrap_cluster" to GCP base module

sionescu:
# Description

The module uses data.kubernetes_all_namespaces to dynamically get the list of all namespaces, but that can only be done after a cluster has been turned up.
On cluster creation, TF attempts to connect to the K8s control place to fetch the list of namespaces, and fails.

The solution is to split cluster bootstrap in two phases. This is only necessary while we're in the middle of the migration from PSP to PSS.
Once all clusters have been upgraded to PSS and K8s version >=1.25, the bootstrapping dance can be removed.

# Test plan

Tested by turning up a GCP devnet.

### DIFF
--- a/terraform/aptos-node-testnet/gcp/main.tf
+++ b/terraform/aptos-node-testnet/gcp/main.tf
@@ -17,7 +17,8 @@ provider "helm" {
 module "validator" {
   source = "../../aptos-node/gcp"
 
-  manage_via_tf = var.manage_via_tf
+  cluster_bootstrap = var.cluster_bootstrap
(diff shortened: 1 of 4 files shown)


sherry-x (approved):

rustielin (approved):
LGTM makes sense. Should we do the same for all the other `aptos-node/*` modules?

Also this pattern makes it seems like we should actually separate all the PSP stuff into another TF module.

sionescu:
@rustielin I'll add this to the AWS modules later if necessary. Right now I'm focused on turning up a bunch of GCP clusters.

PR #8126, opened 2023-05-10
[Spec] Update specs of `transaction_fee` & `transaction_validation`

xudon9:
### Description
This is a part of spec work for Aptos Framework from MoveBit. The change is:
- Specified all aborts conditions in `transaction_validation`
- Specified `burn_fee` in `transaction_fee`
### Test Plan
`aptos move prove -f transaction_fee` and `aptos move prove -f transaction_validation` passes.


rahxephon89 (approved):
Thanks!

junkil-park (approved):
LGTM! Thank you!

PR #8128, opened 2023-05-10
[CLI] Create aptos-ledger crate

0xjinn:
### Description
Create a `aptos-ledger` crate, which allow Rust application to communicate with our Aptos app on Ledger

Functions that are being created on this initial PR
- get_app_version
- get_app_name
- get_public_key
- sign_txn

### Test Plan

Manually tested to make sure These method are able to get the correct result from my Ledger Nano X

```
Version: Ok("0.4.17")
App Name: Ok("Aptos")
Public Key: Ok("05a8ace09d1136181029be3e817de3619562b0da2eedbff210e2b2f92c71be70")
```


crates/aptos-ledger/src/lib.rs
```diff
@@ -0,0 +1,301 @@
...
+const CLA_APTOS: u8 = 0x5B; // Aptos CLA Instruction class
+const INS_GET_VERSION: u8 = 0x03; // Get version instruction code
+const INS_GET_APP_NAME: u8 = 0x04; // Get app name instruction code
+const INS_GET_PUB_KEY: u8 = 0x05; // Get public key instruction code
+const INS_SIGN_TXN: u8 = 0x06; // Sign the transaction
+const APDU_ANSWER_CODE: u16 = 36864; // success code for transport.exchange
+
+const MAX_APDU_LEN: u16 = 255;
+const CHUNK_SIZE: usize = 128; // Chunk size to be sent to Ledger
+const P1_NON_CONFIRM: u8 = 0x00;
```
0xmaayan, line 21: 21, 23 and 25 have the same value, maybe unify?

crates/aptos-ledger/src/lib.rs
```diff
@@ -0,0 +1,301 @@
...
+/// Returns the current version of the Aptos app on Ledger
+pub fn get_app_version() -> Result<String, AptosLedgerError> {
+    // open connection to ledger
+    let transport = match open_ledger_transport() {
+        Ok(transport) => transport,
+        Err(err) => return Err(err),
+    };
+
+    match transport.exchange(&APDUCommand {
+        cla: CLA_APTOS,
```
0xmaayan, line 44: what are all those parameters?

crates/aptos-ledger/src/lib.rs
```diff
@@ -0,0 +1,301 @@
...
+                let major = response.data()[0];
+                let minor = response.data()[1];
+                let patch = response.data()[2];
+                let version = format!("{}.{}.{}", major, minor, patch);
+                Ok(version)
+            } else {
+                let error_string = response
+                    .error_code()
+                    .map(|error_code| error_code.to_string())
+                    .unwrap_or_else(|retcode| format!("Error with retcode: {}", retcode));
```
0xmaayan, line 62: Error with "retcode"?

crates/aptos-ledger/src/lib.rs
```diff
@@ -0,0 +1,301 @@
...
+        data: cdata,
+    }) {
+        Ok(response) => {
+            // Got the response from ledger after user has confirmed on the ledger wallet
+            if response.retcode() == APDU_ANSWER_CODE {
+                // extract the Public key from the response data
+                let mut offset = 0;
+                let response_buffer = response.data();
+                let pub_key_len: usize = (response_buffer[offset] - 1).into();
+                offset += 1;
```
0xmaayan, line 139: `offset += 1` is duplicated

crates/aptos-ledger/src/lib.rs
```diff
@@ -0,0 +1,301 @@
...
+    // open connection to ledger
+    let transport = match open_ledger_transport() {
+        Ok(transport) => transport,
+        Err(err) => return Err(err),
+    };
+
+    // serialize the derivative path
+    let derivative_path_bytes = serialize_bip32(DERIVATIVE_PATH);
+
+    // await this.sendToDevice(INS.SIGN_TX, P1_START, P2_MORE, pathBuffer);
```
0xmaayan, line 169: remove comment?

0xmaayan (approved):
LGTM! left some comments and questions

crates/aptos-ledger/src/lib.rs
0xjinn: Ah yeah, I was hestating unifying them. But each one serve a different purpose, so I decided to separate them out. (Same format as the Petra wallet - ledger integration)

crates/aptos-ledger/src/lib.rs
0xjinn: Good question! These parameters are APDU parameters for communicating with Ledger, or smart card device. 

cla - specific to app
ins - instruction parameter (or method we want to invoke on ledger side)
p1 - p2 - instruction parameter
data - payload in bytes

crates/aptos-ledger/src/lib.rs
0xjinn: Yep, it is the SW defined on our Aptos app on Ledger - https://github.com/aptos-labs/ledger-app-aptos/blob/main/doc/COMMANDS.md

crates/aptos-ledger/src/lib.rs
0xjinn: Its actually intentional :) 

Reference from the Petra Ledger side - https://github.com/aptos-labs/wallet/blob/main/apps/extension/src/core/ledger/index.ts#L163

crates/aptos-ledger/src/lib.rs
```diff
@@ -0,0 +1,260 @@
+// Copyright © Aptos Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+use hex::encode;
+use ledger_apdu::APDUCommand;
+use ledger_transport_hid::{hidapi::HidApi, TransportNativeHID};
+use std::str;
+
+const DERIVATIVE_PATH: &str = "m/44'/637'/0'/0'/0'"; // TODO: Add support for multiple index
```
banool, line 9: Could you add a comment explaining what this is for, where it comes from, etc.

crates/aptos-ledger/src/lib.rs
```diff
@@ -0,0 +1,260 @@
...
+const P2_MORE: u8 = 0x80;
+const P2_LAST: u8 = 0x00;
+
+#[derive(Debug)]
+pub enum AptosLedgerError {
+    /// Error when trying to open a connection to the Ledger device
+    DeviceNotFound,
+    /// Unexpected error
+    UnexpectedError(String),
+}
```
banool, line 31: Let's use `thiserror` for this?

crates/aptos-ledger/src/lib.rs
```diff
@@ -0,0 +1,260 @@
...
+        p2: P2_LAST,
+        data: vec![],
+    }) {
+        Ok(response) => {
+            // Received response from Ledger
+            if response.retcode() == APDU_CODE_SUCCESS {
+                let major = response.data()[0];
+                let minor = response.data()[1];
+                let patch = response.data()[2];
+                let version = format!("{}.{}.{}", major, minor, patch);
```
banool, line 54: Instead of turning this into a string, could we make a `Version` struct with major, minor, and patch fields, and then impl `Display` for it?

crates/aptos-ledger/src/lib.rs
```diff
@@ -0,0 +1,260 @@
...
+                Err(AptosLedgerError::UnexpectedError(error_string))
+            }
+        },
+        Err(err) => Err(AptosLedgerError::UnexpectedError(err.to_string())),
+    }
+}
+
+/// Returns the official app name register in Ledger
+pub fn get_app_name() -> Result<String, AptosLedgerError> {
+    // open connection to ledger
```
banool, line 70: Mega nit, some comments start with Capital letters, others don't.
```suggestion
    // Open connection to ledger
```

crates/aptos-ledger/src/lib.rs
```diff
@@ -0,0 +1,260 @@
...
+    let transport = match open_ledger_transport() {
+        Ok(transport) => transport,
+        Err(err) => return Err(err),
+    };
```
banool, line 74:
```suggestion
    let transport = open_ledger_transport()?;
```

crates/aptos-ledger/src/lib.rs
```diff
@@ -0,0 +1,260 @@
...
+                    .unwrap_or_else(|retcode| format!("Error with retcode: {:x}", retcode));
+                Err(AptosLedgerError::UnexpectedError(error_string))
+            }
+        },
+        Err(err) => Err(AptosLedgerError::UnexpectedError(err.to_string())),
+    }
+}
+
+/// Returns the official app name register in Ledger
+pub fn get_app_name() -> Result<String, AptosLedgerError> {
```
banool, line 69: Beyond just preferred style, is there a concrete reason why these functions don't take in the `transport`? It seems to me if I wanted to do multiple things, it'd be better to create a transport and then use it for multiple functions, rather than making one each time.

Also, with that in mind, have we tested what happens when you trying to keep multiple transports open, e.g. if you're calling these functions in a multithreaded environment?

crates/aptos-ledger/src/lib.rs
```diff
@@ -0,0 +1,260 @@
...
+    };
+
+    match transport.exchange(&APDUCommand {
+        cla: CLA_APTOS,
+        ins: INS_GET_APP_NAME,
+        p1: P1_NON_CONFIRM,
+        p2: P2_LAST,
+        data: vec![],
+    }) {
+        Ok(response) => {
```
banool, line 83: Dang this is such an awkward struct, very un-Rust-like: https://docs.rs/ledger-transport/latest/ledger_transport/struct.APDUAnswer.html. I wish it was just a native Result and we could then use the error in our error class.

My concern is in the error case we're throwing out some potentially useful information, e.g. `response.data()`. Will that ever have something in it in the error case?

crates/aptos-ledger/src/lib.rs
```diff
@@ -0,0 +1,260 @@
...
+                Ok(app_name.to_string())
+            } else {
+                let error_string = response
+                    .error_code()
+                    .map(|error_code| error_code.to_string())
+                    .unwrap_or_else(|retcode| format!("Error with retcode: {:x}", retcode));
+                Err(AptosLedgerError::UnexpectedError(error_string))
+            }
+        },
+        Err(err) => Err(AptosLedgerError::UnexpectedError(err.to_string())),
```
banool, line 98: If you make `AptosLedgerError` an actual error using `thiserror`, this should be simpler, you could avoid the `err.to_string()` and perhaps even just use `?` if you define a `From`. I realize this is a bit vague lol but just start with `thiserror` and see how you go.

crates/aptos-ledger/src/lib.rs
```diff
@@ -0,0 +1,260 @@
...
+/// * `display` - If true, the public key will be displayed on the Ledger device, and confirmation is needed
+pub fn get_public_key(display: bool) -> Result<String, AptosLedgerError> {
+    // open connection to ledger
+    let transport = match open_ledger_transport() {
+        Ok(transport) => transport,
+        Err(err) => return Err(err),
+    };
+
+    // serialize the derivative path
+    let cdata = serialize_bip32(DERIVATIVE_PATH);
```
banool, line 115: It seems like this is the same every time, let's use this: https://docs.rs/once_cell/latest/once_cell/sync/struct.Lazy.html.

crates/aptos-ledger/src/lib.rs
```diff
@@ -0,0 +1,260 @@
...
+        Ok(response) => {
+            // Got the response from ledger after user has confirmed on the ledger wallet
+            if response.retcode() == APDU_CODE_SUCCESS {
+                // extract the Public key from the response data
+                let mut offset = 0;
+                let response_buffer = response.data();
+                let pub_key_len: usize = (response_buffer[offset] - 1).into();
+                offset += 1;
+
+                // Skipping weird 0x04
```
banool, line 139: Maybe a little more explanation of why we do this.

crates/aptos-ledger/src/lib.rs
```diff
@@ -0,0 +1,260 @@
...
+                            format!("Unknown Ledger APDU retcode: {:x}", retcode)
+                        });
+                    return Err(AptosLedgerError::UnexpectedError(error_string));
+                }
+            },
+            Err(err) => return Err(AptosLedgerError::UnexpectedError(err.to_string())),
+        };
+    }
+    Err(AptosLedgerError::UnexpectedError(
+        "Unable to process request".to_owned(),
```
banool, line 216: Nit: `to_string` would be more standard.

crates/aptos-ledger/src/lib.rs
```diff
@@ -0,0 +1,260 @@
...
+    ))
+}
+
+/// This is the Rust version of the serialization of BIP32 from Petra Wallet
+/// https://github.com/aptos-labs/wallet/blob/main/apps/extension/src/core/ledger/index.ts#L47
```
banool, line 221: If linking to code in another repo, don't point to main but to a specific rev, otherwise this link might break down the line.

crates/aptos-ledger/src/lib.rs
```diff
@@ -0,0 +1,260 @@
+// Copyright © Aptos Foundation
```
banool, line 1: Given we intend for this to be used as a library, I recommend we add this:
```
#![deny(missing_docs)]
```

This will enforce that all public functions / structs this crate exposes have doc comments.

banool (changes requested):
Looking great!!

crates/aptos-ledger/src/lib.rs
0xjinn: Good call!

crates/aptos-ledger/src/lib.rs
0xjinn: Ah, so basically adding the `thiserror` crate, and define `#[error()]` for each error?

crates/aptos-ledger/src/lib.rs
0xjinn: Yes for now. In the future, we will need to support multiple index account, and the derivative_path will be different depending on the account/address index. (Current version only support Account created at index 0)

crates/aptos-ledger/src/lib.rs
0xjinn: Not that I know of. The payload will only contains data if there is no error.  This contains the format of the return payload - https://github.com/aptos-labs/ledger-app-aptos/blob/main/doc/COMMANDS.md

crates/aptos-ledger/src/lib.rs
0xjinn: Good question! I don't think you can open multiple transport at the same time, it will lock the device - https://github.com/ruabmbua/hidapi-rs/issues/81

And passing in the transport for each function works, but I think it would be better to have each individual method to open the device, so the developer does not have to create/open device from their end, and pass in the transport value. 

Thoughts?

crates/aptos-ledger/src/lib.rs
banool: We should still use Lazy for this and each of those derivative paths if we know them ahead of time.

banool (changes requested):
Mostly nits, though requesting changes for the Debug thing.

crates/aptos-ledger/src/lib.rs
0xjinn: Yep, added Lazy for now. I will follow up with a new design on how to handle the multiple index accounts, since there could be a lot of them (10+)

crates/aptos-ledger/src/lib.rs
```diff
@@ -0,0 +1,309 @@
...
+#[derive(Debug, Error)]
+/// Aptos Ledger Error
+pub enum AptosLedgerError {
+    /// Error when trying to open a connection to the Ledger device
+    #[error("Device not found")]
+    DeviceNotFound,
+
+    /// Unexpected error
+    #[error("Unexpected Error: {0}")]
+    UnexpectedError(String, Option<u16>),
```
banool, line 51: Could you leave a comment explaining what the Option<u16> is here?

banool: Also you should probably show the error code, something like this
```suggestion
    /// Unexpected error
    #[error("Unexpected Error: {0} (code {1:?})")]
    UnexpectedError(String, Option<u16>),
```

banool (approved):
Awesome looking great, just one final nit!

PR #8129, opened 2023-05-10
Buffer aptos telemetry metrics during when victoria metrics is down

vusirikala:
### Description
When victoria metrics is down, the aptos telemetry service will buffer those metrics. Aptos telemetry service spawns a thread which resubmits these buffered metrics once every 1 minute.


crates/aptos-telemetry-service/src/prometheus_push_metrics.rs
```diff
@@ -86,6 +99,16 @@ pub async fn handle_metrics_ingest(
...
+                context
+                    .downtime_metrics_cache()
+                    .write()
+                    .push_back(MetricsEntry {
+                        context: context.clone(),
+                        claims: claims.clone(),
+                        encoding: encoding.clone(),
+                        metrics_body: metrics_body.clone(),
+                        timestamp,
+                    });
```
ibalajiarun, line 111: Caching here will lead to duplication. We have the ability to push to multiple prometheus backends for high availability (for mainnet, we push to two backends), so if we push to cache on each failure, we will end up caching multiple times. See line 73 which iterates over client vec.

We should instead cache if we weren't able to send to any backends.

crates/aptos-telemetry-service/src/lib.rs
```diff
@@ -107,6 +109,14 @@ impl AptosTelemetryServiceArgs {
         let validator_fullnodes = Arc::new(aptos_infallible::RwLock::new(HashMap::new()));
         let public_fullnodes = config.pfn_allowlist.clone();
 
+        let downtime_metrics_cache = Arc::new(aptos_infallible::RwLock::new(VecDeque::<
+            MetricsEntry,
+        >::new()));
+        let downtime_metrics_cache_updater = DowntimeMetricsCacheUpdater::new(
+            downtime_metrics_cache.clone(),
+            Duration::from_secs(60000),
```
ibalajiarun, line 117: Also would suggest extracting into a constant
```suggestion
            Duration::from_secs(60),
```

crates/aptos-telemetry-service/src/context.rs
```diff
@@ -12,7 +13,7 @@ use aptos_types::{chain_id::ChainId, PeerId};
 use jsonwebtoken::{Algorithm, DecodingKey, EncodingKey, TokenData, Validation};
 use serde::{de::DeserializeOwned, Serialize};
 use std::{
-    collections::{HashMap, HashSet},
+    collections::{HashMap, HashSet, VecDeque},
```
ibalajiarun, line 16: nit: Why VecDeque instead of a Vec?

crates/aptos-telemetry-service/src/types/auth.rs
```diff
@@ -25,7 +25,7 @@ pub struct AuthResponse {
     pub handshake_msg: Vec<u8>,
 }
 
-#[derive(Debug, Serialize, Deserialize, PartialEq, Eq)]
+#[derive(Debug, Serialize, Deserialize, PartialEq, Eq, Clone)]
```
ibalajiarun, line 28: Do we need to clone the claim? Can we not simply move into the cache?

crates/aptos-telemetry-service/src/prometheus_push_metrics.rs
```diff
@@ -39,7 +42,17 @@ pub async fn handle_metrics_ingest(
...
+    let timestamp = Utc::now().timestamp() as usize;
+    push_metrics_to_clients(context, claims, encoding, metrics_body, timestamp).await
+}
 
+pub async fn push_metrics_to_clients(
+    context: Context,
+    claims: Claims,
+    encoding: Option<String>,
+    metrics_body: Bytes,
+    timestamp: usize,
```
ibalajiarun, line 54: This timestamp is not used except to cache. This should be passed to `client.post_prometheus_metrics` right?

ibalajiarun (changes requested):

crates/aptos-telemetry-service/src/context.rs
vusirikala: Makes it more efficient to pop elements from front of the queue.

crates/aptos-telemetry-service/src/types/auth.rs
vusirikala: I tried that before adding `clone`. I got this error. 
Screenshot 2023-05-10 at 12 42 43 PM

crates/aptos-telemetry-service/src/prometheus_push_metrics.rs
vusirikala: Thanks for pointing it out. Fixed it now.

crates/aptos-telemetry-service/src/prometheus_push_metrics.rs
vusirikala: Thanks for pointing it out. Fixed it.

crates/aptos-telemetry-service/src/clients/mod.rs
```diff
@@ -38,14 +38,18 @@ pub mod victoria_metrics_api {
...
                 .iter()
                 .map(|label| ("extra_label".into(), label.into()))
                 .collect();
 
             self.inner
-                .post(format!("{}api/v1/import/prometheus", self.base_url))
+                .post(format!(
+                    "{}api/v1/import/prometheus?timestamp={}",
+                    self.base_url, timestamp
+                ))
```
ibalajiarun, line 52: The timestamp should be part of the `query()` in line 55. We already send some extra labels in  query params, and timestamp should be part of that.

crates/aptos-telemetry-service/src/metrics.rs
```diff
@@ -142,10 +143,15 @@ impl PrometheusExporter {
         ];
 
         let start_timer = Instant::now();
-
+        let timestamp = Utc::now().timestamp() as usize;
```
ibalajiarun, line 146: I would suggest confirming if the timestamp is indeed in Utc, otherwise we might corrupt data unnecessarily.

crates/aptos-telemetry-service/src/prometheus_push_metrics.rs
```diff
@@ -56,13 +78,20 @@ pub async fn handle_metrics_ingest(
     };
 
     let start_timer = Instant::now();
+    let successful_clients = aptos_infallible::RwLock::new(Vec::<String>::new());
```
ibalajiarun, line 81: We should use a futures lock here which supports async/await. With this kind of lock, we would end up sleeping the current thread, which means we might be out of threads to run other concurrent futures. With async/await the tokio runtime will schedule another future on the thread instead. Important in this case because we are joining on futures here.

crates/aptos-telemetry-service/src/metrics.rs
vusirikala: Good catch. The documentation (https://docs.victoriametrics.com/#timestamp-formats) says unix timestamps. I changes from UTC to unix timestamps.

crates/aptos-telemetry-service/src/downtime_metrics_cache.rs
```diff
@@ -0,0 +1,66 @@
...
+        });
+    }
+
+    async fn resubmit(&mut self) {
+        let len = self.downtime_metrics.read().len();
+
+        for _i in 0..len {
+            let entry = self.downtime_metrics.write().pop_front();
+            if let Some(entry) = entry {
+                let _result = push_metrics_to_clients(
```
ibalajiarun, line 54: Okay, deadlock scenario here: there is a write lock on `self.downtime_metrics` and then you call `push_metrics_to_clients` which will hold another write lock on `downtime_metrics`. I think you need to guard the write() call to make sure you release the lock after you pop_front()

```suggestion
            let entry = {
                self.downtime_metrics.write().pop_front()
             };
            if let Some(entry) = entry {
                let _result = push_metrics_to_clients(
```

Issue #8077, opened 2023-05-05
[Bug]


thepomeranian:
[Feature Request] Mask private key in log output

https://github.com/aptos-labs/aptos-core/blob/f9c9b1a7fe005ce6704f599665998f8549c7f9f1/aptos-node/src/logger.rs#L92
This log output does not hide the private key of the VFN network
Most people currently should be using the same key, so I think it might not be a problem. But if there is a problem with the filtering should be a thing?

## Expected Behavior

TODO:
* add data masking to private key similar to: https://github.com/aptos-labs/aptos-core/blob/main/aptos-node/src/logger.rs#L86

## Additional context

This was originally submitted as a security concern and triaged with @thepomeranian @gedigi and @sherry-x

PR #8131, opened 2023-05-10
Print config using "debug" mode to avoid printing secret keys

vusirikala:
### Description

Fixed[ issue 8077 ]( https://github.com/aptos-labs/aptos-core/issues/8077). This prints node config in debug mode to avoid printing secret keys.


gedigi (approved):

JoshLind (approved):
Thanks @vusirikala, seems like a reasonable first step to harden things.

aptos-node/src/logger.rs
```diff
@@ -89,5 +89,5 @@ fn log_config_and_build_information(node_config: &NodeConfig) {
         }
     }
 
-    info!(config = config, "Loaded AptosNode config");
+    info!("Loaded node config: {:?}, config");
```
vusirikala, line 92:
```suggestion
    info!("Loaded node config: {:?}", config);
```

PR #8132, opened 2023-05-10
[TS] Added a missing dependency

junkil-park:
### Description

- Added "node-gyp": "9.3.1" because it is required in building when using node 20
- Removed "ref-array-di": "^1.2.2" because it's not used any longer.

### Test Plan
pnpm i


0xjinn (approved):
Thank you for fixing the issue!

junkil-park:
I close this PR, which is not needed.

0xmaayan:
not completely sure that I understand `Added "node-gyp": "9.3.1" because it is required in building when using node 20` , node 20 requires `node-gyp`?

PR #8134, opened 2023-05-10
[move-examples] Recreate mint_nft example and unit tests with TokenV2

xbtmatt:
### Description

This is a tutorial intended to mirror the `mint_nft` example in aptos-move/move-examples as closely as possible, with the only difference being that we're using the no code token from aptos_token to demonstrate how to do the tutorial with token v2 instead of token v1.

All 4 parts have been updated to reflect the new flow of creating a token. The tutorial comments that explain each file have also been updated and tailored to the new tutorial.

### Test Plan

The corresponding unit tests are located in aptos-move/e2e-move-tests/src/tests/mint_nft_v2.rs

You have to run with a larger stack size for some tests due to the test harness

```
RUST_MIN_STACK=4297152 cargo test mint_nft_v2_e2e -- --nocapture
RUST_MIN_STACK=4297152 cargo test sample_mint_nft_v2_part4_unit_test_signature -- --nocapture

cargo test generate_nft_v2_tutorial_part4_signature -- --nocapture
```

### DIFF
--- a/aptos-move/framework/aptos-framework/sources/aggregator/aggregator_factory.move
+++ b/aptos-move/framework/aptos-framework/sources/aggregator/aggregator_factory.move
@@ -7,7 +7,7 @@ module aptos_framework::aggregator_factory {
     use std::error;
 
     use aptos_framework::system_addresses;
-    use aptos_std::aggregator::Aggregator;
+    use aptos_framework::aggregator::Aggregator;
     use aptos_std::table::{Self, Table};
 
     friend aptos_framework::genesis;
--- a/aptos-move/framework/aptos-framework/sources/block.spec.move
+++ b/aptos-move/framework/aptos-framework/sources/block.spec.move
@@ -1,6 +1,6 @@
 spec aptos_framework::block {
     spec module {
-        use aptos_std::chain_status;
+        use aptos_framework::chain_status;
         // After genesis, `BlockResource` exist.
         invariant [suspendable] chain_status::is_operating() ==> exists<BlockResource>(@aptos_framework);
     }
--- a/aptos-move/framework/aptos-framework/sources/object.move
+++ b/aptos-move/framework/aptos-framework/sources/object.move
@@ -21,11 +21,12 @@ module aptos_framework::object {
     use std::signer;
     use std::vector;
 
+    use aptos_std::from_bcs;
+
     use aptos_framework::account;
     use aptos_framework::transaction_context;
     use aptos_framework::create_signer::create_signer;
     use aptos_framework::event;
-    use aptos_framework::from_bcs;
     use aptos_framework::guid;
 
     friend aptos_framework::primary_fungible_store;
--- a/aptos-move/framework/aptos-framework/sources/staking_contract.spec.move
+++ b/aptos-move/framework/aptos-framework/sources/staking_contract.spec.move
@@ -309,7 +309,7 @@ spec aptos_framework::staking_contract {
     }
 
     spec schema IncreaseLockupWithCapAbortsIf{
-        use std::timestamp;
+        use aptos_framework::timestamp;
         staker: address;
         operator: address;
 
--- a/aptos-move/framework/aptos-framework/sources/staking_proxy.spec.move
+++ b/aptos-move/framework/aptos-framework/sources/staking_proxy.spec.move
@@ -25,8 +25,8 @@ spec aptos_framework::staking_proxy {
     }
 
     spec set_staking_contract_operator(owner: &signer, old_operator: address, new_operator: address) {
+        use aptos_std::simple_map;
         use aptos_framework::staking_contract::{Store};
-        use aptos_framework::simple_map;
         // TODO: Verify timeout and can't verify `staking_contract::switch_operator`.
         pragma aborts_if_is_partial;
 
--- a/aptos-move/framework/aptos-framework/sources/state_storage.spec.move
+++ b/aptos-move/framework/aptos-framework/sources/state_storage.spec.move
@@ -1,6 +1,6 @@
 spec aptos_framework::state_storage {
     spec module {
-        use aptos_std::chain_status;
+        use aptos_framework::chain_status;
         pragma verify = true;
         pragma aborts_if_is_strict;
         // After genesis, `StateStorageUsage` and `GasParameter` exist.
@@ -21,7 +21,7 @@ spec aptos_framework::state_storage {
     }
 
     spec on_new_block(epoch: u64) {
-        use aptos_std::chain_status;
+        use aptos_framework::chain_status;
         requires chain_status::is_operating();
         aborts_if false;
         ensures epoch == global<StateStorageUsage>(@aptos_framework).epoch;
--- a/aptos-move/framework/aptos-framework/sources/storage_gas.spec.move
+++ b/aptos-move/framework/aptos-framework/sources/storage_gas.spec.move
@@ -27,7 +27,7 @@ spec aptos_framework::storage_gas {
     // -----------------
 
     spec module {
-        use aptos_std::chain_status;
+        use aptos_framework::chain_status;
         pragma verify = true;
         pragma aborts_if_is_strict;
         // After genesis, `StateStorageUsage` and `GasParameter` exist.
@@ -125,7 +125,7 @@ spec aptos_framework::storage_gas {
 
     /// Address @aptos_framework must exist StorageGasConfig and StorageGas and StateStorageUsage.
     spec on_reconfig {
-        use aptos_std::chain_status;
+        use aptos_framework::chain_status;
         requires chain_status::is_operating();
         aborts_if !exists<StorageGasConfig>(@aptos_framework);
         aborts_if !exists<StorageGas>(@aptos_framework);
--- a/aptos-move/framework/aptos-stdlib/Move.toml
+++ b/aptos-move/framework/aptos-stdlib/Move.toml
@@ -5,7 +5,6 @@ version = "1.0.0"
 [addresses]
 std = "0x1"
 aptos_std = "0x1"
-aptos_framework = "0x1"
 Extensions = "0x1" # For Prover to instantiate `{{Ext}}` in prelude.
 
 [dependencies]
--- a/aptos-move/framework/aptos-stdlib/sources/math128.move
+++ b/aptos-move/framework/aptos-stdlib/sources/math128.move
@@ -3,8 +3,8 @@ module aptos_std::math128 {
 
     use std::fixed_point32::FixedPoint32;
     use std::fixed_point32;
-    use std::fixed_point64::FixedPoint64;
-    use std::fixed_point64;
+    use aptos_std::fixed_point64::FixedPoint64;
+    use aptos_std::fixed_point64;
 
     /// Abort value when an invalid argument is provided.
     const EINVALID_ARG_FLOOR_LOG2: u64 = 1;


aptos-move/move-examples/mint_nft_v2/1-Create-NFT/sources/create_nft.move
```diff
@@ -0,0 +1,187 @@
+/// This is a tutorial intended to mirror the `mint_nft` example in move-examples as closely as possible, with the only difference
```
movekevin, line 1: I think it's better to just put these instructions on aptos.dev (may be https://aptos.dev/category/nft, which would need some reorg to account for token v2) and link to that in this module. This is not the best format to read the instructions anyway

aptos-move/move-examples/mint_nft_v2/1-Create-NFT/sources/create_nft.move
```diff
@@ -0,0 +1,187 @@
...
+/// being that we're using the no code token from aptos_token to demonstrate how to do the tutorial with token v2 instead of token v1.
```
movekevin, line 2: For after you turn this into a tutorial/doc on aptos.dev, we should just include the link to @gregnazario's doc on Token v2 which should already compare token v1 and v2. This tutorial shouldn't mention much about token v1 at all as it could be confusing to new devs who only see token v2 (which is the recommended approach now) and don't need to worry about v1.

aptos-move/move-examples/mint_nft_v2/1-Create-NFT/sources/create_nft.move
```diff
@@ -0,0 +1,187 @@
...
+///
```
movekevin, line 3: We should also include some brief explanations of the core parts of the Move code:
1. Ticket model (tickets are used to mint real tokens later on?)
2. Collection creation
3. Token minting
4. Property Metadata

aptos-move/move-examples/mint_nft_v2/1-Create-NFT/sources/create_nft.move
```diff
@@ -0,0 +1,187 @@
...
+    /// Action not authorized because the signer is not the admin of this module
+    const ENOT_AUTHORIZED: u64 = 1;
+
+    /// `init_module` is automatically called when publishing the module.
+    /// In this function, we create an example NFT collection and an example token.
+    fun init_module(source_account: &signer) {
+        let collection_name = string::utf8(b"Collection name");
+        let description = string::utf8(b"Description");
+        let collection_uri = string::utf8(b"Collection uri");
+        let token_name = string::utf8(b"Token name");
```
movekevin, line 115: Why not just make these constants? These are hardcoded anyway

aptos-move/move-examples/mint_nft_v2/1-Create-NFT/sources/create_nft.move
```diff
@@ -0,0 +1,187 @@
...
+            false, // mutable_description
+            false, // mutable_royalty
+            false, // mutable_uri
+            false, // mutable_token_description
+            false, // mutable_token_name
+            true, // mutable_token_properties
+            false, // mutable_token_uri
+            false, // tokens_burnable_by_creator
+            false, // tokens_freezable_by_creator
+            5, // royalty_numerator
```
movekevin, line 134: Let's add a quick comment here to explain that this is 5% royalty - 5% of all transfers should be sent to the creator.

aptos-move/move-examples/mint_nft_v2/1-Create-NFT/sources/create_nft.move
```diff
@@ -0,0 +1,187 @@
...
+            100, // royalty_denominator
+        );
+
+        let source_address = signer::address_of(source_account);
+        let collection_object_address = collection::create_collection_address(&source_address, &collection_name);
+
+        move_to(source_account, ModuleData {
```
movekevin, line 141: We should add a quick explanation here that we're storing these information in the ModuleData resource that will be stored with the code deployer's account. This is used later when minting the ticket tokens.

aptos-move/move-examples/mint_nft_v2/1-Create-NFT/sources/create_nft.move
```diff
@@ -0,0 +1,187 @@
...
+            module_data.token_name,
+            module_data.token_uri,
+            module_data.property_keys,
+            module_data.property_types,
+            module_data.property_values,
+        );
+        let token_object = object::address_to_object<AptosToken>(object::create_guid_object_address(module_owner_address, token_creation_num));
+        object::transfer(module_owner, token_object, receiver_address);
+
+        // update "given_to" to the value of the new receiver.
```
movekevin, line 178: Is this still useful in token v2?

aptos-move/move-examples/mint_nft_v2/2-Using-Resource-Account/Move.toml
```diff
@@ -0,0 +1,11 @@
...
+name = "Examples"
+version = "0.0.0"
+
+[dependencies]
+AptosFramework = { local = "../../../framework/aptos-framework" }
+AptosTokenObjects = { local = "../../../framework/aptos-token-objects" }
+
+[addresses]
+aptos_framework = "0x1"
+aptos_token_objects = "0x4"
```
movekevin, line 11: nit: Add an empty line at the end

aptos-move/move-examples/mint_nft_v2/2-Using-Resource-Account/sources/create_nft_with_resource_account.move
```diff
@@ -0,0 +1,201 @@
...
+            token_name,
+            token_uri,
+            property_keys: vector<String>[string::utf8(b"given_to")],
+            property_types: vector<String>[ string::utf8(b"address") ],
+            property_values: vector<vector<u8>>[bcs::to_bytes(&@source_addr)],
+            collection_object_address,
+        });
+    }
+
+    /// Mint an NFT to the receiver. Note the difference from the tutorial in part 1, here we only ask for the receiver's
```
movekevin, line 154: We technically need the receiver's signer here, not just address. The real diff between this and step 1 code is that this function can now be called by the user/receiver instead of the admin

aptos-move/move-examples/mint_nft_v2/1-Create-NFT/sources/create_nft.move
xbtmatt: Yeah I figured this is really not a great way to do it but was just getting the instructions out there without having to format it on aptos.dev and link to token v2 etc. I can remove these from the code and re-write and format it for the docs, or I can just link to [the current tutorial](https://aptos.dev/guides/nfts/mint-nft-cli/) or a dead link to `https://aptos.dev/guides/nfts/mint-nft-cli-v2`

aptos-move/move-examples/mint_nft_v2/2-Using-Resource-Account/sources/create_nft_with_resource_account.move
```diff
@@ -0,0 +1,201 @@
...
+            module_data.collection_name,
+            module_data.token_description,
+            module_data.token_name,
+            module_data.token_uri,
+            module_data.property_keys,
+            module_data.property_types,
+            module_data.property_values,
+        );
+
+        let token_object = object::address_to_object<AptosToken>(object::create_guid_object_address(resource_address, token_creation_num));
```
movekevin, line 181: Hmm this adds more blockers for parallelism (beside the supply being tracked) as it uses a single account's guid. Let's add a quick comment to mention this and we can come back and fix this later once we have the pieces ready for full parallelism

aptos-move/move-examples/mint_nft_v2/2-Using-Resource-Account/sources/create_nft_with_resource_account.move
```diff
@@ -0,0 +1,201 @@
...
+        let receiver_address = signer::address_of(receiver);
+        object::transfer(&resource_signer, token_object, receiver_address);
+
+        // remove the property key "given_to" entirely.
```
movekevin, line 185: Same as previous comment. Do we still need this?

aptos-move/move-examples/mint_nft_v2/4-Getting-Production-Ready/output.key
```diff
@@ -0,0 +1 @@
+C6FB9CC1022DF74BE74EEC60F93D0F0C3CC4F75C92DEEC4486BDFE57FE67A011
```
movekevin, line 1: What is this?

aptos-move/move-examples/mint_nft_v2/4-Getting-Production-Ready/sources/create_nft_getting_production_ready.move
```diff
@@ -0,0 +1,478 @@
...
+        // the resource account has the same authentication key as the source account so we had control over the resource account.
+        let resource_signer_cap = resource_account::retrieve_resource_account_cap(resource_signer, @source_addr);
+        let resource_address = signer::address_of(resource_signer);
+        let collection_object_address = collection::create_collection_address(&resource_address, &collection_name);
+
+        // hardcoded public key - we will update it to the real one by calling `set_public_key` from the admin account
+        let pk_bytes = x"f66bf0ce5ceb582b93d6780820c2025b9967aedaa259bdbb9f3d0297eced0e18";
+        let public_key = std::option::extract(&mut ed25519::new_validated_public_key_from_bytes(pk_bytes));
+        move_to(resource_signer, ModuleData {
+            public_key: public_key,
```
movekevin, line 216: nit: Just public_key,

aptos-move/move-examples/mint_nft_v2/4-Getting-Production-Ready/sources/create_nft_getting_production_ready.move
```diff
@@ -0,0 +1,478 @@
...
+///         Do you want to submit a transaction for a range of [540000 - 810000] Octas at a gas unit price of 100 Octas? [yes/no] >
+///         yes
+///         {
+///           "Result": "Success"
+///         }
+///
+///     Note the resource account address in the above output: a2534c87a046bc9dfe768a5f6d05ca9e5d5de7cc99aad60435f2325f6e9cf84c
+///
+///
+/// 3. Call `mint_event_ticket()` with a valid signature to mint a token.
```
movekevin, line 58: We need to explain the valid signature part more

aptos-move/move-examples/mint_nft_v2/4-Getting-Production-Ready/sources/create_nft_getting_production_ready.move
```diff
@@ -0,0 +1,478 @@
...
+            module_data.property_types,
+            module_data.property_values,
+        );
+
+        let token_object_address = object::create_guid_object_address(resource_address, token_creation_num);
+        let token_object = object::address_to_object<AptosToken>(object::create_guid_object_address(resource_address, token_creation_num));
+        object::transfer(&resource_signer, token_object, receiver_addr);
+
+        let object_addresses = borrow_global_mut<ObjectAddresses>(@mint_nft_v2);
+        if (option::is_some(&object_addresses.last_minted_token_object_address)) {
```
movekevin, line 277: Do you know what last_minted_token_object_address  is used for?

aptos-move/move-examples/mint_nft_v2/1-Create-NFT/sources/create_nft.move
xbtmatt: yea I'll add that and explain how you could pass in those arguments if you want but we'll use constants for simplicity

aptos-move/move-examples/mint_nft_v2/1-Create-NFT/sources/create_nft.move
xbtmatt: Showing who the owner is in the property map? No, not really. I'll change it to something random like "Date Minted" just to show how to use the property map still

aptos-move/move-examples/mint_nft_v2/2-Using-Resource-Account/sources/create_nft_with_resource_account.move
xbtmatt: if I'm not misunderstanding what you mean, I meant to just clarify that you can send the token object without the `receiver`s approval (signer) now, and the only reason we have the receiver as a signer is so they're approving of receiving the mint. I can explain it/word it better in the tutorial

movekevin: Explaining it better in the tutorial makes sense

aptos-move/move-examples/mint_nft_v2/1-Create-NFT/sources/create_nft.move
xbtmatt: Yea that's what I'll do. We're not going to keep/maintain the original tutorial/doc page right? Can I just move this to `mint_nft` and replace the contracts and docs for anything linked to it?

junkil-park:
Great job! Would this PR also update the content in `aptos.dev`? Where is the menu that leads to the tutorial of this example in `aptos.dev`?

developer-docs-site/docs/guides/nfts/mint-nft-cli.md
```diff
@@ -4,587 +4,291 @@ title: "Mint NFTs with the Aptos CLI"
...
 
-* a [GitHub account](https://docs.github.com/en/get-started/signing-up-for-github/signing-up-for-a-new-github-account)
-* the [GitHub CLI](https://cli.github.com/)
 * the [Aptos CLI](../../tools/install-cli/index.md) (or you can run from [aptos-core](https://github.com/aptos-labs/aptos-core) source via `cargo run`)
 * the `aptos-core` repository checked out: `git clone https://github.com/aptos-labs/aptos-core.git`
+* a general understanding of NFTs and NFT Collections
+* an understanding of [the Token V2 standard](https://aptos.dev/guides/nfts/token-v2)
 
-Then:
+## 1. Create a simple smart contract to mint an NFT
```
movekevin, line 18: Let's not override the v1 version of the doc and keep it around until we can fully deprecate token v1

developer-docs-site/docs/guides/nfts/mint-nft-cli.md
```diff
@@ -4,587 +4,291 @@ title: "Mint NFTs with the Aptos CLI"
...
 
-* Review the [mint_nft](https://github.com/aptos-labs/aptos-core/tree/main/aptos-move/move-examples/mint_nft) source code and code comments within each subdirectory.
-* Explore the `mint_event_ticket` function defined within each subdirectory.
+We're going to start by making the simplest form of the flow for creating a collection and minting a token and sending it to a user. The code for this part of the tutorial is in the first section of the `move-examples/mint_nft` folder in the aptos-core repository.
 
-## NFT types
+Here are the most basic things we need to do first:
```
movekevin, line 22: This sentence sounds a bit redundant. How about "Here are the things we need to do first:"

developer-docs-site/docs/guides/nfts/mint-nft-cli.md
```diff
@@ -4,587 +4,291 @@ title: "Mint NFTs with the Aptos CLI"
...
-The two most common types of NFT are event ticket / certificates and profile picture NFTs.
+* Create the NFT collection and store the configuration options for it
+* Mint a non-fungible token within that collection using the configuration options
+* Send the minted token to a user account
 
-### Event tickets and certificates
+### Defining configuration options
 
-This kind of NFT has a base token, and every new NFT generated from this base token has the same token data ID and image. They are generally used as certificates, meaning each NFT created from the base token is considered a printing edition of the base token.
+The first thing we need to do is store the collection information somewhere. These will be fields we get on-chain later to mint the token.
```
movekevin, line 30: I'd clarify what "collection information" includes

developer-docs-site/docs/guides/nfts/mint-nft-cli.md
```diff
@@ -4,587 +4,291 @@ title: "Mint NFTs with the Aptos CLI"
...
 
-You might use this type of NFT for event tickets where each NFT is a ticket and has properties representing expiration date and if the ticket has been used. When you mint the NFT, you may set an expiration time for the event ticket and `is_ticket_used` to `false`. When the ticket is used, you update `is_ticket_used` to `true`.
+We'll store it in a `struct` called `MintConfiguration`. These fields aren't the only fields used to create a collection, but they are the most basic ones meant to demonstrate how to store configuration details.
```
movekevin, line 32: This sentence can be merged into the previous sentence to say something along the line of:

We need to store collection information so we can mint tokens later such as:
[MintConfiguration struct code with annotated comments explaining the fields]

developer-docs-site/docs/guides/nfts/mint-nft-cli.md
```diff
@@ -4,587 +4,291 @@ title: "Mint NFTs with the Aptos CLI"
...
+```rust
+struct MintConfiguration has key {
+    collection_name: String,
+    token_name: String,
+    token_uri: String,
+}
 ```
 
-### Publish the module under a resource account
+We give `MintConfiguration` the `key` ability so that the module contract can store it at an account's resources, which the module can find later if supplied with the account address. The account address we store it to will be the creator of the collection.
```
movekevin, line 42: We should talk about how this MintConfiguration will be initially stored as a resource in the creator's account. So we can later say we'll move this to the resource account

developer-docs-site/docs/guides/nfts/mint-nft-cli.md
```diff
@@ -4,587 +4,291 @@ title: "Mint NFTs with the Aptos CLI"
...
 
-1. In your `aptos-core` source checkout, navigate to the `aptos-core/aptos-move/move-examples/mint_nft/3-Adding-Admin` directory:
+Some collection configuration parameters will be hard-coded in the contract as `const` variables:
 
-```shell
-cd aptos-move/move-examples/mint_nft/3-Adding-Admin
+```rust
```
movekevin, line 46: We might not need to have this code here - I don't think it adds much value

developer-docs-site/docs/guides/nfts/mint-nft-cli.md
```diff
@@ -4,587 +4,291 @@ title: "Mint NFTs with the Aptos CLI"
...
 
-2. Edit `Move.toml` in that directory to replace `admin_addr = "0xcafe"` with the new `admin` address from `.aptos/config.yaml`.
+You can make these configurable on your own by adding them to the list of parameters sent into `initialize_collection` function below.
 
-3. Run the following CLI command to publish the module under a resource account, updating the `<seed>` and `default-account-address>` with your own:
+### Writing a collection creation function
 
-```shell
-aptos move create-resource-account-and-publish-package --seed <seed> --address-name mint_nft --profile default --named-addresses source_addr=<default-account-address>
+```rust
```
movekevin, line 58: Let's not have this entire snippet here (any changes later would need to be made to both the code and this doc). Just the aptos_token::create_collection call should be good.

developer-docs-site/docs/guides/nfts/mint-nft-cli.md
```diff
@@ -4,587 +4,291 @@ title: "Mint NFTs with the Aptos CLI"
...
-```shell
-"Result": "Success"`
+```rust
+let creator_addr = signer::address_of(creator);
+assert!(creator_addr == @mint_nft, error::permission_denied(ENOT_AUTHORIZED));
 ```
 
-### Include admin functions
+This gates the ability to run this function to the owner of the contract. This isn't necessary, but unless you specifically design a contract around
+letting multiple accounts use it, it's simpler to only allow the account that deploys the contract to use it. This avoids a few potential vulnerabilities where you unintentionally allow an account to modify internal module resources for another account.
```
movekevin, line 107: Let's mention we'll show to migrate from deployer-only to an admin model in later steps (parts?)

developer-docs-site/docs/guides/nfts/mint-nft-cli.md
```diff
@@ -4,587 +4,291 @@ title: "Mint NFTs with the Aptos CLI"
...
+:::
 
+After that, we see the logic for our very basic initialize collection function. It creates a collection based off the parameters given in the
+function arguments and then stores the minting config info to the creator's account resources with `move_to`.
 
-* In struct `ModuleData`, we add two additional fields: `expiration_timestamp` and `minting_enabled`. These allow us to set and update when this collection will expire and enable or disable minting ad-hoc, respectively.
-* We included two admin functions - `set_minting_enabled()` and `set_timestamp()` - to update the `expiration_timestamp` and `minting_enabled` fields.
-* In the admin functions, we check if the caller is calling from the valid administrator's address. If not, we abort because the caller does not have permission to update the config of this module.
-* In `mint_event_ticket()`, we add two `assert` statements to make sure the user can mint tokens from this collection only if minting is enabled and the collection has not expired.
+Some of these fields are self-explanatory, but here's a quick rundown of what you're actually setting here:
```
movekevin, line 116: Might as well just link to the collection.move code for a detailed explanation of the fields?

aptos-move/move-examples/mint_nft_v2/4-Getting-Production-Ready/sources/create_nft_getting_production_ready.move
xbtmatt: I was using it to view it myself, meant to take it out, hah

aptos-move/move-examples/mint_nft_v2/4-Getting-Production-Ready/output.key
xbtmatt: To be quite honest, I didn't fully understand the purpose of generating a public key in the old tutorial. I think it may have been a relic of a convoluted process to test the module, since there is no easy way to generate a public key and private key in a `[#test]` function

xbtmatt:
> Great job! Would this PR also update the content in `aptos.dev`? Where is the menu that leads to the tutorial of this example in `aptos.dev`?

I'm not sure how we're going to change the structure to add it, since we're going to be keeping the old one. For now, I'm going to show it next to the other one, just add a `v2` to the name:

`Mint NFTs (v2) with the Aptos CLI`

What do you guys think? @movekevin

aptos-move/move-examples/mint_nft_v2/1-Create-NFT/sources/create_nft.move
movekevin: I'd prefer not replacing the existing tutorial for minting token v1. Let's keep the tutorial for token v2 completely separate (even if there's some redundancy or copy pasta)

developer-docs-site/docs/guides/nfts/mint-nft-cli.md
xbtmatt: Yea, you're right. I think I was going to explain the fields more intuitively but ended up mostly just explaining what's already in collection.move

I'll leave the explanation of the royalty numerator/denominator and remove the rest

developer-docs-site/docs/guides/nfts/mint-nft-cli.md
xbtmatt: Yea I think I overexplained a bit, hah. Removed it

aptos-move/move-examples/mint_nft_v2/4-Getting-Production-Ready/output.key
movekevin: > > Great job! Would this PR also update the content in `aptos.dev`? Where is the menu that leads to the tutorial of this example in `aptos.dev`?
> 
> I'm not sure how we're going to change the structure to add it, since we're going to be keeping the old one. For now, I'm going to show it next to the other one, just add a `v2` to the name:
> 
> `Mint NFTs (v2) with the Aptos CLI`
> 
> the url would be https://aptos.dev/guides/nfts/mint-nft-v2-cli
> 
> What do you guys think? @movekevin

That sounds fine on paper. I'd just make the change and we can then look at preview to see if it makes sense

developer-docs-site/docs/guides/nfts/mint-nft-v2-cli.md
```diff
@@ -0,0 +1,918 @@
...
+
+To achieve this, in this section we'll show you how to:
+
+- Create the NFT collection with a resource account
+- Store the capability to sign things with the resource account, a `SignerCapability`, into the owner's resources on-chain
+- Automate minting the token to the user; that is, write a mint function that works without the collection creator's signature
+
+### What is a resource account?
+
+A resource account is essentially an account that another account can own. They are useful for separating and managing different types of resources, but they're also capable of delegating decisions to sign transactions to the logic in a smart contract.
```
movekevin, line 265: This is not the best way to explain resource accounts IMO. Especially for this example, the resource account is actually used as a way to store resources and control (over the collection) that will later be used by the module itself.

developer-docs-site/docs/guides/nfts/mint-nft-v2-cli.md
```diff
@@ -0,0 +1,918 @@
...
+
+Now, we can require the receiver to sign so that a user can mint whenever they like, and the owner doesn't have to approve of it beforehand.
+
+```rust
+public entry fun mint(receiver: &signer, resource_addr: address) acquires MintConfiguration {
+    //...
+}
+```
+
+Note that we require the user to pass in the resource address- we've provided a view function as one way for you to calculate it. We show you how to use this function later in the [Running the contract section](#running-the-contract-1).
```
movekevin, line 334: I wonder if we can just deploy the module at the resource account instead. This would remove the requirement that the user needs to pass in the resource account address

developer-docs-site/docs/guides/nfts/mint-nft-v2-cli.md
```diff
@@ -0,0 +1,918 @@
...
+
+In the first section, the user has to wait for the owner of the contract to mint and send them an NFT. In the second section, the user can request to mint and receive an NFT themselves.
+
+## 3. Adding restrictions: a whitelist, an end time, an admin, and an enabled flag
+
+We're still missing some very common features for NFT minting contracts:
+
+1. A whitelist that restricts minting to whitelisted addresses
+2. The ability to add/remove addresses from the whitelist
+3. A start* and end time
```
movekevin, line 419: Why not just say "An end time" then if start time is not being added :)

developer-docs-site/docs/guides/nfts/mint-nft-v2-cli.md
```diff
@@ -0,0 +1,918 @@
...
+
+When we initialize the collection, we create a default empty whitelist, an expiration timestamp that's one second in the past, and disable the mint:
+
+```rust
+public entry fun initialize_collection( /* ... */ ) {
+    // ...
+
+    move_to(&resource_signer, MintConfiguration {
+        // ...
+        whitelist: table::new<address, bool>(),
```
movekevin, line 449: Let's use smart_table here instead

aptos-move/move-examples/mint_nft_v2/3-Adding-Admin/sources/create_nft_with_resource_and_admin_accounts.move
```diff
@@ -0,0 +1,201 @@
...
+
+    use aptos_token_objects::aptos_token::{Self, AptosToken};
+
+    // This struct stores an NFT collection's relevant information
+    struct MintConfiguration has key {
+        signer_capability: SignerCapability,
+        collection_name: String,
+        token_name: String,
+        token_uri: String,
+        whitelist: Table<address, bool>,
```
movekevin, line 20: Let's use smart_table here instead for better gas and storage efficiency

aptos-move/move-examples/mint_nft_v2/3-Adding-Admin/sources/create_nft_with_resource_and_admin_accounts.move
```diff
@@ -0,0 +1,201 @@
...
+    ) acquires MintConfiguration {
+        let mint_configuration = borrow_global_mut<MintConfiguration>(resource_addr);
+        let admin_addr = signer::address_of(admin);
+        // abort if the signer is not the admin
+        assert!(admin_addr == mint_configuration.admin, error::permission_denied(ENOT_AUTHORIZED));
+        mint_configuration.expiration_timestamp = expiration_timestamp;
+    }
+
+    #[view]
+    public fun get_resource_address(collection_name: String): address {
```
movekevin, line 198: Shouldn't this just be @mint_nft_v2_part3?

movekevin (approved):
This is coming together really amazingly! Beside a few comments here and there, I think this is good to go. You can iterate on this more after gathering feedback from the community

developer-docs-site/docs/guides/nfts/mint-nft-v2-cli.md
xbtmatt: Yeah, but I was trying to move away from the flow of publishing modules to resource accounts because of how difficult it is to upgrade. When I was doing the tutorial myself it was especially frustrating when I screwed anything up and had to re-deploy to a new account because of a typo or because I messed with something, haha

I also considered just storing it in a resource, but then I realized whenever I actually used a minting contract in an e2e app I always ended up generating the resource address and storing it in state at some point, so it didn't matter, especially with an easy `#[view]` function for it

developer-docs-site/docs/guides/nfts/mint-nft-v2-cli.md
xbtmatt: I was trying to clarify that a start time is usually needed, but that we weren't adding one yet, lol. I'll remove, it's kinda pointless to have

developer-docs-site/docs/guides/nfts/mint-nft-v2-cli.md
movekevin: It's not a great UX IMO that the user needs to pass the resource account address every time. Upgrading is not great with resource accounts currently for sure but (1) that's a problem for another tutorial :) and (2) we can improve the flow (it's part of our Q2 Move roadmap actually)

developer-docs-site/docs/guides/nfts/mint-nft-v2-cli.md
xbtmatt: I use it in the next section! Is it confusing if I go from `table` to `smart_table` or you think I should swap it still

aptos-move/move-examples/mint_nft_v2/3-Adding-Admin/sources/create_nft_with_resource_and_admin_accounts.move
xbtmatt: No because I create the resource account with a seed
 
```
        assert!(owner_addr == @mint_nft_v2_part2, error::permission_denied(ENOT_AUTHORIZED));

        let seed = *string::bytes(&collection_name);
        let (resource_signer, resource_signer_cap) = account::create_resource_account(owner, seed);
```

I could use an empty seed, but I always try to make sure the seed + signer (for creating the resource account) is a unique combination as a good habit, in this case creator + collection because I assert the creator is also the module deployer

But mainly this is just due to not using the `publish module to resource account` flow

movekevin: I still think we should publish the module to the resource account in this example. Users having to pass in the resource address reminds me of one similar pain point in Sui where users have to pass in the object address where the contract state is stored. This is a rather painful UX IMO as the addresses would need to be hardcoded in all off chain components (FE, indexing, explorer, etc.) for users to interact or find the data associated.

I think it's worthwhile in our system to eliminate this pain if possible so that common data of a contract is stored along with the module's code and is thus much easier to find and interact with

developer-docs-site/docs/guides/nfts/mint-nft-v2-cli.md
xbtmatt: Hm yea, this is not a particularly good explanation. It's actually pretty confusing in its current form, I'll revise it.

developer-docs-site/docs/guides/nfts/mint-nft-v2-cli.md
xbtmatt: Ya I agree, I may have been overthinking it in this case, but my thought process was the above + there's a comment right above the function call to generate a resource address that says:

>  /// This is a helper function to compute resource addresses. Computation of the address
>  /// involves the use of a cryptographic hash operation and should be use thoughtfully.

So I tried to avoid computing it for every single mint transaction.

I was going to store it on contract account itself, which in this case makes sense, but I didn't want to encourage a bad habit of storing resources there.

In this case, it probably makes the most sense to store it on the owner's account, which in this case is the contract account anyway, yea?

aptos-move/move-examples/mint_nft_v2/3-Adding-Admin/sources/create_nft_with_resource_and_admin_accounts.move
xbtmatt: Okay so I thought through this-originally with an admin model I thought that you could lose track of the original creator, in which case you'd still need to provide the original creator address + collection name.

However that's only the case if the module supports multiple collections from different "owners" otherwise you only need the collection name (and not even that if the module only supports a single collection), because if you restrict it to a single "owner" you can just store that information at the contract address

aptos-move/move-examples/mint_nft_v2/3-Adding-Admin/sources/create_nft_with_resource_and_admin_accounts.move
davidiw: @xbtmatt

developer-docs-site/docs/guides/nfts/mint-nft-v2-cli.md
davidiw: always good to provide a link to any dedicated material if we have it.

developer-docs-site/docs/guides/nfts/mint-nft-v2-cli.md
```diff
@@ -0,0 +1,918 @@
...
+```rust
+
+```
+
+For the sake of brevity, we'll only explain a single example of a negative test condition here. We'll test that setting a new admin results in the old admin being unable to call admin-only functions:
+
+```rust
+
+```
+
```
davidiw, line 900: hmm?

davidiw (commented):
unblocking but needs some work still

developer-docs-site/docs/guides/nfts/mint-nft-v2-cli.md
xbtmatt: I think I went overboard with some of the explanations here and tried to fit way too much into one PR...

I'm going to revamp a lot of this stuff (mostly by removing lots of content that will be linked in other sections of the docs) and leave the unit tests out of the tutorial explanations entirely.

developer-docs-site/docs/guides/nfts/mint-nft-v2-cli.md
```diff
@@ -0,0 +1,918 @@
...
+This tutorial assumes you have:
+
+* the [Aptos CLI](../../tools/install-cli/index.md) (or you can run from [aptos-core](https://github.com/aptos-labs/aptos-core) source via `cargo run`)
+* the `aptos-core` repository checked out: `git clone https://github.com/aptos-labs/aptos-core.git`
+* a basic understanding of Move, NFTs and NFT Collections
+
+## 1. Creating a simple smart contract to mint an NFT
+
+We're going to start by making the simplest form of the flow for creating a collection and minting a token and sending it to a user. The code for this part of the tutorial is in the first section of the `move-examples/mint_nft_v2_part1` folder in the aptos-core repository.
```
davidiw, line 25: might be good to add a clickable link to this.

developer-docs-site/docs/guides/nfts/mint-nft-v2-cli.md
davidiw: I mean isn't this more akin to minting from an object?

We're passing in our mint resource?

Also have we considered using an object instead of a resource account? It seems to be a more sensible approach ... considering.

davidiw (commented):
if you don't really want to use the resource account, would definitely consider moving the minting state to an object.

PR #8135, opened 2023-05-10
[Aptos Data Client] Clean up names and split files.

JoshLind:
### Description

The Aptos Data Client needs a lot of love. This PR starts by cleaning up some of the struct names, and splits up the large monolithic files into separate files, so that the structure is a little simpler. Note: nothing should change logically in this PR (it's just renames and file splitting).

This PR offers several commits:
1. Rename `summary_poll_interval_ms` to `summary_poll_loop_interval_ms` to signify that the interval is for the loop execution.
2. Move the `DataSummaryPoller` into a separate file (`poller.rs`).
3. Move the `AptosNetDataClient` into a separate file (`client.rs`).
4. Move the `Error` struct (and related code) into a separate file (`error.rs`).
5. Rename `AptosDataClient` to `AptosDataClientInterface` (as the declaration defines the trait interface).
6. Move the `GlobalDataSummary` into a separate file (`global_summary.rs`).
7. Move the `AptosDataClientInterface` (and other interface code) into a separate file (`interface.rs`).
8. Remove the unnecessary `aptosnet` directory.
9. Rename `AptosNetDataClient` to `AptosDataClient` (as this is the actual client implementation used by the node!).
10. Rename `state.rs` to `peer_states.rs` (to make it more obvious that we're talking about the peer states).

### Test Plan
Existing test infrastructure.


msmouse (approved):
i'm blind

brianolson (approved):

PR #8137, opened 2023-05-10
[single node benchmark] Breakdown of time spent across phases

igor-aptos:
### <samp>🤖 Generated by Copilot at 741d478</samp>

This pull request enhances the executor benchmark and its testing script. It adds metrics and logging to `lib.rs` to measure and report the executor performance. It also refactors and improves `single_node_performance.py` to run, parse, and print the benchmark output more efficiently and clearly. It also lowers the benchmark size for faster testing.

### DIFF
--- a/execution/executor-benchmark/src/lib.rs
+++ b/execution/executor-benchmark/src/lib.rs
@@ -21,7 +21,13 @@ use crate::{
 };
 use aptos_config::config::{NodeConfig, PrunerConfig};
 use aptos_db::AptosDB;
-use aptos_executor::block_executor::{BlockExecutor, TransactionBlockExecutor};
+use aptos_executor::{
+    block_executor::{BlockExecutor, TransactionBlockExecutor},
+    metrics::{
+        APTOS_EXECUTOR_COMMIT_BLOCKS_SECONDS, APTOS_EXECUTOR_EXECUTE_BLOCK_SECONDS,
+        APTOS_EXECUTOR_OTHER_TIMERS_SECONDS, APTOS_EXECUTOR_VM_EXECUTE_BLOCK_SECONDS,
+    },
+};
 use aptos_jellyfish_merkle::metrics::{
     APTOS_JELLYFISH_INTERNAL_ENCODED_BYTES, APTOS_JELLYFISH_LEAF_ENCODED_BYTES,
 };
@@ -34,6 +40,7 @@ use aptos_vm::counters::TXN_GAS_USAGE;
 use gen_executor::DbGenInitTransactionExecutor;
 use pipeline::PipelineConfig;
 use std::{
+    collections::HashMap,
     fs,
     path::Path,
     sync::{atomic::AtomicUsize, Arc},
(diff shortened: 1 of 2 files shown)


execution/executor-benchmark/src/lib.rs
```diff
@@ -146,6 +146,27 @@ pub fn run_benchmark<V>(
...
+        ("2.2.2.", false, "calculate_block_state_updates"),
+        ("2.2.3.", false, "calculate_usage"),
+        ("2.2.4.", false, "make_checkpoint"),
+        ("2.3.", false, "assemble_ledger_diff_for_block"),
+        ("2.3.1.", false, "calculate_events_and_writeset_hashes"),
+        ("3.", true, "as_state_compute_result"),
+        ("4.", true, "get_txns_to_commit"),
+    ];
+
+    let start_by_other = other_labels.iter().map(|other_label| (other_label.2.to_string(), APTOS_EXECUTOR_OTHER_TIMERS_SECONDS.with_label_values(&[other_label.2]).get_sample_sum())).collect::<HashMap<_, _>>();
```
msmouse, line 167: nit: I hate to access tuple fields by `.n` and prefer giving them names by assignment. And looks like `cargo fmt` is needed.

testsuite/single_node_performance.py
```diff
@@ -74,16 +76,132 @@
...
+        text=True,
+        cwd=target_directory,
+        stdout=PIPE,
+        bufsize=1,
+        universal_newlines=True,
+    ) as p:
+        # stream to output while command is executing
+        for line in p.stdout:
+            print(line, end="")
+            result.append(line)
```
msmouse, line 92: why not just `capture_output=True`?

testsuite/single_node_performance.py
```diff
@@ -74,16 +76,132 @@
...
+    single_node_result: RunResults
+    concurrency_level_results: dict[int, RunResults]
+    block_size: int
+    expected_tps: float
+
+
+def extract_run_results(output: str, execution_only: bool) -> RunResults:
+    if execution_only:
+        tps = float(re.findall(r"Overall execution TPS: (\d+\.?\d*) txn/s", output)[-1])
+        gps = float(re.findall(r"Overall execution GPS: (\d+\.?\d*) gas/s", output)[-1])
```
msmouse, line 129: :trollface:  should really just print integers.

msmouse (approved):

execution/executor-benchmark/src/lib.rs
```diff
@@ -174,8 +195,19 @@ pub fn run_benchmark<V>(
...
+    info!("Overall TPS: {} txn/s", delta_v as f64 / elapsed);
+    info!("Overall GPS: {} gas/s", delta_gas as f64 / elapsed);
+
+    let time_in_execution = APTOS_EXECUTOR_EXECUTE_BLOCK_SECONDS.get_sample_sum() - start_execution_total;
+    info!("Overall fraction of total: {:.3} in execution", time_in_execution / elapsed);
+    info!("Overall fraction of execution {:.3} in VM", (APTOS_EXECUTOR_VM_EXECUTE_BLOCK_SECONDS.get_sample_sum() - start_vm_only) / time_in_execution);
+    for (prefix, top_level,  other_label) in other_labels {
+        let fraction_in_label = (APTOS_EXECUTOR_OTHER_TIMERS_SECONDS.with_label_values(&[other_label]).get_sample_sum() - start_by_other.get(other_label).unwrap()) / time_in_execution;
+        if top_level || fraction_in_label > 0.01 {
+            info!("Overall fraction of execution {:.3} in {} {}", fraction_in_label, prefix, other_label);
```
sitalkedia, line 207: nit: I find TPS metrics more intuitive than execution time. Thoughts on output TPS per state in addition to execution time or may be just the TPS?

testsuite/single_node_performance.py
igor-aptos: command runs for a while, and with capture_output=True we are not able to see incremental progress. 

Also if there is a deadlock, nothing will be printed.

testsuite/single_node_performance.py
igor-aptos: we are printing integers in the aggregate table in the regression test. When you are doing single run, having more precision might be useful (or if you have workload that is in <100 tps.

execution/executor-benchmark/src/lib.rs
igor-aptos: @sitalkedia  - added

execution/executor-benchmark/src/lib.rs
igor-aptos: updated. forgot this works in lambdas in rust

sitalkedia (approved):
LGTM

grao1991:
I've made some changes on metrics, can you rebase on https://github.com/aptos-labs/aptos-core/pull/8180?

PR #8142, opened 2023-05-11
[api] for the debugger: get_raw_state_value

msmouse:
To support getting resource groups and state metadata.

### DIFF
--- a/api/src/state.rs
+++ b/api/src/state.rs
@@ -14,8 +14,8 @@ use crate::{
 use anyhow::Context as AnyhowContext;
 use aptos_api_types::{
     verify_module_identifier, Address, AptosErrorCode, AsConverter, IdentifierWrapper,
-    MoveModuleBytecode, MoveResource, MoveStructTag, MoveValue, RawTableItemRequest,
-    TableItemRequest, VerifyInput, VerifyInputWithRecursion, U64,
+    MoveModuleBytecode, MoveResource, MoveStructTag, MoveValue, RawStateValueRequest,
+    RawTableItemRequest, TableItemRequest, VerifyInput, VerifyInputWithRecursion, U64,
 };
 use aptos_state_view::TStateView;
 use aptos_types::{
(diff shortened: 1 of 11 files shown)


runtian-zhou (commented):
Thank you! Would love to see an api like that as long as others agree on this.

grao1991:
Shell we expose this as a "debug API" and guard it behind some flags, and add a note saying it's not considered as part of our "stable API".  I want to leave some flexibility here.

msmouse:
@banool  is there a way to mark an endpoint "not promised, do not use"?

davidiw:
> @banool is there a way to mark an endpoint "not promised, do not use"?

if there isn't one, I might recommend we just put it on something like host/experimental/API in contrast to host/v1/API

api/src/state.rs
```diff
@@ -214,6 +214,43 @@ impl StateApi {
...
+    /// in the request body.
+    ///
+    /// The Aptos nodes prune account state history, via a configurable time window.
+    /// If the requested ledger version has been pruned, the server responds with a 410.
+    #[oai(
+        path = "/state_values/raw",
+        method = "post",
+        operation_id = "get_raw_state_value",
+        tag = "ApiTags::StateValues"
+    )]
```
banool, line 230:
```suggestion
        tag = "ApiTags::StateValues",
        hidden
    )]
```

banool (commented):
There is no way to mark it as "do not use" but if we upgrade our version of poem (https://github.com/aptos-labs/aptos-core/pull/8197) it is possible to hide an endpoint. I've left a suggestion in the code.

msmouse:
Thanks @banool. I added "hidden" and changed the tag. PTAL

msmouse:
moved to `/experimental/stata_values/raw`

grao1991 (approved):

areshand (approved):

banool (approved):
Looks good, nice nice.

PR #8143, opened 2023-05-11
[Docker] Pin also libtinfo6 Debian package

sionescu:
The lack of a pin is causing failures([example](https://github.com/aptos-labs/aptos-core/actions/runs/4942740670/jobs/8836548806
)) because apt likes to update all packages and fails to backtrack to older versions if some dependencies are pinned.

The long-term solution is to figure out a way to avoid pinning packages altogether, especially foundation packages like perl-base and libtinfo, because the pins are fragile and will break often.

### DIFF
--- a/docker/experimental/tools.Dockerfile
+++ b/docker/experimental/tools.Dockerfile
@@ -11,6 +11,7 @@ RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
     wget \
     curl \
     perl-base=5.32.1-4+deb11u1 \
+    libtinfo6=6.2+20201114-2+deb11u1 \
     git \
     libssl1.1 \
     ca-certificates \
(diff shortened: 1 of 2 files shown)


rustielin (approved):

docker/experimental/tools.Dockerfile
```diff
@@ -11,6 +11,7 @@ RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
     wget \
     curl \
     perl-base=5.32.1-4+deb11u1 \
+    libtinfo6=6.2+20201114-2+deb11u1 \
```
perryjrandall, line 14: who is going to update these? (when do they become stale?)

cc @gedigi

sionescu: I tried removing the pin in perl-base before this, and the install failed with another unresolved dependency on libncursesw. My guess is that either python3-botocore/bullseye or awscli/bullseye depends on older packages and is not getting rebuilt frequently to keep up with Debian.
This is definitely a bad thing and we should try to avoid it but on the short term this is the best we can do.

Issue #7731, opened 2023-04-13
[Bug] Logic to determine a script in move model is broken


wrwg:
There is a bug report about this here: https://github.com/move-language/move/issues/1033

But the bug goes even further: it appears that logic for determining what is a script is broken independent of address size. This not only effects the ABI generator but also some verification logic for scripts in the Move prover.

PR #8151, opened 2023-05-11
[model compiler] Refactor move-model to be independent of bytecode

wrwg:
This is a large refactor of the `move-model` to become independent of bytecode, and an essential step towards the new compiler.

Originally the move-model was designed to work from bytecode as generated by the Move compiler. As such it was input the prover, docgen, and other tools. This PR makes this dependency optional:

- The model can still operate base off bytecode, supporting all exisiting use cases and tests. This is important for the incremental introduction of the new compiler infra.
- In a new execution path enabled by this PR, the model can also operate without having bytecode as input. In this case, the model builder will completely compile the model from the Move sources.

The focus of this PR is preparing the 2nd operation mode without breaking the first. There are further steps needed for making the 2nd mode fully functional.

This PR also removes a number of unused crates to simplify refactoring:

- stackless bytecode interpreter
- legacy read-write set analysis
- legacy escape analysis

### Test Plan

Existing tests


third_party/move/move-model/tests/testsuite.rs
```diff
@@ -3,22 +3,16 @@
...
-};
+use move_model::{options::ModelBuilderOptions, run_model_builder_with_options};
 use move_prover_test_utils::baseline_test::verify_or_update_baseline;
 use std::path::Path;
 
 fn test_runner(
     path: &Path,
     options: ModelBuilderOptions,
-    check_from_bytecode: bool,
+    _check_from_bytecode: bool,
```
davidiw, line 15: why was this removed and what are the implications?

third_party/move/move-model/src/lib.rs
```diff
@@ -317,54 +314,6 @@ fn collect_related_modules_recursive<'a>(
     }
 }
 
-/// Build a `GlobalEnv` from a collection of `CompiledModule`'s. The `modules` list must be
-/// topologically sorted by the dependency relation (i.e., a child node in the dependency graph
-/// should appear earlier in the vector than its parents).
-pub fn run_bytecode_model_builder<'a>(
```
davidiw, line 323: what was the motivation in removing this?

davidiw (commented):
This is amazing step forward!

Could we consider breaking this up into a few commits? Maybe others have deep familiarity of the code base and can swiftly cut through the noise, but I can't. Right now, I can see at least 5 different distinct pieces that I'd love to be able to review as independently as possible: 
* removal of stackless bytecode interpreter
* removal of legacy read-write set analysis
* removal of legacy escape analysis
* some updates of documentation and other clean ups
* the actual refactor of the move-model

Maybe others will have more time to dig through this.

wrwg:
> This is amazing step forward!
> 
> Could we consider breaking this up into a few commits? Maybe others have deep familiarity of the code base and can swiftly cut through the noise, but I can't. Right now, I can see at least 5 different distinct pieces that I'd love to be able to review as independently as possible:
> 
> * removal of stackless bytecode interpreter
> * removal of legacy read-write set analysis
> * removal of legacy escape analysis
> * some updates of documentation and other clean ups
> * the actual refactor of the move-model
> 
> Maybe others will have more time to dig through this.

The removal of full crates doesn't add mental overhead, so I see no practical reason why to split those parts off.

The other things are unfortunately interconnected and cannot be easily serialized. However, this is really just a refactor. There is no deep review needed IMO. Understanding this PR does not contribute to understanding the new compiler.

wrwg:
> > This is amazing step forward!
> > Could we consider breaking this up into a few commits? Maybe others have deep familiarity of the code base and can swiftly cut through the noise, but I can't. Right now, I can see at least 5 different distinct pieces that I'd love to be able to review as independently as possible:
> > 
> > * removal of stackless bytecode interpreter
> > * removal of legacy read-write set analysis
> > * removal of legacy escape analysis
> > * some updates of documentation and other clean ups
> > * the actual refactor of the move-model
> > 
> > Maybe others will have more time to dig through this.
> 
> The removal of full crates doesn't add mental overhead, so I see no practical reason why to split those parts off.

Forgot to note that those crates are removed because otherwise their code would have to be refactored as well. By removing dead code, I tried to reduce complexity of the refactor.

> 
> The other things are unfortunately interconnected and cannot be easily serialized. However, this is really just a refactor. There is no deep review needed IMO. Understanding this PR does not contribute to understanding the new compiler.

third_party/move/move-model/tests/testsuite.rs
wrwg: This is dead code (only used in tests), and it would have been rather difficult to keep it working.

third_party/move/move-model/src/lib.rs
wrwg: > 

Same as above. Dead code.

gerben-stavenga:
I see you did refactor move-to-yul is that still a package we need to maintain

aptos-move/framework/aptos-stdlib/sources/copyable_any.spec.move
```diff
@@ -4,7 +4,7 @@ spec aptos_std::copyable_any {
     // Function specifications
     // -----------------------
 
-    spec pack<T: drop + store>(x: T): Any {
+    spec pack<T: drop + store + copy>(x: T): Any {
```
gerben-stavenga, line 7: Why did this change happen in this PR?

third_party/move/move-model/src/ast.rs
```diff
@@ -1188,30 +1225,27 @@ impl ModuleName {
     /// Determine whether this is a script. The move-compiler infrastructure uses MAX_ADDR
     /// for pseudo modules created from scripts, so use this address to check.
     pub fn is_script(&self) -> bool {
-        static MAX_ADDR: Lazy<BigUint> = Lazy::new(|| {
-            BigUint::from_str_radix("ffffffffffffffffffffffffffffffff", 16).expect("valid hex")
-        });
-        self.0 == *MAX_ADDR
+        self.0 == Address::Numerical(AccountAddress::new([0xFF; AccountAddress::LENGTH]))
```
gerben-stavenga, line 1228: Perhaps make MAX_ADDR  part of AccountAddress

gerben-stavenga: Also I thought address was 32 bytes, I count 32 f's in the previous version which is 16 bytes. Was that a bug?

gerben-stavenga (approved):

aptos-move/framework/aptos-stdlib/doc/math128.md
```diff
@@ -10,12 +10,14 @@ Standard math utilities missing in the Move Language.
 -  [Function `max`](#0x1_math128_max)
 -  [Function `min`](#0x1_math128_min)
 -  [Function `average`](#0x1_math128_average)
+-  [Function `mul_div`](#0x1_math128_mul_div)
```
brmataptos, line 13: These doc changes are unrelated to the refactoring, right?

aptos-move/framework/aptos-stdlib/sources/copyable_any.spec.move
wrwg: This is because abilities were wrongly implemented before. You could list here whatever abilities you want before instead of those the original fun decl had. The refactor fixed this (kind of unintentional) and so I needed to also correct the source here.

third_party/move/move-model/src/ast.rs
wrwg: Yes, it was in fact a bug, someone reported this in the Move repo. It only effected docgen for scripts however, which I think no one uses.

aptos-move/framework/aptos-stdlib/doc/math128.md
wrwg: No actually not. All inline functions have become visible in the model, because the models function tables are now derived from the parsed AST, and not the bytecode. In the bytecode, inlined functions are gone, so there were not visible in the model before (and hence docgen).

wrwg: So basically -- and unintentionally -- this PRC fixes a known incompleteness of docgen. It similar fixes some link decoration because the parsing of addresses had bugs which are fixed with this PR.

All those changes are _not_ piggybacked on this PR but a consequence of it.

aptos-move/framework/aptos-stdlib/sources/copyable_any.spec.move
gerben-stavenga: I suspected as much, could you add it to the description which bugs were fixed?

third_party/move/move-model/src/ast.rs
gerben-stavenga: Thx, please at to description as fixed as well.

gerben-stavenga: Maybe close https://github.com/aptos-labs/aptos-core/issues/7731

third_party/move/move-prover/bytecode/src/lib.rs
```diff
@@ -50,6 +46,9 @@ pub mod verification_analysis;
 pub mod verification_analysis_v2;
 pub mod well_formed_instrumentation;
 
+/// An error message used for cases where a compiled module is expected to be attached
+pub(crate) const COMPILED_MODULE_AVAILABLE: &str = "compiled module missing";
```
meng-xu-cs, line 50: Just a comment that this error message marks a significant portion of the refactoring and can be used to trace where a compiled module is expected in the code.

The name and the static string seems a bit out-of-sync at first sight, but when seeing how it is used, i.e., in `.expect(COMPILED_MODULE_AVAILABLE)`, the naming makes perfect sense.

meng-xu-cs (approved):
Thanks a lot for this huge effort! I understand this is a large PR but I don't see a way of making it in smaller pieces. The compiled module expectation is a general assumption in the old code and removing this assumption will affect a lot of things.

Fortunately, this is mostly a refactoring to make the Rust type system happy (mainly `Option<T>` to `T` or vice versa) and does not intend to have an impact on the functionality. Giving a green light to the PR based on this.

PR #8153, opened 2023-05-11
Limit num ristretto points

zjma:


types/src/on_chain_config/timed_features.rs
```diff
@@ -92,6 +93,8 @@ impl TimedFeatureFlag {
             (EntryTypeSizeLimit, TESTNET) => 1680231600000, /* Thu Mar 30 2023 20:00:00 GMT-0700 */
             (EntryTypeSizeLimit, MAINNET) => 1680318000000, /* Fri Mar 31 2023 20:00:00 GMT-0700 */
 
+            (LimitedNumOfRistretto255PointsPerSession, TESTNET) => 1683306000000, /* Fri May 5 2023 10:00:00 GMT-0700 */
+            (LimitedNumOfRistretto255PointsPerSession, MAINNET) => 1683824400000, /* Thu May 11 2023 10:00:00 GMT-0700 */
```
sherry-x, line 97: you can probably remove those since the time already passed

alinush (changes requested):
As @sherry-x suggested, let's remove the timed feature. This means removing the calls to `add_point` and marking the function as private, instead of `pub`. Then, we'll know for sure our code only uses `safe_add_point`.

alinush (changes requested):
As @sherry-x suggested, let's remove the timed feature. This means removing the calls to `add_point` and marking the function as private, instead of `pub`. Then, we'll know for sure our code only uses `safe_add_point`.

alinush (changes requested):
As @sherry-x suggested, let's remove the timed feature. This means removing the calls to `add_point` and marking the function as private, instead of `pub`. Then, we'll know for sure our code only uses `safe_add_point`.

sherry-x (approved):

alinush (approved):

PR #8154, opened 2023-05-11
[transaction generator] Allow using move-example modules directly (ie. ambassador)

igor-aptos:


aptos-move/move-examples/token_objects/ambassador/move/Move.toml
```diff
@@ -3,7 +3,7 @@ name = 'ambassador_token'
 version = '1.0.0'
 
 [addresses]
-token_objects = "_"
+token_objects = "0xABCD"
```
igor-aptos, line 6: in order for me to compile this, I cannot have placeholder here, but an actual address (that gets replaced in the compiled objects later with correct address)

let me know if you have any concerns here

aptos-move/move-examples/token_objects/ambassador/move/sources/ambassador.move
```diff
@@ -105,6 +105,16 @@ module token_objects::ambassador {
         );
     }
 
+    public entry fun mint_ambassador_token_by_user(
```
igor-aptos, line 108: @junkil-park @movekevin I need to add this entry function, and use in the transaction generator, as otherwise throughput is miniscule (as we can have only 100 transactions per signer pending at any time)

movekevin (approved):

aptos-move/move-examples/token_objects/ambassador/move/Move.toml
junkil-park: I think this should be fine.

aptos-move/move-examples/token_objects/ambassador/move/sources/ambassador.move
junkil-park: OK.

junkil-park (approved):
LGTM. @igor-aptos , are you benchmarking only the minting operation? I have a PR (https://github.com/aptos-labs/aptos-core/pull/8200/files#diff-f94c1473d02bd66860d9533e327813253c183928f35d5cebf1e7136ce51162a7) that adds more behaviors to `set_ambassador_level`, thus resulting in making the function a bit slower.

testsuite/module-publish/src/main.rs
```diff
@@ -117,12 +139,10 @@ use once_cell::sync::Lazy;"#,
 }
 
 // Write out given package
-fn write_pacakge(file: &mut File, package_folder: DirEntry) -> String {
-    let file_name = package_folder.file_name();
-    let package_name = file_name.to_str().unwrap();
+fn write_pacakge(file: &mut File, package_path: PathBuf, package_name: &str) -> String {
```
grao1991, line 142: nit: s/pacakge/package

grao1991 (approved):

igor-aptos:
> LGTM. @igor-aptos , are you benchmarking only the minting operation? I have a PR (https://github.com/aptos-labs/aptos-core/pull/8200/files#diff-f94c1473d02bd66860d9533e327813253c183928f35d5cebf1e7136ce51162a7) that adds more behaviors to `set_ambassador_level`, thus resulting in making the function a bit slower.

I am not yet using set_ambassador_level or other functions, though we can add benchmarks for those too

PR #8158, opened 2023-05-11
remove dead rocksdb and github secure storage implementation to reduce bloat in aptos-config package

geekflyer:
This PR removes the dead GitHub and RocksDb backends from the `secure/storage` crate and moves the `aptos-github-client` crate to a new location in the project structure. Main motivation is to reduce bloat/size of the commonly used aptos-config package such that it doesn't depend on rocksdb. This will likely reduce our binary and images sizes and reduce build times of some targets.


msmouse (approved):

davidiw (approved):
I guess if there's really no use....

PR #8160, opened 2023-05-11
[tf/gcp] add dns support  for GCP testnet

rustielin:
### Description

Add variable `var.create_dns_records` to control whether or not to create validator and fullnode specific DNS records directly on the first validator. These are only really useful in the case of deploying just a single validator/VFN, since there's 2 records that need to be created (one for validator and one for fullnode). It's simpler to create the record directly in TF rather than using an external dependency like external-dns.

Since we're hosting a bunch of validators and fullnodes for testnets, we'd want each one to have a DNS name. That's a use case for external-dns. So in testnets, set `create_dns_records = false`, and let external-dns handle the rest

### Test Plan

Lint & apply, unblocking new devnet GCP deployment

### DIFF
--- a/terraform/aptos-node-testnet/gcp/main.tf
+++ b/terraform/aptos-node-testnet/gcp/main.tf
@@ -29,6 +29,9 @@ module "validator" {
   zone_name    = var.zone_name # keep empty if you don't want a DNS name
   zone_project = var.zone_project
   record_name  = var.record_name
+  # do not create the main fullnode and validator DNS records
+  # instead, rely on external-dns from the testnet-addons
+  create_dns_records = false
 
   # General chain config
   era            = var.era
--- a/terraform/aptos-node/gcp/dns.tf
+++ b/terraform/aptos-node/gcp/dns.tf
@@ -14,17 +14,19 @@ locals {
 }
 
 data "kubernetes_service" "validator-lb" {
-  count = var.zone_name != "" ? 1 : 0
+  count = var.zone_name != "" && var.create_dns_records ? 1 : 0
   metadata {
-    name = "${local.workspace_name}-aptos-node-validator-lb"
+    # This is the main validator LB service that is created by the aptos-node helm chart
+    name = "${local.workspace_name}-aptos-node-0-validator-lb"
   }
   depends_on = [time_sleep.lb_creation]
 }
 
 data "kubernetes_service" "fullnode-lb" {
-  count = var.zone_name != "" ? 1 : 0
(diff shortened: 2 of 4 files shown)


sionescu (approved):

sionescu (approved):

terraform/aptos-node/gcp/dns.tf
```diff
@@ -14,17 +14,17 @@ locals {
 }
 
 data "kubernetes_service" "validator-lb" {
-  count = var.zone_name != "" ? 1 : 0
+  count = var.zone_name != "" && var.create_records ? 1 : 0
   metadata {
-    name = "${local.workspace_name}-aptos-node-validator-lb"
+    name = "${local.workspace_name}-aptos-node-0-validator-lb"
```
perryjrandall, line 19: Why are we hardcoding these to 0?

Maybe worth a comment

terraform/aptos-node/gcp/dns.tf
```diff
@@ -56,9 +56,9 @@ resource "google_dns_record_set" "fullnode" {
 }
 
 output "validator_endpoint" {
-  value = var.zone_name != "" ? "/dns4/${trimsuffix(google_dns_record_set.validator[0].name, ".")}/tcp/${data.kubernetes_service.validator-lb[0].spec[0].port[0].port}" : null
+  value = var.zone_name != "" && var.create_records ? "/dns4/${trimsuffix(google_dns_record_set.validator[0].name, ".")}/tcp/${data.kubernetes_service.validator-lb[0].spec[0].port[0].port}" : null
```
perryjrandall, line 59: This access path seems error prone if the service isnt present?

perryjrandall (approved):

terraform/aptos-node-testnet/gcp/main.tf
```diff
@@ -29,6 +29,9 @@ module "validator" {
   zone_name    = var.zone_name # keep empty if you don't want a DNS name
   zone_project = var.zone_project
   record_name  = var.record_name
+  # do not create the main fullnode and validator DNS records
+  # instead, rely on external-dns from the testnet-addons
+  create_records = false
```
geekflyer, line 34: let's call this `create_dns_records` for clarity

terraform/aptos-node/gcp/dns.tf
rustielin: yea we're kinda making the assumption that the k8s services come up without issue. no way around this imo since we're still coupling infra and application release. At least it will fail with a good error

PR #8170, opened 2023-05-12
[aptos-node] add --config-override

rustielin:
# Overview

Add `--config-override` flag to `aptos-node`, so we can override the default nodeconfig, which is especially useful when used with `--test`, since the single node testnet mode will auto-populate a lot of fields. This PR uses `serde_merge` to handle the merging and deserialization for us.

Run a local testnet and merge it with a config that enables indexer GRPC:

```
storage:
  enable_indexer: true

indexer_grpc:
  enabled: true
  address: 0.0.0.0:50051
  processor_task_count: 10
  processor_batch_size: 100
  output_batch_size: 100

```

Then start the node:

```
cargo run -p aptos-node -- --test --config-override test_config.yaml
```

The output NodeConfig and also the logs show that the GRPC endpoint is now working!

```
WARNING: Entering test mode! This should never be used in production!
        **Merged default config with override from path: "test_config.yaml"
        Log file: "/private/var/folders/0r/0cqndyk91kb6s_xv23gdvsdw0000gn/T/bd8fd545c16b34856518744d3ab8aa08/validator.log"
        Test dir: "/private/var/folders/0r/0cqndyk91kb6s_xv23gdvsdw0000gn/T/bd8fd545c16b34856518744d3ab8aa08"
        Aptos root key path: "/private/var/folders/0r/0cqndyk91kb6s_xv23gdvsdw0000gn/T/bd8fd545c16b34856518744d3ab8aa08/mint.key"
        Waypoint: 0:4b7a0c3d90ee83433ddabdf033b35a6c742b010a9b0aa404a0e83397c60b756c
        ChainId: testing
        Indexer gRPC endpoint: 0.0.0.0:50053

```

# Test plan

New unit test, and existing CI

### DIFF
--- a/aptos-node/src/lib.rs
+++ b/aptos-node/src/lib.rs
@@ -29,7 +29,7 @@ use hex::FromHex;
 use rand::{rngs::StdRng, SeedableRng};
 use std::{
     fs,
-    io::Write,
+    io::{Read, Write},
     path::PathBuf,
     sync::{
         atomic::{AtomicBool, Ordering},
@@ -60,6 +60,10 @@ pub struct AptosNodeArgs {
     /// Repeated runs will start up from previous state.
     #[clap(long, parse(from_os_str), requires("test"))]
     test_dir: Option<PathBuf>,
+
+    /// Path to node configuration file override for local test mode. Cannot be used with --config
+    #[clap(long, parse(from_os_str), requires("test"), conflicts_with("config"))]
+    test_config_override: Option<PathBuf>,
 
     /// Run only a single validator node testnet.
     #[clap(long)]
@@ -119,6 +123,7 @@ impl AptosNodeArgs {
                 .unwrap_or_else(StdRng::from_entropy);
             setup_test_environment_and_start_node(
                 self.config,
+                self.test_config_override,
                 self.test_dir,
                 self.random_ports,
                 self.lazy,
@@ -216,6 +221,7 @@ pub fn start(
 /// Creates a simple test environment and starts the node
 pub fn setup_test_environment_and_start_node<R>(
     config_path: Option<PathBuf>,
+    test_config_override_path: Option<PathBuf>,
     test_dir: Option<PathBuf>,
     random_ports: bool,
     enable_lazy_mode: bool,
@@ -243,7 +249,11 @@ where
             .map_err(|error| anyhow!("Unable to load config: {:?}", error))?
     } else {
         // Create a test only config for a single validator node
-        let node_config = create_single_node_test_config(config_path, enable_lazy_mode);
+        let node_config = create_single_node_test_config(
+            config_path.clone(),
+            test_config_override_path.clone(),
+            enable_lazy_mode,
+        )?;
 
         // Build genesis and the validator node
         let builder = aptos_genesis::builder::Builder::new(&test_dir, framework.clone())?
@@ -279,6 +289,18 @@ where
 
     // Print out useful information about the environment and the node
     println!("Completed generating configuration:");
+    if test_config_override_path.is_some() {
+        println!(
+            "\tMerged default config with override from path: {:?}",
+            test_config_override_path.unwrap()
+        );
+    }
+    if config_path.is_some() {
+        println!(
+            "\tUsed user-provided config from path: {:?}",
+            config_path.unwrap()
+        );
+    }
     println!("\tLog file: {:?}", log_file);
     println!("\tTest dir: {:?}", test_dir);
     println!("\tAptos root key path: {:?}", aptos_root_key_path);
@@ -293,6 +315,11 @@ where
         "\tAptosnet fullnode network endpoint: {}",
         &config.full_node_networks[0].listen_address
     );
+    if config.indexer_grpc.enabled {
+        if let Some(ref indexer_grpc_address) = config.indexer_grpc.address {
+            println!("\tIndexer gRPC endpoint: {}", indexer_grpc_address);
+        }
+    }
     if enable_lazy_mode {
         println!("\tLazy mode is enabled");
     }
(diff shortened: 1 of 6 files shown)


aptos-node/src/lib.rs
```diff
@@ -301,14 +328,74 @@ where
     start(config, Some(log_file), false)
 }
 
+/// Merge two YAML values, useful for merging NodeConfigs
+fn merge_yaml(a: &mut serde_yaml::Value, b: serde_yaml::Value) {
```
perryjrandall, line 332: You can do this or you can add serde_merge

https://docs.rs/serde_merge/latest/serde_merge/

aptos-node/src/lib.rs
```diff
@@ -301,14 +328,74 @@ where
...
+                    continue;
+                }
+                if !a.contains_key(&k) {
+                    a.insert(k.to_owned(), v.to_owned());
+                } else {
+                    merge_yaml(&mut a[&k], v);
+                }
+            }
+        },
+        (a, b) => *a = b,
```
perryjrandall, line 350: does this behavior make sense?

Or should we error on merge of things which are not mapping?

Also you can merge sequence, although not sure if you want to do that at a high level

aptos-node/src/lib.rs
```diff
@@ -301,14 +328,74 @@ where
...
+                    anyhow::anyhow!(
+                        "Unable to open config override file {}. Error: {}",
+                        config_override_path.to_str().unwrap(),
+                        e
+                    )
+                })
+                .unwrap();
+            let mut contents = String::new();
+            file.read_to_string(&mut contents)
+                .map_err(|e| {
```
perryjrandall, line 375: Can we dedupe these error blocks?

perryjrandall:
Please add some unittests, consider using a merge alternative

aptos-node/src/lib.rs
rustielin: ah yea that sounds like a better idea

aptos-node/src/lib.rs
```diff
@@ -301,14 +328,60 @@ where
...
 /// Creates a single node test config, with a few config tweaks to reduce
 /// the overhead of running the node on a local machine.
 fn create_single_node_test_config(
     config_path: Option<PathBuf>,
+    config_override_path: Option<PathBuf>,
     enable_lazy_mode: bool,
 ) -> NodeConfig {
-    // Build a single validator network with a generated config
-    let mut node_config = NodeConfig::get_default_validator_config();
+    let mut node_config = match config_override_path {
```
perryjrandall, line 348: If you're doing a lot of `if Ok or if Some` then you might want to consider chaining with map instead of deeply nesting, sometimes it can be more readable

aptos-node/src/lib.rs
```diff
@@ -301,14 +328,60 @@ where
...
+                        Err(e) => {
+                            panic!(
+                                "Unable to read contents of config override file {}. Error: {}",
+                                config_override_path.to_str().unwrap(),
+                                e
+                            )
+                        },
+                    }
+                },
+                Err(e) => {
```
perryjrandall, line 373: Is panic'ing a good idea here or should we propagate the error upwards?

aptos-node/src/tests.rs
```diff
@@ -51,3 +51,39 @@ fn test_mutual_authentication_validators() {
 fn test_aptos_vm_does_not_have_test_natives() {
     aptos_vm::natives::assert_no_test_natives(crate::utils::ERROR_MSG_BAD_FEATURE_FLAGS)
 }
+
+#[test]
+fn test_config_override_merge() {
```
perryjrandall, line 56: noice!

Although it would be very helpful to add a test for `create_single_node_test_config` since that actually has more logic than the merge itself

Ideally the merge logic is covered by their own unittests so this is really just testing their library which is less useful of a test

perryjrandall (approved):
So clean

aptos-node/src/lib.rs
rustielin: ah true yea. slightly more readable now with `map_err`

aptos-node/src/lib.rs
rustielin: good point. I've changed it to return `anyhow::Result` so we can propagate the errors upwards

aptos-node/src/lib.rs
```diff
@@ -372,12 +432,24 @@ fn create_single_node_test_config(
...
+            Err(error) => {
+                bail!(
+                    "Unable to load config from path: {:?}. Error: {:?}",
+                    config_path,
+                    error
+                );
+            },
+        }
+
+        // .unwrap_or_else(|error| {
```
perryjrandall, line 446: remove or uncomment this?

rustielin: agh good catch

aptos-node/src/lib.rs
```diff
@@ -61,6 +61,10 @@ pub struct AptosNodeArgs {
     #[clap(long, parse(from_os_str), requires("test"))]
     test_dir: Option<PathBuf>,
 
+    /// Path to node configuration file override for local test mode. Cannot be used with --config
+    #[clap(long, parse(from_os_str), requires("test"), conflicts_with("config"))]
+    config_override: Option<PathBuf>,
```
JoshLind, line 66: Can we name this `--test-config-override`? AFAICT, this is only relevant for the test mode, right? Because the normal mode will do this by default (i.e., you specify the config options in the file and anything you don't specify is chosen by the default or config optimizer). Does that make sense?

aptos-node/src/lib.rs
```diff
@@ -61,6 +61,10 @@ pub struct AptosNodeArgs {
...
+
```
gregnazario, line 67: I don't think you necessarily need this, it'll override all fields with default fields if you don't them already

aptos-node/src/lib.rs
```diff
@@ -293,6 +315,11 @@ where
         "\tAptosnet fullnode network endpoint: {}",
         &config.full_node_networks[0].listen_address
     );
+    if config.indexer_grpc.enabled {
+        if let Some(ref indexer_grpc_address) = config.indexer_grpc.address {
+            println!("\tIndexer gRPC endpoint: {}", indexer_grpc_address);
+        }
+    }
```
gregnazario, line 322: This is a good change :)

aptos-node/src/lib.rs
```diff
@@ -301,14 +328,47 @@ where
     start(config, Some(log_file), false)
 }
 
+/// Merges node_config with the config override file
+fn merge_nodeconfig_override(
```
JoshLind, line 332: nit: `nodeconfig` -> `node_config`, two words? 😄

aptos-node/src/lib.rs
```diff
@@ -372,12 +432,16 @@ fn create_single_node_test_config(
 
     // If a config path was provided, use that as the template
     if let Some(config_path) = config_path {
-        node_config = NodeConfig::load_config(&config_path).unwrap_or_else(|error| {
-            panic!(
-                "Failed to load persisted config at path: {:?}. Error: {:?}",
-                config_path, error
-            )
-        });
+        node_config = match NodeConfig::load_config(&config_path) {
```
JoshLind, line 435: nit: use `map_error` and `?` instead of this? e.g.,:

```
node _config = match XXX.map_err(|error| Error::...)?
```

crates/aptos/src/node/mod.rs
```diff
@@ -1055,6 +1055,13 @@ pub struct RunLocalTestnet {
     #[clap(long, parse(from_os_str))]
     test_dir: Option<PathBuf>,
 
+    /// Path to node configuration file override for local test mode.
+    ///
+    /// If provided, the default node config will be overridden by the config in the given file.
+    /// Cannot be used with --config
+    #[clap(long, parse(from_os_str), requires("test"), conflicts_with("config"))]
+    config_override: Option<PathBuf>,
```
JoshLind, line 1063: Likewise, maybe `test_config_override`?

aptos-node/src/tests.rs
```diff
@@ -51,3 +51,57 @@ fn test_mutual_authentication_validators() {
...
+#[test]
+fn test_create_single_node_test_config() {
+    // create a test config override and merge it with the default config
+    // this will get cleaned up by the tempdir when it goes out of scope
+    let test_dir = aptos_temppath::TempPath::new().as_ref().to_path_buf();
+    fs::DirBuilder::new()
+        .recursive(true)
+        .create(&test_dir)
+        .expect("Failed to create test_dir");
+    let config_path = None;
```
JoshLind, line 64: Why do you need a variable for this? 😄 Just pass `None` into `create_single_node_test_config`

aptos-node/src/lib.rs
rustielin: yea makes sense!

aptos-node/src/tests.rs
rustielin: 🤦 thanks for catching this lol

JoshLind (approved):
LGTM! 😄 Thanks @rustielin.

PR #8177, opened 2023-05-12
[Forge] Adjust some of the thresholds that are flaky

bchocho:
### Description

Need to dig deeper on why the fluctuation in TPS is so large, but for now reducing the thresholds to reduce noise. If these fail frequently, they can mask other issues.

### DIFF
--- a/testsuite/forge-cli/src/main.rs
+++ b/testsuite/forge-cli/src/main.rs
@@ -865,7 +865,7 @@ fn three_region_sim_graceful_overload(config: ForgeConfig) -> ForgeConfig {
                 // Additionally - we are not really gracefully handling overlaods,
                 // setting limits based on current reality, to make sure they
                 // don't regress, but something to investigate
-                avg_tps: 1400,
+                avg_tps: 1200,
                 latency_thresholds: &[],
             },
         }])
@@ -952,7 +952,7 @@ fn individual_workload_tests(test_name: String, config: ForgeConfig) -> ForgeCon
         )
         .with_success_criteria(
             SuccessCriteria::new(match test_name.as_str() {
-                "account_creation" => 3700,
+                "account_creation" => 3600,
                 "nft_mint" => 1000,
                 "publishing" => 60,
                 "write_new_resource" => 3700,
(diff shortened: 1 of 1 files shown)


testsuite/forge-cli/src/main.rs
```diff
@@ -1003,7 +1003,7 @@ fn single_vfn_perf(config: ForgeConfig) -> ForgeConfig {
         .with_initial_fullnode_count(1)
         .with_network_tests(vec![&PerformanceBenchmark])
         .with_success_criteria(
-            SuccessCriteria::new(5000)
+            SuccessCriteria::new(4250)
```
sherry-x, line 1006: slightly worried that we might not be able to come back from lowering this, it’s easy to lower threshold than recovering regression. This is quite a big change from 5k to 4.2k. have we be able to figure out when did we start to see this regression?

ibalajiarun (approved):

testsuite/forge-cli/src/main.rs
bchocho: The regression comes after enabling QS as the default. It's known that single node performance behaves different with QS (because of backpressure) and we haven't quite tuned this.

Here's a chart: [internal link](https://cloud.us.humio.com/github/search?connect-points=false&imputation=none&live=false&query=%23type%3Dgithub-job-log%20%7C%20github.job.conclusion%20!%3D%20cancelled%20github.job.conclusion%20!%3D%20skipped%20github.job.conclusion%20!%3D%20null%20github.workflow.head_branch%3Dmain%20github.job.name%3D/run-forge-.*forge$/%0A%7C%20github.job.name%3D%22run-forge-single-vfn-perf%20/%20forge%22%0A%7C%20message%20%3D%20%22*Test%20stats*:%20submitted:*%22%0A%7C%20regex(%22committed:%20(?%3Ctps%3E%5B0-9%5D+)%20txn/s%22)%0A%7C%20timechart(span%3D1day,%20series%3Dgithub.job.name,%20limit%3D30,%20function%3Dmax(field%3Dtps))&start=30d&tz=Asia/Seoul&widgetType=time-chart)

sitalkedia (approved):

PR #8178, opened 2023-05-13
[Sharding][Execution] Integrate Sharded block executor in the VM execution path

sitalkedia:
### Description

Integrate the shared block executor to the VM execution path. Please note that this is a No-Op as the newly introduced sharding functions are only used in the executor benchmark. 

### Test Plan
Existing UTs

### DIFF
--- a/aptos-move/aptos-vm/src/sharded_block_executor/executor_shard.rs
+++ b/aptos-move/aptos-vm/src/sharded_block_executor/executor_shard.rs
@@ -61,6 +61,7 @@ impl<S: StateView + Sync + Send + 'static> ExecutorShard<S> {
                         state_view.as_ref(),
                         concurrency_level_per_shard,
                     );
+                    drop(state_view);
                     self.result_tx.send(ret).unwrap();
                 },
                 ExecutorShardCommand::Stop => {
--- a/aptos-move/aptos-vm/src/sharded_block_executor/executor_shard.rs
+++ b/aptos-move/aptos-vm/src/sharded_block_executor/executor_shard.rs
@@ -2,7 +2,6 @@
 // Parts of the project are originally copyright © Meta Platforms, Inc.
 // SPDX-License-Identifier: Apache-2.0
 use crate::{block_executor::BlockAptosVM, sharded_block_executor::ExecutorShardCommand};
-use aptos_block_executor::errors::Error;
 use aptos_logger::trace;
 use aptos_state_view::StateView;
 use aptos_types::transaction::TransactionOutput;
@@ -14,22 +13,20 @@ use std::sync::{
 
 /// A remote block executor that receives transactions from a channel and executes them in parallel.
 /// Currently it runs in the local machine and it will be further extended to run in a remote machine.
-pub struct ExecutorShard<S: StateView + Sync> {
+pub struct ExecutorShard<S: StateView + Sync + Send + 'static> {
     shard_id: usize,
     executor_thread_pool: Arc<rayon::ThreadPool>,
     num_executor_threads: usize,
-    command_rx: Receiver<ExecutorShardCommand>,
-    result_tx: Sender<Result<Vec<TransactionOutput>, Error<VMStatus>>>,
-    state_view: Arc<S>,
+    command_rx: Receiver<ExecutorShardCommand<S>>,
+    result_tx: Sender<Result<Vec<TransactionOutput>, VMStatus>>,
 }
 
-impl<S: StateView + Sync> ExecutorShard<S> {
+impl<S: StateView + Sync + Send + 'static> ExecutorShard<S> {
     pub fn new(
         shard_id: usize,
         concurrency_level: usize,
-        state_view: Arc<S>,
-        command_rx: Receiver<ExecutorShardCommand>,
-        result_tx: Sender<Result<Vec<TransactionOutput>, Error<VMStatus>>>,
+        command_rx: Receiver<ExecutorShardCommand<S>>,
+        result_tx: Sender<Result<Vec<TransactionOutput>, VMStatus>>,
     ) -> Self {
         let executor_thread_pool = Arc::new(
             rayon::ThreadPoolBuilder::new()
@@ -43,7 +40,6 @@ impl<S: StateView + Sync> ExecutorShard<S> {
             num_executor_threads: concurrency_level,
             command_rx,
             result_tx,
-            state_view,
         }
     }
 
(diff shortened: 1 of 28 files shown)


aptos-move/aptos-vm/src/sharded_block_executor/executor_shard.rs
```diff
@@ -14,22 +13,20 @@ use std::sync::{
...
     num_executor_threads: usize,
-    command_rx: Receiver<ExecutorShardCommand>,
-    result_tx: Sender<Result<Vec<TransactionOutput>, Error<VMStatus>>>,
-    state_view: Arc<S>,
+    command_rx: Receiver<ExecutorShardCommand<S>>,
+    result_tx: Sender<Result<Vec<TransactionOutput>, VMStatus>>,
 }
 
-impl<S: StateView + Sync> ExecutorShard<S> {
+impl<S: StateView + Sync + Send + 'static> ExecutorShard<S> {
```
grao1991, line 24: The static looks weird here.

storage/storage-interface/src/cached_state_view.rs
```diff
@@ -89,6 +91,12 @@ pub struct CachedStateView {
     proof_fetcher: Arc<AsyncProofFetcher>,
 }
 
+impl Debug for CachedStateView {
```
grao1991, line 94: Do we really need this?

execution/executor/src/components/chunk_output.rs
```diff
@@ -116,9 +132,13 @@ impl ChunkOutput {
     #[cfg(not(feature = "consensus-only-perf-test"))]
     fn execute_block<V: VMExecutor>(
         transactions: Vec<Transaction>,
-        state_view: &CachedStateView,
+        state_view: Arc<CachedStateView>,
```
grao1991, line 135: Why do we need Arc?

storage/storage-interface/src/cached_state_view.rs
sitalkedia: Yes, unfortunately this is needed for Arc::try_unwrap function weirdly.

aptos-move/aptos-vm/src/sharded_block_executor/executor_shard.rs
sitalkedia: Rust complains otherwise :).

execution/executor/src/components/chunk_output.rs
sitalkedia: So, this is because the ExecutorShard runs as a separate thread and there is no way to pass a reference to that thread without lifetime issues.

storage/storage-interface/src/cached_state_view.rs
```diff
@@ -89,6 +91,12 @@ pub struct CachedStateView {
...
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        write!(f, "{:?}::{:?}", self.id, &self.state_cache)
```
msmouse, line 96: I hope DashMap::debug doesn't print out all the KVs..

storage/storage-interface/src/lib.rs
```diff
@@ -2,6 +2,8 @@
 // Parts of the project are originally copyright © Meta Platforms, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
+extern crate core;
```
msmouse, line 5: what's this for?

msmouse (commented):
I'm not sure we are ready to put this onto the critical path yet:
1. It doesn't work for production traffic (when it's actually sharded)
2. I'm yet clear if we are gonna base our p-txn executor on top of BlockSTM.. maybe (I tend to think it's likely) a NaiveExecutor is more proper.

execution/executor/src/components/chunk_output.rs
sitalkedia: Yes, we were seeing around 30% performance degradation with that implementation. I still need to dig into the root cause of that but I think this is clean as well?

sitalkedia:
> I'm not sure we are ready to put this onto the critical path yet:

@msmouse - To be clear, sharded block-stm is ofcourse not in the critical path. All this does is by default set the number of shards to 1 and execute sharded block-stm, the outcome is exactly same as non-sharded one. Its just makes it easier to integrate with executor benchmark and play around with Forge with sharded block-stm.

msmouse:
I'm sorta okay with this, because it doesn't look like any real world performance impact or too much complexity.. But still, let's spend a little time to see if it's easy to make the sharded executor "on the side" and keep production flow untouched as much as possible. -- if it's too much work, maybe we land it as is. Thanks man!

sitalkedia:
@msmouse - Updated to call the sharded execution only for the executor benchmark.

storage/storage-interface/src/cached_state_view.rs
sitalkedia: You are right, removed it.

storage/storage-interface/src/lib.rs
sitalkedia: accidental change, removed

msmouse (approved):

aptos-move/aptos-vm/src/sharded_block_executor/executor_shard.rs
igor-aptos: what does it complain on, do you have a message? it might be something else that needs to be changed

igor-aptos (approved):

aptos-move/aptos-vm/src/sharded_block_executor/executor_shard.rs
sitalkedia: It complains about the size of the type S might not be available in compile time - which is a bit weird but adding 'static here helps

execution/executor/src/components/chunk_output.rs
```diff
@@ -37,16 +47,22 @@ impl ChunkOutput {
...
 
         update_counters_for_processed_chunk(&transactions, &transaction_outputs, "executed");
 
+        let state_view = Arc::try_unwrap(state_view_arc).unwrap();
+
         Ok(Self {
             transactions,
             transaction_outputs,
+            // Unwrapping here is safe because the execution has finished and it is guaranteed that
+            // the state view is not used anymore.
```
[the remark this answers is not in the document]
sitalkedia: Sorry missed this comment, will fix it in next PR.

PR #8183, opened 2023-05-13
Add zero() to fungible_asset module

movekevin:
### Description
It's very common in Defi to have the follow pattern:
```
let zero = fungible_asset::zero(metadata);
while (some conditions) {
  // Loop through a series of positions and calculate interest
  let interest = fungible_asset::extract(&mut position, interest_amount);

  // Take interest
  fungible_asset::merge(&mut zero, interest);
}
```

It's very hard to do this without zero(). We can add the fungible assets to a vector and then merge them at the end but this is just unnecessarily extra work that costs extra gas (once to insert into a vector and another one when passing through to merge).

### DIFF
--- a/aptos-move/framework/aptos-framework/sources/fungible_asset.move
+++ b/aptos-move/framework/aptos-framework/sources/fungible_asset.move
@@ -321,9 +321,8 @@ module aptos_framework::fungible_asset {
         metadata: Object<T>,
     ): Object<FungibleStore> {
         let store_obj = &object::generate_signer(constructor_ref);
-        let metadata = object::convert<T, Metadata>(metadata);
         move_to(store_obj, FungibleStore {
-            metadata,
+            metadata: object::convert(metadata),
             balance: 0,
             frozen: false,
         });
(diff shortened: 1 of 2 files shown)


aptos-move/framework/aptos-framework/sources/fungible_asset.move
```diff
@@ -466,6 +466,16 @@ module aptos_framework::fungible_asset {
         deposit_with_ref(transfer_ref, to, fa);
     }
 
+    /// Create a fungible asset with zero amount.
+    /// This can be useful when starting a series of computations where the initial value is 0.
+    public fun zero<T: key>(metadata: Object<T>): FungibleAsset {
+        let metadata = object::convert<T, Metadata>(metadata);
```
davidiw, line 472:
```suggestion
        let metadata = object::convert(metadata);
```
that shouldn't be necessary.

aptos-move/framework/aptos-framework/sources/fungible_asset.move
```diff
@@ -466,6 +466,16 @@ module aptos_framework::fungible_asset {
...
+        FungibleAsset {
+            metadata,
```
davidiw, line 474: then you could just call object::convert here...

davidiw (approved):

lightmark (approved):

aptos-move/framework/aptos-framework/sources/fungible_asset.move
movekevin: Updated

aptos-move/framework/aptos-framework/sources/fungible_asset.move
movekevin: Updated

PR #8186, opened 2023-05-14
[Gas Estimation] Implement block-based gas estimation

bchocho:
### Description

Implementation of [AIP-34](https://github.com/aptos-foundation/AIPs/blob/main/aips/aip-34.md)

### Test Plan

Added api tests and smoke test.


api/src/context.rs
```diff
@@ -797,27 +806,168 @@ impl Context {
...
+            }
+        }
+        None
+    }
+
+    pub fn estimate_gas_price<E: InternalError>(
+        &self,
+        ledger_info: &LedgerInfo,
+    ) -> Result<GasEstimation, E> {
+        let max_block_history = 120;
```
JoshLind, line 849: nit: do we want this to be configurable, too?

api/goldens/aptos_api__tests__transactions_test__test_gas_estimation_one_block.json
```diff
@@ -0,0 +1,5 @@
+{
+  "deprioritized_gas_estimate": 1,
```
JoshLind, line 2: FWIW, these labels aren't as descriptive to me as those used in the AIP. Should we consider changing them? For example: `low gas fee`, `market gas fee`, `aggressive (or priority) gas fee` seem more explanatory? 😄

Also, silly question, do we define these labels anywhere? The spec doesn't seem super descriptive: https://fullnode.devnet.aptoslabs.com/v1/spec#/operations/estimate_gas_price. Might be worth updating the spec and/or docs to call the new bands out?

api/src/context.rs
```diff
@@ -797,27 +806,168 @@ impl Context {
...
+        let aggressive_price = self.next_bucket(p90_price);
+
+        let estimation = GasEstimation {
+            deprioritized_gas_estimate: Some(low_price),
+            gas_estimate: market_price,
+            prioritized_gas_estimate: Some(aggressive_price),
+        };
+        // 4. Update cache
+        // GC old entries
+        if cache.min_inclusion_prices.len() > max_block_history {
```
JoshLind, line 962: Do we want to add a unit test to ensure we are doing garbage collection, i.e., that the cache never grows indefinitely?

api/src/context.rs
```diff
@@ -797,27 +806,168 @@ impl Context {
...
+        if blocks.is_empty() && !cached_blocks_hit {
+            let estimation = self.default_gas_estimation(min_gas_unit_price);
+            cache.estimation = Some(estimation);
+            cache.last_updated_epoch = Some(epoch);
+            cache.last_updated_time = Some(Instant::now());
+            return Ok(estimation);
+        }
+        let remaining = max_block_history - blocks.len();
+
+        // 2. Get gas prices per block
```
JoshLind, line 902: nit: should this code block be in a helper function?

JoshLind (commented):
Awesome, thanks @bchocho 😄 First pass looks reasonable to me -- only nits and questions.

api/goldens/aptos_api__tests__transactions_test__test_gas_estimation_one_block.json
bchocho: (1) Good callout about changing the spec/docs. I'll update the PR with these changes.

(2) I'm not sure what breaks if we change the json field names, so I'm playing it safe by reusing what's already there.
@gregnazario do you have any insight here? Is it possible to change the names, or would we have to create a new API endpoint and then deprecate this one?

api/src/context.rs
bchocho: I guess it's a style issue - I'd rather not put this into a helper function as it has a few shared variables with the other parts and won't be reusable.

bchocho:
@JoshLind thanks! I've addressed your comments (except replied to the last one)

api/src/context.rs
JoshLind: Makes sense 😄 Generally, I'll try add configs for things that I suspect will be worth tuning at runtime, so that I have some levers to pull in case things don't behave as expected, but I get that there's a trade-off. Your call 😃

JoshLind (approved):
Nice, thanks @bchocho! Looks solid to me 😄

bchocho:
@JoshLind I actually updated the spec a bit, as reflected in https://github.com/aptos-foundation/AIPs/pull/145. Hope this makes sense!

api/src/context.rs
```diff
@@ -797,27 +806,185 @@ impl Context {
...
-        })
+
+        // (2) market
+        let mut latest_prices: Vec<_> = min_inclusion_prices
+            .iter()
+            .take(config.market_block_history)
+            .cloned()
+            .collect();
+        latest_prices.sort();
+        let market_price = latest_prices[latest_prices.len() / 2];
```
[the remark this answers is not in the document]
bchocho: Generally I agree. In these cases though we are using a fraction of the vector length, so I'm highly confident there's no out of bounds error that we can hit.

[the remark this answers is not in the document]
JoshLind: > so I'm highly confident there's no out of bounds error that we can hit.

Bold 😄 SGTM.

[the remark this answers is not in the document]
bchocho: I just took your advice and used get. Not feeling so bold anymore :P

pr-aptos (approved):
stamping to unblock devnet

PR #8194, opened 2023-05-15
Define fungible asset, token, and unit in the glossary

saharct:
### Description

Let's generically define these terms to consider current and future ways of creating fungible assets. We can rephrase or add a bullet if more specificity is needed, but we should first work toward clarifying any confusion within the standards docs.

### DIFF
--- a/developer-docs-site/docs/reference/glossary.md
+++ b/developer-docs-site/docs/reference/glossary.md
@@ -187,6 +187,15 @@ then there is a guarantee that T_N will never be included in the blockchain.
 
 - **Fullnodes** are clients that ensure data are stored up-to-date on the network. They replicate blockchain state and transactions from other fullnodes and validator nodes.
 
+### Fungible Asset
+
+- A **fungible asset** is an asset, such as a currency, share, in-game resource, etc., that can be interchanged with another identical asset without any loss in its value. For example, APT is fungible because you can exchange one APT for another.
+- Follow the [Digital Asset Standards](../standards/index.md#digital-asset-standards) to create fungible assets on the blockchain, including tokens and coins. 
+
+### Fungible Unit
+
+- A **fungible unit** is an individual unit of a fungible asset. These units are identical and can be interchanged without any loss in value. For example, each Octa (the smallest unit of APT) is a fungible unit.
+
 ## G
 
 ### Gas
--- a/developer-docs-site/docs/reference/glossary.md
+++ b/developer-docs-site/docs/reference/glossary.md
@@ -191,10 +191,12 @@ then there is a guarantee that T_N will never be included in the blockchain.
 
 - A **fungible asset** is an asset, such as a currency, share, in-game resource, etc., that is interchangeable with another identical asset without any loss in its value. For example, APT is a fungible asset because you can exchange one APT for another.
 - Follow the [Digital Asset Standards](../standards/index.md#digital-asset-standards) to create fungible assets on the Aptos blockchain. 
+- Next generation of the Coin standard that addresses shortcomings of `aptos_framework::coin` such as lack of guaranteed enforcement of freeze and burn and advanced functionalities such as programmable transfers, e.g., approve in ERC-20.
 
 ### Fungible Token
 
-- A **fungible token** is a token that is interchangeable with other identical tokens (i.e., tokens that share the same `TokenId`). This means the tokens have the same `creator address`, `collection name`, `token name`, and `property version`.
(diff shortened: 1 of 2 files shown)


developer-docs-site/docs/reference/glossary.md
```diff
@@ -187,6 +187,15 @@ then there is a guarantee that T_N will never be included in the blockchain.
 
 - **Fullnodes** are clients that ensure data are stored up-to-date on the network. They replicate blockchain state and transactions from other fullnodes and validator nodes.
 
+### Fungible Asset
+
+- A **fungible asset** is an asset, such as a currency, share, in-game resource, etc., that can be interchanged with another identical asset without any loss in its value. For example, APT is fungible because you can exchange one APT for another.
+- Follow the [Digital Asset Standards](../standards/index.md#digital-asset-standards) to create fungible assets on the blockchain, including tokens and coins.
```
lightmark, line 193: tokens usually means nft in our docs, which is not FA.

lightmark (commented):
I think another term is `fungible token`?

developer-docs-site/docs/reference/glossary.md
saharct: But we also have docs where we refer to FTs as tokens. [AIP-21](https://github.com/aptos-foundation/AIPs/blob/main/aips/aip-21.md) uses terms like "tokenization" and "tokenized" to describe FAs. That being said, I'll remove the last bit from the second bullet point so that we can let the context in each doc decide.

saharct:
> I think another term is `fungible token`?

Yes, I'll add `Fungible Token` too. I initially wanted to mention tokens as a potential type of FA to avoid a separate definition, but it's worth defining `Fungible Token` separately.

developer-docs-site/docs/reference/glossary.md
```diff
@@ -187,6 +187,19 @@ then there is a guarantee that T_N will never be included in the blockchain.
...
 - **Fullnodes** are clients that ensure data are stored up-to-date on the network. They replicate blockchain state and transactions from other fullnodes and validator nodes.
 
+### Fungible Asset
+
+- A **fungible asset** is an asset, such as a currency, share, in-game resource, etc., that is interchangeable with another identical asset without any loss in its value. For example, APT is a fungible asset because you can exchange one APT for another.
+- Follow the [Digital Asset Standards](../standards/index.md#digital-asset-standards) to create fungible assets on the Aptos blockchain. 
+
+### Fungible Token
+
+- A **fungible token** is a token that is interchangeable with other identical tokens (i.e., tokens that share the same `TokenId`). This means the tokens have the same `creator address`, `collection name`, `token name`, and `property version`.
```
lightmark, line 197: this is different in token v2. fungible token should be a special type of fungible asset that the asset metadata is a token.

saharct: I defined v1 first since the v2 docs are not public yet. I'm okay with updating the definition in this PR to include v1 and v2, then waiting until the other docs land before landing this PR.

lightmark: Do we want to keep it in general?

developer-docs-site/docs/reference/glossary.md
```diff
@@ -187,6 +187,19 @@ then there is a guarantee that T_N will never be included in the blockchain.
 
 - **Fullnodes** are clients that ensure data are stored up-to-date on the network. They replicate blockchain state and transactions from other fullnodes and validator nodes.
 
+### Fungible Asset
+
+- A **fungible asset** is an asset, such as a currency, share, in-game resource, etc., that is interchangeable with another identical asset without any loss in its value. For example, APT is a fungible asset because you can exchange one APT for another.
+- Follow the [Digital Asset Standards](../standards/index.md#digital-asset-standards) to create fungible assets on the Aptos blockchain. 
+
```
davidiw, line 194:
```suggestion
- Follow the [Digital Asset Standards](../standards/index.md#digital-asset-standards) to create fungible assets on the Aptos blockchain. 
- Next generation of the Coin standard that addresses shortcomings of `aptos_framework::coin` such as lack of guaranteed enforcement of freeze and burn and advanced functionalities such as programmable transfers, e.g., approve in ERC-20.

```

developer-docs-site/docs/reference/glossary.md
davidiw: agreed, let's define two bullets here.

davidiw, line 197:
```suggestion
- For TokenV1 (aptos_token::token), a **fungible token** is a token that is interchangeable with other identical tokens (i.e., tokens that share the same `TokenId`). This means the tokens have the same `creator address`, `collection name`, `token name`, and `property version`.
- For TokenV2 (aptos_token_objects::token), a **fungible token** is a fungible asset with metadata object includes a TokenV2 resource.
```

davidiw (approved):

PR #8200, opened 2023-05-15
[TS] Added a typescript client for the Ambassador token example

junkil-park:
### Description
* Added a typescript client that manages the Ambassador token.
* Updated the Ambassador token contract regarding "rank" and "uri"

### Test Plan
pnpm run ambassador


aptos-move/move-examples/token_objects/ambassador/client/ambassador.ts
```diff
@@ -0,0 +1,171 @@
...
+      }
+      token_standard
+    }
+  }
+`;
+    const data = await fetchGraphQL(operation, "TokenOwnership", {});
+    const tokenOwnership = data.data;
+
+    // TODO: `tokenOwnership` can be obtained by the following code when a new SDK version is ready.
+    // const tokenOwnership = await provider.getTokenOwnershipV2(ownerAddr);
```
movekevin, line 49: Can you integrate https://github.com/aptos-labs/aptos-core/pull/8210 (might need to change tsconfig.json to depend on local code)? If we can test this out, we can avoid hardcoding the query here. cc @0xmaayan

aptos-move/move-examples/token_objects/ambassador/client/ambassador.ts
```diff
@@ -0,0 +1,171 @@
...
+
+    for (const ownership of tokenOwnership.current_token_ownerships_v2) {
+        if (ownership.current_token_data.token_name === tokenName) {
```
movekevin, line 52: Can we allow filtering by token name in the query? cc @0xmaayan @bowenyang007

aptos-move/move-examples/token_objects/ambassador/client/ambassador.ts
```diff
@@ -0,0 +1,171 @@
...
+        return pendingTxn.hash;
+    }
+}
+
+/** run our demo! */
+async function main(): Promise<void> {
+    const client = new AmbassadorClient();
+
+    // Set a test admin account
+    const privateKeyBytes_admin = Uint8Array.from(Buffer.from('f21423f436f7d44c2abd95b5a25323e81fc737040ab17ae8fe40dbf1b1de9e66', 'hex'));
```
movekevin, line 124: Can we generate a key instead of hardcoding private keys? I'd avoid having any private key even throwaway ones in the repo

aptos-move/move-examples/token_objects/ambassador/client/ambassador.ts
```diff
@@ -0,0 +1,171 @@
...
+    const userAddr = new HexString("4db1582c315ddd9f29db3dfcf0aa7f7467b1a4f2d1190bb93b8304cdc164490c");
+    // Set a test token name
+    const tokenName = "Aptos Ambassador #23";
+
+    console.log("\n=== Addresses ===");
+    console.log(`Admin: ${admin.address()}`);
+    console.log(`User: ${userAddr}`);
+
+    // Mint Ambassador Token
+    let txnHash = await client.mintAmbassadorToken(admin, 'Aptos Ambassador Token', tokenName, 'https://raw.githubusercontent.com/junkil-park/metadata/main/ambassador_1/', userAddr);
```
movekevin, line 136: Can we dump the metadata into this directory so we don't need to point to the link?

aptos-move/move-examples/token_objects/ambassador/client/ambassador.ts
```diff
@@ -0,0 +1,171 @@
...
+        token_name
+        current_collection {
+          collection_name
+        }
+      }
+      token_standard
+    }
+  }
+`;
+    const data = await fetchGraphQL(operation, "TokenOwnership", {});
```
movekevin, line 45: Is there a way to fetch the token's data (i.e. level and rank) using a single request/query? cc @0xmaayan @bowenyang007

movekevin (commented):
Nice!!!

aptos-move/move-examples/token_objects/ambassador/client/ambassador.ts
junkil-park: This single query also fetches the `rank` data. It is an entry of the property map, so `token_properties` should contain it. 
However, `level` is different because it's a top-level Move resource. Currently, a specific view function or a custom indexer can get the value. (@0xmaayan , @bowenyang007 )

aptos-move/move-examples/token_objects/ambassador/client/ambassador.ts
0xmaayan: SDK can provide something like `getToken(owner, tokenName)` ?  @bowenyang007 any concerns about performance?

aptos-move/move-examples/token_objects/ambassador/client/ambassador.ts
0xmaayan: we talked about it. I created an example PR on how to use a local SDK build https://github.com/aptos-labs/aptos-core/pull/8133. 
Once https://github.com/aptos-labs/aptos-core/pull/8210 is merged, we can publish a new version and @junkil-park can use the updated version instead of building from local.

aptos-move/move-examples/token_objects/ambassador/client/ambassador.ts
0xmaayan: SDK provides view function call https://github.com/aptos-labs/aptos-core/blob/main/ecosystem/typescript/sdk/src/tests/e2e/aptos_client.test.ts#L602-L658
and Indexer has `move_resources` query

aptos-move/move-examples/token_objects/ambassador/client/ambassador.ts
```diff
@@ -0,0 +1,171 @@
+// Copyright © Aptos Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+import { AptosAccount, AptosClient, HexString } from "aptos";
+
+const NODE_URL = process.env.APTOS_NODE_URL || "https://fullnode.testnet.aptoslabs.com";
```
movekevin, line 6: nit: Use these values from common.ts instead

aptos-move/move-examples/token_objects/ambassador/client/ambassador.ts
movekevin: Yeah I was thinking in the query you can join with move_resources to get level. This way you wouldn't need to send 2 separate requests to get all the data you need

aptos-move/move-examples/token_objects/ambassador/client/ambassador.ts
junkil-park: Updated to use the published `aptos` sdk.

aptos-move/move-examples/token_objects/ambassador/client/ambassador.ts
junkil-park: Good point. Done.

aptos-move/move-examples/token_objects/ambassador/client/ambassador.ts
junkil-park: This Typescript client example is in `aptos-move/move-examples/token_objects/ambassador/` (near the Move contracts), not in the same directory where `common.ts` is.

aptos-move/move-examples/token_objects/ambassador/client/ambassador.ts
```diff
@@ -0,0 +1,144 @@
...
+        });
+
+        rl.question('Please press the Enter key to proceed ...\n', () => {
+            rl.close();
+            resolve();
+        });
+    });
+}
+
+class AmbassadorClient extends AptosClient {
```
0xmaayan, line 34: why do you need to extend `AptosClient`? I think you dont need it since you use Provider class.

aptos-move/move-examples/token_objects/ambassador/client/ambassador.ts
```diff
@@ -0,0 +1,144 @@
...
+        };
+
+        const result = await this.view(payload);
+        return BigInt(result[0].toString());
+    }
+}
+
+/** run our demo! */
+async function main(): Promise<void> {
+    const client = new AmbassadorClient();
```
0xmaayan, line 92: you dont need it, we [initialize](https://github.com/aptos-labs/aptos-core/pull/8200/files#diff-7207e18b0700e512a3b28cc6c31fa794b4ce636e9517d3ef48e5fd086ec6a0d8R7) a Provider class so can simply use that, i.e every `client.` can become `provider.`

aptos-move/move-examples/token_objects/ambassador/client/ambassador.ts
junkil-park: This example requests a separate query via calling the view function to get the AmbassadorLevel.

We could do it in a single query if we extended our custom indexer and interact with it: https://github.com/aptos-labs/aptos-indexer-processors/pull/24

aptos-move/move-examples/token_objects/ambassador/client/ambassador.ts
junkil-park: Done.

aptos-move/move-examples/token_objects/ambassador/client/ambassador.ts
junkil-park: Removed.

aptos-move/move-examples/token_objects/ambassador/client/ambassador.ts
junkil-park: I am trying, but it's a bit tricky because of the compilation and publication steps.

aptos-move/move-examples/token_objects/ambassador/move/sources/ambassador.move
```diff
@@ -47,12 +49,16 @@ module token_objects::ambassador {
...
+        /// Used to mutate the token uri
+        mutator_ref: token::MutatorRef,
         /// Used to burn.
         burn_ref: token::BurnRef,
         /// Used to mutate properties
         property_mutator_ref: property_map::MutatorRef,
         /// Used to emit LevelUpdateEvent
         level_update_events: event::EventHandle<LevelUpdateEvent>,
+        /// the base URI of the token
+        base_uri: String,
```
movekevin, line 61: This is an interesting choice to have this here vs the collection since this does enable more parallelism for operations that modify the rank of the token. I'd add a line here (or in the tutorial to be written) to explain this

aptos-move/move-examples/token_objects/ambassador/client/ambassador.ts
movekevin: I see. May be you can leave that as a follow up for that example after the other examples are done

aptos-move/move-examples/token_objects/ambassador/client/ambassador.ts
movekevin: Hmm I didn't notice this. I think we should move ambassador.ts there under ecosystem/ instead of here. This is how we structure tutorials - the move modules in aptos-move/move-examples while the TS parts are in ecosystem/

aptos-move/move-examples/token_objects/ambassador/client/ambassador.ts
junkil-park: After creating a random "admin" account, we need to compile and publish the Move package with the account. Especially the compilation step is not easy to automate. For this, I will just follow this way letting the user compile it manually: https://github.com/aptos-labs/aptos-core/blob/main/ecosystem/typescript/sdk/examples/typescript/your_coin.ts#L105

aptos-move/move-examples/token_objects/ambassador/client/ambassador.ts
junkil-park: I see. I was following the example of `https://github.com/aptos-labs/aptos-core/tree/main/aptos-move/move-examples/my_first_dapp`. OK, I will move the TS parts into `ecosystem`.

aptos-move/move-examples/token_objects/ambassador/move/sources/ambassador.move
junkil-park: "parallelism" is an interesting point! `base_uri` here is intended to be the base URI for a token rather than a collection. In this example, each token has three URIs for gold, silver, and bronze images, so needs a common prefix.

ecosystem/typescript/sdk/examples/typescript/ambassador.ts
```diff
@@ -0,0 +1,153 @@
...
+    const adminAddr = admin.address()
+    const userAddr = user.address()
+    const tokenName = "Aptos Ambassador #1";
+
+    console.log("\n=== Addresses ===");
+    console.log(`Admin: ${adminAddr} `);
+    console.log(`User: ${userAddr} `);
+
+    // Mint Ambassador Token
+    let txnHash = await client.mintAmbassadorToken(admin, 'Aptos Ambassador Token', tokenName, 'https://raw.githubusercontent.com/aptos-labs/aptos-core/main/aptos-move/move-examples/token_objects/ambassador/metadata/', userAddr);
```
movekevin, line 116: Why not also move these images to here in ecosystem/ ?

movekevin (approved):

ecosystem/typescript/sdk/examples/typescript/ambassador.ts
junkil-park: Done.e

ecosystem/typescript/sdk/examples/typescript/ambassador.ts
```diff
@@ -0,0 +1,171 @@
...
+    await faucetClient.fundAccount(user.address(), 100_000_000);
+
+    console.log(
+        "\nCompile and publish the Ambassador module (`aptos-core/aptos-move/move-examples/token_objects/ambassador`) using the following profile, and press enter:",
+    );
+    console.log(`  ambassador_admin:`);
+    console.log(`    private_key: "${admin.toPrivateKeyObject().privateKeyHex}"`);
+    console.log(`    public_key: "${admin.pubKey()}"`);
+    console.log(`    account: ${admin.address()}`);
+    console.log(`    rest_url: "https://fullnode.testnet.aptoslabs.com"`);
```
0xmaayan, line 114: devnet?
```suggestion
    console.log(`    rest_url: "https://fullnode.devnet.aptoslabs.com"`);
```

PR #8202, opened 2023-05-15
Add emit_events transaction to transaction generator

vusirikala:
### Description

Add an `emit_events` transaction in transaction-generator that generates the given number of events. This transaction is added to test if the state sync works fine when the output of a transaction is huge.


igor-aptos (approved):

JoshLind (approved):
Nice! Thanks @vusirikala 😄

PR #8207, opened 2023-05-15
[docker] add docker-compose for indexer-grpc

rustielin:
### Description

Edits the existing `validator-testnet` docker-compose:
* Enable indexer grpc in the default config. (Since this whole setup is for testing, it makes sense to just enable all features. If folks feel strongly, we can remove this and just add in a README to edit the config to enable indexer GRPC instead)

New `indexer-grpc` docker-compose, including:
* cache worker
* file store (with local filestore)
* data service
* redis

Mounts into existing docker compose networks, using static IPs. (We can re-evaluate getting rid of these static IPs, as they were only necessary when we faced docker networking issues in CircleCI)

A simple check is implemented in GHA that checks if we can connect to the indexer-grpc via `grpcurl`. For now (since the workflow is not landed in main yet), we test on `pull_request`, and use a hard-coded recent docker image build to spin up the whole docker-compose setup.

Bonus:
* Get the validator-testnet docker-compose to work with tap, after finally updating the image (this devnet release via https://github.com/aptos-labs/aptos-core/pull/8136). With this week's devnet release, we should publish an announcement calling this out.

### Test Plan

New e2e test https://github.com/aptos-labs/aptos-core/actions/runs/4986678324/jobs/8927692834?pr=8207
Also verify locally (which is what the test does): https://github.com/aptos-labs/aptos-core/blob/08c67319c6dc15f6585a3db2ae94b1ae4d087b7e/.github/workflows/docker-indexer-grpc.yaml
* export some vars to select docker image to use: `FAUCET_IMAGE_REPO`, `VALIDATOR_IMAGE_REPO`, `INDEXER_GRPC_IMAGE_REPO`, `IMAGE_TAG`
* in docker/compose/validator-testnet: `docker-compose up -d`
* in docker/compose/indexer-grpc: `docker-compose up -d`
* hit some endpoints using `grpcurl`

### DIFF
--- a/docker/compose/indexer-grpc/test_indexer_grpc_docker_compose.sh
+++ b/docker/compose/indexer-grpc/test_indexer_grpc_docker_compose.sh
@@ -1,0 +1,46 @@
+#!/bin/bash
+
+# A quick script that checks the e2e setup for the indexer-grpc service on docker-compose
+
+if ! command -v grpcurl &>/dev/null; then
+    wget https://github.com/fullstorydev/grpcurl/releases/download/v1.8.7/grpcurl_1.8.7_linux_x86_64.tar.gz
+    sha=$(shasum -a 256 grpcurl_1.8.7_linux_x86_64.tar.gz | awk '{ print $1 }')
+    [ "$sha" != "b50a9c9cdbabab03c0460a7218eab4a954913d696b4d69ffb720f42d869dbdd5" ] && echo "shasum mismatch" && exit 1
+    tar -xvf grpcurl_1.8.7_linux_x86_64.tar.gz
+    chmod +x grpcurl
+    mv grpcurl /usr/local/bin/grpcurl
+    grpcurl -version
+fi
+
+# Try hitting the indexer-grpc setup in a number of ways
+# 
+
+# try getting the internal grpc on the fullnode itself
+stream_time_seconds=30
+start_time=$(date +%s)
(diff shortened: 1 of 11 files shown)


.github/workflows/docker-indexer-grpc.yaml
```diff
@@ -0,0 +1,61 @@
+name: "Docker Indexer gRPC"
```
sherry-x, line 1: rename to "Docker Indexer gRPC test" to avoid confusion this for building indexer docker

sherry-x (approved):
Love it! Let's make sure we create documentation on how to use this later.

docker/compose/indexer-grpc/test_indexer_grpc_docker_compose.sh
```diff
@@ -0,0 +1,46 @@
...
+echo "grpcurl took ${total_time} seconds to run"
+
+if [ $total_time -lt "${stream_time_seconds}" ]; then
+    echo "grpcurl exited early, which indicates failure"
+    echo "RawData/GetTransactions on the aptos-node should be an endless stream"
+    exit 1
+fi
+
+# try hitting the data service
+stream_time_seconds=30
```
larry-aptos, line 33: is the validator node up at this point? the reason is, sometimes it may take sometime to setup the node and data can stream out after it.

larry-aptos (approved):

docker/compose/indexer-grpc/test_indexer_grpc_docker_compose.sh
rustielin: In my testing, the node at least in `--test` mode spins up pretty instantaneously. The proper way to do this would be to wait on the node status by polling its endpoints, but I've just added in a sleep after spinning up the node, since the indexer GRPC will also crash loop currently if the node is unavailable.

rustielin:
I'm going to hold off on landing this until the next devnet release, when the new faucet and indexer grpc images are released. This docker-compose only works with those up to date images

PR #8208, opened 2023-05-15
[Gas Estimation] Include gas used as a threshold for marking blocks full

bchocho:
### Description

When doing gas estimation, add criteria for considering a block full if it has reached the gas limit. (Previously, only the txn count was considered.) The gas limit is read from the onchain execution config.

### Test Plan

Added smoke test that relies on the gas limit block full value.

### DIFF
--- a/api/src/context.rs
+++ b/api/src/context.rs
@@ -35,7 +35,7 @@ use aptos_types::{
     contract_event::EventWithVersion,
     event::EventKey,
     ledger_info::LedgerInfoWithSignatures,
-    on_chain_config::{GasSchedule, GasScheduleV2, OnChainConfig},
+    on_chain_config::{GasSchedule, GasScheduleV2, OnChainConfig, OnChainExecutionConfig},
     state_store::{
         state_key::{StateKey, StateKeyInner},
         state_key_prefix::StateKeyPrefix,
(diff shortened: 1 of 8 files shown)


api/src/context.rs
```diff
@@ -95,6 +96,10 @@ impl Context {
                 estimation: None,
                 min_inclusion_prices: BTreeMap::new(),
             })),
+            gas_limit_cache: Arc::new(RwLock::new(GasLimitCache {
```
danielxiangzl, line 99: When will `Context` be initialized? Every epoch change?

msmouse:
Nice moving the API out of DbReader. Looks neat. Thanks!

api/src/context.rs
```diff
@@ -65,6 +65,7 @@ pub struct Context {
     pub node_config: NodeConfig,
     gas_schedule_cache: Arc<RwLock<GasScheduleCache>>,
     gas_estimation_cache: Arc<RwLock<GasEstimationCache>>,
+    gas_limit_cache: Arc<RwLock<GasLimitCache>>,
```
JoshLind, line 68: nit: FWIW, it might make sense to create a single `gas_caches` struct that contains all sub-structs. I worry that we're starting to litter the `Context` object (which is used in more places than just the gas estimation endpoints, IIUC).

api/src/context.rs
```diff
@@ -855,12 +860,43 @@ impl Context {
...
+        &self,
+        start_version: Version,
+        limit: u64,
+        ledger_version: Version,
+    ) -> Result<Vec<(u64, u64)>> {
+        if start_version > ledger_version || limit == 0 {
+            return Ok(vec![]);
+        }
+
+        // This is just an estimation, so we cna just skip over errors
```
JoshLind, line 873: cna -> can

JoshLind (approved):
Unblocking -- LGTM.

JoshLind (approved):
Unblocking -- LGTM.

bchocho:
Thanks @JoshLind let me clean these up in a follow-up

PR #8213, opened 2023-05-15
[aptos-vm] Charge invariant violation txn

runtian-zhou:
### Description

Charges gas for TXNs that lead to an invariant violation error, when the feature flag is set.

Will need to decide if we need to enable this flag in the upcoming release.

### Test Plan

Uses failpoint injection to make sure that:
1. With the flag enabled, an invariant violation error TXN will be kept and charged gas.
2. With the flag disabled, the TXN will be marked as discarded.

### DIFF
--- a/aptos-move/aptos-vm/src/aptos_vm.rs
+++ b/aptos-move/aptos-vm/src/aptos_vm.rs
@@ -259,7 +259,12 @@ impl AptosVM {
             .0
             .new_session(resolver, SessionId::txn_meta(txn_data), true);
 
-        match TransactionStatus::from(error_code.clone()) {
+        match TransactionStatus::from_vm_status(
+            error_code.clone(),
+            self.0
+                .get_features()
+                .is_enabled(FeatureFlag::CHARGE_INVARIANT_VIOLATION),
+        ) {
             TransactionStatus::Keep(status) => {
                 // Inject abort info if available.
                 let status = match status {
@@ -1098,7 +1103,12 @@ impl AptosVM {
                     );
                 }
 
-                let txn_status = TransactionStatus::from(err.clone());
+                let txn_status = TransactionStatus::from_vm_status(
+                    err.clone(),
+                    self.0
+                        .get_features()
+                        .is_enabled(FeatureFlag::CHARGE_INVARIANT_VIOLATION),
+                );
                 if txn_status.is_discarded() {
                     discard_error_vm_status(err)
                 } else {
@@ -1857,7 +1867,13 @@ impl AptosSimulationVM {
                 if new_published_modules_loaded {
                     self.0 .0.mark_loader_cache_as_invalid();
                 };
-                let txn_status = TransactionStatus::from(err.clone());
+                let txn_status = TransactionStatus::from_vm_status(
+                    err.clone(),
+                    self.0
+                         .0
+                        .get_features()
+                        .is_enabled(FeatureFlag::CHARGE_INVARIANT_VIOLATION),
+                );
                 if txn_status.is_discarded() {
                     discard_error_vm_status(err)
                 } else {
--- a/aptos-move/aptos-vm/src/aptos_vm.rs
+++ b/aptos-move/aptos-vm/src/aptos_vm.rs
@@ -41,7 +41,7 @@ use aptos_types::{
         TransactionOutput, TransactionPayload, TransactionStatus, VMValidatorResult,
         WriteSetPayload,
     },
-    vm_status::{AbortLocation, DiscardedVMStatus, StatusCode, VMStatus},
+    vm_status::{AbortLocation, StatusCode, VMStatus},
     write_set::WriteSet,
 };
 use aptos_utils::{aptos_try, return_on_failure};
@@ -1671,9 +1671,7 @@ impl VMAdapter for AptosVM {
                 let _timer = TXN_TOTAL_SECONDS.start_timer();
                 let (vm_status, output) = self.execute_user_transaction(resolver, txn, log_context);
 
-                if let Err(DiscardedVMStatus::UNKNOWN_INVARIANT_VIOLATION_ERROR) =
-                    vm_status.clone().keep_or_discard()
-                {
+                if let StatusType::InvariantViolation = vm_status.status_type() {
                     error!(
                         *log_context,
                         "[aptos_vm] Transaction breaking invariant violation. txn: {:?}, status: {:?}",
--- a/aptos-move/e2e-testsuite/src/tests/invariant_violation.rs
+++ b/aptos-move/e2e-testsuite/src/tests/invariant_violation.rs
@@ -1,0 +1,66 @@
+// Copyright © Aptos Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+use aptos_language_e2e_tests::{common_transactions::peer_to_peer_txn, executor::FakeExecutor};
+use aptos_types::{
+    account_address::AccountAddress,
+    on_chain_config::FeatureFlag,
+    transaction::{ExecutionStatus, TransactionStatus},
+    vm_status::DiscardedVMStatus,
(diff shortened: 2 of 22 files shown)


aptos-move/aptos-vm/src/aptos_vm.rs
```diff
@@ -245,7 +245,11 @@ impl AptosVM {
     ) -> (VMStatus, TransactionOutputExt) {
         let mut session = self.0.new_session(resolver, SessionId::txn_meta(txn_data));
 
-        match TransactionStatus::from(error_code.clone()) {
+        match TransactionStatus::from(error_code.clone()).translate_invariant_violation(
```
runtian-zhou, line 248: The logic change here is pretty ugly but I couldn't find a better way to clean this up. Would appreciate any suggestions. cc @gerben-stavenga @vgao1996

types/src/transaction/mod.rs
```diff
@@ -825,6 +825,18 @@ impl TransactionStatus {
             _ => Err(format_err!("Not Keep.")),
         }
     }
+
+    pub fn translate_invariant_violation(self, charge_invariant_violation: bool) -> Self {
```
gerben-stavenga, line 829: Is this the best solution? To go from invariant_violation -> discard -> keep anyway

movekevin: +1 this is janky

runtian-zhou:
I slightly changed the logic to move the error code conversion into `aptos_types` but most of the heavy lifting is still in the move repo. Lemme know if this looks better and I can creat a separate PR to move those logic into `aptos_types` as well

vgao1996:
> but most of the heavy lifting is still in the move repo

IIUC this is because that logic shouldn't be in the move repo in the first place. Concepts like Keep or Discard were specific to Diem, and now Aptos.

types/src/transaction/mod.rs
```diff
@@ -837,7 +835,15 @@ impl From<VMStatus> for TransactionStatus {
                 },
                 _ => TransactionStatus::Keep(recorded.into()),
             },
-            Err(code) => TransactionStatus::Discard(code),
+            Err(code) => {
+                if code.status_type() == StatusType::InvariantViolation
+                    && charge_invariant_violation
+                {
+                    TransactionStatus::Keep(ExecutionStatus::MiscellaneousError(Some(code)))
```
vgao1996, line 842: It might be cleaner to add a separate variant `ExecutionStatus::InvariantViolation` since `ExecutionStatus::MiscellaneousError` has been historically used for execution errors that can be triggered by user inputs. Don't know if we'll run into issues with this though. The whole error conversion story is a mess that was left out in the previous rounds of refactoring.

runtian-zhou: I'm a bit concerned about adding another category cuz the storage might have trouble deserializing this new error category

vgao1996 (approved):

runtian-zhou:
Also drafted a tiny AIP for this: https://github.com/aptos-foundation/AIPs/pull/143

types/src/transaction/mod.rs
```diff
@@ -837,11 +835,25 @@ impl From<VMStatus> for TransactionStatus {
...
+                    TransactionStatus::Discard(code)
+                }
+            },
         }
     }
 }
 
+impl From<VMStatus> for TransactionStatus {
+    fn from(vm_status: VMStatus) -> Self {
+        TransactionStatus::from_vm_status(vm_status, true)
```
alinush, line 853: Hopefully, having the flag enabled here by default is the intended behavior.

alinush (approved):
Just an ACK stamp: PR makes sense, not sure if implementation is correct since I am no Move VM expert.

gerben-stavenga (approved):

aptos-move/vm-genesis/src/lib.rs
```diff
@@ -434,6 +434,7 @@ pub fn default_features() -> Vec<FeatureFlag> {
         FeatureFlag::CRYPTOGRAPHY_ALGEBRA_NATIVES,
         FeatureFlag::BLS12_381_STRUCTURES,
         FeatureFlag::STORAGE_SLOT_METADATA,
+        FeatureFlag::CHARGE_INVARIANT_VIOLATION,
```
movekevin, line 437: cc @perryjrandall Let's look into removing these hardcoded values in genesis and favoring a testnet-like upgrade process for devnet

types/src/transaction/mod.rs
movekevin: Why would storage have trouble deserializing this if we change all the locations that look at the different variations of ExecutionStatus?

types/src/transaction/mod.rs
```diff
@@ -844,11 +842,25 @@ impl From<VMStatus> for TransactionStatus {
                 },
                 _ => TransactionStatus::Keep(recorded.into()),
             },
-            Err(code) => TransactionStatus::Discard(code),
+            Err(code) => {
+                if code.status_type() == StatusType::InvariantViolation
+                    && charge_invariant_violation
+                {
+                    TransactionStatus::Keep(ExecutionStatus::MiscellaneousError(Some(code)))
```
movekevin, line 849: This still looks like a bandaid solution to me. How much work is it to potentially move https://github.com/aptos-labs/aptos-core/blob/main/third_party/move/move-core/types/src/vm_status.rs#L167 to the code here to decide whether a VMStatus is discarded or not?

movekevin (commented):
I'd prefer either doing this right (rather than a 2-step conversion bandaid solution or at least making sure this would be addressed in the upcoming VM refactoring). Aptos VM is already convoluted when it comes to converting and handling errors/failures that adding even more complication would make the code effectively a minefield and making adding features or refactoring very risky in the future

pacan4ikx (approved):

movekevin (approved):
Approved to unblock with a TODO for moving keep_or_discard to transaction status code in a separate PR

aptos-move/e2e-testsuite/src/tests/invariant_violation.rs
```diff
@@ -0,0 +1,66 @@
...
+    // execute transaction
+    let output = executor.execute_transaction(txn.clone());
+
+    // CHARGE_INVARIANT_VIOLATION enabled at genesis so this txn is kept.
+    assert_eq!(
+        output.status(),
+        &TransactionStatus::Keep(ExecutionStatus::MiscellaneousError(Some(
+            StatusCode::UNKNOWN_INVARIANT_VIOLATION_ERROR
+        ))),
+    );
```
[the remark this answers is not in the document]
runtian-zhou: Yes there's a specific test for that!

aptos-move/e2e-testsuite/src/tests/invariant_violation.rs
```diff
@@ -0,0 +1,66 @@
...
+            FeatureFlag::CHARGE_INVARIANT_VIOLATION as u64,
+        )])
+        .simple_serialize()
+        .unwrap(),
+    ]);
+
+    let output = executor.execute_transaction(txn);
+
+    // With CHARGE_INVARIANT_VIOLATION disabled this transaction will be discarded.
+    assert_eq!(
```
runtian-zhou, line 62: Here's the test @movekevin

movekevin: This only tests the output though and not the gas charge

runtian-zhou: If it is discarded then it will definitelty not charge for gas right?

movekevin: Yeah but just adding an explicit check that gas has been deducted from the account (balance is less than in the beginning) provides absolute absolute confidence

PR #8217, opened 2023-05-16
Persisted dag and reliable_broadcast state

zjma:
### Description

Now `DagDriver` gets a `DagStorage`. `DagStorage` can give you a `DagStoreWriteBatch` for you to collect intended storage diffs, then accept your commit request at any time (typically at the end of the event handling/before send a message).

When being persisted to `DagStorage`, the big dag state is broken up into small objects with pointers pointing to each other. In-memory structs (and their helper methods) are tweaked to make the break-up effort easier.

Object partial updates are now possible with the `shallow_save()` API.

### TODOs
- [ ] Figure out a testing plan.
- [ ] Some write can still be smaller. See all the "only write the diff" comments.
- [ ] Apply the pattern to reliable_broadcast.


consensus/src/dag/dag.rs
```diff
@@ -5,344 +5,47 @@ use crate::{
...
+use crate::dag::dag_storage::naive::NaiveDagStoreWriteBatch;
+use crate::dag::types::{AbsentInfo, MissingDagNodeStatus, PendingInfo};
+use crate::dag::types::dag_in_mem::{DagInMem, DagInMem_Key};
+use crate::dag::types::dag_round_list::{DagRoundList, DagRoundListItem};
+use crate::dag::types::missing_node_status_map::{MissingNodeStatusMap, MissingNodeStatusMapEntry, MissingNodeStatusMapEntry_Key};
+use crate::dag::types::peer_node_map::{PeerNodeMap, PeerNodeMapEntry};
+use crate::dag::types::week_link_creator::WeakLinksCreator;
 
-// TODO: bug - what if I link to a node but before broadcasting I already create a node in the next round.
-enum PeerStatus {
```
zjma, line 20: These are just moved.

consensus/src/dag/dag_driver.rs
```diff
@@ -177,7 +188,9 @@ impl DagDriver {
     async fn handle_certified_node(&mut self, certified_node: CertifiedNode, ack_required: bool) {
         let digest = certified_node.digest();
         let source = certified_node.source();
-        self.dag.try_add_node(certified_node).await;
+        let mut storage_diff = self.dag_store.new_write_batch();
```
zjma, line 191: An example of using `DagStorage` + `DagStorageWriteBatch`.

PR #8219, opened 2023-05-16
[AptosVM] RespawnedSession to avoid duplication

msmouse:
### Description

Consolidate change set processing to one single function.
In preparation for introducing storage deletion refund, where the logic
needs to run in the respawned session introduced in this PR.

As a side effect, the multisig transaction flow now creates one less
session.

### Test Plan

unit tests and

replay-verify running
https://github.com/aptos-labs/aptos-core/actions/runs/4995140782


aptos-move/aptos-vm/src/move_vm_ext/respawned_session.rs
```diff
@@ -0,0 +1,122 @@
...
+use aptos_types::{
+    state_store::{
+        state_key::StateKey, state_storage_usage::StateStorageUsage, state_value::StateValue,
+    },
+    write_set::{TransactionWrite, WriteSet},
+};
+use move_core_types::vm_status::{err_msg, StatusCode, VMStatus};
+
+#[ouroboros::self_referencing]
+pub struct RespawnedSession<'r, 'l> {
```
georgemitenkov, line 22: Can we have documentation please? I find it really hard to work with VM sometimes because things are never documented and connected in a  very unintuitive way😂

aptos-move/aptos-aggregator/src/delta_change_set.rs
```diff
@@ -171,7 +171,7 @@ impl DeltaOp {
     /// error VM status is returned.
     pub fn try_into_write_op(
         self,
-        state_view: &impl StateView,
+        state_view: &dyn StateView,
```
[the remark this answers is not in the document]
msmouse: So the new types I introduce here don't need to carry `<S>`, similar to what you did for `SessionExt` I believe?

aptos-move/aptos-vm/src/move_vm_ext/respawned_session.rs
```diff
@@ -0,0 +1,127 @@
...
+            state_view,
+            resolver_builder: |state_view| state_view.as_move_resolver(),
+            session_builder: |resolver| Some(vm.new_session(resolver, session_id)),
+        }
+        .build())
+    }
+
+    pub fn execute<R>(&mut self, fun: impl FnOnce(&mut SessionExt) -> R) -> R {
+        self.with_session_mut(|session| fun(session.as_mut().unwrap()))
+    }
```
vgao1996, line 55: Would it be cleaner to just implement Deref?

aptos-move/aptos-vm/src/aptos_vm.rs
```diff
@@ -495,6 +464,28 @@ impl AptosVM {
...
+        gas_meter.charge_storage_fee_for_all(
+            change_set_ext.write_set().iter(),
+            change_set_ext.change_set().events(),
+            txn_data.transaction_size,
+            txn_data.gas_unit_price,
+        )?;
+
+        // TODO(Gas): Charge for aggregator writes
+        let session_id = SessionId::txn_meta(txn_data);
+        RespawnedSession::spawn(&self.0, session_id, resolver, change_set_ext)
```
vgao1996, line 486: Maybe I missed something, but here we're still spawning a new (Move VM) Session, right? Could you further elaborate on exactly what duplicated work this PR saves?

aptos-move/aptos-vm/src/move_vm_ext/respawned_session.rs
msmouse: You mean DerefMut? And I can't.. the struct. Ouroboros doesn't generate `.borrow_FIELD_mut()` https://docs.rs/ouroboros/latest/ouroboros/attr.self_referencing.html (search for "There is no borrow_FIELD_mut") (I'm not good enough to figure out exactly why)

aptos-move/aptos-vm/src/aptos_vm.rs
msmouse: It doesn't create one less session (except for the multi-sig flow, where yet another session was created). It removes duplicaton in code -- `charge_storage_fee_for_all()` for example is now called at only one place instead of 3.

aptos-move/aptos-vm/src/move_vm_ext/respawned_session.rs
```diff
@@ -0,0 +1,127 @@
...
+/// the base state view, and this struct implements that.
+#[ouroboros::self_referencing]
+pub struct RespawnedSession<'r, 'l> {
+    state_view: ChangeSetStateView<'r>,
+    #[borrows(state_view)]
+    #[covariant]
+    resolver: StorageAdapter<'this, ChangeSetStateView<'r>>,
+    #[borrows(resolver)]
+    #[not_covariant]
+    session: Option<SessionExt<'this, 'l>>,
```
gerben-stavenga, line 33: why is this option?

aptos-move/aptos-vm/src/move_vm_ext/respawned_session.rs
```diff
@@ -0,0 +1,127 @@
...
+}
+
+impl<'r> TStateView for ChangeSetStateView<'r> {
+    type Key = StateKey;
+
+    fn id(&self) -> StateViewId {
+        self.base.id()
+    }
+
+    fn get_state_value(&self, state_key: &Self::Key) -> Result<Option<StateValue>> {
```
gerben-stavenga, line 110: I think this is really good, no need for a double layer DeltaStateView

aptos-move/aptos-vm/src/move_vm_ext/respawned_session.rs
msmouse: I need to `take` and consume it on line #62, to get the new change set before destructing `self`.

aptos-move/aptos-vm/src/aptos_vm.rs
vgao1996: Ah now I see what you mean by "removing the duplication". Agreed that it makes sense to define a session at the Aptos VM level that stores and provides all the info one needs to execute a transaction, even if the Move VM stops doing this in the future.

aptos-move/aptos-vm/src/move_vm_ext/respawned_session.rs
```diff
@@ -0,0 +1,127 @@
...
+    write_set::{TransactionWrite, WriteSet},
+};
+use move_core_types::vm_status::{err_msg, StatusCode, VMStatus};
+
+/// We finish the session after the user transaction is done running to get the change set and
+/// charge gas and storage fee based on it before running storage refunds and the transaction
+/// epilogue. The latter needs to see the state view as if the change set is applied on top of
+/// the base state view, and this struct implements that.
+#[ouroboros::self_referencing]
+pub struct RespawnedSession<'r, 'l> {
```
gerben-stavenga, line 26: Overall I like the direction, but I'm not sold on this respawned session especially it requiring ouroboros

msmouse: What's the alternative in your mind? 
(The worst is we just keep repeating things and create one more session for now.).

gerben-stavenga (approved):

aptos-move/aptos-vm/src/move_vm_ext/respawned_session.rs
```diff
@@ -0,0 +1,127 @@
...
+    pub fn new(base: &'r dyn StateView, change_set: ChangeSetExt) -> Result<Self, VMStatus> {
+        // TODO: at this point we know that delta application failed
+        // (and it should have occurred in user transaction in general).
+        // We need to rerun the epilogue and charge gas. Currently, the use
+        // case of an aggregator is for gas fees (which are computed in
+        // the epilogue), and therefore this should never happen.
+        // Also, it is worth mentioning that current VM error handling is
+        // rather ugly and has a lot of legacy code. This makes proper error
+        // handling quite challenging.
+        let materialized_delta_change_set = change_set
```
georgemitenkov, line 91: cc @vusirikala

PR #8222, opened 2023-05-16
[move-examples] pre-minted managed fungible asset (coin)

lightmark:
### Description
1. Change the implementation of `managed_fungible_asset` to make it full-fledged with some flexibility. I feel it will be used by a lot of examples and we don't like to copy/paste the same part of code again and again.
3. Add an example with pre-minted managed coin using 1.
4. Add readme.

### Test Plan
ut


aptos-move/move-examples/fungible_asset/pre_minted_managed_coin/sources/preminted_managed_coin.move
```diff
@@ -0,0 +1,142 @@
+/// Same as `managed_coin` except that the total supply of coins are pre-minted when publishing the module. Nobody has
```
movekevin, line 1: Instead of saying "same as ...", I think it'd be better to list all the features that this supports here. It's more work but would be much less work for devs to understand what this does and whether it fits their needs

aptos-move/move-examples/fungible_asset/pre_minted_managed_coin/sources/preminted_managed_coin.move
```diff
@@ -0,0 +1,142 @@
...
+        let mint_ref = fungible_asset::generate_mint_ref(constructor_ref);
+        let admin_wallet = primary_fungible_store::ensure_primary_store_exists(
+            signer::address_of(admin),
+            get_metadata()
+        );
+        fungible_asset::mint_to(&mint_ref, admin_wallet, PRE_MINTED_TOTAL_SUPPLY);
+
+        // Create mint/burn/transfer refs to allow creator to manage the fungible asset.
+        let burn_ref = fungible_asset::generate_burn_ref(constructor_ref);
+        let transfer_ref = fungible_asset::generate_transfer_ref(constructor_ref);
```
movekevin, line 49: Let's kill transfer ref in this example since this feature (freezing stores or force transfer) is not core IMO for this example. This would keep the code here very concise

lightmark: I am going to refactor everything and reuse most of the code to solve all the issues you mentioned.

lightmark:
@movekevin I refactored everything, plz take another look.

aptos-move/move-examples/fungible_asset/preminted_managed_coin/sources/preminted_managed_coin.move
```diff
@@ -0,0 +1,63 @@
...
+
+    const ASSET_SYMBOL: vector<u8> = b"MEME";
+    const PRE_MINTED_TOTAL_SUPPLY: u64 = 10000;
+
+    /// Initialize metadata object and store the refs.
+    fun init_module(admin: &signer) {
+        let constructor_ref = &object::create_named_object(admin, ASSET_SYMBOL);
+        managed_fungible_asset::initialize(
+            constructor_ref,
+            0, /* maximum_supply. 0 means no maximum */
```
movekevin, line 22: Let's may be still add a maximum supply here (1B?) as an example

aptos-move/move-examples/fungible_asset/preminted_managed_coin/sources/preminted_managed_coin.move
```diff
@@ -0,0 +1,63 @@
...
+
+        // Create mint ref to premint fungible asset with a fixed supply volume into a specific account.
+        // This account can be any account including normal user account, resource account, multi-sig account, etc.
+        // We just use the creator account to show the proof of concept.
+        let mint_ref = fungible_asset::generate_mint_ref(constructor_ref);
+        let admin_primary_store = primary_fungible_store::ensure_primary_store_exists(
+            signer::address_of(admin),
+            get_metadata()
+        );
+        fungible_asset::mint_to(&mint_ref, admin_primary_store, PRE_MINTED_TOTAL_SUPPLY);
```
movekevin, line 39: May be we should considering adding mint/burn to primary_fungible_store to make it more convenient (not having to call 2 functions like here)

movekevin (approved):

aptos-move/move-examples/fungible_asset/preminted_managed_coin/sources/preminted_managed_coin.move
lightmark: sgtm. lemme send a PR for this

larry-aptos (approved):

PR #8228, opened 2023-05-16
[transaction generator] Cleanup ways to create different workloads

igor-aptos:
Have list of default transaction types only in TransactionTypeArg class, and have all others use it directly

Remove stale nft minting as well


crates/transaction-emitter-lib/src/args.rs
```diff
@@ -137,6 +137,12 @@ pub struct EmitArgs {
     )]
     pub transaction_type: Vec<TransactionTypeArg>,
 
+    #[clap(long)]
+    pub module_working_set_size: Option<usize>,
```
vusirikala, line 141: Would be nice to have comments on what the arguments represent.

crates/transaction-generator-lib/src/args.rs
```diff
@@ -15,9 +15,9 @@ pub enum TransactionTypeArg {
     CoinTransferWithInvalid,
     AccountGeneration,
     AccountGenerationLargePool,
-    NftMintAndTransfer,
     PublishPackage,
-    CreateNewAccountResource,
+    AccountResourceSmall,
```
grao1991, line 19: Can we rename to AccountResource32B, AccountResource1K, etc.

Also can we also have 2k, 4k, 5k?

grao1991 (approved):

PR #8233, opened 2023-05-16
[transaction-generator] Rename TransactionExecutor to ReliableTransactionSubmitter

igor-aptos:
cleanup follow-up from this comment - https://github.com/aptos-labs/aptos-core/pull/7651#discussion_r1164756699

### DIFF
--- a/crates/transaction-generator-lib/src/lib.rs
+++ b/crates/transaction-generator-lib/src/lib.rs
@@ -100,7 +100,7 @@ pub struct CounterState {
 }
 
 #[async_trait]
-pub trait TransactionExecutor: Sync + Send {
+pub trait ReliableTransactionSubmitter: Sync + Send {
     async fn get_account_balance(&self, account_address: AccountAddress) -> Result<u64>;
 
     async fn query_sequence_number(&self, account_address: AccountAddress) -> Result<u64>;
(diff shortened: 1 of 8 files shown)


danielxiangzl (approved):

crates/transaction-generator-lib/src/lib.rs
```diff
@@ -100,7 +100,7 @@ pub struct CounterState {
 }
 
 #[async_trait]
-pub trait TransactionExecutor: Sync + Send {
+pub trait ReliableTransactionSubmitter: Sync + Send {
```
sitalkedia, line 103: Not sure if I like this naming - calling this `Reliable` doesn't add much value since we don't have any `Unreliable` alternative. How about just `TransactionSubmitter` ?

grao1991: +1

grao1991 (approved):

crates/transaction-generator-lib/src/lib.rs
igor-aptos: We do have unreliable client - that is raw RestClient, and submit_worker sends transactions being fine with transactions failing. 

This one is adding expectations to the API that it will wait for transactions to be committed successfully,  and fail otherwise.  It does retries,  resubmission of transactions etc. Check RestReliabkeTransactionSubmitter implementation.


This is used exclusively for initialization before transaction generation for the given workload starts.

sitalkedia: May be `RetryingTransactionSubmitter` is a better name in that case?

igor-aptos: I prefer Reliable, as retrying doesn't confer the meaning.

initialization needs to succeed, and so every single transaction that initialization creates needs to be executed successfully

PR #8237, opened 2023-05-16
[Indexer] Token v2 part 2

bowenyang007:
### Description
1. Fix nft transfer (when nft is transferred we need to soft delete previous ownership record)
2. Add token activities
3. Pad some addresses that weren't padded properly in token v2 tables

### Test Plan
image

### DIFF
--- a/crates/indexer/src/models/token_models/v2_token_activities.rs
+++ b/crates/indexer/src/models/token_models/v2_token_activities.rs
@@ -1,0 +1,226 @@
+// Copyright © Aptos Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+// This is required because a diesel macro makes clippy sad
+#![allow(clippy::extra_unused_lifetimes)]
+#![allow(clippy::unused_unit)]
+
+use super::{
+    token_utils::{TokenDataIdType, TokenEvent},
+    v2_token_utils::{TokenStandard, TokenV2AggregatedDataMapping, V2TokenEvent},
+};
+use crate::{schema::token_activities_v2, util::standardize_address};
+use aptos_api_types::Event as APIEvent;
+use bigdecimal::{BigDecimal, One, Zero};
+use field_count::FieldCount;
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Deserialize, FieldCount, Identifiable, Insertable, Serialize)]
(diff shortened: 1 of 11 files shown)


crates/indexer/src/models/token_models/v2_token_activities.rs
```diff
@@ -0,0 +1,226 @@
...
+                        from_address: Some(object_core.get_owner_address()),
+                        to_address: None,
+                        token_amount: BigDecimal::one(),
+                        before_value: None,
+                        after_value: None,
+                    },
+                    V2TokenEvent::TokenMutationEvent(inner) => TokenActivityHelperV2 {
+                        from_address: Some(object_core.get_owner_address()),
+                        to_address: None,
+                        token_amount: BigDecimal::zero(),
```
rtso, line 93: nit: should this be 1? because 1 token is being mutated. I guess a larger question is what is token_amount supposed to represent?

rtso (approved):

crates/indexer/src/models/token_models/v2_token_activities.rs
```diff
@@ -0,0 +1,226 @@
...
+    pub property_version_v1: BigDecimal,
+    pub type_: String,
+    pub from_address: Option<String>,
+    pub to_address: Option<String>,
+    pub token_amount: BigDecimal,
+    pub before_value: Option<String>,
+    pub after_value: Option<String>,
+    pub entry_function_id_str: Option<String>,
+    pub token_standard: String,
+    pub is_fungible_v2: Option<bool>,
```
0xmaayan, line 35: we don't keep `is_soulbound_v2`?

crates/indexer/src/models/token_models/v2_token_activities.rs
bowenyang007: the option here are really 0 or null b/c this is modifying token data level properties (e.g. changing URI) so in the case of fungible tokens we can't really get the right amount if that makes sense.

larry-aptos (approved):

crates/indexer/src/models/token_models/v2_token_activities.rs
bowenyang007: yea exactly. you can think of soulbound as metadata. fungible...i can kinda track it but it really doesn't have to be here either.

PR #8238, opened 2023-05-17
[Network] Fix bug in VFN role and improve role propagation.

JoshLind:
Note: most of this PR is just new tests.

### Description
This PR fixes a bug in the network stack whereby VFNs connecting to validators are assigned the "Unknown" role, instead of the "ValidatorFullnode" role (as they should be). The PR also helps to improve role assignment logic, specifically that peer roles should always be assigned when a noise handshake completes (regardless of inbound or outbound origin). This is not currently the case.

To achieve this, the PR offers the following commits:
1. Update the peer role logic to assign VFNs the "ValidatorFullnode" role when connecting to validators.
2. Update the noise handshake code to assign peer roles during handshake time for outbound connections. This removes a workaround that assigns the role after the handshake is done.
3. Update the stale peer logic to not remove inbound connections with the "ValidatorFullnode" role.

The PR also adds a new set of unit tests to better verify this behaviour (in each commit).

### Test Plan
Existing and new unit tests.

### DIFF
--- a/network/src/noise/handshake.rs
+++ b/network/src/noise/handshake.rs
@@ -375,19 +375,37 @@ impl NoiseUpgrader {
                         Self::authenticate_inbound(remote_peer_short, &peer, &remote_public_key)
                     },
                     None => {
-                        // if not, verify that their peerid is constructed correctly from their public key
+                        // The peer is not in the trusted peer set. Verify that the Peer ID is
+                        // constructed correctly from the public key.
                         let derived_remote_peer_id =
                             aptos_types::account_address::from_identity_public_key(
                                 remote_public_key,
                             );
                         if derived_remote_peer_id != remote_peer_id {
+                            // The peer ID is not constructed correctly from the public key
                             Err(NoiseHandshakeError::ClientPeerIdMismatch(
                                 remote_peer_short,
                                 remote_peer_id,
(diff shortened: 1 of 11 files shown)


network/src/noise/handshake.rs
```diff
@@ -726,15 +788,257 @@ mod test {
...
+        client.network_context = client_network_context;
+
+        // Update the server network context
+        let server_peer_id = server.network_context.peer_id();
+        let server_network_context =
+            NetworkContext::new(RoleType::FullNode, NetworkId::Public, server_peer_id);
+        server.network_context = server_network_context;
+
+        // Add the server VFN to the trusted peers set
+        let trusted_peers = peers_and_metadata
```
bchocho, line 1005: I see that in this test we only add the server VFN to the trusted_peers.

In a "normal" network, do we expect both to be in trusted_peers, so they both identify each other as vfns?

JoshLind: Aah, good point. I just double checked this and in a "normal" network, both are in the trusted set, so I've updated the test to ensure they both connect as VFNs 😄

bchocho (approved):
LGTM

brianolson (approved):

PR #8241, opened 2023-05-17
Tune backpressure to kick-in later

igor-aptos:
Increase all threshold values by 1s

### DIFF
--- a/config/src/config/consensus_config.rs
+++ b/config/src/config/consensus_config.rs
@@ -132,31 +132,31 @@ impl Default for ConsensusConfig {
             vote_back_pressure_limit: 30,
             pipeline_backpressure: vec![
                 PipelineBackpressureValues {
-                    back_pressure_pipeline_latency_limit_ms: 1000,
+                    back_pressure_pipeline_latency_limit_ms: 2000,
                     max_sending_block_txns_override: 10000,
                     max_sending_block_bytes_override: 5 * 1024 * 1024,
                     backpressure_proposal_delay_ms: 100,
                 },
                 PipelineBackpressureValues {
-                    back_pressure_pipeline_latency_limit_ms: 1500,
+                    back_pressure_pipeline_latency_limit_ms: 2500,
                     max_sending_block_txns_override: 10000,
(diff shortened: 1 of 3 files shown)


config/src/config/consensus_config.rs
```diff
@@ -132,31 +132,31 @@ impl Default for ConsensusConfig {
...
                 PipelineBackpressureValues {
-                    back_pressure_pipeline_latency_limit_ms: 2000,
+                    back_pressure_pipeline_latency_limit_ms: 3000,
                     max_sending_block_txns_override: 10000,
                     max_sending_block_bytes_override: 5 * 1024 * 1024,
                     backpressure_proposal_delay_ms: 300,
                 },
                 PipelineBackpressureValues {
-                    back_pressure_pipeline_latency_limit_ms: 2500,
+                    back_pressure_pipeline_latency_limit_ms: 3500,
```
bchocho, line 153: We expect 300ms * 2 increase in latency compared to the previous (@ 3000 ms) so it seems like we expect to cross over to this value unless there is a sudden drop in the latency. Is this what we want? Or should this threshold be something > 3000 + 300*2 ?

igor-aptos: this is not additional 300ms, this is if we cross 3s with delay included. 

so crossing 3000 adds additional 200ms. so there is additional 300ms of gap before the next level - things are not escalating automatically (it is just that gaps are between levels are 300ms not 500ms as the values would suggest.

bchocho: Ah, I see the increase will be the delta * 2.

bchocho (approved):
Makes sense to me.

bchocho:
As follow-up though let's make sure the backpressure still kicks in for some/all of the forge tests

sitalkedia (changes requested):
This should not be needed if we fix the pipeline latency calculation logic as discussed in the SEV thread.

igor-aptos:
@sitalkedia , as discussed,  we will use this as a better approach here

sherry-x (approved):

sitalkedia (approved):

PR #8244, opened 2023-05-17
[helm][genesis] Disable aptos CLI telemetry 

ibalajiarun:
### Description

This PR fixes an issue where the genesis operation in Forge can take arbitrarily long time because the CLI is slow, which sometimes leads to genesis job timing out and forge failing. 

Turns out, the CLI is slow because it sends metrics to Google Analytics and can sometimes have to retry multiple times after backoff. This change essentially disables sending telemetry to Google Analytics.

### Test Plan

Forge test


rustielin (approved):
Great find!

sionescu (approved):

PR #8245, opened 2023-05-17
more shards for rocksdb table cache to speedup deep DB reads

zjma:


zjma:
image



image

image

zjma:
Above are some additional executor-benchmark results showing that it reduces DB read latencies.

PR #8246, opened 2023-05-17
Change consensus pipeline backpressure to exclude consensus ordering latency

igor-aptos:
Current pipeline backpressure constants need to be tuned for different networks, based on average network latency (i.e. a test network with minimal latencies needs to have a different thresholds compared to mainnet with real-world latencies)

With this change, we exclude block_creation -> block_ordered part of the latency, and only use block_ordered -> block_committed as a pipeline latency, reducing the need for tuning backpressure constants across different networks.

### DIFF
--- a/consensus/src/block_storage/block_store.rs
+++ b/consensus/src/block_storage/block_store.rs
@@ -42,6 +42,8 @@ mod block_store_test;
 #[path = "sync_manager.rs"]
 pub mod sync_manager;
 
+const MAX_ORDERING_PIPELINE_LATENCY_REDUCTION: Duration = Duration::from_secs(1);
+
 fn update_counters_for_ordered_blocks(ordered_blocks: &[Arc<ExecutedBlock>]) {
     for block in ordered_blocks {
         observe_block(block.block().timestamp_usecs(), BlockStage::ORDERED);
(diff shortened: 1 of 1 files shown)


consensus/src/block_storage/block_store.rs
```diff
@@ -556,7 +556,7 @@ impl BlockStore {
         counters::CONSENSUS_PROPOSAL_PENDING_ROUNDS.observe(pending_rounds as f64);
         counters::CONSENSUS_PROPOSAL_PENDING_DURATION.observe(latency_to_committed.as_secs_f64());
 
-        latency_to_committed
+        latency_to_committed.saturating_sub(latency_to_ordered_ms)
```
sitalkedia, line 559: LGTM, can you add some comments in the function explaining what latency is it calculating and justification behind that? Also, should we change the name of the function to `pipeline_ordered_to_committed_latency` to make it clearer?

sitalkedia (approved):

sitalkedia:
Also, let's make sure to test this out e2e (in mainnet like Forge env) before rolling in mainnet again :)

JoshLind (approved):
Unblocking.

PR #8247, opened 2023-05-17
[Execution/Storage][Sharding] Shard state updates in StateDelta.

grao1991:
### Description
Avoid the flatten in the save_transaction_block, and move it to the async state commit stage in the background (will also be removed later when the JMT calculation part is sharded). By removing it from foreground thread, it gives us 5%+ performance gain for commit (measured by account creation benchmark from 0 to 20M accounts).

### DIFF
--- a/execution/executor-types/src/in_memory_state_calculator.rs
+++ b/execution/executor-types/src/in_memory_state_calculator.rs
@@ -14,12 +14,14 @@ use aptos_types::{
     event::EventKey,
     on_chain_config,
     state_store::{
-        state_key::StateKey, state_storage_usage::StateStorageUsage, state_value::StateValue,
+        create_empty_sharded_state_updates, state_key::StateKey,
+        state_storage_usage::StateStorageUsage, state_value::StateValue, ShardedStateUpdates,
     },
     transaction::{Transaction, Version},
     write_set::{TransactionWrite, WriteOp, WriteSet},
 };
 use dashmap::DashMap;
+use itertools::zip_eq;
 use once_cell::sync::Lazy;
 use std::collections::HashMap;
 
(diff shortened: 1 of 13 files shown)


execution/executor-types/src/in_memory_state_calculator.rs
```diff
@@ -245,7 +256,10 @@ impl InMemoryStateCalculator {
                     &mut self.usage,
                     (*write_set).clone(),
                 )?;
-                self.updates_after_latest.extend(state_updates.into_iter());
+                state_updates.iter().for_each(|(k, v)| {
```
areshand, line 259: nit: consolidate here with line 278 - 280 using a util function? would be easier to read as well.

areshand (approved):

msmouse (approved):
please address comments

execution/executor-types/src/in_memory_state_calculator.rs
grao1991: Done

storage/aptosdb/src/lib.rs
```diff
@@ -2002,19 +2007,13 @@ impl DbWriter for AptosDB {
...
-                    let _timer = OTHER_TIMERS_SECONDS
-                        .with_label_values(&["flatten_states"])
-                        .start_timer();
-                    Some(block_state_updates.into_iter().flatten().collect())
-                };
                 if !txns_to_commit.is_empty() {
                     let _timer = OTHER_TIMERS_SECONDS
                         .with_label_values(&["buffered_state___update"])
                         .start_timer();
                     // TODO(grao): Make BufferedState take sharded updates.
```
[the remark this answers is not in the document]
grao1991: Done

PR #8250, opened 2023-05-17
[indexer-grpc] integration tests with docker-compose

rustielin:
## Description

Set up e2e testing with cargo test and existing docker-compose setup. A python script is written to wrap the 2 docker-composes required (validator-testnet, indexer-grpc) for the full setup, as well as some healthchecks and convenience methods for start/stop/wipe cluster.

```
$ ./testsuite/indexer_grpc_local.py --help                           
usage: Indexer GRPC Local [-h] [--verbose] {start,stop,wipe} ...

Spins up an indexer GRPC locally using a single validator testnet

positional arguments:
  {start,stop,wipe}
    start            Start the indexer GRPC setup
    stop             Stop the indexer GRPC setup
    wipe             Completely wipe the storage

options:
  -h, --help         show this help message and exit
```

NOTE: It bears a close resemblance to the faucet integration tests, which use the same method of spinning up docker containers using some python scripts on the side, and then invoking `cargo test --features integration-tests`. We can unify both these in the future, but they vary in some slight ways (e.g. using the `aptos` CLI rather than `docker-compose` directly)

### New tests

New crate `aptos-indexer-grpc-integration-tests` contains integration tests that rely on docker-compose on the same host. It's hidden behind the `integration-tests` feature flag. To run it locally:

```
# start the docker compose
./testsuite/indexer_grpc_local.py --verbose start --no-indexer-grpc

# run the tests
cargo nextest run --features integration-tests --package aptos-indexer-grpc-integration-tests
```

* `verify_docker_compose_setup` -- just sanity check that we can connect to docker compose on the host. healthcheck exists separately in the docker-compose start up
* `test_cold_start_cache_worker_progress` -- just cold start the cache worker by itself and check liveness
* `test_cold_start_file_store_worker_progress` -- cold start both the cache worker and filestore worker. Verify that the file store metadata is constantly updated while emitting traffic 

### Choice of test setup
The choice of docker-compose + cargo test is common in our codebase. It spins up some critical dependencies on the side on the same host, and invokes part of the codebase directly in a test binary (e.g. via `cargo test` or `nextest`). There are some other ways to spin up and manage the deps, but I ultimately chose not to use them. As a summary:

* [testcontainers](https://docs.rs/testcontainers/latest/testcontainers/) -- does not support arbitrary command args, which we use extensively elsewhere
* [dagger](https://dagger.io) -- could work in conjunction with docker-compose, since you can import containers by ID (didn't work out of the box though). If all we want to do is interact with some containers that are set up elsewhere though, we can use the docker api directly
* [kurtosis](https://www.kurtosis.com/) -- should work, need more investigation. We can plug it in to https://github.com/kurtosis-tech/aptos-package/tree/main. One downside though is that we'd have to create a parallel deployment implementation in kurtosis vs docker-compose. Also seems more of a fit for full e2e test suites, rather than what we're looking for here, which is some light tool to spin up deps (and the deps don't really require anything too complicated)

Misc:
* Change some `info` logs to `debug` to improve log signal
* FileStoreWorker writes to metadata if it does not exist already. This makes it such that we don't have to crash-loop the CacheWorker until the minimum (1000 versions) are processed. This is apparent in test environments with not enough traffic.
* Use `redis` docker image based on system architecture. `redis` vs `arm64v8/redis` works well on M1 Mac via local testing

## Test Plan

Existing unit tests and new integration tests
* [Indexer gRPC Integration Tests/ run-tests-local-testnet](https://github.com/aptos-labs/aptos-core/actions/runs/5084580577/jobs/9137055361?pr=8250)

Updated e2e docker compose test
* [Docker Indexer gRPC test / test-indexer-grpc-docker-compose](https://github.com/aptos-labs/aptos-core/actions/runs/5084580579/jobs/9137054886?pr=8250)


testsuite/indexer_grpc_local.py
```diff
@@ -0,0 +1,290 @@
...
+class StartSubcommand(Enum):
+    NO_INDEXER_GRPC = "no-indexer-grpc"
+
+
+def normalize_var_name(var_name: str) -> str:
+    return var_name.replace("-", "_").upper()
+
+
+# set envs based on platform, if it's not already overriden
+if not os.environ.get("REDIS_IMAGE_REPO"):
```
perryjrandall, line 61: Can we do this in the main function instead this happens on module import

testsuite/indexer_grpc_local.py
```diff
@@ -0,0 +1,290 @@
...
+    stop_indexer_grpc(context.shell)
+    stop_single_validator_testnet(context.shell)
+
+
+def wipe(context: SystemContext):
+    stop(context)  # call stop() just for sanity
+    context.shell.run(["docker", "volume", "rm"] + SHARED_DOCKER_VOLUME_NAMES)
+
+
+def main():
```
perryjrandall, line 244: Can you add return type annotations?

crates/transaction-emitter-lib/src/wrappers.rs
```diff
@@ -18,6 +18,7 @@ pub async fn emit_transactions(
     cluster_args: &ClusterArgs,
     emit_args: &EmitArgs,
 ) -> Result<TxnStats> {
+    println!("Emitting transactions...");
```
perryjrandall, line 21: Should this be using info! logger?

rustielin: oof I forgot to remove that :P

ecosystem/indexer-grpc/indexer-grpc-integration-tests/src/tests/fullnode_tests.rs
```diff
@@ -0,0 +1,360 @@
...
+        match reqwest::get(format!("http://localhost:{}/metrics", health_check_port)).await {
+            Ok(_) => break,
+            Err(e) => {
+                if i == startup_timeout_secs - 1 {
+                    let msg = if join_handle.is_finished() {
+                        format!("Server failed on startup: {:#?}", join_handle.await)
+                    } else {
+                        "Server was still starting up".to_string()
+                    };
+                    println!("waiting...");
```
perryjrandall, line 174: do we wanna use logger here?

ecosystem/indexer-grpc/indexer-grpc-integration-tests/src/tests/mod.rs
```diff
@@ -0,0 +1,10 @@
+// Copyright © Aptos Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+#[test]
+fn sanity_check() {
+    assert_eq!(2 + 2, 4);
```
perryjrandall, line 6: we probably dont need this

larry-aptos (approved):

larry-aptos (approved):

ecosystem/indexer-grpc/indexer-grpc-integration-tests/src/tests/mod.rs
rustielin: lol yea forgot to delete :p

ecosystem/indexer-grpc/indexer-grpc-cache-worker/src/worker.rs
```diff
@@ -294,7 +282,7 @@ async fn process_streaming_response(
                     PROCESSED_VERSIONS_COUNT.inc_by(num_of_transactions);
                     LATEST_PROCESSED_VERSION.set(current_version as i64);
                     PROCESSED_BATCH_SIZE.set(num_of_transactions as i64);
-                    info!(
+                    debug!(
```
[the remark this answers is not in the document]
rustielin: actually, I'll revert these changes to the log levels. We should probably re-evaluate some of these though, but not in this PR

.github/workflows/docker-indexer-grpc-test.yaml
```diff
@@ -20,7 +20,7 @@ jobs:
       VALIDATOR_IMAGE_REPO: ${{ secrets.GCP_DOCKER_ARTIFACT_REPO }}/validator
       FAUCET_IMAGE_REPO: ${{ secrets.GCP_DOCKER_ARTIFACT_REPO }}/faucet
       INDEXER_GRPC_IMAGE_REPO: ${{ secrets.GCP_DOCKER_ARTIFACT_REPO }}/indexer-grpc
-      IMAGE_TAG: ${{ inputs.GIT_SHA || 'f4100b21da4e9ba10fadd184e92e3d1c22bc282e' }} # hardcode to a known good build when not running on workflow_call
+      IMAGE_TAG: ${{ inputs.GIT_SHA || '3e29accd9d9aa2ca4802bcfa35847d13c3af7ccd' }} # hardcode to a known good build when not running on workflow_call
```
perryjrandall, line 23: Lets make this a GH secret at least so that we can change without a code change

.github/workflows/docker-indexer-grpc-test.yaml
```diff
@@ -36,23 +36,17 @@ jobs:
           AWS_DOCKER_ARTIFACT_REPO: ${{ secrets.AWS_DOCKER_ARTIFACT_REPO }}
           GIT_CREDENTIALS: ${{ secrets.GIT_CREDENTIALS }}
 
-      - name: Run validator-testnet docker compose
-        run: docker-compose up -d
+      - name: Install grpcurl
```
perryjrandall, line 39: This should probably be a composable action which we can share

crates/transaction-emitter-lib/src/wrappers.rs
```diff
@@ -173,6 +173,8 @@ pub async fn emit_transactions_with_cluster(
     if !cluster.coin_source_is_root {
         emit_job_request = emit_job_request.prompt_before_spending();
     }
+
+    println!("Start emitting....");
```
perryjrandall, line 177: debugging? use logger?

docker/compose/indexer-grpc/cache-worker-config.yaml
```diff
@@ -5,4 +5,4 @@ server_config:
   file_store_config:
     file_store_type: LocalFileStore
     local_file_store_path: /opt/aptos/file-store
-  redis_main_instance_address: 172.16.1.12:6379
+  redis_main_instance_address: 172.16.1.12:6379 # use the primary
```
perryjrandall, line 8: should these be hardcoded? Should we be reading these from some place?

testsuite/test_framework/reqwest.py
```diff
@@ -0,0 +1,19 @@
...
+from requests import Response
+import logging
+
+
+class HttpClient:
+    def get(self, url: str, headers: dict = {}) -> requests.Response:
+        raise NotImplementedError()
+
+
+class SimpleHttpClient(HttpClient):
```
perryjrandall, line 15: I like this interface, look forward to a mock one!

testsuite/indexer_grpc_local.py
```diff
@@ -0,0 +1,289 @@
+#!/usr/bin/env python3
```
perryjrandall, line 1: Is this file using strict typing?

perryjrandall (approved):
Amazing <3

Left some comments, most in my mind making sure python is adhering to strict typing / lint

.github/workflows/docker-indexer-grpc-test.yaml
rustielin: Actually I'll just change it to `devnet` tag, which should be up to date enough

.github/workflows/docker-indexer-grpc-test.yaml
rustielin: good call

docker/compose/indexer-grpc/cache-worker-config.yaml
rustielin: they're hard coded from the base docker-compose, where we're using static IPs

testsuite/indexer_grpc_local.py
rustielin: nope. I'm planning on doing that for all scripts in `testsuite` soon...

PR #8251, opened 2023-05-17
[forge] Automatically generate a useful namespace

perryjrandall:
Make --namespace optional and give the default a somewhat easy to remember name

Test Plan: unittests

### DIFF
--- a/testsuite/forge-cli/src/main.rs
+++ b/testsuite/forge-cli/src/main.rs
@@ -40,6 +40,7 @@ use aptos_testcases::{
     validator_reboot_stress_test::ValidatorRebootStressTest,
     CompositeNetworkTest,
 };
+use rand::{rngs::ThreadRng, seq::SliceRandom, Rng};
 use std::{
     env,
     num::NonZeroUsize,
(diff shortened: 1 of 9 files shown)


testsuite/forge-cli/src/main.rs
```diff
@@ -259,11 +274,24 @@ fn main() -> Result<()> {
                     if let Some(move_modules_dir) = &k8s.move_modules_dir {
                         test_suite = test_suite.with_genesis_modules_path(move_modules_dir.clone());
                     }
+                    let namespace = if k8s.namespace.is_none() {
+                        let mut rng : ThreadRng= rand::thread_rng();
+                        let words = std::fs::File::open("/usr/share/dict/words")
```
sionescu, line 279: It probably makes sense to hard-code a sanitized list of words, and append a short (6-8 chars) random string. The `words` dict has plenty of NSFW words.

perryjrandall: I personally dont see a huge problem with that, however I can pick words of a certain length

testsuite/forge-cli/src/main.rs
```diff
@@ -866,7 +895,7 @@ fn three_region_sim_graceful_overload(config: ForgeConfig) -> ForgeConfig {
                 // Additionally - we are not really gracefully handling overlaods,
                 // setting limits based on current reality, to make sure they
                 // don't regress, but something to investigate
-                avg_tps: 1200,
+                avg_tps: 1400,
```
rustielin, line 898: accident?

rustielin (approved):
LGTM! Thanks for making forge better tested and documented :) 

One small comment on (accidental?) bumping of success criteria

testsuite/forge-cli/src/main.rs
perryjrandall: How the fuck...? Yes accident

testsuite/forge-cli/src/main.rs
geekflyer: does this work across OS?

geekflyer (approved):
approving for now, but I must wonder - maybe it's better to use a library for the random name generation (to handle OS differences and whatnot)?

perryjrandall:
> approving for now, but I must wonder - maybe it's better to use a library for the random name generation (to handle OS differences and whatnot)?

So I checked out random_words and actually that has a built in set of words and some nice convenience functions, thanks for the suggestion!

PR #8252, opened 2023-05-17
[scripts] make dev_setup.sh more resilient

rustielin:
### Description

Try to repro some network issues 

There's some nasty redirects in our .NET installation it turns out. Using `wget` instead of `curl` seems to fix it for the most part.

Sometimes it redirects back to https://dot.net/v1/dotnet-install.sh (which then errors out), and sometimes to https://dotnet.microsoft.com/download/dotnet/scripts/v1/dotnet-install.sh (the correct one)

Also set `-x` in `dev_setup.sh` when running in CI only. Don't set it when running on laptops since it will become too verbose.

### Test Plan

Try `wget` and `curl` the URL a bunch of times. I'm able to repro the main issue like so:

```
for i in `seq 0 100`; do curl -Ls -o /dev/null -w %{url_effective} https://dot.net/v1/dotnet-install.sh && echo ; done
```

all the CI that relies on dev_setup.sh working on this PR, despite a spike in errors across other PRs

### DIFF
--- a/scripts/dev_setup.sh
+++ b/scripts/dev_setup.sh
@@ -14,7 +14,7 @@
 # 3 ${HOME}/bin/, or ${INSTALL_DIR} is expected to be on the path - hashicorp tools/etc.  will be installed there on linux systems.
 
 # fast fail.
-set -eo pipefail
+set -eox pipefail
 
 SHELLCHECK_VERSION=0.7.1
 GRCOV_VERSION=0.8.2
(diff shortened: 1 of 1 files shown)


sionescu (approved):

0xmaayan (approved):
Thanks!

perryjrandall:
We can completely reduce this error by building this as an image layer then using that in subsequent github action workflows

PR #8253, opened 2023-05-17
[TF] More fixes for GCP DNS

sionescu:
### Description

* add variable to control the record TTL and lower the default from 1h to 5min, as the rest of the records on GCP
* add variable *create_dns_records* to the testnet module as well, to be useable from the devnet module

### Test Plan

Tested on the new GCP Devnet.

### DIFF
--- a/terraform/helm/testnet-addons/templates/ingress.yaml
+++ b/terraform/helm/testnet-addons/templates/ingress.yaml
@@ -5,14 +5,16 @@ metadata:
   labels:
     {{- include "testnet-addons.labels" . | nindent 4 }}
   annotations:
+    {{- if .Values.service.domain }}
+    external-dns.alpha.kubernetes.io/hostname: {{ .Values.service.domain }}
+    {{- end }}
+    # EKS annotations
+    {{- if eq .Values.cloud "EKS" }}
     kubernetes.io/ingress.class: alb
     alb.ingress.kubernetes.io/scheme: internet-facing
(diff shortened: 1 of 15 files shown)


rustielin (approved):
do you also still need this https://github.com/aptos-labs/aptos-core/pull/8160 in conjunction with this PR?

rustielin:
@sionescu I think it makes sense in a world where we don't care about DNS but still want a testnet or single node. It's mostly for test cases, and also brings it up to parity to the AWS setup

sionescu:
@rustielin Yes, in that case let's merge #8160.

sherry-x (approved):

terraform/helm/testnet-addons/templates/ingress.yaml
```diff
@@ -5,15 +5,17 @@ metadata:
   labels:
     {{- include "testnet-addons.labels" . | nindent 4 }}
   annotations:
+    {{- if .Values.service.domain }}
+    external-dns.alpha.kubernetes.io/hostname: {{ .Values.service.domain }}
+    {{- end }}
+    # EKS annotations
+    {{- if eq .Values.cloud "EKS" }}
```
rustielin, line 12: nit: `aws` and `gcp` rather than `EKS` and `GKE`?

rustielin (changes requested):

rustielin (approved):

PR #8258, opened 2023-05-17
Add MakeOrChangeTable transaction to transaction generator

vusirikala:
### Description

Add an `make_or_change_table` transaction in transaction-generator that generates modifies the give number of table entries. This transaction is added to test if the state sync works fine when the output of a transaction is huge.


testsuite/module-publish/src/packages/simple/sources/Simple.move
```diff
@@ -411,6 +412,25 @@ module 0xABCD::simple {
         }
     }
 
+    struct TableStore has key {
+        table_entries: Table<u64, u64>
+    }
+
+    fun make_or_change_table(owner: &signer, count: u64) acquires TableStore {
```
igor-aptos, line 419: this always changes the same set of elements. 
maybe more flexible is to also pass offset here, and then allow changing different sub-interval of the table if so desired.

You can pass 0 for offset in your MakeOrChange entry point, but someone can add another more flexible entry point if they want to (i.e. we could have a MakeOrChangeRandom(max_count, max_offset), which randomly select count of keys and offset based on those max values.

igor-aptos (approved):
one minor nit inline, the rest looks good

testsuite/module-publish/src/packages/simple/sources/Simple.move
vusirikala: Done.

.github/workflows/docker-build-test.yaml
```diff
@@ -92,13 +92,13 @@ jobs:
...
   determine-docker-build-metadata:
+    needs: [permission-check]
     runs-on: ubuntu-latest
     steps:
       - name: collect metadata
         run: |
-          echo "GIT_SHA: ${{ env.GIT_SHA }}"
-          echo "TARGET_CACHE_ID: ${{ env.TARGET_CACHE_ID }}"
-          echo "CONCURRENCY: ${{ github.workflow }}-${{ github.event_name }}-${{ (github.event_name == 'push' || github.event_name == 'workflow_dispatch') && github.sha || github.head_ref || github.ref }}"
+          echo "GIT_SHA: ${GIT_SHA}"
```
JoshLind, line 100: Uhh.. silly question: why do we need these changes?

crates/transaction-generator-lib/src/publishing/module_simple.rs
```diff
@@ -292,6 +304,32 @@ impl EntryPoints {
...
             },
+            EntryPoints::MakeOrChangeTable { offset, count } => get_payload(
+                module_id,
+                ident_str!("make_or_change_table").to_owned(),
+                vec![
+                    bcs::to_bytes(offset).unwrap(),
+                    bcs::to_bytes(count).unwrap(),
+                ],
+            ),
+            EntryPoints::MakeOrChangeTableRandom {
```
JoshLind, line 315: Silly question: instead of duplicating `MakeOrChangeTable`, would having a `bool` inside of the `MakeOrChangeTable` enum, e.g., `random`, enough to determine which mode to use?

JoshLind (approved):
LGTM. Just a question around why we need to change the docker files?

.github/workflows/docker-build-test.yaml
vusirikala: That's actually not my change. Yesterday, there was a security fix added by Stelian on all branches. This is part of that.

crates/transaction-generator-lib/src/publishing/module_simple.rs
vusirikala: Should I combine MakeOrChangeTable and MakeOrChangeTableRandom into a single transaction as per your comment? The semantics of offset and count is different in both the transactions though.

PR #8259, opened 2023-05-17
[gha] ability to call run-fullnode-sync from workflow_dispatch

rustielin:
### Description

Adds the `workflow_dispatch` trigger to the reusable workflow `Run Fullnode Sync` (`run-fullnode-sync.yaml`). This lets someone go to the UI to trigger it, with all the options like branch (or git ref), branch, bootstrapping/continuous-syncing mode, etc. like how each of the `fullnode-execute-{network}-{branch}.yaml` jobs call.

### Test Plan

`gh` cli. It can also equivalently be run from the GHA UI

```
# because it's not landed in main branch yet, have to specify --ref <this_branch>
gh workflow run run-fullnode-sync.yaml --ref rustielin/fullnode-sync-gh -F=TEST_NAME=rustie-test -F=BRANCH=5d067dbbd9468397a7618b040a4306ee3a36c6d1 -F=NETWORK=devnet -F=BOOTSTRAPPING_MODE=ExecuteTransactionsFromGenesis -F=CONTINUOUS_SYNCING_MODE=ExecuteTransactions
```

it makes a run like this: https://github.com/aptos-labs/aptos-core/actions/runs/5008782115/jobs/8977013420

### DIFF
--- a/.github/workflows/run-fullnode-sync.yaml
+++ b/.github/workflows/run-fullnode-sync.yaml
@@ -11,8 +11,8 @@ on:
         description: "The unique name of the fullnode test."
         type: string
         required: true
-      BRANCH:
-        description: "The aptos-core branch (or ref) to switch to before running the fullnode."
+      GIT_REF:
+        description: "The aptos-core git ref (GIT_REF or commit hash) to switch to before running the fullnode."
         type: string
         required: true
       NETWORK:
@@ -38,8 +38,8 @@ on:
         description: "The unique name of the fullnode test."
         type: string
         required: true
-      BRANCH:
-        description: "The aptos-core branch (or ref) to switch to before running the fullnode."
+      GIT_REF:
+        description: "The aptos-core GIT_REF (or ref) to switch to before running the fullnode."
         type: string
         required: true
       NETWORK:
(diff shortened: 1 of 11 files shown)


.github/workflows/run-fullnode-sync.yaml
```diff
@@ -32,17 +32,39 @@ on:
...
+      CONTINUOUS_SYNCING_MODE:
+        description: "The state sync continuous syncing mode for the fullnode."
+        type: string
+        required: true
 
 jobs:
   fullnode-sync:
     runs-on: medium-perf-docker-with-local-ssd
-    timeout-minutes: ${{ inputs.TIMEOUT_MINUTES }}
+    timeout-minutes: ${{ inputs.TIMEOUT_MINUTES || 300 }}
```
sionescu, line 61: How about turning this into an input on `workflow_dispatch` too ?

sionescu (approved):

msmouse (approved):
Thanks boss! 
Would be nicer if we can put in a commit hash, but this isn't too bad at all.

.github/workflows/run-fullnode-sync.yaml
rustielin: There were some issues with the type conversion between string and number actually. Need to dig more into what GH wants in each scenario lol

rustielin:
@msmouse actually you can put in a commit hash (see my test plan), but for now the input argument is still called `BRANCH`. Lemme rename it to `GIT_REF` actually, which would be more accurate and less confusing....

sionescu:
@msmouse The name `BRANCH` is a bit of a misnomer, you can use any ref.

PR #8265, opened 2023-05-18
reduce aptos-api error size

simonjiao:
### Description

- Boxing api errors to reduce error size

### Test Plan
cargo build && cargo test


banool (changes requested):
Hey thanks for this PR! 

The reason we pin the version is because future Poem versions change the content type string (from `application/json` to `application/json; charset: utf-8` and I don't want to invest the effort to figure out if this is a breaking change right now. Please find a way to resolve the compilation errors without relaxing the Poem version requirements if at all possible.

simonjiao:
> Hey thanks for this PR!
> 
> The reason we pin the version is because future Poem versions change the content type string (from `application/json` to `application/json; charset: utf-8` and I don't want to invest the effort to figure out if this is a breaking change right now. Please find a way to resolve the compilation errors without relaxing the Poem version requirements if at all possible.

sure. Will remove that changes.

banool (approved):
Thanks! Looks good to me now, let's see how it goes in CI.

davidiw (approved):

PR #8266, opened 2023-05-18
[aggregator][vm] Avoid materializing deltas before running epilogue

vusirikala:
### Description

This PR avoids materializing deltas before running epilogue. Instead, DeltaStateView stores both writes and deltas. DeltaStateView materializes a delta into a write only if the aggregator is read in epilogue.


aptos-move/aptos-aggregator/src/delta_change_set.rs
```diff
@@ -295,6 +295,23 @@ impl DeltaChangeSet {
         }
     }
 
+    pub fn contains_key(&self, key: &StateKey) -> bool {
+        self.delta_change_set.contains_key(key)
+    }
+
+    pub fn get_write_op(
```
georgemitenkov, line 302: I would argue we do not need this with extra expects, etc. Why not just get the delta?

```
// DeltaStateView
match self.delta_set.get(state_key) {
   Some(...) => ...
   None => match self.write_set.get(state_key) {
     // Old logic.
   }
}
```

See other comment

aptos-move/aptos-vm/src/delta_state_view.rs
```diff
@@ -2,22 +2,23 @@
...
         state_key::StateKey, state_storage_usage::StateStorageUsage, state_value::StateValue,
     },
-    write_set::{TransactionWrite, WriteSet},
+    write_set::TransactionWrite,
 };
 
 pub struct DeltaStateView<'a, 'b, S> {
     base: &'a S,
-    write_set: &'b WriteSet,
+    change_set: &'b ChangeSetExt,
```
georgemitenkov, line 16: FYI, `ChangeSetExt` will be gone soon (If I get some reviews🙃) see here: https://github.com/aptos-labs/aptos-core/pull/8195.

aptos-move/aptos-aggregator/src/delta_change_set.rs
vusirikala: Done.

aptos-move/aptos-vm/src/delta_state_view.rs
vusirikala: Currently, `success_transaction_cleanup` still takes `ChangeSetExt` as input. Shall we wait to land this PR until your PR is landed?

georgemitenkov: No no, land this first, it's just that you are aware that in future we will remove `TransactionOutputExt` and `ChangeSetExt` :)

georgemitenkov:
@vusirikala this looks good! Maybe we can add tests that `DeltaStateView` resolves data correctly? Or maybe we land as is, let's see what @gelash @zekun000 think

msmouse (approved):

aptos-move/aptos-vm/src/aptos_vm.rs
```diff
@@ -319,8 +319,7 @@ impl AptosVM {
         log_context: &AdapterLogSchema,
         change_set_configs: &ChangeSetConfigs,
     ) -> Result<(VMStatus, TransactionOutputExt), VMStatus> {
-        let storage_with_changes =
-            DeltaStateView::new(resolver, user_txn_change_set_ext.write_set());
+        let storage_with_changes = DeltaStateView::new(resolver, &user_txn_change_set_ext);
         // TODO: at this point we know that delta application failed
```
gelash, line 323: This comment seems very weird and outdated, we should probably fix it. @georgemitenkov can you look as well?
Do we know that the delta application failed?
Should we remove the part where it says the only use-case is for gas fees? (soon to not be the case)

aptos-move/aptos-vm/src/aptos_vm.rs
```diff
@@ -329,12 +328,7 @@ impl AptosVM {
         // Also, it is worth mentioning that current VM error handling is
         // rather ugly and has a lot of legacy code. This makes proper error
         // handling quite challenging.
-        let delta_write_set = user_txn_change_set_ext
-            .delta_change_set()
-            .clone()
-            .try_into_write_set(resolver)?;
-        let delta_state_view = DeltaStateView::new(&storage_with_changes, &delta_write_set);
-        let resolver = delta_state_view.as_move_resolver();
+        let resolver = storage_with_changes.as_move_resolver();
```
gelash, line 331: could combine the definition of storage_with_changes in the same line, I think that makes sense.

aptos-move/aptos-vm/src/aptos_vm.rs
```diff
@@ -678,14 +672,8 @@ impl AptosVM {
...
-        let storage_with_changes =
-            DeltaStateView::new(resolver, inner_function_change_set_ext.write_set());
-        let delta_write_set = inner_function_change_set_ext
-            .delta_change_set()
-            .clone()
-            .try_into_write_set(resolver)?;
-        let delta_state_view = DeltaStateView::new(&storage_with_changes, &delta_write_set);
-        let resolver = delta_state_view.as_move_resolver();
+        let storage_with_changes = DeltaStateView::new(resolver, &inner_function_change_set_ext);
+        let resolver = storage_with_changes.as_move_resolver();
```
gelash, line 676: same - could combine in one line if we never use 'storage_with_changes'

aptos-move/aptos-vm/src/delta_state_view.rs
```diff
@@ -32,9 +33,14 @@ where
     }
 
     fn get_state_value(&self, state_key: &StateKey) -> Result<Option<StateValue>> {
-        match self.write_set.get(state_key) {
-            Some(write_op) => Ok(write_op.as_state_value()),
-            None => self.base.get_state_value(state_key),
+        match self.change_set.delta_change_set().get(state_key) {
```
gelash, line 36: One question here is whether we want to adjust the write_set and delta_set when we resolve something from the delta_set. Maybe it's not that relevant? @georgemitenkov

gelash (commented):
Minor comments - let's have a look what makes sense to address @vusirikala, @georgemitenkov, and 🚢

gelash (commented):
Minor comments - let's have a look what makes sense to address @vusirikala, @georgemitenkov, and 🚢

aptos-move/aptos-vm/src/aptos_vm.rs
vusirikala: Done.

aptos-move/aptos-vm/src/aptos_vm.rs
vusirikala: When I combine the two lines into one, it gives me "creates a temporary which is freed while still in use" error. So, I left it like this.

aptos-move/aptos-vm/src/aptos_vm.rs
georgemitenkov: Yes its super-outdated. @vusirikala let's get rid of this comment.

aptos-move/aptos-vm/src/delta_state_view.rs
georgemitenkov: Do you mean re-cache materialized delta as a write op?

aptos-move/aptos-vm/src/delta_state_view.rs
```diff
@@ -47,3 +53,80 @@ where
...
+    use aptos_types::{
+        state_store::{state_key::StateKey, state_value::StateValue, table::TableHandle},
+        transaction::{ChangeSet, NoOpChangeSetChecker},
+        write_set::{WriteOp, WriteSetMut},
+    };
+    use move_core_types::account_address::AccountAddress;
+    use std::sync::Arc;
+
+    fn state_value_to_int(state_value: StateValue) -> u128 {
+        u128::from_be_bytes(
```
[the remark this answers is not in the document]
vusirikala: That gives an error. 
I didn't understand it though. I need to reverse a vector of bytes and convert the vector into an integer. How is this going to help?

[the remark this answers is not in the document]
georgemitenkov: I haven't tried compiling the snippet I mentioned above, that's why I said "something like this"😂 This one actiually does work:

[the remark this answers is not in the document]
georgemitenkov:
```rust
use aptos_aggregator::delta_change_set::{delta_add, deserialize, serialize, DeltaChangeSet},
```

aptos-move/aptos-vm/src/delta_state_view.rs
gelash: yes, but thinking about it, shouldn't really matter here

aptos-move/aptos-vm/src/aptos_vm.rs
gelash: oh, makes sense, I remember being able to have a resolver that took ownership but can only find the one now that holds a reference, so it needs a binding / owner.
Let's leave it as is then.

gelash (approved):

aptos-move/aptos-vm/src/delta_state_view.rs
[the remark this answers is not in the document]
vusirikala: Done.

PR #8269, opened 2023-05-18
[Peer Monitoring Service] Update metrics and logs.

JoshLind:
Note: this PR is built on: https://github.com/aptos-labs/aptos-core/pull/8238.

### Description

This PR updates the metrics and logs for the peer monitoring service. Specifically, it offers two commits:
1. Cleans up the network ID label in the client and server metrics.
2. Adds new metrics to track the most obvious monitoring data, e.g., average ping latencies, distance from validators, number of connected peers and node uptime. We also periodically log this information.

### Test Plan
Manual verification and existing test infrastructure.

### DIFF
--- a/network/peer-monitoring-service/client/src/metrics.rs
+++ b/network/peer-monitoring-service/client/src/metrics.rs
@@ -3,13 +3,77 @@
 
 use aptos_config::network_id::PeerNetworkId;
 use aptos_metrics_core::{
-    register_histogram_vec, register_int_counter_vec, register_int_gauge_vec, HistogramVec,
-    IntCounterVec, IntGaugeVec,
+    histogram_opts, register_histogram_vec, register_int_counter_vec, register_int_gauge_vec,
+    HistogramVec, IntCounterVec, IntGaugeVec,
 };
 use once_cell::sync::Lazy;
 
 /// The special label TOTAL_COUNT stores the sum of all values in the counter
 pub const TOTAL_COUNT_LABEL: &str = "TOTAL_COUNT";
+
+// Histogram buckets for tracking average ping latencies (secs)
+const AVERAGE_PING_LATENCY_BUCKETS: &[f64] = &[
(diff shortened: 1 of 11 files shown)


bchocho (approved):
LGTM

network/peer-monitoring-service/client/src/metrics.rs
```diff
@@ -3,20 +3,84 @@
...
-    IntCounterVec, IntGaugeVec,
+    histogram_opts, register_histogram_vec, register_int_counter_vec, register_int_gauge_vec,
+    HistogramVec, IntCounterVec, IntGaugeVec,
 };
 use once_cell::sync::Lazy;
 
 /// The special label TOTAL_COUNT stores the sum of all values in the counter
 pub const TOTAL_COUNT_LABEL: &str = "TOTAL_COUNT";
 
+// Histogram buckets for tracking average ping latencies (secs)
```
brianolson, line 14: nit: I think it would look better interleaved so that the buckets are right next to the thing they go with

brianolson (approved):

network/peer-monitoring-service/client/src/peer_states/mod.rs
```diff
@@ -17,6 +21,9 @@ pub mod node_info;
 pub mod peer_state;
 mod request_tracker;
 
+/// The frequency at which we update the logs and metrics for the peer monitoring states
+const LOGS_AND_METRICS_FREQUENCY_SECS: u64 = 120; // 2 minutes
```
sitalkedia, line 25: Wondering if this needs to be so high - is it too expensive to collect? If not, may be keep it 30-60s so that we don't loose too much of granularity?

network/peer-monitoring-service/client/src/metrics.rs
JoshLind: Updated -- don't feel super strongly though 😄

PR #8280, opened 2023-05-18
Increase readability of nodes landing page

saharct:
### Description

Update custom.css to set the color of small text within list-group-item to black/white (depending on the selected theme). The page currently looks like a wall of turquoise-colored text. 

Before:
image

After:
image


davidiw (approved):
nice

PR #8283, opened 2023-05-18
[Storage server] Split unit tests into separate files.

JoshLind:
### Description
This PR splits the super large `test.rs` file in the storage server into multiple separate files. This should make the test layout more readable/manageable. Nothing else should change 😸 

### Test Plan
Existing test infrastructure. Also, I ran all the tests before this change and counted that we ended up with the same number of tests after we split things up, so none were lost 😄


bchocho (approved):
LGTM (somewhat of a blind accept :))

msmouse (approved):

PR #8290, opened 2023-05-19
[Gas Estimation] gate (disable) the v2 logic

bchocho:
### Description

Gate with a regular config so change is not released accidentally. Maybe want to change this to onchain eventually.

### DIFF
--- a/api/goldens/aptos_api__tests__transactions_test__test_gas_estimation_disabled.json
+++ b/api/goldens/aptos_api__tests__transactions_test__test_gas_estimation_disabled.json
@@ -1,0 +1,5 @@
+{
+  "deprioritized_gas_estimate": 0,
+  "gas_estimate": 0,
+  "prioritized_gas_estimate": 150
+}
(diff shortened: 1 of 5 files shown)


sherry-x (approved):

api/goldens/aptos_api__tests__transactions_test__test_gas_estimation_disabled.json
```diff
@@ -0,0 +1,5 @@
+{
+  "deprioritized_gas_estimate": 0,
+  "gas_estimate": 0,
```
banool, line 3: So this is how it worked before I suppose? I recall that in our test environment the gas unit price is zero, is that why this estimate is zero? Because in reality if we're returning `gas_estimate = 0` that would be really bad.

bchocho: Yup only expected in the test environment because of https://github.com/aptos-labs/aptos-core/blob/1c5f49514116d545f7cd3efe2a37204da4ee07da/config/global-constants/src/lib.rs#L25

banool: Okay all good then!

banool (approved):

api/goldens/aptos_api__tests__transactions_test__test_gas_estimation_disabled.json
banool: For later on: https://github.com/aptos-labs/aptos-core/issues/8293.

PR #8292, opened 2023-05-19
docker build 1

ibalajiarun:

### DIFF
--- a/consensus/src/dag/bullshark.rs
+++ b/consensus/src/dag/bullshark.rs
@@ -185,6 +185,11 @@ impl Bullshark {
             let anchor_author = self
                 .proposer_election
                 .get_round_anchor_peer_id(current_round);
+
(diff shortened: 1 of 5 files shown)


consensus/src/dag/bullshark.rs
```diff
@@ -97,61 +106,56 @@ impl Bullshark {
...
-
-            // wave -= 1;
-            // round -= 2;
         }
 
-        anchor_stack.push(current_anchor);
-        self.lowest_unordered_anchor_wave = new_ordered_wave + 1;
-        anchor_stack
-    }
+        let anchor_round = current_anchor.round();
```
ibalajiarun, line 133: I think `current_anchor` here should be the original one passed to the function, but this one uses the reset one in line 126.

davidiw:
looks stale?

PR #8294, opened 2023-05-19
[TS SDK] Support write operations for Fungible Assets

0xmaayan:
### Description
This PR adds support to Fungible Assets
- new `FungibleAssetClient` class that has a `transfer` and `balance` methods
- a `generateTransfer` method that generates and returns a transfer amount raw transaction (this can be used to simulate txn before signing and submitting to chain)
- support for fungible assets in the coin client class.

### Test Plan
tests are passing


ecosystem/typescript/sdk/src/plugins/coin_client.ts
```diff
@@ -54,8 +61,23 @@ export class CoinClient {
       // If this is the first time an account has received the specified coinType,
       // and this is set to false, the transaction would fail.
       createReceiverIfMissing?: boolean;
+      // fungible asset address if you wish to transfer a fungible asset
+      assetAddress?: MaybeHexString;
     },
   ): Promise<string> {
+    if (extraArgs?.assetAddress) {
```
0xmaayan, line 68: is this too much? what do you think @hariria as someone who worked with this class recently

ecosystem/typescript/sdk/src/plugins/fungible_asset_client.ts
```diff
@@ -0,0 +1,108 @@
...
+   * first simulate the transaction and then sign and submit it.
+   *
+   * @param sender The sender account
+   * @param assetAddress The fungible asset address - For example if you’re transferring USDT, this would be the USDT address
+   * @param recipient Recipient address
+   * @param amount Number of assets to transfer
+   * @param assetType (optional) The fungible asset type - default to `0x1::fungible_asset::Metadata`
+   * @returns Raw Transaction
+   */
+  async generateTransferAmount(
```
0xmaayan, line 78: @kent-white does that work for the wallet case?

kent-white: yes, I think this will work

ecosystem/typescript/sdk/src/plugins/fungible_asset_client.ts
```diff
@@ -0,0 +1,113 @@
...
+   *
+   * @param sender The sender account
+   * @param assetAddress The fungible asset address.
+   * For example if you’re transferring USDT this would be the USDT address
+   * @param recipient Recipient address
+   * @param amount Number of assets to transfer
+   * @param assetType (optional) The fungible asset type - default to `0x1::fungible_asset::Metadata`
+   * @returns The hash of the transaction submitted to the API
+   */
+  async transferAmount(
```
movekevin, line 35: nit: Let's just name this transfer instead of transferAmount

ecosystem/typescript/sdk/src/plugins/fungible_asset_client.ts
```diff
@@ -0,0 +1,113 @@
+import { AptosAccount } from "../account";
+import { RawTransaction } from "../aptos_types";
+import * as Gen from "../generated/index";
+import { OptionalTransactionArgs, Provider } from "../providers";
+import { TransactionBuilderRemoteABI } from "../transaction_builder";
+import { MaybeHexString, HexString } from "../utils";
+
+export class FungibleAssetClient {
```
movekevin, line 8: Did you consider combining this with the existing coin client at all?

0xmaayan: you mean https://github.com/aptos-labs/aptos-core/pull/8294/files#diff-dca373a5371526fcac7c2fb7088983125df962743aea81253771be2ea19dd87bR36 ?

movekevin: Oh weird sorry. I clicked on a link from email so I was accidentally only looking at changes from the latest commit

ecosystem/typescript/sdk/src/plugins/coin_client.ts
```diff
@@ -54,8 +61,32 @@ export class CoinClient {
       // If this is the first time an account has received the specified coinType,
       // and this is set to false, the transaction would fail.
       createReceiverIfMissing?: boolean;
+      // fungible asset address if you wish to transfer a fungible asset
+      assetAddress?: MaybeHexString;
```
movekevin, line 65: I wonder if we can merge this with coinType? We can check if coinType is just an address or containing the fully qualified name (address::module::struct)

ecosystem/typescript/sdk/src/plugins/coin_client.ts
```diff
@@ -54,8 +61,32 @@ export class CoinClient {
...
     },
   ): Promise<string> {
+    if (extraArgs?.assetAddress) {
+      /* eslint-disable no-console */
+      console.warn("to transfer a fungible asset, use `FungibleAssetClient()` class for a better support");
```
movekevin, line 70: Is it good practice to use console.log in the SDK/library? This might lead to varying effects depending on where the client runs, e.g. in a Cloud Scheduler job

ecosystem/typescript/sdk/src/plugins/coin_client.ts
```diff
@@ -54,8 +61,32 @@ export class CoinClient {
...
+      const provider = new Provider({
+        fullnodeUrl: this.aptosClient.nodeUrl,
+        indexerUrl: NetworkToIndexerAPI[NodeAPIToNetwork[this.aptosClient.nodeUrl]] ?? this.aptosClient.nodeUrl,
+      });
+      const fungibleAsset = new FungibleAssetClient(provider);
+      let recipient = to;
```
movekevin, line 76: This can be removed by moving this line up from below:

const toAddress = getAddressFromAccountOrAddress(to);

ecosystem/typescript/sdk/src/plugins/coin_client.ts
0xmaayan: in the new fungible asset world, 
`coinType` is the asset type (default to 0x1::fungible_asset::Metadata)
`assetAddress` is the asset address

i.e https://github.com/aptos-labs/aptos-core/pull/8294/files#diff-cb86fcb259943ce56d237a9a0b938ce90ea148ab8a7e16f1545dff606c3fa659R37-R41
`assetAddress = assetAddress`
`coinTypes = assetType`

ecosystem/typescript/sdk/src/plugins/coin_client.ts
0xmaayan: this is just a logging mechanism, I don't think it can lead to anything special other than logs being printed.

ecosystem/typescript/sdk/src/plugins/coin_client.ts
movekevin: I know that but coinType is a string here and thus it could be either 0x1::aptos_coin::AptosCoin or 0x123 (representing some fungible asset). Wouldn't that be easier to use and understand?

ecosystem/typescript/sdk/src/plugins/coin_client.ts
movekevin: In certain setups, logs can get scraped and piped into logging analytics system like Datadog automatically. I've seen a few setups like this.

ecosystem/typescript/sdk/src/tests/e2e/fungible_asset_client.test.ts
```diff
@@ -0,0 +1,129 @@
...
...3b95164525d57d96ba99219c9409a967707d5503cf27b5dd4e76da765d0ddde770e30beb44e8b95cefb1b714d420c6030a4cee1a43860d2d735aea7c3f3639b5e3a6dad502a45937c6e84df2860032c364102037858fe2cee577413dec52116b41cc641b5a5ddb219865817e3f858b37fdc9063eef105e5021d4609d9292c2acdd7e9e9876dc6bb55ee46c8ef36743723d6fc4f026ff89818a77cfa22e1782fa766d0f791702f6b3841e8880309e63411328086c15b8b7deccb7facd83af7bb76f79f997cc66b573c73160d1e1ad18733069bbf1c5ee458613d2b984d4d3aedd1ca207bbbba17fded07dcb9a4451a3612ca2c171188227f04ea94319eded19291e76a3daed611a71c725dd240515b6bc8fc911a2471d76cd383c7674e89a5a59f2336ce2f1e85a24448fbb77c7bdbeee136cf2dd2d8cf2633718d59c376e909f76dd84ba29d8ac095b94dae8bbf7815c2c2192d5c509a639948300431e90f089292540bd7bf1687777f741887296607dc3b887d467653f78fb3966430c6b9b29919af56f25b4cdcd9e126d6dc85e6f5d6ffd0b207ad810cb20000000000400000000000000000000000000000000000000000000000000000000000000010e4170746f734672616d65776f726b00000000000000000000000000000000000000000000000000000000000000010b4170746f735374646c696200000000000000000000000000000000000000000000000000000000000000010a4d6f76655374646c69620000000000000000000000000000000000000000000000000000000000000004114170746f73546f6b656e4f626a6563747300",
```
movekevin, line 40: May be not in the scope of this PR: This is hardcoded so if the contract ever changes, these tests would be outdated right? Have you explored any alternatives to doing these e2e tests?

ecosystem/typescript/sdk/src/plugins/coin_client.ts
0xmaayan: sorry not following, what we would do then if someone passes coinType (i.e assetType) and assetAddress? we need to use both

movekevin: Why do we need to use both?
1. If user passes a coinType of the form 0x1::aptos_coin::AptosCoin, this means they want to transfer coin v1
2. If user passes a coinType that's an address (0x123), we can assume they want to transfer fungible assets

ecosystem/typescript/sdk/src/plugins/coin_client.ts
0xmaayan: even with random `console.warn`? npm packages out there use that method to warn users from present/future changes in the sdk https://github.com/axios/axios/blob/21a5ad34c4a5956d81d338059ac0dd34a19ed094/lib/helpers/deprecatedMethod.js#L17

ecosystem/typescript/sdk/src/plugins/coin_client.ts
0xmaayan: and if they pass a coinType that's an address (0x123) and also want to set the asset type and not use the default asset type?

ecosystem/typescript/sdk/src/plugins/coin_client.ts
movekevin: I just wanted to bring this up. Not a blocking comment :)

ecosystem/typescript/sdk/src/tests/e2e/fungible_asset_client.test.ts
0xmaayan: yes this is where it becomes tricky. we need to somehow compile it and then publish it so we can use this module in tests. to publish we can use the SDK function, but we most pre-compile it somehow before
1. compile locally and using this string
2. compile locally and include the compiled file and then read from it

both would have a static compiled result

ecosystem/typescript/sdk/src/plugins/coin_client.ts
movekevin: What do you mean by setting the asset type? Isn't the address the fungible asset type?

0xmaayan: hmm no.. there is an assetType and assetAddress https://github.com/aptos-labs/aptos-core/pull/8294/files#diff-cb86fcb259943ce56d237a9a0b938ce90ea148ab8a7e16f1545dff606c3fa659R37-R40

assetAddress is an object address (0x123) and asset type is by [default](https://github.com/aptos-labs/aptos-core/pull/8294/files#diff-cb86fcb259943ce56d237a9a0b938ce90ea148ab8a7e16f1545dff606c3fa659R11) `0x1::fungible_asset::Metadata` but can be changed to a different type if needed

movekevin: assetType is not really required when sending a tx from off chain so defaulting to 0x1::fungible_asset::Metadata is okay. Passing a custom type is only really useful on chain in Move code

ecosystem/typescript/sdk/src/tests/e2e/fungible_asset_client.test.ts
movekevin: One idea I can come up with is change the test process to have a step that compiles and deploys the move-examples code against a local network via bash script (to invoke the CLI)

ecosystem/typescript/sdk/src/plugins/coin_client.ts
0xmaayan: looking at this [entry](https://github.com/aptos-labs/aptos-core/blob/main/aptos-move/framework/aptos-framework/sources/primary_fungible_store.move#L140-L144) function, we need to pass a type (`<T: key>`). SDK defaults to  `0x1::fungible_asset::Metadata` but gives the option to override it and pass any type they want.

ecosystem/typescript/sdk/src/plugins/fungible_asset_client.ts
```diff
@@ -0,0 +1,113 @@
...
+    assetType?: string,
+    extraArgs?: OptionalTransactionArgs,
+  ): Promise<string> {
+    const rawTxn = await this.generateTransferAmount(sender, assetAddress, recipient, amount, assetType, extraArgs);
+    const txnHash = await this.submit(sender, rawTxn);
+    return txnHash;
+  }
+
+  /**
+   * Get the balance of an account's fungible asset.
```
movekevin, line 49: We should document that this looks at an account's primary store, which would exclude their balances in all other untracked stores. cc @lightmark. We should provide a separate function (something like storeBalance) that takes a store address and return the balance.

ecosystem/typescript/sdk/src/plugins/coin_client.ts
movekevin: Hmm, in that case how about naming it as fungibleAssetType to be clear?

ecosystem/typescript/sdk/src/tests/e2e/fungible_asset_client.test.ts
movekevin: Yeah

ecosystem/typescript/sdk/src/plugins/fungible_asset_client.ts
```diff
@@ -0,0 +1,113 @@
...
+   *
+   * @param account Account that you want to get the balance of.
+   * @param assetAddress The fungible asset address you want to check the balance of
+   * @param assetType (optional) The fungible asset type - default to `0x1::fungible_asset::Metadata`
+   * @returns Promise that resolves to the balance
+   */
+  async balance(account: MaybeHexString, assetAddress: MaybeHexString, assetType?: string): Promise<bigint> {
```
geekflyer, line 56: let's name this `getBalance`.

geekflyer (changes requested):

ecosystem/typescript/sdk/src/plugins/coin_client.ts
geekflyer, line 65:
```suggestion
      assetMetadataAddress?: MaybeHexString;
```

Let's call this `assetMetadataAddress` or something like that. Just calling this `assetAddress` is confusing as intuitively this sounds like the address of the actual asset we're transferring, but really what this argument is referring to is just the metadata object address.

ecosystem/typescript/sdk/src/plugins/fungible_asset_client.ts
```diff
@@ -0,0 +1,113 @@
...
+   *
+   * This method can be used if you want/need to get the raw transaction so you can
+   * first simulate the transaction and then sign and submit it.
+   *
+   * @param sender The sender account
+   * @param assetAddress The fungible asset address.
+   * For example if you’re transferring USDT this would be the USDT address
+   * @param recipient Recipient address
+   * @param amount Number of assets to transfer
+   * @param assetType (optional) The fungible asset type - default to `0x1::fungible_asset::Metadata`
```
geekflyer, line 79: `assetMetadataType`

ecosystem/typescript/sdk/CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ All notable changes to the Aptos Node SDK will be captured in this file. This ch
 
 ## Unreleased
 
+- Add `FungibleAsset` plugin to support fungible assets
```
banool, line 9: FungibleAssetClient?

ecosystem/typescript/sdk/src/plugins/coin_client.ts
```diff
@@ -32,6 +33,12 @@ export class CoinClient {
    * this to true, the transaction will fail if the receiver account does not
    * exist on-chain.
    *
+   * TS SDK supports fungible assets operations. For CoinClient
```
banool, line 36: nit
```suggestion
   * The SDK supports fungible assets operations. For CoinClient
```

ecosystem/typescript/sdk/src/plugins/coin_client.ts
```diff
@@ -88,8 +125,26 @@ export class CoinClient {
     extraArgs?: {
       // The coin type to use, defaults to 0x1::aptos_coin::AptosCoin
       coinType?: string;
+      // fungible asset address if you wish to transfer a fungible asset
+      assetAddress?: MaybeHexString;
     },
   ): Promise<bigint> {
+    if (extraArgs?.assetAddress) {
+      /* eslint-disable no-console */
+      console.warn("to check balance of a fungible asset, use `FungibleAssetClient()` class for a better support");
```
banool, line 134: nit
```suggestion
      console.warn("to check balance of a fungible asset, use `FungibleAssetClient()` class for better support");
```

ecosystem/typescript/sdk/src/plugins/fungible_asset_client.ts
```diff
@@ -0,0 +1,113 @@
...
+   * Get the balance of an account's fungible asset.
+   *
+   * @param account Account that you want to get the balance of.
+   * @param assetAddress The fungible asset address you want to check the balance of
+   * @param assetType (optional) The fungible asset type - default to `0x1::fungible_asset::Metadata`
+   * @returns Promise that resolves to the balance
+   */
+  async balance(account: MaybeHexString, assetAddress: MaybeHexString, assetType?: string): Promise<bigint> {
+    const payload: Gen.ViewRequest = {
+      function: "0x1::primary_fungible_store::balance",
```
banool, line 58: Is there something particularly complicated about the underlying move structs explaining why we need to use a view function for this? I guess it needs to fetch the metadata by following an "object pointer"?

If not and it's just a single resource read, given reading a coin / fungible asset is a pretty common operation, it could be better to do it via the API normally.

ecosystem/typescript/sdk/src/tests/e2e/fungible_asset_client.test.ts
```diff
@@ -0,0 +1,129 @@
...
+const publisher = new AptosAccount(
+  new HexString("0x1c2b344cdc1ca1cc33d5810cf93278fd3c2a8e8ba9cd78240c1193766b06a724").toUint8Array(),
+);
+const alice = new AptosAccount();
+const bob = new AptosAccount();
+let assetAddress = "";
+/**
+ * Since there is no ready-to-use fungible asset contract/module on an aptos framework address
+ * we pre compiled ../../../aptos-move/move-examples/fungible_token contract and publish
+ * it here to local testnet so we can intercat with it to mint a fungible asset and then
```
banool, line 21:
```suggestion
 * it here to local testnet so we can interact with it to mint a fungible asset and then
```

ecosystem/typescript/sdk/src/plugins/coin_client.ts
0xmaayan: _but IMO that's confusing and would not support the case when the user wants to provide both._

That is what I say. We give the user the option to provide 2 "things"
1. asset address - to transfer an amount of assets from
2. asset type - the asset type

We default it to a `0x1::fungible_asset::Metadata` type but the user can override it if they need to. Because of that, we have 2 arguments.

0xmaayan: the asset is not always `Metadata`, it can also be `Object`, so naming it `Metadata` can be slightly wrong. It is mostly depends on the contract you interact with.

geekflyer: then maybe we need to give this a different name then what I suggested, that being said that doesn't change the fact that the current name is even more confusing imo.

0xmaayan: maybe we can have `fungibleAssetAddress` since this is the fungible asset address, (i.e 0x123)

lightmark: I said you could give the user that option to override but `Metadata` always works so actually they are no benefit of letting them override. 
So if we don't let them override, what @movekevin suggests makes sense.

ecosystem/typescript/sdk/src/plugins/fungible_asset_client.ts
lightmark: Yes, I refactored managed_fungible_asset example. @0xmaayan We can talk about this on slack.

ecosystem/typescript/sdk/src/plugins/fungible_asset_client.ts
lightmark: It should be supported natively in API at object level since we don't have indexer inside node and the ownership is not that intuitive as coin v1. So you know the balance of someone you'll query indexer first to get the fungible store object address and then call the api.

ecosystem/typescript/sdk/src/plugins/coin_client.ts
movekevin: Yeah I don't really see any benefits in letting users override when sending a tx or calling a view function. The ability to pass custom type arg is meant to be more useful when writing code in Move

ecosystem/typescript/sdk/src/plugins/coin_client.ts
```diff
@@ -54,8 +60,27 @@ export class CoinClient {
       // If this is the first time an account has received the specified coinType,
       // and this is set to false, the transaction would fail.
       createReceiverIfMissing?: boolean;
+      // fungible asset address if you wish to transfer a fungible asset
+      fungibleAssetMetadataAddress?: MaybeHexString;
```
movekevin, line 64: I don't really see the benefit of having this here where people need to call coin_client.transfer with different params anyway vs just calling fungible_asset_client directly. I was thinking that if we can combine this with coinType, at least the benefit would be that people can still use the same function coin_client.transfer with same exact inputs and args, except that coinType would be an address (instead of a struct type) as these data would be coming from indexer queries. Can you think through this flow end to end?

ecosystem/typescript/sdk/src/plugins/fungible_asset_client.ts
0xmaayan: now that we have `check balance for primary store` and `check balance for secondary store` I guess having the `coin_client.balance()` function support both or one of them doesn't make sense.

gregnazario (approved):
Didn't see any blockers.  Would be more performant / simpler to get the balance from the resource for primary store rather than using the view function, though I'm assuming you'd have to do local hashing

ecosystem/typescript/sdk/src/plugins/fungible_asset_client.ts
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  /**
+   * Get the balance of a fungible asset from the account's primary fungible store.
+   *
+   * @param account Account that you want to get the balance of.
+   * @param fungibleAssetMetadataAddress The fungible asset address you want to check the balance of
+   * @returns Promise that resolves to the balance
+   */
+  async getBalanceFromPrimaryFungibleStore(
```
movekevin, line 59: I'm not really a fan of appending FromPrimaryFungibleStore to every single function here. Although we want to distinguish from functions interacting directly with store addresses, I think a better way to name and distinguish these are to name them as:
1. Without any suffix, such as getBalance for anything interacting with the (default) primary store
2. Add a suffix [...]ForStore such as getBalanceForStore (or alternatively a Store suffix such as getStoreBalance) for functions interacting with a store address specifically

movekevin (approved):

ecosystem/typescript/sdk/src/plugins/fungible_asset_client.ts
gregnazario: getPrimaryBalance may be a simpler naming for this one

movekevin: That works too

PR #8298, opened 2023-05-19
[docs] Setup Redirects & Fix Dead Links

moonclavedev:
### Description
Fixes some dead links and set up redirects in the case of referring domains.

### Test Plan
Verify redirects are working by visiting urls in the `from:` path below `@docusaurus/plugin-client-redirects` inside `docusaurus.config.js`


0xmaayan (approved):
Thank you!

saharct (approved):

PR #8300, opened 2023-05-20
[CLI] Add init from ledger account

0xjinn:
### Description

First PR to allow user to init and create profile from their ledger account

NOTE: This is PR only assume you have your ledger account at the index 0 derivative path.

### Test Plan

Success init
```
➜  aptos-core git:(jin/ledger_cli_integration) ✗ aptos-debug init --ledger
Aptos already initialized for profile default, do you want to overwrite the existing config? [yes/no] >
yes
Configuring for profile default

Please choose a derivative path from the following 5 ledger accounts, or choose an arbitrary path that you want to use:
Path: m/44'/637'/1'/0'/0' Address 21563230cf6d69ee72a51d21920430d844ee48235e708edbafbc69708075a86e
Path: m/44'/637'/2'/0'/0' Address 667446181b3b980ef29f5145a7a2cc34d433fc3ee8c97fc044fd978435f2cb8d
Path: m/44'/637'/0'/0'/0' Address 59836ba1dd0c845713bdab34346688d6f1dba290dbf677929f2fc20593ba0cfb
Path: m/44'/637'/3'/0'/0' Address 2dcf037a9f31d93e202c074229a1b69ea8ee4d2f2d63323476001c65b0ec4f31
Path: m/44'/637'/4'/0'/0' Address 23c579a9bdde1a59f1c9d36d8d379aeefe7a5997b5b58bd5a5b0c12a4f170431
m/44'/637'/1'/0'/0'
Account 21563230cf6d69ee72a51d21920430d844ee48235e708edbafbc69708075a86e doesn't exist, creating it and funding it with 100000000 Octas
Account 21563230cf6d69ee72a51d21920430d844ee48235e708edbafbc69708075a86e funded successfully

---
Aptos CLI is now set up for account 21563230cf6d69ee72a51d21920430d844ee48235e708edbafbc69708075a86e as profile default!  Run `aptos --help` for more information about commands
{
  "Result": "Success"
}

```

profile created
```
  default:
    public_key: "0xf15355adc165b2d9136985e79e49d2ba509db90af9c3fdcbbf1eb2fa6a29a9dd"
    account: 21563230cf6d69ee72a51d21920430d844ee48235e708edbafbc69708075a86e
    rest_url: "https://fullnode.devnet.aptoslabs.com"
    faucet_url: "https://faucet.devnet.aptoslabs.com"
    derivative_path: "m/44'/637'/1'/0'/0'"
```

### DIFF
--- a/crates/aptos-ledger/src/lib.rs
+++ b/crates/aptos-ledger/src/lib.rs
@@ -24,9 +24,10 @@ use std::{
 };
 use thiserror::Error;
 
-// A piece of data which tells a wallet how to derive a specific key within a tree of keys
-// 637 is the key for Aptos
-const DERIVATIVE_PATH: &str = "m/44'/637'/{index}'/0'/0'";
+/// Derivative path template
+/// A piece of data which tells a wallet how to derive a specific key within a tree of keys
+/// 637 is the key for Aptos
+pub const DERIVATIVE_PATH: &str = "m/44'/637'/{index}'/0'/0'";
 
 const CLA_APTOS: u8 = 0x5B; // Aptos CLA Instruction class
 const INS_GET_VERSION: u8 = 0x03; // Get version instruction code
@@ -72,6 +73,33 @@ impl Display for Version {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         write!(f, "{}.{}.{}", self.major, self.minor, self.patch)
     }
+}
+
+/// Validate the input derivative path to check if it's valid
+pub fn validate_derivative_path(input: &str) -> bool {
+    let prefix = "m/44'/637'/";
+    let suffix = "'";
+
+    if input.starts_with(prefix) && input.ends_with(suffix) {
+        let inner_input = &input[prefix.len()..input.len()];
+
+        // Sample: 0'/0'/0'
+        let sections: Vec<&str> = inner_input.split('/').collect();
+        if sections.len() != 3 {
+            return false;
+        }
+
+        for section in sections {
+            let section_value = &section.trim_end_matches('\'');
+            if section_value.parse::<u32>().is_ok() {
+                continue;
+            }
+            return false;
+        }
+
+        return true;
+    }
+    false
 }
 
 /// Returns the current version of the Aptos app on Ledger
--- a/crates/aptos/Cargo.toml
+++ b/crates/aptos/Cargo.toml
@@ -31,6 +31,7 @@ aptos-genesis = { workspace = true }
 aptos-github-client = { workspace = true }
 aptos-global-constants = { workspace = true }
 aptos-keygen = { workspace = true }
+aptos-ledger = "0.1.0"
 aptos-logger = { workspace = true }
 aptos-network-checker = { workspace = true }
 aptos-node = { workspace = true }
--- a/crates/aptos/Cargo.toml
+++ b/crates/aptos/Cargo.toml
@@ -31,7 +31,7 @@ aptos-genesis = { workspace = true }
 aptos-github-client = { workspace = true }
 aptos-global-constants = { workspace = true }
 aptos-keygen = { workspace = true }
-aptos-ledger = "0.1.0"
+aptos-ledger = { workspace = true }
 aptos-logger = { workspace = true }
 aptos-network-checker = { workspace = true }
 aptos-node = { workspace = true }
--- a/crates/aptos/src/common/init.rs
+++ b/crates/aptos/src/common/init.rs
@@ -53,6 +53,10 @@ pub struct InitTool {
     #[clap(long)]
     pub skip_faucet: bool,
 
+    /// Whether you want to create a profile from your ledger account
+    ///
+    /// Make sure that you have Ledger device connected, unlocked, and have Aptos app installed
+    /// and opened. Otherwise CLI would not be able to open the connection to your account
     #[clap(long)]
     pub from_ledger: bool,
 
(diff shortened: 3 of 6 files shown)


crates/aptos/src/common/types.rs
```diff
@@ -215,6 +215,8 @@ pub struct ProfileConfig {
     /// URL for the Faucet endpoint (if applicable)
     #[serde(skip_serializing_if = "Option::is_none")]
     pub faucet_url: Option<String>,
+    /// Derivative path index of the account on ledger
+    pub bip44_index: Option<String>,
```
0xjinn, line 219: I wonder if I should store the whole "derivative path" here? Instead of the account index.

m/44'/637'/{bip44_index}'/0'/0'

.github/workflows/docker-build-test.yaml
```diff
@@ -92,13 +92,13 @@ jobs:
   # Because the docker build happens in a reusable workflow, have a separate job that collects the right metadata
   # for the subsequent docker builds. Reusable workflows do not currently have the "env" context: https://github.com/orgs/community/discussions/26671
   determine-docker-build-metadata:
+    needs: [permission-check]
```
banool, line 95: I think you need to rebase?

crates/aptos/Cargo.toml
```diff
@@ -30,6 +30,7 @@ aptos-genesis = { workspace = true }
 aptos-github-client = { workspace = true }
 aptos-global-constants = { workspace = true }
 aptos-keygen = { workspace = true }
+aptos-ledger = "0.1.0"
```
banool, line 33: re our chat irl, let's update this
```suggestion
aptos-ledger = { workspace = true }
```

crates/aptos/src/common/init.rs
```diff
@@ -131,33 +141,64 @@ impl CliCommand<()> for InitTool {
...
         }
 
+        // Set the BIP44 index to 0
+        // TODO: Need to update this allow account at different bip44 derivative path index
+        profile_config.bip44_index = if self.from_ledger {
+            Some("0".to_string())
+        } else {
+            None
+        };
+
```
banool, line 151: Your other PR deals with this indexing issue right, could we do that here too?

crates/aptos/src/common/init.rs
```diff
@@ -131,33 +141,64 @@ impl CliCommand<()> for InitTool {
...
+            None
         } else {
-            eprintln!("Enter your private key as a hex literal (0x...) [Current: {} | No input: Generate new key (or keep one if present)]", profile_config.private_key.as_ref().map(|_| "Redacted").unwrap_or("None"));
-            let input = read_line("Private key")?;
-            let input = input.trim();
-            if input.is_empty() {
-                if let Some(private_key) = profile_config.private_key {
-                    eprintln!("No key given, keeping existing key...");
-                    private_key
+            let ed25519_private_key = if let Some(key) = self
```
banool, line 157: Could you flatten this, like just make it `else if let Some(key)` etc.

crates/aptos/src/common/types.rs
banool: Could you explain why you'd choose one or the other?

0xjinn: Yep good question!
 
I actually prefer the full derivative path here, so I can just use it directly. Though I am not sure if this full derivative path is considered partial "secret".

hardsetting: Yeah full derivative path is safer imho, and no it's not secret so all good :)

crates/aptos/src/common/init.rs
hardsetting: Yeah agreed, why not making a simple option `--account-index` or similar, for selecting the account index?

crates/aptos/src/common/types.rs
hardsetting: But also it can be easily migrated in case we later on decide to change the path format. So also wouldn't worry too much.
For reference, in Petra we're storing the full path

hardsetting (approved):
Looks good! 
I would maybe include the `--account-index` option in this PR too, it should be a simple change

crates/aptos/src/common/types.rs
0xjinn: Good call! Will update this to full path

crates/aptos-ledger/src/lib.rs
```diff
@@ -74,6 +75,33 @@ impl Display for Version {
     }
 }
 
+/// Validate the input derivative path to check if it's valid
+pub fn validate_derivative_path(input: &str) -> bool {
+    let prefix = "m/44'/637'/";
+    let suffix = "'";
+
+    if input.starts_with(prefix) && input.ends_with(suffix) {
+        let inner_input = &input[prefix.len()..input.len()];
```
0xmaayan, line 84: `len()..input` suppose to have two `..`?

0xjinn: Yep, so this is like an range in Rust. Similar to `input.substring(prefix.len(), input.len())`

0xmaayan (approved):

crates/aptos/src/common/init.rs
```diff
@@ -131,33 +138,90 @@ impl CliCommand<()> for InitTool {
             Network::Custom => self.custom_network(&mut profile_config)?,
         }
 
+        // Fetch the top 5 (index 0-4) accounts from Ledger
+        let derivative_path = if self.ledger {
+            let account_map = aptos_ledger::fetch_batch_accounts(Some(0..5))?;
```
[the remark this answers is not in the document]
banool: While index would be easier, I figure if they wanna pick an index they would just use the 5. So maybe allow the full path?

PR #8301, opened 2023-05-20
quorum store sharded cache

zjma:


aptos-move/aptos-vm/src/block_executor/mod.rs
```diff
@@ -135,6 +138,19 @@ impl BlockExecutorTransactionOutput for AptosTransactionOutput {
...
 pub struct BlockAptosVM();
 
+
+/// A best-effort 4MB cache that help de-dup duplicated transactions potentially from quorum store.
+static TXN_CACHE: Lazy<SegmentedCache<HashValue, u64>> = Lazy::new(|| SegmentedCache::new(100000, 256));
+
+/// Insert a transaction hash into `TXN_CACHE` if not present.
+/// Return whether the insertion happened (i.e., cache miss).
+fn insert_into_txn_cache(txn: &Transaction) -> bool {
+    let key = txn.hash();
```
zjma, line 148: this can be slow. any cheaper ID?

PR #8305, opened 2023-05-21
Add support for staking Coins directly

banool:
### Description
If you want to interact with a delegated stake pool currently you have the to deal with the following constraints:
1. You can't stake `Coin<AptosCoin>` directly / unstake and receive `Coin<AptosCoin>`.
2. Because of this, you can only interact with the stake pool from an account with an Account resource, which excludes interactions from objects.

More context here: https://aptos-org.slack.com/archives/C036X27DZNG/p1684629998327789. 

More discussion required on if we want to actually land this. If we do, we might need an AIP. Probably not, but we should discuss it.

### Test Plan
```
cd aptos-move/framework/aptos-framework
aptos move test
```


banool:
Closing this, it seems like the correct model is to always store coins in an account, not an object (at least, that's the best practice today).

PR #8306, opened 2023-05-21
Add `entry_at_idx` function to SimpleMap

banool:
### Description
This function makes it possible to iterate through all items in a SimpleMap:
```
let len = simple_map::length(&my_map);
let i = 0;
while (i < len) {
    let key = *simple_map::key_at_idx(&my_map, i);
    // Do something.
    i = i + 1;
};
```

The inability to do this today results in a pattern where you have to have a SimpleMap with a vector alongside containing all the keys, which is unnecessary additional complexity / storage waste.

Internal discussion: https://aptos-org.slack.com/archives/C036X27DZNG/p1684789482036019.

### Test Plan
```
cd aptos-move/framework/aptos-stdlib
aptos move test
```

### DIFF
--- a/aptos-move/framework/aptos-stdlib/sources/simple_map.move
+++ b/aptos-move/framework/aptos-stdlib/sources/simple_map.move
@@ -134,6 +134,14 @@ module aptos_std::simple_map {
         (key, value)
     }
 
+    /// Get the key at a given index in the underlying vector.
+    public fun key_at_idx<Key: store, Value: store>(
+        map: &SimpleMap<Key, Value>,
+        idx: u64
+    ): &Key {
+        &vector::borrow(&map.data, idx).key
+    }
+
     fun find<Key: store, Value: store>(
         map: &SimpleMap<Key, Value>,
         key: &Key,
--- a/aptos-move/framework/aptos-stdlib/sources/simple_map.move
+++ b/aptos-move/framework/aptos-stdlib/sources/simple_map.move
@@ -134,12 +134,13 @@ module aptos_std::simple_map {
         (key, value)
     }
 
-    /// Get the key at a given index in the underlying vector.
-    public fun key_at_idx<Key: store, Value: store>(
+    /// Get the key and value at a given index in the underlying vector.
+    public fun entry_at_idx<Key: store, Value: store>(
         map: &SimpleMap<Key, Value>,
         idx: u64
-    ): &Key {
-        &vector::borrow(&map.data, idx).key
+    ): (&Key, &Value) {
+        let element = vector::borrow(&map.data, idx);
+        (&element.key, &element.value)
     }
 
     fun find<Key: store, Value: store>(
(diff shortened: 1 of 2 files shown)


0xbe1:
@banool I think the current way of iterating through a SimpleMap is to call `to_vec_pair` https://github.com/aptos-labs/aptos-core/blob/main/aptos-move/framework/aptos-stdlib/sources/simple_map.move#L105

this way no need to maintain an additional vector that contains all keys

movekevin:
> @banool I think the current way of iterating through a SimpleMap is to call `to_vec_pair` https://github.com/aptos-labs/aptos-core/blob/main/aptos-move/framework/aptos-stdlib/sources/simple_map.move#L105
> 
> this way no need to maintain an additional vector that contains all keys

+1

banool:
Calling `to_vec_pair` takes ownership of the simple map though, what if I don't want to do that.

banool:
I figure this pattern is pretty simple:
```
        let key = simple_map::key_at_idx(&my_struct.map, i);
        let value = simple_map::borrow(&my_struct.map, key);
```

But I guess you're right, an `entry_at_idx` function would be easier to use. Let me modify this.

aptos-move/framework/aptos-stdlib/sources/simple_map.move
```diff
@@ -134,6 +134,14 @@ module aptos_std::simple_map {
         (key, value)
     }
 
+    /// Get the key at a given index in the underlying vector.
+    public fun key_at_idx<Key: store, Value: store>(
```
gerben-stavenga: It seems a bit wrong to expose the fact that our SimpleMap is just a vector.

banool: I could change the comment? Of course some approach where there is a "for each" but with references would be ideal, but it doesn't seem like that's possible.

lightmark: I think Gerben's point is we instead should have sth like an iterator but the way to use that iterator would be very cumbersome.

banool: What would it look like to use this iterator? Could you share a hypothetical code snippet?

lightmark: There is no ideal implementation now as move struct cannot have reference inside. What I can think of that may be close to idea is to have a inline function like this:
```
inline fun for_each<K, V>(m: &SimpleMap, f: |&Element|) {
}
```
and expose `Element`. But IIRC, there is some issues with inline functions.
So my suggestion would generating all the keys.

banool:
We have decided this kind of exposure of the internals is bad practice. Alternatives such as returning a vec of the keys / values is unacceptable because we'd have to copy them. Once Move is able to return structs containing references we can revisit some kind of iterator over references.

If anyone wants access to this functionality now through the `key_at_idx` function you can use the module in https://github.com/banool/aptos-dport-std.

PR #8308, opened 2023-05-22
[testsuite][gha] use poetry everywhere and refactor testsuite

rustielin:
## Description

Attempts to unify all python usage across the entire repo. In general, this means that poetry is our entrypoint for all python utilities. This can generally be split into the categories of:
* test utilities in `testsuite`
* test utilities elsewhere (e.g. python SDK examples, aptos CLI and faucet integration tests, etc)
* extra scripts (e.g. in `scripts/` and elsewhere -- not supported)

#### Github Actions
* new `python-setup` action to unify and de-dup all places where we're installing python and basic deps
* pin `snok/install-poetry` to v1.3.3 by git commit hash
* use poetry everywhere we're calling python from an action

#### Testsuites
* New `Filesystem.exists(file_name)` method to test file existence
* `forge.py` no longer has to be run from repo root
* Make pyright happy again. Add types and assertions
* Remove deprecated Forge link shortening logic

#### Test framework
* Create a new testsuite entrypoint `testrun`, to do common things like calling `poetry` the right way and project path discovery. This gets rid of all the shell poetry wrappers (introduced https://github.com/aptos-labs/aptos-core/pull/7785) that start add a bit of clutter at scale
* Rename `forge_wrapper_core` to `test_framework`, as we have many python test utilities now that reuse the same framework
* Remove the `logger` decorator to make it play nice with pyright static checking. Simplify and just use a global logger definition

## How to land

Changes to test utilities and these actions should be landed against all supported branches, to ensure CI continues working as intended.

## Test Plan

Unit tests locally and in CI (using the updated actions). Tests have been written for all new test framework components, and some existing core framework stuff has new tests as well.

Some tests are expected to fail until this PR is landed in the `main` branch:
* Continuous Forge Tests - GCP / determine-test-metadata (we should just remove this test now that GCP forge is stable for 1+ weeks)


.github/actions/python-setup/action.yaml
```diff
@@ -0,0 +1,27 @@
+name: "Python setup"
+description: |
+  Runs an opinionated and unified python setup action. It does the following:
+  * Installs python
+  * Installs poetry
+  * Installs a specified poetry project, if given
+inputs:
+  POETRY_DIRECTORY:
```
geekflyer, line 8: might be clearer to call this `pyproject_location` location or something since this doesn't actually refer to the location of poetry.

perryjrandall (approved):
Should we consider moving our python deps out of testsuite altogether?

Accepting to unblock think this is going to be gnarly to land

.github/workflows/workflow-run-forge.yaml
```diff
@@ -148,7 +147,8 @@ jobs:
         shell: bash
         env:
           FORGE_RUNNER_MODE: pre-forge
-        run: testsuite/run_forge.sh
+        run: poetry run python3 forge.py test
```
sionescu, line 150: Why not `testrun forge.py` ?

sionescu (approved):

.github/workflows/workflow-run-forge.yaml
rustielin: ah yea good catch, I'll update that

PR #8321, opened 2023-05-23
[Multisig v2] Implement atomic add/remove owners with signature update funcs

alnoki:
@movekevin 

# Summary of changes

This PR adds two new functions to `multisig_account.move` for simultaneous updating owners and modifying the signature count. The remaining file changes are simply the result of auto-generated code from the Move modifications.

# Multisig account flow

Without this PR, the multisig account flow prohibits owners from being added/removed in the same transaction as a required signature count update. However this presents a UX problem because typically modifications to the security group comes with a signature count requirement update. For example:

## Removal

Consider a 4-of-7 multisig account that wishes to remove 3 signatories and go to 2-of-4. Before this PR, this modification would have to be broken down into two operations: if the signatory count is updated before removal, then the owners who are supposed to be removed could theoretically collude on a malicious transaction while the multisig is a 2-of-7. If the signatory count is updated after the removal, there could be a potential gridlock situation while the multisig is a 4-of-4 if one of the signatories has lost their key but hasn't noticed it since the last vote. This PR avoids potential conflicts that could arise during intermediate phases.

## Addition

Consider a 3-of-4 that wishes to go to 5-of-9 to dilute the votes of suspect actors: the signatory count can not be updated prior to addition (5 > 4), but if it is updated after addition then the multisig would reach a temporary state of 3-of-9 minority control. This PR avoids such a potential conflict.

### DIFF
--- a/aptos-move/framework/aptos-framework/sources/multisig_account.move
+++ b/aptos-move/framework/aptos-framework/sources/multisig_account.move
@@ -521,7 +521,7 @@ module aptos_framework::multisig_account {
     }
 
     /// Add owners then update number of signatures required, in a single operation.
-    entry fun add_owners_update_signatures_required(
+    entry fun add_owners_and_update_signatures_required(
         multisig_account: &signer,
         new_owners: vector<address>,
         new_num_signatures_required: u64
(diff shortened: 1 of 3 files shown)


aptos-move/framework/aptos-framework/sources/multisig_account.move
```diff
@@ -520,6 +520,16 @@ module aptos_framework::multisig_account {
         });
     }
 
+    /// Add owners then update number of signatures required, in a single operation.
+    entry fun add_owners_update_signatures_required(
```
movekevin, line 524: nit: Name this add_owners_and_update_signatures_required

movekevin (approved):

lightmark (approved):

PR #8323, opened 2023-05-23
[Sharding] [Execution] Add support for Analyzed txn and sharded graph partitioning

sitalkedia:
### Description

- Adds support `AnalyzedTransaction`- that returns the read and write set for a set of transactions. Currently only supports P2P and account creation but this can be further extended into supporting other types of transactions. 
- Add support for a sharded block partitioner that takes a block of analyzed transactions and returns a vector of transaction chunk that can be almost independently executed across shards. It works as follows. 

```

1. A block is partitioned into equally sized transaction chunks and sent to each shard.

   Block:

   T1 {write set: A, B}
   T2 {write set: B, C}
   T3 {write set: C, D}
   T4 {write set: D, E}
   T5 {write set: E, F}
   T6 {write set: F, G}
   T7 {write set: G, H}
   T8 {write set: H, I}
   T9 {write set: I, J}

2. Discard a bunch of transactions from the chunks and create new chunks so that
   there is no cross-shard dependency between transactions in a chunk.
  2.1 Following information is passed to each shard:
     - candidate transaction chunks to be partitioned
     - previously frozen transaction chunks (if any)
     - read-write set index mapping from the previous iteration (if any) - this contains the maximum absolute index
       of the transaction that read/wrote to a storage location indexed by the storage location.
  2.1 Each shard creates a read-write set for all transactions in the chunk and broadcasts it to all other shards.
     Shard 0                          Shard 1                           Shard 2
   +----------------------------+  +-------------------------------+  +-------------------------------+
   |        Read-Write Set      |  |         Read-Write Set         |  |         Read-Write Set         |
   |                            |  |                               |  |                               |
   |   T1 {A, B}                |  |   T4 {D, E}                   |  |   T7 {G, H}                   |
   |   T2 {B, C}                |  |   T5 {E, F}                   |  |   T8 {H, I}                   |
   |   T3 {C, D}                |  |   T6 {F, G}                   |  |   T9 {I, J}                   |
   +----------------------------+  +-------------------------------+  +-------------------------------+
  2.2 Each shard collects read-write sets from all other shards and discards transactions that have cross-shard dependencies.
             Shard 0                          Shard 1                           Shard 2
   +----------------------------+  +-------------------------------+  +-------------------------------+
   |        Discarded Txns      |  |         Discarded Txns         |  |         Discarded Txns         |
   |                            |  |                               |  |                               |
   |   - T3 (cross-shard dependency with T4) |  |   - T6 (cross-shard dependency with T7) |  | No discard |
   +----------------------------+  +-------------------------------+  +-------------------------------+
  2.3 Each shard broadcasts the number of transactions that it plans to put in the current chunk.
     Shard 0                          Shard 1                           Shard 2
   +----------------------------+  +-------------------------------+  +-------------------------------+
   |          Chunk Count       |  |          Chunk Count          |  |          Chunk Count          |
   |                            |  |                               |  |                               |
   |   2                        |  |   2                           |  |      3                        |
   +----------------------------+  +-------------------------------+  +-------------------------------+
  2.4 Each shard collects the number of transactions that all other shards plan to put in the current chunk and based
     on that, it finalizes the absolute index offset of the current chunk. It uses this information to create a read-write set
     index, which is a mapping of all the storage location to the maximum absolute index of the transaction that read/wrote to that location.
     Shard 0                          Shard 1                           Shard 2
   +----------------------------+  +-------------------------------+  +-------------------------------+
   |          Index Offset      |  |          Index Offset         |  |          Index Offset         |
   |                            |  |                               |  |                               |
   |   0                        |  |   2                           |  |   4                           |
   +----------------------------+  +-------------------------------+  +-------------------------------+
  2.5 It also uses the read-write set index mapping passed in the previous iteration to add cross-shard dependencies to the transactions. This is
     done by looking up the read-write set index for each storage location that a transaction reads/writes to and adding a cross-shard dependency.
  2.6 Returns two lists of transactions: one list of transactions that are discarded and another list of transactions that are kept.

3. Use the discarded transactions to create new chunks and repeat step 2 until N iterations.

4. For the remaining transaction chunks, add cross-shard dependencies to the transactions. This is done as follows:
   4.1 Create a read-write set with index mapping for all the transactions in the remaining chunks.
   4.2 Broadcast and collect read-write set with index mapping from all shards.
   4.3 Add cross-shard dependencies to the transactions in the remaining chunks by looking up the read-write set index
       for each storage location that a transaction reads/writes to. The idea is to find the maximum transaction index
       that reads/writes to the same location and add that as a dependency. This can be done as follows: First look up the read-write set index
       mapping received from other shards in the current iteration in descending order of shard id. If the read-write set index is not found,
       look up the read-write set index mapping received from other shards in previous iteration(s) in descending order of shard id.

```

### Test Plan
- Added Unit tests
- End to end test can partition around 0.5m transactions in a second for randomly generated P2P txns.

### DIFF
--- a/execution/block-partitioner/src/main.rs
+++ b/execution/block-partitioner/src/main.rs
@@ -6,7 +6,7 @@ use aptos_block_partitioner::{
 };
 use aptos_types::transaction::analyzed_transaction::AnalyzedTransaction;
 use clap::Parser;
-use rand::{rngs::OsRng, Rng};
+use rand::rngs::OsRng;
 use rayon::iter::{IntoParallelIterator, ParallelIterator};
 use std::{sync::Mutex, time::Instant};
 
@@ -41,10 +41,10 @@ fn main() {
         .map(|_| {
             // randomly select a sender and receiver from accounts
             let mut rng = OsRng;
-            let sender_index = rng.gen_range(0, num_accounts);
-            let receiver_index = rng.gen_range(0, num_accounts);
-            let receiver = accounts[receiver_index].lock().unwrap();
-            let mut sender = accounts[sender_index].lock().unwrap();
+
+            let indices = rand::seq::index::sample(&mut rng, num_accounts, 2);
+            let receiver = accounts[indices.index(1)].lock().unwrap();
+            let mut sender = accounts[indices.index(0)].lock().unwrap();
             create_signed_p2p_transaction(&mut sender, vec![&receiver]).remove(0)
         })
         .collect();
--- a/execution/block-partitioner/src/main.rs
+++ b/execution/block-partitioner/src/main.rs
@@ -1,0 +1,61 @@
+// Copyright © Aptos Foundation
+
+use aptos_block_partitioner::{
+    sharded_block_partitioner::ShardedBlockPartitioner,
+    test_utils::{create_signed_p2p_transaction, generate_test_account, TestAccount},
+};
+use aptos_types::transaction::analyzed_transaction::AnalyzedTransaction;
+use clap::Parser;
+use rand::{rngs::OsRng, Rng};
+use rayon::iter::{IntoParallelIterator, ParallelIterator};
+use std::{sync::Mutex, time::Instant};
+
+#[derive(Debug, Parser)]
+struct Args {
+    #[clap(long, default_value = "2000000")]
+    pub num_accounts: usize,
+
+    #[clap(long, default_value = "100000")]
+    pub block_size: usize,
+
+    #[clap(long, default_value = "10")]
+    pub num_blocks: usize,
+
+    #[clap(long, default_value = "12")]
+    pub num_shards: usize,
+}
+
+fn main() {
+    println!("Starting the block partitioning benchmark");
+    let args = Args::parse();
+    let num_accounts = args.num_accounts;
+    println!("Creating {} accounts", num_accounts);
+    let accounts: Vec<Mutex<TestAccount>> = (0..num_accounts)
+        .into_par_iter()
+        .map(|_i| Mutex::new(generate_test_account()))
+        .collect();
+    println!("Created {} accounts", num_accounts);
+    println!("Creating {} transactions", args.block_size);
+    let transactions: Vec<AnalyzedTransaction> = (0..args.block_size)
+        .into_iter()
+        .map(|_| {
+            // randomly select a sender and receiver from accounts
+            let mut rng = OsRng;
+            let sender_index = rng.gen_range(0, num_accounts);
+            let receiver_index = rng.gen_range(0, num_accounts);
+            let receiver = accounts[receiver_index].lock().unwrap();
+            let mut sender = accounts[sender_index].lock().unwrap();
+            create_signed_p2p_transaction(&mut sender, vec![&receiver]).remove(0)
+        })
+        .collect();
+
+    let partitioner = ShardedBlockPartitioner::new(args.num_shards);
+    for _ in 0..args.num_blocks {
+        let transactions = transactions.clone();
+        println!("Starting to partition");
+        let now = Instant::now();
+        partitioner.partition(transactions, 1);
+        let elapsed = now.elapsed();
+        println!("Time taken to partition: {:?}", elapsed);
+    }
+}
--- a/execution/block-partitioner/src/sharded_block_partitioner/conflict_detector.rs
+++ b/execution/block-partitioner/src/sharded_block_partitioner/conflict_detector.rs
@@ -2,9 +2,7 @@
 
 use crate::{
     sharded_block_partitioner::dependency_analysis::{RWSet, RWSetWithTxnIndex},
-    types::{
-        CrossShardDependencies, ShardId, TransactionWithDependencies, TransactionsChunk, TxnIndex,
-    },
+    types::{CrossShardDependencies, ShardId, SubBlock, TransactionWithDependencies, TxnIndex},
 };
 use aptos_crypto::hash::CryptoHash;
 use aptos_types::transaction::analyzed_transaction::AnalyzedTransaction;
@@ -118,7 +116,7 @@ impl CrossShardConflictDetector {
         current_round_rw_set_with_index: Arc<Vec<RWSetWithTxnIndex>>,
         prev_round_rw_set_with_index: Arc<Vec<RWSetWithTxnIndex>>,
         index_offset: TxnIndex,
-    ) -> TransactionsChunk {
+    ) -> SubBlock {
         let mut frozen_txns = Vec::new();
         for txn in txns.into_iter() {
             let dependency = self.get_dependencies_for_frozen_txn(
@@ -128,7 +126,7 @@ impl CrossShardConflictDetector {
             );
             frozen_txns.push(TransactionWithDependencies::new(txn, dependency));
         }
-        TransactionsChunk::new(index_offset, frozen_txns)
+        SubBlock::new(index_offset, frozen_txns)
     }
 
     fn check_for_cross_shard_conflict(
--- a/execution/block-partitioner/src/sharded_block_partitioner/conflict_detector.rs
+++ b/execution/block-partitioner/src/sharded_block_partitioner/conflict_detector.rs
@@ -4,9 +4,12 @@ use crate::{
     sharded_block_partitioner::dependency_analysis::{RWSet, WriteSetWithTxnIndex},
     types::{CrossShardDependencies, ShardId, SubBlock, TransactionWithDependencies, TxnIndex},
 };
-use aptos_crypto::hash::CryptoHash;
-use aptos_types::transaction::analyzed_transaction::AnalyzedTransaction;
-use std::sync::Arc;
+use aptos_types::transaction::analyzed_transaction::{AnalyzedTransaction, StorageLocation};
+use std::{
+    collections::hash_map::DefaultHasher,
+    hash::{Hash, Hasher},
+    sync::Arc,
+};
 
 pub struct CrossShardConflictDetector {
     shard_id: ShardId,
@@ -129,6 +132,12 @@ impl CrossShardConflictDetector {
         false
     }
 
+    fn get_anchor_shard_id(&self, storage_location: &StorageLocation) -> ShardId {
+        let mut hasher = DefaultHasher::new();
+        storage_location.hash(&mut hasher);
+        (hasher.finish() % self.num_shards as u64) as usize
+    }
+
     fn check_for_read_conflict(
         &self,
         current_shard_id: ShardId,
@@ -140,7 +149,7 @@ impl CrossShardConflictDetector {
             // During conflict resolution, shards starts scanning from the anchor shard id and
             // first shard id that has taken a read/write lock on this storage location is the owner of this storage location.
             // Please note another alternative is scan from first shard id, but this will result in non-uniform load across shards in case of conflicts.
-            let anchor_shard_id = read_location.hash().byte(0) as usize % self.num_shards;
+            let anchor_shard_id = self.get_anchor_shard_id(read_location);
             for offset in 0..self.num_shards {
                 let shard_id = (anchor_shard_id + offset) % self.num_shards;
                 // Ignore if this is from the same shard
@@ -163,7 +172,7 @@ impl CrossShardConflictDetector {
         cross_shard_rw_set: &[RWSet],
     ) -> bool {
         for write_location in txn.write_hints().iter() {
-            let anchor_shard_id = write_location.hash().byte(0) as usize % self.num_shards;
+            let anchor_shard_id = self.get_anchor_shard_id(write_location);
             for offset in 0..self.num_shards {
                 let shard_id = (anchor_shard_id + offset) % self.num_shards;
                 // Ignore if this is from the same shard
--- a/execution/block-partitioner/src/sharded_block_partitioner/conflict_detector.rs
+++ b/execution/block-partitioner/src/sharded_block_partitioner/conflict_detector.rs
@@ -1,7 +1,7 @@
 // Copyright © Aptos Foundation
 
 use crate::{
-    sharded_block_partitioner::dependency_analysis::{RWSet, RWSetWithTxnIndex},
+    sharded_block_partitioner::dependency_analysis::{RWSet, WriteSetWithTxnIndex},
     types::{CrossShardDependencies, ShardId, SubBlock, TransactionWithDependencies, TxnIndex},
 };
 use aptos_crypto::hash::CryptoHash;
@@ -25,7 +25,7 @@ impl CrossShardConflictDetector {
         &mut self,
         txns: Vec<AnalyzedTransaction>,
         cross_shard_rw_set: &[RWSet],
-        prev_rounds_rw_set_with_index: Arc<Vec<RWSetWithTxnIndex>>,
+        prev_rounds_rw_set_with_index: Arc<Vec<WriteSetWithTxnIndex>>,
     ) -> (
         Vec<AnalyzedTransaction>,
         Vec<CrossShardDependencies>,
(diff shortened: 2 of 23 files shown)


types/src/transaction/analyzed_transaction.rs
```diff
@@ -0,0 +1,259 @@
...
+                        ),
+                    }
+                },
+                _ => AnalyzedTransaction::new(
+                    Transaction::UserTransaction(signed_txn),
+                    vec![],
+                    vec![],
+                ),
+            },
+            _ => AnalyzedTransaction::new(txn, vec![], vec![]),
```
zjma, line 250: `_ => todo!()` instead? we need to pass in the r/w set but we don't know how to calculate.

execution/block-partitioner/src/main.rs
```diff
@@ -0,0 +1,61 @@
...
+    println!("Creating {} transactions", args.block_size);
+    let transactions: Vec<AnalyzedTransaction> = (0..args.block_size)
+        .into_iter()
+        .map(|_| {
+            // randomly select a sender and receiver from accounts
+            let mut rng = OsRng;
+            let sender_index = rng.gen_range(0, num_accounts);
+            let receiver_index = rng.gen_range(0, num_accounts);
+            let receiver = accounts[receiver_index].lock().unwrap();
+            let mut sender = accounts[sender_index].lock().unwrap();
```
zjma, line 47: potential double lock acquisition when `num_accounts` is small.

types/src/transaction/mod.rs
```diff
@@ -946,6 +947,10 @@ impl TransactionOutput {
         }
     }
 
+    pub fn retried() -> Self {
```
zjma, line 950: This method is not used.

execution/block-partitioner/src/test_utils.rs
```diff
@@ -0,0 +1,109 @@
...
+
+pub fn generate_test_account_for_address(account_address: AccountAddress) -> TestAccount {
+    TestAccount {
+        account_address,
+        private_key: Ed25519PrivateKey::generate_for_testing(),
+        sequence_number: 0,
+    }
+}
+
+pub fn create_no_dependency_transaction(num_transactions: usize) -> Vec<AnalyzedTransaction> {
```
zjma, line 40: Unused.

execution/block-partitioner/src/sharded_block_partitioner/mod.rs
```diff
@@ -0,0 +1,685 @@
...
+        }
+
+        let mut result = Vec::new();
+        result.push(Vec::new());
+
+        for sender in sender_order {
+            let txns = sender_to_txns.remove(&sender).unwrap();
+            let txns_in_shard = result.last().unwrap().len();
+
+            if txns_in_shard < approx_txns_per_shard {
```
zjma, line 183: The distribution can be slightly more balanced with `if txns_in_shard + txns.len() < approx_txns_per_shard {`?

execution/block-partitioner/src/sharded_block_partitioner/conflict_detector.rs
```diff
@@ -0,0 +1,196 @@
...
+                if rw_set_with_index.has_write_lock(read_location) {
+                    cross_shard_dependencies.add_depends_on_txn(
+                        rw_set_with_index.get_write_lock_txn_index(read_location),
+                    );
+                    break;
+                }
+            }
+        }
+
+        for write_location in frozen_txn.write_set().iter() {
```
zjma, line 89: can be chained to L71 to save some lines?

zjma (approved):

types/src/transaction/analyzed_transaction.rs
```diff
@@ -0,0 +1,259 @@
...
+                        (AccountAddress::ONE, "aptos_account", "create_account") => {
+                            let sender_address = signed_txn.sender();
+                            let receiver_address = bcs::from_bytes(&func.args()[0]).unwrap();
+                            AnalyzedTransaction::analyzed_transaction_for_create_account(
+                                signed_txn,
+                                sender_address,
+                                receiver_address,
+                            )
+                        },
+                        _ => AnalyzedTransaction::new(
```
msmouse, line 237: nit: I'd add a `AnalyzedTransaction::no_hints(txn)` constructor

msmouse (commented):
First pass (only looked at easy things :trollface: ), thanks man!

msmouse (commented):
First pass (only looked at easy things :trollface: ), thanks man!

msmouse (commented):
First pass (only looked at easy things :trollface: ), thanks man!

execution/block-partitioner/src/sharded_block_partitioner/conflict_detector.rs
zjma: nvm i missed something.

execution/block-partitioner/src/sharded_block_partitioner/conflict_detector.rs
```diff
@@ -0,0 +1,196 @@
...
+        }
+        // Iterate through the frozen dependencies and add the max transaction index for each storage location
+        let mut cross_shard_dependencies = CrossShardDependencies::default();
+        for read_location in frozen_txn.read_set().iter() {
+            // For current round, iterate through all shards less than current shards in the reverse order and for previous rounds iterate through all shards in the reverse order
+            // and find the first shard id that has taken a write lock on the storage location. This ensures that we find the highest txn index that is conflicting
+            // with the current transaction.
+            for rw_set_with_index in current_round_rw_set_with_index
+                .iter()
+                .take(self.shard_id)
```
zjma, line 77: missing a `.rev()`? I didn't find `current_round_rw_set_with_index` was reversed anywhere else...

execution/block-partitioner/src/sharded_block_partitioner/conflict_detector.rs
```diff
@@ -0,0 +1,196 @@
...
+                .chain(prev_rounds_rw_set_with_index.iter())
```
zjma, line 78: missing a .rev()? I didn't find `prev_rounds_rw_set_with_index` was reversed anywhere else...

execution/block-partitioner/src/sharded_block_partitioner/partitioning_shard.rs
```diff
@@ -0,0 +1,242 @@
...
+        // this will be used to determine the absolute index of accepted transactions in this shard.
+        self.broadcast_num_accepted_txns(accepted_txns.len());
+        let accepted_txns_vec = self.collect_num_accepted_txns();
+        // Calculate the absolute index of accepted transactions in this shard, which is the sum of all accepted transactions
+        // from other shards whose shard id is smaller than the current shard id and the number of accepted transactions in the
+        // previous rounds
+        let mut index_offset = prev_rounds_frozen_chunks
+            .iter()
+            .map(|chunk| chunk.len())
+            .sum::<usize>();
```
msmouse, line 164: it seems wasteful to do `sum()` repeatedly, how about tracking it with a number across rounds?

execution/block-partitioner/src/sharded_block_partitioner/mod.rs
```diff
@@ -0,0 +1,685 @@
...
+        // First round, we filter all transactions with cross-shard dependencies
+        let mut txns_to_partition = self.reorder_txns_by_senders(transactions);
+        let mut frozen_rw_set_with_index = Arc::new(Vec::new());
+        let mut frozen_chunks = Arc::new(Vec::new());
+
+        for _ in 0..num_partitioning_round {
+            let (
+                current_frozen_chunks_vec,
+                current_frozen_rw_set_with_index_vec,
+                latest_txns_to_partition,
```
msmouse, line 303: More logical to call it "discarded" (provided the name of the function)

execution/block-partitioner/src/sharded_block_partitioner/messages.rs
```diff
@@ -0,0 +1,87 @@
...
+            prev_rounds_rw_set_with_index,
+            prev_rounds_frozen_chunks,
+        }
+    }
+}
+
+pub struct PartitioningBlockResponse {
+    pub frozen_chunk: SubBlock,
+    pub rw_set_with_index: RWSetWithTxnIndex,
+    pub rejected_txns: Vec<AnalyzedTransaction>,
```
msmouse, line 72: "rejected" and "discarded" seem to mean the same thing across the change, how about unifying the concept?

execution/block-partitioner/src/sharded_block_partitioner/mod.rs
```diff
@@ -0,0 +1,685 @@
...
+        transactions: Vec<AnalyzedTransaction>,
+        num_partitioning_round: usize,
+    ) -> Vec<SubBlock> {
+        let total_txns = transactions.len();
+        if total_txns == 0 {
+            return vec![];
+        }
+
+        // First round, we filter all transactions with cross-shard dependencies
+        let mut txns_to_partition = self.reorder_txns_by_senders(transactions);
```
msmouse, line 295: So this is reordering and partitioning..

execution/block-partitioner/src/sharded_block_partitioner/mod.rs
```diff
@@ -0,0 +1,685 @@
...
+        let mut frozen_rw_set_with_index = Arc::new(Vec::new());
+        let mut frozen_chunks = Arc::new(Vec::new());
+
+        for _ in 0..num_partitioning_round {
```
msmouse, line 299: And these rounds don't really partition, but merely reordering?

execution/block-partitioner/src/sharded_block_partitioner/partitioning_shard.rs
```diff
@@ -0,0 +1,242 @@
...
+                rejected_txns,
+            ))
+            .unwrap();
+    }
+
+    fn add_txns_with_cross_shard_deps(&self, partition_msg: AddTxnsWithCrossShardDep) {
+        let AddTxnsWithCrossShardDep {
+            transactions,
+            index_offset,
+            prev_rounds_frozen_chunks,
```
msmouse, line 194: not used?

msmouse (commented):
second pass 
(`add_txns_with_cross_shard_deps` not finished)
---- 
update: finished

execution/block-partitioner/src/sharded_block_partitioner/mod.rs
sitalkedia: Can you elaborate why the above will lead to a better balanced distribution?

execution/block-partitioner/src/sharded_block_partitioner/conflict_detector.rs
sitalkedia: Good eye, fixed it.

execution/block-partitioner/src/sharded_block_partitioner/conflict_detector.rs
```diff
@@ -0,0 +1,196 @@
...
+        current_round_rw_set_with_index: Arc<Vec<RWSetWithTxnIndex>>,
+        prev_rounds_rw_set_with_index: Arc<Vec<RWSetWithTxnIndex>>,
+    ) -> CrossShardDependencies {
+        if current_round_rw_set_with_index.is_empty() && prev_rounds_rw_set_with_index.is_empty() {
+            return CrossShardDependencies::default();
+        }
+        // Iterate through the frozen dependencies and add the max transaction index for each storage location
+        let mut cross_shard_dependencies = CrossShardDependencies::default();
+        for read_location in frozen_txn.read_set().iter() {
+            // For current round, iterate through all shards less than current shards in the reverse order and for previous rounds iterate through all shards in the reverse order
```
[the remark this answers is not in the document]
sitalkedia: Good point, the code missed the reverse ordering. Fixed it.

execution/block-partitioner/src/sharded_block_partitioner/conflict_detector.rs
```diff
@@ -0,0 +1,196 @@
...
+            }
+        }
+
+        for write_location in frozen_txn.write_set().iter() {
+            for rw_set_with_index in current_round_rw_set_with_index
+                .iter()
+                .take(self.shard_id)
+                .chain(prev_rounds_rw_set_with_index.iter())
+            {
+                if rw_set_with_index.has_read_lock(write_location) {
```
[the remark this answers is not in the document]
sitalkedia: Here the dependency means that this transaction needs to wait for the previous txn to complete.  Does it make sense?

execution/block-partitioner/src/sharded_block_partitioner/partitioning_shard.rs
```diff
@@ -0,0 +1,242 @@
...
+        let current_rw_set_with_index = RWSetWithTxnIndex::new(&accepted_txns, index_offset);
+
+        let accepted_txns_with_dependencies = accepted_txns
+            .into_iter()
+            .zip(accepted_cross_shard_dependencies.into_iter())
+            .map(|(txn, dependencies)| TransactionWithDependencies::new(txn, dependencies))
+            .collect::<Vec<TransactionWithDependencies>>();
+
+        let frozen_chunk = SubBlock::new(index_offset, accepted_txns_with_dependencies);
+        drop(prev_rounds_frozen_chunks);
```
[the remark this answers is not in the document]
sitalkedia: This is needed because, the reference needs to be dropped explicitly before sending the response back as on the other side, we rely on unwrapping the Arc.

execution/block-partitioner/src/sharded_block_partitioner/mod.rs
```diff
@@ -0,0 +1,685 @@
...
+            let (
+                current_frozen_chunks_vec,
+                current_frozen_rw_set_with_index_vec,
+                latest_txns_to_partition,
+            ) = self.discard_txns_with_cross_shard_dependencies(
+                txns_to_partition,
+                frozen_chunks.clone(),
+                frozen_rw_set_with_index.clone(),
+            );
+            let mut prev_frozen_chunk = Arc::try_unwrap(frozen_chunks).unwrap();
```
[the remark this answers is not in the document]
sitalkedia: Renamed everything from "chunk" -> "sub block"

execution/block-partitioner/src/sharded_block_partitioner/partitioning_shard.rs
sitalkedia: It doesn't add too much overhead as the number of chunks are not that hight, doing this in the master actually might make it more expensive because of single thread computation. Let me leave a TODO here to re-evaluate if it becomes expensive?

execution/block-partitioner/src/sharded_block_partitioner/mod.rs
sitalkedia: Yes, the naming is a bit confusing - open to suggestion here.

execution/block-partitioner/src/sharded_block_partitioner/mod.rs
sitalkedia: Yes correct.

execution/block-partitioner/src/sharded_block_partitioner/messages.rs
sitalkedia: Renamed all "rejected" -> "discarded".

types/src/transaction/analyzed_transaction.rs
sitalkedia: Not needed with updated code anymore.

types/src/transaction/analyzed_transaction.rs
```diff
@@ -0,0 +1,259 @@
...
+    WildCardTable(TableHandle),
+}
+
+impl CryptoHash for StorageLocation {
+    type Hasher = DummyHasher;
+
+    fn hash(&self) -> HashValue {
+        match self {
+            StorageLocation::Specific(state_key) => CryptoHash::hash(state_key),
+            _ => todo!("hashing of wildcard storage location is not supported yet"),
```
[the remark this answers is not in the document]
sitalkedia: For other variants as well, we will need to evaluate. In the worst case, we can cache the hash in the storage location itself, if the variants themselves don't cache. But avoid the hash calculation is critical for performance of the partitioning. LMK your thoughts.

sitalkedia:
@zjma , @msmouse - Thank you so much for reviewing this long PR. Addressed all the comments. Please take another look.

execution/block-partitioner/src/sharded_block_partitioner/mod.rs
zjma: Actually not much difference... please ignore this comment :D

types/src/transaction/analyzed_transaction.rs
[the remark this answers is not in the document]
msmouse: well, it'll be a domain separation issue if we don't distinguish between variants IIUC. @alinush @zjma

execution/block-partitioner/src/sharded_block_partitioner/conflict_detector.rs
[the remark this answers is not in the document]
msmouse: The storage / cache is versioned, i.e. for exampel txn 1 reads location A at version 0 shouldn't prevent txn 5 from writing to it (at version 5).

types/src/transaction/analyzed_transaction.rs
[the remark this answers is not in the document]
sitalkedia: Can you clarify why there might be conflict if we don't have domain separation? In any case, once we introduce other variants, we can have variants cache the hash or we can change this to cache the hash in the storage location itself. For now, its not needed as the hash is already cached in `StateKey`.

execution/block-partitioner/src/sharded_block_partitioner/partitioning_shard.rs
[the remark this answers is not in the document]
msmouse: Oh I see. Document it. (Also looks like we can just pass in a number instead of this `Arc`?)

execution/block-partitioner/src/sharded_block_partitioner/partitioning_shard.rs
msmouse: That's fine.

execution/block-partitioner/src/sharded_block_partitioner/conflict_detector.rs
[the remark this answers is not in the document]
sitalkedia: You are right, that was stupid of me. Let me fix this.

execution/block-partitioner/src/sharded_block_partitioner/mod.rs
msmouse: `partition_by_senders` sounds more right..

types/src/transaction/analyzed_transaction.rs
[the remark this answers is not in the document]
msmouse: I mean the hash of a StateKey (which is also that of `StorageLocation::Specifict(key)` of the same key) can conflict with that of a `StorageLocation::WildCard(type)`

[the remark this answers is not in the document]
zjma: actually why does `StorageLocation` need to implement `CryptoHash`? IIRC it only needs to be the key of some hash map.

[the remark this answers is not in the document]
sitalkedia: > I mean the hash of a StateKey (which is also that of `StorageLocation::Specifict(key)` of the same key) can conflict with that of a `StorageLocation::WildCard(type)`

@msmouse - In that case, we can cache the hash in the `StorageLocation` itself when we add the new variants? For now, we don't need it as we are supporting only one. 


@zjma - Really `CryptoHash` is not needed but Since we already cache the `CryptoHash` for underlying `StateKey`, we can just reuse it for free without hash recalculation.

execution/block-partitioner/src/sharded_block_partitioner/partitioning_shard.rs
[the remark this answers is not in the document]
sitalkedia: Documented, I am passing `Arc` instead of number because to avoid the additional copy.

types/src/transaction/analyzed_transaction.rs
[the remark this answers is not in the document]
msmouse: @zjma  Good point.. 
@sitalkedia  Looks like you are not getting the benefit because the hash sets uses the derived `Hash` trait..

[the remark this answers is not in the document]
zjma: should be:
```
impl Hash for StorageLocation {
    fn hash<H: Hasher>(&self, state: &mut H) {
        match &self {
            Specific(key) => CryptoHash::hash(key),
            ...
        }
    }
}
```
i guess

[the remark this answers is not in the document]
msmouse: And let's note in the code that this is not conflict free, so can't be simply changed to CryptoHash implementation directly. -- just in case. (sorry I'm a little paranoid about it since evil @alinush sit close to me)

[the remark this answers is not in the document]
sitalkedia: Good point @msmouse , @zjma . I have removed the `CryptoHash` for now - we were using this in one place but that didn't cause any noticeable regression. 

Regarding conflict, that should not matter as we are using this only for hashmap lookup i.e., we don't need conflict free semantic like the `StateKey`

PR #8334, opened 2023-05-23
Deprecate materialize_deltas flag in block executor

vusirikala:
### Description

A small PR to deprecate `materialize_deltas` flag in block executor and replace it with `aggregators_enabled` flag instead.


aptos-move/aptos-vm/src/block_executor/vm_wrapper.rs
```diff
@@ -61,23 +60,15 @@ impl<'a, S: 'a + StateView + Sync> ExecutorTask for AptosExecutorTask<'a, S> {
...
         txn: &PreprocessedTransaction,
         txn_idx: TxnIndex,
-        materialize_deltas: bool,
+        _aggregator_enabled: bool,
     ) -> ExecutionStatus<AptosTransactionOutput, VMStatus> {
         let log_context = AdapterLogSchema::new(self.base_view.id(), txn_idx as usize);
 
         match self
             .vm
             .execute_single_transaction(txn, &view.as_move_resolver(), &log_context)
```
gelash, line 69: hm, but we would have to propagate this to the execution and set the flag in there to materialize the aggregators, right? Otherwise the sequential execution should break currently, as it would still return deltas but nobody would materialize the deltas anymore?

gelash (commented):
Thanks looking at this one

aptos-move/aptos-vm/src/block_executor/vm_wrapper.rs
vusirikala: Done.

gelash (approved):

aptos-move/aptos-vm/src/aptos_vm.rs
```diff
@@ -1589,7 +1626,7 @@ impl VMValidator for AptosVM {
         };
 
         let resolver = &state_view.as_move_resolver();
-        let mut session = self.0.new_session(resolver, SessionId::txn(&txn), true);
+        let mut session = self.0.new_session(resolver, SessionId::txn(&txn), false);
```
gelash, line 1629: makes sense to me but let's add a comment mentioning we do txn validation with aggregator disabled (and maybe a rationale). I suppose based on the method comment that here anyway we only run the prologue (so no aggregators currently) and want to make sure those checks happen in a straightforward way.

aptos-move/aptos-vm/src/aptos_vm.rs
```diff
@@ -1781,9 +1824,9 @@ impl AptosSimulationVM {
 
         // Revalidate the transaction.
         let txn_data = TransactionMetadata::new(txn);
-        let mut session =
-            self.0
-                .new_session(resolver, SessionId::txn_meta(&txn_data), aggregator_enabled);
+        let mut session = self
+            .0
+            .new_session(resolver, SessionId::txn_meta(&txn_data), false);
```
gelash, line 1829: same. let's comment to note we disable aggregators here in simulation, and maybe the rationale.

aptos-move/aptos-vm/src/aptos_vm.rs
```diff
@@ -1844,6 +1888,7 @@ impl AptosSimulationVM {
                                         &mut gas_meter,
                                         &storage_gas_params.change_set_configs,
                                         &txn_data,
+                                        false,
```
gelash, line 1891: same

aptos-move/aptos-vm/src/aptos_vm.rs
```diff
@@ -1871,6 +1916,7 @@ impl AptosSimulationVM {
                 log_context,
                 &mut new_published_modules_loaded,
                 &storage_gas_params.change_set_configs,
+                false,
```
gelash, line 1919: this one's fine, the above comment says it's deprecated.

aptos-move/aptos-vm/src/block_executor/vm_wrapper.rs
```diff
@@ -60,22 +60,21 @@ impl<'a, S: 'a + StateView + Sync> ExecutorTask for AptosExecutorTask<'a, S> {
...
-                        .try_materialize(view)
-                        .expect("Delta materialization failed");
+        match self.vm.execute_single_transaction(
+            txn,
+            &view.as_move_resolver(),
+            &log_context,
+            aggregator_enabled,
+        ) {
+            Ok((vm_status, vm_output, sender)) => {
+                if let PreprocessedTransaction::UserTransaction(_) = txn {
```
gelash, line 73: awesome for fixing this flow!

I think you can do here: matches!(txn, PreprocessedTransaction::UserTransaction(_)) or the appropriate matches syntax. Useful because you can && that with aggregator_enabled to avoid nested ifs.

Could also put the whole predicate inside assert now, or have if matches && !aggregator_enabled followed by the current assert - I don't have a strong preference. But let's add a 1-line comment to the assert, smt like "with aggregators disabled, there should be no delta outputs"

aptos-move/aptos-vm/src/move_vm_ext/session.rs
```diff
@@ -344,7 +344,9 @@ impl<'r, 'l> SessionExt<'r, 'l> {
                     let write_op = woc.convert_aggregator_mod(&state_key, value)?;
                     write_set_mut.insert((state_key, write_op));
                 },
-                AggregatorChange::Merge(delta_op) => delta_change_set.insert((state_key, delta_op)),
+                AggregatorChange::Merge(delta_op) => {
```
gelash, line 347: what happened here, just lint?

gelash (approved):

aptos-move/aptos-vm/src/move_vm_ext/session.rs
vusirikala: Yeah. Just lint.

aptos-move/aptos-vm/src/aptos_vm.rs
georgemitenkov: Actually I think we should keep this to be `aggregator_enabled` in simulation as well. @vusirikala in `pub fn simulate_signed_transaction` we re-execute simulation if aggregators fail. Rationale for keeping this in simulation is that simulation should mimic execution flow, and so if we have a gas model for agregators/re-execution it should be reflected there?

georgemitenkov: Or alternatively we clean-up the re-execution part if we always disable aggregators :)

aptos-move/aptos-vm/src/aptos_vm.rs
vusirikala: Added a comment.

aptos-move/aptos-vm/src/aptos_vm.rs
vusirikala: As you mentioned, I enabled aggregators for `simulate_signed_transaction` and disabled aggregators for `failure_transaction_cleanup`.

aptos-move/aptos-vm/src/aptos_vm.rs
vusirikala: As you mentioned, I enabled aggregators for simulate_signed_transaction and disabled aggregators for failure_transaction_cleanup.

aptos-move/aptos-vm/src/block_executor/vm_wrapper.rs
vusirikala: Done.

msmouse:
@gelash  yeah I realized that 1 or 2 weeks ago and haven't got around to address it. And then @vusirikala found my dark secret, we synced and agreed to output WriteOp::Modification everywhere for aggregators for now. -- I'll follow up if the state metadata thing turned out to be still useful. Thanks guys.

aptos-move/aptos-vm/src/adapter_common.rs
```diff
@@ -53,6 +54,15 @@ pub(crate) trait VMAdapter {
         txn: &PreprocessedTransaction,
         data_cache: &impl MoveResolverExt,
         log_context: &AdapterLogSchema,
+        aggregator_enabled: bool,
+    ) -> Result<(VMStatus, VMOutput, Option<String>), VMStatus>;
+
+    /// Execute a single transaction when running the block sequentially.
```
georgemitenkov, line 60: Can we add a comment describing why we have `execute_single_transaction_sequential` and `execute_single_transaction`, i.e. problems with materialization and aggregator_disabled flag? Just to make sure this is documented in the code :)

aptos-move/aptos-vm/src/aptos_vm.rs
```diff
@@ -1701,6 +1749,41 @@ impl VMAdapter for AptosVM {
             },
         })
     }
+
+    fn execute_single_transaction_sequential(
+        &self,
+        txn: &PreprocessedTransaction,
+        view: &impl StateView,
```
georgemitenkov, line 1756: You can make this `resolver: &impl MoveResolverExt` like in `execute_single_transaction` to avoid `&view.as_move_resolver()` calls?

third_party/move/move-core/types/src/vm_status.rs
```diff
@@ -43,6 +43,12 @@ pub static EXECUTION_STATUS_MIN_CODE: u64 = 4000;
 /// The maximum status code for runtim statuses
 pub static EXECUTION_STATUS_MAX_CODE: u64 = 4999;
 
+/// When `Addition` operation overflows the `limit`.
+pub const EADD_OVERFLOW: u64 = 0x02_0001;
```
georgemitenkov, line 47: Aggregators are not part of Move, we shouldn't put it here. Why not use the codes from `aggregatr_extension.rs`?

third_party/move/move-core/types/src/vm_status.rs
```diff
@@ -162,6 +168,15 @@ impl VMStatus {
         self.status_code().status_type()
     }
 
+    pub fn is_aggregator_error(&self) -> bool {
```
georgemitenkov, line 171: Same, let's move it to aggregator extension?

third_party/move/move-core/types/src/vm_status.rs
```diff
@@ -162,6 +168,15 @@ impl VMStatus {
...
+        if let VMStatus::MoveAbort(_, code) = self {
```
georgemitenkov, line 172: I think something like this can work?

`matches!(self, VMStatus::MoveAbort(_, code) if *code == EADD_OVERFLOW || *code == ESUB_UNDERFLOW)`

aptos-move/block-executor/src/task.rs
```diff
@@ -57,12 +57,20 @@ pub trait ExecutorTask: Sync {
...
-    fn execute_transaction(
+    fn execute_transaction_parallel(
+        &self,
+        view: &impl TStateView<Key = <Self::Txn as Transaction>::Key>,
+        txn: &Self::Txn,
+        txn_idx: TxnIndex,
+        aggregator_enabled: bool,
+    ) -> ExecutionStatus<Self::Output, Self::Error>;
+
+    /// Execute a single transaction given the view of the current state.
```
georgemitenkov, line 68: I would also add a sentence or 2 describing the rationale of having/not-having `aggregator_enabled`

aptos-move/block-executor/src/proptest_types/types.rs
```diff
@@ -466,6 +466,15 @@ where
...
         }
     }
+
+    fn execute_transaction_sequential(
+        &self,
+        view: &impl TStateView<Key = K>,
+        txn: &Self::Txn,
+        txn_idx: TxnIndex,
+    ) -> ExecutionStatus<Self::Output, Self::Error> {
+        self.execute_transaction_parallel(view, txn, txn_idx, true)
```
georgemitenkov, line 476: Is this intended?

aptos-move/aptos-vm/src/move_vm_ext/session.rs
```diff
@@ -472,10 +474,7 @@ impl<'r> WriteOpConverter<'r> {
                     },
                 }
             },
-            Some(existing_value) => match existing_value.into_metadata() {
-                None => WriteOp::Modification(data),
-                Some(metadata) => WriteOp::ModificationWithMetadata { data, metadata },
-            },
+            Some(_) => WriteOp::Modification(data),
```
georgemitenkov, line 477: A few lines above we also have ` Some(metadata) => WriteOp::CreationWithMetadata { ...`. Why keeping this and not `Creation`? Although I think it doesn't really matter because aggregator did not distinguish `Create` vs `Modify` and when we embed it in the future this will be gone anyway

aptos-move/aptos-vm/src/block_executor/vm_wrapper.rs
```diff
@@ -53,29 +53,77 @@ impl<'a, S: 'a + StateView + Sync> ExecutorTask for AptosExecutorTask<'a, S> {
...
-            Ok((vm_status, mut vm_output, sender)) => {
-                if materialize_deltas {
-                    // TODO: Integrate delta application failure.
-                    vm_output = vm_output
-                        .try_materialize(view)
-                        .expect("Delta materialization failed");
-                }
-
+            Ok((vm_status, vm_output, sender)) => {
+                // If aggregators are disabled for user transactions, then delta_change_set should be empty as there should not be any delta outputs.
```
georgemitenkov, line 126: Should we have a debug_assert `delta_change_set .is_empty()` or something like that with this comment?

aptos-move/aptos-vm/src/block_executor/vm_wrapper.rs
```diff
@@ -53,29 +53,77 @@ impl<'a, S: 'a + StateView + Sync> ExecutorTask for AptosExecutorTask<'a, S> {
...
+                    ExecutionStatus::SkipRest(AptosTransactionOutput::new(vm_output))
+                } else {
+                    ExecutionStatus::Success(AptosTransactionOutput::new(vm_output))
+                }
+            },
+            Err(err) => ExecutionStatus::Abort(err),
+        }
+    }
+
+    // This function is called by the BlockExecutor for each transaction is intends
```
georgemitenkov, line 110: `is` --> `it`

aptos-move/aptos-vm/src/aptos_vm.rs
```diff
@@ -1701,6 +1749,41 @@ impl VMAdapter for AptosVM {
...
+                    }
+                },
+            },
+            Err(vm_status) => {
+                if !vm_status.is_aggregator_error() {
+                    return Err(vm_status);
+                }
+            },
+        }
+        match self.execute_single_transaction(txn, &view.as_move_resolver(), log_context, false) {
```
georgemitenkov, line 1776: What do you think about removing this match completely and instead having an assert in `execute_single_transaction`:
```
PreprocessedTransaction::UserTransaction(txn) => {
  ...
  let (vm_status, output) =
                    self.execute_user_transaction(resolver, txn, log_context, aggregator_enabled);
  assert!(aggregator_enabled || output.delta_change_set().is_empty());
  ...

```

aptos-move/aptos-vm/src/aptos_vm.rs
```diff
@@ -1701,6 +1749,41 @@ impl VMAdapter for AptosVM {
...
+        txn: &PreprocessedTransaction,
+        view: &impl StateView,
+        log_context: &AdapterLogSchema,
+    ) -> Result<(VMStatus, VMOutput, Option<String>), VMStatus> {
+        match self.execute_single_transaction(txn, &view.as_move_resolver(), log_context, true) {
+            Ok((vm_status, vm_output, sender)) => match vm_output.try_materialize(view) {
+                Ok(vm_output) => {
+                    return Ok((vm_status, vm_output, sender));
+                },
+                Err(vm_status) => {
```
georgemitenkov, line 1764: Seems like quite a bit of code is duplicated here? Maybe @zekun000 or @gelash know more Rusty ways of writing this

aptos-move/aptos-vm/src/aptos_vm.rs
vusirikala: I initially had `resolver` as input instead of `view`. But in the second line of the function, I needed `view` when computing `vm_output.try_materialize(view)`.

aptos-move/aptos-vm/src/block_executor/vm_wrapper.rs
```diff
@@ -53,29 +53,77 @@ impl<'a, S: 'a + StateView + Sync> ExecutorTask for AptosExecutorTask<'a, S> {
...
+    ) -> ExecutionStatus<AptosTransactionOutput, VMStatus> {
+        let log_context = AdapterLogSchema::new(self.base_view.id(), txn_idx as usize);
+        match self.vm.execute_single_transaction(
+            txn,
+            &view.as_move_resolver(),
+            &log_context,
+            aggregator_enabled,
+        ) {
+            Ok((vm_status, vm_output, sender)) => {
+                // If aggregators are disabled for user transactions, then delta_change_set should be empty as there should not be any delta outputs.
```
georgemitenkov, line 73: nit: split over multiple lines? I am not sure how many we use for our code (it seems quite random) but I usually use 80

aptos-move/aptos-vm/src/aptos_vm.rs
vusirikala: I agree. If anyone has ideas on reducing the duplication, I'm happy to change it.

aptos-move/aptos-vm/src/block_executor/vm_wrapper.rs
```diff
@@ -53,29 +53,77 @@ impl<'a, S: 'a + StateView + Sync> ExecutorTask for AptosExecutorTask<'a, S> {
...
+                assert!(
```
georgemitenkov, line 74: As above: I think we can shift the assert to `execute_single_transaction`?

aptos-move/aptos-vm/src/block_executor/vm_wrapper.rs
```diff
@@ -53,29 +53,77 @@ impl<'a, S: 'a + StateView + Sync> ExecutorTask for AptosExecutorTask<'a, S> {
...
+                    aggregator_enabled
+                        || !matches!(txn, PreprocessedTransaction::UserTransaction(_))
+                        || vm_output.delta_change_set().is_empty()
+                );
+                if vm_output.status().is_discarded() {
```
georgemitenkov, line 79: This is shared with `execute_transaction_sequential`. Let's refactor this out?

third_party/move/move-core/types/src/vm_status.rs
vusirikala: That was what I tried first. But the compiler didn't recognize `code` in `*code == EADD_OVERFLOW`.

georgemitenkov (changes requested):
Nice! Let's move VM statuses / aggregator check from Move repo?

third_party/move/move-core/types/src/vm_status.rs
vusirikala: My idea was that Aggregator extension file deals with `PartialVMError`. And in terms of hierarchy of errors that we have currently, `VMStatus` is a level about `PartialVMError`. I felt it would be a little confusing if we have code in aggregator_extension to deal with `VMStatus`.

third_party/move/move-core/types/src/vm_status.rs
georgemitenkov: I think `*code` should be binding to `code`?

aptos-move/aptos-vm/src/aptos_vm.rs
georgemitenkov: Maybe we can change materialization functions to use resolver instead of view as well ... No strong preference

aptos-move/block-executor/src/proptest_types/types.rs
vusirikala: `execute_transaction_sequential` is supposed to be called only for sequential execution. 
`execute_transaction_parallel` is supposed to be called only for parallel execution. 
If this is confusion, I can change it to `execute_transaction_during_sequential_execution`. Should I change?

aptos-move/aptos-vm/src/aptos_vm.rs
vusirikala: Will have to change the function signature of a series of functions and all their test cases along with it. I can do it if it's really that important.

aptos-move/aptos-vm/src/block_executor/vm_wrapper.rs
vusirikala: Done.

aptos-move/aptos-vm/src/block_executor/vm_wrapper.rs
vusirikala: Done.

aptos-move/aptos-vm/src/aptos_vm.rs
vusirikala: Done.

aptos-move/aptos-vm/src/block_executor/vm_wrapper.rs
vusirikala: Done.

aptos-move/aptos-vm/src/block_executor/vm_wrapper.rs
vusirikala: Done.

third_party/move/move-core/types/src/vm_status.rs
vusirikala: Done.

third_party/move/move-core/types/src/vm_status.rs
vusirikala: Removed it. I shifted the `is_aggregator_error()` check inline in `aptos_vm.rs`.

third_party/move/move-core/types/src/vm_status.rs
vusirikala: Removed it. I shifted the `is_aggregator_error()` check inline in `aptos_vm.rs`.

vusirikala:
> Nice! Let's move VM statuses / aggregator check from Move repo?

Done.

aptos-move/aptos-vm/src/adapter_common.rs
vusirikala: Done.

aptos-move/aptos-vm/src/block_executor/vm_wrapper.rs
vusirikala: Done.

aptos-move/block-executor/src/task.rs
vusirikala: Done.

aptos-move/aptos-vm/src/move_vm_ext/session.rs
vusirikala: Good point. I had to change the `ModificationWithMetadata` to `Modification` earlier when I ran sequential execution only with aggregators disabled. Now, we are enabling aggregators even for sequential execution. So, I reset the changes I made.

aptos-move/block-executor/src/proptest_types/types.rs
gelash: maybe let's add that as a comment

aptos-move/aptos-vm/src/block_executor/vm_wrapper.rs
vusirikala: I did push. Are you looking at the right commit version? Please take a look at the latest one (https://github.com/aptos-labs/aptos-core/pull/8334/files#diff-a8984b647508e3ad3ec3c2aee50ac6aff300d263fabf5b25d357fcbae2dc70c9).

aptos-move/block-executor/src/proptest_types/types.rs
vusirikala: Done.

aptos-move/block-executor/src/proptest_types/types.rs
```diff
@@ -467,6 +467,17 @@ where
...
         }
     }
+
+    fn execute_transaction_in_sequential_execution(
+        &self,
+        view: &impl TStateView<Key = K>,
+        txn: &Self::Txn,
+        txn_idx: TxnIndex,
+    ) -> ExecutionStatus<Self::Output, Self::Error> {
+        // As these functions are dummy, we reuse the code by calling
```
georgemitenkov, line 477: So this is just to satisfy the trait? If it is never called maybe having `unreachable!()` is better?

aptos-move/aptos-vm/src/aptos_vm.rs
```diff
@@ -1900,6 +1982,7 @@ impl AptosSimulationVM {
                         resolver,
                         log_context,
                         &storage_gas_params.change_set_configs,
+                        false,
```
georgemitenkov, line 1985: Why this is false?

aptos-move/aptos-vm/src/aptos_vm.rs
georgemitenkov: It's fine as is :)

aptos-move/aptos-vm/src/aptos_vm.rs
```diff
@@ -1730,6 +1781,34 @@ impl VMAdapter for AptosVM {
...
+                },
+                Err(vm_status) => {
+                    if matches!(vm_status, VMStatus::MoveAbort(_, code) if code == EADD_OVERFLOW || code == ESUB_UNDERFLOW)
+                    {
+                        return Err(vm_status);
+                    }
+                },
+            },
+            Err(vm_status) => {
+                if matches!(vm_status, VMStatus::MoveAbort(_, code) if code == EADD_OVERFLOW || code == ESUB_UNDERFLOW)
```
georgemitenkov, line 1804: This an above is wrong, it should be the negation of this

aptos-move/aptos-vm/src/aptos_vm.rs
```diff
@@ -1730,6 +1781,34 @@ impl VMAdapter for AptosVM {
...
+        view: &impl StateView,
+        log_context: &AdapterLogSchema,
+    ) -> Result<(VMStatus, VMOutput, Option<String>), VMStatus> {
+        match self.execute_single_transaction(txn, &view, log_context, true) {
+            Ok((vm_status, vm_output, sender)) => match vm_output.try_materialize(view) {
+                Ok(vm_output) => {
+                    return Ok((vm_status, vm_output, sender));
+                },
+                Err(vm_status) => {
+                    if matches!(vm_status, VMStatus::MoveAbort(_, code) if code == EADD_OVERFLOW || code == ESUB_UNDERFLOW)
```
georgemitenkov, line 1797: Let's move this to aptos-aggregator? You had a nice function before (`is_aggregator_error`) which is better to use then duplicate the code. This way you also do not have to make `EADD_OVERFLOW` and the other error public.

aptos-move/aptos-vm/src/aptos_vm.rs
georgemitenkov: Some ideas:
What about using map_err and map here? We can map non-aggregator-error to Some(error) and otherwise to None and have a single match instead.

aptos-move/block-executor/src/proptest_types/types.rs
vusirikala: This is in proptest folder. So, maybe this is used for testing.

aptos-move/aptos-vm/src/aptos_vm.rs
vusirikala: delta_change_set.rs uses PartialVMError when returning errors. It doesn't know about VMStatus, which is a couple of levels higher in the hierarchy we have right now. That's the reason, I earlier had `is_aggregator_error` function VMStatus implementation. Since that isn't an option, I felt it's better to have this code here.

aptos-move/aptos-vm/src/aptos_vm.rs
vusirikala: As this part of code is executed when a transaction fails, I disabled aggregators here.

aptos-move/aptos-vm/src/aptos_vm.rs
vusirikala: Good catch. Fixed it.

aptos-move/aptos-vm/src/aptos_vm.rs
georgemitenkov: Making internal error codes public and leaking what is an aggregator error to VM seems worse to me than using a public struct😄 What is the motivation behind delta_change_set.rs not knowing about VMStatus? It's just an import?

aptos-move/aptos-vm/src/aptos_vm.rs
```diff
@@ -1730,6 +1781,34 @@ impl VMAdapter for AptosVM {
...
+                },
+                Err(vm_status) => {
+                    if !matches!(vm_status, VMStatus::MoveAbort(_, code) if code == EADD_OVERFLOW || code == ESUB_UNDERFLOW)
+                    {
+                        return Err(vm_status);
+                    }
+                },
+            },
+            Err(vm_status) => {
+                if !matches!(vm_status, VMStatus::MoveAbort(_, code) if code == EADD_OVERFLOW || code == ESUB_UNDERFLOW)
```
georgemitenkov, line 1804: Can you please refactor this into a common function (you can have at least a lambda if you want to keep it here). It's used twice within the same code.
```
let is_aggregator_error = | ... | { ... };
```

aptos-move/aptos-vm/src/aptos_vm.rs
georgemitenkov: I see, thanks!

aptos-move/block-executor/src/proptest_types/types.rs
georgemitenkov: Can we know for sure how it is used? Having a good testing infra is important.

If it is used for testing it the comment is misleading then, it is not dummy. Otherwise you can call sequential and compare to parallel, assuming it tests your code but it does not.

If it is not used for testing (explicitly called) then having unreachable there is nicer in my opinion because it will panic in case it is used incorrectly?

vusirikala: Or I can just copy-paste the same code the for both _sequential and _parallel functions?

aptos-move/aptos-vm/src/aptos_vm.rs
vusirikala: Done.

aptos-move/aptos-vm/src/aptos_vm.rs
vusirikala: Tried it after your comment. Didn't simplify things. Now, we need to selectively return Ok(..) or Err(..) statements in the match statements. If there is an aggregator error, we need to re-execute transaction later on. It didn't simply the code when using map_err and map.

aptos-move/block-executor/src/proptest_types/types.rs
georgemitenkov: Why copy-paste? It is even worse than using a call to `..._parallel`. Why not `unreachable!("Some message")`?

aptos-move/aptos-vm/src/aptos_vm.rs
georgemitenkov: So what I meant was something like:
```
// returns none if it is an aggregator error, otherwise Some(VMStatus).
let clear_aggregator_error = |status: VMStatus| -> Option<VMStatus> { .. };
 
let result = self
  .execute_single_transaction(txn, &view, log_context, true)
  .map_err(|status| clear_aggregator_error(status))
  .map(|(status, output, sender)| {
    // If execution succeeds, try to apply deltas.
    output.try_materialize(view)
       .map(|output| (status, output, sender))
       .map_err(|status| clear_aggregator_error(status))
   });

// Now we just re-execute if Ok(Err(None)) or Err(None)
```
I think this is a more idiomatic Rust way (cc @gelash @zekun000 again), but if you think this is worse, let's keep it as is now. 

On the other note, if aggregator fails in ` self.execute_single_transaction(txn, &view, log_context, true)`, why do we need this re-execution? it can happen if you have an aggregator in Data state or do `aggregator::add(&agg, 2 * limit))` which clearly overflows. Maybe I am missing something?

georgemitenkov:
@vusirikala PR looks great, I think we can almost land it. I am thinking we should add tests here because the bug with `is_aggregator_error` instead of `!is_aggregator_error` seems quite serious and because aggregators do not fail, it can come unnoticed. What do you think? 

For example, we can have a dummy transaction with aggregator operation which errors, etc. and then test `execute_single_transaction` works correctly?

cc @zekun000 @gelash

vusirikala:
> @vusirikala PR looks great, I think we can almost land it. I am thinking we should add tests here because the bug with `is_aggregator_error` instead of `!is_aggregator_error` seems quite serious and because aggregators do not fail, it can come unnoticed. What do you think?
> 
> For example, we can have a dummy transaction with aggregator operation which errors, etc. and then test `execute_single_transaction` works correctly?
> 
> cc @zekun000 @gelash

Sure. I will think about some test cases. Please let me know if you have any tests in mind.

zekun000 (approved):

aptos-move/aptos-vm/src/adapter_common.rs
```diff
@@ -51,7 +52,21 @@ pub(crate) trait VMAdapter {
...
+        data_cache: &impl StateView,
+        log_context: &AdapterLogSchema,
+        aggregator_enabled: bool,
+    ) -> Result<(VMStatus, VMOutput, Option<String>), VMStatus>;
+
+    /// Execute a single transaction when running the block sequentially.
+    /// This function will first run the `execute_single_transaction` function
+    /// with aggregators enabled. If there is an aggregator error (overflow/underflow),
+    /// we are not sure if the error is triggered in the right location in the code.
+    /// Therefore, for proper error handlig, we will re-run the `execute_transaction`
```
gelash, line 64: typo: handling

aptos-move/block-executor/src/task.rs
```diff
@@ -58,12 +58,29 @@ pub trait ExecutorTask: Sync {
     fn init(args: Self::Argument) -> Self;
 
     /// Execute a single transaction given the view of the current state.
-    fn execute_transaction(
+    /// This function is only run during parallel execution.
+    /// BlockSTM will first call this function with aggregators enabled.
```
gelash, line 62: This part of the comment (from here below) is outdated, let's just add a TODO: parallel execution has to ensure to follow the same semantics w. aggregators (e.g. enable/disable) as it happens during the sequential execution.

gelash (changes requested):
I think the structure still isn't great, because it seems hard to test the proper fallback behavior with aggregator flag even for the sequential execution, which is what this PR is concerned with. I'd suggest the following below, and let's see if it's possible:

Instead of adding a new trait method to the VM/adapter itself, we can do the fallback within the executor.rs itself (inside execute_transactions_sequential), we'd only need to have a new type of error inside ExecutionStatus in task.rs for this, i.e. AggregatorError(E).

Inside the VM wrapper then we can again have a single method (instead of parallel and sequential), and merge back the processing logic, but add the condition that if the result is an aggregator error, and aggregator_enabled is true, then we return the new AggregatorError to the executor.

Why does this help?
(a) currently the PR has no tests, and this way we can at least test the fallback logic by using the mock VM. In particular, we can implement a new mock VM like in the proptest types that can return anything, so we can make it return an aggregator error, and make sure the executor re-runs it (because now the logic is within the executor).
(b) the flow is aligned with how it will be for parallel as well, and also will allow x-testing.

aptos-move/aptos-vm/src/adapter_common.rs
```diff
@@ -51,7 +52,21 @@ pub(crate) trait VMAdapter {
...
+    /// function with aggregators disabled and output the final output.
+    fn execute_single_transaction_sequential(
```
igor-aptos, line 66: I don't like _sequential and _parallel in the names of the functions, and needing to read comments to understand what they do.

It would be better if function explains what it does, instead of who calls it.

maybe:
execute_single_transaction => execute_single_transaction_one_pass
execute_single_transaction_sequential => execute_single_transaction_with_materialization

or any better names that you can think of :)

igor-aptos (commented):
is everything we do here backwards compatible?

aptos-move/aptos-vm/src/adapter_common.rs
vusirikala: Based on Rati's comments, we will move this logic to executor.rs.

aptos-move/aptos-vm/src/adapter_common.rs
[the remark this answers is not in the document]
vusirikala: Done.

aptos-move/aptos-vm/src/aptos_vm.rs
```diff
@@ -1772,6 +1826,26 @@ impl VMAdapter for AptosVM {
             },
         })
     }
+
+    fn execute_single_transaction_sequential(
```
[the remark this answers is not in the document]
vusirikala: It could live inside vm_wrapper as well. When you explained the logic to me initially, you asked me to implement it this way. I guess we are now going to have re-execution logic in exector.rs instead?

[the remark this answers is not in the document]
gelash: I thought initially it was in the wrapper, trait has different implications (for anything implementing that trait) that's why I was asking, but indeed this comment isn't relevant as we move the logic all the way to the executor (which makes sense for having materialization logic there and for testing)

gelash:
decided for a different solution.

PR #8342, opened 2023-05-23
[aptos-ledger] added fetching and update get public key method

0xjinn:
### Description

This PR
1. Added `pub fn fetch_batch_accounts()` to allow fetching batch accounts from Ledger
2. Updated the `get_public_key()` to return the Ed25519PublicKey instead

### Test Plan

Tested manually and make sure Ledger receive it .


crates/aptos-ledger/Cargo.toml
```diff
@@ -13,8 +13,10 @@ repository = "https://github.com/aptos-labs/aptos-core/tree/main/crates/aptos-le
 rust-version = { workspace = true }
 
 [dependencies]
-hex = "0.4.3"
+aptos-crypto = { workspace = true }
```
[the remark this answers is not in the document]
0xjinn: Good question!

I previously thought that if I set workspace link in this crate, it won't be able to publish to crates.io. But I was wrong lol

So I am switching it back to workspace version, to make sure everything are in sync

crates/aptos-ledger/src/lib.rs
```diff
@@ -143,17 +143,86 @@ pub fn get_app_name() -> Result<String, AptosLedgerError> {
...
 }
 
-/// Returns the public key of your Aptos account in Ledger device
+/// Returns the next available account in Ledger starting from start_index to end_index, inclusive
+/// Note: We only allow a range of 20 for performance purpose
+///
+/// # Arguments
+///
+/// * `start_index` - The index to start fetch for accounts
+/// * `end_index` - The last index to fetch the accounts
```
banool, line 152: Could you leave a comment about whether this is an inclusive or exclusive range?

crates/aptos-ledger/src/lib.rs
```diff
@@ -143,17 +143,86 @@ pub fn get_app_name() -> Result<String, AptosLedgerError> {
...
+pub fn fetch_batch_accounts(
+    start_index: u32,
+    end_index: u32,
```
banool, line 155: There is a native representation of this idea already, consider taking a [Range](https://doc.rust-lang.org/std/ops/struct.Range.html) instead.

crates/aptos-ledger/src/lib.rs
```diff
@@ -143,17 +143,86 @@ pub fn get_app_name() -> Result<String, AptosLedgerError> {
...
+    if start_index > end_index || end_index - start_index > 20 {
+        return Err(AptosLedgerError::UnexpectedError("Unexpected Error: Incorrect start and end index, make sure the range is less than or equal to 20".to_string(), None));
+    }
+
+    // Open connection to ledger
+    let transport = open_ledger_transport()?;
+
+    let mut accounts = HashMap::new();
+    for i in start_index..end_index {
+        let path = DERIVATIVE_PATH.replace("{index}", &i.to_string());
```
banool, line 167: Rather than an actual string replacement I'd recommend using a template string: https://doc.rust-lang.org/std/fmt/#named-parameters.

crates/aptos-ledger/src/lib.rs
```diff
@@ -143,17 +143,86 @@ pub fn get_app_name() -> Result<String, AptosLedgerError> {
...
+            data: cdata,
+        }) {
+            Ok(response) => {
+                // Got the response from ledger after user has confirmed on the ledger wallet
+                if response.retcode() == APDU_CODE_SUCCESS {
+                    // Extract the Public key from the response data
+                    let mut offset = 0;
+                    let response_buffer = response.data();
+                    let pub_key_len: usize = (response_buffer[offset] - 1).into();
+                    offset += 1;
```
banool, line 184: Could you explain this line?

crates/aptos-ledger/src/lib.rs
```diff
@@ -143,17 +143,86 @@ pub fn get_app_name() -> Result<String, AptosLedgerError> {
...
+
+                    // Skipping weird 0x04 - because of how the Aptos Ledger parse works when return pub key
+                    offset += 1;
+
+                    let pub_key_buffer = response_buffer[offset..offset + pub_key_len].to_vec();
+                    let hex_string = encode(pub_key_buffer);
+                    let public_key = match Ed25519PublicKey::from_encoded_string(&hex_string) {
+                        Ok(pk) => Ok(pk),
+                        Err(err) => Err(AptosLedgerError::UnexpectedError(err.to_string(), None)),
```
banool, line 193:
```suggestion
                        Err(err) => Err(AptosLedgerError::UnexpectedError(err.to_string(), Some(response.retcode()))),
```

banool (changes requested):
Looking good! Mostly requesting changes for the template string.

0xmaayan (approved):
LGTM

crates/aptos-ledger/src/lib.rs
0xjinn: Yeah good question. Basically we want to increment the offset or index, because index 0 is where is pk length is. Then the actual PK starts at the next index.

crates/aptos-ledger/src/lib.rs
```diff
@@ -203,12 +286,12 @@ pub fn get_public_key(display: bool) -> Result<String, AptosLedgerError> {
 /// # Arguments
 ///
 /// * `raw_txn` - the serialized raw transaction that need to be signed
-pub fn sign_txn(raw_txn: Vec<u8>) -> Result<Vec<u8>, AptosLedgerError> {
+pub fn sign_txn(raw_txn: Vec<u8>, path: &str) -> Result<Vec<u8>, AptosLedgerError> {
```
hardsetting, line 289: nit / personal preference: I would swap the order of args here, `path` first and `raw_txn` after

hardsetting (approved):
lgtm!

hardsetting:
As discussed offline, to make this a little more efficient we can do partial key derivation on the ledger, and then compute the public keys for each account on the client side (should be much faster).

i.e.
- compute the key and chain code for "m/44'/637" on the ledger and retrieve it using the ledger transport
- use the key and chain code to compute the remaining "{i}'/0'/0'" for each account in Rust (not sure if there's a library that supports that, we can look at it together)

for a follow-up PR, this works great for now :D

0xjinn:
@hardsetting Sounds good! Though I am not sure how partial key derivation works, but agree. This first version should be good enough, since we will most likely just fetch the top 0-9 indexed account for user to use in the CLI :)

crates/aptos-ledger/src/lib.rs
```diff
@@ -143,17 +144,96 @@ pub fn get_app_name() -> Result<String, AptosLedgerError> {
     }
 }
 
-/// Returns the public key of your Aptos account in Ledger device
+/// Returns the next available account in Ledger starting from start_index to end_index, inclusive
```
banool, line 147: Are we sure it is inclusive? I think when you iterate through a `Range` it is inclusive on the lower bound and exclusive on the upper bound: https://doc.rust-lang.org/std/ops/struct.Range.html.

crates/aptos-ledger/src/lib.rs
```diff
@@ -143,17 +144,96 @@ pub fn get_app_name() -> Result<String, AptosLedgerError> {
...
+/// Note: We only allow a range of 20 for performance purpose
+///
+/// # Arguments
+///
+/// * `index_range` - The range of the accounts in index, that you want to fetch, if None default to 0-20
```
banool, line 152: Clarify here too whether this is inclusive / exclusive.

crates/aptos-ledger/src/lib.rs
banool: Oh true, good point.

PR #8344, opened 2023-05-24
[Executor Logging] enable speculative logging

danielxiangzl:
### Description
Only disable speculative logging for sharded BlockSTM, to remove the speculative error loggings when running non-sharded BlockSTM (such as single-node-benchmark).


sitalkedia (approved):

gelash (approved):
thanks!

PR #8349, opened 2023-05-24
Change code example move/book/modules-and-scripts.md

dmshvetsov:
### Description

Change code example move/book/modules-and-scripts.md to clarify that it is not a test module that contains automated tests but an code example

### Test Plan

book/docs changes, I assume not tests needed


davidiw (approved):
good thinking

PR #8355, opened 2023-05-24
[Aptos Data Client] Split up tests into multiple files.

JoshLind:
### Description
This PR splits the large `test.rs` file in the Aptos data client into multiple separate files. This should make the test layout more readable/manageable. Nothing else should change 😸 

### Test Plan
Existing test infrastructure. Also, I ran all the tests before this change and counted that we ended up with the same number of tests after we split things up, so none were lost 😄


ASISBusiness (approved):

perryjrandall (approved):
mrclean

bchocho (approved):
LG, mostly blind accept :P

PR #8356, opened 2023-05-24
[TS SDK] Add generic to AptosClient.getAccountResource

BriungRi:
### Description

AptosClient.getAccountResource returns a `MoveResource` with the scheme

```
export type MoveResource = {
    type: MoveStructTag;
    data: MoveStructValue;
};
export type MoveStructValue = {
    };
```

This is almost always inconvenient because `data` is almost never `{}`.

This diff wraps `MoveResource` to `GenericMoveResource<TData>` so that users can provide a generic when calling `getAccountResource`.

### Additional thoughts

Ideally, since `MoveResource` is generated from the openapi Typescript generator, the generator should generate `MoveResource` so that it accepts a generic. I'm not sure this is necessarily possible while conforming to the openapi spec.

A less ideal solution would be to change the spec.yml to the following

```
    MoveResource:
      type: object
      description: A parsed Move resource
      required:
      - type
      - data
      properties:
        type:
          $ref: '#/components/schemas/MoveStructTag'
        data:
+         type: object
-         $ref: '#/components/schemas/MoveStructValue'
```

which /surprisingly/ results in the following generated code

```
export type MoveResource = {
    type: MoveStructTag;
    data: any;
};
```

which is actually significantly different from the original. This makes it slightly nicer to define the following interface without any casting

```
  @parseApiError
  async getAccountResource<TData = {}>(
    accountAddress: MaybeHexString,
    resourceType: Gen.MoveStructTag,
    query?: { ledgerVersion?: AnyNumber },
  ): Promise<
    Gen.MoveResource & {
      data: TData;
    }
  > {
    return this.client.accounts.getAccountResource(
      HexString.ensure(accountAddress).hex(),
      resourceType,
      query?.ledgerVersion?.toString(),
    );
  }
```

### Test Plan

```
pnpm test
```

```
Test Suites: 24 passed, 24 total
Tests:       1 skipped, 240 passed, 241 total
Snapshots:   0 total
Time:        39.16 s
Ran all test suites.
```


0xmaayan:
@BriungRi Thank you for the contribution! Can we also add test to show how one can use the function with a generic type? also, mind adding this change to the CHANGELOG?

PR #8366, opened 2023-05-25
[State Sync] Add latency monitor.

JoshLind:
Note: most of this PR is just tests.

### Description
This PR adds a new latency monitor to state sync, which tracks a number of important syncing latencies. This will help us to better understand, debug and improve overall end-to-end latencies in the system. The latencies now being tracked, are:
1. **Propose to sync**: The time taken from when a transaction was included in a proposed block (i.e., block timestamp), until it is eventually synced by the node.
2. **Propose to seen**: The time taken from when a transaction was included in a proposed block (i.e., block timestamp), until it is seen by the state sync data advertisements (but not necessarily synced).
3. **Seen to sync**: The time taken from when a transaction is seen by the state sync data advertisements, until it is synced.

### Test Plan
New and existing unit tests. I've also run this manually and verified that things look okay.

### DIFF
--- a/config/src/config/state_sync_config.rs
+++ b/config/src/config/state_sync_config.rs
@@ -236,6 +236,8 @@ impl Default for DataStreamingServiceConfig {
 #[derive(Clone, Copy, Debug, Deserialize, PartialEq, Eq, Serialize)]
 #[serde(default, deny_unknown_fields)]
 pub struct AptosDataClientConfig {
+    /// The interval (milliseconds) at which to refresh the latency monitor
+    pub latency_monitor_loop_interval_ms: u64,
     /// Maximum number of epoch ending ledger infos per chunk
     pub max_epoch_chunk_size: u64,
     /// Maximum number of in-flight polls for priority peers
@@ -265,6 +267,7 @@ pub struct AptosDataClientConfig {
 impl Default for AptosDataClientConfig {
     fn default() -> Self {
         Self {
+            latency_monitor_loop_interval_ms: 50, // 50 milliseconds
             max_epoch_chunk_size: MAX_EPOCH_CHUNK_SIZE,
             max_num_in_flight_priority_polls: 10,
             max_num_in_flight_regular_polls: 10,
--- a/state-sync/aptos-data-client/src/latency_monitor.rs
+++ b/state-sync/aptos-data-client/src/latency_monitor.rs
@@ -1,0 +1,495 @@
+// Copyright © Aptos Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::{
+    interface::AptosDataClientInterface,
+    logging::{LogEntry, LogEvent, LogSchema},
+    metrics,
+};
+use aptos_config::config::AptosDataClientConfig;
+use aptos_logger::{info, sample, sample::SampleRate, warn};
+use aptos_storage_interface::DbReader;
+use aptos_time_service::{TimeService, TimeServiceTrait};
+use futures::StreamExt;
+use std::{
+    collections::BTreeMap,
+    sync::Arc,
+    time::{Duration, Instant},
+};
+
+// Useful constants
+const LATENCY_MONITOR_LOG_FREQ_SECS: u64 = 5;
+const MAX_NUM_TRACKED_VERSION_ENTRIES: usize = 10_000;
+const MAX_VERSION_LAG_TO_TOLERATE: u64 = 10_000;
+
+/// A simple monitor that tracks the latencies taken to see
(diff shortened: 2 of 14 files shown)


ASISBusiness (approved):

state-sync/aptos-data-client/src/latency_monitor.rs
```diff
@@ -0,0 +1,498 @@
...
+        self.advertised_version_timestamps.write().insert(
+            highest_advertised_version,
+            (time_now_instant, timestamp_now_usecs),
+        );
+
+        // If the map is too large, garbage collect the old versions
+        while self.advertised_version_timestamps.read().len() > MAX_NUM_TRACKED_VERSION_ENTRIES {
+            // Remove the lowest version from the map by popping the first
+            // item. This is possible because BTreeMaps are sorted by key.
+            self.advertised_version_timestamps.write().pop_first();
```
bchocho, line 211: Can we just hold on to the previous write lock a little longer for the GC? It seems like getting the write lock, read lock, then write lock again just adds lock contention.

Given the current loop, maybe a really unlucky thread can starve because the size increases each time it's waiting for the write lock.

JoshLind: Aah, it's worth calling out that this is single threaded, i.e., there is only thread that runs the `latency_monitor` loop, so there shouldn't be any contention. 😄 In fact, I've removed the locks entirely. I originally added them so that the monitor could be cloneable (made it easier to write tests, etc.,) but I've worked around it now 😸

config/src/config/state_sync_config.rs
```diff
@@ -265,6 +267,7 @@ pub struct AptosDataClientConfig {
 impl Default for AptosDataClientConfig {
     fn default() -> Self {
         Self {
+            latency_monitor_loop_interval_ms: 50, // 50 milliseconds
```
bchocho, line 270: Is 50ms a reasonable number?

bchocho (approved):
LGTM, one question below

config/src/config/state_sync_config.rs
JoshLind: Good question. When I look at the data, I see latency variance across 100ms buckets (which makes me reluctant to run this loop less frequently, because it'll skew the results, e.g., if I run it only every 200ms, then we won't get data on fine-grained improvements). Likewise, I don't want to run it too frequently, because of resource costs.

Do you have a specific concern? One thing we could do is make this less frequent in mainnet (if we're worried about resource usage?)

perryjrandall: Can this be overridden in config?

JoshLind: Yeah, it can be. I can also add it to the config optimizer (so we have different defaults). But, I'm not sure there's much resource usage here at 50ms. Will wait for @bchocho to clarify the exact concern 😄

state-sync/aptos-data-client/src/latency_monitor.rs
```diff
@@ -0,0 +1,495 @@
...
+            metrics::observe_value_with_label(
+                &metrics::SYNC_LATENCIES,
+                metrics::SEEN_TO_SYNC_LATENCY_LABEL,
+                duration_from_seen_to_synced.as_secs_f64(),
+            );
+
+            // Update the proposal latencies
+            if let Ok(block_timestamp_usecs) = self.storage.get_block_timestamp(*synced_version) {
+                // Update the propose to seen latencies
+                if let Some(duration_from_propose_to_seen) =
```
perryjrandall, line 136: What happens here if we get Err / None do we just drop things?

state-sync/aptos-data-client/src/latency_monitor.rs
```diff
@@ -0,0 +1,495 @@
...
+            MAX_VERSION_LAG_TO_TOLERATE,
+        },
+        tests::mock::{create_mock_data_client, create_mock_db_reader},
+    };
+    use aptos_config::config::AptosDataClientConfig;
+    use aptos_time_service::{TimeService, TimeServiceTrait};
+    use std::time::{Duration, Instant};
+
+    #[test]
+    fn test_calculate_duration_from_proposal() {
```
perryjrandall, line 260: beautiful tests <3

perryjrandall (approved):

state-sync/aptos-data-client/src/latency_monitor.rs
JoshLind: Nothing should be dropped here 😄 If we don't hit these, we won't update the metrics, which I think is the safest behaviour.

PR #8368, opened 2023-05-25
[CLI][Rest] added default custom header

0xjinn:


crates/aptos-rest-client/src/lib.rs
```diff
@@ -59,6 +52,17 @@ static DEFAULT_INTERVAL_DURATION: Duration = Duration::from_millis(DEFAULT_INTER
 const DEFAULT_MAX_SERVER_LAG_WAIT_DURATION: Duration = Duration::from_secs(60);
 const RESOURCES_PER_CALL_PAGINATION: u64 = 9999;
 const MODULES_PER_CALL_PAGINATION: u64 = 1000;
+const X_APTOS_SDK_HEADER_VALUE: &str = concat!("aptos-sdk/", env!("CARGO_PKG_VERSION"));
+
+static NODE_URL: Lazy<Url> = Lazy::new(|| {
```
banool, line 57: We shouldn't have this default. More comments below.

crates/aptos-rest-client/src/lib.rs
```diff
@@ -69,46 +73,65 @@ pub struct Client {
...
+pub struct ClientBuilder {
+    reqwestBuilder: ReqwestClientBuilder,
+    version_path_base: String,
+    base_url: Url, // Default to mainnet if None
+    timeout: Option<Duration>, // Default to 10 seconds if None
+    headers: HeaderMap,
+}
+
+impl ClientBuilder {
+    pub fn new() -> Self {
```
banool, line 85: So with the builder pattern it is important to distinguish between:
1. Arguments that are required + have no sensible default.
1. Arguments that are optional + have no sensible default.
1. Arguments that are optional + have a sensible default.

Looking at all these arguments below:
- URL is 1
   - It is not possible to build a client without the user providing a URL. This means the builder should take the argument in to `new`.
- Timeout is 3
   - In `new` it makes sense to set a default, but we also allow people to override it.

As for `header` I think this is a helpful function (which would be 2), but we should also have a function on the builder for `X_APTOS_CLIENT`, like `x_aptos_client` (which would be 3). This way the caller doesn't have to know what header to use, that kinda thing. I'm less opinionated about this part though :P

crates/aptos-rest-client/src/lib.rs
```diff
@@ -69,46 +73,65 @@ pub struct Client {
...
-        let inner = ReqwestClient::builder()
-            .timeout(timeout)
-            .user_agent(user_agent)
-            .cookie_store(true)
-            .build()
-            .unwrap();
-
-        // If the user provided no version in the path, use the default. If the
-        // provided version has no trailing slash, add it, otherwise url.join
-        // will ignore the version path base.
```
banool, line 91: Can you include this as a docstring for the `get_version_path_with_base` function?

crates/aptos/src/common/types.rs
```diff
@@ -53,13 +50,18 @@ use std::{
...
+use crate::common::types::KeyType::X25519;
 
 pub const USER_AGENT: &str = concat!("aptos-cli/", env!("CARGO_PKG_VERSION"));
 const US_IN_SECS: u64 = 1_000_000;
 const ACCEPTED_CLOCK_SKEW_US: u64 = 5 * US_IN_SECS;
 pub const DEFAULT_EXPIRATION_SECS: u64 = 30;
 pub const DEFAULT_PROFILE: &str = "default";
 
+// Custom header to identify the client
+const X_APTOS_CLIENT_HEADER: &str = "x-aptos-client";
```
banool, line 62: Let's use the constant we've already got in the API types: https://github.com/aptos-labs/aptos-core/blob/main/api/types/src/headers.rs#L21.

crates/aptos/src/common/types.rs
```diff
@@ -902,11 +904,7 @@ impl RestOptions {
     }
 
     pub fn client(&self, profile: &ProfileOptions) -> CliTypedResult<Client> {
-        Ok(Client::new_with_timeout_and_user_agent(
-            self.url(profile)?,
-            Duration::from_secs(self.connection_timeout_secs),
-            USER_AGENT,
-        ))
+        Ok(Client::builder().timeout(Duration::from_secs(self.connection_timeout_secs)).base_url(self.url(profile)?).header(X_APTOS_CLIENT_HEADER, X_APTOS_CLIENT_VALUE).build())
```
banool, line 907: I think we might need to rustfmt this?

banool (changes requested):
Thanks for working on this!!

Left some feedback about the way we're doing the builder.

crates/aptos-rest-client/src/lib.rs
```diff
@@ -69,46 +73,65 @@ pub struct Client {
...
-                    format!("{}/", path)
-                } else {
-                    path.to_string()
-                }
-            },
-        };
+pub struct ClientBuilder {
+    reqwestBuilder: ReqwestClientBuilder,
+    version_path_base: String,
+    base_url: Url, // Default to mainnet if None
```
0xmaayan, line 79: let's default to devnet

sdk/src/coin_client.rs
```diff
@@ -21,6 +21,9 @@ use std::{
     str::FromStr,
     time::{SystemTime, UNIX_EPOCH},
 };
+use aptos_api_types::X_APTOS_CLIENT;
+
+const X_APTOS_SDK_HEADER_VALUE: &str = concat!("aptos-client-sdk-rust / ", env!("CARGO_PKG_VERSION"));
```
0xmaayan, line 26:
```suggestion
const X_APTOS_SDK_HEADER_VALUE: &str = concat!("aptos-rust-sdk/ ", env!("CARGO_PKG_VERSION"));
```

crates/aptos-rest-client/src/lib.rs
banool: Let's not default to anything, re my comment on the builder pattern below.

crates/aptos-rest-client/src/client_builder.rs
```diff
@@ -39,25 +39,25 @@ pub struct ClientBuilder {
     reqwest_builder: ReqwestClientBuilder,
     version_path_base: String,
     base_url: Url,
-    timeout: Option<Duration>, // Default to 10 seconds if None
+    timeout: Duration, // Default to 10 seconds if None
```
banool, line 42: nit: out of date comment

crates/aptos-rest-client/src/lib.rs
```diff
@@ -73,30 +73,11 @@ pub struct Client {
...
-            .unwrap();
-
-        let version_path_base = get_version_path_with_base(base_url.clone());
-
-        Self {
-            inner,
-            base_url,
-            version_path_base,
-        }
+        Self::builder(AptosBaseUrl::Custom(base_url)).build()
```
banool, line 80: Cleaaaaaaaannnnnn!!!!!!!!

Saying that, I think we still need to use `get_version_path_with_base` otherwise the URL might be missing `/v1`. But rather than putting it here let's do it in the `ClientBuilder`.

banool (changes requested):

crates/aptos-rest-client/src/client_builder.rs
0xjinn: nice catch!

crates/aptos-rest-client/src/lib.rs
banool: Oop Jin pointed out to me we are indeed doing this in the ClientBuilder!

banool (approved):
Awesome work!!

0xmaayan (approved):

PR #8372, opened 2023-05-25
[TS SDK] add `x-aptos-client` header to indexer queries

0xmaayan:
### Description
Adds `x-aptos-client` header to Indexer/Hasura requests
### Test Plan
Tested request locally
Screen Shot 2023-05-25 at 2 01 11 PM

### DIFF
--- a/ecosystem/typescript/sdk/src/providers/indexer.ts
+++ b/ecosystem/typescript/sdk/src/providers/indexer.ts
@@ -1,7 +1,7 @@
 import axios from "axios";
 
 import { AnyNumber } from "../bcs/types";
-import { HexString, MaybeHexString } from "../utils";
+import { HexString, MaybeHexString, getCustomRequestHeader } from "../utils";
 import {
   GetAccountTokensCountQuery,
   GetAccountCoinsDataQuery,
@@ -40,7 +40,6 @@ import {
   GetTokenOwnedFromCollection,
   GetCollectionData,
 } from "../indexer/generated/queries";
-import { getCustomRequestHeader } from "../utils";
 
 /**
  * Controls the number of results that are returned and the starting position of those results.
--- a/ecosystem/typescript/sdk/src/providers/indexer.ts
+++ b/ecosystem/typescript/sdk/src/providers/indexer.ts
@@ -40,6 +40,7 @@ import {
   GetTokenOwnedFromCollection,
   GetCollectionData,
 } from "../indexer/generated/queries";
+import { VERSION } from "../version";
 
 /**
  * Controls the number of results that are returned and the starting position of those results.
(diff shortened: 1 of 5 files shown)


ecosystem/typescript/sdk/src/providers/indexer.ts
```diff
@@ -92,7 +93,9 @@ export class IndexerClient {
    * @param graphqlQuery A GraphQL query to pass in the `data` axios call.
    */
   async queryIndexer<T>(graphqlQuery: GraphqlQuery): Promise<T> {
-    const { data } = await axios.post(this.endpoint, graphqlQuery);
+    const { data } = await axios.post(this.endpoint, graphqlQuery, {
+      headers: { "x-aptos-client": `aptos-ts-sdk/${VERSION}` },
```
banool, line 97: We're using this header key in two places now, let's make it a constant.

Similarly we're building the value of the header in two places, let's make it a function.

banool (changes requested):

banool (approved):

ecosystem/typescript/sdk/src/providers/indexer.ts
geekflyer: imo we can just make the header a global constant instead of function. .e.g. something like `export const commonHeaders = { .... }` and then just pass it to the client calls

ecosystem/typescript/sdk/src/utils/misc.ts
```diff
@@ -30,3 +32,7 @@ export const DEFAULT_TXN_EXP_SEC_FROM_NOW = 20;
 // How long does SDK wait for txhn to finish
 export const DEFAULT_TXN_TIMEOUT_SEC = 20;
 export const APTOS_COIN = "0x1::aptos_coin::AptosCoin";
+
+export function getCustomRequestHeader() {
+  return { "x-aptos-client": `aptos-ts-sdk/${VERSION}` };
```
geekflyer, line 37: let's make this a constant instead of function

geekflyer (approved):
let's make it a constant

ecosystem/typescript/sdk/src/utils/misc.ts
```diff
@@ -30,3 +32,5 @@ export const DEFAULT_TXN_EXP_SEC_FROM_NOW = 20;
 // How long does SDK wait for txhn to finish
 export const DEFAULT_TXN_TIMEOUT_SEC = 20;
 export const APTOS_COIN = "0x1::aptos_coin::AptosCoin";
+
+export const CUSTOM_REQUEST_HEADER = { "x-aptos-client": `aptos-ts-sdk/${VERSION}` };
```
geekflyer, line 36:
```suggestion
export const COMMON_REQUEST_HEADERS = { "x-aptos-client": `aptos-ts-sdk/${VERSION}` };
```

PR #8377, opened 2023-05-25
[Cli] add db restore to aptos cli

areshand:

### DIFF
--- a/crates/aptos/src/node/mod.rs
+++ b/crates/aptos/src/node/mod.rs
@@ -21,15 +21,15 @@ use crate::{
 };
 use aptos_backup_cli::{
     coordinators::restore::{RestoreCoordinator, RestoreCoordinatorOpt},
-    metadata::cache::MetadataCacheOpt,
-    storage::command_adapter::{config::CommandAdapterConfig, CommandAdapter},
-    utils::{ConcurrentDownloadsOpt, GlobalRestoreOpt, ReplayConcurrencyLevelOpt, RocksdbOpt},
+    storage::DBToolStorageOpt,
+    utils::GlobalRestoreOpt,
 };
 use aptos_cached_packages::aptos_stdlib;
 use aptos_config::config::NodeConfig;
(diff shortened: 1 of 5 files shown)


crates/aptos/src/node/mod.rs
```diff
@@ -1486,65 +1488,27 @@ impl CliCommand<()> for AnalyzeValidatorPerformance {
...
-        // hack: get around this error, related to use of `async_trait`:
-        //   error: higher-ranked lifetime error
-        //   ...
-        //   = note: could not prove for<'r, 's> Pin<Box<impl futures::Future<Output = std::result::Result<(), CliError>>>>: CoerceUnsized<Pin<Box<(dyn futures::Future<Output = std::result::Result<(), CliError>> + std::marker::Send + 's)>>>
+        let storage = self.storage.init_storage().await?;
         tokio::task::spawn_blocking(|| {
             let runtime = tokio::runtime::Runtime::new().unwrap();
-            runtime.block_on(RestoreCoordinator::new(opt, global_opt, storage).run())
+            runtime
+                .block_on(RestoreCoordinator::new(self.opt, self.global.try_into()?, storage).run())
```
banool, line 1511: I wonder if this is still necessary, vs just awaiting the inner function directly? If it _is_ still necessary, can you please leave that comment from before explaining the hack.

0xjinn:
Though this seems more internal to me, I wonder if we should update the changelog to reflect this? cc: @banool Thoughts?

msmouse (approved):
❤️  LGTM. Address others' concerns please.

crates/aptos/src/node/mod.rs
areshand: still necessary. Otherwise same compiler error with just inner function. added back the comment

areshand:
> changelog

updated the changelog.

0xjinn (approved):
LGTM.

Please ping me and let me know if this is urgent, I can start a release today if needed.

areshand:
> LGTM.
> 
> Please ping me and let me know if this is urgent, I can start a release today if needed.

will let you know once it lands, a release today would be timely needed. Thank you.

PR #7375, opened 2023-03-24
[storage] add cron backup compaction

areshand:
### Description

Currently, I did manual compaction for gcp and was testnet backup files.
Now, want to create a daily cron job to run manual compaction job in both gcp and aws backup fullnodes

### Test Plan
After the (1) v1.5 testnet release and  (2) terraform fix, the test will be
1. terraform apply the config file in this PR for testnet GCP cluster
2. check metadata folder to see if metadata files are compacted as expected, and check if the old metadata files are cleaned up from the metadata folder.

PR #8382, opened 2023-05-26
[helm][fullnode] add lookup for all workloads

rustielin:
### Description

So terraform apply doesn't blow away updates to these workloads through external means:
* restore job
* backup-verify cron
* backup stateful set

This has been done for validators, VFNs, and PFNs, but the `fullnode` chart also includes the above workfloads.

This unblocks some of the work @areshand need to do with updating the fullnode helm chart with compaction https://github.com/aptos-labs/aptos-core/pull/7375, as that will require a `helm upgrade`. We want to make sure that all workloads are unaffected by that.

### Test Plan

Verify locally using kind.
1. create a helm values file `kind_local_fullnode.yaml` specifying image tag A for all workloads (fullnode, backup, backup_verify, restore)
2. verify that all workloads come up with the right tag
3. change the tags in the values file, and also change `manageImages: false`. The result of this should be that subsequent `helm upgrade` should reuse the images already running, and not use the ones in the values file. ✅ 
4. flip `manageImages: true`, and subsequent `helm upgrade` uses the images specified in the values file ✅

### DIFF
--- a/terraform/helm/fullnode/templates/backup-verify.yaml
+++ b/terraform/helm/fullnode/templates/backup-verify.yaml
@@ -1,3 +1,4 @@
+{{ $backup_verify_cronjob := lookup "batch/v1" "CronJob" $.Release.Namespace (print (include "backup.fullname" .) "-backup-verify")}}
 apiVersion: batch/v1
 kind: CronJob
 metadata:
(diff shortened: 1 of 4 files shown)


perryjrandall (approved):

terraform/helm/fullnode/templates/backup-verify.yaml
```diff
@@ -23,8 +24,11 @@ spec:
           terminationGracePeriodSeconds: 0
           containers:
           - name: backup-verify
-            # use the same image with the backup sts
+            {{- if and $backup_verify_cronjob (not $.Values.manageImages) }} # if the statefulset already exists and we do not want helm to simply overwrite the image, use the existing image
```
perryjrandall, line 27: bless

sionescu (approved):

PR #8385, opened 2023-05-26
Single node benchmak assert no error log lines

igor-aptos:


runtian-zhou (approved):
Thank you!

sherry-x (approved):
So beautiful!!

msmouse (approved):

sitalkedia:
Do we have plans to do this on Forge as well?

PR #8391, opened 2023-05-26
Update data-pruning.md, call out default prune window

msmouse:
To call out the default pruning window and the fact that majority of the nodes don't have full history.


sherry-x (approved):

JoshLind (approved):
Thanks boss

PR #8401, opened 2023-05-26
[RFC] Add native function create_uuid

vusirikala:
### Description

Add native function `create_uuid` to `NativeTransactionContext`. This function will create a universally unique identifier by hashing `SessionId || uuid_counter`. These universally unique identifiers are helpful when assigning universally unique names for objects in Token V2 standard.


aptos-move/framework/src/natives/transaction_context.rs
```diff
@@ -34,6 +40,40 @@ impl NativeTransactionContext {
...
+    _args: VecDeque<Value>,
+) -> SafeNativeResult<SmallVec<[Value; 1]>> {
+    context.charge(gas_params.base)?;
+
+    let mut transaction_context = context
+        .extensions_mut()
+        .get_mut::<NativeTransactionContext>();
+    transaction_context.guid_counter += 1;
+    let mut hash_arg = Vec::new();
+    hash_arg.extend(transaction_context.txn_hash.clone());
```
igor-aptos, line 67: you need to append this as well:

  const OBJECT_FROM_TRANSACTION_GUID_ADDRESS_SCHEME: u8 = 0xFB;

to make it unique among the ways guids are created

igor-aptos (commented):
you'll need move changes to be able to test it end to end

check how other native functions are tested, and potentially add a test there - i.e. that calling this on different transactions or multiple times in the same transaction, it gives different results.

aptos-move/framework/src/natives/transaction_context.rs
vusirikala: Done.

davidiw (commented):
Cool stuff. Overall the code matches a lot of what I expected. Could we push this to an AIP quickly to discuss how we might migrate from the current account guid concept to the transaction guid?

Furthermore, it makes me wonder in what circumstances we might be blocked on really considering to use this. For example, events would still benefit from either aggregators or eliminating the sequence numbers all together. Objects could immediately make use of this. Though one could argue then the object could be further improved by eliminating the legacy GUID field from object core.

vusirikala:
> you'll need move changes to be able to test it end to end
> 
> check how other native functions are tested, and potentially add a test there - i.e. that calling this on different transactions or multiple times in the same transaction, it gives different results.

Added a unit test that creates 50 GUIDs in a transaction and verifies that all of them are unique.

davidiw:
Another thing to contemplate. We already have a framework for defining GUIDs as an address plus an integer. Would it make more sense to follow that paradigm rather than define a new paradigm and have to port everything over? I'm a little stuck on the hassle it would be to introduce a new GUID type.

davidiw (changes requested):
let's get alignment on what a proper next-gen guid is prior to landing this.

davidiw (dismissed):
let's get alignment on what a proper next-gen guid is prior to landing this.

aptos-move/framework/aptos-framework/sources/transaction_context.move
```diff
@@ -1,4 +1,7 @@
 module aptos_framework::transaction_context {
+    /// Return a globally unique identifier 
+    public native fun create_guid(): address;
```
jolestar, line 3: Or provide `get_tx_hash()` and `guid_counter()` native functions, and implement the `create_guid()` in Move

davidiw: If we go that route, then we can explore multiple solutions and leave it more to the user to dictate their preferred approach.

geekflyer:
if we introduce a new API I would use thuds as opportunity to call this UUID (instead of GUID), which is the much more common term used in other programming languages and libs.

aptos-move/framework/aptos-framework/sources/transaction_context.move
vusirikala: Isn't it generally preferable to provide as simple interface as possible, and reduce the cognitive load on the smart contract developer? In terms of functionality, the developer would achieve the same with both the options, but the developer has to put more effort into how to construct GUIDs if we expose `get_tx_hash` and `guid_counter` without an additional benefit.

jolestar: The `get_tx_hash` is helpful in many scenarios, `guid_counter` may be just for the `create_guid`. And I think the `new_table_handle` is repeated with `create_guid`; if we do not reuse the `guid_counter`, it will produce the repeat `table_handle` and `guid`.

vusirikala:
> Another thing to contemplate. We already have a framework for defining GUIDs as an address plus an integer. Would it make more sense to follow that paradigm rather than define a new paradigm and have to port everything over? I'm a little stuck on the hassle it would be to introduce a new GUID type.

I have made a few changes in object.move to utilize the new `create_guid` method. What do you think?

alinush:
Oh, btw, please make an effort to ensure that GUID hashes are domain separated from everything else in Rust and Move. See "Current domain-separation for hashing and signatures" in Notion.

movekevin (commented):
This looks okay to me high-level wise. Is there a good way to gate this new flow with a feature flag?

aptos-move/framework/aptos-framework/sources/transaction_context.move
```diff
@@ -1,4 +1,7 @@
 module aptos_framework::transaction_context {
+    /// Return a universally unique identifier
+    public native fun create_uuid(): address;
```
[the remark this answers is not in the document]
vgao1996: The native function needs to be feature-gated.

aptos-move/framework/src/natives/transaction_context.rs
```diff
@@ -34,6 +40,43 @@ impl NativeTransactionContext {
...
+    pub base: InternalGas,
+}
+
+fn native_create_uuid(
+    gas_params: &CreateUUIDGasParameters,
+    context: &mut SafeNativeContext,
+    mut _ty_args: Vec<Type>,
+    _args: VecDeque<Value>,
+) -> SafeNativeResult<SmallVec<[Value; 1]>> {
+    context.charge(gas_params.base)?;
```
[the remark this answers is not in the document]
alinush: You could, sure! Just pass those SHA256 gas parameter along as an extra parameter when the native is created. We have an example of this in the `algebra.move` module.

aptos-move/framework/src/natives/transaction_context.rs
```diff
@@ -34,6 +40,43 @@ impl NativeTransactionContext {
...
+
+    let mut transaction_context = context
+        .extensions_mut()
+        .get_mut::<NativeTransactionContext>();
+    transaction_context.uuid_counter += 1;
+    const OBJECT_FROM_TRANSACTION_UUID_ADDRESS_SCHEME: u8 = 0xFB;
```
alinush, line 66: This domain separator **MUST NOT** be a dangling hardcoded constant here. Otherwise, it will be very difficult to account for when adding new things later on that need to be domain separated.

What are you domain-separating from here? The other address derivation methods from [here](https://github.com/aptos-labs/aptos-core/blob/main/types/src/transaction/authenticator.rs#L241)?

aptos-move/framework/aptos-framework/sources/object.move
```diff
@@ -210,6 +210,12 @@ module aptos_framework::object {
         create_object_internal(creator_address, obj_addr, false)
     }
 
+    /// Create a new object by generating a random UUID based on transaction hash
+    public fun create_object(creator: &signer): ConstructorRef {
+        let uuid = aptos_framework::transaction_context::create_uuid();
+        create_object_internal(signer::address_of(creator), uuid, true)
```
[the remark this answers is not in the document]
igor-aptos: it seems like true is used whenever unique (non-repeatable) object address is used (i.e. in create_object_from_guid), and false otherwise.

if we want to allow users to specify if object can be deleted or not, that can be made a function argument, but that doesn't seem to be the current logic?

aptos-move/framework/aptos-framework/sources/transaction_context.move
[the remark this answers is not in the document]
igor-aptos: @movekevin get_txn_hash() is generally useful, should we add it to the same PR, or separate?

aptos-move/framework/src/natives/transaction_context.rs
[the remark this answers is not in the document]
igor-aptos: since uuid's generated here are fully unique, we don't need to namespace them for each usage. 

but if we have multiple namespace-ing already, maybe we need a way to namespace it from all other namespaces? @movekevin  Is that what you meant here?

igor-aptos: @alinush - it would be great to "domain separate it from everything that uses txn_hash", so it can be generally used (instead of having generate_uuid_for_object/generate_uuid_for_account/generate_uuid_for_x)


for this PR it needs to be separated from these : https://github.com/aptos-labs/aptos-core/blob/d65effee8c96b61e2e824a3865e091e6ac220adf/aptos-move/framework/aptos-framework/sources/object.move#L65

aptos-move/framework/aptos-framework/sources/transaction_context.move
igor-aptos: @jolestar - more details on the choice : https://www.notion.so/aptoslabs/Move-Parallelism-for-MetaPixel-982afa7c10ca494fad235fd1aa4318f4?pvs=4#d87bb42cc7a648d39fa3e699a4fffd35

guid_counter is scoped separately, while table_handle isn't

aptos-move/framework/aptos-framework/sources/object.move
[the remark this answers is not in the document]
igor-aptos: @vusirikala  - let's make all functions except create_named_object and create_user_derived_object to use it (i.e. remove all uses of create_object_from_guid here, and remove that function.

aptos-move/framework/src/natives/transaction_context.rs
alinush: That's just this [constant from here](https://github.com/aptos-labs/aptos-core/blob/main/types/src/transaction/authenticator.rs#LL250C5-L250C32) hardcoded in Move because there's no other way to export it.

This means that @vusirikala should move his `OBJECT_FROM_TRANSACTION_UUID_ADDRESS_SCHEME` constant into the [`Scheme` struct from that file](https://github.com/aptos-labs/aptos-core/blob/main/types/src/transaction/authenticator.rs#L241) and read it from there.

This will be easy to track.

aptos-move/aptos-gas/src/aptos_framework.rs
```diff
@@ -164,6 +164,8 @@ crate::natives::define_gas_parameters_for_natives!(GasParameters, "aptos_framewo
     [.util.from_bytes.per_byte, "util.from_bytes.per_byte", 5 * MUL],
 
     [.transaction_context.get_script_hash.base, "transaction_context.get_script_hash.base", 200 * MUL],
+    // Using SHA2-256's cost
+    [.transaction_context.create_uuid.base, "transaction_context.create_uuid.base", 3000 * MUL],
```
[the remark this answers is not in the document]
alinush: Why replicate the gas parameters, when we can use existing ones? 

See an example here for how we use existing SHA256 gas parameters in one of our `crypto_algebra` natives.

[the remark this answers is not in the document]
movekevin: Note - let's use gas schedule version 10 here since we'll be using 9 for the upcoming v1.5 mainnet release

aptos-move/framework/aptos-framework/sources/object.move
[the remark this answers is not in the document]
movekevin: My point is to add documentation :) This type of object being deletable makes sense to me if we can guarantee the same address can never be regenerated with future txs.

aptos-move/framework/src/natives/transaction_context.rs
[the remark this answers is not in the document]
movekevin: This would be a cleaner solution and would also address Alin's comment on the redundant gas schedule entry

aptos-move/framework/src/natives/transaction_context.rs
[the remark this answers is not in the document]
movekevin: We should also keep namespaces separate for different use cases. If the uuid generated from tx hash can be used in general use cases beyond object creation, we should also make sure that object creation uses its own domain separation. Otherwise, if in the future we ever support creating a different type of objects or entities (e.g. accounts, etc.) using these uuids, there would be conflicts.

cc @alinush as well regarding general domain separation best practice

aptos-move/framework/aptos-framework/sources/transaction_context.move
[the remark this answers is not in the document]
movekevin: Same PR as I don't think it's much more code to separate that out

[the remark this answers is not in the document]
vusirikala: Done.

aptos-move/framework/src/natives/transaction_context.rs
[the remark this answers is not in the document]
igor-aptos: I am not sure I followed here

are you saying that native function create_uuid should take in domain/or we should have multiple create_uuid functions (create_uuid_for_object, create_uuid_for_account, etc) ?

All of that seems unnecessary. Code makes sure that create_uuid calls don't conflict with one another (i.e. if you call create_uuid multiple times in a transaction, from different places, you are going to get a different uuid back). as long as the place we plan to use it for didn't already use conflicting (i.e. identical namespace) there before, we should be good?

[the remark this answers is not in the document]
movekevin: There can code that uses the returned uuid directly as an address. In general, I think it's more cautious to always both have built in domain separation to prevent any possible conflicts with addresses generated from other places (e.g. resource account, other object creation functions) and also introduce namespacing in important paths where entities such as accounts or objects are created. Specifically:
1. Add a generic suffix here in uuid creation to make sure it never conflicts with any other way of generating addresses (e.g. resource account, object, etc.)
2. Add a suffix in object creation using the uuid. This prevents any conflicts with other functions that could reuse that same address to create accounts, for example.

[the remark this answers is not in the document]
igor-aptos: @movekevin - I still don't get it :D 

can you explain how this can make object generation and account creation conflict if they use this create_uuid utility?

aptos-move/framework/aptos-framework/sources/object.move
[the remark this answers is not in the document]
vusirikala: I made changes to create_object_from_object, create_object_from_account, and removed create_object_from_guid. I added documentation that the objects created from create_object can be deleted because the created uuids can never be regenerated with future txs.

aptos-move/aptos-gas/src/aptos_framework.rs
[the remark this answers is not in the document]
vusirikala: Done.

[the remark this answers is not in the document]
vgao1996: You should keep LATEST_GAS_FEATURE_VERSION at 9 since it has already been bumped since the last release.

aptos-move/framework/src/natives/transaction_context.rs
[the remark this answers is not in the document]
vusirikala: I talked to @vgao1996 today about both the options. He said "Both would work. Although I'd personally prefer the new gas parameter approach to decouple create_uuid from sha256."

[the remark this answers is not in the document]
vgao1996: Yeah, my concern was is `create_uuid` and `sha256` are conceptually separate things, and reusing the sha256 gas parameters for create_uuid seems like a leaky abstraction to me, as implies the `create_uuid` is implemented using sha256 under the hood. Not totally sure if this is possible at all, but what if we change the hashing algorithm in the future?

aptos-move/framework/aptos-framework/sources/transaction_context.move
```diff
@@ -1,4 +1,10 @@
 module aptos_framework::transaction_context {
+    /// Return the transaction hash of the current transaction
+    public native fun get_txn_hash(): vector<u8>;
+
+    /// Return a universally unique identifier
```
movekevin, line 5: Can you add more documentation here that this uuid is generated based on the txn hash and that multiple uuids can be created in the same tx

aptos-move/framework/src/natives/transaction_context.rs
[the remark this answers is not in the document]
movekevin: Can we separate the gas charging in sha256 into a common function that this can call? This is effectively doing a sha256 of the tx hash. If we ever need to change the cost of sha256 in the future, we'd need to remember there are two places to change if we duplicate the cost formula here

aptos-move/framework/src/natives/transaction_context.rs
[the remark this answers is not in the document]
movekevin: Anyone can create an account at a given address with aptos_account::create_account which does not add any extra domain separator. So if the domain separation is done inside create_guid and object::create_object doesn't add any domain separator, you can actually create an account and an object at the same address. Therefore, we should at least move the domain separation into the object creation function. We can additionally add another domain separator in create_guid, but that could be a bit paranoid as it'd be hard to get the same hex from other forms of generation (cc @alinush on that thought though)

aptos-move/framework/aptos-framework/sources/object.move
```diff
@@ -210,23 +209,24 @@ module aptos_framework::object {
...
-    /// Create a new object from a GUID generated by an account.
-    public fun create_object_from_account(creator: &signer): ConstructorRef {
-        let guid = account::create_guid(creator);
-        create_object_from_guid(signer::address_of(creator), guid)
+    /// Create a new object by generating a random UUID based on transaction hash.
+    /// The created object is deletable as we can guarantee the same UUID can 
+    /// never be regenerated with future txs.
+    public fun create_object(creator: &signer): ConstructorRef {
+        let uuid = aptos_framework::transaction_context::create_uuid();
+        create_object_internal(signer::address_of(creator), uuid, true)
```
movekevin, line 217: Please add a domain separation here (as a suffix to the generated uuid)

aptos-move/framework/aptos-framework/sources/transaction_context.move
vusirikala: Done.

aptos-move/framework/aptos-framework/sources/object.move
vusirikala: UUID has an address type (32 bytes), should instead of appending to UUID, can I replace the last byte with `0xFA`?

aptos-move/framework/src/natives/transaction_context.rs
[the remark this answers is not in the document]
alinush: @vgao1996, I am not sure I follow. Since the `create_uuid` calls an operation that does SHA2-256 it seems perfectly reasonable to me (and more maintainable) to pass the SHA2-256 gas parameters in to that native, as opposed to duplicating their costs.

If we change the hashing algorithm in the future, we can update the native to take those new, potentially-different gas parameters as input, no?

aptos-move/framework/aptos-framework/sources/object.move
```diff
@@ -210,23 +209,24 @@ module aptos_framework::object {
...
+    public fun create_object_from_account(creator: &signer): ConstructorRef {
+        create_object(creator)
     }
 
-    fun create_object_from_guid(creator_address: address, guid: guid::GUID): ConstructorRef {
-        let bytes = bcs::to_bytes(&guid);
-        vector::push_back(&mut bytes, OBJECT_FROM_GUID_ADDRESS_SCHEME);
-        let obj_addr = from_bcs::to_address(hash::sha3_256(bytes));
-        create_object_internal(creator_address, obj_addr, true)
+    /// Deprectated.
```
alinush, line 226: Typo: Should be "Deprecated"

alinush: Also, how come **both** `create_object_from_account` and `create_object_from_object` are marked as "deprecated"?

Did you mean to only mark one of them?

types/src/transaction/authenticator.rs
```diff
@@ -461,6 +470,13 @@ impl AuthenticationKeyPreimage {
         Self::new(public_key.to_bytes(), Scheme::MultiEd25519)
     }
```
alinush, line 472: Add comments:
```
/// Construct a preimage from a transaction-derived UUID as (txn_hash || uuid_scheme_id)
```

aptos-move/framework/aptos-framework/sources/object.move
alinush: Or have they both been deprecated by `create_object`.

If so, please update comments to indicate which function they are deprecated by (i.e., which function should be called instead)

aptos-move/framework/src/natives/transaction_context.rs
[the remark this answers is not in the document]
alinush: Well, I defer to you if you think *reusing* gas parameters might be a good idea. My first instinct is that duplicating gas parameters is likely to lead to forgetting to update them in all the places. Which is why I think we should re-use them....

types/src/transaction/authenticator.rs
```diff
@@ -461,6 +470,13 @@ impl AuthenticationKeyPreimage {
...
 
+    pub fn uuid(transaction_derived_uuid: TransactionDerivedUUID) -> AuthenticationKeyPreimage {
+        let mut hash_arg = Vec::new();
+        hash_arg.push(Scheme::DeriveUuid as u8);
+        hash_arg.extend(transaction_derived_uuid.hash().to_vec());
```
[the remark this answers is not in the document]
alinush: Here, you do not want to hash `hash_arg`. Just return `Self(hash_arg)`. The hashing will happen by calling `AuthenticationKey::from_preimage()` over this returned `AuthenticationKeyPreimage` in the right place. (i.e., in your native).

aptos-move/framework/src/natives/transaction_context.rs
```diff
@@ -34,6 +42,68 @@ impl NativeTransactionContext {
...
+
+    let mut transaction_context = context
+        .extensions_mut()
+        .get_mut::<NativeTransactionContext>();
+    transaction_context.uuid_counter += 1;
+
+    let hash_vec = AuthenticationKeyPreimage::uuid(TransactionDerivedUUID {
+        txn_hash: transaction_context.txn_hash.clone(),
+        uuid_counter: transaction_context.uuid_counter,
+    });
```
alinush, line 98: Here, you want to call `AuthenticationKey::from_preimage` over the returned `AuthenticationKeyPreimage`.

aptos-move/framework/aptos-framework/sources/object.move
```diff
@@ -210,23 +209,24 @@ module aptos_framework::object {
...
-    public fun create_object_from_object(creator: &signer): ConstructorRef acquires ObjectCore {
-        let guid = create_guid(creator);
-        create_object_from_guid(signer::address_of(creator), guid)
+    /// Deprectated. 
+    /// Create a new object by generating a random UUID based on transaction hash.
+    public fun create_object_from_account(creator: &signer): ConstructorRef {
+        create_object(creator)
     }
 
-    fun create_object_from_guid(creator_address: address, guid: guid::GUID): ConstructorRef {
```
alinush, line 225: Can you update the comments in `ConstructorRef` after deleting this? They still refer to this function.

aptos-move/framework/src/natives/transaction_context.rs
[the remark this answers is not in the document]
igor-aptos: Is the point here that we should be charging here cost of SHA2-256 + extra ? is that possible to specify, and have only extra be configurable?

types/src/transaction/authenticator.rs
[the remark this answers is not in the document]
igor-aptos: @alinush  - should we be doing two hashes ? (i.e. transaction_derived_uuid.hash() and then AuthenticationKey::from_preimage on top)

should we instead of transaction_derived_uuid.hash()  use bcs::to_bytes(transaction_derived_uuid) ? 

are all other places that use AuthenticationKeyPreimage doing double hashing?

aptos-move/framework/src/natives/transaction_context.rs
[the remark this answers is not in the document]
vusirikala: @alinush With the updated code, then `UUID = Hash ( Hash(TransactionDerivedUUID) || 0xFB )`. So, there are two hashes right? So, we should be charging more for UUID generation?

types/src/transaction/authenticator.rs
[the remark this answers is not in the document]
vusirikala: @alinush Done.

types/src/transaction/authenticator.rs
vusirikala: Done.

aptos-move/framework/src/natives/transaction_context.rs
vusirikala: Done.

aptos-move/framework/aptos-framework/sources/object.move
vusirikala: Both the functions are deprecated by `create_object`. I will remove "deprecated" comment for now.

aptos-move/framework/aptos-framework/sources/object.move
igor-aptos: I don't think we need to do anything here, since we add Scheme::DeriveUuid already in native code.

aptos-move/framework/aptos-framework/sources/object.move
vusirikala: Done.

aptos-move/framework/aptos-framework/sources/object.move
alinush: Yep. AFAICT, no need to.

aptos-move/framework/src/natives/transaction_context.rs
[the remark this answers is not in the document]
alinush: Correct, but like @igor-aptos is saying below, the `CryptoHasher` domain separation is not actually necessary. This is true AFAICT. So we can just do `UUID = Hash ( Bytes(TransactionDerivedUUID) || 0xFB )`

types/src/transaction/authenticator.rs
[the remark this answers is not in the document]
alinush: Igor, I think so. We might've overdone it here a little, since I didn't originally understand this PR.

Be sure to remove the `CryptoHash` traits from the struct too.

aptos-move/framework/aptos-framework/sources/object.move
movekevin: They're not really deprecated as they offer alternatives for creating objects. We need to more deeply evaluate if they no longer have any useful purposes

aptos-move/framework/aptos-framework/sources/object.move
movekevin: It makes more sense IMO to have 0xFA here in the object code since all these 0xF something domain separations are object domain separators. Another benefits of this is to simplify the prediction or off-chain determination of tx-hash based uuids as they'd always just simply hash(txhash + creation index in tx)

igor-aptos: @movekevin this function will be used in other places/new applications as well. Having it separated from other schemes as well is useful (i.e. DeriveResourceAccountAddress and Ed25519 are non-object schemes)

And Scheme (from types/src/transaction/authenticator.rs) seems like a right place to track all the different schemes.
It feels like object.move is lacking, i.e. it has listed only subset of schemes, and can create conflict with Scheme, and should have a back-reference to Scheme enum.

But that's just my understanding, @alinush knows better how these were designed

types/src/transaction/authenticator.rs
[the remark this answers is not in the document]
vusirikala: I changed the implementation to use only one hash.

aptos-move/framework/aptos-framework/sources/object.move
alinush: Well, it's a bit of a hodge-podge unfortunately... Here's a short history lesson.

1. We had a domain separation bug a while ago between MultiEd25519 addresses and resource account addresses.
2. We as a result added a domain separator to resource account address derivation **in Move**.
3. But, in order to remember this domain separator exists, we duplicated it in in Rust in the `Scheme` struct in `types/src/transaction/authenticator.rs`
    a. i.e., we added `DeriveResourceAccountAddress` and you'll notice it's **never used in Rust**
4. Then, when Aptos Objects were proposed, some new address derivation code was introduced there too.
5. As a result, a new `DeriveObjectId` separator was added to `Scheme` too. Once again, this is **never used in Rust**; just kept track of.
6. Incidentally, this code had a small, [non-critical domain-separation bug](https://github.com/aptos-labs/aptos-core/pull/6481) too.
7. After fixing it, we replaced `DeriveObjectId` with two other domain separators in `Scheme`: `DeriveObjectIdFromGuid` and `DeriveObjectIdFromSeed`; also **never used in Rust**

Now, we are adding another domain separator there called `DeriveUuid`, which is actually used **in Rust** to derive a (domain-separated) address.

aptos-move/framework/aptos-framework/sources/object.move
```diff
@@ -210,23 +208,22 @@ module aptos_framework::object {
...
     }
 
-    /// Create a new object from a GUID generated by an account.
-    public fun create_object_from_account(creator: &signer): ConstructorRef {
-        let guid = account::create_guid(creator);
-        create_object_from_guid(signer::address_of(creator), guid)
+    /// Create a new object by generating a random UUID based on transaction hash.
+    /// The created object is deletable as we can guarantee the same UUID can 
+    /// never be regenerated with future txs.
+    public fun create_object(creator: &signer): ConstructorRef {
```
movekevin, line 214: I just realized passing the creator signer doesn't make sense. This should take an owner address instead, to whom the newly created owner will be transferred to

vusirikala: I changed the function to take `creator_address: address` as input.

aptos-move/framework/aptos-framework/sources/transaction_context.move
```diff
@@ -1,4 +1,29 @@
...
+
+    /// Return the transaction hash of the current transaction
+    public native fun get_txn_hash(): vector<u8>;
+
+    /// Return a universally unique identifier (of type address) generated
+    /// by hashing the transaction hash of this transaction and a sequence number 
+    /// specific to this transaction. This function can be called any
+    /// number of times inside a single transaction. Each such call increments
+    /// the sequence number and generates a new unique address
+    public native fun create_unique_address(): address;
```
igor-aptos: Add to the comment:

- uses Scheme in types/src/transaction/authenticator.rs for domain separation from other ways of generating unique addresses.

aptos-move/framework/aptos-framework/sources/object.move
```diff
@@ -108,8 +108,7 @@ module aptos_framework::object {
     /// This is a one time ability given to the creator to configure the object as necessary
     struct ConstructorRef has drop {
         self: address,
-        /// Set to true so long as deleting the object is possible. For example, the object was
-        /// created via create_object_from_guid.
+        /// Set to true so long as deleting the object is possible.
```
igor-aptos, line 111: add to 53rd line of this file:

// transaction_context::create_unique_address uses this for domain separation within it's native implementation
#[test_only]
const DERIVE_UUID_ADDRESS_SCHEME: u8 = 0xFB;


and add unit test that checks that aptos_framework::transaction_context::create_unique_address() returns the same as if you took 
        let guid = Guid(aptos_framework::transaction_context::get_txn_hash(), 0 /* because 0 is the first sequence number I assume */)
        let bytes = bcs::to_bytes(&guid);
        vector::push_back(&mut bytes, DERIVE_UUID_ADDRESS_SCHEME);
        let obj_addr = from_bcs::to_address(hash::sha3_256(bytes));


to make sure that serialization, hashing and domain separation works the same way and correctly

aptos-move/framework/aptos-framework/sources/object.move
```diff
@@ -210,12 +209,23 @@ module aptos_framework::object {
...
+    }
+
+    /// Deprecated. Use `create_object` instead.
     /// Create a new object from a GUID generated by an account.
     public fun create_object_from_account(creator: &signer): ConstructorRef {
         let guid = account::create_guid(creator);
         create_object_from_guid(signer::address_of(creator), guid)
     }
 
+    /// Deprecated. Use `create_object` instead.
```
movekevin, line 228: You can use #[deprecated] here

movekevin (approved):

aptos-move/e2e-move-tests/src/tests/transaction_context.data/sources/transaction_context_test.move
```diff
@@ -0,0 +1,25 @@
...
+    use std::vector;
+    use aptos_framework::transaction_context::create_uuid;
+    
+    /// When checking the value of aggregator fails.
+    const ENOT_UNIQUE: u64 = 20;
+
+    public entry fun create_many_uuids(_account: &signer, count: u64) {
+        let uuids: vector<address> = vector<address>[];
+        let i: u64 = 0;
+        while (i < count) {
```
gedigi, line 11: would this allocate `u64::MAX * 32`?

aptos-move/framework/aptos-framework/doc/transaction_context.md
```diff
@@ -5,14 +5,143 @@
...
+
+
+</details>
+
+<a name="0x1_transaction_context_create_unique_address"></a>
+
+## Function `create_unique_address`
+
+Return a universally unique identifier (of type address) generated
+by hashing the transaction hash of this transaction and a sequence number
```
gedigi, line 96: do we need some sort of domain separation?

aptos-move/framework/src/natives/transaction_context.rs
```diff
@@ -34,6 +42,68 @@ impl NativeTransactionContext {
...
+
+    let mut transaction_context = context
+        .extensions_mut()
+        .get_mut::<NativeTransactionContext>();
+    transaction_context.uuid_counter += 1;
+
+    let hash_vec = AuthenticationKey::from_preimage(&AuthenticationKeyPreimage::uuid(
+        transaction_context.txn_hash.clone(),
+        transaction_context.uuid_counter,
+    ));
```
gedigi, line 98: should this be domain separated as well?

aptos-move/framework/aptos-framework/sources/transaction_context.move
```diff
@@ -1,4 +1,43 @@
...
+        assert!(features::uuids_enabled(), EUUID_NOT_SUPPORTED);
+        create_unique_address()
+    }
+ 
     /// Return the script hash of the current entry function.
     public native fun get_script_hash(): vector<u8>;
+
+    /// This method runs `create_unique_address` native function and returns 
+    /// the generated unique address wrapped in the UUID class.
+    public fun create_uuid(): UUID {
```
movekevin, line 37: How is this going to be use? UUID is an internal type so other modules would not be able to see the inner type anyway.

vusirikala: David Wolinsky suggested this change. The idea is that, if we only expose `create_unique_address`,  then some malicious code could actually say that it is creating a unique address but instead use some address like 0x1. If we expose, `create_uuid`, then we know that UUIDs can be generated only by `create_uuid` method and are more confident that the address is generated uniquely and randomly. Does that make sense?

aptos-move/framework/aptos-framework/doc/transaction_context.md
vusirikala: We have domain separation. I will change this comment to reflect that.

aptos-move/framework/aptos-framework/sources/object.move
vusirikala: Done.

aptos-move/framework/aptos-framework/sources/object.move
vusirikala: I added the test case to `transaction_context.move` instead of `object.move`. Is that okay?

aptos-move/framework/aptos-framework/sources/transaction_context.move
vusirikala: Done.

aptos-move/framework/aptos-framework/sources/transaction_context.move
igor-aptos: We do need to have a function in this module that given UUID gives you the unique_address it wraps

aptos-move/framework/aptos-framework/sources/transaction_context.move
```diff
@@ -1,4 +1,58 @@
 module aptos_framework::transaction_context {
+
+    use std::features;
+
+    /// create_unique_address uses this for domain separation within its native implementation
+    #[test_only]
+    const DERIVE_UUID_ADDRESS_SCHEME: u8 = 0xFB;
```
igor-aptos, line 7: the point of having this and the test in object.move is to make sure that it is domain separated from other object methods correctly - and to make sure nobody adds a new domain separation that reuses 0xFB.

i.e. this constant needs to be next to OBJECT_DERIVED_SCHEME, OBJECT_FROM_GUID_ADDRESS_SCHEME etc.

vusirikala: Done.

aptos-move/framework/aptos-framework/sources/transaction_context.move
vusirikala: Done.

aptos-move/framework/aptos-framework/sources/transaction_context.move
```diff
@@ -1,4 +1,49 @@
...
+    /// This method runs `create_unique_address` native function and returns
+    /// the generated unique address wrapped in the AUID class.
+    public fun create_auid(): AUID {
+        assert!(features::auids_enabled(), EAUID_NOT_SUPPORTED);
+        return AUID {
+            unique_address: create_unique_address()
+        }
+    }
+
+    public fun get_unique_address(auid: AUID): address {
```
movekevin, line 46: This should take a reference

aptos-move/framework/aptos-framework/sources/transaction_context.move
```diff
@@ -1,4 +1,49 @@
 module aptos_framework::transaction_context {
+
+    use std::features;
+
+    /// AUID feature is not supported.
+    const EAUID_NOT_SUPPORTED: u64 = 3;
+
+    /// A wrapper denoting aptos unique identifer (AUID)
+    /// for storing an address
+    struct AUID has drop, store {
```
movekevin, line 10: Should this be copyable as well?

movekevin (changes requested):

aptos-move/framework/aptos-framework/sources/transaction_context.move
```diff
@@ -1,4 +1,49 @@
...
+    /// number of times inside a single transaction. Each such call increments
+    /// the sequence number and generates a new unique address.
+    /// Uses Scheme in types/src/transaction/authenticator.rs for domain separation
+    /// from other ways of generating unique addresses.
+    native fun create_unique_address(): address;
+
+    /// Return a universally unique identifier. Internally calls
+    /// the private function `create_unique_address`. This function is
+    /// created for to feature gate the `create_unique_address` function.
+    public fun create_unique_addr(): address {
```
movekevin, line 29: Nit: I’d name this generate_unique_address while naming the native version generate_unique_address_internal

aptos-move/framework/aptos-framework/sources/transaction_context.move
```diff
@@ -1,4 +1,49 @@
 module aptos_framework::transaction_context {
+
+    use std::features;
+
+    /// AUID feature is not supported.
+    const EAUID_NOT_SUPPORTED: u64 = 3;
```
movekevin, line 6: Why is this starting at 3? Shouldnt this be 1?

aptos-move/framework/aptos-framework/sources/object.move
```diff
@@ -47,6 +48,10 @@ module aptos_framework::object {
     /// Explicitly separate the GUID space between Object and Account to prevent accidental overlap.
     const INIT_GUID_CREATION_NUM: u64 = 0x4000000000000;
 
+    #[test_only]
```
movekevin, line 51: Id remove the test only annotation as its not publicly accessible anyway. And its nice to see all these domain separators related to object creation in the same place

aptos-move/e2e-move-tests/src/tests/transaction_context.rs
```diff
@@ -0,0 +1,22 @@
...
+    transaction_context::{create_many_uuids, initialize},
+    MoveHarness,
+};
+use aptos_language_e2e_tests::account::Account;
+
+fn setup() -> (MoveHarness, Account) {
+    initialize(common::test_dir_path("transaction_context.data"))
+}
+
+#[test]
```
movekevin, line 15: Can you add a test where an object is created and check that you can fetch resources there?

types/Cargo.toml
```diff
@@ -36,6 +36,7 @@ serde = { workspace = true }
 serde_bytes = { workspace = true }
 serde_json = { workspace = true }
 serde_yaml = { workspace = true }
+sha2 = { workspace = true }
```
movekevin, line 39: Is this needed? We should have already had all the deps needed for sha3256

aptos-move/framework/aptos-framework/sources/transaction_context.move
vusirikala: GUID correctly has only drop, store. So, I used the capabilities here. I added copy now.

aptos-move/framework/aptos-framework/sources/transaction_context.move
vusirikala: Done.

aptos-move/framework/aptos-framework/sources/transaction_context.move
vusirikala: I actually used 1 initially. After I changed the variable name a few types, I think the linter automatically bumped up the number. Changed it back to 1.

aptos-move/framework/aptos-framework/sources/object.move
vusirikala: Done.

types/Cargo.toml
vusirikala: I used sha2 earlier for the hash. Removing this dependency as we shifted to sha3256.

aptos-move/framework/aptos-framework/sources/transaction_context.move
vusirikala: I don't have a strong opinion here. If we make the change, we will have to update the AIP accordingly. 
@igor-aptos What do you think?

igor-aptos: Agreed, public function should have a nice name :)

aptos-move/framework/aptos-framework/sources/transaction_context.move
igor-aptos: It shouldn't be copyable.

Main point is if you have two instances of AUID, they are guaranteed to be different. 
copyable breaks that

aptos-move/framework/aptos-framework/sources/transaction_context.move
vusirikala: Changed the names.

aptos-move/e2e-move-tests/src/tests/transaction_context.rs
vusirikala: Still working on it. Not entirely sure how to interact with objects.

movekevin: You can use the same /resources API to fetch resources at an object address. As for how to get the object address, this would be a good exercise for you to understand how people would be able to get the created addresses from the newly introduced functions.

aptos-move/e2e-move-tests/src/tests/transaction_context.data/sources/transaction_context_test.move
```diff
@@ -0,0 +1,28 @@
...
+
+    /// When checking the value of aggregator fails.
+    const ENOT_UNIQUE: u64 = 20;
+
+    public entry fun create_many_auids(_account: &signer, count: u64) {
+        if (features::auids_enabled()) {
+            let auids: vector<address> = vector<address>[];
+            let i: u64 = 0;
+            while (i < count) {
+                i = i+1;
```
movekevin, line 14: nit: i+1 => i + 1

aptos-move/e2e-move-tests/src/tests/transaction_context.data/sources/transaction_context_test.move
```diff
@@ -0,0 +1,28 @@
...
+    use std::vector;
+    use aptos_framework::transaction_context::generate_unique_address;
+    use std::features;
+
+    /// When checking the value of aggregator fails.
+    const ENOT_UNIQUE: u64 = 20;
+
+    public entry fun create_many_auids(_account: &signer, count: u64) {
+        if (features::auids_enabled()) {
+            let auids: vector<address> = vector<address>[];
```
movekevin, line 11: nit: This can just be let auids = vector[];
The type can be inferred

aptos-move/e2e-move-tests/src/tests/transaction_context.data/sources/transaction_context_test.move
```diff
@@ -0,0 +1,28 @@
...
+            let i: u64 = 0;
+            while (i < count) {
+                i = i+1;
+                vector::push_back(&mut auids, generate_unique_address());
+            };
+            i = 0;
+            while (i < count - 1) {
```
movekevin, line 18: This validation looks like something that should be done inside a unit test for the transaction_context.move module, not here?

movekevin (approved):

aptos-move/e2e-move-tests/src/tests/transaction_context.data/sources/transaction_context_test.move
```diff
@@ -0,0 +1,28 @@
+module 0x1::transaction_context_test {
+    use std::vector;
+    use aptos_framework::transaction_context::generate_unique_address;
+    use std::features;
+
+    /// When checking the value of aggregator fails.
+    const ENOT_UNIQUE: u64 = 20;
+
+    public entry fun create_many_auids(_account: &signer, count: u64) {
```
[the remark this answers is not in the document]
vusirikala: Done.

aptos-move/e2e-move-tests/src/tests/transaction_context.data/sources/transaction_context_test.move
vusirikala: Tried it just now. Got this error. So, switching back. 
Screenshot 2023-06-26 at 10 59 41 AM

aptos-move/framework/aptos-framework/sources/object.move
```diff
@@ -108,8 +112,7 @@ module aptos_framework::object {
     /// This is a one time ability given to the creator to configure the object as necessary
     struct ConstructorRef has drop {
         self: address,
-        /// Set to true so long as deleting the object is possible. For example, the object was
-        /// created via create_object_from_guid.
+        /// Set to true so long as deleting the object is possible.
```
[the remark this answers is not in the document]
vusirikala: Done.

aptos-move/aptos-gas/src/aptos_framework.rs
```diff
@@ -163,7 +163,10 @@ crate::natives::define_gas_parameters_for_natives!(GasParameters, "aptos_framewo
     [.util.from_bytes.base, "util.from_bytes.base", 300 * MUL],
     [.util.from_bytes.per_byte, "util.from_bytes.per_byte", 5 * MUL],
 
+    [.transaction_context.get_txn_hash.base, { 9.. => "transaction_context.get_txn_hash.base" }, 200 * MUL],
```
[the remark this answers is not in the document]
vusirikala: Done.

aptos-move/e2e-move-tests/src/tests/transaction_context.data/sources/transaction_context_test.move
vusirikala: Moved the the test from e2e-move-tests to a unit test in transaction_context.move.

aptos-move/framework/aptos-framework/sources/object.move
```diff
@@ -210,12 +213,25 @@ module aptos_framework::object {
...
+    /// The unique address is computed sha3_256([transaction hash | auid counter | 0xFB]).
+    /// The created object is deletable as we can guarantee the same unique address can
+    /// never be regenerated with future txs.
+    public fun create_object(owner_address: address): ConstructorRef {
+        let unique_address = transaction_context::generate_unique_address();
+        create_object_internal(owner_address, unique_address, true)
+    }
+
+    #[deprecated]
+    /// Use `create_object` instead.
```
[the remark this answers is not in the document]
vusirikala: Done.

aptos-move/framework/aptos-framework/sources/configs/staking_config.spec.move
```diff
@@ -83,6 +83,7 @@ spec aptos_framework::staking_config {
     }
 
     spec calculate_and_save_latest_epoch_rewards_rate(): FixedPoint64 {
+        pragma verify_duration_estimate = 120;
```
igor-aptos, line 86: what is this for?

aptos-move/framework/aptos-framework/sources/object.move
```diff
@@ -47,6 +48,9 @@ module aptos_framework::object {
     /// Explicitly separate the GUID space between Object and Account to prevent accidental overlap.
     const INIT_GUID_CREATION_NUM: u64 = 0x4000000000000;
 
+    /// generate_unique_address uses this for domain separation within its native implementation
+    const DERIVE_AUID_ADDRESS_SCHEME: u8 = 0xFB;
```
igor-aptos, line 52: move it below MAXIMUM_OBJECT_NESTING, so it is right before 0xFC constant

aptos-move/framework/aptos-framework/sources/transaction_context.move
```diff
@@ -1,4 +1,75 @@
...
+    /// number of times inside a single transaction. Each such call increments
+    /// the sequence number and generates a new unique address.
+    /// Uses Scheme in types/src/transaction/authenticator.rs for domain separation
+    /// from other ways of generating unique addresses.
+    native fun generate_unique_address_internal(): address;
+
+    /// Return a universally unique identifier. Internally calls
+    /// the private function `generate_unique_address_internal`. This function is
+    /// created for to feature gate the `generate_unique_address_internal` function.
+    public fun generate_unique_address(): address {
```
igor-aptos, line 29: maybe we should rename this to generate_auid_address , to align with the rest?

not a big deal though.

aptos-move/framework/aptos-framework/sources/transaction_context.move
```diff
@@ -1,4 +1,75 @@
...
+    const EAUID_NOT_SUPPORTED: u64 = 1;
+
+    /// A wrapper denoting aptos unique identifer (AUID)
+    /// for storing an address
+    struct AUID has drop, store {
+        unique_address: address
+    }
+
+    /// Return the transaction hash of the current transaction
+    public native fun get_txn_hash(): vector<u8>;
```
igor-aptos, line 15: even though we are not using it, does this need to be gated?

igor-aptos (approved):

aptos-move/framework/aptos-framework/sources/configs/staking_config.spec.move
vusirikala: Yesterday, I got some move prover test errors in Github actions. I think Junkil added this to fix those errors.

aptos-move/framework/aptos-framework/sources/object.move
vusirikala: Done.

aptos-move/framework/aptos-framework/sources/transaction_context.move
vusirikala: Done.

aptos-move/framework/aptos-framework/sources/transaction_context.move
vusirikala: I couldn't really think of a reason to gate it.

PR #8402, opened 2023-05-26
[State Sync] Rename "subscription" to "optimistic fetch".

JoshLind:
### Description
This PR updates the "subscription" terminology used by the storage service in state sync to "optimistic fetch". This is because: (i) "optimistic fetch" better describes what is happening, i.e., a peer is optimistically fetching only one chunk of data, and not subscribing to it; and (ii) we want to explore "subscription" style requests in the future, thus it makes sense to avoid the confusion.

Aside from renaming, nothing else should change. The only "breakage" here is the renaming of two (unimportant) metrics. I'll update the dashboard when the times comes 😄 

### Test Plan
Existing test infrastructure.


bchocho (approved):

state-sync/storage-service/server/src/optimistic_fetch.rs
```diff
@@ -127,7 +127,7 @@ impl DataSubscriptionRequest {
             DataRequest::GetNewTransactionOutputsWithProof(request) => request.known_version,
             DataRequest::GetNewTransactionsWithProof(request) => request.known_version,
             DataRequest::GetNewTransactionsOrOutputsWithProof(request) => request.known_version,
-            request => unreachable!("Unexpected subscription request: {:?}", request),
+            request => unreachable!("Unexpected optimistic fetch request: {:?}", request),
```
perryjrandall, line 130: nit this string can be a const since its used a couple places

perryjrandall (approved):

state-sync/storage-service/server/src/optimistic_fetch.rs
JoshLind: Maybe... I think it's used 4 times? But the rust compiler is very very smart. I've given up trying to outsmart it 😄

PR #8403, opened 2023-05-26
[API] Fix request client regex

banool:
### Description
The current regex does not match `aptos-ts-sdk/1.9.1` because it only expects a single `-` in the part before the `/`. This was an oversight because I tested it initially with just `aptos-cli`.

### Test Plan
```
cargo run -p aptos -- node run-local-testnet --force-restart --assume-yes --with-faucet
```
```
curl -H 'x-aptos-client: aptos-ts-sdk/7.8.9' http://127.0.0.1:8080/v1
```
```
curl -s http://127.0.0.1:9101/metrics | grep -i request_source
```

Before:
```
aptos_api_request_source_client{operation_id="get_ledger_info",request_source_client="unknown",status="200"} 2
```

After:
```
aptos_api_request_source_client{operation_id="get_ledger_info",request_source_client="aptos-ts-sdk/7.8.9",status="200"} 1
```

### DIFF
--- a/api/src/log.rs
+++ b/api/src/log.rs
@@ -17,7 +17,7 @@ use std::time::Duration;
 
 const REQUEST_SOURCE_CLIENT_UNKNOWN: &str = "unknown";
 static REQUEST_SOURCE_CLIENT_REGEX: Lazy<Regex> =
-    Lazy::new(|| Regex::new(r"aptos-[a-zA-Z]+/[0-9A-Za-z\.\-]+").unwrap());
+    Lazy::new(|| Regex::new(r"aptos-[a-zA-Z\-]+/[0-9A-Za-z\.\-]+").unwrap());
 
 /// Logs information about the request and response if the response status code
 /// is >= 500, to help us debug since this will be an error on our side.


api/src/log.rs
```diff
@@ -17,7 +17,7 @@ use std::time::Duration;
 
 const REQUEST_SOURCE_CLIENT_UNKNOWN: &str = "unknown";
 static REQUEST_SOURCE_CLIENT_REGEX: Lazy<Regex> =
-    Lazy::new(|| Regex::new(r"aptos-[a-zA-Z]+/[0-9A-Za-z\.\-]+").unwrap());
+    Lazy::new(|| Regex::new(r"aptos-[a-zA-Z\-]+/[0-9A-Za-z\.\-]+").unwrap());
```
0xmaayan, line 20: does `\-` mean "any amount of `-`"?

0xmaayan (approved):

api/src/log.rs
banool: `[a-zA-Z\-]+` means "at least one of these chars", where "these chars" is a-z, A-Z, and -.

0xmaayan: I wonder if we should also consider digits. not important for now but who knows what names others would come up with. we ask for the format `<sdk-id>/<sdk-version>` where `<sdk-id>` can be anything

0xjinn (approved):

PR #8405, opened 2023-05-27
generate non-conflicting coin transfer block

zjma:


sitalkedia (commented):
@zjma - It seems like your change is on top of my previous branch, that I indent to discard. It is a bit confusing to review as your change doesn't reflect everything. Would you mind making this PR on `main` instead?

crates/transaction-generator-lib/src/p2p_transaction_generator.rs
```diff
@@ -91,6 +90,38 @@ impl P2PTransactionGenerator {
...
         }
     }
+
+    /// Generate a given number (`num_txns`) of transactions that satisfies some constraints.
+    pub fn generate_block(&mut self, num_txns: usize, transfer_amount: u64, no_rw_conflict: bool, num_txns_per_sender: usize) -> Vec<(SignedTransaction, AccountAddress)> {
+        let num_accounts = self.accounts().len();
+        if no_rw_conflict {
+            // We need `num_txns` distinct senders and `num_txns` distinct recipients.
+            assert!(num_accounts >= num_txns * 2);
+            let indices = rand::seq::index::sample(&mut self.rng, self.accounts.len(), num_txns * 2);
```
sitalkedia, line 100: I might be wrong here but IIUC, this doesn't guarantee uniqueness, it is possible that the random selection will yield the same index twice.

execution/executor-benchmark/src/transaction_generator.rs
```diff
@@ -440,49 +446,29 @@ impl TransactionGenerator {
         transactions_per_sender: usize,
         non_conflicting_txns: bool,
     ) {
```
sitalkedia, line 448: You need to make changes in the `run_workload` function instead to use the transaction generator.

sitalkedia (changes requested):

zjma:
> 

@sitalkedia I think it's easier if we update the transaction generation part and reuse your branch. Do you mind keeping your branch/PR?

crates/transaction-generator-lib/src/p2p_transaction_generator.rs
zjma: According to [the doc](https://docs.rs/rand/0.7.3/rand/seq/index/fn.sample.html) we don't need to worry. (I also tested a bit.)

sitalkedia:
> @sitalkedia I think it's easier if we update the transaction generation part and reuse your branch. Do you mind keeping your branch/PR?

@zjma - since I don't plan to land my PR, you can cherry-pick relevant changes and update this PR instead.

crates/transaction-generator-lib/src/p2p_transaction_generator.rs
```diff
@@ -7,35 +7,108 @@ use aptos_sdk::{
...
+    init_shuffle_done: bool
+}
+
+impl BurnAndRecycleSampler {
+    fn new(num_items: usize, segment_size: usize) -> Self {
+        Self {
+            item_pool: (0..num_items).collect(),
+            next_index: 0,
+            segment_size,
+            init_shuffle_done: false,
```
sitalkedia, line 50: Why do you need this? Can't you just shuffle the item_pool in the constructor?

zjma: you are right, it's probably more straightforward if we shuffle in the constructor... It's just the constructor then need to take an extra parameter `&mut StdRng` which seems a bit annoying to me.

crates/transaction-generator-lib/src/p2p_transaction_generator.rs
```diff
@@ -7,35 +7,108 @@ use aptos_sdk::{
...
+        }
+    }
+
+    fn sample_one(&mut self, rng: &mut StdRng) -> usize {
```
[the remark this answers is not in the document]
zjma: With `BasicSampler`, no matter how big the address pool is, conflicts are possible: the sampled address right before recycling and the one right after can be the same. (Technically you can avoid dealing with it if `num_recipients % num_signers == 0`, but that's not what we want to do here?)

Only by using 2+ sub-pools and each having a large enough size (>= block_size) can you avoid the potential conflicts. And with this change, you can achieve it by using a `P2PTransactionGenerator` with:
- a `BurnAndRecycleSampler` with `segment_size>=block_size`
- at least `2*segment_size` addresses

igor-aptos (commented):
looks much better!

One question, why do we use non_conflicting_txns_per_block flag, and override CoinTransfer transaction type within executor benchmark, instead of having a new NonConflictingCoinTransfer transaction type?

this way it can only be used in executor benchmark, but not in transaction emitter. but at some point we will want to try out sharding in a full system (with consensus and stuff), so that will be needed?

crates/transaction-generator-lib/src/p2p_transaction_generator.rs
```diff
@@ -7,35 +7,108 @@ use aptos_sdk::{
...
+        println!("s={s}");
+    }
+}
+
 
 pub struct P2PTransactionGenerator {
     rng: StdRng,
     send_amount: u64,
     txn_factory: TransactionFactory,
     all_addresses: Arc<RwLock<Vec<AccountAddress>>>,
```
igor-aptos, line 91: also sampling by index is generally fine, but code is made so that these Arc's are communication mechanisms between transaction generators.

i.e:
- we can have first phase of account creation that populates all_addresses, and second phase of coin transfers to consume it
- or you can have a mix of account creation that populate all_addresses, and p2p transfers that consumes it. 
- or you can first populate all_addresses from account creation, and then run a mix of p2p transfers and another transaction type that consumes it

so all_addresses can change

it might be a bit nicer with the rest of the system, to remove accounts from all_addresses, and then place them back later (instead of just operating based on index, initially set)

zjma: @igor-aptos which one sounds better, to change P2PTxnGenerator to hold the burnt address, or to augment the shared pool to have a burnt queue or something?

PR #8414, opened 2023-05-29
Comprehensive marketplace demo for tokenv1, tokenv2

davidiw:
This introduces the core for a potential Aptos standard around marketplace for assets on-chain. I base a lot of the design and patterns from https://github.com/aptos-labs/aptos-core/pull/8112 but with the intent of hammering more at the modularity goals given objects.

The goals of this package are to
* Separate core logical components for readability and expansion over time. Eliminate data from being the layer of compatibility and instead the exposed functions.
* Leverage of objects and resource groups to unify common logic without wasting storage.
* Single definition of a fee schedule for the marketplace where the listing was created.
* Unified framework for auctions and fixed-price listings.
* Support for TokenV1, TokenV2, and Object based assets.
* Support for receiving funds in either Coin or FungibleAsset.

FeeSchedule includes:
* Listing, bidding, and commission
* Clean interface that allows newer business logic to be added over time, by passing in current pricing information

All listings support:
* Ability to specify a fixed purchase price
* Define when purchasing may begin
* Embed a fee schedule for the hosting marketplace

Auctions support:
* Buy-it-now
* Incremental end times based upon the last bid time
* Minimum bid increments
Fixed-price support:
* Seller can end at any time.

TODO:
* Collection offers for Tokens.

This is intended as an exploration into the ideal marketplace framework. Please make pull requests to extend it and generalize our use cases. This may never actually be deployed on
Mainnet unless the community rallies behind a common marketplace and harness.


japko1980 (approved):

aptos-move/move-examples/marketplace/README.md
```diff
@@ -0,0 +1,32 @@
+This introduces the core for a potential Aptos standard around marketplace for assets on-chain.
+
+The goals of this package are to
+* Separate core logical components for readability and expansion over time. Eliminate data from being the layer of compatibility and instead the exposed functions.
```
movekevin, line 4: "Eliminate data from being the layer of compatibility and instead the exposed functions." doesn't make sense to me at all. Can you paraphrase or elaborate more?

aptos-move/move-examples/marketplace/sources/coin_listing.move
```diff
@@ -0,0 +1,1080 @@
+address marketplace {
```
movekevin, line 1: Are you experimenting with this for the test module?

aptos-move/move-examples/marketplace/sources/coin_listing.move
```diff
@@ -0,0 +1,1080 @@
...
+    struct PurchaseEvent has drop, store {
+        purchaser: address,
+        price: u64,
+        commission: u64,
+        royalties: u64,
+    }
+
+    // Init functions
+
+    #[legacy_entry_fun]
```
movekevin, line 97: Although this does allow the entry fun to return a value, did you check that this entry function is still callable as a tx? Also, it's not good to encourage using this attribute as it also skips all args checking during compilation, not just return value

aptos-move/move-examples/marketplace/sources/coin_listing.move
```diff
@@ -0,0 +1,1080 @@
...
+            start_time,
+            starting_bid,
+            bid_increment,
+            auction_end_time,
+            minimum_bid_time_before_end,
+            buy_it_now_price,
+        )
+    }
+
+    inline fun init<CoinType>(
```
movekevin, line 218: Does this have to be inline?

aptos-move/move-examples/marketplace/sources/coin_listing.move
```diff
@@ -0,0 +1,1080 @@
...
+/// Defines a single listing or an item for sale or auction. This is an escrow service that
+/// enables two parties to exchange one asset for another.
+/// Each listing has the following properties:
+/// * FeeSchedule specifying payment flows
+/// * Owner or the person that can end the sale or auction
+/// * Optional buy it now price
+/// * Ending time at which point it can be claimed by the highest bidder or left in escrow.
+/// * For auctions, the minimum bid rate and optional increase in duration of the auction if bids
+///   are made toward the end of the auction.
+module coin_listing {
```
movekevin, line 11: Are you still working on the fungible asset listing?

aptos-move/move-examples/marketplace/sources/listing.move
```diff
@@ -0,0 +1,259 @@
+/// Defines a single listing or an item for sale or auction. This is an escrow service that
+/// enables two parties to exchange one asset for another.
+/// Each listing has the following properties:
+/// * FeeSchedule specifying payment flows
+/// * Owner or the person that can end the sale or auction
+/// * Starting time
+/// * Logic for cleanup
+module marketplace::listing {
```
movekevin, line 8: Nice examples for the object extension pattern

aptos-move/move-examples/marketplace/sources/listing.move
```diff
@@ -0,0 +1,259 @@
...
+    use std::error;
+    use std::option;
+    use std::signer;
+    use std::string::String;
+
+    use aptos_framework::object::{Self, ConstructorRef, DeleteRef, ExtendRef, Object, ObjectCore, TransferRef};
+    use aptos_framework::timestamp;
+
+    use aptos_token::token::{Self as tokenv1, Token as TokenV1};
+    use aptos_token_objects::token;
```
movekevin, line 18: nit: Why not alias this as tokenv2 to be consistent with tokenv1

aptos-move/move-examples/marketplace/sources/listing.move
```diff
@@ -0,0 +1,259 @@
...
+            seller: signer::address_of(creator),
+            fee_schedule,
+            start_time,
+            delete_ref: object::generate_delete_ref(&constructor_ref),
+            extend_ref: object::generate_extend_ref(&constructor_ref),
+        };
+        move_to(&listing_signer, listing);
+
+        let listing_addr = object::address_from_constructor_ref(&constructor_ref);
+        object::transfer(creator, object, listing_addr);
```
movekevin, line 89: For later - I think for convenience we should add an object::transfer_to_object() that takes an Object<T> reference and does these 2 steps in a single call

aptos-move/move-examples/marketplace/sources/listing.move
```diff
@@ -0,0 +1,259 @@
...
+
+        (listing_signer, constructor_ref)
+    }
+
+    public(friend) fun create_tokenv1_container(
```
movekevin, line 94: IMO it's better to just separate this out into a wrapping example (e.g. Greg has one here https://github.com/aptos-labs/aptos-core/pull/8032)

aptos-move/move-examples/marketplace/sources/listing.move
```diff
@@ -0,0 +1,259 @@
...
+        delete_ref: DeleteRef,
+        /// Used to transfer the tokenv1 at the conclusion of a purchase.
+        transfer_ref: TransferRef,
+    }
+
+    // Init functions
+
+    public(friend) fun init(
+        creator: &signer,
+        object: Object<ObjectCore>,
```
movekevin, line 68: Shouldn't this be a token? Object < Token >. This could be generalized to object marketplace but other functionalities in this example are token-related anyway so having a general object input here is rather confusing IMO

aptos-move/move-examples/marketplace/sources/listing.move
```diff
@@ -0,0 +1,259 @@
...
+
+    #[view]
+    public fun compute_royalty(
+        object: Object<Listing>,
+        amount: u64,
+    ): (address, u64) acquires Listing, TokenV1Container {
+        let listing = borrow_listing(object);
+        let obj_addr = object::object_address(&listing.object);
+        if (exists<TokenV1Container>(obj_addr)) {
+            let token_container = borrow_global<TokenV1Container>(obj_addr);
```
movekevin, line 230: This math can be almost one-lined by creating a FixedPoint64 from numerator/denom and using the math libraries provided there

aptos-move/move-examples/marketplace/sources/listing.move
```diff
@@ -0,0 +1,259 @@
...
+
+            let payee_address = tokenv1::get_royalty_payee(&royalty);
+            let numerator = tokenv1::get_royalty_numerator(&royalty);
+            let denominator = tokenv1::get_royalty_denominator(&royalty);
+
+            let royalty_amount = amount * numerator / denominator;
+            (payee_address, royalty_amount)
+        } else {
+            let royalty = token::royalty(listing.object);
+            if (option::is_some(&royalty)) {
```
movekevin, line 242: Same here

aptos-move/move-examples/marketplace/sources/listing.move
```diff
@@ -0,0 +1,259 @@
...
+        if (exists<TokenV1Container>(obj_addr)) {
+            let token_container = borrow_global<TokenV1Container>(obj_addr);
+            let token_id = tokenv1::get_token_id(&token_container.token);
+            let royalty = tokenv1::get_royalty(token_id);
+
+            let payee_address = tokenv1::get_royalty_payee(&royalty);
+            let numerator = tokenv1::get_royalty_numerator(&royalty);
+            let denominator = tokenv1::get_royalty_denominator(&royalty);
+
+            let royalty_amount = amount * numerator / denominator;
```
movekevin, line 238: amount * numerator can overflow (both are u64). Please use proper math libraries in aptos-stdlib which prevent overflow

aptos-move/move-examples/marketplace/sources/coin_listing.move
```diff
@@ -0,0 +1,1080 @@
...
+        let now = timestamp::now_seconds();
+        assert!(now < auction_listing.auction_end_time, error::invalid_state(EAUCTION_ENDED));
+
+        let (previous_bidder, previous_bid, minimum_bid) = if (option::is_some(&auction_listing.current_bid)) {
+            let Bid { bidder, coins } = option::extract(&mut auction_listing.current_bid);
+            let current_bid = coin::value(&coins);
+            coin::deposit(bidder, coins);
+            (option::some(bidder), option::some(current_bid), current_bid + auction_listing.bid_increment)
+        } else {
+            (option::none(), option::none(), auction_listing.starting_bid)
```
movekevin, line 350: I think having option::none everywhere makes the code more complex and less readable. Wdyt about using 0 to indicate there's been no bid?

aptos-move/move-examples/marketplace/sources/coin_listing.move
```diff
@@ -0,0 +1,1080 @@
...
+
+        let fee_schedule = listing::fee_schedule(object);
+        coin::transfer<CoinType>(
+            bidder,
+            fee_schedule::fee_address(fee_schedule),
+            fee_schedule::bidding_fee(fee_schedule, bid_amount),
+        );
+
+        let now = timestamp::now_seconds();
+        let minimum_end_time = now + auction_listing.minimum_bid_time_before_end;
```
movekevin, line 369: Might be good to include this new end time in the bid event as well?

aptos-move/move-examples/marketplace/sources/coin_listing.move
```diff
@@ -0,0 +1,1080 @@
...
+
+        let seller = listing::seller(object);
+
+        let (purchaser, price, royalties, commission)  = if (option::is_some(&current_bid)) {
+            let Bid { bidder, coins } = option::destroy_some(current_bid);
+
+            let fee_schedule = listing::fee_schedule(object);
+            let price = coin::value(&coins);
+
+            let commission_charge = fee_schedule::commission(fee_schedule, price);
```
movekevin, line 415: A lot of this logic is redundant with purchase(). Can we refactor this into a different function (e.g. purchase_internal_with_price or something)

aptos-move/move-examples/marketplace/sources/listing.move
```diff
@@ -0,0 +1,259 @@
...
+    #[resource_group_member(group = aptos_framework::object::ObjectGroup)]
+    /// Corner-stone for all listings, represents the core utility layer including object
+    /// cleanup.
+    struct Listing has key {
+        /// The item owned by this listing, transferred to the new owner at the end.
+        object: Object<ObjectCore>,
+        /// The seller of the object
+        seller: address,
+        /// The fees associated with claiming this listing.
+        fee_schedule: Object<FeeSchedule>,
```
movekevin, line 43: I wonder if this should be optional

aptos-move/move-examples/marketplace/sources/fee_schedule.move
```diff
@@ -0,0 +1,498 @@
...
+            commission_numerator <= commission_denominator,
+            error::invalid_argument(EEXCEEDS_MAXIMUM),
+        );
+        assert!(
+            commission_denominator != 0,
+            error::out_of_range(EDENOMINATOR_IS_ZERO),
+        );
+
+        let (constructor_ref, fee_schedule_signer) = empty_init(creator, fee_address);
+        move_to(&fee_schedule_signer, FixedRateBiddingFee { bidding_fee });
```
movekevin, line 114: Why are these all different resources if they're always initialized together?

aptos-move/move-examples/marketplace/sources/fee_schedule.move
```diff
@@ -0,0 +1,498 @@
...
+
+    /// Create a marketplace with no fees.
+    public entry fun empty(
+        creator: &signer,
+        fee_address: address,
+    ) {
+        empty_init(creator, fee_address);
+    }
+
+    inline fun empty_init(creator: &signer, fee_address: address): (ConstructorRef, signer) {
```
movekevin, line 132: Related to my earlier comment in listing.move, it might be better to not have FeeSchedule then. That might be easier to index and query than having FeeSchedule with none of the individual fee resources (which are also confusing and tricky to deal with indexing and off chain wise)

aptos-move/move-examples/marketplace/sources/fee_schedule.move
```diff
@@ -0,0 +1,498 @@
...
+        commission: u64,
+    ) acquires FeeSchedule, FixedRateCommission, PercentageRateCommission {
+        let fee_schedule_signer = remove_commission(creator, marketplace);
+        move_to(&fee_schedule_signer, FixedRateCommission { commission });
+        let updated_resource = type_info::type_name<FixedRateListingFee>();
+        emit_mutation_event(signer::address_of(&fee_schedule_signer), updated_resource);
+    }
+
+    /// Remove any existing commission and set a percentage rate commission.
+    public entry fun set_percentage_rate_commission(
```
movekevin, line 225: This can allow funky behaviors if the marketplace owner changes the commissions after listings are already created when they see that there's a large purchase going through

aptos-move/move-examples/marketplace/sources/fee_schedule.move
```diff
@@ -0,0 +1,498 @@
...
+        creator: &signer,
+        marketplace: Object<FeeSchedule>,
+        denominator: u64,
+        numerator: u64,
+    ) acquires FeeSchedule, FixedRateCommission, PercentageRateCommission {
+        assert!(
+            numerator <= denominator,
```
movekevin, line 232: Should this be strictly < 100%. 100% commission doesn't quite make sense to me

aptos-move/move-examples/marketplace/sources/fee_schedule.move
```diff
@@ -0,0 +1,498 @@
...
+
+    #[resource_group_member(group = aptos_framework::object::ObjectGroup)]
+    /// Defines marketplace fees
+    struct FeeSchedule has key {
+        /// Address to send fees to
+        fee_address: address,
+        /// Ref for changing the configuration of the marketplace
+        extend_ref: ExtendRef,
+        /// An event stream of changes to the fee schedule
+        mutation_events: EventHandle<MutationEvent>,
```
movekevin, line 35: As a design pattern, I wonder if we should encourage events to be in a different resource now that we have resource groups

aptos-move/move-examples/marketplace/sources/fee_schedule.move
davidiw: well I think the trickier thing is that right now we don't actually take the minimum of the asked versus what's available... we have a lot of ratios or amounts... but if there is insufficient coin from the purchase... we need to either allow the transaction to fail or to support the minimum available.

Currently the transaction fails.

For example, if the creator has a 100% royalty and the commission is > 0%, the txn will fail.

aptos-move/move-examples/marketplace/sources/listing.move
davidiw: the original idea was to make this general, but agreed...

aptos-move/move-examples/marketplace/Move.toml
```diff
@@ -1,11 +1,11 @@
...
-name = "MarketPlace"
-version = "0.0.1"
+name = "Marketplace"
+version = "0.0.0"
+
+[addresses]
+marketplace = "_"
 
 [dependencies]
 AptosFramework = { local = "../../framework/aptos-framework" }
```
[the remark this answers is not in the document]
movekevin: There's no right approach :P I usually personally prefer as little code as possible

aptos-move/move-examples/marketplace/sources/coin_listing.move
movekevin: Alright sounds good

aptos-move/move-examples/marketplace/sources/listing.move
movekevin: Ah yeah that's true. Ignore my comment then 🙏

aptos-move/move-examples/marketplace/sources/coin_listing.move
movekevin: May be. I don't feel strongly here though

aptos-move/move-examples/marketplace/sources/listing.move
davidiw: actually, recall this takes the tokenv1 container...

aptos-move/move-examples/marketplace/sources/listing.move
davidiw: This is different. This makes the assumption that we want the tokenv1 in the tokenv2. This in turn also creates an undiscoverable collection... that needs more work and isn't compatible with the goals of this project. Let me remove the tokenwarapper into its own file though.

aptos-move/move-examples/marketplace/sources/coin_listing.move
davidiw: ahh... so the separate module is great because otherwise you end up with a ton of #[test_only] that aren't readable. same file is great because you know exactly where the demo code is.

aptos-move/move-examples/marketplace/Move.toml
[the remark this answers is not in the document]
davidiw: there should always be a right approach :)

aptos-move/move-examples/marketplace/sources/listing.move
davidiw: I feel fixedpoint is too heavy... as it converts everything to u128 and then u256, but agreed we can make this safer.

aptos-move/move-examples/marketplace/sources/coin_listing.move
davidiw: I see, that's fair, but we should be able to adjust if we hit compiler errors :shrug:

aptos-move/move-examples/marketplace/sources/coin_listing.move
davidiw: ahh, I remember, this result in slightly less optimal code for cases where auctions don't end with a bidder... but whatever... I think most of the outcome results in no real impact.

davidiw (commented):
So I'm done with all the changes albeit the tokenv1 work.

So one of the issues we encounter is that the contents of a token cannot be unlocked until the end of an auction or purchased.

For a tokenv1, we place it in a stasis until this point in time.

We enforce that by ensuring that only friend functions can call that unlock behavior.

Even if we move the tokenv1 into a dedicated shape, if we want to provide a generic unlock, we'd have to do it by providing the `TransferRef` to another party for transferring when they deem appropriate.

I'm not inclined to the increased level of complexity for whatever UX or DevX gains there might be.

aptos-move/move-examples/marketplace/sources/listing.move
movekevin: You can also use https://github.com/aptos-labs/aptos-core/blob/main/aptos-move/framework/aptos-stdlib/sources/math64.move#L32

aptos-move/move-examples/marketplace/sources/listing.move
```diff
@@ -0,0 +1,263 @@
...
+            let denominator = tokenv1::get_royalty_denominator(&royalty);
+
+            let royalty_amount = amount * numerator / denominator;
+            (payee_address, royalty_amount)
+        } else {
+            let royalty = tokenv2::royalty(listing.object);
+            if (option::is_some(&royalty)) {
+                let royalty = option::destroy_some(royalty);
+                let payee_address = royalty::payee_address(&royalty);
+                let royalty_amount =
```
movekevin, line 247: You can use https://github.com/aptos-labs/aptos-core/blob/main/aptos-move/framework/aptos-stdlib/sources/math64.move#L32

movekevin (approved):

movekevin (approved):

aptos-move/move-examples/marketplace/sources/coin_listing.move
```diff
@@ -0,0 +1,1156 @@
...
+        let fee_schedule = listing::fee_schedule(object);
+        coin::transfer<CoinType>(
+            bidder,
+            fee_schedule::fee_address(fee_schedule),
+            fee_schedule::bidding_fee(fee_schedule, bid_amount),
+        );
+
+        let now = timestamp::now_seconds();
+        let current_end_time = auction_listing.auction_end_time;
+        let minimum_end_time = now + auction_listing.minimum_bid_time_before_end;
```
lightmark, line 437: if the seller cannot cancel the auction and the bid_increment is too low, would this allow ppl to endlessly extend the auction duration?
Why do we need this?

aptos-move/move-examples/marketplace/sources/collection_offer.move
```diff
@@ -0,0 +1,750 @@
...
+    const ENOT_OWNER: u64 = 4;
+    /// The offered token is not within the expected collection.
+    const EINCORRECT_COLLECTION: u64 = 5;
+    /// The collection offer has expired.
+    const EEXPIRED: u64 = 6;
+
+    // Core data structures
+
+    #[resource_group_member(group = aptos_framework::object::ObjectGroup)]
+    /// Create a limed lifetime offer to buy tokens from a collection. The collection and
```
lightmark, line 45: typo

lightmark (approved):
some nits

aptos-move/move-examples/marketplace/sources/listing.move
```diff
@@ -0,0 +1,263 @@
...
+    }
+
+    /// If the account did not have tokenv1 enabled, then it must call this after making the
+    /// purchase to extract the token.
+    public entry fun extract_tokenv1(
+        owner: &signer,
+        object: Object<TokenV1Container>,
+    ) acquires TokenV1Container {
+        let object_addr = object::object_address(&object);
+        assert!(object::is_owner(object, signer::address_of(owner)), 0);
```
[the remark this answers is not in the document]
davidiw: crap

aptos-move/move-examples/marketplace/sources/listing.move
```diff
@@ -0,0 +1,263 @@
...
+        let TokenV1Container {
+            token,
+            delete_ref: _,
```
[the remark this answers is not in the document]
davidiw: oh man

aptos-move/move-examples/marketplace/sources/coin_listing.move
davidiw: this is actually quite common... we can update based upon community feedback

aptos-move/move-examples/marketplace/sources/collection_offer.move
davidiw: gah

aptos-move/move-examples/marketplace/sources/fee_schedule.move
```diff
@@ -0,0 +1,495 @@
...
+
+    #[resource_group_member(group = aptos_framework::object::ObjectGroup)]
+    /// Fixed rate for commission
+    struct FixedRateCommission has drop, key {
+        /// Fixed rate for commission
+        commission: u64,
+    }
+
+    #[resource_group_member(group = aptos_framework::object::ObjectGroup)]
+    /// Percentage-based rate for commission
```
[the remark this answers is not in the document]
davidiw: since this is an RFC, demo, I'll leave this open for discussion.

aptos-move/move-examples/marketplace/sources/fee_schedule.move
```diff
@@ -0,0 +1,495 @@
...
+        marketplace: Object<FeeSchedule>,
+        fee: u64,
+    ) acquires FeeSchedule, FixedRateListingFee {
+        let fee_schedule_signer = remove_listing_fee(creator, marketplace);
+        move_to(&fee_schedule_signer, FixedRateListingFee { listing_fee: fee });
+        let updated_resource = type_info::type_name<FixedRateListingFee>();
+        emit_mutation_event(signer::address_of(&fee_schedule_signer), updated_resource);
+    }
+
+    inline fun remove_listing_fee(
```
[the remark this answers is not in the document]
davidiw: not sure what this means.

PR #8417, opened 2023-05-30
[Spec] Fix spec on mainnet branch

rahxephon89:
### Description

Fix spec to pass compiler check


davidiw (approved):
verified locally, thanks!

PR #8418, opened 2023-05-30
[Spec] Fix spec on testnet branch

rahxephon89:
### Description

Fix specs on testnet branch to pass the compiler check


davidiw (approved):
verified locally. thanks!

davidiw (approved):
verified locally. thanks!

PR #8419, opened 2023-05-30
restrict pull_request_target trigger on docker-build-test to only run…

geekflyer:
… when specific branches are targeted

### DIFF
--- a/.github/workflows/docker-build-test.yaml
+++ b/.github/workflows/docker-build-test.yaml
@@ -171,43 +171,8 @@ jobs:
       FEATURES: consensus-only-perf-test
       BUILD_ADDL_TESTING_IMAGES: true
 
-  rust-images-all:
-    needs:
-      [
-        determine-docker-build-metadata,
-        rust-images,
-        rust-images-indexer,
-        rust-images-failpoints,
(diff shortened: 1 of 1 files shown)


rustielin (commented):
IIUC this change needs to be applied against every single branch to have any real effect, since any branch can be the base of a PR. (https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target)

Also this removes the possibility of testing on stacked PRs, where a PR targets the branch of a previous PR.

I don't think this is really needed

.github/workflows/docker-build-test.yaml
```diff
@@ -25,6 +25,18 @@ on: # build on main branch OR when a PR is labeled with `CICD:build-images`
   workflow_dispatch:
   pull_request_target:
     types: [labeled, opened, synchronize, reopened, auto_merge_enabled]
+    branches:
+        - main
+        # release branches
+        - devnet
+        - testnet
+        - mainnet
+        - aptos-node-v*
```
sionescu, line 34: aptos-node-* are no longer used for releases, you can omit this.

sionescu (approved):

PR #8420, opened 2023-05-30
[Fix] Fix flaky test.

grao1991:


msmouse (approved):
Nice find anyway. Let's see if it fixes the test.

areshand (approved):

PR #8421, opened 2023-05-30
rosetta delegator address balances

zizhengjiang:
### Description
Implement the delegator address balances rosetta endpoint.

The rosetta client should manage its own delegator and pool address relationship when fetching relevant data.
The request payload should look like this (added the metadata field under sub_account_identifier):
```
"account_identifier": {
    "address": "0x899d44dca988af7895cdc9828ae0e7f6c5c42d74fa1610e693fb5f613cb8c09a", // owner/delegator address
    "sub_account": {
        "address": "active_stake", // string active
        # this is the new field added
        "metadata": {
            "pool_address": "0xbd50bb20e1972747db365c3a6e31255d78867c6b8df9877ea6054bb335dd6031" // pool_address
        }
    }
}
```

**About balance fetching**:
Using delegator address [0x65b1e4d62e803d7e3a5c715baa33391757ae530c9fef287bf2d7d15c1ea6e3e6](https://explorer.aptoslabs.com/account/0x65b1e4d62e803d7e3a5c715baa33391757ae530c9fef287bf2d7d15c1ea6e3e6?network=mainnet) and its validator address [0x2ded75e99c6efbe143a9648a3e88ebf9d0cf249b2af44d510bdd0287e8adcc79](https://explorer.aptoslabs.com/account/0x2ded75e99c6efbe143a9648a3e88ebf9d0cf249b2af44d510bdd0287e8adcc79?network=mainnet) as an example

The response should be similar to native staking:
Variables that should be fetched from the **delegator address:**
1. Sequence number from its `0x1::account::Account` resource
2. Balances from delegation_pool.move get_stake() function

Variables should be fetched from the **pool address:**
~~1. Operator from `0x1::stake::StakePool` resource~~
2. Lockup until secs from `0x1::stake::StakePool` resource

### Test Plan

Tested with remote online node from e2e
**delegated staking balance**
Screenshot 2023-06-01 at 4 15 45 PM

**native staking balance**
Screenshot 2023-06-01 at 4 16 37 PM

**direct balance**
Screenshot 2023-06-01 at 4 17 25 PM

### DIFF
--- a/crates/aptos-rosetta/src/types/identifiers.rs
+++ b/crates/aptos-rosetta/src/types/identifiers.rs
@@ -39,7 +39,7 @@ impl AccountIdentifier {
     pub fn pool_address(&self) -> ApiResult<Option<AccountAddress>> {
         if let Some(sub_account) = &self.sub_account {
             if let Some(metadata) = &sub_account.metadata {
-                return str_to_account_address(&metadata.pool_address.as_str()).map(Some);
+                return str_to_account_address(metadata.pool_address.as_str()).map(Some);
             }
         }
 
--- a/crates/aptos-rosetta/src/types/identifiers.rs
+++ b/crates/aptos-rosetta/src/types/identifiers.rs
@@ -43,10 +43,7 @@ impl AccountIdentifier {
             }
         }
 
-        Err(ApiError::InvalidInput(Some(format!(
-            "Invalid pool address {:?}",
-            self
-        ))))
+        Ok(None)
     }
 
     pub fn base_account(address: AccountAddress) -> Self {
--- a/crates/aptos-rosetta/src/types/identifiers.rs
+++ b/crates/aptos-rosetta/src/types/identifiers.rs
@@ -36,6 +36,19 @@ impl AccountIdentifier {
         str_to_account_address(self.address.as_str())
     }
 
+    pub fn pool_address(&self) -> ApiResult<Option<AccountAddress>> {
+        if let Some(sub_account) = &self.sub_account {
+            if let Some(metadata) = &sub_account.metadata {
+                return str_to_account_address(&metadata.pool_address.as_str()).map(Some);
+            }
+        }
+
+        Err(ApiError::InvalidInput(Some(format!(
+            "Invalid pool address {:?}",
+            self
+        ))))
+    }
+
     pub fn base_account(address: AccountAddress) -> Self {
         AccountIdentifier {
             address: to_hex_lower(&address),
@@ -132,6 +145,30 @@ impl AccountIdentifier {
         }
     }
 
+    pub fn is_delegator_active_stake(&self) -> bool {
+        if let Some(ref inner) = self.sub_account {
+            inner.is_delegator_active_stake()
+        } else {
+            false
+        }
+    }
+
+    pub fn is_delegator_inactive_stake(&self) -> bool {
+        if let Some(ref inner) = self.sub_account {
+            inner.is_delegator_inactive_stake()
+        } else {
+            false
+        }
+    }
+
+    pub fn is_delegator_pending_inactive_stake(&self) -> bool {
+        if let Some(ref inner) = self.sub_account {
+            inner.is_delegator_pending_inactive_stake()
+        } else {
+            false
+        }
+    }
+
     pub fn is_operator_stake(&self) -> bool {
         if let Some(ref inner) = self.sub_account {
             !(inner.is_total_stake()
@@ -167,6 +204,9 @@ fn str_to_account_address(address: &str) -> Result<AccountAddress, ApiError> {
 pub struct SubAccountIdentifier {
     /// Hex encoded AccountAddress beginning with 0x
     pub address: String,
+    /// Metadata only used for delegated staking
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub metadata: Option<SubAccountIdentifierMetadata>,
 }
 
 const STAKE: &str = "stake";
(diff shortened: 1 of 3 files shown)


crates/aptos-rosetta/src/types/identifiers.rs
```diff
@@ -222,15 +268,27 @@ impl SubAccountIdentifier {
...
     }
 
     pub fn is_pending_inactive_stake(&self) -> bool {
-        self.address.as_str() == PENDING_INACTIVE_STAKE
+        self.address.as_str() == PENDING_INACTIVE_STAKE && self.metadata.is_none()
     }
 
     pub fn is_inactive_stake(&self) -> bool {
-        self.address.as_str() == INACTIVE_STAKE
+        self.address.as_str() == INACTIVE_STAKE && self.metadata.is_none()
```
xindingw, line 279: I understand this is not from this PR but why we have to do this hack? by any means, Neither `STAKE`, `PENDING_INACTIVE_STAKE ` or `INACTIVE_STAKE` is an address. If we just need a enumeration, I believe `metadata` field is for this purpose, right?

crates/aptos-rosetta/src/types/misc.rs
```diff
@@ -86,6 +99,8 @@ pub struct BalanceResult {
     pub balance: Option<Amount>,
     /// Time at which the lockup expires and pending_inactive balance becomes inactive
     pub lockup_expiration: u64,
+    /// Operator account address
```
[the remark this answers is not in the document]
zizhengjiang: will remove it

crates/aptos-rosetta/src/types/misc.rs
```diff
@@ -315,3 +331,69 @@ pub async fn get_stake_balances(
...
+
+        if account_identifier.is_delegator_active_stake() {
+            requested_balance = Some(result.get(0).to_string());
+        } else if account_identifier.is_delegator_inactive_stake() {
+            requested_balance = Some(result.get(1).to_string());
+        } else if account_identifier.is_delegator_pending_inactive_stake() {
+            requested_balance = Some(result.get(2).to_string());
+        }
+    } else {
+        Ok(None)
```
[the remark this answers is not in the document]
zizhengjiang: here i'd like to just mimic the `get_stake_balances` func and let the upper layer (account.rs) handle the error

crates/aptos-rosetta/src/types/identifiers.rs
zizhengjiang: I feel we probably don't want to deal with extra fields that's why it was initially implemented like this
given the current code is already in use internally i'd prefer just leave it as it is and use the new field get the target data

crates/aptos-rosetta/src/types/misc.rs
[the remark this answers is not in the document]
xindingw: No, this is different from the existing code to handle `requested_balance`(it's also problematic though). `rest_client.view` might return an error and in this case the else branch will be entered. In the else branch, the error value is ignored and `OK` is returned so the upper layer cannot even know there is an error.

crates/aptos-rosetta/src/types/identifiers.rs
xindingw: I won't block this PR for this. But I highly suggest to fix it ASAP because it doesn't make sense at all and the schema will be incompatible after fixing.

crates/aptos-rosetta/src/types/misc.rs
```diff
@@ -315,3 +331,69 @@ pub async fn get_stake_balances(
...
+        Ok(Some(BalanceResult {
+            balance: Some(Amount {
+                value: balance,
+                currency: native_coin(),
+            }),
+            lockup_expiration,
+            operator_address,
+        }))
+    } else {
+        Ok(None)
```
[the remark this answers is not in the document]
zizhengjiang: will remove the other two, just left this one
it will be handled by the `account.rs` layer

crates/aptos-rosetta/src/types/misc.rs
```diff
@@ -315,3 +326,80 @@ pub async fn get_stake_balances(
...
+        } else if account_identifier.is_delegator_inactive_stake() {
+            requested_balance = result.get(1).map(|v| v.to_string());
+        } else if account_identifier.is_delegator_pending_inactive_stake() {
+            requested_balance = result.get(2).map(|v| v.to_string());
+        }
+    } else {
+        return Err(ApiError::InternalError(Some(
+            "Unable to get delegation stake balances".to_string(),
+        )));
+    }
```
xindingw, line 369: This still discards the error from `rest_client.view`. You can try `.await?`, which throws the error if there is one.:
```suggestion
    let balances_response = rest_client.view(
        &ViewRequest {
            function: DELEGATION_POOL_GET_STAKE_FUNCTION.clone(),
            type_arguments: vec![],
            arguments: vec![
                serde_json::Value::String(String::from(
                    pool_address.to_string(),
                )),
                serde_json::Value::String(String::from(
                    owner_address.to_string()
                )),
            ],
        },
        Some(version),
    ).await?;
    let result = balances_response.into_inner();
    ...
```

xiaoxiaff (approved):

xiaoxiaff (approved):

xiaoxiaff (approved):

nicolas-martin (approved):

crates/aptos-rosetta/src/types/misc.rs
```diff
@@ -315,3 +326,64 @@ pub async fn get_stake_balances(
...
+    ).await?;
+
+    let balances_result = balances_response.into_inner();
+    if account_identifier.is_delegator_active_stake() {
+        requested_balance = balances_result.get(0).and_then(|v| v.as_str().map(|s| s.to_owned()));
+    } else if account_identifier.is_delegator_inactive_stake() {
+        requested_balance = balances_result.get(1).and_then(|v| v.as_str().map(|s| s.to_owned()));
+    } else if account_identifier.is_delegator_pending_inactive_stake() {
+        requested_balance = balances_result.get(2).and_then(|v| v.as_str().map(|s| s.to_owned()));
+    }
```
xindingw, line 359: nit: Maybe throw "Unable to construct BalanceResult instance" error here?

xindingw (approved):
LGTM! Let me know if you need me to find another approver.

davidiw (approved):

PR #8425, opened 2023-05-30
Deploy and Upgrade tutorial for MOVE

vusirikala:
### Description

Added some MOVE files in `move-examples` folder. Will use these examples to create a tutorial in `aptos.dev/tutorials` page  for deploying and upgrading MOVE smart contracts. 

Step 1: Hello blockchain example
Step 2: Before storing a message, trim the message to 8 characters. For this add a private function called `trim_message`. 
Step 3: Change the function signature of the private function `trim_message`. `trim_message` function now additionally takes length as input.


aptos-move/move-examples/upgrade-tutorial/step_1/sources/message_store.move
```diff
@@ -0,0 +1,59 @@
+module upgrade_tutorial::message_store {
```
movekevin, line 1: This looks similar to the existing message example we already have. Can we combine somehow?

movekevin (commented):
Thanks for getting started on this. Are you planning to add the tutorial text here or in a separate followup?

PR #8431, opened 2023-05-30
[gha] check aptos-node version against branch

rustielin:
### Description
Quick GHA so we can check the aptos-node version against the current release branch/tag. This way, `aptos-node --version` does not fall out of sync

### Test Plan
Unit tests. 

Also check canary on this PR, which is expected to fail. `aptos-node/Cargo.toml` has 1.4.0, so we expect it to match the branch name (which is `rustielin/check-aptos-node-version-v1.5` in this case, so it fails)


thepomeranian (approved):

rustielin:
closing in favor of https://github.com/aptos-labs/aptos-core/pull/8432

PR #8447, opened 2023-05-31
Add get-latest-cli action

banool:
### Description
This PR adds a GH action to install the CLI.

We could theoretically expand this to create a profile as well but I felt that made the action sort of complicated (just look at `aptos init`, it has a lot of flags we'd probably want to expose). We can chat about that though.

Depends on https://github.com/aptos-labs/aptos-core/pull/8446.

### Test Plan
@brianolson tested this in internal-ops, thanks Brian!!


brianolson (approved):

rustielin (approved):
Amazing! Left a comment above, but otherwise looks good

brianolson (approved):

PR #8449, opened 2023-05-31
Add vector::remove_value

movekevin:
### Description
This is a commonly used operation

### Test Plan
Unit tests


aptos-move/framework/move-stdlib/sources/vector.move
```diff
@@ -204,6 +204,26 @@ module std::vector {
         pragma intrinsic = true;
     }
 
+    /// Remove the first occurrence of a given value in the vector `v` and return it in a vector, shifting all
+    /// subsequent elements.
+    /// This is O(n) and preserves ordering of elements in the vector.
+    /// This returns an empty vector if the value isn't present in the vector.
+    /// Note that this cannot return an option as option uses vector and there'd be a circular dependency between option
+    /// and vector.
+    public fun remove_value<Element>(v: &mut vector<Element>, val: &Element): vector<Element> {
```
gerben-stavenga, line 213: Option<Element> to indicate oneness

movekevin: Can't use option here directly as explained in the comment. option.move depends on vector.move so if vector also depends on option, that'd be a circular dependency

lightmark: hmmm, I think for this case, the `Element` should have copy. So you could return a `bool`. Or could you share an example that `Element` is not copyable.

gerben-stavenga: You mean drop instead of copy right? I think this is fine

lightmark: I mean both. to get the returned value you can just copy `val`

movekevin: Yeah but that'd add arbitrary constraints to this function. Also, in the scenario mentioned you need drop, not copy. It's actually pretty easy for caller to call vector::remove_value and just ignore the returned value if it's droppable. This would make this function useful regardless of if the element has drop

davidiw: so ugly, so ugyl

davidiw (approved):

davidiw (approved):

aptos-move/framework/move-stdlib/sources/vector.move
movekevin: Circular dependency between modules is way too strict :(

gerben-stavenga: vector is a fundamental type so not tied to a module. It's likely one can remove the dependence on vector from option

movekevin: Not if option is calling functions on vector such as vector::length(). Those unfortunately can't be replaced with native syntax (yet)

PR #8450, opened 2023-05-31
[storage] use a valid target version in db restore

areshand:
### Description
If the user doesn't specify a target version or specify a target version larger than the max backup txn version, it could trigger a bug to restore more ledger ending info than required. 

### Test Plan

manual test the command


msmouse (approved):
❤️

grao1991 (approved):

PR #8451, opened 2023-05-31
[telemetry-svc][bug-fix] Add timeouts for requests

ibalajiarun:
### Description

This PR should fix a bug with telemetry service that causes metrics push requests to take a long time to complete. The issue is:

First, there is no timeout on the client side. This can affect the 15-second cadence to send metrics periodically to telemetry service. A 5-second timeout has been added on the client-side.

Second, the server waits for all backend push metrics API calls to complete either successfully or fail. This is problematic because the API calls themselves have reties associated with failures, so waiting on all the backend calls means waiting on the slowest call. To fix this, we will wait for a timeout and decide as success if any of the backend calls have succeeded.

There is now a limit on maximum retry duration of 10 seconds on the node-side and the metrics are dropped otherwise and metrics are sent in the next interval.

### Test Plan

Existing tests.

### DIFF
--- a/crates/aptos-telemetry-service/src/prometheus_push_metrics.rs
+++ b/crates/aptos-telemetry-service/src/prometheus_push_metrics.rs
@@ -10,7 +10,6 @@ use crate::{
     metrics::METRICS_INGEST_BACKEND_REQUEST_DURATION,
     types::{auth::Claims, common::NodeType},
 };
-use futures::{select, stream::FuturesUnordered, FutureExt, StreamExt};
 use reqwest::{header::CONTENT_ENCODING, StatusCode};
 use std::time::Duration;
 use tokio::time::Instant;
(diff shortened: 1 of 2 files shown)


crates/aptos-telemetry-service/src/prometheus_push_metrics.rs
```diff
@@ -57,52 +61,82 @@ pub async fn handle_metrics_ingest(
...
-                METRICS_INGEST_BACKEND_REQUEST_DURATION
-                    .with_label_values(&[&claims.peer_id.to_string(), name, res.status().as_str()])
-                    .observe(start_timer.elapsed().as_secs_f64());
-                if res.status().is_success() {
-                    debug!("remote write to victoria metrics succeeded");
-                } else {
+    let mut post_futures: FuturesUnordered<_> = client
+        .into_iter()
+        .map(|(name, client)| async {
+            let result = client
```
JoshLind, line 67: nit: you could use tokio::timeout, right? e.g., https://rcos.io/static/internal_docs/tokio/time/fn.timeout.html. 😄 e.g.., just use `timeout` in here, and then below, the `select!` will just wait for them all to return?

Either way, LGTM 😄

JoshLind (approved):

crates/aptos-telemetry-service/src/prometheus_push_metrics.rs
ibalajiarun: That makes it much simpler. Thanks Josh! 

I was trying to timeout with join_all and it did not make sense, so I came up with a complex solution.

PR #8454, opened 2023-05-31
[TS SDK ] run ts sdk indexer tests in a separate CI job

0xmaayan:
### Description
indexer service can be flaky and we dont want those tests to be land blocking for any PR on the aptos repo. This is why we run indexer tests separate from the main sdk test run
### Test Plan
tests are passing


banool (approved):

.github/workflows/ts-sdk-indexer-tests.yaml
rustielin, line 1: would it make sense to run this in the same workflow as `sdk-release.yaml`, but just as a different job?

rustielin (approved):
stamping to unblock, but technically this can be in the same workflow iiuc

.github/workflows/sdk-release.yaml
```diff
@@ -141,3 +141,28 @@ jobs:
...
+
+  # Run the TS SDK indexer tests. Note: indexer service can be flaky and we
+  # dont want those tests to be land blocking for any PR on the aptos repo.
+  # This is why we run those tests separate from
+  # test-sdk-confirm-client-generated-publish.
+  run-indexer-test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
```
rustielin, line 153: can you pin the version of these third party action deps like in the other job? e.g.
```
actions/setup-node@969bd2663942d722d85b6a8626225850c2f7be4b # pin@v3
```

0xmaayan: should I do the same for checkout?
```
actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8 # pin@v3
```

rustielin: yep, and `pnpm/action-setup@v2` as well

rustielin:
```
pnpm/action-setup@537643d491d20c2712d11533497cb47b2d0eb9d5 # pin https://github.com/pnpm/action-setup/releases/tag/v2.2.3
```

0xmaayan: we use it like this in the other job thought https://github.com/aptos-labs/aptos-core/pull/8454/files#diff-f9bdc9424e6b94fc4cdf8b6974e1c90d62ffc04c88d91c80c721022b183df298R62-R64

rustielin: ah that should best be changed as well. pinning the third party dep by SHA is a best practice

PR #8457, opened 2023-05-31
bcs <-> proto conversion

gerben-stavenga:
### Description

Prototype protobuf (alliteration for the win) API for move values.

Each move struct would correspond with a .proto definition with custom field options. The custom field options allows imbuing the Move semantics of the fields in the .proto definitions.

This means that a proto compiler plugin and can automatically generate code with MoveStructLayout data. I added two routines. 

1) proto_serialize(val: &MoveStruct) -> Vec<u8> serializing a move struct value as a protobuf. This can be directly parsed by the client side generated protobuffers.
2) proto_deserialize(proto: &[u8], layout: &MoveStructLayout) -> MoveStruct directly parse a binary proto stream into a MoveStruct


davidiw:
what's the plan here? looks interesting.

gerben-stavenga:
> what's the plan here? looks interesting.

This was just prototyping. I discovered there are some unfortunate issues.

Move structs and BCS have Vec<Vec<u64>> types, but protos don't. So to support it one needs to add additional wrapper protos. Similarly you lose type information for types like u128 which don't have native proto representation.

Still I thought it's interesting enough to at least keep around

davidiw:
closing out as no work planned

PR #8462, opened 2023-06-01
[Faucet] Make outstanding txn wait duration configurable, update defaults

banool:
> We should also decrease the expiration secs on the txn from 30 seconds to 10 seconds, the E2E latency on the txns should never be above 10 seconds. This is the value we generally use everywhere else (e.g. in the Rust and TS SDKs).

More information here: https://aptos-org.slack.com/archives/C05ATSJJCSV/p1685609837398169?thread_ts=1685606365.531029&cid=C05ATSJJCSV.

### Test Plan
E2E tests in CI.

### DIFF
--- a/crates/aptos-faucet/cli/src/main.rs
+++ b/crates/aptos-faucet/cli/src/main.rs
@@ -74,8 +74,8 @@ impl FaucetCliArgs {
             Duration::from_secs(30),
             None,
             self.max_gas_amount,
-            10,
             15,
+            20,
             true,
         );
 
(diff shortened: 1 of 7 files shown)


crates/aptos-faucet/cli/src/main.rs
```diff
@@ -74,7 +74,8 @@ impl FaucetCliArgs {
             Duration::from_secs(30),
             None,
             self.max_gas_amount,
-            30,
+            15,
```
perryjrandall, line 77: Should this be configurable in some way?

perryjrandall (approved):

crates/aptos-faucet/cli/src/main.rs
banool: This CLI doesn't get used I don't think so it's probably not necessary. It's configurable for the actual faucet which is what matters.

0xmaayan (approved):

PR #8463, opened 2023-06-01
[docs]update module file name recommendation

fgfm999:
### Description
Update `module file name` recommendation. 
Since we recommend to use  lower snake case for `module name`, and keep `module name` and `module file name` to be same, the example should be `option.move`.


davidiw (approved):
nice finds

PR #8465, opened 2023-06-01
[CLI][E2E] Expose metrics port of local testnet node

banool:
### Description
Currently the metrics of the node are inaccessible because the metrics port on the container is not published. This PR fixes that and tests for it.

### Test Plan
```
poetry run python main.py --base-network mainnet --test-cli-tag mainnet
```

See CI also.

### DIFF
--- a/crates/aptos/e2e/cases/init.py
+++ b/crates/aptos/e2e/cases/init.py
@@ -3,6 +3,7 @@
 
 import os
 
+import requests
 from common import TestError
 from test_helpers import RunHelper
 from test_results import test_case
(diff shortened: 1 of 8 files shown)


crates/aptos/e2e/cases/init.py
```diff
@@ -41,3 +42,10 @@ def test_init(run_helper: RunHelper, test_name=None):
         raise TestError(
             f"Failed to query local testnet for account {account_info.account_address}"
         ) from e
+
+
+@test_case
+def test_metrics_accessible(run_helper: RunHelper, test_name=None):
+    # Assert that the metrics endpoint is accessible.
+    metrics_url = run_helper.get_metrics_url(json=True)
+    requests.get(metrics_url).json()
```
0xjinn, line 51: Do we need newline at the bottom of the file?

banool: Doesn't look like it:
```
$ pre-commit run --all-files
Trim Trailing Whitespace.................................................Passed
Fix End of Files.........................................................Passed
Check for added large files..............................................Passed
Insert license in comments...............................................Passed
```

0xjinn: Interesting. I guess I shouldn't manually add a new /n, and just run black and let linter do its thing lol

0xjinn (approved):
Niiiiiice!

crates/aptos/e2e/cases/init.py
banool: Yeah `pre-commit` handles it hahah.

0xmaayan (approved):

PR #8474, opened 2023-06-01
Add accurate resource group gas metering

gerben-stavenga:
### Description

Refactor the code to return actual bytes read from storage to the gas metering. Use gas version feature to enable this when the next version is bumped.

### DIFF
--- a/aptos-move/aptos-vm/src/data_cache.rs
+++ b/aptos-move/aptos-vm/src/data_cache.rs
@@ -54,9 +54,7 @@ impl<'a, S: StateView> StorageAdapter<'a, S> {
             accurate_byte_count: false,
             resource_group_cache: RefCell::new(BTreeMap::new()),
         };
-        if gas_config(&s).1 >= 9 {
-            s.accurate_byte_count = true;
-        };
+        s.accurate_byte_count = true;
         s
     }
 
--- a/aptos-move/aptos-vm/src/data_cache.rs
+++ b/aptos-move/aptos-vm/src/data_cache.rs
@@ -21,7 +21,7 @@ use move_core_types::{
     account_address::AccountAddress,
     language_storage::{ModuleId, StructTag},
     metadata::Metadata,
-    resolver::{ModuleResolver, ResourceResolver},
+    resolver::{resource_add_cost, ModuleResolver, ResourceResolver},
     vm_status::StatusCode,
 };
 use move_table_extension::{TableHandle, TableResolver};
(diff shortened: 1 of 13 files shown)


movekevin (approved):

third_party/move/move-core/types/src/resolver.rs
```diff
@@ -41,7 +42,7 @@ pub trait ResourceResolver {
         address: &AccountAddress,
         typ: &StructTag,
         metadata: &[Metadata],
-    ) -> Result<Option<Vec<u8>>, Error>;
+    ) -> Result<(Option<Vec<u8>>, Option<NumBytes>), Error>;
```
vgao1996, line 45: Should make this `Option<(Vec<u8>, NumBytes)>` instead.

aptos-move/aptos-vm/src/data_cache.rs
```diff
@@ -65,32 +76,40 @@ impl<'a, S: StateView> StorageAdapter<'a, S> {
         address: &AccountAddress,
         struct_tag: &StructTag,
         metadata: &[Metadata],
-    ) -> Result<Option<Vec<u8>>, VMError> {
+    ) -> Result<(Option<Vec<u8>>, Option<NumBytes>), VMError> {
```
vgao1996, line 79: Same here. Should merge the two options.

aptos-move/aptos-vm/src/data_cache.rs
```diff
@@ -42,16 +47,22 @@ pub(crate) fn get_resource_group_from_metadata(
 /// Adapter to convert a `StateView` into a `MoveResolverExt`.
 pub struct StorageAdapter<'a, S> {
     state_store: &'a S,
+    accurate_byte_count: Cell<bool>,
```
vgao1996, line 50: Why does this need to be a cell? It's set only during creation time, right?

vgao1996 (commented):
Some early feedback

aptos-move/aptos-vm/src/data_cache.rs
```diff
@@ -65,32 +76,40 @@ impl<'a, S: StateView> StorageAdapter<'a, S> {
...
+                let mut len = Some(NumBytes::new(group_data.len() as u64));
                 let group_data: BTreeMap<StructTag, Vec<u8>> = bcs::from_bytes(&group_data)
                     .map_err(|_| {
                         PartialVMError::new(StatusCode::UNKNOWN_INVARIANT_VIOLATION_ERROR)
                             .finish(Location::Undefined)
                     })?;
                 let res = group_data.get(struct_tag).cloned();
+                if !self.accurate_byte_count.get() {
+                    len = res.as_ref().map(|b| NumBytes::new(b.len() as u64));
+                }
```
vgao1996, line 102: Yeah, as we discussed offline, let's try to make the amount charged independent of the ordering by which you load the resources in the same group.

vgao1996:
Also a reminder that we need to take care of the lifetime issue

gerben-stavenga:
> Also a reminder that we need to take care of the lifetime issue we just discussed

I addressed the issue by releasing the cache after proloque. This was infrastructure that was already there for assembling resource groups for creating the writeset

aptos-move/aptos-vm/src/data_cache.rs
```diff
@@ -42,16 +42,22 @@ pub(crate) fn get_resource_group_from_metadata(
...
 impl<'a, S: StateView> StorageAdapter<'a, S> {
     pub fn new(state_store: &'a S) -> Self {
-        Self {
+        let mut s = Self {
             state_store,
+            accurate_byte_count: false,
             resource_group_cache: RefCell::new(BTreeMap::new()),
-        }
+        };
+        if gas_config(&s).1 >= 9 {
```
davidiw, line 57: you need to set the version to 9 in this PR too otherwise during upgrade it won't be set.

vgao1996 (approved):
@gerben-stavenga I think David made a good point about bumping LATEST_GAS_FEATURE_VERSION. Please do that before you land this!

sherry-x:
Triggering re-run failed job to confirm the performance numbers.

PR #8482, opened 2023-06-01
 [Sharding][Benchmark] Add block partitioning support for executor benchmark

sitalkedia:
### Description

Integrating block partitioning as a separate phase in executor benchmark.

### Test Plan

Ran the benchmark with a single shard and partitioning is a no-op.


sitalkedia:
Yes, part of this is already committed as separate PR.  There is still part of the functionality left to be integrated. Will open a separate PR for that.

PR #8483, opened 2023-06-01
[Faucet] Adjust faucet txn expiration / seqnum related values

banool:
### Description
Based on https://github.com/aptos-labs/aptos-core/actions/runs/5149399311/jobs/9272587458?pr=7533 it seems like, especially for use in CI, I lowered the txn expiration too low. This PR adjusts them. For the standard case I don't change the values completely back to what they were but when the faucet is used from the CLI (for a local testnet) they're back to what they were.

### Test Plan
Let's see how it goes in CI.


0xmaayan (approved):
❤️

0xjinn (approved):

PR #8484, opened 2023-06-01
[move-examples] Adding a whitelist move-example

xbtmatt:
# Modular Whitelist

This is a modular whitelist smart contract intended to gate an account's access to do "Something" based on time, price, and the # of times they've already done "Something". For example, an NFT minting contract with a whitelist.

The contract leverages objects by giving a whitelist object to the creator and using it to manage the whitelisting logistics.

The intention is to offer a modular whitelist that can be used by a separate smart contract without having to write too much boilerplate code.

The contract sorts the whitelist in ascending order by comparing each tier's start time. If the start time is equal, it compares the price. The contract automatically selects a mint tier based on the earliest, cheapest available tier in the sorted tier list for a minter.

## Setting up the whitelist
```rust
whitelist::init_tiers(owner);

whitelist::upsert_tier_config(
	owner,
	string::utf8(b"public"),
	true, // open_to_public, users don't need to be registered in the list
	PUBLIC_PRICE,
	PUBLIC_START_TIME,
	PUBLIC_END_TIME,
	PUBLIC_PER_USER_LIMIT,
);

whitelist::upsert_tier_config(
	owner,
	string::utf8(b"whitelist"),
	false, // open_to_public, users need to be registered in the whitelist
	WHITELIST_PRICE,
	WHITELIST_START_TIME,
	WHITELIST_END_TIME,
	WHITELIST_PER_USER_LIMIT,
);
```

## Using the whitelist to count the # of interactions
```rust
public entry fun mint(...) {
	// ... other minting logic

	// update the account's info in the corresponding tier
   whitelist::increment(owner, receiver);

	// ... other minting logic
}
```

## Clean up
```
// the owner of the whitelist object can destroy it
destroy(owner);
```

Whitelists are most commonly used with minting contracts, so we show an example that involves a very simple minting smart contract.

Unit tests in whitelist.move and mint.move can be run by specifying `@whitelist_example` as a named address.

```
aptos move test --named-addresses whitelist_example=default
```


aptos-move/move-examples/whitelist/sources/mint.move
```diff
@@ -0,0 +1,273 @@
...
+        });
+    }
+
+    public entry fun initialize_collection(
+        owner: &signer,
+    ) acquires MintConfiguration {
+        assert!(signer::address_of(owner) == @owner, error::permission_denied(ENOT_AUTHORIZED));
+
+        let mint_configuration = borrow_global<MintConfiguration>(@whitelist_example);
+        let resource_signer = &account::create_signer_with_capability(&mint_configuration.signer_capability);
```
movekevin, line 62: How about we keep the whitelist as an object instead? The collection (not the focus of this example anyway) can be created with the owner being the creator

xbtmatt: Ah yes, that would make a lot more sense...too used to using resource accounts

aptos-move/move-examples/whitelist/sources/mint.move
```diff
@@ -0,0 +1,273 @@
...
+            false, // open_to_public, users need to be registered in the whitelist
+            WHITELIST_PRICE,
+            WHITELIST_START_TIME,
+            WHITELIST_END_TIME,
+            WHITELIST_PER_USER_LIMIT,
+        );
+    }
+
+    /// simple mint function to demonstrate how to call the whitelist functions
+    public entry fun mint(receiver: &signer, tier_name: String) acquires MintConfiguration {
```
lightmark, line 107: do you expect the developer to switch function arguments for different phases/tiers? For example, when public mint starts, the front-end would change the `tier_name` to `public` manually? From my feeling this is not that ergonomic. Could we make tiers/phases into a ordered vector ordered by their start time and make sure the minting periods don't overlap? Then every time when you call mint, you don't need a name. but just check the first phase that has started but not ended yet. For example, you have
[0, 10)[10, 20)[20, 30)
when you call mint at 5, it will follow tier/phase 0. 11 will go to tier/phase 1, etc. 
An optimization is to reverse the order to remove the expired phase.

aptos-move/move-examples/whitelist/sources/mint.move
```diff
@@ -0,0 +1,273 @@
...
+    public entry fun remove_addresses_from_tier(
+        owner: &signer,
+        tier_name: String,
+        addresses: vector<address>,
+    ) acquires MintConfiguration {
+        assert!(signer::address_of(owner) == @owner, error::permission_denied(ENOT_AUTHORIZED));
+        let mint_configuration = borrow_global_mut<MintConfiguration>(@whitelist_example);
+        let resource_signer = &account::create_signer_with_capability(&mint_configuration.signer_capability);
+        whitelist::remove_addresses_from_tier(resource_signer, tier_name, addresses);
+    }
```
lightmark, line 176: Also, could you add a function to clean the whitelist object since it is ephemeral.

lightmark (commented):
If we use object for whitelist, I think it makes much more sense to publish the code to the original creator account since the code could be reused but whiltelist object will be cleaned up eventually so we don't want to publish code onto that object address.

aptos-move/move-examples/whitelist/sources/mint.move
xbtmatt: Yea, I originally didn't do this because I thought sometimes you may want more control over the tiers with overlapping times, but I actually don't think this is very common at all so it is a decrease in dx for poor reasons. I'll change it to use increasing, adjacent times

> An optimization is to reverse the order to remove the expired phase.

Not sure what you mean by this, but I am going to do something like this:

Only include starting times, assume all tiers are time-adjacent, and sort the vector by the latest starting time, in descending order (latest first). If the tier has started, it will use that, if it hasn't it will try to use the next one down

aptos-move/move-examples/whitelist/sources/mint.move
xbtmatt: Yea, will do!

aptos-move/move-examples/whitelist/sources/mint.move
xbtmatt: @movekevin you mean the publisher is also the creator of the contract and the owner of the whitelist object, yea? Just want to clarify

movekevin: The creator of the collection

aptos-move/move-examples/whitelist/sources/mint.move
lightmark: I think it totally depends on how do you want FE integrated with BE. In a traditional web2 design, all the phase transition happens on the BE side so for this case, I am not sure what's the best practice. Especially when you mentioned
> "Sometimes, whitelisting contracts will let a user continue to mint from an earlier (whitelist) phase even when a public phase has started."

That's very ambiguous when it comes to real nft projects. How do they deal with minting with different phases concurrently? I think it is very adhoc from case to case; While my perspective is always from web2 service experience. 
> Passing in the tier name isn't ideal, but whenever you're using a whitelist the front-end is almost always aware of the current phase, because you want to indicate to the user when transitioning from something that is free to something that costs something
Even the reason FE is aware of the phase is querying to the BE. Because FE can be tricked in many ways usually. Even you do that correct at FE, ppl can send their own SDK request by themselves (not via your FE code) to trigger some unfavorable behavior.

So if we cannot define proper behavior for overlapping phase, I don't understand how this design can be robust.

xbtmatt: > ppl can send their own SDK request by themselves (not via your FE code) to trigger some unfavorable behavior.

Ultimately the contract still enforces the internal rules about how many from each tier a user can mint. Even if a user bypasses the front end, it's still tracking the # of mints and will verify that they have mints left in that tier.

> So if we cannot define proper behavior for overlapping phase, I don't understand how this design can be robust.

My thinking was that it's up to the creator to determine how to display that information to the user, but yeah maybe that doesn't make sense, because the user is always going to want the cheapest one.

So overlapping is fine as long as the contract can determine which tier to mint from if multiple are available. If we assume the cheapest tier, we just need to sort the vector of tiers any time the price changes.

Then when the user mints, it would iterate over the sorted vector of each tier, cheapest first, to see if:

1. the phase has begun and
2. if the user has any mints left in that tier

I think this is similar to what you were saying earlier, but it would allow for overlapping and just choose the cheaper of the two phases.

lightmark: I don't really mind giving flexibility to the creator. My only concern here is that FE has to be updated when a new phase begin calling with a different phase key. If this is a normal operation, I am cool with the design.

lightmark (approved):

aptos-move/move-examples/whitelist/sources/mint.move
xbtmatt: I see what you're saying, yea, it is a little odd. I don't mind redoing it, I think assuming cheapest first makes sense too

xbtmatt: @lightmark I thought more about what we discussed here and I realized you were right- it doesn't actually make sense if the contract can't deterministically select what tier a user will mint from.

I implemented a simple insertion sort for the keys of the simple map that sorts by `start_time` and then `price` if there's a tie, so the user always mints from the earliest, cheapest mint tier that's available to them.

I didn't implement overlapping of tiers, nor did I check that they're contiguous (so you could have start-end times of [0, 10], [11, 20]), mostly just to leave the creator some flexibility when using the contract.

The developer/front end can also now just call `get_earliest_tier_available(creator_addr, minter_addr)` to view what `tier_name` the contract will mint from for the given minter address.

PR #8486, opened 2023-06-01
[mempool] classify client-submitted vs. broadcasted transactions

bchocho:
### Description

Added a new label for classifying whether a mempool-to-commit latency was measured on a transaction that was directly submitted by the client to this node (via REST api) or came to this node via mempool broadcast.

### DIFF
--- a/mempool/src/core_mempool/transaction.rs
+++ b/mempool/src/core_mempool/transaction.rs
@@ -45,11 +45,7 @@ impl MempoolTransaction {
             expiration_time,
             ranking_score,
             timeline_state,
-            insertion_info: InsertionInfo {
-                insertion_time,
-                client_submitted,
-                validator_end_to_end: timeline_state != TimelineState::NonQualified,
-            },
+            insertion_info: InsertionInfo::new(insertion_time, client_submitted, timeline_state),
             was_parked: false,
         }
     }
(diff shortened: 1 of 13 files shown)


mempool/src/core_mempool/transaction.rs
```diff
@@ -84,6 +89,13 @@ pub struct SequenceInfo {
     pub account_sequence_number: u64,
 }
 
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Hash)]
+pub struct InsertionInfo {
+    pub insertion_time: SystemTime,
+    pub client_submitted: bool,
+    pub is_end_to_end: bool,
```
JoshLind, line 96: nit: can we add a comment to this line? I'm a noob and I'm not sure what `is_end_to_end` means? 🤔

JoshLind (approved):
Solid. Thanks @bchocho 😄

mempool/src/core_mempool/transaction.rs
bchocho: Good call. I renamed is_end_to_end and added a comment to both of those, as unfortunately they're quite nuanced.

igor-aptos (changes requested):

mempool/src/core_mempool/transaction.rs
```diff
@@ -84,6 +89,41 @@ pub struct SequenceInfo {
...
+    }
+
+    pub fn validator_scope_label(&self) -> &'static str {
+        if self.validator_end_to_end {
+            counters::E2E_LABEL
+        } else {
+            counters::LOCAL_LABEL
+        }
+    }
+}
```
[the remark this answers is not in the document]
bchocho: Thanks, made the suggested change to an enum. Please take a look!

mempool/src/core_mempool/transaction.rs
```diff
@@ -84,6 +85,61 @@ pub struct SequenceInfo {
     pub account_sequence_number: u64,
 }
 
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Hash)]
+pub enum SubmittedBy {
+    /// The transaction was received from a client REST API submission, rather than a mempool
+    /// broadcast. This can be used as the time a transaction first entered the network,
+    /// to measure end-to-end latency within the entire network. However, if a transaction is
+    /// submitted to multiple nodes (by the client) then the end-to-end latency measured will not
+    /// be accurate.
```
igor-aptos, line 94: add in all comments after "will not be accurate" a comment ("will be lower than the correct value")

igor-aptos (approved):

Issue #6376, opened 2023-01-28
[Bug] Adopt inline functions in all framework modules


movekevin:
# 🐛 Bug

Inline functions such as map, filter, etc. have been introduced to all core data structures. We should update all framework modules to use them for conciseness + efficiency.

e99243506bigplay:
ah no, more gas consumption is incoming.

movekevin:
> ah no, more gas consumption is incoming.

??? Why do you say so?

e99243506bigplay:
> > ah no, more gas consumption is incoming.
> 
> Why do you say so?

Because inline functions could be black box and they are still not very optimized. they are not equal to macro function or you put code for inlining manually. If you could ensure the gas consumption about "current" function is better than previous version, then you are free to go.
https://github.com/aptos-foundation/AIPs/issues/33#issuecomment-1399097948

movekevin:
Inline functions should lead to the same gas cost as the expanded version (it's implicitly expanded).

e99243506bigplay:
#6507

wrwg:
@xindingw are you actually working on this? If not move to backlog

movekevin:
> @xindingw are you actually working on this? If not move to backlog

Yes he is currently :)

e99243506bigplay:
Notice: currently inline function has unnecessary overhead.
```
module me::test3 {
    /// Returns a * b / c going through u128 to prevent intermediate overflow
    public inline fun mul_div(a: u128, b: u128, c: u128): u128 {
        (((a as u256) * (b as u256) / (c as u256)) as u128)
    }
    public fun main(a: u128, b: u128, c: u128): u128 {
        mul_div(a, b, c) + mul_div(a, b, c) / mul_div(a, b, c) * mul_div(a, b, c)
    }
}
```
will be
```
// Move bytecode v6
module 0.test3 {


public main(Arg0: u128, Arg1: u128, Arg2: u128): u128 {
L0:	loc3: u128
L1:	loc4: u128
L2:	loc5: u128
L3:	loc6: u128
L4:	loc7: u128
L5:	loc8: u128
L6:	loc9: u128
L7:	loc10: u128
L8:	loc11: u128
B0:
	0: CopyLoc[0](Arg0: u128)
	1: CopyLoc[1](Arg1: u128)
	2: CopyLoc[2](Arg2: u128)
	3: StLoc[11](loc8: u128)
	4: StLoc[7](loc4: u128)
[...]

PR #8490, opened 2023-06-02
Refactor to replace vanilla loops with inline functions

xindingw:
### Description
Refactor for #6376.
Add `vector::enumerate_ref` and `vector::enumerate_mut` to loop with index. 

### Test Plan

Unit testing

### DIFF
--- a/aptos-move/framework/aptos-framework/sources/genesis.move
+++ b/aptos-move/framework/aptos-framework/sources/genesis.move
@@ -158,12 +158,9 @@ module aptos_framework::genesis {
     }
 
     fun create_accounts(aptos_framework: &signer, accounts: vector<AccountMap>) {
-        let i = 0;
-        let num_accounts = vector::length(&accounts);
         let unique_accounts = vector::empty();
-
-        while (i < num_accounts) {
-            let account_map = vector::borrow(&accounts, i);
+        vector::for_each_ref(&accounts, |account_map| {
+            let account_map: &AccountMap = account_map;
             assert!(
                 !vector::contains(&unique_accounts, &account_map.account_address),
                 error::already_exists(EDUPLICATE_ACCOUNT),
@@ -175,9 +172,7 @@ module aptos_framework::genesis {
                 account_map.account_address,
                 account_map.balance,
             );
-
-            i = i + 1;
-        };
+        });
     }
 
     /// This creates an funds an account if it doesn't exist.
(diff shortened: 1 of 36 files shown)


aptos-move/framework/aptos-framework/sources/genesis.move
```diff
@@ -315,21 +302,15 @@ module aptos_framework::genesis {
     /// Network address fields are a vector per account, where each entry is a vector of addresses
     /// encoded in a single BCS byte array.
     fun create_initialize_validators(aptos_framework: &signer, validators: vector<ValidatorConfiguration>) {
-        let i = 0;
-        let num_validators = vector::length(&validators);
-
         let validators_with_commission = vector::empty();
-
-        while (i < num_validators) {
+        vector::for_each_reverse(validators, |validator| {
```
movekevin, line 306: Why use for_each_reverse instead of for_each here?

aptos-move/framework/aptos-framework/sources/vesting.move
```diff
@@ -401,18 +401,17 @@ module aptos_framework::vesting {
...
-        while (i < len) {
-            let shareholder = *vector::borrow(shareholders, i);
-            // This will still return the shareholder if shareholder == beneficiary.
-            if (shareholder_or_beneficiary == get_beneficiary(vesting_contract, shareholder)) {
-                return shareholder
-            };
-            i = i + 1;
-        };
+        let result = @0x0;
+        vector::any(shareholders, |shareholder| {
```
movekevin, line 405: I wonder if we can add a find() function in vector that returns the ref to the first element that matches a predicate. I've seen this use case in other places updated in this PR as well

movekevin (approved):

aptos-move/framework/aptos-framework/sources/vesting.move
xindingw: I used consider this but didn't come up with a good solution. My rationales:
1. find() cannot return &Element because the element that matches the predicate might not exist.
2. find() cannot return Element due to ownership.
3. We cannot use `Option` because its implement is based on `vector`. I think it doesn't make sense though.
4. If we return the index of the element, we still need to compare with the length to check if we find something then `borrow` that element with the index. In this case `any` is even more convenient...

aptos-move/framework/aptos-framework/sources/genesis.move
xindingw: The loop is using `vector::pop_back(&mut validators)` so actually it iterates in the reverse order. In fact `for_each` is implemented with `for_each_reverse`([link](https://github.com/aptos-labs/aptos-core/blob/bd3eb542ef22a2067653417295a02a2fb571efd8/aptos-move/framework/move-stdlib/sources/vector.move#L243)). This is because of ownership. So `for_each_reverse` is more efficient.

aptos-move/framework/aptos-framework/sources/vesting.move
movekevin: I see that makes sense

.github/workflows/docker-build-test.yaml
```diff
@@ -171,43 +171,8 @@ jobs:
       FEATURES: consensus-only-perf-test
       BUILD_ADDL_TESTING_IMAGES: true
 
-  rust-images-all:
```
lightmark, line 174: why there are changes to this file?

movekevin: This is from a force push by Stelian. Check Slack more context 😄

lightmark (approved):

PR #8492, opened 2023-06-02
Fix docker-update-images workflow

gedigi:
### Description
This PR adapts the script to update Aptos debian and rust containers to the new setup.

### Test Plan


perryjrandall (approved):

sionescu (approved):
Looks OK, but I would suggest putting the image info into a json file, and sourcing that into the .hcl file. That way you can parse/edit the json without using a regex.

https://dev.to/lucassha/consuming-and-decoding-json-in-terraform-309p

gedigi:
> Looks OK, but I would suggest putting the image info into a json file, and sourcing that into the .hcl file. That way you can parse/edit the json without using a regex.
> 
> https://dev.to/lucassha/consuming-and-decoding-json-in-terraform-309p

Thanks for the suggestion @sionescu I'll do that!

PR #8499, opened 2023-06-02
[aptoswap] A uniswap-v2-style AMM core smart contract based on fungible asset

lightmark:
### Description
An AMM core smart contract using only fungible asset, based on uniswap v2.
1. add liquidity with two fungible assets
2. remove liquidity by redeeming lp tokens
3. do the swap with a fee.

I may add more peripheral code gradually if needed.

### Test Plan
aptos move test


aptos-move/move-examples/defi/aptoswap/core/sources/liquidity_pool.move
```diff
@@ -0,0 +1,417 @@
...
+                string::utf8(b"http://aptoswap.com/favicon.ico"),
+                string::utf8(b"http://aptoswap.com/")
+            );
+            (
+                fungible_asset::generate_mint_ref(cref),
+                fungible_asset::generate_burn_ref(cref),
+            )
+        };
+
+        let x_store = fungible_asset::create_store(&object::create_object_from_object(pool_signer), x);
```
0xbe1, line 101: any chance adding docs/diagram to explain the object layout, given readers are not very familiar with these object apis including:

```
let pool_cref = &object::create_object_from_account(pool_manager_signer);
...
let x_store = fungible_asset::create_store(&object::create_object_from_object(pool_signer), x);
```

the main goal is to help users understand what are the different objects and what resources do they hold. it seems to be a bit difficult to reason about the layout compared to the vanilla account model that we're quite familiar with

aptos-move/move-examples/defi/aptoswap/core/sources/liquidity_pool.move
```diff
@@ -0,0 +1,417 @@
...
+    public fun register(acc: &signer, x: Object<Metadata>, y: Object<Metadata>) acquires LiquidityPoolManager {
+        if (!is_valid_pair(x, y)) {
+            return register(acc, y, x)
+        };
+        assert!(!lp_exists(x, y), error::already_exists(ERR_POOL_EXISTENCE));
+
+        let (lp_token_name, lp_token_symbol) = utilities::swap_lp_token_name_and_symbol(x, y);
+
+        let pool_manager_signer = &get_pool_account_signer();
+        let pool_cref = &object::create_object_from_account(pool_manager_signer);
```
0xbe1, line 74: i wonder when should we choose:
- create_object_from_account
- create_named_object
- create_object_from_object
?

aptos-move/move-examples/defi/aptoswap/init/sources/pool_manager.move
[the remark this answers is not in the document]
lightmark: cuz it may be deployed separately.

aptos-move/move-examples/defi/aptoswap/core/sources/liquidity_pool.move
```diff
@@ -0,0 +1,417 @@
...
+        lp_token_mint_ref: fungible_asset::MintRef,
+        lp_token_burn_ref: fungible_asset::BurnRef,
+        fee: u64,
+        // basis points
+    }
+
+    /// The onchain index of the liquidity pools of different pairs
+    struct LiquidityPoolManager has key {
+        extend_ref: ExtendRef,
+        map: SmartTable<vector<Object<Metadata>>, Object<LiquidityPool>>
```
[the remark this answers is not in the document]
lightmark: I can use pair as a key. Just the creation is more verbose but the vector is just `vector[a, b]`

aptos-move/move-examples/defi/aptoswap/core/sources/liquidity_pool.move
lightmark: You are right. The object is actually a bag of things instead of one thing. cc @movekevin

aptos-move/move-examples/defi/aptoswap/core/sources/liquidity_pool.move
lightmark: 1. create_object_from_account -> your new object owner is an account.
2. create_named_object -> your new object need deterministic naming, such as primary fungible store, or cannot be deleted, such as metadata object based on which we issue fungible assets.
3. create_object_from_object -> similar to 1, but the owner now becomes an account.
1 or 3 doesn't matter because you can always create an object from either can transfer the ownership. So the comparison is actually `named` vs `randomly generated`

aptos-move/move-examples/defi/aptoswap/init/sources/pool_manager.move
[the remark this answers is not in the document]
0xbe1: how does deploying separately help?

aptos-move/move-examples/defi/aptoswap/core/sources/liquidity_pool.move
0xbe1: @lightmark why does object ownership matter - i guess that's a follow-up question

aptos-move/move-examples/defi/aptoswap/init/sources/pool_manager.move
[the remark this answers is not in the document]
lightmark: Hmmm, it is a cryptic topic that how to make a `contract account` in aptos move.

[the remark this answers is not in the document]
0xbe1: @lightmark I am under the impression that liquidswap does so in order to deploy the amm under a resource account. Thala Swap also deploys the amm under a resource account but we manage to have only 1 rather than 2 packages - is this something that you're interested at?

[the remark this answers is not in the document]
lightmark: liquidswap put the data under the resource account but the code is deployed at the admin address IIRC. I have to read your code before making the statement... Now there is some tricky part that how to make both code and data under resource account cleanly. I think Kevin was considering a writeup.

aptos-move/move-examples/defi/aptoswap/core/sources/liquidity_pool.move
lightmark: let's assume you have a `FungibleStore` object, owner has the authority to withdraw. That's why ownership matters. you can build the permission model in your way, but the defacto pattern is the ownership model, at least for fungible asset store.

aptos-move/move-examples/defi/aptoswap/core/sources/liquidity_pool.move
```diff
@@ -0,0 +1,417 @@
...
+    use aptos_framework::fungible_asset::{Metadata, FungibleStore, FungibleAsset, deposit};
+    use aptos_framework::object::{Object, ExtendRef, generate_signer_for_extending, generate_extend_ref};
+    use aptos_framework::fungible_asset;
+    use aptos_framework::object;
+    use aptos_framework::primary_fungible_store;
+    use aptos_std::comparator;
+    use aptos_std::math128;
+    use aptos_std::math64;
+    use aptos_std::smart_table::SmartTable;
+    use aptos_std::smart_table;
```
movekevin, line 11: nit: Merge with line 10. We should probably file a request in the intellij Move plugin repo to request automatic import merging

aptos-move/move-examples/defi/aptoswap/core/sources/liquidity_pool.move
```diff
@@ -0,0 +1,417 @@
...
+    use std::vector;
+
+    use pool_manager::utilities;
+
+    const MINIMAL_LIQUIDITY: u64 = 1000;
+
+    // fee is basis point based so the scale is 10000.
+    const FEE_SCALE: u64 = 10000;
+
+    const ERR_SWAP_PAIR_CANNOT_BE_THE_SAME: u64 = 1;
```
movekevin, line 25: minor: Add error messages for all these errors

aptos-move/move-examples/defi/aptoswap/core/sources/liquidity_pool.move
```diff
@@ -0,0 +1,417 @@
+module pool_manager::liquidity_pool {
```
movekevin, line 1: Can you add more documentation for this module? We should also write a tutorial to walk people through this

aptos-move/move-examples/defi/aptoswap/core/sources/liquidity_pool.move
```diff
@@ -0,0 +1,417 @@
...
+
+    const ERR_SWAP_PAIR_CANNOT_BE_THE_SAME: u64 = 1;
+    const ERR_POOL_EXISTENCE: u64 = 2;
+    const ERR_NOT_ENOUGH_LIQUIDITY: u64 = 3;
+    const ERR_NOT_ENOUGH_LP_TOKEN_TO_BURN: u64 = 4;
+    const ERR_ZERO_COIN_TO_SWAP: u64 = 5;
+    const ERR_INCORRECT_SWAP_DUE_TO_REDUCED_LIQUIDITY: u64 = 6;
+
+    /// Stored at lp token metadata object to find the liquidity pool.
+    /// x address must be lexicographically smaller than y address.
```
movekevin, line 33: Is this still needed for fungible assets outside of Solidity and CoinType?

aptos-move/move-examples/defi/aptoswap/core/sources/liquidity_pool.move
```diff
@@ -0,0 +1,417 @@
...
+        let extend_ref = aptoswap::init::retrieve_extend_ref(admin);
+        let pool_manager_obj_signer = object::generate_signer_for_extending(&extend_ref);
+        move_to(&pool_manager_obj_signer, LiquidityPoolManager {
+            extend_ref,
+            map: smart_table::new()
+        })
+    }
+
+    /// Register liquidity pool `X`/`Y`.
+    public fun register(acc: &signer, x: Object<Metadata>, y: Object<Metadata>) acquires LiquidityPoolManager {
```
movekevin, line 65: nit: create might be a better name

aptos-move/move-examples/defi/aptoswap/core/sources/utilities.move
```diff
@@ -0,0 +1,20 @@
+module pool_manager::utilities {
+    use aptos_framework::object::Object;
+    use aptos_framework::fungible_asset::Metadata;
+    use std::string::String;
+    use aptos_framework::fungible_asset;
+    use std::string;
+
+    friend pool_manager::liquidity_pool;
+
+    public(friend) fun swap_lp_token_name_and_symbol(x: Object<Metadata>, y: Object<Metadata>): (String, String) {
```
movekevin, line 10: nit: remove swap_. It makes me think this is swapping something

aptos-move/move-examples/defi/aptoswap/core/sources/utilities.move
```diff
@@ -0,0 +1,20 @@
...
+        let name = fungible_asset::name(x);
```
movekevin, line 11: You can now use the cool functions in aptos_std::string_utils to do all this more easily

aptos-move/move-examples/defi/aptoswap/core/sources/liquidity_pool.move
movekevin, line 74: My vote is on the new create function being introduced in https://github.com/aptos-labs/aptos-core/pull/8401

aptos-move/move-examples/defi/aptoswap/core/sources/liquidity_pool.move
```diff
@@ -0,0 +1,417 @@
...
+        };
+        assert!(!lp_exists(x, y), error::already_exists(ERR_POOL_EXISTENCE));
+
+        let (lp_token_name, lp_token_symbol) = utilities::swap_lp_token_name_and_symbol(x, y);
+
+        let pool_manager_signer = &get_pool_account_signer();
+        let pool_cref = &object::create_object_from_account(pool_manager_signer);
+        let pool_signer = &object::generate_signer(pool_cref);
+
+        // Create lp fungible token metadata object.
```
movekevin, line 77: I wonder if we can combine the LP token metadata object and the liquidity pool object into a single object. This would be fewer addresses to remember. It also logically makes sense that the liquidity pool and the LP token are the same. What do you think @0xbe1

movekevin (approved):
Great example overall! I think this just needs more documentation + tutorial

aptos-move/move-examples/defi/aptoswap/core/sources/liquidity_pool.move
0xbe1: @movekevin do you mean merging `LiquidityPool` and `SwapPair`?

> I wonder if we can combine the LP token metadata object and the liquidity pool object into a single object.

Otherwise I think `LiquidityPool` already composes both LP token metadata and liquidity pool data

aptos-move/move-examples/defi/aptoswap/core/sources/liquidity_pool.move
lightmark: yeah, I never noticed this cuz it's folded.

aptos-move/move-examples/defi/aptoswap/core/sources/liquidity_pool.move
lightmark: (a, b) and (b, a) should share a unique id.

aptos-move/move-examples/defi/aptoswap/core/sources/liquidity_pool.move
lightmark: It didn't compose the metadata. Now it does as I re-pushed.

aptos-move/move-examples/defi/aptoswap/core/sources/liquidity_pool.move
movekevin: But why?

aptos-move/move-examples/defi/aptoswap/core/sources/liquidity_pool.move
```diff
@@ -0,0 +1,442 @@
+/// This module demonstrates how to create a core module of AMM following uniswap v2 algorithm with Aptos fungible
```
movekevin, line 1: I'd replace "uniswap v2" with constant product

aptos-move/move-examples/defi/aptoswap/init/sources/pool_manager.move
```diff
@@ -0,0 +1,43 @@
...
+        admin: &signer,
+        code_metadata_serialized: vector<u8>,
+        code: vector<u8>
+    ) {
+        assert!(signer::address_of(admin) == @aptoswap, std::error::permission_denied(ERR_NOT_APTOSWAP_ADMIN));
+
+        let contructor_ref = &object::create_named_object(admin, b"aptoswap_pool_manager_seed");
+        let obj_signer = object::generate_signer(contructor_ref);
+
+        aptos_framework::code::publish_package_txn(
```
movekevin, line 24: nit: Why not import code?

aptos-move/move-examples/defi/aptoswap/init/sources/pool_manager.move
```diff
@@ -0,0 +1,43 @@
+module aptoswap::init {
```
movekevin, line 1: This can be done in a followup but we should make a separate example for the object-based deployment flow

movekevin (approved):

aptos-move/move-examples/defi/aptoswap/init/sources/pool_manager.move
lightmark: wdym?

movekevin: use aptos_framework::code;
code::publish...

aptos-move/move-examples/defi/aptoswap/init/sources/pool_manager.move
```diff
@@ -0,0 +1,43 @@
...
+    const ERR_NOT_APTOSWAP_ADMIN: u64 = 1;
+    const ERR_EXTEND_REF_NOT_MATCH: u64 = 2;
+
+    struct PoolManagerExtendRef has key {
+        ref: ExtendRef
+    }
+
+    /// Creates new object for Aptoswap to be the pool manager object where the lp code will be deployed at.
+    /// Can only be executed only from Aptoswap admin account.
+    public entry fun initialize_lp_account(
```
movekevin, line 14: This is a strange flow. Why create a different lp object every time, store ExtendRef in admin account, and then retrieve + take over ExtendRef later?

Why not allow liquidity_pool to call this directly to create a liquidity pool object, then still do the management here (store extend ref, provide a signer for extension that's a friend only function and can be called by modules in the same package)?

aptos-move/move-examples/defi/aptoswap/core/sources/liquidity_pool.move
```diff
@@ -0,0 +1,442 @@
...
+    public fun create(acc: &signer, x: Object<Metadata>, y: Object<Metadata>) acquires LiquidityPoolManager {
+        if (!is_valid_pair(x, y)) {
+            return create(acc, y, x)
+        };
+        assert!(option::is_none(&get_liquidity_pool(x, y)), error::already_exists(ERR_POOL_EXISTENCE));
+
+        let (lp_token_name, lp_token_symbol) = lp_token_name_and_symbol(x, y);
+
+        let pool_manager_signer = &get_pool_account_signer();
+        let pool_cref = &object::create_named_object(pool_manager_signer, pool_address_seed(x, y));
```
movekevin, line 73: If we'recreating named objects for the liquidity pool anyway, do we still need the on-chain map? (LiquidityPoolManager)

aptos-move/move-examples/defi/aptoswap/core/sources/liquidity_pool.move
```diff
@@ -0,0 +1,442 @@
...
+        let y_value = fungible_asset::amount(&coin_y);
+
+        let lp_token_supply = lp_token_supply(pool);
+        let provided_liq = if (lp_token_supply == 0) {
+            let initial_liq = (math128::sqrt((x_value as u128) * (y_value as u128)) as u64);
+            let min_liquidity = fungible_asset::mint(&pool.lp_token_mint_ref, MINIMAL_LIQUIDITY);
+            primary_fungible_store::deposit(object::address_from_extend_ref(&pool.extend_ref), min_liquidity);
+            initial_liq - MINIMAL_LIQUIDITY
+        } else {
+            let x_liq = math64::mul_div(x_value, (lp_token_supply as u64), x_reserve_size);
```
movekevin, line 140: Are you planning on writing the router code as well? Without it, this is a possible foot gun as you would only receive the smaller amount based on constant product

aptos-move/move-examples/defi/aptoswap/core/sources/liquidity_pool.move
```diff
@@ -0,0 +1,442 @@
...
+        let y_reserve_size = fungible_asset::balance(y_store);
+
+        let x_value = fungible_asset::amount(&coin_x);
+        let y_value = fungible_asset::amount(&coin_y);
+
+        let lp_token_supply = lp_token_supply(pool);
+        let provided_liq = if (lp_token_supply == 0) {
+            let initial_liq = (math128::sqrt((x_value as u128) * (y_value as u128)) as u64);
+            let min_liquidity = fungible_asset::mint(&pool.lp_token_mint_ref, MINIMAL_LIQUIDITY);
+            primary_fungible_store::deposit(object::address_from_extend_ref(&pool.extend_ref), min_liquidity);
```
movekevin, line 137: These 2 lines can be combined with fungible_asset::mint_to

aptos-move/move-examples/defi/aptoswap/core/sources/liquidity_pool.move
```diff
@@ -0,0 +1,442 @@
...
+        let from = fungible_asset::metadata_from_asset(&in);
+        let (from_index, pool) = if (is_valid_pair(from, to)) {
+            (0, borrow_lp(from, to))
+        } else {
+            (1, borrow_lp(to, from))
+        };
+        let from_reserve = fungible_asset::balance(*vector::borrow(&pool.store_pair, from_index));
+        let to_reserve = fungible_asset::balance(*vector::borrow(&pool.store_pair, 1 - from_index));
+        let k_before_swap = (from_reserve as u256) * (to_reserve as u256) * (FEE_SCALE * FEE_SCALE as u256);
+        let (out, k_after_swap_and_fee) = swap_and_calculate_new_k(pool, from_index, in, to_amount);
```
movekevin, line 212: This doesn't seem right. What is to_amount for? Technically, to can be calculated by maintaining the product k, so to_amount is not needed. The only reason you'd often these in DEXes is for users to provide a min_out as a way to indicate max slippage tolerated (if slippage is too high, you can get a lot less back than expected so providing a min guarantees the tx would fail if you receive less than that)

aptos-move/framework/aptos-framework/sources/object.move
```diff
@@ -537,6 +537,14 @@ module aptos_framework::object {
...
     }
 
+    #[test_only]
+    public fun create_constructor_ref(s: &signer, can_delete: bool): ConstructorRef {
+        ConstructorRef {
+            self: signer::address_of(s),
+            can_delete
+        }
+    }
+
```
davidiw, line 547: should this be updated now that you add sticky objects?

davidiw (commented):
status update?

PR #8500, opened 2023-06-02
[gha][docker] only push to GAR on PR; push to ECR on postcommit

rustielin:
### Description

`TARGET_REGISTRY` is determined in GHA based on the workflow trigger, and passed through to the docker bake file, which determines the image (tags) to push.

`TARGET_REGISTRY` can be:
* local -- doesn't push
* remote -- just the default, pushing to all supported internal remote docker repos
* gcp -- just to GAR

This saves us some time during the post-build image export phase, around 2min. So around 15% cost savings on the docker build pipeline

### Test Plan

```
# targeting just GCP
gh workflow run workflow-run-docker-rust-build.yaml --ref rustielin/docker-images-gar-only-on-pr -F=GIT_SHA=$(git rev-parse HEAD)  -F=TARGET_REGISTRY=gcp 
```
https://github.com/aptos-labs/aptos-core/actions/runs/5158877766/jobs/9293093246

```
# targeting all remotes
gh workflow run workflow-run-docker-rust-build.yaml --ref rustielin/docker-images-gar-only-on-pr -F=GIT_SHA=$(git rev-parse HEAD)
```
https://github.com/aptos-labs/aptos-core/actions/runs/5158863097/jobs/9293061340

### DIFF
--- a/.github/workflows/workflow-run-docker-rust-build.yaml
+++ b/.github/workflows/workflow-run-docker-rust-build.yaml
@@ -25,6 +25,12 @@ on:
         required: false
         type: boolean
         description: Whether to build additional testing images. If not specified, only the base release images will be built
+      TARGET_REGISTRY:
+        default: remote
+        required: false
+        type: string
+        description: The target docker registry to push to
+
   workflow_dispatch:
     inputs:
       GIT_SHA:
@@ -45,6 +51,11 @@ on:
         required: false
         type: boolean
         description: Whether to build additional testing images. If not specified, only the base release images will be built
+      TARGET_REGISTRY:
+        default: remote
+        required: false
+        type: string
+        description: The target docker registry to push to
 
 env:
   GIT_SHA: ${{ inputs.GIT_SHA }}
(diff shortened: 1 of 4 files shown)


.github/workflows/workflow-run-docker-rust-build.yaml
```diff
@@ -45,6 +45,11 @@ on:
         required: false
         type: boolean
         description: Whether to build additional testing images. If not specified, only the base release images will be built
+      TARGET_REGISTRY:
```
perryjrandall, line 48: Can this be choice or enum?

Can you validate

Nit: instead of remote call it all?

perryjrandall:
Can we figure out a better test plan?

perryjrandall (changes requested):

.github/workflows/workflow-run-docker-rust-build.yaml
rustielin: iirc  `choice` is only available in `workflow_dispatch` and not `workflow_call` or something like that

perryjrandall: RIP

https://docs.github.com/en/enterprise-cloud@latest/actions/using-workflows/workflow-syntax-for-github-actions#onworkflow_callinputsinput_idtype

Alternatively can we validate its valid in a step? that way you can fail early if the value doesnt make sense?

perryjrandall:
Do we have a rollout / revert plan?

perryjrandall (approved):
nice test plan, consider rollout / revert plan

ibalajiarun (approved):

sionescu (approved):

rustielin:
For rollout, the new `TARGET_REGISTRY` options are backwards compatible with the existing `remote` option, so nothing needs to be done  there.

For revert, we can just revert this PR (and announce to all branches that have included this change)

PR #8501, opened 2023-06-02
Add functions to extract all keys and values from simple_map and (token v1's property_map)

movekevin:
### Description
These functions are useful to have

### Test Plan
Unit tests

### DIFF
--- a/aptos-move/framework/aptos-stdlib/sources/simple_map.move
+++ b/aptos-move/framework/aptos-stdlib/sources/simple_map.move
@@ -100,6 +100,22 @@ module aptos_std::simple_map {
         (std::option::none(), std::option::none())
     }
 
+    /// Return all keys in the map. This requires keys to be copyable.
+    public fun keys<Key: copy, Value>(map: &SimpleMap<Key, Value>): vector<Key> {
+        vector::map_ref(&map.data, |e| {
+            let e: &Element<Key, Value> = e;
+            e.key
+        })
+    }
+
+    /// Return all values in the map. This requires values to be copyable.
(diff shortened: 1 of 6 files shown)


gerben-stavenga (approved):

aptos-move/framework/aptos-stdlib/sources/simple_map.move
```diff
@@ -183,6 +199,26 @@ module aptos_std::simple_map {
...
+    }
+
+    #[test]
+    public fun test_values() {
+        let map = create<u64, u64>();
+        assert!(values(&map) == vector[], 0);
+        add(&mut map, 2, 1);
+        add(&mut map, 3, 1);
+
+        assert!(values(&map) == vector[1, 1], 0);
```
xindingw, line 219: Can you test with difference values? In this case we cannot tell if all 1 come from the same entry.

lightmark (approved):

aptos-move/framework/aptos-stdlib/sources/simple_map.move
movekevin: Done

xindingw (approved):

aptos-move/framework/aptos-stdlib/sources/simple_map.move
```diff
@@ -100,6 +100,22 @@ module aptos_std::simple_map {
         (std::option::none(), std::option::none())
     }
 
+    /// Return all keys in the map. This requires keys to be copyable.
```
banool, line 103: lol seems familiar

banool (approved):

PR #8515, opened 2023-06-04
fix nvmrc version and futures version

geekflyer:
### Description

to things:
1. fix npm version in npmrc. The `n` there just to be a typo of some sort
2. Unpin `futures` version. This is because I'm indirectly using futures in the api-gateway (another repo) and I found a version incompatibility due to futures being hard-pinned in aptos-core. This removes the hard pinning which seems kinda unnecessary. This won't actually cause any differences in the effective futures version of aptos-core due to the lockfile still pinning this to `.24`, so this is a relatively safe change by itself.


davidiw (approved):

davidiw:
closing out... feel free to re-open

PR #8460, opened 2023-06-01
[CLI] Add customer header checking to e2e

0xjinn:
### Description

Add custom header "aptos-cli" checking e2e

### Test Plan

```
➜  e2e git:(jin/add_e2e_for_aptos_header) ✗ DOCKER_DEFAULT_PLATFORM=linux/amd64 poetry run python main.py --base-network testnet --test-cli-path ~/Projects/aptos-core/target/debug/aptos
2023-06-01 15:18:34,453 - INFO - Trying to run aptos CLI local testnet from image: aptoslabs/tools:testnet
testnet: Pulling from aptoslabs/tools
Digest: sha256:e2cd216dcd7225546ba47caadd8d6cbe34508028988eba474145bd3032320c15
Status: Image is up to date for aptoslabs/tools:testnet
2023-06-01 15:18:42,202 - INFO - Running aptos CLI local testnet from image: aptoslabs/tools:testnet
2023-06-01 15:18:42,202 - INFO - Waiting for node and faucet APIs for aptos-tools-testnet to come up
2023-06-01 15:18:58,505 - INFO - Node and faucet APIs for aptos-tools-testnet came up
2023-06-01 15:18:58,522 - INFO - Running test: test_metrics_accessible
2023-06-01 15:18:58,605 - INFO - Running test: test_init
2023-06-01 15:19:00,234 - INFO - Running test: test_account_fund_with_faucet
2023-06-01 15:19:01,512 - INFO - Running test: test_account_create
2023-06-01 15:19:02,874 - INFO - Running test: test_aptos_header_included
2023-06-01 15:19:02,912 - INFO - Stopping container: aptos-tools-testnet
2023-06-01 15:19:03,629 - INFO - Stopped container: aptos-tools-testnet
2023-06-01 15:19:03,629 - INFO - These tests passed:
2023-06-01 15:19:03,629 - INFO - test_metrics_accessible
2023-06-01 15:19:03,629 - INFO - test_init
2023-06-01 15:19:03,629 - INFO - test_account_fund_with_faucet
2023-06-01 15:19:03,629 - INFO - test_account_create
2023-06-01 15:19:03,629 - INFO - test_aptos_header_included
2023-06-01 15:19:03,629 - INFO - All tests passed!
```

PR #8517, opened 2023-06-04
Revert "Revert "[CLI] Add customer header checking to e2e (#8460)" (#8496)"

banool:
This reverts commit 56aee8ba47c1f0dbd4ad8cd11bfc1fc3eddd7d8e.


crates/aptos/e2e/main.py
```diff
@@ -118,28 +121,29 @@ def main():
...
-
-    # Prepare the run helper. This ensures in advance that everything needed is there.
-    run_helper.prepare()
-
-    # Run tests.
-    run_tests(run_helper)
-
-    # Stop the node + faucet.
-    stop_node(container_name)
+    try:
```
0xjinn, line 124: Should we catch it as well, so we know what went wrong?

0xjinn:
Curious, what is that "nightly" tag?

crates/aptos/e2e/main.py
banool: We'll see what went wrong from the exception being raised, same as before. It's just this way the stop_node function runs unconditionally.

```
>>> try:
...     raise RuntimeError("hey")
... finally:
...     print("finally")
...
finally
Traceback (most recent call last):
  File "<stdin>", line 2, in <module>
RuntimeError: hey
```

banool:
I just wanted to use a recent CLI without building it myself so I used the nightly tag. This is an image that gets built once every day.

0xjinn (approved):

movekevin (approved):

PR #8525, opened 2023-06-05
[Aptos Framework] Add multisig v2 owner helpers, abstractions

alnoki:
@davidiw @movekevin @wrwg 

This PR includes modifications to `multisig_account.move` with additional autogen file changes.

Functionality includes assorted convenience functions for multisig v2 owner schema modifications, and an abstracted inner function (`update_owner_schema`) in accordance with [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) design principles)

### DIFF
--- a/aptos-move/framework/aptos-framework/sources/multisig_account.move
+++ b/aptos-move/framework/aptos-framework/sources/multisig_account.move
@@ -576,20 +576,6 @@ module aptos_framework::multisig_account {
         );
     }
 
-    /// Update the number of signatures required then remove owners, in a single operation.
-    entry fun remove_owners_and_update_signatures_required(
-        multisig_account: &signer,
-        owners_to_remove: vector<address>,
-        new_num_signatures_required: u64
-    ) acquires MultisigAccount {
-        update_owner_schema(
-            address_of(multisig_account),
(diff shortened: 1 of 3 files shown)


aptos-move/framework/aptos-framework/sources/multisig_account.move
```diff
@@ -919,15 +947,31 @@ module aptos_framework::multisig_account {
...
-            assert!(!found, error::invalid_argument(EDUPLICATE_OWNER));
-            vector::push_back(&mut distinct_owners, owner);
-        });
+    fun validate_owners(
+        owners_ref: &vector<address>,
+        multisig_address: address
+    ) {
+        assert!(
+            !vector::contains(owners_ref, &multisig_address),
+            EOWNER_CANNOT_BE_MULTISIG_ACCOUNT_ITSELF
```
movekevin, line 956: Can you use error::invalid_argument wrapper here to be consistent as we do that everywhere in the framework?

movekevin (approved):
Can you add a quick AIP for this (should be separate from the one we might have for t he flush discussion)?

xindingw (approved):

PR #8528, opened 2023-06-05
remove QS

thepomeranian:


igor-aptos (approved):

igor-aptos (changes requested):
not needed on main branch, per perryjrandall

perryjrandall:
> not needed on main branch, per perryjrandall

We may want to keep this on in main since we also want to test it in main on devnet cut etc

Unless we decide for some reason we need to turn it off everywhere

PR #8536, opened 2023-06-06
Update release builder to use the correct reconfigure function

movekevin:
### Description
The current feature flags component of release builder is using the wrong (newly added) reconfigure function. We have an existing one in aptos_governance that was added before mainnet for precisely this purpose and should be used instead.


perryjrandall (approved):
How is this tested?

Can you set / unset some flags on devnet?

movekevin:
> How is this tested?
> 
> Can you set / unset some flags on devnet?

We'll test this in testnet/mainnet soon for enabling sig checker v2

runtian-zhou (approved):

PR #8544, opened 2023-06-06
Add a new function to allow migrating existings account to multisig account while immediately revoking auth key access

movekevin:
### Description
This provides a single convenient function so auth key rotation/revocation doesn't need to be done in a separate transaction

### Test Plan
Unit tests

### DIFF
--- a/aptos-move/framework/aptos-framework/sources/multisig_account.move
+++ b/aptos-move/framework/aptos-framework/sources/multisig_account.move
@@ -94,6 +94,8 @@ module aptos_framework::multisig_account {
     /// Provided owners to remove and new owners overlap.
     const EOWNERS_TO_REMOVE_NEW_OWNERS_OVERLAP: u64 = 18;
 
+    const ZERO_AUTH_KEY: vector<u8> = x"0000000000000000000000000000000000000000000000000000000000000000";
+
     /// Represents a multisig account's configurations and transactions.
     /// This will be stored in the multisig account (created as a resource account separate from any owner accounts).
     struct MultisigAccount has key {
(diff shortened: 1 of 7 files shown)


aptos-move/framework/aptos-framework/sources/multisig_account.move
```diff
@@ -413,6 +419,46 @@ module aptos_framework::multisig_account {
         );
     }
 
+    /// Creates a new multisig account on top of an existing account and immediately rotate the origin auth key to 0x0.
+    ///
+    /// Note: If the original account is a resource account, this does not revoke all control over it as if any
+    /// SignerCapability of the resource account still exists, it can still be used to generate the signer for the
```
lightmark, line 425: How is it possible? I think resource_account cannot generate create_multisig_account_signed_message because nobody has the private key?

lightmark (approved):

aptos-move/framework/aptos-framework/sources/multisig_account.move
movekevin: The resource account could have had the auth key set to an actual key (via the flow in resource_account.move) that they used to to create rotation or signer capability offers

aptos-move/framework/aptos-framework/sources/multisig_account.move
```diff
@@ -415,6 +436,64 @@ module aptos_framework::multisig_account {
...
+    /// account.
+    public entry fun create_with_existing_account_and_revoke_auth_key(
```
alinush, line 444: Can you explain what the `metadata_*` args are?

alinush (approved):
LGTM

PR #8545, opened 2023-06-06
[Sharding] [ Execution] Remote Executor service support

sitalkedia:
### Description

- This PR adds a support for running execution (block-stm) remotely in a separate process (or in the same process but in a separate thread for testing).  The network calls currently are blocking, which means that the server thread is blocked until the execution finishes - will add non blocking support as a followup. 
- Integrate the block-stm benchmark to add support for remote execution

### Test Plan
- Added unit tests
-  Tested by running a executor service locally with the benchmark. 

```
 cargo run --release -p  aptos-executor-service
```

```
cargo run --release -p  aptos-transaction-benchmarks -- execute --concurrency-level-per-shard 8 --num-executor-shards 10 --num-blocks 15  --block-size 100000 --remote-executor-addresses 127.0.0.1:8080
```


storage/state-view/src/lib.rs
```diff
@@ -47,6 +51,10 @@ pub trait TStateView {
 
     /// Get state storage usage info at epoch ending.
     fn get_usage(&self) -> Result<StateStorageUsage>;
+
+    fn as_in_memory_state_view(&self) -> InMemoryStateView {
+        unreachable!("in-memory state view conversion not supported yet")
+    }
```
msmouse, line 57: hmm.. if possible I'd avoid adding this to the trait.

third_party/move/move-core/types/src/vm_status.rs
```diff
@@ -756,6 +756,12 @@ pub enum StatusCode {
     RESERVED_RUNTIME_ERROR_4 = 4036,
     RESERVED_RUNTIME_ERROR_5 = 4037,
 
+    // Error happens due to error in communicating to the remote executor service
+    REMOTE_EXECUTION_SERVER_READ_ERROR = 5000,
+    REMOTE_EXECUTION_SERVER_WRITE_ERROR = 5001,
+    REMOTE_REQUEST_SERIALIZATION_ERROR = 5002,
+    REMOTE_RESPONSE_DESERIALIZATION_ERROR = 5003,
+
```
msmouse, line 764: Not sure we want to introduce these. Especially not in the 5xxx range, see the error code categorization code below. 
These errors are supposed to be part of the VM transaction output, I assume it's irrelevant to your change anyway?

execution/executor-service/src/remote_executor_client.rs
```diff
@@ -0,0 +1,85 @@
...
+        let input_message = bcs::to_bytes(&input).map_err(|e| {
+            VMStatus::Error(
+                REMOTE_EXECUTION_SERVER_WRITE_ERROR,
+                Some(format!(
+                    "Failed to serialize request to remote execution server: {}",
+                    e
+                )),
+            )
+        })?;
+        loop {
```
msmouse, line 65: this is a bit brutal.. shall we add limited retries? or just error out.

aptos-move/aptos-vm/src/sharded_block_executor/block_executor_client.rs
```diff
@@ -0,0 +1,62 @@
...
+use std::sync::Arc;
+
+pub trait TBlockExecutorClient {
+    fn execute_block<S: StateView + Sync>(
+        &self,
+        transactions: Vec<Transaction>,
+        state_view: &S,
+        concurrency_level: usize,
+        maybe_block_gas_limit: Option<u64>,
+    ) -> Result<Vec<TransactionOutput>, VMStatus>;
```
msmouse, line 16: Do we really need to return VMStatus here?

execution/executor-service/src/thread_executor_service.rs
```diff
@@ -0,0 +1,52 @@
...
+
+use crate::{remote_executor_service, remote_executor_service::RemoteExecutorService};
+use aptos_config::utils;
+use std::{
+    net::{IpAddr, Ipv4Addr, SocketAddr},
+    thread::{self, JoinHandle},
+};
+
+/// This is a simple implementation of RemoteExecutorService that runs the executor service in a
+/// separate thread. This should be used for testing only.
```
msmouse, line 13: > This should be used for testing only.
was wondering why we needed this. Let's put it under the tests? or explicitly mark it testonly

aptos-move/aptos-vm/src/sharded_block_executor/executor_shard.rs
```diff
@@ -2,49 +2,41 @@
...
-use std::sync::{
-    mpsc::{Receiver, Sender},
-    Arc,
-};
+use std::sync::mpsc::{Receiver, Sender};
 
 /// A remote block executor that receives transactions from a channel and executes them in parallel.
 /// Currently it runs in the local machine and it will be further extended to run in a remote machine.
-pub struct ExecutorShard<S: StateView + Sync + Send + 'static> {
+pub struct ExecutorShard<S: StateView + Sync + Send + 'static, E: TBlockExecutorClient> {
```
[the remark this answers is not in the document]
sitalkedia: May be I am missing something but how would we do that in this case? The struct stores the `executor_client`, which is constraint to E.

storage/state-view/src/lib.rs
sitalkedia: I can't think of any cleaner way to do this. Any suggestion? :)

third_party/move/move-core/types/src/vm_status.rs
sitalkedia: You are right, these doesn't belong here. Let me clean them up.

aptos-move/aptos-vm/src/sharded_block_executor/block_executor_client.rs
sitalkedia: Yes, otherwise, we would have to change the VM apis and that is going to be messy. This ensures that the client doesn't need to distinguish between local execution vs remote. What is the concern here?

execution/executor-service/src/remote_executor_client.rs
sitalkedia: Added retries and panic instead.

aptos-move/aptos-vm/src/sharded_block_executor/executor_shard.rs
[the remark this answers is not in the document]
msmouse: try to just do `struct ExecutorShard<S, E>` and only specify the constraints on the `impl`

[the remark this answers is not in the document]
sitalkedia: I see.. didn't know this sorcery 😄

sitalkedia:
@msmouse - I messed up the current branch and not able to fix it. Opened a new PR - https://github.com/aptos-labs/aptos-core/pull/8572. Closing this one - sorry about the confusion.

storage/state-view/src/lib.rs
msmouse: Maybe isolate it to a `triat AsInMemoryStateview`? only the client will require this trait.

aptos-move/aptos-vm/src/sharded_block_executor/block_executor_client.rs
msmouse: I'm confused.. if you mean you wanna use the client in place of `AptosVM`, why not just implement `VMExecutor` for the client? otherwise the caller side wouldn't require you to return VMStatus anyway.. Guess I'll need to read your next PR.

sitalkedia: @msmouse - Even if we implement `VMExecutor` for the client, the interface still requires it to return `VMStatus`? Also, IMO its not clean for the client to implement `VMExecutor` as the executor exposes both sharded and unsharded execution APIs, which this client can't expose.

sitalkedia:
@msmouse - Apparently some issue with github that prevents push to existing PRs.

sitalkedia:
Let's continue the discussion in this PR instead. Hopefully, that issue will be resolved soon.

aptos-move/aptos-vm/src/sharded_block_executor/block_executor_client.rs
```diff
@@ -0,0 +1,62 @@
...
+}
+
+impl BlockExecutorClient for LocalExecutorClient {
```
zjma, line 19: Maybe only me, but I'm a bit confused by the naming: here the `Client` in `BlockExecutorClient` gives me the impression that anyone implementing it can talk to some service (in this case `ExecutorService`). Then `ExecutorService` holds a `LocalExecutorClient` which implements BlockExecutorClient...

IMO `BlockExecutor` is a better name.

sitalkedia: The issue is that `BlockExecutor` is already being used for the upper execution layer.  Naming is hard :(

execution/executor-service/src/remote_executor_client.rs
```diff
@@ -0,0 +1,75 @@
...
+    fn execute_block<S: StateView + Sync>(
+        &self,
+        transactions: Vec<Transaction>,
+        state_view: &S,
+        concurrency_level: usize,
+        maybe_block_gas_limit: Option<u64>,
+    ) -> Result<Vec<TransactionOutput>, VMStatus> {
+        let input = BlockExecutionRequest::ExecuteBlock(ExecuteBlockCommand {
+            transactions,
+            state_view: S::as_in_memory_state_view(state_view),
```
msmouse, line 69: I'm afraid even with 100k accounts serializing this would kill the performance completely..

aptos-move/aptos-transaction-benchmarks/src/transactions.rs
```diff
@@ -218,8 +229,20 @@ where
 
         let state_view = Arc::new(executor.get_state_view().clone());
         let parallel_block_executor =
```
msmouse, line 231: I mean, were you arguing this thing returns `VMStatus` so the client must return `VMStatus`? I'd argue this thing is at the same level of the `executor::BlockExecutor` which doesn't return `VMStatus`. I think it's awkward to convert things like your `executor_service::Error` into VMStatus when it's not necessary (okay I see you unwraped on it, but it's not really necessary, right?).

execution/executor-service/src/remote_executor_client.rs
sitalkedia: Clarified offline, we will do partitioning so that we are not serializing the same state key/value for all shards.

aptos-move/aptos-transaction-benchmarks/src/transactions.rs
sitalkedia: Discussed offline, this is one level lower than the `BlockExecutor`. For now, this is okay - we can revisit this infra further once we evolve the design.

storage/state-view/src/lib.rs
sitalkedia: Tried this out and this doesn't work unfortunately.

```
error[E0119]: conflicting implementations of trait `in_memory_state_view::AsInMemoryStateView` for type `in_memory_state_view::InMemoryStateView`
  --> storage/state-view/src/in_memory_state_view.rs:51:1
   |
17 | impl<T: StateView> AsInMemoryStateView for T {
   | -------------------------------------------- first implementation here
...
51 | impl AsInMemoryStateView for InMemoryStateView {
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ conflicting implementation for `in_memory_state_view::InMemoryStateView`
   ```

msmouse (approved):
Thanks boss!

zjma (approved):

PR #8549, opened 2023-06-06
[gas] fix gas metering for resource groups

vgao1996:
This implements proper (hopefully) gas metering for resource groups so that they charged in spite of the resource existing or not.

### DIFF
--- a/aptos-move/aptos-gas/src/gas_meter.rs
+++ b/aptos-move/aptos-gas/src/gas_meter.rs
@@ -491,11 +491,12 @@ impl MoveGasMeter for StandardGasMeter {
         &mut self,
         _addr: AccountAddress,
         _ty: impl TypeView,
-        loaded: Option<(NumBytes, impl ValueView)>,
+        val: Option<impl ValueView>,
+        bytes_loaded: NumBytes,
     ) -> PartialVMResult<()> {
         if self.feature_version != 0 {
             // TODO(Gas): Rewrite this in a better way.
-            if let Some((_, val)) = &loaded {
+            if let Some(val) = &val {
                 self.use_heap_memory(
                     self.gas_params
                         .misc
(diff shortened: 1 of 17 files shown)


gerben-stavenga (approved):

third_party/move/move-core/types/src/resolver.rs
```diff
@@ -41,14 +45,7 @@ pub trait ResourceResolver {
...
         metadata: &[Metadata],
-    ) -> Result<Option<(Vec<u8>, u64)>, Error>;
-}
-
-pub fn resource_add_cost(buf: Option<Vec<u8>>, extra: u64) -> Option<(Vec<u8>, u64)> {
-    buf.map(|b| {
-        let len = b.len() as u64 + extra;
-        (b, len)
-    })
+    ) -> Result<(Option<Vec<u8>>, usize), Error>;
```
runtian-zhou, line 48: I would try to avoid changes in those apis. Given that we expect the Move repo to be used by other entities, changes like this might be breaking for them.

runtian-zhou (approved):

aptos-move/aptos-gas/src/gas_meter.rs
```diff
@@ -504,10 +505,13 @@ impl MoveGasMeter for StandardGasMeter {
                 )?;
             }
         }
+        if self.feature_version <= 8 && val.is_none() && bytes_loaded != 0.into() {
+            return Err(PartialVMError::new(StatusCode::UNKNOWN_INVARIANT_VIOLATION_ERROR).with_message("in legacy versions, number of bytes loaded must be zero when the resource does not exist ".to_string()));
```
movekevin, line 509: Should we just ignore these values instead of returning an error here? This would be a safer way to maintain backward incompatibility without just erroring out.

movekevin (approved):

third_party/move/move-core/types/src/resolver.rs
vgao1996: Good point. Although I'm not aware of a good way to support the new semantics (bytes loaded could differ from the data read) and keep the old interface at the same time. 
I feel like what we can do in the future, is to evolve this into a clearer & more generic interface and try to upstream it.

vgao1996: Anyway, let me think a little bit more about this and see if I can come up with something.

aptos-move/aptos-gas/src/gas_meter.rs
vgao1996: I would actually argue the opposite: this check helps us ensure that changes I made elsewhere are indeed compatible with the old implementation, or otherwise this could fire during replay.

PR #8550, opened 2023-06-07
[CLI][E2E] Add test for aptos move publish

banool:
### Description
This PR adds a test to check that `aptos move publish` works.

In addition to the test itself I make a couple of framework and CI changes to make this actually work.

### Test Plan
Tested locally:
```
DOCKER_DEFAULT_PLATFORM=linux/amd64 poetry run python main.py -d --base-network mainnet --test-cli-tag nightly
```

I tested in CI with `pull_request` instead of `pull_request_target` and it worked too.

### DIFF
--- a/crates/aptos/e2e/cases/move.py
+++ b/crates/aptos/e2e/cases/move.py
@@ -1,0 +1,60 @@
+# Copyright © Aptos Foundation
+# SPDX-License-Identifier: Apache-2.0
+
+import json
+
+from common import TestError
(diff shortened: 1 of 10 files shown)


banool:
I accidentally merged this into another PR due to the magic of automerge. See the new PR here: https://github.com/aptos-labs/aptos-core/pull/8590.

movekevin (commented):
Approved after talking on Slack

PR #8551, opened 2023-06-07
[dag] reliable broadcast

zekun000:
This commit introduces an abstracted reliable broadcast concept, it's responsible to keep trying sending message until either it aggregates enough ack or it's cancelled.

### DIFF
--- a/consensus/src/dag/reliable_broadcast.rs
+++ b/consensus/src/dag/reliable_broadcast.rs
@@ -1,0 +1,98 @@
+// Copyright © Aptos Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::network_interface::ConsensusMsg;
+use aptos_consensus_types::common::Author;
+use async_trait::async_trait;
+use futures::{stream::FuturesUnordered, StreamExt};
+use std::{future::Future, sync::Arc, time::Duration};
+use tokio::sync::oneshot;
+
+pub trait DAGMessage: Sized + Clone {
+    fn from_network_message(msg: ConsensusMsg) -> anyhow::Result<Self>;
+
+    fn into_network_message(self) -> ConsensusMsg;
+}
+
+pub trait BroadcastStatus {
+    type Message: DAGMessage;
+    type Ack: DAGMessage;
+    type Aggregated;
+
+    fn empty(validators: Vec<Author>) -> Self;
+
+    fn add(&mut self, peer: Author, ack: Self::Ack) -> anyhow::Result<Option<Self::Aggregated>>;
+}
+
+#[async_trait]
+pub trait DAGNetworkSender: Send + Sync {
+    async fn send_rpc(
+        &self,
+        receiver: Author,
+        message: ConsensusMsg,
+        timeout: Duration,
+    ) -> anyhow::Result<ConsensusMsg>;
+}
+
+pub struct ReliableBroadcast {
+    validators: Vec<Author>,
+    network_sender: Arc<dyn DAGNetworkSender>,
+}
+
+impl ReliableBroadcast {
+    pub fn new(validators: Vec<Author>, network_sender: Arc<dyn DAGNetworkSender>) -> Self {
+        Self {
+            validators,
+            network_sender,
+        }
+    }
+
+    pub fn broadcast<S: BroadcastStatus>(
+        &self,
+        message: S::Message,
+        return_tx: oneshot::Sender<S::Aggregated>,
+        mut cancel_rx: oneshot::Receiver<()>,
+    ) -> impl Future<Output = ()> {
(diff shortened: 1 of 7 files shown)


consensus/src/dag/reliable_broadcast.rs
```diff
@@ -0,0 +1,101 @@
...
+                        match result {
+                            Ok(msg) =>  {
+                                if let Ok(ack) = S::Ack::from_network_message(msg) {
+                                    if let Ok(Some(aggregated)) = aggregating.add(receiver, ack) {
+                                        let _ = return_tx.send(aggregated);
+                                        return;
+                                    }
+                                }
+                            },
+                            Err(_) => fut.push(send_message(receiver, network_message.clone())),
```
sitalkedia, line 91: When will this happen? Do we need some limit to this retry?

consensus/src/dag/reliable_broadcast.rs
```diff
@@ -0,0 +1,101 @@
...
+            network_sender,
+        }
+    }
+
+    pub fn broadcast<S: BroadcastStatus>(
+        &self,
+        message: S::Message,
+        return_tx: oneshot::Sender<S::Aggregated>,
+        mut cancel_rx: oneshot::Receiver<()>,
+    ) -> impl Future<Output = ()> {
```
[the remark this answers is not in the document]
zekun000: we're spawning the Future so it can't take &self into it, we clone things from self and create a 'static lifetime future

consensus/src/dag/reliable_broadcast.rs
zekun000: timeout, peer not connect etc, the point of reliable broadcast is to keep trying until receiving 2f+1 responses or explicitly cancelled

consensus/src/dag/tests/reliable_broadcast_tests.rs
```diff
@@ -0,0 +1,151 @@
...
+    tokio::spawn(rb.broadcast::<TestBroadcastStatus>(message.clone(), tx, cancel_rx));
+    assert!(rx.await.is_err());
+
+    // implicit drop cancel
+    let (tx, rx) = oneshot::channel();
+    let (cancel_tx, cancel_rx) = oneshot::channel();
+    drop(cancel_tx);
+    tokio::spawn(rb.broadcast::<TestBroadcastStatus>(message, tx, cancel_rx));
+    assert!(rx.await.is_err());
+}
```
ibalajiarun, line 151: dropping the future should also cancel right?

consensus/src/dag/tests/reliable_broadcast_tests.rs
```diff
@@ -0,0 +1,151 @@
...
+    type Aggregated = HashSet<Author>;
+
+    fn empty(receivers: Vec<Author>) -> Self {
+        Self {
+            threshold: receivers.len(),
+            received: HashSet::new(),
+        }
+    }
+
+    fn add(&mut self, peer: Author, _ack: Self::Ack) -> anyhow::Result<Option<Self::Aggregated>> {
```
ibalajiarun, line 65: Do we expect to return an Err() result here? If not, this can be just an Option<>

ibalajiarun (approved):
nit: I suggest updating the description to say its the sender side of the reliable broadcast.

consensus/src/dag/tests/reliable_broadcast_tests.rs
zekun000: I think so, like wrong signature for example

consensus/src/dag/tests/reliable_broadcast_tests.rs
zekun000: drop this receiver won't cancel it

consensus/src/dag/reliable_broadcast.rs
```diff
@@ -0,0 +1,101 @@
...
+                            .send_rpc(receiver, message, Duration::from_millis(500))
+                            .await,
+                    )
+                }
+            };
+            for receiver in receivers {
+                fut.push(send_message(receiver, network_message.clone()));
+            }
+            loop {
+                tokio::select! {
```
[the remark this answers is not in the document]
zekun000: generally I found tokio is more modern than futures crate

sasha8 (commented):
A few points:

1. Zekun, as you mentioned offline, we still need to implement RPC broadcast, otherwise we serialize too many times, right?

2. We have this optimization that we only care about the latest node and reset we can drop. So this we will implement on top of this via the cancel channel?

3. We reliably broadcast two things. First, the node for which we need to aggregate a signature. Second, the certificate for we which we retry until received acks from all or need to broadcast a new node. Do your fancy traits support this?

PR #8552, opened 2023-06-07
[TF] Use a ManagedCertificate for testnet-addons ingress

sionescu:
Also, rename the Helm values to be similar to the ones in pfn-addons.

### DIFF
--- a/terraform/helm/testnet-addons/values.yaml
+++ b/terraform/helm/testnet-addons/values.yaml
@@ -83,8 +83,10 @@ service:
 ingress:
   # -- The ACM certificate to install on the ingress
   acm_certificate:
-  # -- The GCP certificate to install on the ingress
-  gcp_certificate:
+  # -- The GCE static IP to install on the ingress
+  gce_static_ip:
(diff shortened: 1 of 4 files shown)


terraform/helm/testnet-addons/values.yaml
```diff
@@ -83,8 +83,10 @@ service:
 ingress:
   # -- The ACM certificate to install on the ingress
   acm_certificate:
-  # -- The GCP certificate to install on the ingress
-  gcp_certificate:
+  # -- The GCE static IP to install on the ingress
+  gce_static_ip:
+  # -- The GCE certificate to install on the ingress
+  gce_managed_certificate:
```
rustielin, line 89: can you re-generate the README? `helm-docs .`

rustielin (approved):
LGTM

terraform/helm/testnet-addons/values.yaml
sionescu: Done.

PR #8558, opened 2023-06-07
[Spec] update specs of transaction_fee

UIZorrot:
### Description

Completed all specs left as TODO in transaction_fee.move

### Test Plan

aptos move prove -f transaction_fee.move


aptos-move/framework/aptos-framework/sources/transaction_fee.spec.move
```diff
@@ -133,4 +161,4 @@ spec aptos_framework::transaction_fee {
         aborts_if exists<AptosCoinCapabilities>(addr);
         ensures exists<AptosCoinCapabilities>(addr);
     }
-}
+}
```
junkil-park, line 164: Add a new line after this line. Otherwise, the linter fails: https://github.com/aptos-labs/aptos-core/actions/runs/5206823455/jobs/9393770795?pr=8558#step:5:80

junkil-park (approved):

PR #8564, opened 2023-06-07
[TS SDK] bump to version 1.10.0

0xmaayan:


0xmaayan:
> Are you waiting for fungible assets?

fungible assets is in

PR #8566, opened 2023-06-07
[Sharding] Integrate executor benchmark with sharded block-stm

zjma:
### Description

New incarnation of #8188 .

### Test Plan

Single-node performance tests. If should have no impact on the performance testing results.


execution/executor-benchmark/src/transaction_generator.rs
```diff
@@ -401,21 +412,33 @@ impl TransactionGenerator {
         block_size: usize,
         num_blocks: usize,
         transactions_per_sender: usize,
+        non_conflicting_txns: bool,
     ) {
         for _ in 0..num_blocks {
             // TODO: handle when block_size isn't divisible by transactions_per_sender
+            let mut random_account_generator: Box<dyn RandomAccountGenerator> =
```
zjma, line 419: @sitalkedia should your changes to `run_transfer` go away?

sitalkedia: Yes, no need for any change in `run_transfer` function anymore since we are supporting that in `run_workload` function.

execution/executor-benchmark/src/account_generator.rs
```diff
@@ -59,10 +63,23 @@ impl AccountGenerator {
...
+        &mut self,
+        batch_size: usize,
+    ) -> (Arc<RwLock<LocalAccount>>, Vec<AccountAddress>);
+}
+
 pub struct AccountCache {
     generator: AccountGenerator,
-    pub accounts: VecDeque<LocalAccount>,
-    rng: StdRng,
+    pub accounts: VecDeque<Arc<RwLock<LocalAccount>>>,
```
sitalkedia, line 81: Can you clarify why `RwLock` is needed?

execution/executor/src/components/chunk_output.rs
```diff
@@ -76,7 +76,7 @@ impl ChunkOutput {
             maybe_block_gas_limit,
         )?;
 
-        // TODO(skedia) add logic to emit counters per shard instead of doing it globally.
+        update_counters_for_processed_chunk(&transactions, &transaction_outputs, "executed");
```
sitalkedia, line 79: Why did this change? We don't have a good way to support these counters in sharded execution for now.

execution/executor-benchmark/src/account_generator.rs
zjma: Not needed and fixed.

crates/transaction-generator-lib/src/p2p_transaction_generator.rs
```diff
@@ -13,29 +13,166 @@ use rand::{
...
+            for replace_batch_start in (0..num_addresses).step_by(self.sub_pool_size) {
+                let end = min(replace_batch_start + self.sub_pool_size, num_addresses);
+                self.to_be_replaced[replace_batch_start..end].shuffle(rng);
+            }
+            for _ in 0..num_addresses {
+                pool.push(self.to_be_replaced.pop().unwrap());
+            }
+        }
+        let sample = pool.pop().unwrap();
+        self.to_be_replaced.push(sample.clone());
```
igor-aptos, line 107: having this be a template is nice, but .clone() will not work for the other type (LocalAccount). so if we end up needing it for that, we would need to return &T

But since I'm not sure we will need that, this is fine :)

igor-aptos (approved):

PR #8567, opened 2023-06-07
[docs] Fix the docs to use new args format

CapCap:
### Description
Fix the CLI examples in the docs to use new args format

### Test Plan
doc change

### DIFF
--- a/developer-docs-site/docs/guides/nfts/mint-onchain-data.md
+++ b/developer-docs-site/docs/guides/nfts/mint-onchain-data.md
@@ -16,7 +16,7 @@ called *silver coin* by using the [Aptos command line interface (CLI)](../../too
 ## Create a collection
 Token must belong to a collection. First, you can create a collection by running Aptos CLI:
 ```bash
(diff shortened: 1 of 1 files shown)


crates/aptos/src/common/types.rs
```diff
@@ -1704,7 +1704,9 @@ pub struct MultisigAccountWithSequenceNumber {
 pub struct TypeArgVec {
     /// TypeTag arguments separated by spaces.
     ///
-    /// Example: `u8 u16 u32 u64 u128 u256 bool address vector signer`
+    /// Supported types: `bool u8 u16 u32 u64 u128 u256 address string`
+    ///
+    /// Example: `--args u8:25 'address:["0x1", "0x2]' bool:[true, true, false] bool:true string:"Aptos"`
```
gregnazario, line 1709: This is actually unfortunately the wrong field.

`type-args` is for generics

`args` is for the thing you've got up here

gregnazario (approved):

saharct (approved):

PR #8576, opened 2023-06-07
[Helm] Add charts for kube-state-metrics and prometheus-node-exporter

sionescu:
### Description

We don't want to tie their installation to the "monitoring" helm chart which installed Prometheus & Grafana.
Their installation will be done using internal-ops/infra/observability/deploy-agents.mts

### Test Plan

Installing first in the GCP Forge clusters.


perryjrandall (approved):

rustielin (changes requested):
It seems like all these charts are just including a single subchart dependency. Can we just install directly the chart inside without nesting?

sionescu:
@rustielin We could do that, but we're already using this pattern of wrapping an external chart and providing default values and perhaps extra resources. It's easier to use this way because you just point Helm to a local directory and it will install that.

terraform/helm/kube-state-metrics/values.yaml
```diff
@@ -0,0 +1,5 @@
+kube-state-metrics:
+  namespaceOverride: kube-system
```
geekflyer, line 2: let's stop installing stuff into kube-system imo. We should just install these things into their own namespaces.

sionescu: Ok, I'll move the to another namespace, like "aptos-monitoring".

sionescu: Moved them to namespace "monitoring".

rustielin (approved):

PR #8579, opened 2023-06-07
clean AptosVmImpl::new() up a bit

msmouse:
### Description

Mainly to make it clear:
1. do not load `storage_gas_params` if there's no global `gas_params`
2. only override certain table gas parameters with io gas parameters when pricing v2 is active (for now it's equivalent to that there is the storage gas schedule stored on chain; gonna change after we deprecate the curves)

and make the above logic all exposed in `new()` (instead of partially under `StorageGasParams::new()`

### Test Plan
started full replay run here
https://github.com/aptos-labs/aptos-core/actions/runs/5212425745

### DIFF
--- a/aptos-move/aptos-gas/src/transaction/storage.rs
+++ b/aptos-move/aptos-gas/src/transaction/storage.rs
@@ -308,14 +308,9 @@ pub struct StorageGasParameters {
 impl StorageGasParameters {
     pub fn new(
         feature_version: u64,
-        gas_params: Option<&AptosGasParameters>,
+        gas_params: &AptosGasParameters,
         storage_gas_schedule: Option<&StorageGasSchedule>,
-    ) -> Option<Self> {
-        if feature_version == 0 || gas_params.is_none() {
-            return None;
-        }
-        let gas_params = gas_params.unwrap();
-
+    ) -> Self {
         let pricing = match storage_gas_schedule {
             Some(schedule) => {
                 StoragePricing::V2(StoragePricingV2::new(feature_version, schedule, gas_params))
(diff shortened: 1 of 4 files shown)


vgao1996:
Love it! Need to reason about backward compatibility but maybe it's easier to do a replay.

msmouse:
@vgao1996 Yeah I'll try the replay jobs after tests passed.

vgao1996 (approved):

grao1991 (approved):

aptos-move/aptos-gas/src/transaction/storage.rs
```diff
@@ -308,14 +308,9 @@ pub struct StorageGasParameters {
 impl StorageGasParameters {
     pub fn new(
         feature_version: u64,
-        gas_params: Option<&AptosGasParameters>,
+        gas_params: &AptosGasParameters,
         storage_gas_schedule: Option<&StorageGasSchedule>,
-    ) -> Option<Self> {
-        if feature_version == 0 || gas_params.is_none() {
```
zekun000, line 314: what happens to feature_version = 0?

zekun000 (approved):

aptos-move/aptos-gas/src/transaction/storage.rs
msmouse: `feature_version == 0 && param_param.is_none()` is only when running the genesis, where a `UnmeteredGasMeter` is used (in fact if we don't construct the `StorageGasParameters` here, a standard gas meter will fail to construct)

PR #8584, opened 2023-06-08
Add zip functions in vector module to iterate over 2 vectors concurrently

movekevin:
### Description
This is a common operations for iterating over vector inputs.

### Test Plan
New unit tests

### DIFF
--- a/aptos-move/framework/move-stdlib/sources/vector.move
+++ b/aptos-move/framework/move-stdlib/sources/vector.move
@@ -15,6 +15,9 @@ module std::vector {
     /// The index into the vector is out of bounds
     const EINVALID_RANGE: u64 = 0x20001;
 
+    /// The length of the vectors are not equal.
+    const EVECTORS_LENGTH_MISMATCH: u64 = 0x20002;
+
     #[bytecode_instruction]
     /// Create an empty vector.
     native public fun empty<Element>(): vector<Element>;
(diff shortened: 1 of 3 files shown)


aptos-move/framework/move-stdlib/sources/vector.move
```diff
@@ -263,6 +266,51 @@ module std::vector {
         }
     }
 
+    /// Apply the function to each pair of elements in the two given vectors, consuming them.
+    public inline fun zip<Element1, Element2>(v1: vector<Element1>, v2: vector<Element2>, f: |Element1, Element2|) {
```
0xbe1, line 270: can you also add a version that instead returns `zip<Element1, Element2, A>(v1: vector<Element1>, v2: vector<Element2>, f: |Element1, Element2| A)`

movekevin: Added zip_map and zip_map_ref

xindingw (approved):

aptos-move/framework/move-stdlib/sources/vector.move
```diff
@@ -263,6 +266,51 @@ module std::vector {
...
+
+    /// Apply the function to each pair of elements in the two given vectors in the reverse order, consuming them.
+    /// This errors out if the vectors are not of the same length.
+    public inline fun zip_reverse<Element1, Element2>(
+        v1: vector<Element1>,
+        v2: vector<Element2>,
+        f: |Element1, Element2|,
+    ) {
+        let len = length(&v1);
+        // We can't use the constant EVECTORS_LENGTH_MISMATCH here as all calling code would then need to define it
```
lightmark, line 285: when could this be solved... This is ugly. Can we make this not inline and later convert it to inline if we solved the scope issue of inline function?

lightmark (approved):

aptos-move/framework/move-stdlib/sources/vector.move
movekevin: There's a Slack thread on this where we're discussing :)

brmataptos: We need a way to refer to this module's const in the place where it is inlined.  But I suppose that's a bytecode issue.

PR #8586, opened 2023-06-08
[gha] set up merge gatekeeper

rustielin:
### Description

Try out merge gatekeeper (https://github.com/upsidr/merge-gatekeeper)

This will allow us to get close to target determination in our monorepo, if we use it in conjunction with modified GHA triggers that only run certain checks when certain file paths are changed.

### Test Plan

Just check out the workflow output for now. We can run the merge gatekeeper in parallel with the rest of our CI without side effects, so long as we don't change the branch protection to require it.

### DIFF
--- a/.github/workflows/merge-gatekeeper.yaml
+++ b/.github/workflows/merge-gatekeeper.yaml
@@ -1,0 +1,34 @@
+name: "*Merge Gatekeeper"
+
+on:
+  pull_request:
+    types: [labeled, opened, synchronize, reopened, auto_merge_enabled]
+
+env:
+  MERGE_GATEKEEPER_ALLOWLIST: "rustielin,perryjrandall,geekflyer,sherry-x,sionescu,ibalajiarun,igor-aptos,sitalkedia"
+
+jobs:
+  merge-gatekeeper:
+    runs-on: ubuntu-latest
+    # Restrict permissions of the GITHUB_TOKEN.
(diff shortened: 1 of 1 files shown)


sionescu (approved):

.github/workflows/merge-gatekeeper.yaml
```diff
@@ -0,0 +1,19 @@
+name: "*Merge Gatekeeper"
+
+on:
+  pull_request:
+
+jobs:
+  merge-gatekeeper:
```
perryjrandall, line 7: Can we gate this to user / branch for some time first?

For instance gate to our prod eng group or a set of usernames

perryjrandall (approved):

.github/workflows/merge-gatekeeper.yaml
rustielin: no easy way to do this, unless we block on a specific set of branch prefixes. I think it's worth getting the signal everywhere for now

perryjrandall: Cant we do something silly to start? at least for a couple days, opt in some blockchain people and our team

```
          if [[ "${{ github.actor }}" == "your-username" ]]; then
```

rustielin: ok lfg, always down for silly

PR #8590, opened 2023-06-08
[CLI][E2E] Add test for aptos move publish

banool:
### Description
This PR adds a test to check that `aptos move publish` works.

### Test Plan
Tested locally:
```
DOCKER_DEFAULT_PLATFORM=linux/amd64 poetry run python main.py -d --base-network mainnet --test-cli-tag nightly
```

I also tested locally with `--test-cli-path`.

### DIFF
--- a/aptos-move/move-examples/cli-e2e-tests/common/sources/cli_e2e_tests.move
+++ b/aptos-move/move-examples/cli-e2e-tests/common/sources/cli_e2e_tests.move
@@ -1,0 +1,327 @@
+module addr::cli_e2e_tests {
+    use std::error;
+    use std::option::{Self, Option};
+    use std::signer;
+    use std::string::{Self, String};
+
+    use aptos_framework::object::{Self, ConstructorRef, Object};
+
+    use aptos_token_objects::collection;
+    use aptos_token_objects::token;
+    use aptos_std::string_utils;
+
+    const ENOT_A_HERO: u64 = 1;
+    const ENOT_A_WEAPON: u64 = 2;
+    const ENOT_A_GEM: u64 = 3;
+    const ENOT_CREATOR: u64 = 4;
+    const EINVALID_WEAPON_UNEQUIP: u64 = 5;
+    const EINVALID_GEM_UNEQUIP: u64 = 6;
+    const EINVALID_TYPE: u64 = 7;
+
+    struct OnChainConfig has key {
+        collection: String,
+    }
+
+    #[resource_group_member(group = aptos_framework::object::ObjectGroup)]
+    struct Hero has key {
+        armor: Option<Object<Armor>>,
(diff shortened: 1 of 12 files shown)


aptos-move/move-examples/cli-e2e-tests/common/sources/cli_e2e_tests.move
```diff
@@ -0,0 +1,327 @@
+module addr::cli_e2e_tests {
```
0xmaayan, line 1: I am a bit confused as for why we need this move module? python has a compiled module we use for testing https://github.com/aptos-labs/aptos-core/blob/main/ecosystem/python/sdk/examples/two_by_two_transfer.mv can't we use this one to test the `aptos move` subcommands?

banool: Well the point of these tests is to be end to end, so we do the same steps the users would do. No one really precompiles modules, takes the mv file, and then publishes that.

banool:
Thanks to the previous changes to the CI to make it check out the correct code, the tests actually ran in this CI. As such we can see that the new test indeed works: https://github.com/aptos-labs/aptos-core/actions/runs/5223635795.

aptos-move/move-examples/cli-e2e-tests/common/sources/cli_e2e_tests.move
gregnazario: We should likely then, put one place where this can be used across multiple test suites, including somehow autocompiling it for the Python SDK.

gregnazario (commented):
Looks good, but maybe we can simply compile an already existing code and test?  Like the hero move example you're copying here?

wijesinghe1967:
0x546ccaAA604B1C2E30313577071C9D4BBf0E64e2

wijesinghe1967:
0x546ccaAA604B1C2E30313577071C9D4BBf0E64e2

wijesinghe1967:
0x546ccaAA604B1C2E30313577071C9D4BBf0E64e2

0xjinn:
Should we just use the `hello_blockchain` example for testing instead of creating a new one?

banool:
Same problem I mentioned before, that example uses a local path for the framework dep but we need to use a path from git.

0xjinn:
> Same problem I mentioned before, that example uses a local path for the framework dep but we need to use a path from git.

@banool Ah I see! Just saw your comment above - https://github.com/aptos-labs/aptos-core/pull/8590#issuecomment-1584816048

LGTM

0xjinn (approved):

gregnazario (approved):
Please make sure to see if you can combine the test with the already existing similar test in Move later

PR #8594, opened 2023-06-08
[FA] add utility functions in primary_fungible_store

lightmark:
### Description
add a couple of mint, transfer, freeze/unfreeze function to primary_store.

### Test Plan
ut


movekevin (approved):

aptos-move/framework/aptos-framework/sources/primary_fungible_store.move
```diff
@@ -148,8 +148,73 @@ module aptos_framework::primary_fungible_store {
         fungible_asset::transfer(sender, sender_store, recipient_store, amount);
     }
 
+    /// Mint to the primary store of `owner`.
+    public fun mint(
```
movekevin, line 152: nit: I think the full function signature can fit on the same line without exceeding the char limit. Same for other functions below

xindingw (approved):

PR #8595, opened 2023-06-08
Forge test for testing throughput on network purely tuned for throughput

igor-aptos:

### DIFF
--- a/config/src/config/consensus_config.rs
+++ b/config/src/config/consensus_config.rs
@@ -102,8 +102,8 @@ impl Default for ConsensusConfig {
             max_sending_block_bytes: 600 * 1024, // 600 KB
             max_sending_block_bytes_quorum_store_override: 5 * 1024 * 1024, // 5MB
             max_receiving_block_txns: 10000,
-            max_receiving_block_txns_quorum_store_override: 2
-                * MAX_SENDING_BLOCK_TXNS_QUORUM_STORE_OVERRIDE,
+            max_receiving_block_txns_quorum_store_override: 10000
+                .max(2 * MAX_SENDING_BLOCK_TXNS_QUORUM_STORE_OVERRIDE),
             max_receiving_block_bytes: 3 * 1024 * 1024, // 3MB
             max_receiving_block_bytes_quorum_store_override: 6 * 1024 * 1024, // 6MB
             max_pruned_blocks_in_mem: 100,
--- a/testsuite/forge-cli/src/main.rs
+++ b/testsuite/forge-cli/src/main.rs
@@ -3,7 +3,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use anyhow::{format_err, Context, Result};
-use aptos_config::config::ConsensusConfig;
+use aptos_config::config::{ChainHealthBackoffValues, ConsensusConfig, PipelineBackpressureValues};
 use aptos_forge::{
     args::TransactionTypeArg,
     success_criteria::{LatencyType, StateProgressThreshold, SuccessCriteria},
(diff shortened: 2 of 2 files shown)


config/src/config/consensus_config.rs
```diff
@@ -102,8 +102,8 @@ impl Default for ConsensusConfig {
             max_sending_block_bytes: 600 * 1024, // 600 KB
             max_sending_block_bytes_quorum_store_override: 5 * 1024 * 1024, // 5MB
             max_receiving_block_txns: 10000,
-            max_receiving_block_txns_quorum_store_override: 2
-                * MAX_SENDING_BLOCK_TXNS_QUORUM_STORE_OVERRIDE,
+            max_receiving_block_txns_quorum_store_override: 10000.max(
```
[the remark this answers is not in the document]
bchocho: Yikes. Will need to make these a part of the sanitizer

igor-aptos:
this is ready for review/landing

testsuite/forge-cli/src/main.rs
```diff
@@ -1420,6 +1421,65 @@ fn realistic_env_max_throughput_test_suite(duration: Duration) -> ForgeConfig {
         )
 }
 
+fn realistic_network_tuned_for_throughput_test() -> ForgeConfig {
+    ForgeConfig::default()
+        .with_initial_validator_count(NonZeroUsize::new(12).unwrap())
+        // if we have full nodes for subset of validators, TPS drops.
+        // Validators without VFN are proposing almost empty blocks,
```
bchocho, line 1428: Seems unexpected. I would think QS batch broadcast has an equal opportunity to make it to a VN-only node vs. VN+VFN node proposal. Is there a forge run you can point me to here?

testsuite/forge-cli/src/main.rs
```diff
@@ -1420,6 +1421,65 @@ fn realistic_env_max_throughput_test_suite(duration: Duration) -> ForgeConfig {
...
+                serde_yaml::to_value(Vec::<ChainHealthBackoffValues>::new()).unwrap();
+
+            helm_values["validator"]["config"]["consensus"]
+                ["wait_for_full_blocks_above_recent_fill_threshold"] = (0.8).into();
+            helm_values["validator"]["config"]["consensus"]
+                ["wait_for_full_blocks_above_pending_blocks"] = 8.into();
+
+            helm_values["validator"]["config"]["consensus"]["quorum_store"]["back_pressure"]
+                ["backlog_txn_limit_count"] = 100000.into();
+            helm_values["validator"]["config"]["consensus"]["quorum_store"]["back_pressure"]
```
bchocho, line 1454: Did you have a chance to play with this number? I actually think a higher value might not be better for throughput. It likely just affects latency on low load.

bchocho (approved):
I have some questions below on the QS side, but probably easier for me to play around with those values once this is landed.

perryjrandall (approved):

testsuite/forge-cli/src/main.rs
```diff
@@ -1420,6 +1421,65 @@ fn realistic_env_max_throughput_test_suite(duration: Duration) -> ForgeConfig {
...
+                ["dynamic_max_txn_per_s"] = 6000.into();
+
+            // Experimental storage optimizations
+            helm_values["validator"]["config"]["storage"]["rocksdb_configs"]["use_state_kv_db"] =
+                true.into();
+            helm_values["validator"]["config"]["storage"]["rocksdb_configs"]
+                ["use_sharded_state_merkle_db"] = true.into();
+        }))
+        .with_success_criteria(
```
perryjrandall, line 1466: Seems like a lot of success criteria can be static?

Not in this diff but it would be great to be able to configure this outside of code

testsuite/forge-cli/src/main.rs
bchocho: Makes sense - maybe we can change the comment to reflect this since quorum store is the default. If its already landing I'll go hunt down the other test too and change them

igor-aptos: I'll update in the stacked PR

PR #8598, opened 2023-06-09
[forge][fix] Track and remove chaos properly

ibalajiarun:
### Description

This PR fixes a bug where some tests remove all chaoses, but it only should remove those that the test adds.

### Test Plan

Multi-region test now succeeds properly.

```
two traffics test: inner traffic : committed: 5967 txn/s, latency: 6540 ms, (p50: 6300 ms, p90: 7800 ms, p99: 12600 ms), latency samples: 2589920
two traffics test : committed: 100 txn/s, latency: 2559 ms, (p50: 2400 ms, p90: 3000 ms, p99: 4500 ms), latency samples: 1820
Max round gap was 1 [limit 4] at version 655404. Max no progress secs was 3.913393 [limit 10] at version 655404.
Test Ok
```

### DIFF
--- a/testsuite/testcases/src/modifiers.rs
+++ b/testsuite/testcases/src/modifiers.rs
@@ -217,6 +217,16 @@ pub struct CpuChaosTest {
     pub override_config: Option<CpuChaosConfig>,
 }
 
+impl CpuChaosTest {
+    fn create_cpu_chaos(&self, swarm: &mut dyn Swarm) -> SwarmCpuStress {
+        let all_validators = swarm.validators().map(|v| v.peer_id()).collect::<Vec<_>>();
+
+        let config = self.override_config.as_ref().cloned().unwrap_or_default();
+
+        create_cpu_stress_template(all_validators, &config)
+    }
+}
+
 impl Test for CpuChaosTest {
     fn name(&self) -> &'static str {
         "CpuChaosWrapper"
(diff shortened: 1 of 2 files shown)


rustielin (approved):
I'm not familiar with the big, but the changes make sense. 

What happens when you try to remove all chaos vs just those that were added? Since we're only running single test in a suite, shouldn't that be the same thing?

testsuite/forge-cli/src/main.rs
```diff
@@ -231,6 +231,12 @@ fn main() -> Result<()> {
     let duration = Duration::from_secs(args.duration_secs as u64);
     let suite_name: &str = args.suite.as_ref();
 
+    let suite_name = if suite_name == "land_blocking" {
```
igor-aptos, line 234: remove

igor-aptos (approved):

igor-aptos:
> What happens when you try to remove all chaos vs just those that were added? Since we're only running single test in a suite, shouldn't that be the same thing?

@rustielin we are using CompositeNetworkTest to merge multiple different types of chaos , so that you can more easily mix and match, i.e:

```
        CompositeNetworkTest::new(
            MultiRegionNetworkEmulationTest {
                override_config: None,
            },
            CpuChaosTest {
                override_config: None,
            },
        )
```

PR #8604, opened 2023-06-09
[Helm] Update autoscaling charts

sionescu:
### Description

Update the version of cluster-autoscaler and metrics-server.

### Test Plan

Deploy to the old AWS devnet, which is no longer in use.

### DIFF
--- a/.github/actions/docker-setup/action.yaml
+++ b/.github/actions/docker-setup/action.yaml
@@ -30,6 +30,12 @@ inputs:
   GIT_CREDENTIALS:
     description: "Optional credentials to pass to git. Useful if you need to pull private repos for dependencies"
     required: false
+  GCP_AUTH_DURATION:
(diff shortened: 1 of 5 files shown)


terraform/helm/autoscaling/Chart.yaml
```diff
@@ -4,5 +4,5 @@ version: 0.1.0
 
 dependencies:
   - name: metrics-server
-    version: 3.8.2
+    version: 3.10.0
```
rustielin, line 7: you'll need to update the cached version of the chart in `charts/` too

rustielin (changes requested):

terraform/helm/autoscaling/Chart.yaml
sionescu: Done.

PR #8605, opened 2023-06-09
Deprecate the storage gas curves 

msmouse:
### Description

two commits:

#### refactor: storage gas parameters: load onchain config from within constructors

Whether or not to load the curve values will be depending on the gas
feature verion, so refactoring to bring the logic consolidated into the
StorageGasParameters constructors.

#### Deprecate the storage gas curves
On gas feature version 10, skip loading the storage gas schedule from
the storage curves. Rather, the gas schedule is updated with the gas
curve initial values loaded in a few (static) entries that don't change with 
the growth of the global state storage.

Gas schedule values are according to https://www.notion.so/aptoslabs/Gas-schedule-change-proposal-after-Storage-Fee-eec5de304fd1493899b2740ce9c76e11?pvs=4#3b7f35cf07f8467c927bd33d0701ac91

### Test Plan
tests and replaying mainnet
https://github.com/aptos-labs/aptos-core/actions/runs/5249272288

prices are slightly down after the change (due to that we don't over charge creations any more):

  | before (units) | after | diff | diff_pct
-- | -- | -- | -- | --
Transfer | 600 | 600 | 0 | 0.00%
CreateAccount | 101600 | 100600 | -1000 | -0.98%
CreateTransfer | 101900 | 100900 | -1000 | -0.98%
CreateStakePool | 207400 | 205500 | -1900 | -0.92%


aptos-move/aptos-gas/src/transaction/mod.rs
```diff
@@ -79,28 +79,32 @@ crate::params::define_gas_parameters!(
             GAS_SCALING_FACTOR
         ],
         // Gas Parameters for reading data from storage.
-        [load_data_base: InternalGas, "load_data.base", 16_000],
         [
-            load_data_per_byte: InternalGasPerByte,
-            "load_data.per_byte",
-            1_000
+            storage_io_per_state_slot_read: InternalGasPerArg,
```
msmouse, line 83: @vgao1996 I assume it's okay to change the unit from `InternalGas` to `InternalGasPerArg`?

vgao1996: Totally okay. It's also known to me that there is currently some friction with converting back and forth between `SomeUnit` and `SomeUnitPerSomeOtherUnit` and I'll see what I can do.

aptos-move/aptos-gas/src/gas_meter.rs
```diff
@@ -61,7 +63,7 @@ use std::collections::BTreeMap;
 //       global operations.
 // - V1
 //   - TBA
-pub const LATEST_GAS_FEATURE_VERSION: u64 = 9;
+pub const LATEST_GAS_FEATURE_VERSION: u64 = 10;
```
vgao1996, line 66: Should this still be 9?

msmouse: 9 is going out with 1.5 I think

grao1991 (approved):

vgao1996 (approved):

PR #8607, opened 2023-06-09
[terraform][gcp][testnet] make default daily 24h

ibalajiarun:
### Description

We want to keep the maintenance window defaults to match GKE defaults because this template could be used by anyone to deploy nodes. We will override them for our nodes internally.

### DIFF
--- a/terraform/aptos-node/gcp/variables.tf
+++ b/terraform/aptos-node/gcp/variables.tf
@@ -252,9 +252,9 @@ variable "gke_maintenance_policy" {
   })
   default = {
     recurring_window = {
-      start_time = "2023-06-01T14:00:00Z"
-      end_time   = "2023-06-01T18:00:00Z"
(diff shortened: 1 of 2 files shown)


terraform/aptos-node/gcp/variables.tf
```diff
@@ -252,9 +252,9 @@ variable "gke_maintenance_policy" {
   })
   default = {
     recurring_window = {
-      start_time = "2023-06-01T14:00:00Z"
-      end_time   = "2023-06-01T18:00:00Z"
-      recurrence = "FREQ=WEEKLY;BYDAY=MO,TU,WE,TH,FR"
+      start_time = "2023-06-15T00:00:00Z"
+      end_time   = "2023-06-15T11:59:00Z"
```
sionescu, line 256: All day should be 23:59.

sionescu (approved):

perryjrandall (approved):

PR #8608, opened 2023-06-09
Partitioning dependent txns support

sitalkedia:

### DIFF
--- a/execution/block-partitioner/src/sharded_block_partitioner/messages.rs
+++ b/execution/block-partitioner/src/sharded_block_partitioner/messages.rs
@@ -9,7 +9,7 @@ use aptos_types::transaction::analyzed_transaction::AnalyzedTransaction;
 use std::sync::Arc;
 
 
-pub struct DiscardTxnsWithCrossShardDep {
+pub struct DiscardCrossShardDep {
     pub transactions: Vec<AnalyzedTransaction>,
     // The frozen dependencies in previous chunks.
     pub prev_rounds_write_set_with_index: Arc<Vec<WriteSetWithTxnIndex>>,
@@ -19,7 +19,7 @@ pub struct DiscardTxnsWithCrossShardDep {
     pub frozen_sub_blocks: Vec<SubBlock>,
 }
 
-impl DiscardTxnsWithCrossShardDep {
+impl DiscardCrossShardDep {
     pub fn new(
         transactions: Vec<AnalyzedTransaction>,
         prev_rounds_write_set_with_index: Arc<Vec<WriteSetWithTxnIndex>>,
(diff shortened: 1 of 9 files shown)


execution/block-partitioner/src/sharded_block_partitioner/messages.rs
```diff
@@ -1,84 +1,84 @@
...
-pub struct DiscardTxnsWithCrossShardDep {
+pub struct DiscardCrossShardDep {
     pub transactions: Vec<AnalyzedTransaction>,
     // The frozen dependencies in previous chunks.
     pub prev_rounds_write_set_with_index: Arc<Vec<WriteSetWithTxnIndex>>,
-    pub prev_rounds_frozen_sub_blocks: Arc<Vec<SubBlock>>,
+    pub current_round_start_index: TxnIndex,
+    // This is the frozen sub block for the current shard and is passed because we want to modify
+    // it to add dependency back edges.
+    pub frozen_sub_blocks: SubBlocksForShard,
```
zjma, line 17: It looks like the final sub-blocks won't be available until the last round. Maybe have every shard hold them and keep updating them every round? Passing them back and forth for every round seems inefficient.

sitalkedia: That's a good observation, I thought about it but didn't do it for 3 reasons 

1. Current design ensures the partitioning workers are stateless, which makes the design cleaner and also, going forward, ensures enhancements like adding fault tolerance/dynamic load balancing etc possible. I would like avoid making it a stateful design by storing the sub-blocks between rounds. 
2. Passing the block back and forth is minimal overhead as we pass ownership and don't do any cloning.
3. The partitioning APIs also is cleaner this way as otherwise, I would have to pass `SubBlocksForShard` only in first message and None in subsequent rounds. 

For above reasons, I would like to stick with current design.

msmouse (approved):

zjma (approved):

execution/block-partitioner/src/types.rs
```diff
@@ -1,48 +1,148 @@
...
-/// 2. `dependents`: The transactions in the set depend on the execution of the transaction. In this
+/// 2. `dependent_edges`: The transactions in the set depend on the execution of the transaction. In this
 /// case, the transactions in the set can only be executed after the transaction has been executed.
+/// Dependent edge is a reverse of required edge, for example if txn 20 in shard 2 requires txn 10 in shard 1,
+/// then txn 10 in shard 1 will have a dependent edge to txn 20 in shard 2.
 pub struct CrossShardDependencies {
-    depends_on: HashSet<TxnIndex>,
-    // TODO (skedia) add support for this.
-    _dependents: HashSet<TxnIndex>,
+    required_edges: CrossShardEdges,
```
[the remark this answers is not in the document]
sitalkedia: I had that naming convention before but that got me very very confused repeatedly, so I changed it. I would like to keep it this way unless strong opinion :)

PR #8610, opened 2023-06-10
[forge] Fix long running job failure

perryjrandall:
Since the migration to GCP long duration forge jobs have been failing.

I looked at this with Rustie and initially we were stumped.

The code that determines the success is based on fetching the phase of the pod that is running the job, if the phase == "succeeded" then pass skip etc or anything else fail.

When looking at the debugging output (which had auth failure) I realized that the status check could possibly have auth failure as well.

This lead me to the auth duration, which is default to 1.5hr previously in aws forge we had a longer expiration by default!

So this PR addresses this by setting the auth expiration to double the runner duration.

I also add a little error handling in case someone hits this again

Test Plan: unittest + run the test on this branch

### DIFF
--- a/.github/workflows/workflow-run-forge.yaml
+++ b/.github/workflows/workflow-run-forge.yaml
@@ -128,6 +128,7 @@ jobs:
           AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
           AWS_DOCKER_ARTIFACT_REPO: ${{ secrets.AWS_DOCKER_ARTIFACT_REPO }}
           GIT_CREDENTIALS: ${{ secrets.GIT_CREDENTIALS }}
+          GCP_AUTH_DURATION: ${{ inputs.FORGE_RUNNER_DURATION_SECS * 2 }}
 
       - name: "Install GCloud SDK"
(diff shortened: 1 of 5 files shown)


rustielin (approved):

igor-aptos (approved):

.github/workflows/workflow-run-forge.yaml
```diff
@@ -128,6 +128,7 @@ jobs:
           AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
           AWS_DOCKER_ARTIFACT_REPO: ${{ secrets.AWS_DOCKER_ARTIFACT_REPO }}
           GIT_CREDENTIALS: ${{ secrets.GIT_CREDENTIALS }}
+          GCP_AUTH_DURATION: ${{ inputs.FORGE_RUNNER_DURATION_SECS * 2 }}
```
igor-aptos, line 131: @perryjrandall I think this broke the run:

 Invalid workflow file: .github/workflows/forge-stable.yaml#L103
The workflow is not valid. In .github/workflows/forge-stable.yaml (Line: 103, Col: 11): Error from called workflow aptos-labs/aptos-core/.github/workflows/workflow-run-forge.yaml@main (Line: 131, Col: 30): Unexpected symbol: '*'. Located at position 35 within expression: inputs.FORGE_RUNNER_DURATION_SECS * 2

PR #8611, opened 2023-06-10
[framework][spec] Refactor the coin module to express total supply as a global invariant

rahxephon89:
### Description

The global invariant of total supply property does not hold during execution of several operations such as burn, extract, extract_all, mint and so on. The check of the invariant needs to be delegated to all callers of these operations. However, the Move prover only allow internal or public friend functions to delegate check of global invariants to their callers because in reality we do not know which functions will call public APIs. 

To specify the total supply property as a global invariant, this PR refactors the coin module by adding public friend helper functions `burn_internal`, `extract_internal`,  `merge_internal`, `extract_all_internal` and `mint_internal` and replacing corresponding public APIs with them in the aptos framework. With this change, we can use `pragma delegate_invariants_to_caller` to prove the global invariant.

### DIFF
--- a/aptos-move/framework/aptos-framework/doc/aptos_coin.md
+++ b/aptos-move/framework/aptos-framework/doc/aptos_coin.md
@@ -304,7 +304,7 @@ accounts.
 
     // Mint the core resource <a href="account.md#0x1_account">account</a> <a href="aptos_coin.md#0x1_aptos_coin_AptosCoin">AptosCoin</a> for gas so it can execute system transactions.
     <a href="coin.md#0x1_coin_register">coin::register</a>&lt;<a href="aptos_coin.md#0x1_aptos_coin_AptosCoin">AptosCoin</a>&gt;(core_resources);
-    <b>let</b> coins = <a href="coin.md#0x1_coin_mint">coin::mint</a>&lt;<a href="aptos_coin.md#0x1_aptos_coin_AptosCoin">AptosCoin</a>&gt;(
+    <b>let</b> coins = <a href="coin.md#0x1_coin_mint_internal">coin::mint_internal</a>&lt;<a href="aptos_coin.md#0x1_aptos_coin_AptosCoin">AptosCoin</a>&gt;(
         18446744073709551615,
         &mint_cap,
     );
@@ -349,7 +349,7 @@ Create new coins and deposit them into dst_addr's account.
     );
 
     <b>let</b> mint_cap = &<b>borrow_global</b>&lt;<a href="aptos_coin.md#0x1_aptos_coin_MintCapStore">MintCapStore</a>&gt;(account_addr).mint_cap;
-    <b>let</b> coins_minted = <a href="coin.md#0x1_coin_mint">coin::mint</a>&lt;<a href="aptos_coin.md#0x1_aptos_coin_AptosCoin">AptosCoin</a>&gt;(amount, mint_cap);
+    <b>let</b> coins_minted = <a href="coin.md#0x1_coin_mint_internal">coin::mint_internal</a>&lt;<a href="aptos_coin.md#0x1_aptos_coin_AptosCoin">AptosCoin</a>&gt;(amount, mint_cap);
     <a href="coin.md#0x1_coin_deposit">coin::deposit</a>&lt;<a href="aptos_coin.md#0x1_aptos_coin_AptosCoin">AptosCoin</a>&gt;(dst_addr, coins_minted);
 }
 </code></pre>
--- a/aptos-move/framework/aptos-framework/doc/coin.md
+++ b/aptos-move/framework/aptos-framework/doc/coin.md
@@ -35,11 +35,14 @@ This module provides the foundation for typesafe Coins.
 -  [Function `symbol`](#0x1_coin_symbol)
 -  [Function `decimals`](#0x1_coin_decimals)
 -  [Function `supply`](#0x1_coin_supply)
+-  [Function `burn_internal`](#0x1_coin_burn_internal)
 -  [Function `burn`](#0x1_coin_burn)
 -  [Function `burn_from`](#0x1_coin_burn_from)
 -  [Function `deposit`](#0x1_coin_deposit)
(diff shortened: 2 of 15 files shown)


movekevin (commented):
Would you mind elaborating more on why we need the friend-only internal versions of these functions (the description only says it's needed for total supply specification but doesn't explain why)? Looks like a decent amount of boilerplate code to me :)

rahxephon89:
> Would you mind elaborating more on why we need the friend-only internal versions of these functions (the description only says it's needed for total supply specification but doesn't explain why)? Looks like a decent amount of boilerplate code to me :)

Sorry I did not clearly explain it. I have updated the description of this PR. Please let me know if there are any concerns with these changes.

davidiw (changes requested):
Considering that calling functions is not free, it might be a better solution to update the move prover than coercing developers to pick up what looks like questionable programming paradigms.

wrwg:
> Considering that calling functions is not free, it might be a better solution to update the move prover than coercing developers to pick up what looks like questionable programming paradigms.

I disagree. The reason why this is refactored for the prover is not because the prover has a deficiency here, but simply because the semantics of the `foo` and `foo_internal` functions is different. The pre-conditions you can assume about those functions is different if they are public entry points vs if they are only called privately, which leads to different reasoning. The extra function call is minimal in gas cost IMO. Happy to discuss if you are not convinced. In general, it is very common in our verification effort that code may need to be rewritten a little for verifiability.

davidiw:
> 

Would love to learn more, I found this line:

> This would be legal only if `setup` is a private or `public (friend)` function. The difference between this and disabling invariants in `setup` is that the invariants would not be assumed at the beginning of `setup` and would be proved after `setup` returns at each site where it is called.

(From https://aptos.dev/move/prover/spec-lang#global-invariants)

Which seems to indicate that there's a nuance that the prover may not actually prove the following invariants inline but when proving the caller of this function?

Which makes me wonder... is it possible to then produce a private function that would prove the public function instead, something to the effect of

```
fun burn_for_prover(...) {
  burn(...)
```

and then prover could just prove that?

wrwg:
> > 
> 
> Would love to learn more, I found this line:
> 
> > This would be legal only if `setup` is a private or `public (friend)` function. The difference between this and disabling invariants in `setup` is that the invariants would not be assumed at the beginning of `setup` and would be proved after `setup` returns at each site where it is called.
> 
> (From https://aptos.dev/move/prover/spec-lang#global-invariants)
> 
> Which seems to indicate that there's a nuance that the prover may not actually prove the following invariants inline but when proving the caller of this function?

Yes, that is the effect of suspended invariants via `delegate_invariants_to_caller`.  However, we do not allow this pragma for public functions because they have unknown callers.

> 
> Which makes me wonder... is it possible to then produce a private function that would prove the public function instead, something to the effect of
> 
> ```
> fun burn_for_prover(...) {
>   burn(...)
> ```
> 
> and then prover could just prove that?

This unfortunately does not work, because what you are effectively are doing is to enforce that after you exit `burn`, whether called from inside the module or from outside, the invariants must hold. But the `delegate_invariants_to_caller` is transitive, and if the caller has the same attribute, the invariants will be postponed to its caller, and so on and so on.

That is why currently we simply disallow `delegate_invariants_to_caller` for public functions.

The alternative (and we had something like this implemented in earlier version of the Prover) would be to actually be able to reflect over from where a function is called. So effectively, there are two burn functions:

(a) the public one, which enforces invariants at exit
(b) the private one, which does not enforce them

Now you choose either the (a) or the (b) semantics depending on whether the next element on the caller stack is an external or internal function...

This is a terribly complicated model, and thats why we retired it. I hope you can see from this argument that independent of prover or not, it can be rather difficult to reason informally about such functions.

rahxephon89:
> The alternative (and we had something like this implemented in earlier version of the Prover) would be to actually be able to reflect over from where a function is called. So effectively, there are two burn functions:
> 
> (a) the public one, which enforces invariants at exit (b) the private one, which does not enforce them
> 

Hi @wrwg, sorry for the stupid question again. If a function is already marked as public, can we just go to the case (a)? Then,  if a public function is called in a private function, we just check whether the assumption holds by an assertion before calling it and output verification error if it does not?

rahxephon89:
Sorry for the delayed reply @davidiw and @wrwg. Here are some further and yet immature thoughts on this issue.

For public(friend) and private functions, the pragma `delegate_invariants_to_caller` can be used without any issues. For the public function, it is different because we do not know how it will be called and thus`delegate_invariants_to_caller` cannot be used. As a result, we introduce `disable_invariants_in_body`, another pragma for suspending the invariant check. With this pragma, we will assume the invariant at the entrance and then assert it at the exit of the function. If a function sets `disable_invariants_in_body`, then all its directly and indirect callees will turn-off invariant checking. Moreover, we have the following restrictions:

- invariant checking in an external facing function cannot be turned off, which means that public or script functions cannot be transitively called by functions disabling or delegating invariants;
- if a function is called (callee) in another function (caller) which has the pragma `disable_invariants_in_body` on, it means a suspendable invariant should not hold at any point in the function and thus the callee should not set the pragma `disable_invariants_in_body` .

And the restrictions above leads to code change in this PR:  the public function `on_new_epoch` in the `stake` module will call public APIs such as `mint`, `merge` or `extract_all` in the `coin` module. To prove a suspendable global invariant in `on_new_epoch` , the pragma `disable_invariants_in_body` is set. To prove some other global invariants in the `coin` module , `mint` and `merge` also have `disable_invariants_in_body` set, which causes the prover error.

I am wondering whether the current rules may be overly restrictive for public functions. In my opinion, for a public function, **no matter where it is called, all related global invariants must hold at both entrance and exit**. Thus, I am proposing that we relax the restrictions above and allow public functions to be called directly or indirectly by another function with `disable_invariants_in_body` being set. Then we insert assert before the call and assume after the call (for related global invariants) for them. If the prover outputs an verification error, it means the precondition is not strong enough and we may need to insert some spec blocks in the code to enhance the precondition.

PR #8623, opened 2023-06-12
[Storage Service] Add simple wait to optimistic fetch tests.

JoshLind:
### Description
This PR adds a simple wait to the optimistic fetch test cases. This should hopefully help to reduce some test flakes (where the test runner thread runs faster than the optimistic fetch handler).

### Test Plan
Existing test infrastructure.

### DIFF
--- a/state-sync/storage-service/server/src/tests/utils.rs
+++ b/state-sync/storage-service/server/src/tests/utils.rs
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use crate::{
+    optimistic_fetch::OptimisticFetchRequest,
     storage::StorageReader,
     tests::mock::{MockClient, MockDatabaseReader},
     StorageServiceServer,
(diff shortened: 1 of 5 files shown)


state-sync/storage-service/server/src/tests/utils.rs
```diff
@@ -429,3 +432,19 @@ pub async fn wait_for_optimistic_fetch_service_to_refresh(
...
 }
+
+/// Waits for the specified number of optimistic fetches to be active
+pub async fn wait_for_active_optimistic_fetches(
+    active_optimistic_fetches: Arc<Mutex<HashMap<PeerNetworkId, OptimisticFetchRequest>>>,
+    expected_num_active_fetches: usize,
+) {
+    loop {
+        let num_active_fetches = active_optimistic_fetches.lock().len();
+        if num_active_fetches == expected_num_active_fetches {
```
bchocho, line 443: Is it always an exact match? If not, >=

bchocho (approved):
Lgtm

state-sync/storage-service/server/src/tests/utils.rs
grao1991: If yes, handle the case for >.

grao1991 (approved):

state-sync/storage-service/server/src/tests/utils.rs
JoshLind: Yeah, it's always an exact match 😄 Anything else would be a bug in the code.

PR #8638, opened 2023-06-13
[mempool] Replace NewPeer/LostPeer events with periodic polling of network connections.

bchocho:
### Description

Instead of using NewPeer/LostPeer events, periodically poll PeersAndMetadata and change broadcasting peers accordingly. This has two advantages:
(1) Get closer to removing NewPeer/LostPeer completely
(2) Prepare for mempool to later use the version information contained in PeersAndMetadata.

### Test Plan

Behavior-wise this should be a noop. So relying on existing tests.

### DIFF
--- a/aptos-node/src/services.rs
+++ b/aptos-node/src/services.rs
@@ -6,6 +6,7 @@ use aptos_build_info::build_information;
 use aptos_config::config::NodeConfig;
 use aptos_consensus::network_interface::ConsensusMsg;
 use aptos_consensus_notifications::ConsensusNotifier;
+use aptos_data_client::interface::AptosPeersInterface;
 use aptos_event_notifications::ReconfigNotificationListener;
 use aptos_indexer_grpc_fullnode::runtime::bootstrap as bootstrap_indexer_grpc;
 use aptos_logger::{debug, telemetry_log_writer::TelemetryLog, LoggerFilterUpdater};
@@ -107,6 +108,7 @@ pub fn start_mempool_runtime_and_get_consensus_sender(
     network_interfaces: ApplicationNetworkInterfaces<MempoolSyncMsg>,
     mempool_listener: MempoolNotificationListener,
     mempool_client_receiver: Receiver<MempoolClientRequest>,
+    peers: Arc<dyn AptosPeersInterface>,
 ) -> (Runtime, Sender<QuorumStoreRequest>) {
     // Create a communication channel between consensus and mempool
     let (consensus_to_mempool_sender, consensus_to_mempool_receiver) =
@@ -123,6 +125,7 @@ pub fn start_mempool_runtime_and_get_consensus_sender(
         consensus_to_mempool_receiver,
         mempool_listener,
         mempool_reconfig_subscription,
+        peers,
     );
     debug!("Mempool started in {} ms", instant.elapsed().as_millis());
 
--- a/aptos-node/src/services.rs
+++ b/aptos-node/src/services.rs
@@ -6,7 +6,6 @@ use aptos_build_info::build_information;
 use aptos_config::config::NodeConfig;
 use aptos_consensus::network_interface::ConsensusMsg;
 use aptos_consensus_notifications::ConsensusNotifier;
-use aptos_data_client::interface::AptosPeersInterface;
 use aptos_event_notifications::ReconfigNotificationListener;
 use aptos_indexer_grpc_fullnode::runtime::bootstrap as bootstrap_indexer_grpc;
 use aptos_logger::{debug, telemetry_log_writer::TelemetryLog, LoggerFilterUpdater};
@@ -108,7 +107,7 @@ pub fn start_mempool_runtime_and_get_consensus_sender(
     network_interfaces: ApplicationNetworkInterfaces<MempoolSyncMsg>,
     mempool_listener: MempoolNotificationListener,
     mempool_client_receiver: Receiver<MempoolClientRequest>,
-    peers: Arc<dyn AptosPeersInterface>,
+    peers: Arc<PeersAndMetadata>,
 ) -> (Runtime, Sender<QuorumStoreRequest>) {
     // Create a communication channel between consensus and mempool
     let (consensus_to_mempool_sender, consensus_to_mempool_receiver) =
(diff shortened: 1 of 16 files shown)


mempool/src/shared_mempool/coordinator.rs
```diff
@@ -258,33 +279,11 @@ async fn handle_network_event<NetworkClient, TransactionValidator>(
...
-            debug!(LogSchema::new(LogEntry::NewPeer)
-                .peer(&peer)
-                .is_upstream_peer(is_upstream_peer));
-            notify_subscribers(SharedMempoolNotification::PeerStateChange, &smp.subscribers);
-            if is_new_peer && is_upstream_peer {
-                tasks::execute_broadcast(peer, false, smp, scheduled_broadcasts, executor.clone())
-                    .await;
-            }
+        Event::NewPeer(_) => {
+            // TODO: remove Event
```
JoshLind, line 283: Awesome, thanks for removing these 😄

mempool/src/shared_mempool/network.rs
```diff
@@ -103,38 +103,66 @@ impl<NetworkClient: NetworkClientInterface<MempoolSyncMsg>> MempoolNetworkInterf
...
+        let to_disable: Vec<_> = sync_states
+            .keys()
+            .filter(|previous_peer| !updated_peers.contains_key(previous_peer))
+            .copied()
+            .collect();
+        let to_add: Vec<_> = updated_peers
+            .iter()
+            .filter(|(peer, _)| !sync_states.contains_key(peer))
+            .map(|(peer, metadata)| (*peer, metadata.get_connection_metadata()))
+            .filter(|(peer, metadata)| self.is_upstream_peer(peer, Some(metadata)))
```
JoshLind, line 121: nit: any chance we could call out in the function comment that we only add upstream peers? 😄 Just to make it obvious that we aren't tracking all peers.

mempool/src/shared_mempool/network.rs
```diff
@@ -103,38 +103,66 @@ impl<NetworkClient: NetworkClientInterface<MempoolSyncMsg>> MempoolNetworkInterf
...
+        to_add: &[(PeerNetworkId, ConnectionMetadata)],
+        to_disable: &[PeerNetworkId],
+    ) {
         let mut sync_states = self.sync_states.write();
-        let is_new_peer = !sync_states.contains_key(&peer);
-        if self.is_upstream_peer(&peer, Some(&metadata)) {
-            // If we have a new peer, let's insert new data, otherwise, let's just update the current state
-            if is_new_peer {
+        for (peer, metadata) in to_add.iter().cloned() {
+            if self.is_upstream_peer(&peer, Some(&metadata)) {
```
JoshLind, line 134: nit: do we still need this check if we only add upstream peers above?

mempool/src/shared_mempool/network.rs
```diff
@@ -103,38 +103,66 @@ impl<NetworkClient: NetworkClientInterface<MempoolSyncMsg>> MempoolNetworkInterf
...
-    /// Disables a peer if it can be restarted, otherwise removes it
-    pub fn disable_peer(&self, peer: PeerNetworkId) {
-        // All other nodes have their state immediately restarted anyways, so let's free them
-        if self.sync_states.write().remove(&peer).is_some() {
-            counters::active_upstream_peers(&peer.network_id()).dec();
+    /// Update peers based on updated view of connected peers. Return (peers newly added that need
+    /// to start broadcasts, peers that will be disabled from broadcasts).
+    pub fn update_peers(
+        &self,
+        updated_peers: &HashMap<PeerNetworkId, PeerMetadata>,
```
JoshLind, line 154: nit: does it make sense for this to be called something like "all_connected_peers"?

state-sync/aptos-data-client/src/peer_states.rs
```diff
@@ -370,6 +370,13 @@ impl PeerStates {
         }
     }
 
+    /// Returns metadata for all peers currently connected to the node
+    pub(crate) fn get_connected_peers_and_metadata(
+        &self,
+    ) -> Result<HashMap<PeerNetworkId, PeerMetadata>, Error> {
+        self.peers_and_metadata.get_connected_peers_and_metadata()
```
JoshLind, line 377: This is the only thing I'm a little iffy on 🤔 Technically, if all you want is the peers and metadata information, we should pass the `PeersAndMetadata` struct to mempool (as originally planned) and not the data client. But, I assume you've done this because you later want access to the more fine-grained synced version info? (Does that make sense?) 😃 

Let's sync offline -- I think there's a couple options we have and we can see which makes the most sense 😄

JoshLind (commented):
Thanks @bchocho! Excited to see the notifications go 😄 Only thing we need to sync on is the peers and metadata info paths.

state-sync/aptos-data-client/src/peer_states.rs
bchocho: Yup! Thats exactly the reason. Lets sync offline -- I'd like to get your feedback on the finer grained peer info my prototype is using.

mempool/src/shared_mempool/network.rs
bchocho: Good catch!

aptos-node/src/services.rs
```diff
@@ -107,6 +107,7 @@ pub fn start_mempool_runtime_and_get_consensus_sender(
     network_interfaces: ApplicationNetworkInterfaces<MempoolSyncMsg>,
     mempool_listener: MempoolNotificationListener,
     mempool_client_receiver: Receiver<MempoolClientRequest>,
+    peers: Arc<PeersAndMetadata>,
```
JoshLind, line 110: nit: `peers` -> `peers_and_metadata`?

JoshLind (approved):
Thanks @bchocho! LGTM.

mempool/src/shared_mempool/coordinator.rs
```diff
@@ -103,7 +110,25 @@ pub(crate) async fn coordinator<NetworkClient, TransactionValidator>(
                 tasks::execute_broadcast(peer, backoff, &mut smp, &mut scheduled_broadcasts, executor.clone()).await;
             },
             (network_id, event) = events.select_next_some() => {
-                handle_network_event(&executor, &bounded_executor, &mut scheduled_broadcasts, &mut smp, network_id, event).await;
+                handle_network_event(&bounded_executor, &mut smp, network_id, event).await;
+            },
+            _ = update_peers_interval.tick().fuse() => {
+                if let Ok(connected_peers) = peers_and_metadata.get_connected_peers_and_metadata() {
```
brianolson, line 116: nit? every other branch of this `select!` has been pulled out into a helper function so that it's a one-liner here

brianolson (approved):

PR #8639, opened 2023-06-13
[Spec] update specs of resource_account

UIZorrot:
### Description

This is a part of spec work for Aptos Framework from MoveBit. The changes are:

Completed most of the TODO in resource_account.move;
There is one function related to code.move is not verified.

### Test Plan

`aptos move prove -f resource_account
`


aptos-move/framework/aptos-framework/sources/resource_account.spec.move
```diff
@@ -9,8 +9,24 @@ spec aptos_framework::resource_account {
         seed: vector<u8>,
         optional_auth_key: vector<u8>,
     ) {
-        // TODO: Could not verify `rotate_account_authentication_key_and_store_capability` because can't get `resource` and `resource_signer_cap`.
-        pragma verify = false;
+        pragma verify = true;
```
junkil-park, line 12: you can remove this line.

aptos-move/framework/aptos-framework/sources/resource_account.spec.move
```diff
@@ -19,8 +35,40 @@ spec aptos_framework::resource_account {
         optional_auth_key: vector<u8>,
         fund_amount: u64,
     ) {
-        // TODO: Could not verify `rotate_account_authentication_key_and_store_capability` because can't get `resource` and `resource_signer_cap`.
-        pragma verify = false;
+        pragma verify = true;
```
junkil-park, line 38: you can remove this line.

aptos-move/framework/aptos-framework/sources/resource_account.spec.move
```diff
@@ -29,9 +77,26 @@ spec aptos_framework::resource_account {
         metadata_serialized: vector<u8>,
         code: vector<vector<u8>>,
     ) {
-        // TODO: Calls `code::publish_package_txn`.
-        // TODO: Could not verify `code::publish_package_txn` because can't get `resource` and `resource_signer_cap`.
         pragma verify = false;
```
junkil-park, line 80: Why is this not verified?

UIZorrot: It is related to a vector enum loop in code.move, which we can't verify yet

UIZorrot: by the way, if we used aborts_if_is_partial, it will throw an boogie error:  Error: wrong number of result variables in call to $1_vector_length'$1_code_PackageMetadata':

junkil-park (approved):
Overall good. It seems many lines of specs are coming from the functions in the `coin` module. Could you use schemas to reduce the repetition?

aptos-move/framework/aptos-framework/sources/resource_account.spec.move
rahxephon89: @UIZorrot Could you file an issue for this internal boogie error? Thanks

aptos-move/framework/aptos-framework/sources/resource_account.spec.move
```diff
@@ -60,6 +74,27 @@ spec aptos_framework::resource_account {
         aborts_if !get && !(exists<Account>(resource_addr) && len(optional_auth_key) == 32);
     }
 
+    spec schema RotateAccountAuthenticationKeyAndStoreCapabilityAbortsIfWithoutAccountLimit {
```
rahxephon89, line 77: I am wondering if we can include RotateAccountAuthenticationKeyAndStoreCapabilityAbortsIf to reduce duplicate?

rahxephon89 (changes requested):

aptos-move/framework/aptos-framework/sources/resource_account.spec.move
UIZorrot: well as you may see, there is a little difference between these two schema, which are the last two lines:

rahxephon89 (approved):

PR #8621, opened 2023-06-12
[move-examples] Add multiple examples for resource accounts

xbtmatt:
### Description
Merged the publishing code from this module with the [ package manager in PR #8641. ](https://github.com/aptos-labs/aptos-core/pull/8641)

Added three sections to the `examples/resource-accounts` section in the docs.

The tutorial shows examples and sample code that you can compile and publish to test it.

### Test Plan
There is a unit test in `upgradeable_resource_account_package.rs` in the unit tests folder:

`RUST_MIN_STACK=4297152 cargo test --package e2e-move-tests --lib -- tests::upgradeable_resource_account_package::code_upgrading_using_resource_account --exact --nocapture`

The tutorial shows a more complete example of how to do it using the `build-publish-package` command to generate a JSON file with the module bytecode.

PR #8641, opened 2023-06-13
[Move examples] Add an example for managing resource accounts when deploying

movekevin:
### Description
/// This module allows publishing to a resource account and retaining control its signer for future upgrades or
/// for other purposes such as creating an NFT collection. It also offer features around managing object and resource
/// account addresses so those can be easily accessed in other modules.
///
/// The deployment flow is as follows:
/// 1. Deploy the package, including this package_manager module, using the Aptos CLI command create-resource-and-publish-package
/// with an appropriate seed. This will create a resource account and deploy the module. The deployer address also needs
/// to be specified in Move.toml.
/// 2. Make sure the created resource address is persisted in the Move.toml for future deployments and upgrades as the
/// CLI doesn't do so by default.
/// 3. During deployment, package_manager::init_module will be called and extract the SignerCapability from the newly
/// created resource account.
/// 4. All other modules from the same package that are friends can call package_manager to obtain the resource account
/// signer when needed. If cross-package access is needed, authorization can be granted via an address-based whitelist
/// instead of friendship, which is limited to the same package.
/// 5. If new modules need to be deployed or existing modules in this package need to be updated, an assigned admin
/// account (defaults to the deployer account) can call package_manager::publish_package to with the new code.
///
/// Other modules can store and obtain stored addreses by calling add_address or get_address. This is useful for
/// storing addresses of other modules in the same package or system addresses such as the NFT collection.

### Test Plan
Unit tests and manual tests

### DIFF
--- a/aptos-move/move-examples/package_manager/Move.toml
+++ b/aptos-move/move-examples/package_manager/Move.toml
@@ -1,0 +1,11 @@
+[package]
+name = "package_manager"
+version = "1.0.0"
+upgrade_policy = "compatible"
+
+[dependencies]
+AptosFramework = { local = "../../framework/aptos-framework" }
+
+[addresses]
+deployer = "_"
+package = "_"
--- a/aptos-move/move-examples/package_manager/sources/package_manager.move
+++ b/aptos-move/move-examples/package_manager/sources/package_manager.move
@@ -1,0 +1,86 @@
+/// This module allows publishing to a resource account and retaining control its signer for future upgrades or
+/// for other purposes such as creating an NFT collection. It also offer features around managing object and resource
+/// account addresses so those can be easily accessed in other modules.
+///
+/// The deployment flow is as follows:
+/// 1. Deploy the package, including this package_manager module, using the Aptos CLI command create-resource-and-publish-package
+/// with an appropriate seed. This will create a resource account and deploy the module. The deployer address also needs
+/// to be specified in Move.toml.
+/// 2. Make sure the created resource address is persisted in the Move.toml for future deployments and upgrades as the
+/// CLI doesn't do so by default.
+/// 3. During deployment, package_manager::init_module will be called and extract the SignerCapability from the newly
+/// created resource account.
+/// 4. All other modules from the same package that are friends can call package_manager to obtain the resource account
+/// signer when needed. If cross-package access is needed, authorization can be granted via an address-based whitelist
+/// instead of friendship, which is limited to the same package.
+/// 5. If new modules need to be deployed or existing modules in this package need to be updated, an assigned admin
+/// account (defaults to the deployer account) can call package_manager::publish_package to with the new code.
+///
+/// Other modules can store and obtain stored addreses by calling add_address or get_address. This is useful for
+/// storing addresses of other modules in the same package or system addresses such as the NFT collection.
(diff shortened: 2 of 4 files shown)


aptos-move/move-examples/resource_account_package/sources/package_manager.move
```diff
@@ -0,0 +1,83 @@
...
+
+    /// Initialize PermissionConfig to establish control over the resource account.
+    /// This function is invoked only when this package is deployed the first time.
+    fun init_module(resource_acc: &signer) {
+        let signer_cap = resource_account::retrieve_resource_account_cap(resource_acc, @deployer);
+        move_to(resource_acc, PermissionConfig { signer_cap, upgrade_admin: @deployer });
+    }
+
+    /// Can be called by friended modules to obtain the resource account signer.
+    public(friend) fun get_signer(): signer acquires PermissionConfig {
```
davidiw, line 48: might it be good to show a test-only example here?

aptos-move/move-examples/resource_account_package/sources/package_manager.move
```diff
@@ -0,0 +1,83 @@
...
+    public entry fun publish(
+        admin: &signer,
+        package_metadata: vector<u8>,
+        code: vector<vector<u8>>,
+    ) acquires PermissionConfig {
+        let package_signer = authorized_create_signer(admin);
+        code::publish_package_txn(package_signer, package_metadata, code);
+    }
+
+    inline fun authorized_create_signer(admin: &signer): &signer {
```
davidiw, line 63: this is only used once, let's not inline it?

davidiw (approved):

aptos-move/move-examples/resource_account_package/sources/package_manager.move
xbtmatt: I think a `#[test_only]` example makes sense here, maybe with a separate test module in the same file to show how to easily declare it a friend and obtain the signer

xbtmatt:
Tested it as if I were actually using it, works nicely and is pretty nifty!

There are issues with the dev ex but it's tangential to this, I'll try to address it in #8621 or in another PR, because I think it boils down to the CLI needing features it doesn't have (e.g. storing resource address in Move.toml file upon publish)

xbtmatt (approved):

aptos-move/move-examples/resource_account_package/sources/package_manager.move
movekevin: There's only a test that gets the signer as a showcase. Furthermore this same package_manager is already incorporated into other examples such as move-examples/swap, rewards pool (https://github.com/aptos-labs/aptos-core/pull/9056) and vote_lockup (https://github.com/aptos-labs/aptos-core/pull/9095)

aptos-move/move-examples/resource_account_package/sources/package_manager.move
movekevin: I want to show the clear patterns of only doing borrow_global and borrow_global_mut in designated functions. This can make code a lot more aware of global storage accesses in case authorizations or validations are needed. I've been doing this consistently in other examples as well (swap, rewards pool, vote_lockup)

PR #8642, opened 2023-06-13
net perf test direct

brianolson:
### Description

Add direct send performance test, normally disabled under compile time flag `network-perf-test`

### Test Plan

This is test code. Will be used in local and cluster experiments.


brianolson:
replaced by https://github.com/aptos-labs/aptos-core/pull/8892

PR #8643, opened 2023-06-13
[Helm] Fix loadtest cronjob

sionescu:
### Description

* change binary name
* add required command line parameters to chart values

### Test Plan

Test live on Devnet.


rustielin (approved):

perryjrandall (approved):
Is there a way to test this or make this easier to modify?

The templating here is intense

PR #8644, opened 2023-06-13
Specify that functions can return references under certain conditions

saharct:
### Description

Added a note in the Move Book under functions.md#return-type

### DIFF
--- a/developer-docs-site/docs/move/book/functions.md
+++ b/developer-docs-site/docs/move/book/functions.md
@@ -327,6 +327,10 @@ fun zero(): u64 { 0 }
 
 Here `: u64` indicates that the function's return type is `u64`.
 
+:::tip
+A function can return an immutable `&` or mutable `&mut` [reference](./references.md) if derived from an input reference. Keep in mind, this means that a function [cannot return a reference to global storage](./references.md#references-cannot-be-stored).
+:::
+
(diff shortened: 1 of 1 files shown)


developer-docs-site/docs/move/book/functions.md
```diff
@@ -327,6 +327,10 @@ fun zero(): u64 { 0 }
 
 Here `: u64` indicates that the function's return type is `u64`.
 
+:::tip
+A function can return an immutable `&` or mutable `&mut` [reference](./references.md) if derived from an input reference. Keep in mind, this means that a function [cannot return a reference to global storage](./references.md#references-cannot-be-stored).
```
movekevin, line 331: I wonder if we should mention here that inline functions can actually return reference to global state as well

movekevin (approved):

developer-docs-site/docs/move/book/functions.md
gregnazario: Yeah, we should probably mention inline as well

gregnazario (approved):

developer-docs-site/docs/move/book/functions.md
saharct: Sounds good. I'll add it to the tip.

PR #8652, opened 2023-06-14
drop pruned blocks asynchronously

msmouse:
### Description

helps with performance when blocks are large.

### Test Plan

tests (fixed)

### DIFF
--- a/execution/executor/src/components/block_tree/mod.rs
+++ b/execution/executor/src/components/block_tree/mod.rs
@@ -7,7 +7,10 @@
 #[cfg(test)]
 mod test;
 
-use crate::logging::{LogEntry, LogSchema};
+use crate::{
+    logging::{LogEntry, LogSchema},
+    metrics::APTOS_EXECUTOR_OTHER_TIMERS_SECONDS,
+};
 use anyhow::{anyhow, ensure, Result};
 use aptos_consensus_types::block::Block as ConsensusBlock;
 use aptos_crypto::HashValue;
@@ -18,7 +21,10 @@ use aptos_storage_interface::DbReader;
 use aptos_types::{ledger_info::LedgerInfo, proof::definition::LeafCount};
 use std::{
     collections::{hash_map::Entry, HashMap},
-    sync::{Arc, Weak},
+    sync::{
+        mpsc::{channel, Receiver},
+        Arc, Weak,
+    },
 };
 
 pub struct Block {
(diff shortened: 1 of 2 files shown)


execution/executor/src/components/block_tree/mod.rs
```diff
@@ -250,8 +261,26 @@ impl BlockTree {
...
+        // guarantees that the drop will not happen in the current thread
+        let (tx, rx) = channel::<()>();
+        rayon::spawn(move || {
+            let _timeer = APTOS_EXECUTOR_OTHER_TIMERS_SECONDS
+                .with_label_values(&["drop_old_root"])
+                .start_timer();
+            drop(old_root);
+            // Error is ignored, since the caller might not care about dropping completion and
+            // has discarded the receiver already.
+            tx.send(()).ok();
```
zekun000, line 281: you just want the join handle right?

msmouse: `spawn()` doesn't return one.

areshand (approved):

execution/executor/src/components/block_tree/mod.rs
msmouse: Spawning a new thread will be slower in theory?

zekun000: I guess it's probably faster than the channel lol

grao1991: The cost of spawning a thread should be ~10us, not too bad.

grao1991 (approved):

execution/executor/src/components/block_tree/mod.rs
zekun000: just allocating the channel is not a fair comparison, you need to send something in a separate rayon task lol

msmouse: Yeah but that's not important because it's async and noboday receive in production.

msmouse: Ideally we have a global drop pool

zekun000: it's just ugly but anyway spawning thread is much more expensive than i expected

msmouse: landing ugly for now. maybe creating the drop pool "some time in the future"

PR #8655, opened 2023-06-14
[Spec] update specs of reconfiguration

UIZorrot:
### Description

There are minor changes in the specification of `reconfiguration.move`

### Test Plan
`aptos move prove -f recon `


aptos-move/framework/aptos-framework/sources/reconfiguration.spec.move
```diff
@@ -67,7 +67,9 @@ spec aptos_framework::reconfiguration {
         use aptos_framework::transaction_fee;
         use aptos_framework::staking_config;
 
-        pragma verify_duration_estimate = 120; // TODO: set because of timeout (property proved)
+        //We use aborts_if_is_partial to avoid timeout
+        //All properties is verifed
+        pragma aborts_if_is_partial;
```
junkil-park, line 72: Why did you change it to `aborts_if_is_partial`? Any problem with `verify_duration_estimate`?

PR #8664, opened 2023-06-14
[Storage] Reorganize calculation and commit tasks in save_transactions/save_transaction_block.

grao1991:
### Description
To get a better performance, still have room to tune/improve.

### DIFF
--- a/storage/aptosdb/src/lib.rs
+++ b/storage/aptosdb/src/lib.rs
@@ -74,7 +74,7 @@ use aptos_config::config::{
 use aptos_config::config::{
     BUFFERED_STATE_TARGET_ITEMS, DEFAULT_MAX_NUM_NODES_PER_LRU_CACHE_SHARD,
 };
-use aptos_crypto::hash::HashValue;
+use aptos_crypto::HashValue;
 use aptos_db_indexer::Indexer;
 use aptos_infallible::Mutex;
 use aptos_logger::prelude::*;
@@ -115,7 +115,6 @@ use aptos_types::{
 };
 use aptos_vm::data_cache::AsMoveResolver;
 use arr_macro::arr;
-use itertools::zip_eq;
 use move_resource_viewer::MoveValueAnnotator;
 use once_cell::sync::Lazy;
 use rayon::prelude::*;
@@ -275,16 +274,6 @@ impl Drop for RocksdbPropertyReporter {
             .join()
             .expect("Rocksdb property reporting thread should join peacefully.");
     }
-}
-
-#[derive(Default)]
-struct LedgerSchemaBatches {
-    ledger_metadata_batch: SchemaBatch,
-    event_batch: SchemaBatch,
-    transaction_batch: SchemaBatch,
-    transaction_info_batch: SchemaBatch,
-    transaction_accumulator_batch: SchemaBatch,
-    write_set_batch: SchemaBatch,
 }
 
 /// This holds a handle to the underlying DB responsible for physical storage and provides APIs for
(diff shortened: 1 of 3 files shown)


storage/aptosdb/src/lib.rs
```diff
@@ -1006,100 +867,278 @@ impl AptosDB {
...
+    ) -> Result<()> {
+        let _timer = OTHER_TIMERS_SECONDS
+            .with_label_values(&["commit_transactions"])
+            .start_timer();
+        let chunk_size = 512;
+        txns_to_commit
+            .par_chunks(chunk_size)
+            .enumerate()
+            .try_for_each(|(chunk_index, txns_in_chunk)| -> Result<()> {
+                let batch = SchemaBatch::new();
```
msmouse, line 1005: Does it really matter to use multiple batches? (and why does it matter only for transactions?)

msmouse (approved):
Looks legit

storage/aptosdb/src/lib.rs
```diff
@@ -1006,100 +867,278 @@ impl AptosDB {
...
+        first_version: Version,
+        expected_state_db_usage: StateStorageUsage,
+        sharded_state_cache: Option<&ShardedStateCache>,
+    ) -> Result<HashValue> {
+        let new_root_hash = thread::scope(|s| {
+            let _timer = OTHER_TIMERS_SECONDS
+                .with_label_values(&["save_transactions__work"])
+                .start_timer();
+            // TODO(grao): Write progress for each of the following databases, and handle the
+            // inconsistency at the startup time.
```
zekun000, line 882: not related to this pr, but we can't recover if it crashes in the middle? or we just ignore anything that's not covered by ledger info?

grao1991: I have a TODO to truncate those versions that are larger than ledger info version. It's probably not that critical if we assume there is no fork.

(It seems there are some places that call get_latest_transaction_info_option to get the latest version, we probably want to get the ledger info version instead, I will work on it in a separate PR)

zekun000: truncate is not a high pri, but what happens if txn has a different version than event for example. I thought state sync is using the last transaction info to fetch instead of last ledger info?

grao1991: Yes, that's a problem I was saying in the "()" above. I need to get it done before 1.6 cut. (it's not in the 1.5)

zekun000 (approved):

storage/aptosdb/src/lib.rs
grao1991: It's ~10% faster in my benchmark.. (I only did it for txn because it is relatively larger comparing to other data types)

btw, I don't like it.

storage/aptosdb/src/lib.rs
msmouse: >  get the ledger info version instead

the overall progress, you mean

grao1991: yes ^

storage/aptosdb/src/lib.rs
```diff
@@ -1006,100 +867,278 @@ impl AptosDB {
...
+        last_version: Version,
+        new_root_hash: HashValue,
+        ledger_info_with_sigs: Option<&LedgerInfoWithSignatures>,
+    ) -> Result<()> {
+        let _timer = OTHER_TIMERS_SECONDS
+            .with_label_values(&["commit_ledger_info"])
+            .start_timer();
 
         let ledger_batch = SchemaBatch::new();
-        self.save_ledger_info(new_root_hash, ledger_info_with_sigs, &ledger_batch)?;
```
msmouse, line 1102: This is missing in the new state sync flow (until a ledger info is saved)?

storage/aptosdb/src/lib.rs
```diff
@@ -1006,100 +867,278 @@ impl AptosDB {
...
-                ledger_schema_batches
-                    .ledger_metadata_batch
-                    .put::<DbMetadataSchema>(
-                        &DbMetadataKey::LedgerCommitProgress,
-                        &DbMetadataValue::Version(last_version),
-                    )
-                    .unwrap();
-                self.ledger_db
-                    .metadata_db()
-                    .write_schemas(ledger_schema_batches.ledger_metadata_batch)
```
msmouse, line 1046: hmm.. this already makes the `LedgerCommitProgress` meaningless?

PR #8665, opened 2023-06-14
[BlockSTM] Remove Commit Lock

danielxiangzl:
### Description
This PR removes the commit lock in the `TxnLastInputOutput`, by letting `Coordinator` send the `txn_id` to the `Worker` only after `Coordinator` finishes its work. Then, there is no race between `Coordinator` and `Worker` and there is no need for the commit lock. 

This will also be useful for adding `FeeStatement` later, as `FeeStatement` will be stored in `VMOutput`, and we want it to be available when `Coordinator` reads it for calculating the block gas. This PR makes sure that `Worker` only takes the `VMOutput` after `Coordinator` finishes its work.
### Test Plan

Existing tests.

### DIFF
--- a/aptos-move/block-executor/src/executor.rs
+++ b/aptos-move/block-executor/src/executor.rs
@@ -229,13 +229,8 @@ where
         scheduler_task: &mut SchedulerTask,
         last_input_output: &TxnLastInputOutput<T::Key, E::Output, E::Error>,
     ) {
+        let mut remaining_work = None;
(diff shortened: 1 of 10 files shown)


zekun000 (commented):
lgtm but wdyt about a fancier drop guard that does the work?

zekun000 (approved):

msmouse (approved):
I'm blind.

PR #8670, opened 2023-06-14
[Fix] Return the correct version for state snapshot when the root is not committed.

grao1991:

### DIFF
--- a/storage/aptosdb/src/state_store/state_store_test.rs
+++ b/storage/aptosdb/src/state_store/state_store_test.rs
@@ -4,18 +4,23 @@
 
 use super::*;
 use crate::{
+    jellyfish_merkle_node::JellyfishMerkleNodeSchema,
     new_sharded_kv_schema_batch,
     state_restore::StateSnapshotRestore,
     test_helper::{arb_state_kv_sets, update_store},
     AptosDB,
 };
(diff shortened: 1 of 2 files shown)


zekun000 (approved):

storage/aptosdb/src/state_store/state_store_test.rs
```diff
@@ -287,7 +292,43 @@ pub fn test_get_state_snapshot_before() {
     assert_eq!(store.get_state_snapshot_before(3).unwrap(), Some((2, hash)));
     assert_eq!(store.get_state_snapshot_before(2).unwrap(), Some((0, hash)));
     assert_eq!(store.get_state_snapshot_before(1).unwrap(), Some((0, hash)));
-    assert_eq!(store.get_state_snapshot_before(0).unwrap(), None,);
+    assert_eq!(store.get_state_snapshot_before(0).unwrap(), None);
+
+    // Only consider a version as available when the root node is there.
```
msmouse, line 297: Comment a bit more on why you put in another node.

msmouse (approved):

storage/aptosdb/src/state_store/state_store_test.rs
grao1991: Done

PR #8674, opened 2023-06-14
[aptos-vm] Skip converting storage error

runtian-zhou:
### Description

Storage error can be a result of speculation error and should thus not be treated as `UNEXPECTED_ERROR_FROM_KNOWN_MOVE_FUNCTION`.

### Test Plan

TBD

### DIFF
--- a/aptos-move/aptos-vm/src/errors.rs
+++ b/aptos-move/aptos-vm/src/errors.rs
@@ -132,6 +132,10 @@ pub fn convert_prologue_error(
             };
             VMStatus::Error(new_major_status, None)
         },
+        // Storage error can be a result of speculation failure so throw the error back for caller to handle.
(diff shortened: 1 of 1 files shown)


zekun000 (approved):

aptos-move/aptos-vm/src/errors.rs
```diff
@@ -132,6 +132,10 @@ pub fn convert_prologue_error(
             };
             VMStatus::Error(new_major_status, None)
         },
+        // Storage error can be a result of speculation failure so throw the error back for caller to handle.
+        VMStatus::Error(StatusCode::STORAGE_ERROR, msg) => {
```
zekun000, line 136: I think you can write `e @ VMStatus::Error(_) => e`

sherry-x (approved):

rahxephon89 (approved):

PR #8677, opened 2023-06-15
[GHA] Use spot VM for executor-performance

ibalajiarun:
### Description

This PR duplicates the executor-performance workflow to run on `spot-runner` GHA runners which are backed by spot VM pool. We will use this to assess the availability and performance of the spot VM pool and evaluate if we can expand its use.

 The duplicated workflow is not required check, so won't block merge on failure.

### Test Plan

Tested it in this run: https://github.com/aptos-labs/aptos-core/actions/runs/5273676507


rustielin (approved):

danielxiangzl (approved):
stamp

PR #8679, opened 2023-06-15
Add readable type info support

alnoki:
@movekevin

### DIFF
--- a/aptos-move/framework/aptos-stdlib/doc/type_info.md
+++ b/aptos-move/framework/aptos-stdlib/doc/type_info.md
@@ -12,6 +12,7 @@
 -  [Function `module_name`](#0x1_type_info_module_name)
 -  [Function `struct_name`](#0x1_type_info_struct_name)
 -  [Function `to_readable_type_info`](#0x1_type_info_to_readable_type_info)
+-  [Function `to_type_info`](#0x1_type_info_to_type_info)
 -  [Function `chain_id`](#0x1_type_info_chain_id)
 -  [Function `type_of`](#0x1_type_info_type_of)
 -  [Function `type_name`](#0x1_type_info_type_name)
@@ -224,6 +225,34 @@ For example, a trading application with a global registry of phantom <code>CoinT
         account_address: <a href="type_info.md#0x1_type_info">type_info</a>.account_address,
         module_name: <a href="../../move-stdlib/doc/string.md#0x1_string_utf8">string::utf8</a>(<a href="type_info.md#0x1_type_info">type_info</a>.module_name),
         struct_name: <a href="../../move-stdlib/doc/string.md#0x1_string_utf8">string::utf8</a>(<a href="type_info.md#0x1_type_info">type_info</a>.struct_name)
+    }
+}
+</code></pre>
+
+
+
+</details>
+
+<a name="0x1_type_info_to_type_info"></a>
+
+## Function `to_type_info`
+
+
+
+<pre><code><b>public</b> <b>fun</b> <a href="type_info.md#0x1_type_info_to_type_info">to_type_info</a>(readable_type_info: &<a href="type_info.md#0x1_type_info_ReadableTypeInfo">type_info::ReadableTypeInfo</a>): <a href="type_info.md#0x1_type_info_TypeInfo">type_info::TypeInfo</a>
+</code></pre>
+
+
+
+<details>
+<summary>Implementation</summary>
+
+
+<pre><code><b>public</b> <b>fun</b> <a href="type_info.md#0x1_type_info_to_type_info">to_type_info</a>(readable_type_info: &<a href="type_info.md#0x1_type_info_ReadableTypeInfo">ReadableTypeInfo</a>): <a href="type_info.md#0x1_type_info_TypeInfo">TypeInfo</a> {
+    <a href="type_info.md#0x1_type_info_TypeInfo">TypeInfo</a>{
+        account_address: readable_type_info.account_address,
+        module_name: *<a href="../../move-stdlib/doc/string.md#0x1_string_bytes">string::bytes</a>(&readable_type_info.module_name),
+        struct_name: *<a href="../../move-stdlib/doc/string.md#0x1_string_bytes">string::bytes</a>(&readable_type_info.struct_name)
     }
 }
 </code></pre>
--- a/aptos-move/framework/aptos-stdlib/doc/type_info.md
+++ b/aptos-move/framework/aptos-stdlib/doc/type_info.md
@@ -6,10 +6,12 @@
 
 
 -  [Struct `TypeInfo`](#0x1_type_info_TypeInfo)
+-  [Struct `ReadableTypeInfo`](#0x1_type_info_ReadableTypeInfo)
 -  [Constants](#@Constants_0)
 -  [Function `account_address`](#0x1_type_info_account_address)
 -  [Function `module_name`](#0x1_type_info_module_name)
 -  [Function `struct_name`](#0x1_type_info_struct_name)
+-  [Function `to_readable_type_info`](#0x1_type_info_to_readable_type_info)
 -  [Function `chain_id`](#0x1_type_info_chain_id)
 -  [Function `type_of`](#0x1_type_info_type_of)
 -  [Function `type_name`](#0x1_type_info_type_name)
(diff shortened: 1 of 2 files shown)


movekevin (approved):

movekevin (changes requested):
Actually thinking about this a bit more, can't you just use type_name() to generate a human readable string such as "0x123::module::struct" and return that in view functions?

alnoki:
> Actually thinking about this a bit more, can't you just use type_name() to generate a human readable string such as "0x123::module::struct" and return that in view functions?

@movekevin That function can generate human readable strings, but there is no way to convert back and forth between the resulting string and a `TypeInfo` without a new native function.

This PR adds comparable functionality in a pure Move implementation, which supports legacy applications that are already using `TypeInfo`.

Your thoughts?

movekevin:
> > Actually thinking about this a bit more, can't you just use type_name() to generate a human readable string such as "0x123::module::struct" and return that in view functions?
> 
> @movekevin That function can generate human readable strings, but there is no way to convert back and forth between the resulting string and a `TypeInfo` without a new native function.
> 
> This PR adds comparable functionality in a pure Move implementation, which supports legacy applications that are already using `TypeInfo`.
> 
> Your thoughts?

Why do you need to convert back from a string to TypeInfo? Also it doesn't look like the newly added ReadableTypeInfo does that either

alnoki:
> > > Actually thinking about this a bit more, can't you just use type_name() to generate a human readable string such as "0x123::module::struct" and return that in view functions?
> > 
> > 
> > @movekevin That function can generate human readable strings, but there is no way to convert back and forth between the resulting string and a `TypeInfo` without a new native function.
> > This PR adds comparable functionality in a pure Move implementation, which supports legacy applications that are already using `TypeInfo`.
> > Your thoughts?
> 
> Why do you need to convert back from a string to TypeInfo? Also it doesn't look like the newly added ReadableTypeInfo does that either

The back-conversion added in `4de358a` is intended so that the contents of a `ReadableTypeInfo` can be verified: if someone wants to return a `ReadableTypeInfo` in a view function, they'll probably want to assert the fields in a unit test, and converting back to a `TypeInfo` allows them to verify via the getters.

davidiw (approved):
I think it is a bit excessive, but there's nothing fundamentally wrong :P

movekevin:
> > > > Actually thinking about this a bit more, can't you just use type_name() to generate a human readable string such as "0x123::module::struct" and return that in view functions?
> > > 
> > > 
> > > @movekevin That function can generate human readable strings, but there is no way to convert back and forth between the resulting string and a `TypeInfo` without a new native function.
> > > This PR adds comparable functionality in a pure Move implementation, which supports legacy applications that are already using `TypeInfo`.
> > > Your thoughts?
> > 
> > 
> > Why do you need to convert back from a string to TypeInfo? Also it doesn't look like the newly added ReadableTypeInfo does that either
> 
> The back-conversion added in [4de358a](https://github.com/aptos-labs/aptos-core/commit/4de358a7fe73a838c09f768f92cdd433a17c1d07) is to allow the contents of a `ReadableTypeInfo` to be verified: if someone wants to return a `ReadableTypeInfo` in a view function, they'll probably want to assert the fields in a unit test, and converting back to a `TypeInfo` allows them to verify via the getters.
> 
> As for returning a `ReadableTypeInfo` rather than a string, this paradigm allows for existing applications that use `TypeInfo` to provide readable outputs, since there is no way to convert from a `TypeInfo` to a string without a new native function.

I'm still not quite sure on the use case here. If this is meant to make testing easier, we should introduce test-only changes. What would be a use case for obtaining ReadableTypeInfo from a view function then converting back to TypeInfo?

alnoki:
> > > > > Actually thinking about this a bit more, can't you just use type_name() to generate a human readable string such as "0x123::module::struct" and return that in view functions?
> > > > 
> > > > 
> > > > @movekevin That function can generate human readable strings, but there is no way to convert back and forth between the resulting string and a `TypeInfo` without a new native function.
> > > > This PR adds comparable functionality in a pure Move implementation, which supports legacy applications that are already using `TypeInfo`.
> > > > Your thoughts?
> > > 
> > > 
> > > Why do you need to convert back from a string to TypeInfo? Also it doesn't look like the newly added ReadableTypeInfo does that either
> > 
> > 
> > The back-conversion added in [4de358a](https://github.com/aptos-labs/aptos-core/commit/4de358a7fe73a838c09f768f92cdd433a17c1d07) is to allow the contents of a `ReadableTypeInfo` to be verified: if someone wants to return a `ReadableTypeInfo` in a view function, they'll probably want to assert the fields in a unit test, and converting back to a `TypeInfo` allows them to verify via the getters.
> > As for returning a `ReadableTypeInfo` rather than a string, this paradigm allows for existing applications that use `TypeInfo` to provide readable outputs, since there is no way to convert from a `TypeInfo` to a string without a new native function.
> 
> I'm still not quite sure on the use case here. If this is meant to make testing easier, we should introduce test-only changes. What would be a use case for obtaining ReadableTypeInfo from a view function then converting back to TypeInfo?

@movekevin the use case would be:

1. Existing protocol uses `TypeInfo` internally
2. New view function is designed to return `ReadableTypeInfo`
3. To verify that the view function returns as expected, a unit test is designed to verify that the returned `ReadableTypeInfo` matches the internally-tracked `TypeInfo`

davidiw:
I'm somewhere in the middle here. I don't really follow the argument of testing view functions. How did the type parameter get there in the first place -- isn't that as part of the input to the view function? Your proposed test seems to be more on testing the ReadableTypeInfo -> TypeInfo conversion.

alnoki:
> I'm somewhere in the middle here. I don't really follow the argument of testing view functions. How did the type parameter get there in the first place -- isn't that as part of the input to the view function? Your proposed test seems to be more on testing the ReadableTypeInfo -> TypeInfo conversion.

@davidiw indeed the proposed unit test is to verify the conversion from `ReadableTypeInfo` to `TypeInfo`, and establish a one-to-one correspondence.

Based on this assumption, a straightforward use case is to cast `TypeInfo` into a `ReadableTypeInfo` for a view function in a dApp. When unit testing the dApp, a dev casts the returned `ReadableTypeInfo` back to `TypeInfo` and verifies it against a type. e.g. `assert!(to_type_info(&readable_view_function_return) == type_of<ExpectedType>() == type_info_from_internal_registry, 0);`

movekevin:
> > I'm somewhere in the middle here. I don't really follow the argument of testing view functions. How did the type parameter get there in the first place -- isn't that as part of the input to the view function? Your proposed test seems to be more on testing the ReadableTypeInfo -> TypeInfo conversion.
> 
> @davidiw indeed the proposed unit test is to verify the conversion from `ReadableTypeInfo` to `TypeInfo`, and establish a one-to-one correspondence.
> 
> Based on this assumption, a straightforward use case is to cast `TypeInfo` into a `ReadableTypeInfo` for a view function in a dApp. When unit testing the dApp, a dev casts the returned `ReadableTypeInfo` back to `TypeInfo` and verifies it against a type. e.g. `assert!(to_type_info(&readable_view_function_return) == type_of<ExpectedType>() == type_info_from_internal_registry, 0);`

While I really appreciate brainstorming on making the framework better, i’d not introduce changes for testonly use cases into the framework. Unless there are real use cases here, I unfortunately can’t approve to let this change move forward

alnoki:
@movekevin here is a real use case where I would have used the functionality in this PR, but since it wasn't accepted, I had to implement it myself:

`ReadableTypeInfo` analog:

https://github.com/econia-labs/econia/blob/1a5f7b8c3127c9801325334298d0ede406d5275e/src/move/econia/sources/registry.move#L228-L237

Nested in another struct for a view function return:

https://github.com/econia-labs/econia/blob/1a5f7b8c3127c9801325334298d0ede406d5275e/src/move/econia/sources/registry.move#L293-L313

View function:

https://github.com/econia-labs/econia/blob/1a5f7b8c3127c9801325334298d0ede406d5275e/src/move/econia/sources/registry.move#L578-L634

Unit test:

https://github.com/econia-labs/econia/blob/1a5f7b8c3127c9801325334298d0ede406d5275e/src/move/econia/sources/registry.move#L2427-L2602

movekevin:
> @movekevin here is a real use case where I would have used the functionality in this PR, but since it wasn't accepted, I had to implement it myself:
> 
> `ReadableTypeInfo` analog:
> 
> https://github.com/econia-labs/econia/blob/1a5f7b8c3127c9801325334298d0ede406d5275e/src/move/econia/sources/registry.move#L228-L237
> 
> Nested in another struct for a view function return:
> 
> https://github.com/econia-labs/econia/blob/1a5f7b8c3127c9801325334298d0ede406d5275e/src/move/econia/sources/registry.move#L293-L313
> 
> View function:
> 
> https://github.com/econia-labs/econia/blob/1a5f7b8c3127c9801325334298d0ede406d5275e/src/move/econia/sources/registry.move#L578-L634
> 
> Unit test:
> 
> https://github.com/econia-labs/econia/blob/1a5f7b8c3127c9801325334298d0ede406d5275e/src/move/econia/sources/registry.move#L2427-L2602

You can store TypeInfo directly in your example there as well. Then in the view function, you can convert to a string and return. Or you can even return TypeInfo directly in the view function too (that's allowed). Then in your test you can do the comparison without an intermediate type.

alnoki:
> Or you can even return TypeInfo directly in the view function too (that's allowed)

@movekevin returning `TypeInfo` in a view function isn't practical because it renders on a client as hex-encoded ASCII, hence this PR

> You can store TypeInfo directly in your example there as well. Then in the view function, you can convert to a string and return.

This is what's going on, but packing strings into a nested struct for organization. The purpose of this PR is to provide helpers for the nested struct so that other developers don't have to re-implement what I already did

PR #8686, opened 2023-06-15
[move-example] tic tac toe contract

wlawt:
### Description
An example of tic tac toe written in Move. 

### Test Plan
- All entry functions
- Valid winning states
- Invalid moves (i.e., taking a player twice, illegal moves, etc.)

### Example

Contract example is deployed to [devnet](https://explorer.aptoslabs.com/account/0x2b01c4d0b3d97ca229d8da5bdc675d0a8ce48655a7f68bfcecec49510e8ca1dd/modules/code/ttt?network=devnet)

### DIFF
--- a/aptos-move/move-examples/tic-tac-toe/Move.toml
+++ b/aptos-move/move-examples/tic-tac-toe/Move.toml
@@ -5,8 +5,6 @@ version = '1.0.0'
 [addresses]
 tic_tac_toe = "_"
 
-[dependencies.AptosFramework]
-git = 'https://github.com/aptos-labs/aptos-core.git'
-rev = 'main'
-subdir = 'aptos-move/framework/aptos-framework'
+[dependencies]
+AptosFramework = { local = "../../framework/aptos-framework" }
 
--- a/aptos-move/move-examples/tic-tac-toe/Move.toml
+++ b/aptos-move/move-examples/tic-tac-toe/Move.toml
@@ -1,0 +1,12 @@
+[package]
+name = 'tic-tac-toe'
+version = '1.0.0'
+
+[addresses]
+tic_tac_toe = "_"
+
+[dependencies.AptosFramework]
+git = 'https://github.com/aptos-labs/aptos-core.git'
+rev = 'main'
+subdir = 'aptos-move/framework/aptos-framework'
+
--- a/aptos-move/move-examples/tic-tac-toe/sources/TicTacToe.move
+++ b/aptos-move/move-examples/tic-tac-toe/sources/TicTacToe.move
@@ -66,6 +66,8 @@ module tic_tac_toe::ttt {
     }
         
     public entry fun choose_move(player: &signer, game_addr: address, x: u64, y: u64) acquires Game {
+        assert!(x < 3, error::invalid_argument(ERR_ILLEGAL_ACTION));
+        assert!(y < 3, error::invalid_argument(ERR_ILLEGAL_ACTION));
         let game: &mut Game = borrow_global_mut(game_addr);
         let player_x = option::borrow_mut(&mut game.player_x);
         let player_o = option::borrow_mut(&mut game.player_o);
@@ -306,6 +308,23 @@ module tic_tac_toe::ttt {
         choose_move(player_x, game_addr, 0, 1);
         choose_move(player_o, game_addr, 2, 1);
         choose_move(player_x, game_addr, 0, 2);
+
+        let game = borrow_global(game_addr);
+        assert!(check_is_game_over(game), error::invalid_state(ERR_GAME_NOT_DONE));
+        cleanup(player_x);
+    }
+
+    #[test(player_x = @0x123, player_o = @0x223, game_addr = @0x123)]
+    #[expected_failure]
+    fun test_out_of_bound_move(
+        player_x: &signer, 
+        player_o: &signer,
+        game_addr: address,
+    ) acquires Game {
+        create_account_for_test(signer::address_of(player_x));
+        start_game(player_x);
+        join_as_player_o(player_o, game_addr);
+        choose_move(player_x, game_addr, 3, 4);
 
         let game = borrow_global(game_addr);
         assert!(check_is_game_over(game), error::invalid_state(ERR_GAME_NOT_DONE));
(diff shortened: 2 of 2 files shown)


aptos-move/move-examples/tic-tac-toe/sources/TicTacToe.move
```diff
@@ -0,0 +1,639 @@
...
+    }
+
+    struct Game has key, store {
+        board: Board,
+        player_x: Option<Player>,
+        player_o: Option<Player>,
+        is_player_x_turn: bool,
+        is_game_over: bool,
+        game_over_events: event::EventHandle<GameOverEvent>,
+    }
```
gregnazario, line 51: I wonder can you make this droppable and then you don't need to have the cleanup method?

aptos-move/move-examples/tic-tac-toe/sources/TicTacToe.move
```diff
@@ -0,0 +1,639 @@
...
+
+    //// PLAYER CONSTANTS
+    const PLAYER_X_TYPE: u64 = 1;
+    const PLAYER_O_TYPE: u64 = 2;
+
+    //// ERROR CODES
+    const ERR_ILLEGAL_ACTION: u64 = 0;
+    const ERR_INVALID_MOVE: u64 = 1;
+    const ERR_PLAYER_TAKEN: u64 = 2;
+    const ERR_GAME_NOT_DONE: u64 = 3;
```
gregnazario, line 27: FYI, the standard for error codes are pretty simple, and add useful messages to errors in the explorer and elsewhere such as:

```
/// Error message that will show when it fails in the transaction
ENAME_OF_ERROR: u64 = 1;
```

gregnazario (approved):
Looks pretty good as far as I can tell.  Could use some comments describing it a bit, but other than that it looks good.

aptos-move/move-examples/tic-tac-toe/sources/TicTacToe.move
wlawt: I had issues with `Game` being droppable because supposedly `event::EventHandle<GameOverEvent` wasn't droppable. I also explored the option of creating a separate struct to maintain `GameEvents` but I figured it was easier to keep it within Game and destroy the object at the end/forfeit

gregnazario: Ah yes, I forgot about that I had similar issues.  I think the events even if they're droppable the handle isn't

aptos-move/move-examples/tic-tac-toe/Move.toml
```diff
@@ -0,0 +1,12 @@
...
+name = 'tic-tac-toe'
+version = '1.0.0'
+
+[addresses]
+tic_tac_toe = "_"
+
+[dependencies.AptosFramework]
+git = 'https://github.com/aptos-labs/aptos-core.git'
+rev = 'main'
+subdir = 'aptos-move/framework/aptos-framework'
```
vgao1996, line 11: Since all other example projects use local dependencies, it might be better we also do it that way.

aptos-move/move-examples/tic-tac-toe/sources/TicTacToe.move
```diff
@@ -0,0 +1,687 @@
...
+        col: u64,
+    }
+
+    struct Game has key, store {
+        board: Board,
+        player_x: Option<Player>,
+        player_o: Option<Player>,
+        is_player_x_turn: bool,
+        is_game_over: bool,
+        game_over_events: event::EventHandle<GameOverEvent>,
```
vgao1996, line 61: @movekevin does this make sense to you? It doesn't feel quite right to me to store the event handle in a struct that will be destroyed after the game has finished.

vgao1996 (approved):
Let's land this now but we should revisit the event handle part. cc @movekevin

aptos-move/move-examples/tic-tac-toe/sources/TicTacToe.move
```diff
@@ -0,0 +1,696 @@
...
+    use std::vector;
+    use std::option::{Self, Option};
+    use std::signer;
+    use std::error;
+    use aptos_framework::account;
+    use aptos_framework::event;
+
+    #[test_only]
+    use aptos_framework::account::create_account_for_test;
+
```
movekevin, line 11: nit: Remove unnecessary empty lines

aptos-move/move-examples/tic-tac-toe/sources/TicTacToe.move
```diff
@@ -0,0 +1,696 @@
+module tic_tac_toe::ttt {
+    use std::vector;
+    use std::option::{Self, Option};
+    use std::signer;
+    use std::error;
+    use aptos_framework::account;
+    use aptos_framework::event;
+
+    #[test_only]
+    use aptos_framework::account::create_account_for_test;
```
movekevin, line 10: Move this closer to the test section at the end

aptos-move/move-examples/tic-tac-toe/sources/TicTacToe.move
```diff
@@ -0,0 +1,696 @@
+module tic_tac_toe::ttt {
+    use std::vector;
```
movekevin, line 2: nit: Order these alphabetically please

aptos-move/move-examples/tic-tac-toe/sources/TicTacToe.move
movekevin: Yeah if the intention here is to allow accessing game records after games are over, this event handle needs to be stored in a permanent struct

aptos-move/move-examples/tic-tac-toe/sources/TicTacToe.move
```diff
@@ -0,0 +1,696 @@
...
+     * @notice initializes a valid, playable Game
+     * @dev stores the Game into global storage
+     */
+    public entry fun start_game(creator: &signer) {
+        // check game doesn't already exist under creator address
+        assert!(!exists<Game>(signer::address_of(creator)), error::already_exists(EGAME_ALREADY_EXISTS_FOR_USER));
+        let game = initalize_game(creator);
+        let creator_addr = signer::address_of(creator);
+        choose_player_x(&mut game, creator_addr);
+        move_to<Game>(creator, game);
```
movekevin, line 80: This is an example so probably not a big deal, but I'd not structure the game data as a temporary state inside the creator's account. It's better practice to create the game as an object (with its own address), then store it a pointer to the active game in the creator account if I really want to make everything discoverable on chain. Otherwise, such a pointer is not needed if the UI component can rely on indexers to find active games for a player

PR #8688, opened 2023-06-15
[dag] remove support for weak links for now

zekun000:

### DIFF
--- a/consensus/src/dag/dag_store.rs
+++ b/consensus/src/dag/dag_store.rs
@@ -11,7 +11,7 @@ use aptos_crypto_derive::CryptoHasher;
 use aptos_types::{aggregate_signature::AggregateSignature, validator_verifier::ValidatorVerifier};
 use serde::{Deserialize, Serialize};
 use std::{
-    collections::{BTreeMap, HashMap, HashSet},
+    collections::{BTreeMap, HashMap},
     ops::Deref,
     sync::Arc,
 };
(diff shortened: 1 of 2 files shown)


consensus/src/dag/dag_store.rs
```diff
@@ -226,46 +215,23 @@ impl Dag {
...
-                            Some(&node.metadata.author)
-                        } else {
-                            None
-                        }
-                    })
-                });
+        let all_nodes_in_round = self.nodes_by_round.get(&round)?.iter().flatten();
         if validator_verifier
-            .check_voting_power(strong_link_authors)
+            .check_voting_power(all_nodes_in_round.clone().map(|node| &node.metadata.author))
```
ibalajiarun, line 225: nit: can we clone the vec<author> instead? Not sure which is better.

consensus/src/dag/dag_store.rs
```diff
@@ -226,46 +215,23 @@ impl Dag {
         self.nodes_by_digest.get(digest).cloned()
     }
 
-    pub fn get_unlinked_nodes_for_new_round(
+    pub fn get_strong_links_for_round(
         &self,
+        round: Round,
         validator_verifier: &ValidatorVerifier,
     ) -> Option<Vec<NodeMetadata>> {
```
ibalajiarun, line 222: no strong opinion but I feel like Vec<NodeMetadata> should be a type called RoundParents.

ibalajiarun (approved):

consensus/src/dag/dag_store.rs
zekun000: it's a flat map type, not the vector, I was trying to save the iter().flattern() piece

sasha8 (approved):
LGTM

PR #8692, opened 2023-06-15
[Storage] Remove get_latest_transaction_info_option API, and move all callers to use get_latest_version.

grao1991:
### Description

After we split ledger commit to multiple batches, the latest transaction info version might not be equal to the latest "committed" version.

### DIFF
--- a/state-sync/state-sync-v2/state-sync-driver/src/utils.rs
+++ b/state-sync/state-sync-v2/state-sync-driver/src/utils.rs
@@ -269,18 +269,9 @@ pub fn fetch_latest_synced_ledger_info(
 
 /// Fetches the latest synced version from the specified storage
 pub fn fetch_latest_synced_version(storage: Arc<dyn DbReader>) -> Result<Version, Error> {
-    let latest_transaction_info =
-        storage
-            .get_latest_transaction_info_option()
-            .map_err(|error| {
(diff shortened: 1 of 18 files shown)


msmouse (approved):

state-sync/state-sync-v2/state-sync-driver/src/utils.rs
```diff
@@ -269,18 +269,9 @@ pub fn fetch_latest_synced_ledger_info(
...
-            .map_err(|error| {
-                Error::StorageError(format!(
-                    "Failed to get the latest transaction info from storage: {:?}",
-                    error
-                ))
-            })?;
-    latest_transaction_info
-        .ok_or_else(|| Error::StorageError("Latest transaction info is missing!".into()))
-        .map(|(latest_synced_version, _)| latest_synced_version)
+    storage.get_latest_version().map_err(|e| {
```
zekun000, line 272: hmm, I think this is intended to fetch the latest txn info instead. cc: @JoshLind

JoshLind: As long as this returns the latest version from storage (i.e., the very latest version even if there has not been a ledger info written to storage for that version), I think it should be okay 😄

PR #8695, opened 2023-06-15
[Sharding][Execution] Integrate partitioning into the block-stm benchmark

sitalkedia:
### Description

Following changes are included 

1. Partitioning returns SubBlocks with `Transaction` instead of `AnalyzedTransaction` as we don't need analyzed transaction during execution for now.
2. Introduce a new type `BlockExecutorTransaction` that supports both unsharded and sharded transactions.
3. Integrate partitioning into the block-stm benchmark and also when sharding is not enabled, don't go through the sharded code path.


execution/executor/src/components/chunk_output.rs
```diff
@@ -89,13 +89,13 @@ impl ChunkOutput {
...
     pub fn by_transaction_execution_sharded<V: VMExecutor>(
-        transactions: Vec<Transaction>,
+        block: Vec<SubBlocksForShard<Transaction>>,
         state_view: CachedStateView,
         maybe_block_gas_limit: Option<u64>,
     ) -> Result<Self> {
         let state_view_arc = Arc::new(state_view);
         let transaction_outputs = Self::execute_block_sharded::<V>(
-            transactions.clone(),
+            block.clone(),
```
zjma, line 98: potential slow clone? I'm fine with it being here now.

sitalkedia: Hard to avoid the clone over here for now - this is similar in the unsharded path as well. This is something we should follow up to see if we can pass reference to the parallel execution instead.

msmouse (approved):

testsuite/sequential_execution_performance.py
```diff
@@ -4,7 +4,7 @@
 
 # Set the tps threshold for block size 1k, 10k and 50k
 BLOCK_SIZES = ["1k", "10k", "50k"]
-THRESHOLDS = {"1k": 3200, "10k": 4300, "50k": 7400}
+THRESHOLDS = {"1k": 3500, "10k": 5200, "50k": 12600}
```
sitalkedia, line 7: Please note that the squential test is significantly faster after I removed the sharded block-stm execution from its path. That is why we need to tune these thresholds as well as parallel execution speed up thresholds.

PR #8696, opened 2023-06-15
add keep alive to data service.

larry-aptos:
### Description

* For long-live connections, they're prune to network errors. In this change, we introduce `HTTP2 Ping` checks to make sure when network error happens, the client or the server can be aware of them.
  * Server can drop the connection and update the metrics accordingly, i.e., garbage collection in time. 
  * Client can crash and re-connect.
    * TODO: Gracefully handle the error
  * refs:
    * tonic client:  https://docs.rs/tonic/latest/tonic/transport/channel/struct.Endpoint.html#method.http2_keep_alive_interval
    * tonic server:  https://docs.rs/tonic/latest/tonic/transport/server/struct.Server.html#method.http2_keepalive_interval 

* now data service support both TLS and non-TLS connections; keep both of them to prevent downstream breakage. 

### Test Plan

Locally tested: connect with remote server and disconnect the wifi, client can detect the disconnection. 

image


bowenyang007:
Can you add some documentation in the readme about what each config means (comments in the config file might be enough), and specifically what keep alive means? User might not be able to find this PR.

larry-aptos:
> Can you add some documentation in the readme about what each config means (comments in the config file might be enough), and specifically what keep alive means? User might not be able to find this PR.

Done.

rustielin (approved):
Nice test plan !

ecosystem/indexer-grpc/indexer-grpc-data-service/README.md
```diff
@@ -15,19 +15,36 @@ Indexer GRPC data service fetches data from both cache and file store.
...
 server_config:
-    data_service_grpc_listen_address: 0.0.0.0:50052
     whitelisted_auth_tokens: 
       - "token1"
       - "token2"
     file_store_config:
       file_store_type: GcsFileStore
       gcs_file_store_bucket_name: indexer-grpc-file-store-bucketname
     data_service_grpc_tls_config:
+      data_service_grpc_listen_address: 0.0.0.0:50055
```
geekflyer, line 25: wouldn't it be better to use the same addresses in the example that we use internally?

ecosystem/indexer-grpc/indexer-grpc-parser/README.md
```diff
@@ -17,9 +17,21 @@ Indexer GRPC parser is to indexer data processor that leverages the indexer grpc
...
+        indexer_grpc_keep_alive_timeout_in_secs: 10
         auth_token: AUTH_TOKEN
     ```
 
+#### Config Explanation
+
+* `processor_name`: purpose of this processor; also used for monitoring purpose.
+* `postgres_connection_string`: PostgresQL DB connection string
+* `indexer_grpc_data_service_addresss`: Data service non-TLS endpoint address.
+* `indexer_grpc_keep_alive_interval_in_secs`: client-side HTTP2 KEEPALIVE check interval.
```
geekflyer, line 30: actually can you add a bit better description here? When I just googled HTTP2 KEEPALIVE I didn't really find any useful results. I think what this is really referring to is using HTTP2 PING frames for health checking, the argument just happened to be name `http2_keepalive` in tonic (btw they name it without an underscore).
What actually gave me a better understanding what this is, is looking at the tonics for this flag.
Anyways, I'd suggest updating the description, potentially renaming the flag and also adding a link to the flag docs in tonic.

geekflyer (approved):

ecosystem/indexer-grpc/indexer-grpc-data-service/README.md
```diff
@@ -15,19 +15,36 @@ Indexer GRPC data service fetches data from both cache and file store.
...
+      data_service_grpc_listen_address: 0.0.0.0:50057
     redis_read_replica_address: 127.0.0.1:6379
 ```
 
+### Config Explanation
+
+* `whitelisted_auth_tokens`: list of tokens that are allowed to consume the endpoint.
+* `file_store_config`: GCS to archive the transactions
+* `data_service_grpc_tls_config`: TLS endpoint exposed
+* `data_service_grpc_non_tls_config`: Non-TLS endpoint exposed
```
bowenyang007, line 38: add some comments on what the difference is and why we have both?

bowenyang007 (changes requested):
Let's address the comments for clearer documentation before shipping. Thanks!

ecosystem/indexer-grpc/indexer-grpc-parser/README.md
larry-aptos: good catch; i wasn't aware of their naming is inconsistent 🤦 .

done

ecosystem/indexer-grpc/indexer-grpc-data-service/README.md
larry-aptos: done.

ecosystem/indexer-grpc/indexer-grpc-data-service/README.md
larry-aptos: updated; the plan is to use tls at 50052 and non-tls 50051.

bowenyang007 (approved):

PR #8697, opened 2023-06-15
Improve txn-emitter stability and visibility on Forge

igor-aptos:
- visibility - add few more log lines, and print txn-emitter stats (every 1m) as well
- make txn-emitter work more precisely. With large number of accounts/submission_workes (and graceful test has 180000 of them), there were a few issues. First creating workers is slow, and so they start with some divergence to begin with - fixing by creating them first before starting, and passing the same "start" instant to all. second divergence wasn't visible because it logged only after 240s divergence (changed to log on 5s divergence)
- fix db flag that got changed recently for test tuned for throughput

### DIFF
--- a/crates/transaction-emitter-lib/src/emitter/mod.rs
+++ b/crates/transaction-emitter-lib/src/emitter/mod.rs
@@ -493,8 +493,40 @@ impl EmitJob {
         self.stats.accumulate(&self.phase_starts)
     }
 
-    pub fn accumulate(&self) -> Vec<TxnStats> {
+    pub fn peek_and_accumulate(&self) -> Vec<TxnStats> {
         self.stats.accumulate(&self.phase_starts)
+    }
+
+    pub async fn stop_job(self) -> Vec<TxnStats> {
+        self.stop_and_accumulate().await
+    }
+
+    pub async fn periodic_stat(&self, duration: Duration, interval_secs: u64) {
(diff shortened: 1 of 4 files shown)


crates/transaction-emitter-lib/src/emitter/mod.rs
```diff
@@ -615,17 +647,23 @@ impl TxnEmitter {
         );
 
         let all_start_sleep_durations = mode_params.get_all_start_sleep_durations(self.from_rng());
-        let mut all_accounts_iter = all_accounts.into_iter();
-        let mut workers = vec![];
+
+        // Creating workers is slow with many workers (TODO check why)
+        // so we create them all first, before starting them - so they start at the right time for
+        // traffic pattern to be correct.
```
igor-aptos, line 653: it takes 5 minutes to go through the loop below, in the graceful overload case (https://aptoslabs.grafana.net/goto/GaWX1nu4g?orgId=1) with 180k workers to create, which makes no sense - but at least with this change it doesn't affect the measurements.

bchocho (approved):
> fix db flag that got changed recently for test tuned for throughput
what was this change?

ibalajiarun (approved):

rustielin (approved):

perryjrandall (approved):

PR #8700, opened 2023-06-16
[Rust SDK] Add HardwareWalletAccount for ledger integration

0xjinn:
### Description

Add HardwareWalletAccount for ledger integration
1. mimic LocalAccount, but replace AccountKey with Derivative path
2. Added new trait for doing txn signing - We should have LocalAccount to implement this trait as well (Follow up item)
3. Add SDK to Ledger integration.

Next Step:
Will take HardwareWalletAccount into CLI, whenever we creating a LocalAccount, we need to check whether its a LocalAccount with private key, or Ledger HardwareWalletAccount

NOTE:
This will be merging into feature branch `[feature_branch/cli_ledger_integration](https://github.com/aptos-labs/aptos-core/tree/feature_branch/cli_ledger_integration)`, which will contains all ledger integration breaking changes

### DIFF
--- a/sdk/src/types.rs
+++ b/sdk/src/types.rs
@@ -180,7 +180,9 @@ pub trait TransactionSigner {
     fn sign_with_transaction_builder(&mut self, builder: TransactionBuilder) -> SignedTransaction;
 }
 
-/// Similar
+/// Similar to LocalAccount, but for hardware wallets.
+/// HardwareWallet does not have private key exported.
+/// Anything that requires private key should be go through HardwareWallet.
 #[derive(Debug)]
 pub struct HardwareWalletAccount {
     address: AccountAddress,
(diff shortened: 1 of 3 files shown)


sdk/src/types.rs
```diff
@@ -165,6 +168,127 @@ impl LocalAccount {
     }
 }
 
+/// Types of hardware wallet the SDK currently supports
+#[derive(Debug)]
+pub enum HardwareWalletType {
```
0xjinn, line 173: In case we want to support additional hardware wallet in the future. e.g Trevor

sdk/src/types.rs
```diff
@@ -165,6 +168,127 @@ impl LocalAccount {
...
+    Ledger,
+}
+
+pub trait TransactionSigner {
```
0xjinn, line 177: Follow up items - Update/refactor LocalAccount to impl this trait as well

issue #8719

gregnazario (approved):
It looks good to me.  Keeping in mind that Local account will likely need to be different than the hardware account

PR #8702, opened 2023-06-16
[compiler v2] Setup basic driver and test environment

wrwg:
This creates a new crate `move-compiler-v2`, containing a few basic components:

- A model of options and experiments for the compiler. This has been adapted from the Move-to-Yul compiler from last year.
- A test driver which supports experiments, with a first smoke test to ensure basic functionality of the setup.
- Top level entry function to call the compiler. It invokes the  new compilation mode of the move-model, which types check and constructs an AST. After that is unimplemented.

The PR also adds an additional entry point for the move-model to support the above.


rahxephon89 (approved):

third_party/move/move-model/src/lib.rs
```diff
@@ -61,10 +61,49 @@ pub mod ty;
...
+        },
+        Flags::model_compilation(),
+    )
+}
+
+// =================================================================================================
+// Entry Point V1
+
+/// Build the move model with default compilation flags and default options. This calls
+/// the move compiler v1 to compiler to bytecode first and attach the generated bytecode to
```
georgemitenkov, line 105: nit: compiler --> compile

georgemitenkov (approved):

third_party/move/move-compiler-v2/src/lib.rs
```diff
@@ -0,0 +1,63 @@
+// Copyright © Aptos Foundation
+// Parts of the project are originally copyright © Meta Platforms, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+mod experiments;
```
brmataptos, line 5: Older packages with a similar file had
`#![forbid(unsafe_code)]`
here.  Did you want to have unsafe code for some reason?

third_party/move/move-compiler-v2/src/lib.rs
```diff
@@ -0,0 +1,63 @@
+// Copyright © Aptos Foundation
+// Parts of the project are originally copyright © Meta Platforms, Inc.
```
brmataptos, line 2: From my reading of https://spdx.org/licenses/Apache-2.0.html, I think you need to retain the original notice on this file verbatim:

// Copyright (c) The Diem Core Contributors
// Copyright (c) The Move Contributors

and not change it to "Meta Platforms".  Adding a new Copyright in addition, as you have done, is fine, but seems more appropriate to be Aptos Labs than Aptos Foundation.

third_party/move/move-compiler-v2/src/options.rs
```diff
@@ -0,0 +1,70 @@
+// Copyright © Aptos Foundation
+// Parts of the project are originally copyright © Meta Platforms, Inc.
```
brmataptos, line 2: See commenn on previous file.  This should probably say:

// Copyright (c) Aptos Labs, Inc.
// Copyright (c) The Diem Core Contributors
// Copyright (c) The Move Contributors

No need to reference "Parts of the project" or Meta, as far as I can tell, since this is derived from move-to-yul/src/options.rs.

third_party/move/move-model/src/lib.rs
```diff
@@ -61,10 +61,49 @@ pub mod ty;
...
+#[derive(Debug, Clone)]
+pub struct PackageInfo {
+    pub sources: Vec<String>,
+    pub address_map: BTreeMap<String, NumericalAddress>,
+}
 
-/// Build the move model with default compilation flags and default options and no named addresses.
-/// This collects transitive dependencies for move sources from the provided directory list.
+/// Builds the Move model for the v2 compiler. This builds the model, compiling both code
+/// and specs from sources into the typed check AST. No bytecode is attached to the model.
```
brmataptos, line 75: "typed check AST" doesn't parse for me.  Is there a "check AST" which has been typed?  Or should it be a typechecked AST?

third_party/move/move-model/src/lib.rs
brmataptos: This is not added code but rather displaced by the added "Entry Point V2" code, so perhaps it should be left alone.  But if Wolfgang wants to correct it, then also fix "Build the move model" --> "Build**s** the move model..."

third_party/move/move-model/src/lib.rs
```diff
@@ -75,9 +114,7 @@ pub fn run_model_builder<
     run_model_builder_with_options(move_sources, deps, ModelBuilderOptions::default())
 }
 
-/// Build the move model with default compilation flags and custom options and a set of provided
-/// named addreses.
-/// This collects transitive dependencies for move sources from the provided directory list.
+/// Build the move model with default compilation flags and custom options.
```
brmataptos, line 117: "Builds the move model"

third_party/move/move-compiler-v2/tests/testsuite.rs
```diff
@@ -0,0 +1,63 @@
+// Copyright © Aptos Foundation
+// Parts of the project are originally copyright © Meta Platforms, Inc.
```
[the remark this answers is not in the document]
wrwg: I tend to disagree because this is just a generic driver which is used in this or another form in multiple places of code based on the Move model. This is boilerplate to get things running. The same applies for the testsuite. This is not the same or copied code. Its just a starting point. In the next PR, it will look already rather different. Stay tuned.

third_party/move/move-compiler-v2/src/lib.rs
wrwg: We don't do this anymore.

third_party/move/move-model/src/lib.rs
wrwg: You see the need for this when you look at the entire file. Without this, its rather difficult to navigate. The Diff alone does not show you this.

brmataptos (commented):
A few nits.

PR #8703, opened 2023-06-16
[dag] basic dag driver and types

zekun000:


consensus/src/dag/dag_driver.rs
```diff
@@ -0,0 +1,103 @@
...
+            dag_writer.add_node(node)?;
+            if self.current_round == round {
+                let maybe_strong_links = dag_writer
+                    .get_strong_links_for_round(self.current_round, &self.epoch_state.verifier);
+                drop(dag_writer);
+                if let Some(strong_links) = maybe_strong_links {
+                    self.enter_new_round(strong_links);
+                }
+            }
+        }
```
ibalajiarun, line 68: We need an else case here to return an `Err` if `all_exists` is false?

consensus/src/dag/types.rs
```diff
@@ -0,0 +1,282 @@
...
+
+    fn add(&mut self, peer: Author, _ack: Self::Ack) -> anyhow::Result<Option<Self::Aggregated>> {
+        self.received.insert(peer);
+        if self.received.len() == self.num_validators {
+            Ok(Some(()))
+        } else {
+            Ok(None)
+        }
+    }
+}
```
ibalajiarun, line 275: nit: move this impl BroadcastStatus above to consolidate same type implementations?

consensus/src/dag/reliable_broadcast.rs
```diff
@@ -1,25 +1,36 @@
...
 use aptos_consensus_types::common::Author;
 use async_trait::async_trait;
 use futures::{stream::FuturesUnordered, StreamExt};
+use serde::{de::DeserializeOwned, Serialize};
 use std::{future::Future, sync::Arc, time::Duration};
-use tokio::sync::oneshot;
 
-pub trait DAGMessage: Sized + Clone {
-    fn from_network_message(msg: ConsensusMsg) -> anyhow::Result<Self>;
+pub trait DAGMessage: Sized + Clone + Serialize + DeserializeOwned {
```
ibalajiarun, line 12: This trait should be in `types.rs` as well?

consensus/src/dag/dag_store.rs
```diff
@@ -208,64 +74,42 @@ impl Dag {
                 .is_none(),
```
ibalajiarun, line 74: existing code but, I think we should not replace here, otherwise even though we catch equivocation, by replacing, we might be accepting it?

consensus/src/dag/tests/reliable_broadcast_tests.rs
```diff
@@ -121,33 +107,71 @@ async fn test_reliable_broadcast() {
...
+                let ret = rb
+                    .broadcast::<TestBroadcastStatus>(message, aggregating)
+                    .await;
+                tx.send(ret)
+            }),
+        abort_registration,
+    );
+    tokio::spawn(fut);
+    abort_handle.abort();
     assert!(rx.await.is_err());
```
ibalajiarun, line 176: this might get flaky no? depending on how fast fut can complete, tx.send can happen before abort?

consensus/src/dag/reliable_broadcast.rs
```diff
@@ -70,29 +78,23 @@ impl ReliableBroadcast {
                     )
                 }
             };
+            let network_message = message.into_network_message();
```
ibalajiarun, line 81: Should we use DAGNetworkMessage throughout and convert between ConsensusMsg in dag_handler and dag_network_sender?

consensus/src/dag/dag_store.rs
```diff
@@ -208,64 +74,42 @@ impl Dag {
...
-            self.highest_unlinked_nodes_by_author[index].replace(node);
-        }
         Ok(())
     }
 
     pub fn exists(&self, digest: &HashValue) -> bool {
         self.nodes_by_digest.contains_key(digest)
     }
 
+    pub fn all_exists<'a>(&self, mut digests: impl Iterator<Item = &'a HashValue>) -> bool {
```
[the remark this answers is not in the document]
zekun000: iterator needs to be mut to iterate

consensus/src/dag/dag_driver.rs
zekun000: yeah we need to handle the fetch case, I'll add a todo

consensus/src/dag/tests/reliable_broadcast_tests.rs
zekun000: I can reorder the abort to before the spawn

consensus/src/dag/dag_store.rs
zekun000: equivocation should not be possible with reliable broadcast, that's why I put it as assertion in the beginning, but sure, I can ensure it doesn't exist first

ibalajiarun (approved):
LGTM

sasha8 (approved):

consensus/src/dag/dag_driver.rs
```diff
@@ -0,0 +1,104 @@
...
+            time_service,
+            rb_abort_handle: None,
+        }
+    }
+
+    pub fn add_node(&mut self, node: CertifiedNode) -> anyhow::Result<()> {
+        let mut dag_writer = self.dag.write();
+        let round = node.metadata().round();
+        if dag_writer.all_exists(node.parents().iter().map(|metadata| metadata.digest())) {
+            dag_writer.add_node(node)?;
```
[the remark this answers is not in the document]
zekun000: we discussed this before, for self-contained validation we should check in message verification stage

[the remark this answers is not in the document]
sasha8: Sure, just making sure we do not forget these two things.

PR #8712, opened 2023-06-16
Add a new() function to smart_vector and simple_map to be more standardized across all data structures

movekevin:


aptos-move/framework/aptos-stdlib/sources/data_structures/smart_vector.move
```diff
@@ -27,6 +27,13 @@ module aptos_std::smart_vector {
 
     /// Regular Vector API
 
+    /// Create an empty vector using default logic to estimate `inline_capacity` and `bucket_size`, which may be
+    /// inaccurate.
+    /// This is exactly the same as empty() but is more standardized as all other data structures have new().
```
wrwg, line 32: If we do not have it, we should likely add an issue for attribute `#[deprecated]`?

wrwg (approved):

aptos-move/framework/aptos-stdlib/sources/data_structures/smart_vector.move
movekevin: Good idea! I added #[deprecated] to the other non-standard create function

lightmark (approved):

PR #8717, opened 2023-06-16
Update `econia` indexer with latest code (merge main)

elliottdehn:
### Description

Right now, we can't synchronize because we fall out of consensus with the rest of the nodes mid-synchronization. It's likely that this is because we lack needed code due to being out of date. We can merge `main` (after having synchronized it with upstream) to attempt at resolving this issue. This is equivalent to simply updating the node.

### Test Plan
Existing tests


elliottdehn:
Made by mistake, pay no mind.

Issue #8379, opened 2023-05-26
[Feature Request] Compile package in move with address injection


lightmark:
# 🚀 Feature Request

## Motivation

It is ideal to be able to dynamically assign @foo in move code when call `publish_package_txn(owner: &signer, metadata_serialized: vector<u8>, code: vector<vector<u8>>)` but the current compiled move code includes the static module address (foo = 0x123..) before calling this function. 

For example, the owner there may be an object generated from guid,  so its address is not known before calling this function and users may want to publish the code to the object address. There is no way to achieve that rn

## Pitch

**Describe the solution you'd like**
Add an address mapping feature when compiling code from move.

**Are you willing to open a pull request?** (See [CONTRIBUTING](../../CONTRIBUTING.md))
If I am very chill.

## Additional context

[internal slack thread](https://aptos-org.slack.com/archives/C036X27DZNG/p1684291155563789)

jolestar:
This is how we solved this problem during the previous development of daospace(https://github.com/starcoinorg/starcoin-dao-frontend/tree/create-dao-new):

1. `UserA` creates a Dao account `DaoB`(like Aptos resource account), and delegates the PublishModule capability to the `UserA` account.
2. The daospace front javascript generates a new Move module based on a template and uses the `DaoB` address to compile it(via MoveJs).
3. `UserA` deploys the modules bytecodes to `DaoB` address and returns the PublishModule capability to `DaoB`.

It would simplify the process if Aptos supported this feature.

PR #8726, opened 2023-06-18
[move] address remapping when publishing module package

lightmark:
### Description
issue https://github.com/aptos-labs/aptos-core/issues/8379
allow dynamically replace module address to another when publishing module. what I do

1. deserialize code, remap addresses in address_identifiers and constant pools. serialize it back.
2. for package metadata,
   - clear manifest, source code and source map because it is too cumbersome to remap all the addresses in those text files/source maps and they are optional
   - remap addresses in `deps`.

### Test Plan
add test case in e2e move tests.
replace `0xcafe` to `0xface`


aptos-move/framework/aptos-framework/sources/code.move
```diff
@@ -188,6 +192,27 @@ module aptos_framework::code {
...
 
+    /// Same as `publish_package_txn` but can dynamically replace addresses in the package with new ones.
+    /// The `address_mapping` must have even number of addresses, such as [A, B, C, D], and A and C in original package
+    /// would be replaced by B and D, respectively. If the same address is replaced more than once, the last mapping
+    /// would overwrite the previous ones.
+    public entry fun publish_package_with_address_remapping_txn(
+        owner: &signer,
+        metadata_serialized: vector<u8>,
+        code: vector<vector<u8>>,
+        address_mapping: vector<address>,
```
banool, line 203: This feels sort of error prone, can you do a vector of vectors instead? So each inner vector must have 2 elements, the current address and the address to remap to.

banool (commented):
How come we need this btw? If a module uses named addresses you can map them to whatever address we want.

lightmark:
@banool  The context is this issue #8379

gerben-stavenga:
Do you have some more info about what this does? Is there a feature request somewhere I can look at?

davidiw (commented):
Really cool,

Sometimes I wish we started these out as discussions in AIPs, the reason being is I can see at least 2-3 solutions here:
* Remapping during on-chain publishing as done here
* Remapping API
* Remapping via the CLI

Shall we get an AIP up for this regardless?

Also what's the cost for doing this? Is it worth it from the modular publishers perspective?

In this solution, are there any implications on upgradeability? I imagine I can start with my code with address A -> B, and I can upgrade from A->B, but I can also upgrade from B? Do we have the upgrade code paths?

wrwg:
> Really cool,
> 
> Sometimes I wish we started these out as discussions in AIPs, the reason being is I can see at least 2-3 solutions here:
> 
> * Remapping during on-chain publishing as done here
> * Remapping API
> * Remapping via the CLI
> 
> Shall we get an AIP up for this regardless?
> 
> Also what's the cost for doing this? Is it worth it from the modular publishers perspective?
> 
> In this solution, are there any implications on upgradeability? I imagine I can start with my code with address A -> B, and I can upgrade from A->B, but I can also upgrade from B? Do we have the upgrade code paths?

Hey @davidiw there were long discussions between @vgao1996 , @lightmark and me about the current design (you were out IIRC). Lets please have a chat in person. And yes indeed, eventually this should be an AIP, and this is prototyping for it.

lightmark:
@davidiw All your three solutions have been discussed between @vgao1996 and I further after talked on slack with @wrwg . Our current plan to put this on hold until reaching an agreement with SDK on how to do this offchain vs onchain and how to make the logic consistent, which may involve the same rust code with even FFI bindings for js sdk. 

I will close this PR for now if it bothers you as open. Later @vgao1996 and I will revisit this once he has a better plan for aptos hardhat as this would be an integral part of it. We can continue to discuss at a bigger scope in the notion design doc.

PR #8727, opened 2023-06-18
Add `to_standard_string` function to Address structs that is compliant with AIP 40

banool:
### Stack
- Previous in stack: N/A
- Next in stack: https://github.com/aptos-labs/aptos-core/pull/9186

### Description

This PR adds a `to_standard_string()` function to the various Address structs. This function outputs the account address in a way that is compliant with the proposed standard: https://github.com/aptos-foundation/AIPs/pull/175. 

Currently the purpose of this PR is to complement the AIP. Once the AIP is accepted, I will publish this PR, making any changes if necessary based on the discussion in the AIP. 

Note that the AIP just standardizes the way we format addresses currently (v1), this is not a full identifier overhaul (v2).

In a later set of PRs I will do the following:
- In every instance where the existing Display impl is used (e.g. in `format` or with `to_string()` I will make it explicitly use the underlying function (so nothing should change).
- After that I will make the Display impl for all of these structs call `to_standard_string` internally. This way the default string formatting for addresses will be standard compliant.
- In all future uses we will use the new standards compliant Display impl.

**Update**: Instead, I will change the Display + Serialize impl and update places that shouldn't use the new formatting to use the old function that we used to use in Display.

This PR affects both `third_party` and non `third_party` code. I can split this PR into two if that's how we prefer to do it.

### Test Plan
```
cd third_party/move/move-core/types
cargo test
```

```
cd api/types
cargo test
```

I'm not totally confident with the binary / hex stuff in this PR, please take a close look at the tests to make sure they're legit.

### DIFF
--- a/third_party/move/move-core/types/src/account_address.rs
+++ b/third_party/move/move-core/types/src/account_address.rs
@@ -47,6 +47,46 @@ impl AccountAddress {
         Self(buf)
     }
 
+    /// Represent an account address in a way that is compliant with the v1 address
+    /// standard. The standard is defined as part of AIP-40, read more here:
+    /// https://github.com/aptos-foundation/AIPs/blob/main/aips/aip-40.md
+    ///
+    /// In short, all special addresses MUST be represented in SHORT form, e.g.
+    ///
+    /// 0x1
+    ///
+    /// All other addresses MUST be represented in LONG form, e.g.
+    ///
+    /// 0x002098630cfad4734812fa37dc18d9b8d59242feabe49259e26318d468a99584
+    ///
+    /// For an explanation of what defines a "special" address, see `is_special`.
+    ///
+    /// All string representations of addresses MUST be prefixed with 0x.
+    pub fn to_standard_string(&self) -> String {
+        let suffix = if self.is_special() {
+            self.short_str_lossless()
+        } else {
+            self.to_canonical_string()
+        };
+        format!("0x{}", suffix)
+    }
+
+    /// Returns whether the address is a "special" address. Addresses are considered
+    /// special if the first 63 characters of the hex string are zero. In other words,
+    /// an address is special if the first 31 bytes are zero and the last byte is
+    /// smaller than than `0b10000` (16). In other words, special is defined as an address
+    /// that matches the following regex: `^0x0{63}[0-9a-f]$`. In short form this means
+    /// the addresses in the range from `0x0` to `0xf` (inclusive) are special.
+    ///
+    /// For more details see the v1 address standard defined as part of AIP-40:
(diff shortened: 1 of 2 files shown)


davidiw (commented):
seems fine, but would like to see the complete (or at least more) work before bringing it in.

another note, your plan doesn't state that we should likely eliminate the legacy functions, is that the plan too? I understand based upon Greg's previous experience, we may face a long path that way and we may actually want to check in this work at some incremental stage?

banool:
This PR is complete, what additional work were you expecting to see?

As for deleting legacy functions, yeah maybe down the line, where that's at least a year out. But it may be more effort than it's worth.

third_party/move/move-core/types/src/account_address.rs
```diff
@@ -47,6 +47,46 @@ impl AccountAddress {
...
+    ///
+    /// All string representations of addresses MUST be prefixed with 0x.
+    pub fn to_standard_string(&self) -> String {
+        let suffix = if self.is_special() {
+            self.short_str_lossless()
+        } else {
+            self.to_canonical_string()
+        };
+        format!("0x{}", suffix)
+    }
```
gregnazario, line 72: How will this interact with things like module addresses?

Currently, the compiler and other places will use the short string mode.  Some pieces are treated like strings and some are treated like addresses.

banool: Not 100% sure what you mean by this bit:

> Some pieces are treated like strings and some are treated like addresses.

Could you elaborate?

gregnazario: Function IDs, struct names, all are converted to strings, and used as a string output / input.

This is the full `0x00000...01::coin::Coin` -> `0x1::coin::Coin`

This applies everywhere, so if that doesn't match it could be particularly problematic

Similarly a ton of other places in the compiler sometimes output long, sometimes short

banool: Right indeed. So this PR is not breaking in any way, it just adds a function that complies with AIP 40. Everywhere that accepts addresses should be updated to use the new from_str equivalent in their language in the next PR before we consider using this widely. Though there are places where I know we can start using this today without problems (new stuff, some particular configs / CLI stuff, etc) which is why I want to get it in.

0xmaayan (commented):
LGTM, I see there are some open questions on this PR so will let one of the others the final stamp.
btw, what is the migration plan?

gregnazario (approved):

banool:
Migration plan is mostly discussed on the AIP. I need to do another pass though.

PR #8729, opened 2023-06-19
[Spec] update spec for staking_contract.move

UIZorrot:
### Description

We finished specification of staking contract without the vector loop
There are some works remaining.

### Test Plan

`aptos move prove -f staking_contract`


aptos-move/framework/aptos-framework/sources/staking_contract.move
```diff
@@ -658,6 +658,11 @@ module aptos_framework::staking_contract {
         let (active, _, pending_active, _) = stake::get_stake(staking_contract.pool_address);
         let total_active_stake = active + pending_active;
         let accumulated_rewards = total_active_stake - staking_contract.principal;
+        spec {
+            // To avoid timeout, we assume staking_contract.commission_percentage is in range (0..100).
+            // This assumption is reasonable becasue a percentage value should be restricted in this range.
+            assume staking_contract.commission_percentage >= 0 && staking_contract.commission_percentage <= 100;
```
junkil-park, line 664: I think we can prove it instead of assuming it because there is an assertion to ensure it.

UIZorrot: Oh, however, we tried to use invariant and require, but it caused a timeout, is there anyway to solve it without assume?

aptos-move/framework/aptos-framework/sources/staking_contract.spec.move
```diff
@@ -62,36 +75,42 @@ spec aptos_framework::staking_contract {
...
+
+        let amount = coins.value;
+        include Create_Staking_Contract_With_Coins_Abortsif { amount };
     }
 
     /// Account is not frozen and sufficient to withdraw.
     /// Staking_contract exists the stacker/operator pair.
     spec add_stake(staker: &signer, operator: address, amount: u64) {
-        // TODO: complex aborts conditions in 'stake::add_stake_with_cap'.
+        // TODO: Verification timeout
```
rahxephon89, line 89: Same here

rahxephon89 (changes requested):

aptos-move/framework/aptos-framework/sources/staking_contract.spec.move
```diff
@@ -216,6 +235,8 @@ spec aptos_framework::staking_contract {
 
     /// The StakePool exists under the pool_address of StakingContract.
     spec get_staking_contract_amounts_internal(staking_contract: &StakingContract): (u64, u64, u64) {
+        // TODO: Verification timeout
```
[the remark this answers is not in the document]
UIZorrot: yes, it could be verify by about 1000s, we will have a comment about the conditions

aptos-move/framework/aptos-framework/sources/staking_contract.spec.move
```diff
@@ -5,32 +5,42 @@ spec aptos_framework::staking_contract {
...
     spec commission_percentage(staker: address, operator: address): u64 {
-        include StakingContractExistsAbortsIf;
+        include ContractExistsAbortsIf;
     }
 
     /// Staking_contract exists the stacker/operator pair.
     spec staking_contract_amounts(staker: address, operator: address): (u64, u64, u64) {
-        // Not verified when verify_duration_estimate > vc_timeout
-        pragma verify_duration_estimate = 120; // TODO: set because of timeout (property proved).
+        // TODO: verfication timeout
```
rahxephon89, line 23: If the property is proved, please use the comment "// TODO: set because of timeout (property proved)."

aptos-move/framework/aptos-framework/sources/staking_contract.spec.move
```diff
@@ -304,6 +333,7 @@ spec aptos_framework::staking_contract {
         let accumulated_rewards = total_active_stake - staking_contract.principal;
         aborts_if !exists<stake::StakePool>(pool_address);
         aborts_if active + pending_active > MAX_U64;
+        // TODO: These function causes the timeout
```
rahxephon89, line 336: Thanks for marking this out! In this case, it is the non-linear arithmetic operations that causes long verification time.

aptos-move/framework/aptos-framework/sources/staking_contract.spec.move
```diff
@@ -329,4 +359,104 @@ spec aptos_framework::staking_contract {
...
+        aborts_if !exists<account::Account>(resource_addr) && len(bcs::to_bytes(resource_addr)) != 32;
+        aborts_if len(account::ZERO_AUTH_KEY) != 32;
+
+        // verify stake::initialize_stake_owner()
+        aborts_if exists<stake::ValidatorConfig>(resource_addr);
+        let allowed = global<stake::AllowedValidators>(@aptos_framework);
+        aborts_if exists<stake::AllowedValidators>(@aptos_framework) && !contains(allowed.accounts, resource_addr);
+        aborts_if exists<stake::StakePool>(resource_addr);
+        aborts_if exists<stake::OwnerCapability>(resource_addr);
+        aborts_if exists<account::Account>(resource_addr) && acc.guid_creation_num + 12 >= account::MAX_GUID_CREATION_NUM;
```
rahxephon89, line 460: Could you explain why 12 is used here?

rahxephon89 (changes requested):

aptos-move/framework/aptos-framework/sources/staking_contract.spec.move
UIZorrot: yeah, without assuming it should timeout, by the way, assume it in the range of 0-100 seemed not right, so it better stay as todo for now.

aptos-move/framework/aptos-framework/sources/staking_contract.spec.move
chan-bing: Because the create_stake_pool function calls the stake::initialize_stake_owner function, the stake::initialize_stake_owner function calls the stake::initialize_owner function, and 12 events' guids are created in the initialize_owner function.

rahxephon89: Could you add the comment to the spec to explain it?

chan-bing: Yes, i could.

rahxephon89 (approved):

junkil-park (approved):

movekevin (approved):

PR #8731, opened 2023-06-19
Update Docker images

gedigi:
Update Docker images


rustielin (approved):

ibalajiarun (commented):
Does the rust image also need update? It also uses bullseye base image.

gedigi:
> Does the rust image also need update? It also uses bullseye base image.

this PR is generated by an automated procedure that checks hash values with what's available in docker, and it checks both debian and rust images. It seems that the rust/bullseye image hasn't been updated for a while likely because bullseye is not Debian stable anymore (bookworm is). @aptos-labs/prod-eng should we bring Debian to a current version?

PR #8737, opened 2023-06-20
[TS SDK] Add examples folder to .npmignore file

0xmaayan:
### Description
We don't need to publish the `examples` folder as part of the npm package. It increases the package size and there is no real need for it to be part of the package that users download but can simply be part of the repo.

Decrease package size by more than 60% from 3.9 MB to 1.3 MB and unpacked size by 29% from 9.1 MB to 6.4 MB

### Test Plan
test are passing

### DIFF
--- a/ecosystem/typescript/sdk/.npmignore
+++ b/ecosystem/typescript/sdk/.npmignore
@@ -2,3 +2,4 @@ coverage
 node_modules
 .aptos
 .env
+examples/


ecosystem/typescript/sdk/.npmignore
```diff
@@ -2,3 +2,4 @@ coverage
 node_modules
 .aptos
 .env
+examples/
```
banool, line 5: nit: I notice all the other dirs don't have a trailing slash, not sure if that matters

gregnazario (approved):

PR #8742, opened 2023-06-20
[move-examples] Remove legacy entry fun from marketplace example

gregnazario:
### Description
This cleans up the usage of `legacy_entry_fun` in the example

### Test Plan
https://nft-playground.gnazar.io/ is able to use it correctly to handle listings and purchases

### DIFF
--- a/aptos-move/move-examples/marketplace/sources/collection_offer.move
+++ b/aptos-move/move-examples/marketplace/sources/collection_offer.move
@@ -82,9 +82,28 @@ module collection_offer {
 
     // Initializers
 
-    #[legacy_entry_fun]
     /// Create a tokenv1 collection offer.
     public entry fun init_for_tokenv1<CoinType>(
+        purchaser: &signer,
+        creator_address: address,
+        collection_name: String,
+        fee_schedule: Object<FeeSchedule>,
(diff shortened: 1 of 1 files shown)


movekevin (commented):
#[legacy_entry_fun] is used to allow these entry funcs to return values and be used on chain directly without having to write redundant non-entry public functions. I'd keep these for now until we're done experimenting with that

bowenyang007 (approved):

aptos-move/move-examples/marketplace/sources/collection_offer.move
```diff
@@ -92,14 +91,33 @@ module collection_offer {
...
+            creator_address,
+            collection_name,
+            fee_schedule,
+            item_price,
+            amount,
+            expiration_time
+        );
+    }
+
+    public fun init_for_tokenv1_internal<CoinType>(
```
movekevin, line 106: _internal is a misleading name for a public function. I prefer "_entry" for the entry function and no suffix for the non-entry public function

movekevin (approved):
Approved except for one minor naming comment

davidiw:
landing this as  part of a stack.

PR #8693, opened 2023-06-15
[Rust Toolchain] Upgrade to rust version 1.7

JoshLind:
### Description
This PR updates the rust toolchain version from 1.66.0 (which is around 6 months old), to the latest version of rust, 1.7.0 (which is around 3 weeks old). All changes made in the PR were suggested by clippy. I've done basic due diligence to make sure no semantic changes were made, but would appreciate some additional eyes 😄 

### Test Plan
Existing test infrastructure.

PR #8744, opened 2023-06-20
[Rust Toolchain] Set rust-version to 1.70.0

bchocho:
### Description

As a follow-up to #8693. To explicitly fail the compilation instead of having potentially confusing compiler mismatch.

https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field
> If the currently selected version of the Rust compiler is older than the stated version, cargo will exit with an error, telling the user what version is required.

The check found the docker image was old and had to be bumped as well.

### Test Plan

CICD

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -240,7 +240,7 @@ homepage = "https://aptoslabs.com"
 license = "Apache-2.0"
 publish = false
 repository = "https://github.com/aptos-labs/aptos-core"
-rust-version = "1.70.0"
+rust-version = "1.70"
 
 [workspace.dependencies]
 # Internal crate dependencies.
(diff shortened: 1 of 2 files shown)


JoshLind (approved):
Thanks @bchocho 😄

ibalajiarun (approved):

Cargo.toml
```diff
@@ -240,7 +240,7 @@ homepage = "https://aptoslabs.com"
 license = "Apache-2.0"
 publish = false
 repository = "https://github.com/aptos-labs/aptos-core"
-rust-version = "1.64"
+rust-version = "1.70.0"
```
banool, line 243: I wonder, do we want to set patch version explicitly, usually it's pretty safe to leave that unspecified right? This way we pick up security fixes for free.

bchocho: This is just a minimum version (according to the doc quoted in the PR description) so it should be fine. IIUC 1.70.0 should be equivalent to 1.70. I'm fine to change it as well if you feel it's different

bchocho: Let me just change it -- the example doesn't have a patch version, so maybe it's not a good idea to include one :)

PR #8747, opened 2023-06-20
[Execution][Sharding] Support cross shard dependency

sitalkedia:
### Description

1. End-to-end change to support cross shard dependency.  On a high level, we spawn a block-stm instance per shard per sub-block and each parallel execution instance is provided a cross shard state view. This state view provides the latest value for a set of keys, which were identified as cross shard dependencies. The state view also blocks, the thread until the cross shard dependency is resolved (if needed). 
2. A transaction commit listener is added to parallel executor - this pushes writeset of a transaction once it is committed. A cross shard sender implementation of this listener pushes the writes to remote shard. This also translates the local transaction index into global index. 

There are a bunch of TODO for this

1. Failure handling is not supported yet - if the local block-stm discards a transaction, we need to mark the storage location as skipped.
2. The implementation only works for two rounds of partitioning, for more than two rounds, we need to get the round id along with txn idx and shard id in partitioning.

### Test Plan

Added end to end unit tests to ensure correctness of cross shard execution


aptos-move/aptos-vm/src/sharded_block_executor/sharded_executor_client.rs
```diff
@@ -0,0 +1,197 @@
...
+
+        let message_rxs = self.message_rx.clone();
+        let self_message_tx = Arc::new(Mutex::new(
+            self.message_txs[self.shard_id].lock().unwrap().clone(),
+        ));
+        let cross_shard_state_view =
+            Arc::new(self.create_cross_shard_state_view(state_view, &sub_block));
+        let cross_shard_state_view_clone1 = cross_shard_state_view.clone();
+        self.executor_thread_pool.scope(|s| {
+            s.spawn(move |_| {
```
zjma, line 136: Here 1 thread is occupied. If the caller gives `concurrency_level==2`, then the block-stm will get only 1 thread and then stuck?...

zjma (approved):
LGTM except the thread count issue!

aptos-move/aptos-vm/src/sharded_block_executor/sharded_executor_client.rs
```diff
@@ -0,0 +1,197 @@
...
+                );
+                // Send a stop command to the cross-shard commit receiver.
+                if round != 0 {
+                    self_message_tx
+                        .lock()
+                        .unwrap()
+                        .send(CrossShardMsg::StopMsg)
+                        .unwrap();
+                }
+                callback.send(ret).unwrap();
```
zjma, line 164: does this have to be in a separate thread? I wonder by putting it in a separate thread we further lose another thread...

sitalkedia: You are right, this doesn't need to be in a separate thread, I can change that. But we are not loosing a thread as we are spawning two additional threads.

aptos-move/aptos-vm/src/sharded_block_executor/sharded_executor_client.rs
zjma: Thanks for confirming, finally found the "+ 2"...

aptos-move/aptos-vm/src/sharded_block_executor/sharded_executor_client.rs
sitalkedia: @zjma - Actually I take it back, the execution needs to be in a separate thread as we are using scope() here, this blocks until all the task spawned inside is finished. If we move execution to outside of the scope(), then basically there is a deadlock.

aptos-move/block-executor/src/executor.rs
```diff
@@ -446,6 +449,7 @@ where
         versioned_cache: &MVHashMap<T::Key, T::Value, X>,
         last_input_output: &TxnLastInputOutput<T::Key, E::Output, E::Error>,
         base_view: &S,
+        txn_commit_listener: &Option<L>,
```
zekun000, line 452: this probably should be hold in the BlockExecutor instead of passing all the way down

aptos-move/block-executor/src/executor.rs
```diff
@@ -699,6 +719,25 @@ where
...
+                    // Apply the writes/deltas to the versioned_data_cache.
+                    ExecutionStatus::Success(output)
+                },
+                ExecutionStatus::SkipRest(output) => {
+                    // Apply the writes/deltas and record status indicating skip.
+                    ExecutionStatus::SkipRest(output)
+                },
+                ExecutionStatus::Abort(err) => {
+                    // Record the status indicating abort.
+                    ExecutionStatus::Abort(Error::UserError(err))
```
zekun000, line 733: not sure i understand why we need this change, also it can be written like
```
a @ ExecutionStatus(_) | ExecutionStatus::SkipRest(_) => a,
```

aptos-move/block-executor/src/txn_commit_listener.rs
```diff
@@ -0,0 +1,49 @@
...
+};
+use aptos_mvhashmap::types::TxnIndex;
+use std::fmt::Debug;
+
+/// An interface for listening to transaction commit events. The listener is called only once
+/// for each transaction commit.
+pub trait TransactionCommitListener: Send + Sync {
+    type ExecutionStatus;
+
+    fn on_transaction_committed(&self, txn_idx: TxnIndex, execution_status: &Self::ExecutionStatus);
```
zekun000, line 16: i'm not entirely sure whether we need this index, but anyway I think we may push this `on_commit` function to the trait TransactionOutput

aptos-move/block-executor/src/executor.rs
sitalkedia: Can you clarify what you have in mind? Is this by adding a callback in TransactionOutput function?

zekun000: initially I thought we can hold this listener in the block executor, but in the end I was wondering about supporting it directly in the TransactionOutput trait. if we could add the `on_commit` function to the trait so that we can directly call `self.output[txn_idx].on_commit()` in the commit_hook().

when we construct AptosTransactionOutput::new, we can pass in Option<Listener> that can be called later in on_commit, we don't need to have the no-op one, it can just be None.

sitalkedia: Hmm, we can do that but not sure if the API will be cleaner that way. basically we need to add something like `on_committed(&self, txn_idx: TxnIndex, execution_status: &Self::ExecutionStatus)` to the `TransactionOutput` trait and the `txn_idx` and the `execution_status` seems out of place to be in the API.

sitalkedia: As discussed offline, renamed the lister to commit hook interface but not adding it to the `TransactionOutput` trait as that makes the trait unclean.

aptos-move/block-executor/src/executor.rs
sitalkedia: I tried that first but it confuses the compiler . This is a little verbose but its the only way I could get this working. We also need to do the same in parallel execution code path BTW.

```

719 |               let res = match res {
    |  _______________________-
720 | |                 status @ ExecutionStatus::Success(_) | ExecutionStatus::SkipRest(_) => status,
    | |                                                                                        ------ this is found to be of type `task::ExecutionStatus<<E as task::ExecutorTask>::Output, <E as task::ExecutorTask>::Error>`
721 | |                 ExecutionStatus::Abort(err) => {
722 | |                     // Record the status indicating abort.
723 | |                     ExecutionStatus::Abort(Error::UserError(err))
    | |                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `ExecutionStatus<<... as ExecutorTask>::Output, ...>`, found `ExecutionStatus<_, Error<...>>`
724 | |                 },
725 | |             };
```

aptos-move/aptos-vm/src/sharded_block_executor/cross_shard_client.rs
```diff
@@ -0,0 +1,160 @@
...
+        }
+    }
+
+    fn send_remote_update_for_success(
+        &self,
+        txn_idx: TxnIndex,
+        txn_output: &AptosTransactionOutput,
+    ) {
+        let edges = self.dependent_edges.get(&txn_idx).unwrap();
+        let write_set = txn_output.committed_output().unwrap().write_set();
```
zjma, line 116: This unwrap failed in my sequential execution testing: inside the `txn_output`, the `VmOutput` hasn't been converted to `TransactionOutput`.

I had to do a workaround in my testing branch to do the conversion explicitly like [this](https://github.com/aptos-labs/aptos-core/pull/8852/commits/2ae60dbf593c92ffcb165e9182f9b747482abb95#r1247026775). See the full commit for the workaround [here](https://github.com/aptos-labs/aptos-core/pull/8852/commits/2ae60dbf593c92ffcb165e9182f9b747482abb95).

sitalkedia: Thanks for looking into this. Let me address this in a separate follow up PR.

aptos-move/block-executor/src/executor.rs
```diff
@@ -699,6 +715,16 @@ where
                 true,
             );
 
+            let res = match res {
+                ExecutionStatus::Success(output) => ExecutionStatus::Success(output),
+                ExecutionStatus::SkipRest(output) => ExecutionStatus::SkipRest(output),
+                ExecutionStatus::Abort(err) => ExecutionStatus::Abort(Error::UserError(err)),
+            };
+
+            if let Some(listener) = &self.transaction_commit_hook {
```
zekun000, line 724: nit: update the name

aptos-move/block-executor/src/executor.rs
```diff
@@ -699,6 +715,16 @@ where
                 true,
             );
 
+            let res = match res {
+                ExecutionStatus::Success(output) => ExecutionStatus::Success(output),
+                ExecutionStatus::SkipRest(output) => ExecutionStatus::SkipRest(output),
+                ExecutionStatus::Abort(err) => ExecutionStatus::Abort(Error::UserError(err)),
```
zekun000, line 721: I don't think this is correct, ExecutionStatus::Abort means nothing gets committed, the whole execution returns error

zekun000 (approved):

aptos-move/block-executor/src/executor.rs
sitalkedia: Not sure what you mean, but we will have a way to handle remote execution error in cross shard soon. This doesn't change the existing behavior as we return `UserError` -  same as previously.

zekun000: note this is not a move abort, basically this invalidates the whole block execution (including previous ones that already goes through commit hooks),

zjma (approved):

aptos-move/aptos-vm/src/sharded_block_executor/cross_shard_client.rs
zjma, line 116:
```suggestion
        let write_set = txn_output.committed_output().unwrap().write_set();
        // TODO: https://github.com/aptos-labs/aptos-core/pull/8747#discussion_r1247029085
```

aptos-move/block-executor/src/executor.rs
sitalkedia: Hmm I see.. so how do we handle this in sharded execution? Just abort the entire sharded execution I guess? When does this happen?

PR #8749, opened 2023-06-20
[aptos-vm] Log cleanups

runtian-zhou:
### Description

Added sub status field in VMStatus so that we can accumulate message from movevm.

Also updated the entrypoint for error logs so we can set different counters for each individual situations in the future.

### Test Plan

NA

### DIFF
--- a/aptos-move/aptos-vm/src/aptos_vm.rs
+++ b/aptos-move/aptos-vm/src/aptos_vm.rs
@@ -1706,7 +1706,40 @@ impl VMAdapter for AptosVM {
                                     vm_status
                                 ),
                             );
-                        }
+                        },
+                        // Type resolution failure could be a result of user input error so need to filter that out.
+                        StatusCode::TYPE_RESOLUTION_FAILURE
+                            if vm_status.sub_status()
+                                == Some(move_vm_types::errors::EUSER_TYPE_LOADING_FAILURE) =>
+                        {
+                            ()
+                        },
+                        // Paranoid mode failure. We need to be alerted about this ASAP.
+                        StatusCode::UNKNOWN_INVARIANT_VIOLATION_ERROR
+                            if vm_status.sub_status()
+                                == Some(move_vm_types::errors::EPARANOID_FAILURE) =>
+                        {
+                            error!(
+                                *log_context,
+                                "[aptos_vm] Transaction breaking paranoid mode. txn: {:?}, status: {:?}",
+                                bcs::to_bytes::<SignedTransaction>(&**txn),
+                                vm_status,
+                            );
+                        },
+                        // Paranoid mode failure but with reference counting
+                        StatusCode::UNKNOWN_INVARIANT_VIOLATION_ERROR
+                            if vm_status.sub_status()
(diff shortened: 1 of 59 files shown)


movekevin (approved):

aptos-move/aptos-vm/src/aptos_vm.rs
```diff
@@ -1695,15 +1695,29 @@ impl VMAdapter for AptosVM {
...
+                                    "[aptos_vm] Transaction breaking invariant violation. txn: {:?}, status: {:?}",
+                                    bcs::to_bytes::<SignedTransaction>(&**txn),
+                                    vm_status
+                                ),
+                            );
+                        },
+                        // We will log the rest of invariant violation directly with regular logger as they shouldn't happen.
+                        //
+                        // TODO: Add different counters for the error categories here.
+                        _ => {
```
zekun000, line 1713: we need to keep the filter for the STORAGE_ERROR, it's used by block stm to signal early terminations.

aptos-move/aptos-vm/src/aptos_vm.rs
```diff
@@ -1695,15 +1695,62 @@ impl VMAdapter for AptosVM {
...
+                            speculative_error!(
+                                log_context,
+                                format!(
+                                    "[aptos_vm] Transaction breaking invariant violation. txn: {:?}, status: {:?}",
+                                    bcs::to_bytes::<SignedTransaction>(&**txn),
+                                    vm_status
+                                ),
+                            );
+                        },
+                        // Type resolution failure could be a result of user input error so need to filter that out.
```
wrwg, line 1710: Actually type resolution could be well result of speculative execution, e.g. if one txn deploys a module and the using module is executed before.

aptos-move/aptos-vm/src/errors.rs
```diff
@@ -82,14 +82,12 @@ pub fn convert_prologue_error(
...
                 },
                 (category, reason) => {
-                    speculative_error!(
-			log_context,
-			format!("[aptos_vm] Unexpected prologue Move abort: {:?}::{:?} (Category: {:?} Reason: {:?})",
-				location, code, category, reason),
-		    );
+                    let err_msg = format!("[aptos_vm] Unexpected prologue Move abort: {:?}::{:?} (Category: {:?} Reason: {:?})",
+                    location, code, category, reason);
+                    speculative_error!(log_context, err_msg.clone());
```
wrwg, line 87: I think this will be double logging but this doesn't hurt.

third_party/move/move-core/types/src/vm_status.rs
```diff
@@ -66,6 +66,7 @@ pub enum VMStatus {
     /// dividing by zero or a missing resource
     ExecutionFailure {
         status_code: StatusCode,
+        sub_status: Option<u64>,
```
wrwg, line 69: Nice.

wrwg (approved):

aptos-move/aptos-vm/src/aptos_vm.rs
runtian-zhou: Oh interesting. Do you know where in our code could be vulnerable because of that? would need to add the sub status there as well.

Issue #8718, opened 2023-06-16
[Prover] Need to lock down the dependency (Z3 and Boogie) version


junkil-park:
The upgrade of Z3 and Boogie is mostly backward compatible, but not always. Prover currently warns users when they have installed some outdated version of Z3 and Boogie. However, Prover also need to warn users for the higher version of Z3 and Boogie because those may be incompatible.

rahxephon89:
PR to fix this: https://github.com/aptos-labs/aptos-core/pull/8750

aalok-t:
PR to fix this #9524

PR #8750, opened 2023-06-20
Fix Z3 for Mac arm64 install in dev_setup.sh

brmataptos:
### Description
Fix dev_setup.sh to get selected version of Z3 on Mac Arm64.

### Fixes
https://github.com/aptos-labs/aptos-core/issues/8718

### Test Plan
Tested on Mac Arm64 with no current Z3 installation.
Code is visibly unchanged for other cases.

### DIFF
--- a/scripts/dev_setup.sh
+++ b/scripts/dev_setup.sh
@@ -14,7 +14,7 @@
 # 3 ${HOME}/bin/, or ${INSTALL_DIR} is expected to be on the path - hashicorp tools/etc.  will be installed there on linux systems.
 
 # fast fail.
-set -eox pipefail
+set -eo pipefail
 
 SHELLCHECK_VERSION=0.7.1
 GRCOV_VERSION=0.8.2
@@ -505,9 +505,7 @@ function install_z3 {
     Z3_PKG="z3-$Z3_VERSION-x64-glibc-2.31"
   elif [[ "$(uname)" == "Darwin" ]]; then
     if [[ "$(uname -m)" == "arm64" ]]; then
-      # brew has a newer, arm64-native version
-      brew install z3
-      return
+      Z3_PKG="z3-$Z3_VERSION-arm64-osx-11.0"
     else
       Z3_PKG="z3-$Z3_VERSION-x64-osx-10.16"
     fi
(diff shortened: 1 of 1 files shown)


scripts/dev_setup.sh
```diff
@@ -14,7 +14,7 @@
 # 3 ${HOME}/bin/, or ${INSTALL_DIR} is expected to be on the path - hashicorp tools/etc.  will be installed there on linux systems.
 
 # fast fail.
-set -eox pipefail
```
brianolson, line 17: I think we should keep `-x` so we can see what's going on in the script

brmataptos: That seems useful for debugging, but it's really too noisy for everyday use.  I note it was added without discussion only recently (May 17).

rahxephon89 (approved):

scripts/dev_setup.sh
brmataptos: I'm tempted to wrap every command with side-effects with (set -x; ...) as discussed [here](https://stackoverflow.com/questions/2853803/how-to-echo-shell-commands-as-they-are-executed) but while it makes the output more useful, it makes the script pretty unreadable.  What do you think?

brmataptos: Btw, another problem is that (if selected and not installed) llvm gets installed by brew, which also installs Z3 using brew.  Maybe we need a version check for everything.

brianolson: In my dev environment, running this script is pretty rare, so noise while it runs isn't a problem, and if something goes wrong having the `-x` verbosity is useful to figure out what just happened.

brmataptos: There is already a -v flag to the script to set -x.  Is that not enough for you?

brmataptos: Ok, I've modified it to set VERBOSE if the shell is not interactive, which was the behavior described in https://github.com/aptos-labs/aptos-core/pull/8252:

"Also set -x in dev_setup.sh when running in CI only. Don't set it when running on laptops since it will become too verbose."

brianolson (approved):

PR #8758, opened 2023-06-20
[Storage][Sharding] Write a separate index to support "read everything by key prefix" operation.

grao1991:
### Description
Historically we support an API that returns everything by key prefix (e.g. return all states under an account) operation. This was implemented by seeking the StateKv db by prefix directly. However in the new sharding world, account (or whatever prefix) is not a first class concept in the storage layer, and we will store data for an account in different shards, based on the hash of StateKey. Therefore the API cannot be supported by doing a single db seek.

Our long term vision, is to move such support into indexer, before they are ready, we add this index for now to temporarily unblock the sharded db migration.

### DIFF
--- a/storage/aptosdb/src/backup/restore_utils.rs
+++ b/storage/aptosdb/src/backup/restore_utils.rs
@@ -111,12 +111,14 @@ pub(crate) fn save_transactions(
     txn_infos: &[TransactionInfo],
     events: &[Vec<ContractEvent>],
     write_sets: Vec<WriteSet>,
-    existing_batch: Option<(&mut SchemaBatch, &mut ShardedStateKvSchemaBatch)>,
+    existing_batch: Option<(
+        &mut SchemaBatch,
+        &mut ShardedStateKvSchemaBatch,
+        &SchemaBatch,
+    )>,
     kv_replay: bool,
 ) -> Result<()> {
-    if let Some(existing_batch) = existing_batch {
-        let batch = existing_batch.0;
-        let state_kv_batches = existing_batch.1;
+    if let Some((batch, state_kv_batches, state_kv_metadata_batch)) = existing_batch {
         save_transactions_impl(
             Arc::clone(&ledger_store),
             transaction_store,
@@ -129,11 +131,13 @@ pub(crate) fn save_transactions(
             write_sets.as_ref(),
             batch,
             state_kv_batches,
+            state_kv_metadata_batch,
             kv_replay,
         )?;
     } else {
         let mut batch = SchemaBatch::new();
         let mut sharded_kv_schema_batch = new_sharded_kv_schema_batch();
+        let state_kv_metadata_batch = SchemaBatch::new();
         save_transactions_impl(
             Arc::clone(&ledger_store),
             transaction_store,
@@ -146,15 +150,17 @@ pub(crate) fn save_transactions(
             write_sets.as_ref(),
             &mut batch,
             &mut sharded_kv_schema_batch,
+            &state_kv_metadata_batch,
             kv_replay,
         )?;
         // get the last version and commit to the state kv db
         // commit the state kv before ledger in case of failure happens
         let last_version = first_version + txns.len() as u64 - 1;
-        state_store
-            .state_db
-            .state_kv_db
-            .commit(last_version, sharded_kv_schema_batch)?;
+        state_store.state_db.state_kv_db.commit(
+            last_version,
+            state_kv_metadata_batch,
+            sharded_kv_schema_batch,
+        )?;
 
         // TODO(grao): Support splitted ledger DBs here.
         ledger_store.ledger_db.metadata_db().write_schemas(batch)?;
@@ -220,6 +226,7 @@ pub(crate) fn save_transactions_impl(
     write_sets: &[WriteSet],
     batch: &mut SchemaBatch,
     state_kv_batches: &mut ShardedStateKvSchemaBatch,
+    state_kv_metadata_batch: &SchemaBatch,
     kv_replay: bool,
 ) -> Result<()> {
     // TODO(grao): Support splited ledger db here.
@@ -234,7 +241,14 @@ pub(crate) fn save_transactions_impl(
     }
 
     if kv_replay && first_version > 0 && state_store.get_usage(Some(first_version - 1)).is_ok() {
-        state_store.put_write_sets(write_sets.to_vec(), first_version, batch, state_kv_batches)?;
+        state_store.put_write_sets(
+            write_sets.to_vec(),
+            first_version,
+            batch,
+            state_kv_batches,
+            state_kv_metadata_batch,
+            state_store.state_kv_db.enabled_sharding(),
+        )?;
     }
 
     Ok(())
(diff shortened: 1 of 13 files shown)


storage/aptosdb/src/backup/restore_utils.rs
```diff
@@ -112,12 +112,17 @@ pub(crate) fn save_transactions(
...
+        &mut SchemaBatch,
+        &mut ShardedStateKvSchemaBatch,
+        &SchemaBatch,
+    )>,
     kv_replay: bool,
 ) -> Result<()> {
     if let Some(existing_batch) = existing_batch {
         let batch = existing_batch.0;
         let state_kv_batches = existing_batch.1;
+        let state_kv_metadata_batch = existing_batch.2;
```
msmouse, line 125: why not just 
```
if let Some((batch, state_kv_baches, state_kv_metadata_batch)) = existing_batch {
```

storage/aptosdb/src/schema/state_value_index/mod.rs
```diff
@@ -0,0 +1,59 @@
...
+    state_store::{state_key::StateKey, state_key_prefix::StateKeyPrefix},
+    transaction::Version,
+};
+use byteorder::{BigEndian, ReadBytesExt, WriteBytesExt};
+use std::{io::Write, mem::size_of};
+
+type Key = (StateKey, Version);
+
+// (Key, is_tombstone)
+define_schema!(StateValueIndexSchema, Key, (), STATE_VALUE_INDEX_CF_NAME);
```
msmouse, line 20: where is `is_tombstone`?

storage/aptosdb/src/schema/state_value_index/mod.rs
```diff
@@ -0,0 +1,59 @@
...
+    define_schema,
+    schema::{KeyCodec, SeekKeyCodec, ValueCodec},
+};
+use aptos_types::{
+    state_store::{state_key::StateKey, state_key_prefix::StateKeyPrefix},
+    transaction::Version,
+};
+use byteorder::{BigEndian, ReadBytesExt, WriteBytesExt};
+use std::{io::Write, mem::size_of};
+
```
areshand, line 16: could we add a brief description on the motivation of having a new schema?

storage/aptosdb/src/state_store/mod.rs
```diff
@@ -706,6 +720,22 @@ impl StateStore {
...
+        // TODO(grao): Remove when we are ready.
+        if put_state_value_indices {
+            value_state_sets
+                .par_iter()
+                .enumerate()
+                .try_for_each(|(i, updates)| {
+                    let version = first_version + i as Version;
+                    updates.iter().flatten().try_for_each(|(k, _)| {
+                        state_kv_metadata_batch
+                            .put::<StateValueIndexSchema>(&(k.clone(), version), &())
```
areshand, line 734: what is the purpose of having empty value &()?

grao1991: It needs to have something as the value.

storage/aptosdb/src/schema/state_value_index/mod.rs
grao1991: Added in the PR description.

storage/aptosdb/src/schema/state_value_index/mod.rs
grao1991: Removed.

storage/aptosdb/src/backup/restore_utils.rs
grao1991: Done

storage/aptosdb/src/schema/state_value_index/mod.rs
areshand: maybe add it to the top of code file as well ?

storage/aptosdb/src/state_store/mod.rs
areshand: I see. where is the value populated? I was not able to find the place in the PR.

storage/aptosdb/src/schema/state_value_index/mod.rs
grao1991: Done

storage/aptosdb/src/state_store/mod.rs
grao1991: Here. we fill an empty thing "&()" as the value, because we don't need a value.

areshand: the key only has stateKey and version. how do we get the state value? I assume the purpose is to store all state values in one schema to "read everything by key prefix".

grao1991: state value is still stored in the old "StateValueSchema" (sharded). We get all keys from the StateValueIndexSchema, then use those keys to lookup the value in StateValueSchema.

storage/aptosdb/src/utils/iterators.rs
```diff
@@ -149,14 +171,19 @@ impl<'a> PrefixedStateValueIterator<'a> {
                 }
 
                 if version > self.desired_version {
-                    self.inner
+                    self.kv_iter
+                        .as_mut()
+                        .unwrap()
```
msmouse, line 176: I feel it's more elegant to pass `kv_iter` in

storage/aptosdb/src/utils/iterators.rs
```diff
@@ -165,13 +192,61 @@ impl<'a> PrefixedStateValueIterator<'a> {
...
+                    .unwrap()
+                    .seek(&(state_key.clone(), 0))?;
+
+                if let Some(Some(state_value)) =
+                    self.db
+                        .db_shard(state_key.get_shard_id())
+                        .get::<StateValueSchema>(&(state_key.clone(), version))?
+                {
+                    return Ok(Some((state_key, state_value)));
+                }
```
msmouse, line 234: it's a db corruption if the `get()` returns `None`, right? (I mean shall we error out?)

grao1991: I originally wrote in that way but feel it is an overkill?

msmouse: well, maybe just unwrap()? not too verbose and the logic is clear

grao1991: I'm not conformable to write `unwrap` here because the index and the state value are not written in the same db transaction.

msmouse: exactly the reason we shouldn't just return None in the case of two dbs mismatch?

storage/aptosdb/src/utils/iterators.rs
grao1991: Made a variable.

storage/aptosdb/src/utils/iterators.rs
grao1991: Sure, changed to return an error.

msmouse (approved):

msmouse (approved):

areshand (approved):

PR #8759, opened 2023-06-20
[Gas Estimation] some improvements based on AIP feedback

bchocho:
### Description

Follow-up of #8186 based on feedback.
1. Add an option to override the API results with fixed constants.
2. Re-read from the cache after acquiring write lock, as the entry could have been written between the initial read and acquiring the write lock.
3. Enforce invariant that aggressive >= market >= low.

### Test Plan

Added unit test.

### DIFF
--- a/api/src/context.rs
+++ b/api/src/context.rs
@@ -1053,7 +1053,7 @@ impl Context {
                 );
                 return Ok(self.default_gas_estimation(min_gas_unit_price));
             },
-            Some(price) => *price,
+            Some(price) => low_price.max(*price),
         };
 
         // (3) aggressive
@@ -1069,7 +1069,7 @@ impl Context {
                 );
                 return Ok(self.default_gas_estimation(min_gas_unit_price));
             },
-            Some(price) => *price,
+            Some(price) => market_price.max(*price),
         };
         // round up to next bucket
         let aggressive_price = self.next_bucket(p90_price);
--- a/api/src/context.rs
+++ b/api/src/context.rs
@@ -51,7 +51,7 @@ use futures::{channel::oneshot, SinkExt};
 use move_core_types::language_storage::{ModuleId, StructTag};
 use std::{
     collections::{BTreeMap, HashMap},
-    ops::Bound::Included,
+    ops::{Bound::Included, Deref},
     sync::{Arc, RwLock, RwLockWriteGuard},
     time::Instant,
 };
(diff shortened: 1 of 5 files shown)


config/src/config/gas_estimation_config.rs
```diff
@@ -13,6 +13,9 @@ use serde::{Deserialize, Serialize};
 pub struct GasEstimationConfig {
     /// A gate for computing GasEstimation. If false, just returns the default.
     pub enabled: bool,
+    /// Provide static values to override, e.g., Some((100, 100, 150)). If set, just returns these
+    /// values instead of computing the actual GasEstimation.
+    pub static_override: Option<(u64, u64, u64)>,
```
banool, line 18: Instead of a tuple could you make this a struct? While it might be obvious in the code, in the configs this will appear to be just three numbers. Being able to set them as named fields would be better.

banool (changes requested):
Looks good, just requesting changes for how we represent static_override in the config.

banool:
Also once that change is made, it'd be great if we could test that setting the static overrides in the config works. As in not just in a test like this, but write a config file and start a node off it.

bchocho:
Thanks @banool I also ran the local test with a custom yaml and confirmed the REST call gives the overrides.

api/src/context.rs
```diff
@@ -900,17 +902,30 @@ impl Context {
...
+                gas_estimate: static_override.market,
+                prioritized_gas_estimate: Some(static_override.aggressive),
+            });
+        }
 
         let epoch = ledger_info.epoch.0;
 
         // 0. (0) Return cached result if it exists
-        if let Some(cached_gas_estimation) = self.cached_gas_estimation(epoch) {
+        let cache = self.gas_estimation_cache.read().unwrap();
```
JoshLind, line 916: Silly question: do you need this read lock here? `self.cached_gas_estimation(&cache, epoch)` internally grabs the read lock, right?

api/src/context.rs
```diff
@@ -900,17 +902,30 @@ impl Context {
...
+        if let Some(cached_gas_estimation) = self.cached_gas_estimation(&cache, epoch) {
             return Ok(cached_gas_estimation);
         }
+        drop(cache);
 
         // 0. (1) Write lock and prepare cache
         let mut cache = self.gas_estimation_cache.write().unwrap();
-        // TODO: retry cached result after acquiring write lock
+        // Retry cached result after acquiring write lock
```
JoshLind, line 924: Will anything bad happen if you don't fix this race? 😄 Just curious.

JoshLind (approved):
Nice, looks reasonable to me. Providing the first stamp to help unblock things for the next cut 😄

api/src/context.rs
JoshLind: Aah, missed that -- makes sense, thanks 😄

api/src/context.rs
JoshLind: Nice, I was wondering if the issue was just recomputation or if there was a more serious correctness concern, but this makes sense 😄

banool (approved):
Looks good!!!!

PR #8762, opened 2023-06-20
[Network] Perform message deserialization in parallel.

JoshLind:
### Description
This PR updates the networking code to perform message deserialization in parallel (as opposed to serially, where each application only has a single thread that processes one message at a time). Now, we use `for_each_concurrent` that spawns up to a fixed number of threads and does deserialization in parallel (per application).

This was originally reported as a bottleneck in the consensus tests, but it has since been reproduced by the network benchmarks.

### Test Plan
Existing test infrastructure.

### DIFF
--- a/aptos-node/src/network.rs
+++ b/aptos-node/src/network.rs
@@ -200,6 +200,7 @@ pub fn setup_networks_and_get_interfaces(
                 consensus_network_handle = Some(register_client_and_service_with_network(
                     &mut network_builder,
                     network_id,
+                    &network_config,
                     consensus_network_configuration(node_config),
                 ));
             }
@@ -209,6 +210,7 @@ pub fn setup_networks_and_get_interfaces(
         let mempool_network_handle = register_client_and_service_with_network(
             &mut network_builder,
             network_id,
+            &network_config,
             mempool_network_configuration(node_config),
         );
         mempool_network_handles.push(mempool_network_handle);
@@ -217,6 +219,7 @@ pub fn setup_networks_and_get_interfaces(
         let peer_monitoring_service_network_handle = register_client_and_service_with_network(
             &mut network_builder,
             network_id,
+            &network_config,
             peer_monitoring_network_configuration(node_config),
         );
         peer_monitoring_service_network_handles.push(peer_monitoring_service_network_handle);
@@ -225,6 +228,7 @@ pub fn setup_networks_and_get_interfaces(
         let storage_service_network_handle = register_client_and_service_with_network(
             &mut network_builder,
             network_id,
+            &network_config,
             storage_service_network_configuration(node_config),
         );
         storage_service_network_handles.push(storage_service_network_handle);
@@ -277,13 +281,18 @@ fn create_network_runtime(network_config: &NetworkConfig) -> Runtime {
 }
 
 /// Registers a new application client and service with the network
-fn register_client_and_service_with_network<T: Serialize + for<'de> Deserialize<'de>>(
+fn register_client_and_service_with_network<
+    T: Serialize + for<'de> Deserialize<'de> + Send + 'static,
+>(
     network_builder: &mut NetworkBuilder,
     network_id: NetworkId,
+    network_config: &NetworkConfig,
     application_config: NetworkApplicationConfig,
 ) -> ApplicationNetworkHandle<T> {
-    let (network_sender, network_events) =
-        network_builder.add_client_and_service(&application_config);
+    let (network_sender, network_events) = network_builder.add_client_and_service(
+        &application_config,
+        network_config.max_parallel_deserialization_tasks,
+    );
     ApplicationNetworkHandle {
         network_id,
         network_sender,
--- a/config/src/config/network_config.rs
+++ b/config/src/config/network_config.rs
@@ -165,7 +165,7 @@ impl NetworkConfig {
         };
 
         // Configure the number of parallel deserialization tasks
-        config.configure_num_deseserialization_tasks();
+        config.configure_num_deserialization_tasks();
 
         // Prepare the identity based on the identity format
         config.prepare_identity();
@@ -176,7 +176,7 @@ impl NetworkConfig {
     /// Configures the number of parallel deserialization tasks
     /// based on the number of CPU cores of the machine. This is
     /// only done if the config does not specify a value.
-    fn configure_num_deseserialization_tasks(&mut self) {
+    fn configure_num_deserialization_tasks(&mut self) {
         if self.max_parallel_deserialization_tasks.is_none() {
             self.max_parallel_deserialization_tasks = Some(num_cpus::get());
         }
@@ -522,7 +522,7 @@ mod tests {
         };
 
         // Configure the number of deserialization tasks and verify that it is not overridden
-        network_config.configure_num_deseserialization_tasks();
+        network_config.configure_num_deserialization_tasks();
         assert_eq!(network_config.max_parallel_deserialization_tasks, Some(1));
     }
 }
--- a/config/src/config/network_config.rs
+++ b/config/src/config/network_config.rs
@@ -119,6 +119,8 @@ pub struct NetworkConfig {
     pub outbound_rate_limit_config: Option<RateLimitConfig>,
     /// The maximum size of an inbound or outbound message (it may be divided into multiple frame)
     pub max_message_size: usize,
+    /// The maximum number of parallel message deserialization tasks that can run (per application)
+    pub max_parallel_deserialization_tasks: Option<usize>,
 }
 
 impl Default for NetworkConfig {
@@ -159,9 +161,25 @@ impl NetworkConfig {
             inbound_tx_buffer_size_bytes: Some(INBOUND_TCP_TX_BUFFER_SIZE),
             outbound_rx_buffer_size_bytes: Some(OUTBOUND_TCP_RX_BUFFER_SIZE),
             outbound_tx_buffer_size_bytes: Some(OUTBOUND_TCP_TX_BUFFER_SIZE),
+            max_parallel_deserialization_tasks: None,
         };
+
+        // Configure the number of parallel deserialization tasks
+        config.configure_num_deseserialization_tasks();
+
+        // Prepare the identity based on the identity format
         config.prepare_identity();
+
         config
+    }
+
+    /// Configures the number of parallel deserialization tasks
+    /// based on the number of CPU cores of the machine. This is
+    /// only done if the config does not specify a value.
+    fn configure_num_deseserialization_tasks(&mut self) {
+        if self.max_parallel_deserialization_tasks.is_none() {
+            self.max_parallel_deserialization_tasks = Some(num_cpus::get());
+        }
     }
 
     pub fn identity_key(&self) -> x25519::PrivateKey {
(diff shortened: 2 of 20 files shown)


network/src/protocols/network/mod.rs
```diff
@@ -172,22 +174,58 @@ pub trait NewNetworkEvents {
     fn new(
         peer_mgr_notifs_rx: aptos_channel::Receiver<(PeerId, ProtocolId), PeerManagerNotification>,
         connection_notifs_rx: aptos_channel::Receiver<PeerId, ConnectionNotification>,
+        bounded_executor: BoundedExecutor,
     ) -> Self;
 }
 
-impl<TMessage: Message> NewNetworkEvents for NetworkEvents<TMessage> {
+impl<TMessage: Message + Send + 'static> NewNetworkEvents for NetworkEvents<TMessage> {
```
grao1991, line 181: Why does the TMessage need to be static?

network/src/protocols/network/mod.rs
```diff
@@ -172,22 +174,58 @@ pub trait NewNetworkEvents {
...
-            peer_mgr_notif_to_event
-                as fn(PeerManagerNotification) -> future::Ready<Option<Event<TMessage>>>,
-        );
+        // Create a channel for deserialized messages
+        let (deserialized_message_sender, deserialized_message_receiver) =
+            mpsc::channel(MAX_DESERIALIZATION_QUEUE_SIZE);
+
+        // Deserialize the peer manager notifications in parallel (for each
+        // network application) and send them to the receiver. Note: this
+        // may cause out of order message delivery.
```
grao1991, line 196: Would "out of order delivery" create any problem?

network/src/protocols/network/mod.rs
```diff
@@ -172,22 +174,58 @@ pub trait NewNetworkEvents {
...
+        Handle::current().spawn(async move {
+            // For each notification, spawn a new task to deserialize the message
+            while let Some(peer_manager_notification) = peer_mgr_notifs_rx.next().await {
+                let mut deserialized_message_sender = deserialized_message_sender.clone();
+                bounded_executor
+                    .spawn(async move {
+                        if let Some(deserialized_message) =
+                            peer_mgr_notif_to_event(peer_manager_notification).await
```
msmouse, line 204: `peer_mgr_notif_to_event()` no longer needs to return a future (it's actually a blocking function)

And using `spawn()` in hope these heavy operations will happen in different threads seems not the rightest thing to do 😂

network/src/protocols/network/mod.rs
```diff
@@ -172,22 +174,58 @@ pub trait NewNetworkEvents {
...
-impl<TMessage: Message> NewNetworkEvents for NetworkEvents<TMessage> {
+impl<TMessage: Message + Send + 'static> NewNetworkEvents for NetworkEvents<TMessage> {
     fn new(
-        peer_mgr_notifs_rx: aptos_channel::Receiver<(PeerId, ProtocolId), PeerManagerNotification>,
+        mut peer_mgr_notifs_rx: aptos_channel::Receiver<
+            (PeerId, ProtocolId),
+            PeerManagerNotification,
+        >,
         connection_notifs_rx: aptos_channel::Receiver<PeerId, ConnectionNotification>,
+        bounded_executor: BoundedExecutor,
```
zekun000, line 188: passing the executor here seems causing a lot of changes. does `for_each_concurrent` on the peer_mgr_notifs_rx not work?

aptos-node/src/network.rs
```diff
@@ -263,6 +268,14 @@ pub fn setup_networks_and_get_interfaces(
     )
 }
 
+/// Creates a bounded executor for the given network config and runtime
+fn create_bounded_executor(network_config: &NetworkConfig, runtime: &Runtime) -> BoundedExecutor {
```
sitalkedia, line 272: nit: `create_bounded_executor` name is pretty generic. A better naming might be `create_deserialization_executor` ?

aptos-node/src/network.rs
```diff
@@ -277,13 +290,16 @@ fn create_network_runtime(network_config: &NetworkConfig) -> Runtime {
...
 
 /// Registers a new application client and service with the network
-fn register_client_and_service_with_network<T: Serialize + for<'de> Deserialize<'de>>(
+fn register_client_and_service_with_network<
+    T: Serialize + for<'de> Deserialize<'de> + Send + 'static,
+>(
     network_builder: &mut NetworkBuilder,
     network_id: NetworkId,
     application_config: NetworkApplicationConfig,
+    bounded_executor: BoundedExecutor,
```
sitalkedia, line 299: also rename this to `deserialization_executor`  for clarity

network/src/protocols/network/mod.rs
```diff
@@ -14,24 +14,30 @@ use crate::{
...
 use pin_project::pin_project;
 use serde::{de::DeserializeOwned, Serialize};
 use std::{cmp::min, fmt::Debug, marker::PhantomData, pin::Pin, time::Duration};
+use tokio::runtime::Handle;
 
 pub trait Message: DeserializeOwned + Serialize {}
 impl<T: DeserializeOwned + Serialize> Message for T {}
 
+// TODO: do we want to make this configurable?
+const MAX_DESERIALIZATION_QUEUE_SIZE: usize = 1024;
```
sitalkedia, line 39: Why does this need to be bounded? What happens if this queue gets full?

brianolson: The effective current depth of queue here is zero? Trying to read a deserialized message causes trying to read from the upstream channel, do the deserialize, and return the result.

network/src/protocols/network/mod.rs
gregnazario: I'm pretty sure the order matters unless the downstream application handles out of order messages.

The RPC protocol is built around ensuring that the messages are properly in order and most of the applications are expecting that they get the correct message during the RPC call

config/src/config/network_config.rs
```diff
@@ -50,6 +50,7 @@ pub const MAX_APPLICATION_MESSAGE_SIZE: usize =
     (MAX_MESSAGE_SIZE - MAX_MESSAGE_METADATA_SIZE) - MESSAGE_PADDING_SIZE; /* The message size that applications should check against */
 pub const MAX_FRAME_SIZE: usize = 4 * 1024 * 1024; /* 4 MiB large messages will be chunked into multiple frames and streamed */
 pub const MAX_MESSAGE_SIZE: usize = 64 * 1024 * 1024; /* 64 MiB */
+pub const MAX_PARALLEL_DESERIALIZATION_TASKS: usize = 500; // This is per application and should be enough for max load
```
gregnazario, line 53: Probably should benchmark / it should be roughly same as the number of cores

network/src/protocols/network/mod.rs
JoshLind: This is required to satisfy the `move` call in `spawn_blocking`, e.g., see: https://tokio.rs/tokio/tutorial/spawning

network/src/protocols/network/mod.rs
JoshLind: > peer_mgr_notif_to_event() no longer needs to return a future (it's actually a blocking function)

Aah, nice -- missed this! Thanks :D Also, I changed to use `spawn_blocking` to satisfy the heavy operation requirements.

network/src/protocols/network/mod.rs
JoshLind: Yeah, good call. I changed to `for_each_concurrent`. We still need to pass through the max number of deserialization tasks parameter, but, it should hopefully be cleaner now 😄

aptos-node/src/network.rs
JoshLind: This should be no longer needed now that we removed the bounded executor requirement 😄

aptos-node/src/network.rs
JoshLind: This should be no longer needed now that we removed the bounded executor 😄

config/src/config/network_config.rs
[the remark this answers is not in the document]
JoshLind: Yeah, I've changed this to default to the number of CPU cores, which seems like the easiest thing to do 😄

JoshLind: Yeah, I've changed this to default to the number of CPU cores and added a test 😄

network/src/protocols/network/mod.rs
JoshLind: I've changed this so that we: (i) provide a queue per peer of max depth 30 (which should be more than enough); and (ii) if the queue fills up (e.g., the application layers fails to handle the deserialized messages fast enough), we'll start dropping messages instead of blocking or consuming large amounts of memory. This seems like the safer approach. Thoughts?

network/src/protocols/network/mod.rs
JoshLind: > Would "out of order delivery" create any problem?

I double checked with @zekun000 and @bchocho and we all agreed that it shouldn't. Applications should not be assuming ordering of messages because that assumption can be trivially broken at any point in the stack. I also had a quick look at various components and saw this not to be the case. All applications seem to handle this already (as they should) 😄

network/src/protocols/network/mod.rs
```diff
@@ -172,22 +172,67 @@ pub trait NewNetworkEvents {
...
+            peer_mgr_notifs_rx
+                .for_each_concurrent(
+                    max_parallel_deserialization_tasks,
+                    move |peer_manager_notification| {
+                        // Get the peer ID for the notification
+                        let deserialized_message_sender = deserialized_message_sender.clone();
+                        let peer_id_for_notification = peer_manager_notification.get_peer_id();
+
+                        // Spawn a new blocking task to deserialize the message
+                        Handle::current()
```
zekun000, line 206: nit: there's a shortcut for this - tokio::task::spawn_blocking

network/src/protocols/network/mod.rs
```diff
@@ -172,22 +172,67 @@ pub trait NewNetworkEvents {
...
+            QueueStyle::FIFO,
+            MAX_DESERIALIZATION_QUEUE_SIZE_PER_PEER,
+            None,
         );
+
+        // Deserialize the peer manager notifications in parallel (for each
+        // network application) and send them to the receiver. Note: this
+        // may cause out of order message delivery, but applications
+        // should already be handling this.
+        Handle::current().spawn(async move {
```
zekun000, line 196: nit: tokio::spawn(..)

zekun000 (approved):

network/src/protocols/network/mod.rs
JoshLind: Done, thanks! 😄

network/src/protocols/network/mod.rs
JoshLind: Done, thanks! 😄

zekun000:
btw not sure if we want to limit a upper bound for the parallelism, beefy machines can have 100+ cores, we probably don't want to have that many

davidiw (approved):

JoshLind:
> btw not sure if we want to limit a upper bound for the parallelism, beefy machines can have 100+ cores, we probably don't want to have that many

Hmm... yeah, I'd need to think about that a little more 🤔. For now, I'll leave things as they are to unblock the existing issues 😄 We can always extend it if we see it become a problem.

PR #8766, opened 2023-06-21
[deps] Upgrade to Clap 4

gregnazario:
### Description
Upgrade to clap 4 for consistency and the latest changes

Take a look at the Clap changelog for what each name was replaced with https://github.com/clap-rs/clap/blob/master/CHANGELOG.md

### Test Plan
CI / tested some commands manually

Colors are now added appropriately:

Screenshot 2023-06-21 at 11 52 29 AM

### DIFF
--- a/crates/aptos-faucet/core/src/funder/common.rs
+++ b/crates/aptos-faucet/core/src/funder/common.rs
@@ -69,7 +69,7 @@ pub struct ApiConnectionConfig {
     /// Chain ID of the network this client is connecting to. For example, for mainnet:
     /// "MAINNET" or 1, testnet: "TESTNET" or 2. If there is no predefined string
     /// alias (e.g. "MAINNET"), just use the number. Note: Chain ID of 0 is not allowed.
-    #[clap(long, default_value = "2")]
+    #[clap(long, default_value_t = ChainId::testnet())]
     pub chain_id: ChainId,
 }
 
--- a/crates/aptos-faucet/core/src/funder/common.rs
+++ b/crates/aptos-faucet/core/src/funder/common.rs
@@ -58,12 +58,12 @@ pub struct ApiConnectionConfig {
     /// To manually generate a keypair, use generate-key:
     /// `cargo run -p generate-keypair -- -o <output_file_path>`
     #[serde(default = "ApiConnectionConfig::default_mint_key_file_path")]
-    #[clap(long, default_value = DEFAULT_KEY_FILE_PATH, parse(from_os_str))]
+    #[clap(long, default_value = DEFAULT_KEY_FILE_PATH, value_parser)]
     key_file_path: PathBuf,
 
     /// Hex string of an Ed25519PrivateKey for minting / transferring coins.
     #[serde(skip_serializing_if = "Option::is_none")]
-    #[clap(long, parse(try_from_str = ConfigKey::from_encoded_string))]
+    #[clap(long, value_parser = ConfigKey::<Ed25519PrivateKey>::from_encoded_string)]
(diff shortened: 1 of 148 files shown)


banool:
So this is a small thing but clap 4 doesn't have colors by default which makes me sad because I like the colors. Instead, you have to use some kind of plugin or something. What if we did that 😛

banool:
I notice this PR does both normal aptos-core and third_party. I think we're meant to make changes to third_party separately?

crates/aptos-faucet/core/src/funder/common.rs
```diff
@@ -58,18 +58,18 @@ pub struct ApiConnectionConfig {
...
     #[serde(skip_serializing_if = "Option::is_none")]
-    #[clap(long, parse(try_from_str = ConfigKey::from_encoded_string))]
+    #[clap(long, value_parser = ConfigKey::<Ed25519PrivateKey>::from_encoded_string)]
     key: Option<ConfigKey<Ed25519PrivateKey>>,
 
     /// Chain ID of the network this client is connecting to. For example, for mainnet:
     /// "MAINNET" or 1, testnet: "TESTNET" or 2. If there is no predefined string
     /// alias (e.g. "MAINNET"), just use the number. Note: Chain ID of 0 is not allowed.
-    #[clap(long, default_value = "2")]
+    #[clap(long, default_value_t = ChainId::testnet())]
```
banool, line 72: Neat this is way better.

crates/aptos/src/common/types.rs
```diff
@@ -953,19 +953,19 @@ impl RestOptions {
...
-    #[clap(long, parse(from_os_str))]
+    #[clap(long, value_parser)]
     pub output_dir: Option<PathBuf>,
     /// Named addresses for the move binary
     ///
     /// Example: alice=0x1234, bob=0x5678
     ///
     /// Note: This will fail if there are duplicates in the Move.toml file remove those first.
-    #[clap(long, parse(try_from_str = crate::common::utils::parse_map), default_value = "")]
+    #[clap(long, value_parser = crate::common::utils::parse_map::<String, AccountAddressWrapper>, default_value = "")]
```
banool, line 968: Unfortunate that it now requires explicit typing.

third_party/move/move-command-line-common/src/values.rs
```diff
@@ -60,7 +60,7 @@ pub enum ParsedValue<Extra: ParsableValue = ()> {
     Custom(Extra),
 }
 
-pub trait ParsableValue: Sized {
+pub trait ParsableValue: Sized + Sync + Send + Clone + 'static {
```
banool, line 63: How come these extra trait bounds are necessary?

banool (approved):
Dang nice job, this is a lot. Took a while to review lol but besides those things I mentioned earlier and some comments here it looks good to me.

movekevin (approved):

ibalajiarun (approved):

third_party/move/move-command-line-common/src/values.rs
gregnazario: There's a bunch of weird explicit inputs are necessary now.  The developers working on it moved from a larger codebase and implicit inputs to a smaller codebase and explicit inputs.  Inputs now need to be clonable, and sometimes lifetimes need to be cleaned up / in this case it needs to be okay to send multithreaded

wrwg (approved):
This CLAP guys should stop changing the APIs all the time for little good reason...

gregnazario:
Currently adding a change to add the colors back in, and then we'll be good to go!

banool:
Weewwwww colors!!!!!

PR #8768, opened 2023-06-21
[compiler v2] Initial setup of bytecode generation and processing

wrwg:
This refines the compiler driver and testsuite by the basic modules for bytecode generation and processing:

- Adds a generator from the Move Model AST to stackless bytecode. This first version only deals with basic expressions.
- Configures a stackless bytecode processor pipeline. This will allow us to add multiple phases like safety and borrow analysis and optimizations. For now the pipeline is empty.
- Adopts the testsuite to support a test configuration which is configured via the test source directory. This allows us to group tests for testing different aspects of a pipeline.

Some of this may look like overkill but this apporach has been worked well for the stackless bytecode processing of the Move Prover. Subsequent PRs will add first processors to the pipeline and iilustrate the usage.


rahxephon89 (approved):

third_party/move/move-compiler-v2/src/bytecode_generator.rs
```diff
@@ -0,0 +1,405 @@
...
+    }
+    let mut results = vec![];
+    for ty in gen.func_env.get_result_type().flatten() {
+        let temp = gen.new_temp(ty);
+        results.push(temp)
+    }
+    gen.scopes.push_back(scope);
+    if let Some(def) = gen.func_env.get_def().cloned() {
+        // Need to clone expression if present because of sharing issues with `gen`. However, because
+        // of internalization, clone is cheap.
```
brmataptos, line 47: I think "interning" is better than "internalization" here, which has too many other meanings even if it might be used in this case (which isn't impossible).

third_party/move/move-compiler-v2/src/bytecode_generator.rs
```diff
@@ -0,0 +1,405 @@
...
+// Generator state and helpers
+
+/// Internal state of the code generator
+struct Generator<'env> {
+    /// Access to the function env and its parent.
+    func_env: FunctionEnv<'env>,
+    /// The temporaries allocated so far.
+    temps: Vec<Type>,
+    /// A list of scopes, where each scope is a map from symbol to  assigned temporary.
+    scopes: VecDeque<BTreeMap<Symbol, TempIndex>>,
```
brmataptos, line 81: Why a VecDeque here?

third_party/move/move-compiler-v2/src/bytecode_generator.rs
```diff
@@ -0,0 +1,405 @@
...
+        let ty = self.env().get_node_type(id);
+        let bytecode_op = match ty {
+            Type::Primitive(PrimitiveType::U8) => BytecodeOperation::CastU8,
+            Type::Primitive(PrimitiveType::U16) => BytecodeOperation::CastU16,
+            Type::Primitive(PrimitiveType::U32) => BytecodeOperation::CastU32,
+            Type::Primitive(PrimitiveType::U64) => BytecodeOperation::CastU64,
+            Type::Primitive(PrimitiveType::U128) => BytecodeOperation::CastU128,
+            Type::Primitive(PrimitiveType::U256) => BytecodeOperation::CastU256,
+            _ => {
+                self.internal_error(id, "inconsistent type");
```
brmataptos, line 351: Ok, so I assume that cast calls only work for uint types.  But how should I know that?  Is there a doc somewhere I should refer to?

brmataptos (commented):
A few naming/comment nits and questions.

third_party/move/move-compiler-v2/tests/testsuite.rs
```diff
@@ -20,44 +40,108 @@ fn test_runner(path: &Path) -> datatest_stable::Result<()> {
...
-        let mut out = match move_compiler_v2::run_move_compiler(&mut error_writer, options) {
-            Ok(_) => "succeeded".to_owned(),
-            Err(e) => format!("failed: {}", e),
-        };
-        let diag = String::from_utf8_lossy(&error_writer.into_inner()).to_string();
-        if !diag.is_empty() {
-            out = format!("{}\nDiagnostics:\n{}", out, diag)
+        // Putting the generated test baseline into a Refcell to avoid problems with mut borrow
+        // in closures.
+        let baseline = RefCell::new(String::new());
```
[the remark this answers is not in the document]
wrwg: Done

third_party/move/move-model/src/ast.rs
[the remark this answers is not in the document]
wrwg: Done

third_party/move/move-compiler-v2/tests/checking/basic.exp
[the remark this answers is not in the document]
wrwg: This is because I haven't even started writing tests yet. Those are just smoke tests to see whether the setup works, They will be renamed later once we add systematic tests.

third_party/move/move-compiler-v2/src/lib.rs
```diff
@@ -18,10 +24,40 @@ pub fn run_move_compiler_to_stderr(options: Options) -> anyhow::Result<()> {
...
+    if options.dump_bytecode {
+        // Dump bytecode to files, using a basename for the individual sources derived
+        // from the first input file.
+        let dump_base_name = options
+            .sources
+            .get(0)
+            .and_then(|f| {
+                Path::new(f)
+                    .file_name()
+                    .map(|f| f.to_string_lossy().to_string())
```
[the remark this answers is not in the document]
wrwg: Agree, though generally, this isn't done consistently in Rust code. Done.

third_party/move/move-compiler-v2/src/bytecode_generator.rs
wrwg: That would be the Move book (at least I hope it clarifies this), which we host at aptos.dev.

brmataptos: Indeed.  "Integer types of one size can be cast to integer types of another size. Integers are the only types in Move that support casting."  (https://move-language.github.io/move/integers.html).

I don't think it would be terrible to add that as a comment: "Integers are the only types in Move that support casting."

third_party/move/move-compiler-v2/src/lib.rs
```diff
@@ -45,9 +45,9 @@ pub fn run_move_compiler(
             .and_then(|f| {
                 Path::new(f)
                     .file_name()
-                    .map(|f| f.to_string_lossy().to_string())
+                    .map(|f| f.to_string_lossy().as_ref().to_owned())
```
brmataptos, line 48: Note that you don't need .as_ref() here.

brmataptos (commented):
Very minor nits.

third_party/move/move-compiler-v2/src/lib.rs
wrwg: > Note that you don't need .as_ref() here.

No I do. That is because to_string_lossy returns not `&str` but some cow string.

brmataptos: Cow<T> implements Fn to_owned(&self) -> T directly.

https://doc.rust-lang.org/std/borrow/enum.Cow.html

third_party/move/move-compiler-v2/src/bytecode_generator.rs
wrwg: Sorry not sure why I should add basic documentation about the language -- which is already encoded in the bytecode format as it does not support any other cast operation -- here.

brmataptos:
Reasonable.

third_party/move/move-compiler-v2/src/lib.rs
wrwg: Yeah but the result is not a string it is a `Cow<str>` ...

wrwg: BTW, if that `as_ref` would be unnecessary, cargo clippy would tell.

third_party/move/move-compiler-v2/src/bytecode_generator.rs
brmataptos: Fine.  It wouldn't be terrible, but it's not necessary.

third_party/move/move-compiler-v2/src/lib.rs
brmataptos: Oop, you're right.  Sorry.

third_party/move/move-compiler-v2/src/bytecode_generator.rs
brmataptos, line 1: You now don't need VecDeque use on line 18.

brmataptos (approved):
Looks ok, just the extra use.

wrwg:
> Looks ok, just the extra use.

You don't need to care about details like this, the linter will not allow me to merge.

PR #8773, opened 2023-06-21
Aptos gas payer

gerben-stavenga:
### Description
Implement alternate gas payer support to aptos-blockchain. This pr reuses the already existing multi-agent framework, as a separate gas payer "is" a special case of a multi-agent transaction. However the existing multi-agent framework requires
a entry function that takes all [sender, secondary_signer...] as the first parameters, a gas payer is typically not a parameter to an entry function, but just a signature to pay for the tx.

This PR uses the MSB (high bit) of the sequence number as a boolean to indicate a gas payer in case of a multi-agent tx.
If this is set the last entry in the list of secondary signer is the "gas payer" and does not participate as signer in the 
tx parameters. This is safe as 64bits is large enough for the high bit never to be reached.


aptos-move/aptos-release-builder/src/components/feature_flags.rs
```diff
@@ -39,7 +39,11 @@ pub enum FeatureFlag {
     SignatureCheckerV2,
     StorageSlotMetadata,
     ChargeInvariantViolation,
+<<<<<<< HEAD
```
movekevin: Unresolved merge conflict

aptos-move/aptos-vm/src/aptos_vm.rs
```diff
@@ -394,6 +394,15 @@ impl AptosVM {
         )?)
     }
 
+    fn get_senders(txn_data: &TransactionMetadata) -> Vec<AccountAddress> {
+        let mut res = vec![txn_data.sender];
+        res.extend(txn_data.secondary_signers());
+        if txn_data.sequence_number & (1u64 << 63) != 0 {
```
movekevin, line 400: What is this doing?

aptos-move/aptos-vm/src/aptos_vm_impl.rs
```diff
@@ -502,28 +505,19 @@ impl AptosVMImpl {
...
-
-        let transaction_validation = self.transaction_validation();
+        transaction_validation: &TransactionValidation,
+    ) -> VMResult<()> {
         let txn_sequence_number = txn_data.sequence_number();
         let txn_gas_price = txn_data.gas_unit_price();
         let txn_max_gas_units = txn_data.max_gas_amount();
-        session
-            .execute_function_bypass_visibility(
+        if txn_sequence_number & (1u64 << 63) == 0 {
```
movekevin, line 518: I don’t really get what this is doing at all

aptos-move/aptos-vm/src/aptos_vm.rs
gerben-stavenga: Checking if it's a gas payer tx

aptos-move/aptos-vm/src/aptos_vm_impl.rs
gerben-stavenga: I thought we'd agree on this in the AIP

aptos-move/framework/move-stdlib/doc/features.md
```diff
@@ -987,13 +1002,23 @@ Lifetime: transient
 
 </details>
 
+<<<<<<< HEAD
```
bowenyang007, line 1005: this looks like a merge conflict.

aptos-move/aptos-vm/src/aptos_vm.rs
bowenyang007: Why do it this way vs a boolean?

gerben-stavenga: Adding a boolean would require a whole new format

aptos-move/aptos-vm/src/aptos_vm.rs
```diff
@@ -394,6 +394,15 @@ impl AptosVM {
         )?)
     }
 
+    fn get_senders(txn_data: &TransactionMetadata) -> Vec<AccountAddress> {
```
movekevin, line 397: Shouldn't this check the feature flag and not pop if it's disabled even if GAS_PAYER_FLAG_BIT is set?

aptos-move/e2e-move-tests/src/tests/scripts.rs
```diff
@@ -80,3 +81,74 @@ fn read_coin(h: &MoveHarness, account: &AccountAddress) -> u64 {
         .unwrap()
         .coin()
 }
+
+#[test]
+fn test_two_to_two_transfer_gas_payer() {
```
movekevin, line 86: Can you add some more tests around cases such as:
1. Feature flag is not enabled.
2. Gas payer doesn't have enough gas
3. Extra senders without gas payer bit set.
4. etc. Any other edge cases? Let's be as thorough as possible here to make sure there are no unforeseen issues

aptos-move/aptos-vm/src/aptos_vm_impl.rs
```diff
@@ -537,8 +532,50 @@ impl AptosVMImpl {
...
-            .map_err(expect_no_verification_errors)
+        } else {
+            // Gas payer tx
+            let gas_payer = *txn_data.secondary_signers.last().ok_or_else(|| {
+                PartialVMError::new(StatusCode::UNKNOWN_INVARIANT_VIOLATION_ERROR)
+                    .finish(Location::Undefined)
+            })?;
+            session.execute_function_bypass_visibility(
+                &transaction_validation.module_id(),
+                &transaction_validation.user_epilogue_gas_payer_name,
```
movekevin, line 543: Can we run the same epilogue function but with gas payer set to sender? This would allow having a single epilogue function instead of 2

aptos-move/aptos-vm/src/aptos_vm.rs
gerben-stavenga: We can never get here as the prologue would reject the tx due to inconsistent sequence number

aptos-move/aptos-vm/src/aptos_vm_impl.rs
gerben-stavenga: No replay and forking requires to call the old epilogue function

aptos-move/aptos-vm/src/aptos_vm.rs
```diff
@@ -394,6 +394,15 @@ impl AptosVM {
...
+        let mut res = vec![txn_data.sender];
+        res.extend(txn_data.secondary_signers());
+        if txn_data.sequence_number & GAS_PAYER_FLAG_BIT != 0 {
```
wrwg, line 400: Please add a comment why this pop is happening.

aptos-move/aptos-vm/src/aptos_vm_impl.rs
zekun000: it needs to be different if i understand correctly, it bumps the sequence number of sender but charges gas of the payer so that becomes two identity instead of one

wrwg (approved):
This looks good to me from the code perspective, wo/ understanding the design goals. More tests would be good.

zekun000 (commented):
lgtm

aptos-move/aptos-vm/src/aptos_vm_impl.rs
movekevin: May be just to be safe but breaking older binaries with new framework compat is not uncommon. Every gas schedule upgrade breaks compat with older binaries

aptos-move/aptos-vm/src/aptos_vm.rs
movekevin: For the current code may be (need to double check) but this is potentially foot gunning if someone calls this as part of prologue

aptos-move/aptos-vm/src/aptos_vm.rs
movekevin: +1 on leaving enough context behind so we dont accidentally break things in the future

aptos-move/aptos-vm/src/aptos_vm_impl.rs
gerben-stavenga: But the issue is not breaking newer binaries with older framework. Ie. new binaries must continue to call epilogue function without a gas_payer, because that's the only one existing in old framework. 

Anyway given that the old epilogue is implemented in terms of the newer epilogue, I don't think it's a big deal.

movekevin: Yeah makes sense

aptos-move/framework/aptos-framework/sources/transaction_validation.move
```diff
@@ -148,7 +153,14 @@ module aptos_framework::transaction_validation {
         txn_expiration_time: u64,
         chain_id: u8,
     ) {
-        prologue_common(sender, txn_sequence_number, txn_sender_public_key, txn_gas_price, txn_max_gas_units, txn_expiration_time, chain_id);
+        let gas_payer = signer::address_of(&sender);
+        if ((txn_sequence_number & GAS_PAYER_FLAG_BIT) != 0) {
+            assert!(features::gas_payer_enabled(), PROLOGUE_ESEQUENCE_NUMBER_TOO_BIG);
+            gas_payer = *std::vector::borrow(&secondary_signer_addresses, std::vector::length(&secondary_signer_addresses) - 1);
```
movekevin, line 159: It might be cleaner to also call a separate multi_agent_script_prologue_gas_payer here to minimize the parts of the code that are aware of this separate gas payer logic. In your current implementation, two places need to know:
1. The VM due to the gas payer bit validation + separate epilogue call 
2. This Move code that does prologue
It's both less risk (of mismatching changes) and easier to update in the future when we want to extend gas payer functionality

aptos-move/aptos-vm/src/errors.rs
```diff
@@ -83,10 +83,10 @@ pub fn convert_prologue_error(
                 },
                 (category, reason) => {
                     speculative_error!(
-			log_context,
```
movekevin, line 86: Automatic IDE formatting?

aptos-move/aptos-vm/src/aptos_vm_impl.rs
```diff
@@ -502,28 +506,19 @@ impl AptosVMImpl {
...
-
-        let transaction_validation = self.transaction_validation();
+        transaction_validation: &TransactionValidation,
+    ) -> VMResult<()> {
         let txn_sequence_number = txn_data.sequence_number();
         let txn_gas_price = txn_data.gas_unit_price();
         let txn_max_gas_units = txn_data.max_gas_amount();
-        session
-            .execute_function_bypass_visibility(
+        if txn_sequence_number & GAS_PAYER_FLAG_BIT == 0 {
```
movekevin, line 519: Can you add comments here to explain that this doesn't need a feature flag guarding because prologue should already check so a tx would have failed there if they try to use the feature before it's enabled?

aptos-move/aptos-vm/src/aptos_vm_impl.rs
```diff
@@ -537,8 +532,50 @@ impl AptosVMImpl {
                 ]),
                 &mut UnmeteredGasMeter,
             )
-            .map(|_return_vals| ())
-            .map_err(expect_no_verification_errors)
+        } else {
+            // Gas payer tx
+            let gas_payer = *txn_data.secondary_signers.last().ok_or_else(|| {
+                PartialVMError::new(StatusCode::UNKNOWN_INVARIANT_VIOLATION_ERROR)
```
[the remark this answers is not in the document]
gerben-stavenga: No it's invariant error, as we shouldn't even get to the epilogue if the gas payer account is missing. Was tempted to do unwrap but decided invariant error is better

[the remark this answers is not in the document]
gerben-stavenga: I added a proper abort code in the multi_agent prologue to verify this

aptos-move/framework/aptos-framework/sources/transaction_validation.move
gerben-stavenga: I don't know what you mean minimize. I think this by far minimizes the amount of logic

movekevin: It does spread the logic across the VM and Move though. Would it be cleaner to just do it only in the VM and pass the appropriate args here into Move?

aptos-move/aptos-vm/src/aptos_vm_impl.rs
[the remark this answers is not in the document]
movekevin: We're being a bit strict and trying to avoid invariant violation as much as possible. cc @runtian-zhou regarding whether this is a legit case

aptos-move/aptos-vm/src/errors.rs
gerben-stavenga: I thinks so

aptos-move/framework/aptos-framework/sources/transaction_validation.move
```diff
@@ -148,7 +153,14 @@ module aptos_framework::transaction_validation {
...
+            // Clear the high bit as it's not part of the sequence number
+            txn_sequence_number = txn_sequence_number & (GAS_PAYER_FLAG_BIT - 1);
```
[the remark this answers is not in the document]
gerben-stavenga: Ideally most of the logic would be in move and not in vm. We try to make as much as possible move only. In this case the only real logic is in move, only for validation we should not include error on signer count mismatch due to extraneous gas payer signer at the end.

aptos-move/framework/aptos-framework/sources/transaction_validation.move
gerben-stavenga: Like my comment. I thought that the typical goal is to move as much as possible of logic into "move". In principle the VM doesnt need to know about gas payer at all. The senders thing is mostly for transaction arguments construction, if we don't demand that #senders == #function signer parameters, it's not needed

aptos-move/e2e-move-tests/src/tests/scripts.rs
```diff
@@ -80,3 +81,181 @@ fn read_coin(h: &MoveHarness, account: &AccountAddress) -> u64 {
...
+
+    let build_options = aptos_framework::BuildOptions {
+        with_srcs: false,
+        with_abis: false,
+        with_source_maps: false,
+        with_error_map: false,
+        ..aptos_framework::BuildOptions::default()
+    };
+
+    let package = aptos_framework::BuiltPackage::build(
```
movekevin, line 178: nit: This looks duplicate. May be refactor this part of constructing a transfer tx with a gas payer out into a separate function?

aptos-move/e2e-move-tests/src/tests/scripts.rs
```diff
@@ -80,3 +81,181 @@ fn read_coin(h: &MoveHarness, account: &AccountAddress) -> u64 {
...
+    let script = Script::new(code, vec![], vec![
+        TransactionArgument::U64(amount_alice),
+        TransactionArgument::U64(amount_bob),
+        TransactionArgument::Address(*carol.address()),
+        TransactionArgument::Address(*david.address()),
+        TransactionArgument::U64(amount_carol),
+    ]);
+
+    let transaction = TransactionBuilder::new(alice.clone())
+        .secondary_signers(vec![bob.clone(), payer.clone()])
```
movekevin, line 130: Let's add a quick case in this same test for when the secondary signers list only contains the gas payer (effectively a normal tx with a single sender but separate gas payer)

aptos-move/e2e-move-tests/src/tests/scripts.rs
movekevin: Still missing some failure cases:
1. Gas payer doesn't have sufficient balances.
2. Invalid gas payer signature
3. Etc.
I'll leave it to you to determine what can be covered with unit vs move e2e tests. We'll also have a more concise integration test later in TS as well for a true e2e.

movekevin (approved):
Approval pending more test coverage to ensure there are no unforeseen edge cases

aptos-move/e2e-move-tests/src/tests/scripts.rs
gerben-stavenga: Given that this code is just using multi-agent signing and reusing the epilogue. Most of these flows are tested by regular tests

gerben-stavenga: I have tests that cover (scripts/entry functions)/(no signer/single signer/multiple signers)/(gas bit/no-gas bit)(feature/no feature)

aptos-move/e2e-move-tests/src/tests/scripts.rs
gerben-stavenga: I added normal entry function for this

movekevin (approved):

PR #8777, opened 2023-06-21
fix support of sharded execution in executor benchmark

zjma:
### Description

This change introduces
- partitioning stage in front of the execution stage in executor benchmark, by default synced with execution stage.
- pipeline option `--async-partitioning` to promote the partitioning stage to be fully async.

### Test plan
Manual testing commands:
```
cargo run -p aptos-executor-benchmark -- --block-size 2000 --concurrency-level 8 --split-ledger-db --use-sharded-state-merkle-db create-db --data-dir /tmp/db2 --num-accounts 100000

cargo run -p aptos-executor-benchmark -- --generate-then-execute --transactions-per-sender 1 --block-size 100 --num-executor-shards 3 --async-partitioning 1 --split-ledger-db --use-sharded-state-merkle-db run-executor --transaction-type non-conflicting-coin-transfer --main-signer-accounts 200 --additional-dst-pool-accounts 300 --data-dir /tmp/db2 --checkpoint-dir /tmp/ck2 --blocks 99
```


aptos-move/aptos-vm/src/block_executor/mod.rs
```diff
@@ -242,7 +250,7 @@ impl BlockAptosVM {
                 // Flush the speculative logs of the committed transactions.
                 let pos = output_vec.partition_point(|o| !o.status().is_retry());
 
-                if state_view.id() != StateViewId::Miscellaneous {
+                if !is_sharded_execution && state_view.id() != StateViewId::Miscellaneous {
```
zjma, line 253: @gelash @sitalkedia here it seems we have to not touch the speculative logging global data structure at all, including not initializing and not flushing.

execution/executor-benchmark/src/main.rs
```diff
@@ -117,9 +123,6 @@ struct Opt {
     #[clap(long)]
     concurrency_level: Option<usize>,
 
-    #[clap(long, default_value = "1")]
-    num_executor_shards: usize,
```
zjma, line 121: this is moved under `PipelineOpt`, should be fine, right?

execution/executor-benchmark/src/pipeline.rs
```diff
@@ -152,4 +197,33 @@ where
...
+            .push(TransactionWithDependencies::new(
+                last_txn,
+                CrossShardDependencies::default(),
+            ));
+        ExecutableBlock::new(block_id, ExecutableTransactions::Sharded(sub_blocks))
+    }
+}
+
+/// Message from partitioning thread to execution thread.
+pub struct ParToExeMsg {
```
sitalkedia, line 226: nit: I would rename this to `ExecuteBlockMsg`

execution/executor-benchmark/src/block_partitioning.rs
```diff
@@ -0,0 +1,87 @@
...
+    ) -> Self {
+        let maybe_partitioner = if num_shards <= 1 {
+            None
+        } else {
+            let partitioner = ShardedBlockPartitioner::new(num_shards);
+            Some(partitioner)
+        };
+
+        Self {
+            num_iterations: 0,
```
sitalkedia, line 41: Is this block number? If so, let's rename it to `block_num`

execution/executor-benchmark/src/block_partitioning.rs
```diff
@@ -0,0 +1,87 @@
...
+            self.num_iterations,
+            txns.len()
+        );
+        let block_id = HashValue::random();
+        let block: ExecutableBlock<Transaction> = match &self.maybe_partitioner {
+            None => (block_id, txns).into(),
+            Some(partitioner) => {
+                let last_txn = txns.pop().unwrap();
+                assert!(matches!(last_txn, Transaction::StateCheckpoint(_)));
+                let analyzed_transactions = txns.into_iter().map(|t| t.into()).collect();
```
sitalkedia, line 61: Hmm.. may be we need to do this in parallel otherwise parsing txn to convert to analyzed txn is going to be to slow..

execution/executor-benchmark/src/block_partitioning.rs
```diff
@@ -0,0 +1,87 @@
...
+                ExecutableBlock::new(block_id, ExecutableTransactions::Sharded(sub_blocks))
+            },
+        };
+        let msg = ParToExeMsg {
+            current_block_start_time,
+            block,
+        };
+        self.executable_block_sender.send(msg).unwrap();
+        if let Some(rx) = &self.maybe_exe_fin_receiver {
+            rx.recv().unwrap();
```
sitalkedia, line 83: Can you explain why is this needed?

execution/executor-benchmark/src/pipeline.rs
```diff
@@ -152,4 +197,33 @@ where
             handle.join().unwrap()
         }
     }
+
+    pub fn partition_block(
```
sitalkedia, line 201: Seems like this function is not being called?

execution/executor-benchmark/src/transaction_executor.rs
```diff
@@ -135,5 +139,11 @@ where
                 .commit_blocks(vec![block_id], ledger_info_with_sigs)
                 .unwrap();
         }
+        self.parent_block_id = block_id;
+
+        if let Some(tx) = &self.maybe_exe_fin_sender {
+            // We are in partition-then-execute mode. Notify the partitioning stage to pick up the next batch.
```
sitalkedia, line 145: Hmm this is a bit convoluted, instead of forcing serialization across multiple threads using this approach, can you think of a cleaner approach? May be if async partitioning is disabled, in the same thread, we first call `partition_block` and then `execute_block`. If it is enabled, then we spawn two separate threads for them?

zjma: @sitalkedia your example sounds like maintaining 2 pipeline architectures. But I had the impression that the existing pattern is to keep 1 pipeline architecture and use channels to fine-grained control the behaviors (e.g. [here](https://github.com/aptos-labs/aptos-core/blob/65f979aa0bf788ecd84e02399e1a70b7608d1bc9/execution/executor-benchmark/src/pipeline.rs#L65)).
@msmouse @grao1991 your thoughts?

IMO adding a new pipeline stage and an optional sync channel is anyway simpler than keeping 2 pipeline architectures. Maybe I just didn't document well?

execution/executor-benchmark/src/pipeline.rs
zjma: fixed

execution/executor-benchmark/src/block_partitioning.rs
zjma: fixed

execution/executor-benchmark/src/block_partitioning.rs
zjma: added comment.

execution/executor-benchmark/src/pipeline.rs
zjma: fixed

execution/executor-benchmark/src/transaction_executor.rs
sitalkedia: @zjma - Sorry I didn't make it clearer. The current pipeline architecture is fine, you don't need to change that (except for removing the fin channel). Unless I am missing something, we can get rid of the fin channel and simplify this by doing something like this - 


```

fn partition_block () {

let mut partitioning_stage = BlockPartitioningStage::new(
                    executable_block_sender,
                    maybe_exe_fin_receiver,
                    num_partitioner_shards,
                );
                while let Ok(txns) = raw_block_receiver.recv() {
                    partitioning_stage.process(txns);
                }
            }
   }
   
   fn execute_block(). -> {
   
   start_execution_rx.map(|rx| rx.recv());
                let mut exe = TransactionExecutor::new(
                    executor_1,
                    parent_block_id,
                    version,
                    Some(commit_sender),
                    config.allow_discards,
                    config.allow_aborts,
                    maybe_exe_fin_sender,
                );
                let start_time = Instant::now();
                let mut executed = 0;
                let start_gas = TXN_GAS_USAGE.get_sample_sum();
                while let Ok(msg) = executable_block_receiver.recv() {
                    let ExecuteBlockMessage {
                        current_block_start_time,
                        partition_time,
                        block,
                    } = msg;
                    let block_size = block.transactions.num_transactions();
                    info!("Received block of size {:?} to execute", block_size);
                    executed += block_size;
                    exe.execute_block(current_block_start_time, partition_time, block);
                    info!("Finished executing block");
                }
                
          }
          
          
          If async_partitioning {
          
             thread.spawn(partioning_block).. 
             thread.spawn(execute_block)...
           } else {
              thread.spawn( {
                partition_block()
                execute_block()
            }
```

zjma: fixed

execution/executor-benchmark/src/block_partitioning.rs
zjma: this is far from being the bottleneck. Fine for now?

aptos-move/aptos-vm/src/block_executor/mod.rs
```diff
@@ -212,13 +212,21 @@ impl BlockAptosVM {
...
         let num_txns = signature_verified_block.num_txns();
-        if state_view.id() != StateViewId::Miscellaneous {
+        if !is_sharded_execution && state_view.id() != StateViewId::Miscellaneous {
             // Speculation is disabled in Miscellaneous context, which is used by testing and
             // can even lead to concurrent execute_block invocations, leading to errors on flush.
             init_speculative_logs(num_txns);
         }
 
+        if is_sharded_execution {
+            aptos_vm_logging::disable_speculative_logging();
```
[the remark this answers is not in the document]
zjma: fixed

execution/executor-benchmark/src/pipeline.rs
```diff
@@ -76,46 +88,115 @@ where
...
+        let mut exe = TransactionExecutor::new(
+            executor_1,
+            parent_block_id,
+            version,
+            Some(commit_sender),
+            config.allow_discards,
+            config.allow_aborts,
+        );
+
+        if config.async_partitioning {
```
sitalkedia, line 104: There is a lot of code duplication here, can you look into refactoring the common code into functions so that we can reduce the duplication?

sitalkedia (approved):
LGTM % minor refactoring suggestion.

grao1991 (approved):

execution/executor-benchmark/src/pipeline.rs
zjma: That's why I implemented it [this way](https://github.com/aptos-labs/aptos-core/blob/56ce342d0906a1a5d7875ed10bbc89cbb9c75fed/execution/executor-benchmark/src/pipeline.rs#L101-L178) at the beginning...

zjma: Hmm I actually don't know a good way to dedup the current code without killing the readability. I will leave it as-is for now and merge. Let's discuss further and make a separate change if needed.

sitalkedia: IIUC, you should be able to avoid duplication by refactoring the partitioning and execution inside a thread to separate functions. 

```
fn partitioning_block () {

while let Ok(txns) = raw_block_receiver.recv() {
                        let exe_block_msg = partitioning_stage.process(txns);
                        executable_block_sender.send(exe_block_msg).unwrap();
                        
  }

fn execute_block() {

start_execution_rx.map(|rx| rx.recv());
                    let start_time = Instant::now();
                    let mut executed = 0;
                    let start_gas = TXN_GAS_USAGE.get_sample_sum();
                    while let Ok(msg) = executable_block_receiver.recv() {
                        let ExecuteBlockMessage {
                            current_block_start_time,
                            partition_time,
                            block,
                        } = msg;
                        let block_size = block.transactions.num_transactions();
                        info!("Received block of size {:?} to execute", block_size);
                        executed += block_size;
                        exe.execute_block(current_block_start_time, partition_time, block);
                        info!("Finished executing block");
                        
    }
    
    ```

zjma: I don't think these 2 functions can't be used in sync-partition mode: they each are looping.

zjma: @sitalkedia  this is the what I came up following your idea.
https://github.com/aptos-labs/aptos-core/commit/2dcd59f0dfaa92f062818f9b5cd8bb8c5adc01d4

sitalkedia: LGTM - thanks for taking care of it.

PR #8781, opened 2023-06-21
[Forge] Add basic PFN performance test.

JoshLind:
### Description

This PR adds support for two basic PFN enabled forge tests: (i) a max load test (to measure throughput and latency under load); and (ii) a const TPS test (to measure latency when the system is not saturated).

The PR also adds a new workflow for specialized forge tests (i.e., tests that don't need to run daily, like the stable and unstable forge tests). This is currently set to run every other day, but I might reduce it to bi-weekly in the future.

Note: I also plan to migrate a number of other forge tests to the new workflow to help reduce job frequencies 😄 

### Test Plan
Manual verification using the `exp` script.

### DIFF
--- a/.github/workflows/forge-specialized.yaml
+++ b/.github/workflows/forge-specialized.yaml
@@ -1,0 +1,140 @@
+# Continuously run specialized forge tests against the latest main branch.
+# These tests run less frequently than regular forge tests (i.e., stable
+# and unstable).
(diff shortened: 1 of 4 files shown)


perryjrandall (approved):
yuge

rustielin (approved):
Very nice!!

PR #8783, opened 2023-06-21
[dag] reliable broadcast receiver

ibalajiarun:
### Description

This PR adds the trait and structs for the receiver-side of the reliable broadcast. 

### Test Plan

added unit tests

### DIFF
--- a/consensus/src/dag/dag_fetcher.rs
+++ b/consensus/src/dag/dag_fetcher.rs
@@ -1,60 +1,23 @@
 // Copyright © Aptos Foundation
 // SPDX-License-Identifier: Apache-2.0
 
-use crate::dag::{
-    dag_store::Dag,
-    types::{CertifiedNode, DAGMessage, DAGNetworkSender, Node, NodeMetadata},
+use crate::{
+    dag::{
+        dag_network::DAGNetworkSender,
+        dag_store::Dag,
+        types::{CertifiedNode, DAGMessage, FetchRequest, FetchResponse, Node},
+    },
+    network::TConsensusMsg,
 };
-use aptos_consensus_types::common::{Author, Round};
+use aptos_consensus_types::common::Author;
 use aptos_infallible::RwLock;
 use aptos_logger::error;
-use aptos_types::{epoch_state::EpochState, validator_verifier::ValidatorVerifier};
-use serde::{Deserialize, Serialize};
+use aptos_types::epoch_state::EpochState;
 use std::{sync::Arc, time::Duration};
 use tokio::sync::{
     mpsc::{Receiver, Sender},
     oneshot,
 };
-
-/// Represents a request to fetch missing dependencies for `target`, `start_round` represents
-/// the first round we care about in the DAG, `exists_bitmask` is a two dimensional bitmask represents
-/// if a node exist at [start_round + index][validator_index].
-#[derive(Serialize, Deserialize, Clone)]
-struct FetchRequest {
-    target: NodeMetadata,
-    start_round: Round,
-    exists_bitmask: Vec<Vec<bool>>,
-}
-
-/// Represents a response to FetchRequest, `certified_nodes` are indexed by [round][validator_index]
-/// It should fill in gaps from the `exists_bitmask` according to the parents from the `target_digest` node.
-#[derive(Serialize, Deserialize, Clone)]
-struct FetchResponse {
-    epoch: u64,
-    certifies_nodes: Vec<Vec<CertifiedNode>>,
-}
-
-impl FetchResponse {
-    pub fn verify(
-        self,
-        _request: &FetchRequest,
-        _validator_verifier: &ValidatorVerifier,
-    ) -> anyhow::Result<Self> {
-        todo!("verification");
-    }
-}
-
-impl DAGMessage for FetchRequest {
-    fn epoch(&self) -> u64 {
-        self.target.epoch()
-    }
-}
-
-impl DAGMessage for FetchResponse {
-    fn epoch(&self) -> u64 {
-        self.epoch
-    }
-}
 
 enum FetchCallback {
     Node(Node, oneshot::Sender<Node>),
@@ -110,17 +73,18 @@ impl DagFetcher {
         while let Some((request, callback)) = self.request_rx.recv().await {
             let responders =
                 callback.responders(&self.epoch_state.verifier.get_ordered_account_addresses());
-            let network_request = request.clone().into_network_message();
+            let network_request = DAGMessage::from(request.clone()).into_network_message();
             if let Ok(response) = self
                 .network
                 .send_rpc_with_fallbacks(responders, network_request, Duration::from_secs(1))
                 .await
-                .and_then(FetchResponse::from_network_message)
+                .and_then(DAGMessage::try_from)
+                .and_then(FetchResponse::try_from)
                 .and_then(|response| response.verify(&request, &self.epoch_state.verifier))
             {
                 // TODO: support chunk response or fallback to state sync
                 let mut dag_writer = self.dag.write();
-                for rounds in response.certifies_nodes {
+                for rounds in response.certified_nodes() {
                     for node in rounds {
                         if let Err(e) = dag_writer.add_node(node) {
                             error!("Failed to add node {}", e);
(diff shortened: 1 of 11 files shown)


consensus/src/dag/types.rs
```diff
@@ -298,3 +290,172 @@ pub struct DAGNetworkMessage {
     #[serde(with = "serde_bytes")]
     pub data: Vec<u8>,
 }
+
+#[derive(Clone, Serialize, Deserialize, Debug)]
+pub enum DAGMessage {
```
zekun000, line 295: so many boilerplate 🤦 but I don't see a simpler way to do it better, maybe we can do a procmacro but not today

consensus/src/network.rs
```diff
@@ -39,11 +40,30 @@ use futures::{
     stream::{select, select_all},
     SinkExt, Stream, StreamExt,
 };
+use serde::{de::DeserializeOwned, Serialize};
 use std::{
     mem::{discriminant, Discriminant},
     time::Duration,
 };
 
+pub trait ConsensusMessageTrait: Sized + Clone + Serialize + DeserializeOwned {
```
zekun000, line 49: i personally prefer shorter names, there's conventions in other languages to name it like TConsensusMessage, TDagMessage

consensus/src/dag/dag_network.rs
```diff
@@ -0,0 +1,21 @@
...
+
+    fn process(&mut self, message: Self::Message) -> anyhow::Result<Self::Ack>;
+}
+
+#[async_trait]
+pub trait DAGNetworkSender: Send + Sync {
+    async fn send_rpc(
+        &self,
+        receiver: Author,
+        message: DAGMessage,
```
zekun000, line 18: does it need to be DAGMessage? since we may want to move it to general consensus

consensus/src/dag/reliable_broadcast.rs
```diff
@@ -78,3 +71,46 @@ impl ReliableBroadcast {
         }
     }
 }
+
+pub struct NodeBroadcastHandler {
+    signatures_by_round_peer: BTreeMap<Round, BTreeMap<Author, NodeDigestSignature>>,
```
zekun000, line 76: do we need to store the Node here?

sasha8 (commented):
A few comments.

consensus/src/dag/types.rs
```diff
@@ -83,12 +94,48 @@ impl NodeMetadata {
     }
 }
 
+#[derive(Clone, Serialize, Deserialize, CryptoHasher, Debug)]
+pub struct NodeCertificateMetadata {
```
zekun000, line 98: I'd just call it NodeCertificate, like quorum certificate, it can contain some metadata without metadata in the name 😄

consensus/src/dag/reliable_broadcast.rs
```diff
@@ -67,3 +84,108 @@ impl ReliableBroadcast {
...
+            signer,
+            verifier,
+        }
+    }
+
+    pub fn gc_before_round(&mut self, min_round: Round) {
+        self.signatures_by_round_peer.retain(|r, _| r >= &min_round);
+    }
+
+    fn validate(&self, node: &Node) -> anyhow::Result<()> {
```
zekun000, line 123: we should add a `fn verify(self, verifier: &ValidatorVerifier) -> Result<Self>` to TDagMessage for all self contained verification

consensus/src/dag/reliable_broadcast.rs
```diff
@@ -67,3 +84,108 @@ impl ReliableBroadcast {
...
+    type Request = Node;
+    type Response = NodeDigestSignature;
+
+    fn process(&mut self, node: Self::Request) -> anyhow::Result<Self::Response> {
+        self.validate(&node)?;
+
+        let signatures_by_peer = self
+            .signatures_by_round_peer
+            .entry(node.metadata().round())
+            .or_insert(BTreeMap::new());
```
zekun000, line 179: need a TODO to persist data

consensus/src/dag/dag_handler.rs
```diff
@@ -0,0 +1,65 @@
...
+                    }
+                }
+            }
+        }
+    }
+
+    async fn process_rpc(&mut self, rpc_request: IncomingDAGRequest) -> anyhow::Result<()> {
+        let dag_message: DAGMessage = ConsensusMessageTrait::from_network_message(rpc_request.req)?;
+        let response: anyhow::Result<DAGMessage> = match dag_message {
+            DAGMessage::NodeMsg(node) => self.node_receiver.process(node).map(|r| r.into()),
```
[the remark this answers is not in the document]
ibalajiarun: I will follow up in the next PR.

ibalajiarun:
I updated the PR. It's getting too long, so I will defer additional validation tests to the next PR.

consensus/src/dag/dag_handler.rs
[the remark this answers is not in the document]
ibalajiarun: I will spin it into a separate task after adding additional more handlers.

ibalajiarun:
Okay, I have added the validation unit tests too.

consensus/src/dag/dag_fetcher.rs
```diff
@@ -110,19 +73,20 @@ impl DagFetcher {
...
                 .await
-                .and_then(FetchResponse::from_network_message)
+                .and_then(DAGMessage::try_from)
+                .and_then(FetchResponse::try_from)
                 .and_then(|response| response.verify(&request, &self.epoch_state.verifier))
             {
                 // TODO: support chunk response or fallback to state sync
                 let mut dag_writer = self.dag.write();
-                for rounds in response.certifies_nodes {
+                for rounds in response.certified_nodes() {
```
zekun000, line 87: nit: we should just consume this to avoid clone

consensus/src/dag/reliable_broadcast.rs
```diff
@@ -67,3 +80,118 @@ impl ReliableBroadcast {
...
+            .filter(|parent| !dag_reader.exists(parent.metadata().digest()))
+            .cloned()
+            .collect();
+        if !missing_parents.is_empty() {
+            // For each missing parent, verify their signatures and voting power
+            ensure!(
+                missing_parents.iter().all(|parent| {
+                    let node_digest = NodeDigest::new(*parent.metadata().digest());
+                    self.verifier
+                        .verify_multi_signatures(&node_digest, parent.signatures())
```
zekun000, line 150: I think this checks voting power inside, I'd slightly prefer we do all the self-contained checks (including parents certificate check) for node outside but let's do that in another pr.

consensus/src/dag/types.rs
```diff
@@ -164,38 +184,86 @@ impl Node {
...
+        let prev_round = current_round - 1;
+        // check if the parents' round is the node's round - 1
+        ensure!(
+            self.parents()
+                .iter()
+                .all(|parent| parent.metadata().round() == prev_round),
+            "invalid parent round"
+        );
+
+        ensure!(
```
zekun000, line 214: need to handle round 0?

zekun000 (approved):

consensus/src/dag/types.rs
ibalajiarun: it is a special case I am handling in `reliable_broadcast.rs`

msmouse (approved):
I'm blind.

consensus/src/dag/reliable_broadcast.rs
```diff
@@ -67,3 +80,107 @@ impl ReliableBroadcast {
...
+        ensure!(
+            prev_round >= dag_reader.lowest_round(),
+            NodeBroadcastHandleError::MissingParents
+        );
+
+        // check which parents are missing in the DAG
+        let missing_parents: Vec<NodeCertificate> = node
+            .parents()
+            .iter()
+            .filter(|parent| !dag_reader.exists(parent.metadata().digest()))
```
sasha8, line 141: We need to compare metadata.

sasha8 (commented):
One TODO we discussed offline.

consensus/src/dag/reliable_broadcast.rs
zekun000: instead of this, I think we should calculate the digest either in signing or verification

sasha8: Balaji was saying we don't have enough info to compute parents digests.

PR #8787, opened 2023-06-21
[storage] remove unecessary plus 1

areshand:
### Description

This plus 1 is no longer needed. Previously was introduced to take care of special logic in db restore

### Test Plan

existing tests


storage/aptosdb/src/state_store/mod.rs
```diff
@@ -473,7 +473,7 @@ impl StateStore {
         // The latest snapshot can be at exactly num_transactions or before it.
```
msmouse, line 473: remove

msmouse (approved):

grao1991 (approved):

lightmark (approved):

Issue #5051, opened 2022-10-14
[APT-13] [Feature Request] Add a -dev flag to aptos move compile


Turbo-Gus:
# 🚀 Feature Request

Add a -dev flag to the `aptos move compile` command, so that users can compile move code using dev dependencies.

## Motivation

Many users want to integrate with external code (such as oracles) in order to build Dapps. Frequent devnet and testnet releases make it difficult to rely on external code existing in a given environment. Additionally, due to the early state of the ecosystem, integrating with multiple codebases can lead to issues with dependency resolution. Forking dependencies in order to ensure compatibility and reliability is a great solution to these problems, but if a developer compiles their code using forked versions of dependencies (e.g. in order to test on devnet), when they deploy on mainnet (i.e. relying on non-forked versions of oracles) there will be module compatibility issues.

I believe giving developers more options for devnet testing and controlling dependencies will be a huge positive in the long run.

## Pitch

I'd be happy to open a pull request closing this issue if the Aptos core team agrees this feature is useful.

<sub>[APT-13](https://linear.app/aptoslabs/issue/APT-13/[feature-request]-add-a-dev-flag-to-aptos-move-compile)</sub>

davidiw:
Appreciate the offer, could you elaborate on what the offer is? Maybe a concrete example would be helpful here.

Turbo-Gus:
Currently the command `aptos move compile` compiles move code using dependencies specified under `[dependencies]` in the Move.toml.  I'd like a way of compiling my code that includes dependencies specified with `[dev-dependencies]` in the Move.toml (without resorting to `aptos move test, because I ultimately want to publish this version compiled with dev-dependencies).

I think the easiest way to implement this is to add a flag to `aptos move compile`. I envision the implementation would ultimately set `build_config.dev_mode = true` in the underlying move_cli like so: https://github.com/move-language/move/blob/main/language/tools/move-cli/src/base/test.rs#L157

I'd probably change the Aptos code here to accomplish this: https://github.com/aptos-labs/aptos-core/blob/154e7b169eb44df56dfe59874546a50f1e0c1d27/crates/aptos/src/move_tool/mod.rs#L272

davidiw:
why would you want to include the dev dependencies?

Turbo-Gus:
[...]

PR #8789, opened 2023-06-22
[cli] Add dev address mode for compiler

gregnazario:
### Description
This allows --dev to be added to use the [dev-addresses] as well as updates that really confusing error message to match

### Test Plan
CI, and will test manually in a bit

https://github.com/aptos-labs/aptos-core/issues/8707

### DIFF
--- a/aptos-move/framework/src/built_package.rs
+++ b/aptos-move/framework/src/built_package.rs
@@ -39,6 +39,9 @@ pub const UPGRADE_POLICY_CUSTOM_FIELD: &str = "upgrade_policy";
 /// Represents a set of options for building artifacts from Move.
 #[derive(Debug, Clone, Parser, Serialize, Deserialize)]
 pub struct BuildOptions {
+    /// Dev mode to use [dev-addresses]
+    #[clap(long)]
+    pub dev: bool,
     #[clap(long)]
     pub with_srcs: bool,
     #[clap(long)]
@@ -67,6 +70,7 @@ pub struct BuildOptions {
 impl Default for BuildOptions {
     fn default() -> Self {
         Self {
+            dev: false,
             with_srcs: false,
             with_abis: false,
             with_source_maps: false,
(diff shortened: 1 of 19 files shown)


aptos-move/framework/src/built_package.rs
```diff
@@ -39,6 +39,9 @@ pub const UPGRADE_POLICY_CUSTOM_FIELD: &str = "upgrade_policy";
 /// Represents a set of options for building artifacts from Move.
 #[derive(Debug, Clone, Parser, Serialize, Deserialize)]
 pub struct BuildOptions {
+    /// Dev mode to use [dev-addresses]
```
banool, line 42: This doesn't really explain what this flag does, can you explain this in way more detail? My assumption is when you set this, it makes the compiler take into account dev-addresses?

banool (commented):
I don't really get the point of this flag, could you explain? Like aren't dev-addresses only for useful for tests?

Also we should get in the practice of writing tests for new features like this, could you please add something to `crates/aptos/e2e`?

movekevin (commented):
This is awesome! Thanks for fixing this. Can we test this on the unit tests in https://github.com/aptos-labs/aptos-core/blob/resource_accounts/aptos-move/move-examples/tests/move_unit_tests.rs#L195 as well ?

aptos-move/framework/src/built_package.rs
gregnazario: `dev-addresses` are never used unless you set this flag, and it's always been set to false on the compiler.

gregnazario:
> This is awesome! Thanks for fixing this. Can we test this on the unit tests in https://github.com/aptos-labs/aptos-core/blob/resource_accounts/aptos-move/move-examples/tests/move_unit_tests.rs#L195 as well ?

Yeah, I was hoping adding this should let us use these as well

wrwg:
Is this also addressing  #5051?

gregnazario:
> Is this also addressing #5051?

Yeah, I believe this should also address this

wrwg (approved):

banool (dismissed):
Please update the comment for the flag to explain how it works with and without the flag.

banool (changes requested):
Just to be explicit that I'm waiting for the comment to be updated hahah.

gregnazario:
> Just to be explicit that I'm waiting for the comment to be updated hahah.

I did update the comment

banool (approved):

brmataptos:
Seems to fix https://github.com/aptos-labs/aptos-core/issues/5051

PR #8794, opened 2023-06-22
[Dev setup] Better handle nightly rustup command failures.

JoshLind:
Related: https://github.com/rust-lang/rustup/issues/3390

### Description

This PR adds a fallback to the `dev_setup.sh` script to allow it to handle `cargo nightly` installation failures. Specifically, if nightly fails to install, we'll fallback to a specific nightly build. Note: the fallback is a bit brittle in that it won't work for non-linux machines. But, given that the fallback is primarily to help unblock CI/CD in such situations, this is okay for the time being.

Once this lands (and we get feedback on the upstream issues), we can improve this.

### Test Plan
Manual verification.

### DIFF
--- a/scripts/dev_setup.sh
+++ b/scripts/dev_setup.sh
@@ -415,6 +415,11 @@ function install_toolchain {
 }
 
 function install_rustup_components_and_nightly {
+    echo "Printing the rustup version and toolchain list"
+    rustup --version
+    rustup show
+    rustup toolchain list -v
+
     echo "Updating rustup and installing rustfmt & clippy"
     rustup update
     rustup component add rustfmt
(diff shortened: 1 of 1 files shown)


scripts/dev_setup.sh
```diff
@@ -415,19 +415,29 @@ function install_toolchain {
...
 
     # We require nightly for strict rust formatting
     echo "Installing the nightly toolchain and rustfmt nightly"
-
-    # TODO: make the nightly error handling more robust
     if ! rustup toolchain install nightly
     then
-      echo "Failed to install the nightly toolchain using rustup."
+      # TODO: remove this once we have an answer: https://github.com/rust-lang/rustup/issues/3390
+      echo "Failed to install the nightly toolchain using rustup! Falling back to an older build..."
```
0xmaayan, line 433: what happens if the older build fails also? is this a known working build?

JoshLind: This really shouldn't happen (i.e., the build has already succeeded and been published online, so it shouldn't flake). If it does happen, then I think we have a much larger problem, e.g., `rustup` has stopped working altogether. In which case, we want things to fail so we can be notified immediately 😱

scripts/dev_setup.sh
```diff
@@ -415,19 +415,29 @@ function install_toolchain {
...
     if ! rustup toolchain install nightly
     then
-      echo "Failed to install the nightly toolchain using rustup."
+      # TODO: remove this once we have an answer: https://github.com/rust-lang/rustup/issues/3390
+      echo "Failed to install the nightly toolchain using rustup! Falling back to an older build at 2023-06-01."
+      rustup toolchain install nightly-2023-06-01 # Fix the date to avoid flakiness
+
+      # Rename the toolchain to nightly (crazy... see: https://github.com/rust-lang/rustup/issues/1299).
+      # Note: this only works for linux. The primary purpose is to unblock CI/CD on flakes.
+      mv ~/.rustup/toolchains/nightly-2023-06-01-x86_64-unknown-linux-gnu ~/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu
```
gregnazario, line 438: Can we detect that this is in Linux and throw an appropriate error if not first?

0xmaayan (approved):

bchocho (approved):

PR #8797, opened 2023-06-22
[storage] add cron job for backup compaction

areshand:
### Test Plan

locally test the command successfully finished

`aptos-db-tool backup-maintenance compact --state-snapshot-file-compact-factor 100 --transaction-file-compact-factor 100 --epoch-ending-file-compact-factor 100 --command-adapter-config ~/Downloads/s3-public.yaml
`

### DIFF
--- a/terraform/fullnode/gcp/cluster.tf
+++ b/terraform/fullnode/gcp/cluster.tf
@@ -7,6 +7,7 @@ resource "google_container_cluster" "aptos" {
   lifecycle {
     ignore_changes = [
       private_cluster_config,
+      cluster_autoscaling[0].auto_provisioning_defaults[0].shielded_instance_config
     ]
     prevent_destroy = true
   }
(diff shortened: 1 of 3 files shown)


perryjrandall (approved):
Test plan?

Is there a way to create this from a cli instead of a helm template?

areshand:
> Test plan?
> 
> Is there a way to create this from a cli instead of a helm template?
we locally tested the command
we plan to apply the change to GCP testnet first and confirm the deployment works as expected before rolling out.

perryjrandall:
> > Test plan?
> > 
> > Is there a way to create this from a cli instead of a helm template?
> we locally tested the command
> we plan to apply the change to GCP testnet first and confirm the deployment works as expected before rolling out.

Can you please include the command you're running to test this locally?

perryjrandall:
> > Test plan?
> > Is there a way to create this from a cli instead of a helm template?
> 
> we locally tested the command we plan to apply the change to GCP testnet first and confirm the deployment works as expected before rolling out.
> 
> This is a new chrons job. would be great if there is a way to do it through cli.

I see so you're not tested on testnet yet, best case for this is to helm apply to cluster

In future, its actually very easy to write csript to generate the script, then kubectl apply that

rustielin:
I usually do some e2e test like deploy to a local (KinD) cluster. You can also do a template dry-run locally like this:

```
# pwd: terraform/helm/fullnode
$ helm template -s templates/backup-compaction.yaml . --debug 
```

rustielin (approved):

PR #8799, opened 2023-06-22
[aptos-vm] Accumulate errors 

runtian-zhou:
### Description

This is a followup for #8749. `VMError` now has a sub status field that could be set and accumulated. The AptosVM can the nuse the sub status there to determine whether a critical error need to be logged down or not.

### Test Plan

The transaction that triggered invariant violation error will no longer be triggered with this extra sub status


runtian-zhou:
The next steps here would be to add our own sub status code to the errors that we created so we can customize our logging logic there.

movekevin (commented):
Introducing a sub status just to clarify an error instead of using the correct error code seems a bit hacky here. Is it because it's very hard to clean up the misuses?

vgao1996 (approved):
LGTM. Let's fix the namespace issue I discussed with you offline, but that doesn't need to happen now.

runtian-zhou:
> Introducing a sub status just to clarify an error instead of using the correct error code seems a bit hacky here. Is it because it's very hard to clean up the misuses?

Yea so the status code is part of the ledger so changing anything there would potentially lead to breakage and we only need those code for better alerting so I believe adding another substatus seem to be the easier thing to do here.

brmataptos (approved):

PR #8801, opened 2023-06-22
[Gas Estimation] enable

bchocho:
### Description

Enable gas estimation as default, so the updated estimation is deployed in prod on the next release.


JoshLind (approved):
Noice!

banool (approved):

PR #8802, opened 2023-06-22
[Storage] Add a flag to skip writing indices and usage which will eventually go away.

grao1991:
### Description
Add a flag to skip indices and storage usage, which will eventually be removed from core AptosDB. This will help us get a better idea in benchmark on how much TPS we can reach later.


areshand:
could you have some context on the purpose of skipping indices and usage?

areshand (approved):

msmouse (approved):

PR #8805, opened 2023-06-22
Release builder support custom aptos-core path

xindingw:
### Description
- By providing arg `--aptos-core-path`, release builder can read `aptos-framework` and scripts to compile from a custom `aptos-core` path. If the arg is not provided, release builder will use the path itself was compiled at.
- Fix a bug. When `--test-dir` is specified, `validate-proposals` will fail because release builder is reading compiled proposals in a wrong path.
### Test Plan

Local test.
Rename `aptos-core` to `aptos-core-r` then run binary `validate-proposals`.
`./aptos-release-builder --aptos-core-path <path to aptos-core> validate-proposals --release-config <path to release.yaml> --output-dir /tmp/release_output --endpoint http://0.0.0.0:8080 --mint-to-validator from-directory --test-dir <path to .aptos/testnet>`

### DIFF
--- a/aptos-move/aptos-release-builder/src/lib.rs
+++ b/aptos-move/aptos-release-builder/src/lib.rs
@@ -4,8 +4,13 @@
 pub mod components;
 mod utils;
 pub mod validate;
+
 pub use components::{ExecutionMode, ReleaseConfig, ReleaseEntry};
-use once_cell::sync::Lazy;
+use once_cell::sync::{Lazy, OnceCell};
+use std::{
+    env,
+    path::{Path, PathBuf},
+};
 
 // Update me after branch cut.
 const RELEASE_CONFIG: &str = include_str!("../data/release.yaml");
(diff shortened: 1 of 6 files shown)


perryjrandall (approved):
Looks great, I'd recommend making this an option in the release builder as well as using environment variables is kind of cumbersome UX

Also for testing id recommend using this to upgrade a local network as well to make sure the validate path works as well

runtian-zhou:
Yea I would prefere being able to pass this in as argument because env vars are so error prone. It is likely that the previous run can set an env var that you forgot to clean up and could thus lead to unexepcted result. I would recommend only being able to set it via the argument IMO.

aptos-move/aptos-release-builder/src/lib.rs
```diff
@@ -17,3 +22,33 @@ static CURRENT_RELEASE_CONFIG: Lazy<ReleaseConfig> =
 pub fn current_release_config() -> &'static ReleaseConfig {
     &CURRENT_RELEASE_CONFIG
 }
+
+static APTOS_CORE_PATH: OnceCell<PathBuf> = OnceCell::new();
+
+pub fn initialize_aptos_core_path(overriden_path: Option<PathBuf>) {
```
perryjrandall, line 28: This works, but it seems a little strange to me that we would use a global here instead of just passing the argument where we're using it since most of the code changed is our own code and we should be able to modify the function adding another arg

perryjrandall (approved):

perryjrandall:
Also can you include the exact command you ran in your test plan?

That way others looking for an invocation can learn from your PR!

aptos-move/aptos-release-builder/src/lib.rs
xindingw: I prefer a global because `--aptos-core-path` is a top-level arg of this tooling. It's widely used by all subcommand and it's supposed to be applied to all subcommands. Passing a thing you need everywhere just makes codes more verbose.

xindingw:
> Also can you include the exact command you ran in your test plan?
> 
> That way others looking for an invocation can learn from your PR!

Added!

xindingw:
> Yea I would prefere being able to pass this in as argument because env vars are so error prone. It is likely that the previous run can set an env var that you forgot to clean up and could thus lead to unexepcted result. I would recommend only being able to set it via the argument IMO.

Refactored!

runtian-zhou (approved):

alokeutpal (approved):

PR #8811, opened 2023-06-23
[dag] basic dag fetcher

zekun000:
Skeleton of supporting fetching missing dependencies based on local dag information.

### DIFF
--- a/consensus/src/dag/dag_fetcher.rs
+++ b/consensus/src/dag/dag_fetcher.rs
@@ -1,0 +1,134 @@
+// Copyright © Aptos Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::dag::{
+    dag_store::Dag,
+    types::{CertifiedNode, DAGMessage, DAGNetworkSender, Node, NodeMetadata},
+};
+use aptos_consensus_types::common::{Author, Round};
+use aptos_infallible::RwLock;
+use aptos_logger::error;
+use aptos_types::{epoch_state::EpochState, validator_verifier::ValidatorVerifier};
+use serde::{Deserialize, Serialize};
+use std::{sync::Arc, time::Duration};
+use tokio::sync::{
+    mpsc::{Receiver, Sender},
+    oneshot,
+};
+
+/// Represents a request to fetch missing dependencies for `target`, `start_round` represents
+/// the first round we care about in the DAG, `exists_bitmask` is a two dimensional bitmask represents
+/// if a node exist at [start_round + index][validator_index].
+#[derive(Serialize, Deserialize, Clone)]
+struct FetchRequest {
+    target: NodeMetadata,
+    start_round: Round,
+    exists_bitmask: Vec<Vec<bool>>,
+}
+
+/// Represents a response to FetchRequest, `certified_nodes` are indexed by [round][validator_index]
+/// It should fill in gaps from the `exists_bitmask` according to the parents from the `target_digest` node.
+#[derive(Serialize, Deserialize, Clone)]
+struct FetchResponse {
+    epoch: u64,
+    certifies_nodes: Vec<Vec<CertifiedNode>>,
+}
+
+impl FetchResponse {
+    pub fn verify(
+        self,
+        _request: &FetchRequest,
+        _validator_verifier: &ValidatorVerifier,
+    ) -> anyhow::Result<Self> {
+        todo!("verification");
+    }
+}
+
+impl DAGMessage for FetchRequest {
+    fn epoch(&self) -> u64 {
+        self.target.epoch()
+    }
+}
+
+impl DAGMessage for FetchResponse {
+    fn epoch(&self) -> u64 {
+        self.epoch
+    }
+}
+
+enum FetchCallback {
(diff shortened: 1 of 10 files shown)


sasha8 (approved):

consensus/src/dag/dag_fetcher.rs
```diff
@@ -0,0 +1,100 @@
...
+    fn epoch(&self) -> u64 {
+        self.epoch
+    }
+}
+
+struct DagFetcher {
+    epoch_state: Arc<EpochState>,
+    network: Arc<dyn DAGNetworkSender>,
+    response_tx: Sender<(FetchResponse, CertifiedNode)>,
+    request_rx: Receiver<(FetchRequest, CertifiedNode)>,
```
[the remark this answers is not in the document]
zekun000: fetch request/response is sent over the network, certified node is used locally

consensus/src/dag/dag_fetcher.rs
```diff
@@ -0,0 +1,100 @@
...
+        epoch_state: Arc<EpochState>,
+        network: Arc<dyn DAGNetworkSender>,
+    ) -> (
+        Self,
+        Sender<(FetchRequest, CertifiedNode)>,
+        Receiver<(FetchResponse, CertifiedNode)>,
+    ) {
+        let (request_tx, request_rx) = tokio::sync::mpsc::channel(16);
+        let (response_tx, response_rx) = tokio::sync::mpsc::channel(16);
+        (
```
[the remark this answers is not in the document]
ibalajiarun: I think we should introduce the DagConfig soon.

consensus/src/dag/types.rs
```diff
@@ -35,6 +36,23 @@ pub trait DAGMessage: Sized + Clone + Serialize + DeserializeOwned {
...
+#[async_trait]
+pub trait DAGNetworkSender: Send + Sync {
+    async fn send_rpc(
+        &self,
+        receiver: Author,
+        message: ConsensusMsg,
+        timeout: Duration,
+    ) -> anyhow::Result<ConsensusMsg>;
+
+    async fn send_rpc_with_retry(
```
ibalajiarun, line 48:
```suggestion
    async fn broadcast_fetch_rpc(
```

consensus/src/dag/dag_fetcher.rs
```diff
@@ -0,0 +1,100 @@
...
+}
+
+impl DAGMessage for FetchResponse {
+    fn epoch(&self) -> u64 {
+        self.epoch
+    }
+}
+
+struct DagFetcher {
+    epoch_state: Arc<EpochState>,
```
ibalajiarun, line 53: I only passed the verifier in broadcast receiver logic. is this better?

consensus/src/dag/dag_fetcher.rs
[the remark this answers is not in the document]
ibalajiarun: Sasha and I just discussed. We may also need to fetch `CertifiedNode` in response to a `Node` request before signing the Node. So, the request should be flexible enough to fetch in both cases.

consensus/src/dag/types.rs
zekun000: this is not a broadcast_fetch though, it's more like exponential fallback with failures

consensus/src/dag/dag_fetcher.rs
zekun000: it's just a verifier plus epoch number, I think there's many cases we need epoch numbers to create the message so it seems easier

consensus/src/dag/dag_fetcher.rs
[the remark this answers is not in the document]
zekun000: that's a good point, i didn't think about fetching dependencies for Node

consensus/src/dag/dag_fetcher.rs
```diff
@@ -0,0 +1,132 @@
...
+            request_tx,
+        )
+    }
+
+    pub async fn start(mut self) {
+        while let Some((request, callback)) = self.request_rx.recv().await {
+            let responders =
+                callback.responders(&self.epoch_state.verifier.get_ordered_account_addresses());
+            let network_request = request.clone().into_network_message();
+            if let Ok(response) = self
```
ibalajiarun, line 112: How do we tackle request deduplication? If we get two nodes in round (r) with missing nodes in r-1, those two node in round r will trigger a fetch, but I believe the fetcher should be smart enough to dedup the request otherwise in the worst case we might be sending fetch rpc for n nodes in worst case. 

also, we may need to check the dag store before sending the request, because the previous request might have already fetched the necessary nodes?

Do you want to add these as a TODO?

ibalajiarun (approved):
Approving to unblock, but left some feedback.

consensus/src/dag/dag_fetcher.rs
zekun000: what I have in mind is the fetcher is single task only, so it won't have all the concurrency issues you mentioned. Given the 2f+1 overlaps per round, the benefit of concurrent seems minimal to me. if there's node from much higher round, we'd fall into state sync instead.

PR #8812, opened 2023-06-23
Cherrypick log related fixes

runtian-zhou:
### Description

These set of PRs should help us fix the endless false positives in the Move oncall.

This is consisted of #8615 #8749 and #8799

### Test Plan

Tested the txn that causes false positive to no longer raise logs. Will run replay testing to make sure this won't incur any backward compat issues.


wrwg (approved):

runtian-zhou:
Replay succeeded with https://github.com/aptos-labs/aptos-core/actions/runs/5353552563

sherry-x (approved):

PR #8815, opened 2023-06-23
[Move-prover][Spec] Add support of generic types for exists in the prover

### LINKED ISSUE #6658: [Prover] the use of a type parameter with `exists` in spec
## How to reproduce
This case is from the `object` module in the `aptos_framework`. To reproduce, run Prover uncommenting the `ensures` line.
```
    native fun exists_at<T: key>(object: address): bool;
    spec exists_at {
        pragma opaque;
        aborts_if false;
        // TODO: Disabled the following post-condition due to an issue with
        // the use of a type parameter in `exists` in spec.
        // ensures result == exists<T>(object);
    }
```
The current version of Prover fails during the model building process, complaining that:
```
error: The type argument to `exists` and `global` must be a struct type but #0 is not a struct type.
  ┌─ ./sources/object.spec.move:7:27
  │
7 │         ensures result == exists<T>(object);
  │                           ^^^^^^^^^^^^^^^^^

Error: exiting with model building errors
```

## Expected output
Prover runs without an error.

rahxephon89:
### Description

This PR
- adds support of generic types for exists in the prover so that one can use `exists` at the spec level for the function `exists_at` in the `object` module.
- adds specs for `exists_at`.

Close #6658

### DIFF
--- a/third_party/move/move-model/src/spec_translator.rs
+++ b/third_party/move/move-model/src/spec_translator.rs
@@ -704,14 +704,13 @@ impl<'a, 'b, T: ExpGenerator<'a>> ExpRewriterFunctions for SpecTranslator<'a, 'b
                 )
                 .into_exp(),
             ),
-            Exists(None) if self.in_old => Some(
-                Call(
-                    id,
-                    Exists(Some(self.save_memory(self.builder.get_memory_of_node(id)))),
-                    args.to_owned(),
-                )
-                .into_exp(),
-            ),
+            Exists(None) if self.in_old => {
+                if let Some(mem) = self.builder.get_memory_of_node_opt(id) {
+                    Some(Call(id, Exists(Some(self.save_memory(mem))), args.to_owned()).into_exp())
+                } else {
+                    Some(Call(id, Exists(None), args.to_owned()).into_exp())
+                }
+            },
             SpecFunction(mid, fid, None) if self.in_old => {
                 let used_memory = {
                     let module_env = self.builder.global_env().get_module(*mid);
--- a/third_party/move/move-model/src/spec_translator.rs
+++ b/third_party/move/move-model/src/spec_translator.rs
@@ -21,7 +21,7 @@ use crate::{
     ty::{PrimitiveType, Type},
 };
 use codespan_reporting::diagnostic::Severity;
-use itertools::Itertools;
+use itertools::{Either, Itertools};
 use std::collections::{BTreeMap, BTreeSet};
 
 /// A helper which reduces specification conditions to assume/assert statements.
@@ -55,7 +55,7 @@ pub struct SpecTranslator<'a, 'b, T: ExpGenerator<'a>> {
 /// Represents a translated spec.
 #[derive(Default)]
 pub struct TranslatedSpec {
-    pub saved_memory: BTreeMap<QualifiedInstId<StructId>, MemoryLabel>,
+    pub saved_memory: BTreeMap<Either<QualifiedInstId<StructId>, u16>, MemoryLabel>,
     pub saved_spec_vars: BTreeMap<QualifiedInstId<SpecVarId>, MemoryLabel>,
     pub saved_params: BTreeMap<TempIndex, TempIndex>,
     pub debug_traces: Vec<(NodeId, TraceKind, Exp)>,
@@ -563,12 +563,12 @@ impl<'a, 'b, T: ExpGenerator<'a>> SpecTranslator<'a, 'b, T> {
         }
     }
 
-    fn save_memory(&mut self, qid: QualifiedInstId<StructId>) -> MemoryLabel {
+    fn save_memory_ty(&mut self, ty: Either<QualifiedInstId<StructId>, u16>) -> MemoryLabel {
         let builder = &mut self.builder;
         *self
             .result
             .saved_memory
-            .entry(qid)
+            .entry(ty)
             .or_insert_with(|| builder.global_env().new_global_id())
     }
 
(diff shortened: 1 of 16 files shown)


third_party/move/move-model/src/ty.rs
```diff
@@ -282,6 +282,19 @@ impl Type {
         }
     }
 
+    pub fn require_struct_opt(&self) -> Option<(ModuleId, StructId, &[Type])> {
+        if let Type::Struct(mid, sid, targs) = self {
+            Some((*mid, *sid, targs.as_slice()))
+        } else if let Type::TypeParameter(_) = self {
```
wrwg, line 288: I think you do not need to have this check for parameters. Just return None (no panic) if its not a struct. You can then also reduce code duplication, and define `require_struct` in terms of `require_struct_opt`.

You do not need this extra check because you already type checking whether its a parameter nor struct.

third_party/move/move-model/src/spec_translator.rs
```diff
@@ -704,14 +704,13 @@ impl<'a, 'b, T: ExpGenerator<'a>> ExpRewriterFunctions for SpecTranslator<'a, 'b
...
-                    Exists(Some(self.save_memory(self.builder.get_memory_of_node(id)))),
-                    args.to_owned(),
-                )
-                .into_exp(),
-            ),
+            Exists(None) if self.in_old => {
+                if let Some(mem) = self.builder.get_memory_of_node_opt(id) {
+                    Some(Call(id, Exists(Some(self.save_memory(mem))), args.to_owned()).into_exp())
+                } else {
+                    Some(Call(id, Exists(None), args.to_owned()).into_exp())
```
wrwg, line 711: Unfortunately, I think this isn't correct. If an `exists` is applied within `old(...exists...)`, we need to create this label to use a different memory. If we do not then the exists will be interpreted within the current memory, which is semantically wrong.

There is no obvious workaround. You could report an error like `exists with type parameter not supported within `old` expression' or similar. Do you need it within old?

third_party/move/move-prover/boogie-backend/src/spec_translator.rs
```diff
@@ -1147,18 +1147,34 @@ impl<'env> SpecTranslator<'env> {
...
-            boogie_resource_memory_name(self.env, memory, memory_label),
-        );
+        let memory_opt = &self.get_memory_inst_from_node_opt(node_id);
+        let mut resource_name = "".to_string();
+        if let Some(memory) = memory_opt {
+            resource_name = boogie_resource_memory_name(self.env, memory, memory_label);
+        } else {
+            let mem_ty = &self.get_node_instantiation(node_id)[0];
+            if let Type::TypeParameter(arg) = mem_ty {
+                resource_name = format!("#{}_$memory", arg)
```
wrwg, line 1172: Unfortunately, this wont work like this, because we do not declare any of this `#<typeparam_number>_$memory` variables. I assume your test cases did not hit this case, otherwise the boogie should have failed to compile.

The reason why you did not hit this might be monomorphization. We should try to see whether we can just always report an error also if its a type parameter, and how far we get with this.

wrwg (commented):
See comments. Can we avoid having `exists<T>` within old context? Also we cannot construct a memory boogie variable from a type parameter, but perhaps we do not need to.

third_party/move/move-prover/boogie-backend/src/spec_translator.rs
rahxephon89: > Unfortunately, this wont work like this, because we do not declare any of this `#<typeparam_number>_$memory` variables. I assume your test cases did not hit this case, otherwise the boogie should have failed to compile.
> 
> The reason why you did not hit this might be monomorphization. We should try to see whether we can just always report an error also if its a type parameter, and how far we get with this.

I see. Actually, the following code is generated in the boogie file:
```
type #0;
function {:inline} $IsEqual'#0'(x1: #0, x2: #0): bool { x1 == x2 }
function {:inline} $IsValid'#0'(x: #0): bool { true }
var #0_info: $TypeParamInfo;
var #0_$memory: $Memory #0;
```
I will dig more into this

third_party/move/move-model/src/ty.rs
rahxephon89: I added the check for parameters to handle the case such as `exists<u64>`.

third_party/move/move-prover/boogie-backend/src/spec_translator.rs
wrwg: Interesting. The boogie code above is the monomorphization of type parameter 0. The type parameter is turned into a given type (`type #0;`). I'm surprised to see we also generate a memory variable for it. I think we may do this because `#0` has the `key` ability. So we could be good. Unfortunately, it is still not clear how to deal with `old(...exists...)`, however.

third_party/move/move-model/src/spec_translator.rs
rahxephon89: The issue here is that expressions in the `aborts_if` condition are in the old context and currently the main use for having `exists<T>` is in the aborts_if condition. Do you have any suggestions?

third_party/move/move-model/src/ty.rs
wrwg: But this is already type checker, so not sure whether you need the panic here.

The least thing to do if you want to keep this function is rename it to `require_struct_or_type_param_opt`. Otherwise it is very confusing because `require_struct` already exists -- but this here is not the same.

rahxephon89: Yes, you are right. I fixed it and also refactored other functions using `_opt` to reduce code duplication.

third_party/move/move-model/src/spec_translator.rs
wrwg: It will be difficult. If you check the `Bytecode::SaveMem(label, mem)` instruction, that is what you eventually need to emit here as well. Currently `mem: QualifiedInstId<StructId>`. This would need to be generalized so it can be either a struct id or a type parameter. The generalization would need to be propagated to multiple places. I wonder whether all this is worth it.

Do you see an alternative way to implement `exists_at` just as native function?

rahxephon89: In my opinion, to implement `exists_at` natively that follows the semantics of memory, we still need to maintain a pair (label, mem) where mem is a type parameter. In that sense, it makes more sense to just extend `Bytecode::SaveMem(label, mem)` in general. I am wondering if `exists_at<T>` is used in for generic types, we can say that `old(...exists<T>)` is always equal to `exists<T>` because global memory cannot be changed if it is a generic type?

PR #8820, opened 2023-06-23
[vm-benchmarking] benchmark txn runtime in VM

wlawt:
### Description

The Aptos VM Benchmark allows anyone to specify a set of Move modules and benchmark the running time of transactions in the AptosVM measured in milliseconds. The Aptos VM Benchmark allows an arbitrary amount of packages to be benchmarked under the `/samples` directory.

### Test Plan

See the current examples under `/samples/add-numbers` and `/samples/do-nothing` to demonstrate basic utility. 

There is also support for filtering & running certain tests (i.e., similar to pattern matching with cargo test). The usage:
`cargo run --release -- PATTERN`

To run the tests, run `cargo run`

### Additional notes
- functions must be marked with `entry`
- additionally, functions then must be prefixed with `benchmark`
- finally, functions then must not have any parameters
- the time currently reflects the prologue, body and epilogue


aptos-move/aptos-vm-benchmarks/src/main.rs
```diff
@@ -0,0 +1,219 @@
...
+                    continue;
+                }
+
+                // skip if it doesn't match pattern
+                let mut fully_qualified_path = String::new();
+                fully_qualified_path.push_str(address.to_string().as_str());
+                fully_qualified_path.push_str("::");
+                fully_qualified_path.push_str(identifier);
+                fully_qualified_path.push_str("::");
+                fully_qualified_path.push_str(func_name);
```
vgao1996, line 164: You can replace this with `format!("{}::{}::{}", address, identifier, func_name)`

vgao1996 (approved):
LGTM as a skeleton for future benchmarks. I have a feeling that we may need to have better control over the caching behavior and filter out more noises for this to be truly useful, but let's worry about those later.

runtian-zhou (commented):
I believe you can use this api to directly invoke into the function you want to bench: https://github.com/aptos-labs/aptos-core/blob/73a09d15db4ad31ab128993de6b9ab796849845b/aptos-move/e2e-tests/src/executor.rs#L611

This should allow you to run function without prologue/epilogue

runtian-zhou:
Generally I would be a bit concerned to have a main file that contains too much logic. I don't mind land it as is but just for sanity I felt like splitting the logics out would look a bit better.

runtian-zhou (approved):

PR #8825, opened 2023-06-23
Add create_token method in token.move

vusirikala:
### Description
This PR adds `create_token` method in token.move file.


aptos-move/framework/aptos-token-objects/sources/token.move
```diff
@@ -106,6 +106,22 @@ module aptos_token_objects::token {
         };
     }
 
+    /// Creates a new token object with a unique address and returns the ConstructorRef
+    /// for additional specialization.
+    public fun create_token(
```
igor-aptos, line 111: Mark create_from_account as deprecated

movekevin (approved):
Can you add a unit test for the new function?

movekevin (commented):
Also can you check the docs as well to make sure we recommend the right function?

movekevin (commented):
Can you update any uses of create_named_token in move-examples that are not justified as well?

aptos-move/framework/aptos-token-objects/sources/token.move
vusirikala: Done.

vusirikala:
> Can you update any uses of create_named_token in move-examples that are not justified as well?

I've check that right now we are using `create_named_token` only in test cases.

aptos-move/framework/aptos-token-objects/sources/aptos_token.move
```diff
@@ -199,7 +199,7 @@ module aptos_token_objects::aptos_token {
         property_types: vector<String>,
         property_values: vector<vector<u8>>,
     ): ConstructorRef acquires AptosCollection {
-        let constructor_ref = token::create_from_account(
+        let constructor_ref = token::create_token(
```
movekevin, line 202: This needs to be behind the same feature flag as create_guid otherwise this would always fail before the flag is enabled

aptos-move/framework/aptos-token-objects/sources/token.move
movekevin, line 111: nit: How about just name this "create"

vusirikala: Done.

aptos-move/framework/aptos-token-objects/sources/aptos_token.move
vusirikala: If auids are enabled, I am now calling `create` method. If auids are disabled, I am calling `create_from_account` method. Is that okay?

vusirikala:
> Can you add a unit test for the new function?

I added a unit test now. But it runs only if the AUIDs are enabled.

aptos-move/framework/aptos-token-objects/sources/aptos_token.move
```diff
@@ -199,14 +199,25 @@ module aptos_token_objects::aptos_token {
...
     ): ConstructorRef acquires AptosCollection {
-        let constructor_ref = token::create_from_account(
-            creator,
-            collection,
-            description,
-            name,
-            option::none(),
-            uri,
-        );
+        let constructor_ref = if (std::features::auids_enabled()) {
```
movekevin, line 202: nit: Can you add an import for std::features and just call features::auids_enabled here to keep it consistent?

movekevin (approved):

movekevin:
> > Can you add a unit test for the new function?
> 
> I added a unit test now. But it runs only if the AUIDs are enabled.

That's fine. It should pass after merging this into the create-guid branch

aptos-move/framework/aptos-token-objects/sources/aptos_token.move
vusirikala: Done.

PR #8826, opened 2023-06-23
[sharding] include round ID in dependency info

zjma:
### Description

Also Include the round ID (in addition to shard ID) in cross-shard edges. This is for supporting 1-blockstm-per-subblock approach, where each blockstm instance needs to know which (round,shard) to send/recv necessary txn results.

### Test Plan

Existing UTs.


execution/block-partitioner/src/sharded_block_partitioner/conflict_detector.rs
```diff
@@ -51,7 +53,7 @@ impl CrossShardConflictDetector {
             } else {
                 accepted_txn_dependencies.push(self.get_deps_for_frozen_txn(
                     &txn,
-                    Arc::new(vec![]),
```
zjma, line 54: This seems a bug. `get_deps_for_frozen_txn` implementation assumes the "current round rw set" vec to be of size equal to num_shards.

zjma: Fixed it in this PR.

sitalkedia: Why is this a bug? This needs to be empty for early return condition in `get_deps_for_frozen_txn` function.

execution/block-partitioner/src/sharded_block_partitioner/conflict_detector.rs
```diff
@@ -72,9 +74,8 @@ impl CrossShardConflictDetector {
         current_round_rw_set_with_index: Arc<Vec<WriteSetWithTxnIndex>>,
         prev_rounds_rw_set_with_index: Arc<Vec<WriteSetWithTxnIndex>>,
     ) -> CrossShardDependencies {
-        if current_round_rw_set_with_index.is_empty() && prev_rounds_rw_set_with_index.is_empty() {
```
sitalkedia, line 75: Why remove this? This makes it efficient in first round to early return.

execution/block-partitioner/src/sharded_block_partitioner/partitioning_shard.rs
```diff
@@ -42,6 +43,7 @@ impl PartitioningShard {
             control_rx,
             result_tx,
             cross_shard_client,
+            num_rounds_finished: 0,
```
sitalkedia, line 46: better leave this partioning shard state less and not store round number here, can you pass it in the partitioning command itself?

types/src/block_executor/partitioner.rs
```diff
@@ -9,16 +9,18 @@ pub type ShardId = usize;
 pub type TxnIndex = usize;
 
 #[derive(Debug, Clone, Eq, Hash, PartialEq, Serialize, Deserialize)]
-pub struct TxnIdxWithShardId {
+pub struct TxnIndexForSharding {
     pub txn_index: TxnIndex,
     pub shard_id: ShardId,
+    pub round: usize,
```
sitalkedia, line 15: nit: May be add an alias type for round as well.

types/src/block_executor/partitioner.rs
```diff
@@ -9,16 +9,18 @@ pub type ShardId = usize;
 pub type TxnIndex = usize;
 
 #[derive(Debug, Clone, Eq, Hash, PartialEq, Serialize, Deserialize)]
-pub struct TxnIdxWithShardId {
+pub struct TxnIndexForSharding {
```
sitalkedia, line 12: How about `ShardedTxnIndex` instead?

sitalkedia (commented):
Can you look into adding some tests - in this PR or as a follow up is fine.

execution/block-partitioner/src/sharded_block_partitioner/conflict_detector.rs
zjma: I think [here](https://github.com/aptos-labs/aptos-core/blob/703c36243b401fdba81aaa33569b54c20290307a/execution/block-partitioner/src/sharded_block_partitioner/conflict_detector.rs#L89) is the implementation that assumes "current round rw set" vec to be of size equal to num_shards.
```
            let mut current_shard_id = (self.shard_id + self.num_shards - 1) % self.num_shards; // current shard id - 1 in a wrapping fashion
            for rw_set_with_index in current_round_rw_set_with_index
                .iter()
                .take(self.shard_id)
                .rev()
                .chain(prev_rounds_rw_set_with_index.iter().rev())
            {
                if rw_set_with_index.has_write_lock(storage_location) {
                    cross_shard_dependencies.add_required_edge(
                        TxnIdxWithShardId::new(
                            rw_set_with_index.get_write_lock_txn_index(storage_location),
                            current_shard_id,
                        ),
                        storage_location.clone(),
                    );
                    break;
                }
                // perform a wrapping substraction
                current_shard_id = (current_shard_id + self.num_shards - 1) % self.num_shards;
            }
```

Imagine when `num_shards ==3, self.shard_id == 2, self.round == 2`, and we are getting deps for txns in sub-block `I`. (Other subblocks are denoted as below.)
```
round 0: A B C 
round 1: D E F
-----
round 2: G H I
```
The correct order of sub-blocks to check is `H, G, F, E, ..., B, A`.
The cursor `current_shard_id` starts with 1 which is correct.
But if `current_round_rw_set_with_index` is empty, the first sub-block being iterated in the following for-loop will be `F`. We may end up associating `F` with shard_id 1 when doing `add_required_edge`

zjma: To fix it, we can either pass in empty `G, H, I` and keep current cursor logic, or fix the initial value for `current_shard_id` as well as `current_round` (introduced in this PR). I took the first approach but both are fine to me.

execution/block-partitioner/src/sharded_block_partitioner/conflict_detector.rs
zjma: `current_round_rw_set_with_index` won't be empty with the proposed fix (see [this](https://github.com/aptos-labs/aptos-core/pull/8826#discussion_r1241302932) thread), making it unreachable.

execution/block-partitioner/src/sharded_block_partitioner/partitioning_shard.rs
zjma: sure, but it looks to me partitioning shards are already stateful? Fields `shard_id` and the rxs/txs are all tied to a partitioning session already.

types/src/block_executor/partitioner.rs
zjma: fixed

execution/block-partitioner/src/sharded_block_partitioner/partitioning_shard.rs
zjma: fixed

types/src/block_executor/partitioner.rs
zjma: fixed

execution/block-partitioner/src/sharded_block_partitioner/conflict_detector.rs
sitalkedia: I see your point, can we still keep the early return condition though? It is crucial to make the first round performant by having this early return. May be we can add a special check for round == 0?

zjma: fixed. added the special check outside this function to save a little bit more.

sitalkedia (approved):
LGTM

zekun000 (approved):

PR #8834, opened 2023-06-24
[Sharding][Partitioning] Fix to ensure relative ordering for sender

sitalkedia:
### Description

Fix partitioning logic to maintain relative ordering of transaction across senders. This is done by moving all remaining txns from the sender to the end of the block, once we detect cross shard conflict for that sender.

### Test Plan

Added UTs.

### DIFF
--- a/execution/block-partitioner/src/sharded_block_partitioner/conflict_detector.rs
+++ b/execution/block-partitioner/src/sharded_block_partitioner/conflict_detector.rs
@@ -47,22 +47,16 @@ impl CrossShardConflictDetector {
         let mut rejected_txns = Vec::new();
         let mut discarded_senders = HashSet::new();
         for (_, txn) in txns.into_iter().enumerate() {
-            if self.check_for_cross_shard_conflict(self.shard_id, &txn, cross_shard_rw_set) {
+            let sender_was_discarded = txn
+                .sender()
+                .map_or(false, |sender| discarded_senders.contains(&sender));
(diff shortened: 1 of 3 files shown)


execution/block-partitioner/src/sharded_block_partitioner/conflict_detector.rs
```diff
@@ -45,9 +45,24 @@ impl CrossShardConflictDetector {
...
+                if discarded_senders.contains(&sender) {
+                    rejected_txns.push(txn);
+                } else {
+                    accepted_txn_dependencies.push(self.get_deps_for_frozen_txn(
+                        &txn,
+                        Arc::new(vec![]),
+                        prev_rounds_rw_set_with_index.clone(),
+                    ));
+                    accepted_txns.push(txn);
+                }
```
zjma, line 65: nit: maybe save a few lines like this:

```suggestion
            let sender_was_discarded = txn.sender().map_or(|sender| discarded_sender.contains(&sender), false);
            if sender_was_discarded || self.check_for_cross_shard_conflict(self.shard_id, &txn, cross_shard_rw_set) {
                if let Some(sender) = txn.sender() {
                    discarded_senders.insert(sender);
                }
                rejected_txns.push(txn);
```

zjma:
Also fixed in-shard cross-round dependencies?

zjma (approved):

ibalajiarun (approved):

PR #8845, opened 2023-06-26
add the file store verifier.

larry-aptos:
### Description

* Add an optional background worker to verify the file store:
  * start version matches with file name
  * transactions stored are ordered and sequential without gap.
  * transactions can be deserialized correctly. 
  * files are sequential without gap
* Persist the latest verification information in `verification.json`.

Note:
* update the tracing-subscriber due to bug: https://github.com/tokio-rs/tracing/issues/2565

### Test Plan

Locally tested; looks good overall.

image

image


ecosystem/indexer-grpc/indexer-grpc-utils/src/file_store_operator/mod.rs
```diff
@@ -42,6 +43,15 @@ pub struct FileStoreMetadata {
     pub version: u64,
 }
 
+/// FileStoreMetadata is the metadata for the file store.
+/// It's a JSON file with name: metadata.json.
+#[derive(Serialize, Deserialize, Copy, Clone, Debug)]
+pub struct VerificationMetadata {
+    pub chain_id: u64,
+    // The current version of the file store.
+    pub next_version_to_verify: u64,
```
rustielin, line 52: how fast is the verification? is it a problem that we now expect the files-store-operator process to be stateful since it now writes a file? vs previously when we just wrote to a separate file-store? Can we keep this metadata in the file store itself?

ecosystem/indexer-grpc/indexer-grpc-post-processor/src/file_storage_verifier.rs
```diff
@@ -0,0 +1,103 @@
...
+        file_store_operator.verify_storage_bucket_existence().await;
+        // Get or create verification metadata file.
+        let verification_metadata = file_store_operator
+            .get_or_create_verification_metadata(self.chain_id)
+            .await?;
+        let file_store_metadata = file_store_operator
+            .get_file_store_metadata()
+            .await
+            .ok_or(anyhow::anyhow!("File Store metadata does not exist"))?;
+        ensure!(
```
rustielin, line 50: instead of returning early with an error (and crashing/panic in the parent process), can we instead return some sort of metric? We don't want to stop running all post-processors if only one of them is failing for instance.

ecosystem/indexer-grpc/indexer-grpc-utils/src/file_store_operator/local.rs
```diff
@@ -209,4 +209,23 @@ impl FileStoreOperator for LocalFileStoreOperator {
 
         Ok(())
     }
+
+    async fn get_or_create_verification_metadata(
+        &self,
+        _chain_id: u64,
+    ) -> Result<VerificationMetadata> {
+        anyhow::bail!("Verification is not impelemented for local file store.")
```
rustielin, line 217: maybe implement this now or follow up with another PR? this would be good to test esp for the post-processor

ecosystem/indexer-grpc/indexer-grpc-utils/src/file_store_operator/gcs.rs
```diff
@@ -4,20 +4,27 @@
...
     latest_metadata_update_timestamp: Option<std::time::Instant>,
+
+    /// The timestamp of the latest verification metadata update; this is to avoid too frequent metadata update.
+    latest_verification_metadata_update_timestamp: Option<std::time::Instant>,
 }
 
 impl GcsFileStoreOperator {
-    pub fn new(bucket_name: String) -> Self {
+    pub fn new(bucket_name: String, service_account_path: String) -> Self {
+        env::set_var(SERVICE_ACCOUNT_ENV_VAR, service_account_path);
```
rustielin, line 23: where/how is this used?

ecosystem/indexer-grpc/indexer-grpc-utils/src/config.rs
```diff
@@ -8,7 +8,7 @@ use std::path::PathBuf;
 #[derive(Serialize, Deserialize, Debug, Clone)]
 pub struct GcsFileStore {
     pub gcs_file_store_bucket_name: String,
-    // TODO: consider adding credentials_path, rather than relying on default location
+    pub gcs_file_store_service_account_key_path: String,
```
rustielin, line 11: maybe make this `Option`, so we can default to the default location if unspecified?

ecosystem/indexer-grpc/indexer-grpc-post-processor/src/main.rs
```diff
@@ -2,28 +2,77 @@
...
+                    })
+                }
+            }));
+        }
+
+        let res = futures::future::join_all(tasks).await;
+
+        for r in res {
+            if let Err(e) = r {
+                error!("Task finished with error: {:?}", e);
```
rustielin, line 70: do we really want all checks to stop running once a single one fails?

rustielin (changes requested):

ecosystem/indexer-grpc/indexer-grpc-utils/src/file_store_operator/mod.rs
larry-aptos: 1. it's like several thousands tps.
2. file processor is a stateful since we have `metadata.json` currently. 
3. yes, it looks like this now:

image

rustielin: ah I missed the part where you're reusing the file store operator, so the verification state is also persisted into the file store. ignore!

ecosystem/indexer-grpc/indexer-grpc-post-processor/src/file_storage_verifier.rs
larry-aptos: ah good call. done. also the job retries up to 3 times and then panic.

ecosystem/indexer-grpc/indexer-grpc-utils/src/file_store_operator/local.rs
larry-aptos: sure!

ecosystem/indexer-grpc/indexer-grpc-utils/src/file_store_operator/gcs.rs
larry-aptos: currently, we implicitly require `SERVICE_ACCOUNT_ENV_VAR` as env var for cloud storage. this is to surface it up in config to unify all configuration into one place.

ecosystem/indexer-grpc/indexer-grpc-utils/src/config.rs
larry-aptos: the problem is, if SERVER_ACCOUNT is not set, there is no default value and the component will panic.. Let's force users to set the service account if they're using gcs.

ecosystem/indexer-grpc/indexer-grpc-post-processor/src/main.rs
larry-aptos: the only concern for now is the verifier; added built-in retry as a short-term solution. In the long run, we should retry at framework level and add metrics if error happens.

larry-aptos: fixed now; yea for now we'll not crash the system.

ecosystem/indexer-grpc/indexer-grpc-utils/src/config.rs
rustielin: I see ok!

ecosystem/indexer-grpc/indexer-grpc-utils/src/file_store_operator/gcs.rs
rustielin: I see. So it's a requirement in GCP?

larry-aptos: yea, and it's the `cloud-storage` requirement. i didn't see it exposes other ways to config the identity.

ecosystem/indexer-grpc/indexer-grpc-post-processor/src/main.rs
```diff
@@ -2,28 +2,79 @@
...
+            tasks.push(tokio::spawn({
+                let config = config.clone();
+                async move {
+                    let public_fullnode_address = config.public_fullnode_address.clone();
+                    loop {
+                        let checker = PfnLedgerChecker::new(public_fullnode_address.clone());
+                        info!("Starting PfnLedgerChecker");
+                        if let Err(err) = checker.run().await {
+                            tracing::error!("PfnLedgerChecker failed: {:?}", err);
+                            TASK_FAILURE_COUNT
```
rustielin, line 50: nice!

ecosystem/indexer-grpc/indexer-grpc-post-processor/src/file_storage_verifier.rs
rustielin: looks good!

rustielin (approved):
Looks good! Thanks for addressing the feedback :) 

Just make sure to follow up with the local file store operator, so we can test it in the existing e2e checks.

ecosystem/indexer-grpc/indexer-grpc-utils/src/file_store_operator/gcs.rs
rustielin: maybe leave some comment there then? it's quite hard to identity lol

larry-aptos: done

ecosystem/indexer-grpc/indexer-grpc-post-processor/src/file_storage_verifier.rs
```diff
@@ -0,0 +1,117 @@
...
+            file_store_metadata.chain_id == self.chain_id,
+            "Chain ID mismatch"
+        );
+        let mut next_version_to_verify = verification_metadata.next_version_to_verify;
+        let mut current_head_version = file_store_metadata.version;
+
+        loop {
+            if next_version_to_verify > current_head_version {
```
rtso, line 60: Since `next_version_to_verify` gets incremented in line 111, won't this condition sometimes be true?

rtso (approved):

ecosystem/indexer-grpc/indexer-grpc-post-processor/src/file_storage_verifier.rs
larry-aptos: good call; change it to `next_version_to_store` from `current_head_version`.

i should improve my skill of naming things 🤦

PR #8849, opened 2023-06-26
[storage] remove genesis in fast sync node

areshand:
### Description
The goal is to completely eliminate genesis from the DB in the fast-synced node.This change is the first step ONLY aims to solve the immediate issues of having genesis in the fast-synced DB. The next step of improvement should revise the interface between different components as mentioned below 

There are 2 steps to accomplish this change:
[IN this PR] Using a new struct for storing genesis DB and another DB for storing the fast-synced. The temporary DB and main DB is wrapped in a new struct that can be used both for bootstrapping and storing the actual data. This new struct providing all data interfaces and hide underly individual DB and selection logic from different components. 

[TODO] The next step is to restructure the interface between different components to rely on this new struct for status sync, mempool, event notification bootstrapping. Once the bootstrapping is done, we then create the actual underlying DB 
https://github.com/aptos-labs/aptos-core/pull/8849

By the way, Backup Handler currently depends on individual ledger, txn, and event stores. I've hidden these individual stores from the Backup Handler and exposed these underlying details through functions in the DB itself."

### Test Plan
Existing test


storage/aptosdb/src/backup/backup_data_accessor.rs
```diff
@@ -0,0 +1,93 @@
+// Copyright © Aptos Foundation
+// Parts of the project are originally copyright © Meta Platforms, Inc.
```
JoshLind, line 2: nit: This is a new file, so we don't need this in the license 😄

storage/aptosdb/src/backup/backup_data_accessor.rs
```diff
@@ -0,0 +1,93 @@
...
+    state_store::{state_key::StateKey, state_value::StateValue},
+    transaction::{Transaction, TransactionInfo, Version},
+    write_set::WriteSet,
+};
+use std::sync::Arc;
+
+/// `BackupDataAccessor` provides functionalities for data backup.
+///  This is supposed to be replace backup handler.
+#[derive(Clone)]
+pub struct BackupDataAccessor {
```
JoshLind, line 21: Silly question: why do we need this? If all calls go directly to the underlying wrapper DB, can we not just pass the wrapper to whatever objects require this?

I likely lack context here though 😄

storage/backup/backup-service/src/handlers/mod.rs
```diff
@@ -129,3 +130,118 @@ pub(crate) fn get_routes(backup_handler: BackupHandler) -> BoxedFilter<(impl Rep
         }))
         .boxed()
 }
+
+pub(crate) fn get_routes_with_backup_accessor(
```
JoshLind, line 134: Any context for these changes? They seem slightly orthogonal to me, e.g., different PR? 🤔

storage/backup/backup-service/src/lib.rs
```diff
@@ -4,12 +4,38 @@
...
+    backup::backup_data_accessor::BackupDataAccessor, fast_sync_aptos_db::FastSyncStorageWrapper,
+    AptosDB,
+};
 use aptos_logger::prelude::*;
 use std::{net::SocketAddr, sync::Arc};
 use tokio::runtime::Runtime;
 
+pub fn start_backup_service_with_fast_sync_wrapper(
+    address: SocketAddr,
+    db: Arc<FastSyncStorageWrapper>,
```
JoshLind, line 18: Silly question: I wonder if there's a way to avoid having to duplicate this, and instead pass in an interface, such that `start_backup_service` below can take in either the `FastSyncStorageWrapper` or the `AptosDB` implementation? Otherwise, it seems like we're doing a lot of duplication, no?

storage/storage-interface/src/lib.rs
```diff
@@ -638,31 +641,89 @@ pub trait DbWriter: Send + Sync {
...
+    STARTED,
+    FINISHED,
+}
+
 #[derive(Clone)]
 pub struct DbReaderWriter {
     pub reader: Arc<dyn DbReader>,
     pub writer: Arc<dyn DbWriter>,
+    // This is updated during StateSync bootstrapping.
+    pub fast_sync_status: Arc<RwLock<FastSyncStatus>>,
```
JoshLind, line 656: Silly question: why can we not place this in the `FastSyncStorageWrapper` implementation itself? Putting it here feels off to me, i.e., we're leaking implementation details. Theoretically, `reader` and `writer` both take in interfaces, which should be `FastSyncStorageWrapper` underneath, right?

storage/aptosdb/src/fast_sync_aptos_db.rs
```diff
@@ -0,0 +1,1110 @@
+// Copyright © Aptos Foundation
+// Parts of the project are originally copyright © Meta Platforms, Inc.
```
JoshLind, line 2: nit: This is a new file, so we don't need this in the license 😄

storage/storage-interface/src/lib.rs
JoshLind: Actually, I see that it's already in there. Why do we need it here? 🤔

storage/aptosdb/src/fast_sync_aptos_db.rs
```diff
@@ -0,0 +1,1110 @@
...
+};
+use move_core_types::account_address::AccountAddress;
+use std::sync::{Arc, RwLock};
+
+const SECONDARY_DB_DIR: &str = "fast_sync_secondary";
+
+/// This is a wrapper around [AptosDB] that is used to bootstrap the node for fast sync mode
+pub struct FastSyncStorageWrapper {
+    // main is used for normal read/write or genesis data during fast sync
+    db_main: AptosDB,
```
JoshLind, line 59: Interesting... some thoughts:

1. Nit: I'd likely call these something more obvious, e.g.: `temporary_db_with_genesis` and `real_db_for_fast_sync`?
2. I probably lack context, but do these types need to be `AptosDB`? Why can they not be `DBReaderWriter`? Is this because of the backup system?
3. If we only use this wrapper in the context of fast sync, you can avoid the `Option<AptosDB>` (for the second DB) and clean up the unwraps below?
4. I don't really know why we need `FastSyncStatus`? Could we not just have a `fast_sync_complete: AtomicBool`? This would avoid the need for grabbing a RWLock for all reads and writes, and simplify the switches (i.e., the `FastSyncStatus::UNKNOWN` and `FastSyncStatus::STARTED` doesn't mean much, right? All we need to know is when it finishes, so we can direct all calls to `real_db_for_fast_sync`?) Does that make sense?

storage/aptosdb/src/fast_sync_aptos_db.rs
```diff
@@ -0,0 +1,1110 @@
...
+    }
+}
+
+impl DbReader for FastSyncStorageWrapper {
+    fn get_epoch_ending_ledger_infos(
+        &self,
+        start_epoch: u64,
+        end_epoch: u64,
+    ) -> Result<EpochChangeProof> {
+        if self.is_fast_sync_bootstrap_finished() {
```
JoshLind, line 435: Eventually, I could imagine us defining a macro for this (if we wanted to clean things up -- especially since all the switch statements are identical 😄 )

JoshLind (commented):
Thanks @areshand! Looks like a solid first pass. 😄  Most of my feedback is about code structure, abstractions and simplifications. Let me know if anything doesn't make sense.

state-sync/state-sync-v2/state-sync-driver/src/storage_synchronizer.rs
```diff
@@ -108,6 +108,12 @@ pub trait StorageSynchronizerInterface {
     /// Finish the chunk executor at this round of state sync by releasing
     /// any in-memory resources to prevent memory leak.
     fn finish_chunk_executor(&self);
+
+    /// Inform storage the fast_sync indeed starts
+    fn notify_storage_fast_sync_starts(&mut self) -> Result<(), Error>;
```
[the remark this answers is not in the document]
areshand: The reason is the status (fast_sync_status) has to be stored in a struct (DbReaderWriter). However, get_state_snapshot_receiver and get_state_snapshot_receiver don't have access to db_rw, which doesn't have access to underlying status flag.

storage/backup/backup-service/src/lib.rs
areshand: will remove `start_backup_service` which is itself a lot of work and want to only delete it once this new function is accepted.

storage/storage-interface/src/lib.rs
areshand: DbReaderWriter is the `interface` struct, which state-sync and other components use to access storage. State-sync is not supposed to directly access underlying FastSyncStorageWrapper's fields.

storage/aptosdb/src/fast_sync_aptos_db.rs
areshand: 1. that was my previous implementation. But there is a corner case. For example, the node can be configured to do fast sync. When all its peers only have genesis data, it would actually do a continuous update of txns one by one. Storage in this case should put the genesis and following updates in the same DB.  
2. Yes, this is related to the backup system since it needs to access the underlying txn_store, event_store, etc of the DB. 
3. I don't know if that makes the logic more complicated. since you need to remember AptosDB is used in this circumstance and FastSyncDBWrapper is used in other circumstances. For now, I just make FastSyncDBWrapper used in all circumstances and people can check one place to find the code behavior.
4. as mentioned in 1. we need to know if fast sync indeed starts to determine if we should use a separate DB. the `Unknown` is used for indicating non-fast-sync cases.

storage/backup/backup-service/src/handlers/mod.rs
areshand: related. backup service needs to know the sub dbs (event, txn, state store, etc).  When we switch DB based on Fast Sync status, the backup handler service will also need to switch the DB

storage/aptosdb/src/backup/backup_data_accessor.rs
areshand: yes, we can directly use the DBWrapper for Backup Service to avoid introducing backup_data_acessor. I was just following the convention of BackupHandler

storage/aptosdb/src/fast_sync_aptos_db.rs
```diff
@@ -0,0 +1,1110 @@
...
+    pub(crate) fn get_transaction_store(&self) -> Result<&TransactionStore> {
+        if self.is_fast_sync_bootstrap_finished() {
+            Ok(&self
+                .db_secondary
+                .as_ref()
+                .expect("db_secondary is not initialized")
+                .transaction_store)
+        } else {
+            Ok(&self.db_main.transaction_store)
+        }
```
msmouse, line 132: Too much duplication. We can at least abstract this piece out as `fn get_aptosdb_db_ref() -> &AptosDB`.  I would go as far as doing a 
```
trait GetAptosDbRef {
    fn get_aptos_db_ref() -> &AptosDB;
}
```
And implement that for both the actual AptosDB and the wrapper, and then
```
impl<T: GetAptosDbRef> DbWriter for T {
...
}
```

msmouse: Or darker (but elegant) magic like https://github.com/aptos-labs/aptos-core/blob/547ad67a0126f9447ef9aaf19863a56203a3e17d/aptos-move/aptos-gas-profiling/src/profiler.rs#L51

aptos-node/src/storage.rs
```diff
@@ -1,44 +1,74 @@
...
 #[cfg(not(feature = "consensus-only-perf-test"))]
 pub(crate) fn bootstrap_db(
-    aptos_db: AptosDB,
-    backup_service_address: SocketAddr,
-) -> (Arc<AptosDB>, DbReaderWriter, Option<Runtime>) {
-    use aptos_backup_service::start_backup_service;
+    node_config: &NodeConfig,
+) -> Result<(Arc<FastSyncStorageWrapper>, DbReaderWriter, Option<Runtime>)> {
+    let fast_sync_status = Arc::new(RwLock::new(FastSyncStatus::UNKNOWN));
+    let db = FastSyncStorageWrapper::new_fast_sync_aptos_db(node_config, fast_sync_status.clone())?;
```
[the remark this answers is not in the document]
areshand: I also thought of this but didn't find a good way.  eventually, I think it is better to have all the logic inside FastSyncStorageWrapper instead of exposing the complexity of selecting different DB inside aptos-node. Eg: the fast sync storage wrapper will decide not to create the secondary DB if the config is not turned on. But happy to know if there is a better way to do this.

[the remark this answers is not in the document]
areshand: I also thought of this but didn't find a good way.  eventually, I think it is better to have all the logic inside FastSyncStorageWrapper instead of exposing the complexity of selecting different DB inside aptos-node. Eg: the fast sync storage wrapper will decide not to create the secondary DB if the config is not turned on. But happy to know if there is a better way to do this.

storage/aptosdb/src/fast_sync_aptos_db.rs
JoshLind: 1. I'd need to look at the state sync code again, but if that's the case, I think we should change state sync such that it will always fast sync (i.e., if it starts up in fast sync mode, and we haven't had a single epoch change, state sync will just loop until the first epoch change occurs). Otherwise, this feels a bit janky to me.
3 + 4. Not really following you -- will sync offline 😄

storage/storage-interface/src/lib.rs
JoshLind: Still not following -- will sync offline 😄

storage/aptosdb/src/fast_sync_aptos_db.rs
areshand: updated based on offline discussion. The test will be correct once the state sync change is landed.

aptos-node/src/storage.rs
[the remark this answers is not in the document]
areshand: updated based on offline discussion

storage/storage-interface/src/lib.rs
```diff
@@ -638,6 +638,13 @@ pub trait DbWriter: Send + Sync {
     }
 }
 
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum FastSyncStatus {
```
JoshLind, line 642: nit: can this live in the fast sync storage wrapper file?

storage/aptosdb/src/fast_sync_aptos_db.rs
```diff
@@ -0,0 +1,563 @@
...
+    ($(fn $name:ident(&self $(, $arg: ident : $ty: ty $(,)?)*) -> $return_type:ty,)+) => {
+        $(fn $name(&self, $($arg: $ty),*) -> $return_type {
+            self.get_aptos_db_read_ref().$name($($arg),*)
+        })+
+    };
+}
+
+impl DbReader for FastSyncStorageWrapper {
+    aptos_db_read_fn!(
+        fn get_transactions(
```
JoshLind, line 389: nit: any chance we could still have empty lines between the functions to make it more readable? 😄

storage/aptosdb/src/fast_sync_aptos_db.rs
```diff
@@ -0,0 +1,563 @@
...
+    write_set::WriteSet,
+};
+use move_core_types::account_address::AccountAddress;
+use std::sync::{Arc, RwLock};
+
+const SECONDARY_DB_DIR: &str = "fast_sync_secondary";
+
+/// This is a wrapper around [AptosDB] that is used to bootstrap the node for fast sync mode
+pub struct FastSyncStorageWrapper {
+    // main is used for normal read/write or genesis data during fast sync
```
JoshLind, line 53: nit: can we update this comment? Also, do you think it makes sense to rename these now? e.g., `temporary_db_with_genesis` and `real_db_for_fast_sync`?

JoshLind (commented):
Thanks @areshand! This looks much cleaner to me 😄

I'll send out a PR to update state sync and then approve this PR! We should still wait for the storage experts to review this to make sure they're happy. But, otherwise, this LGTM.

storage/aptosdb/src/fast_sync_aptos_db.rs
areshand: updated

storage/aptosdb/src/fast_sync_aptos_db.rs
```diff
@@ -0,0 +1,618 @@
...
+            false,
+            config.storage.storage_pruner_config,
+            config.storage.rocksdb_configs,
+            config.storage.enable_indexer,
+            config.storage.buffered_state_target_items,
+            config.storage.max_num_nodes_per_lru_cache_shard,
+        )
+        .map_err(|err| anyhow!("fast sync DB failed to open {}", err))?;
+
+        // when the db is empty and config to do fast sync, we will create a second DB
```
msmouse, line 86: "configured"

aptos-node/src/storage.rs
```diff
@@ -1,44 +1,76 @@
...
+    let (aptos_db_reader, db_rw, backup_service) = if let Some(db) = aptos_db {
+        let (db_arc, db_rw) = DbReaderWriter::wrap(db);
+        let db_backup_service =
+            start_backup_service(node_config.storage.backup_service_address, db_arc.clone());
+        (db_arc as Arc<dyn DbReader>, db_rw, Some(db_backup_service))
+    } else {
+        let (db_arc, db_rw) = DbReaderWriter::wrap(
+            fast_sync_db_wrapper.expect("Fast sync db wrapper should not be None"),
+        );
+        let db_backup_service = start_backup_service_with_fast_sync_wrapper(
```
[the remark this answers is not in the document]
areshand: we already have Aptoshandle taking a backup runtime. why do we want to do this?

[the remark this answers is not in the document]
msmouse: To remove the need for the backup service to be able to handle both the db and the wrapper. Our end goal is to spin up all the components that are not necessary during bootstrapping only after the  bootstrapping is finished, right? the backup service is such a component.
(The runtime on the handle is optional, no need to change anything on that front.)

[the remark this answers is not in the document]
areshand: The FastSyncStorageWrapper is only created for fast sync. For non-fast sync, I don't have a wrapper to hold the runtime. Similarly, backup service just need to know either DB or FastSyncWrapper. However, FastSyncWrapper is currently only for Fast sync not for other non-fast sync cases. Not sure how to do this refactor.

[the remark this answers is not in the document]
msmouse: > For non-fast sync, I don't have a wrapper to hold the runtime
the aptos handle holds it in that case -- that's why I said it's convenient that it's an `Option<Runtime>`

it's a bit awkward to have to hold the runtime in different places, but you don't need to copy paste a few functions as you do today. And at high level it's the right thing to do: spin up the backup service only after bootstrapping is done.

msmouse (approved):

aptos-node/src/storage.rs
```diff
@@ -1,44 +1,78 @@
...
+                let fast_sync_db = db_arc.get_fast_sync_db();
+                let db_backup_service =
+                    start_backup_service(node_config.storage.backup_service_address, fast_sync_db);
+
+                (db_arc as Arc<dyn DbReader>, db_rw, Some(db_backup_service))
+            },
+            _ => unreachable!(),
+        };
+
+    maybe_apply_genesis(&db_rw, node_config)?;
```
msmouse, line 52: This will write to the real DB instead of the temp DB? Will that even work?

storage/aptosdb/src/fast_sync_aptos_db.rs
```diff
@@ -0,0 +1,221 @@
...
+    db_for_fast_sync: Arc<AptosDB>,
+    // This is for reading the fast_sync status to determine which db to use
+    fast_sync_status: Arc<RwLock<FastSyncStatus>>,
+}
+
+impl FastSyncStorageWrapper {
+    /// If the db is empty and configured to do fast sync, we return a FastSyncStorageWrapper
+    /// Otherwise, we returns AptosDB directly and the FastSyncStorageWrapper is None
+    pub fn initialize_dbs(config: &NodeConfig) -> Result<Either<AptosDB, Self>> {
+        // ) -> Result<(Option<AptosDB>, Option<FastSyncStorageWrapper>)> {
```
JoshLind, line 44: Remove?

storage/aptosdb/src/fast_sync_aptos_db.rs
```diff
@@ -0,0 +1,221 @@
...
+            config.storage.storage_pruner_config,
+            config.storage.rocksdb_configs,
+            config.storage.enable_indexer,
+            config.storage.buffered_state_target_items,
+            config.storage.max_num_nodes_per_lru_cache_shard,
+        )
+        .map_err(|err| anyhow!("fast sync DB failed to open {}", err))?;
+
+        // when the db is empty and configured to do fast sync, we will create a second DB
+        if config.state_sync.state_sync_driver.bootstrapping_mode
```
JoshLind, line 58: nit: you should be able to call:

```if config.state_sync.state_sync_driver.bootstrapping_mode.is_fast_sync()```

File: storage/aptosdb/src/fast_sync_aptos_db.rs [hunk 20]
```diff
@@ -0,0 +1,221 @@
...
+    StateSnapshotReceiver,
+};
+use aptos_types::{
+    epoch_change::EpochChangeProof,
+    ledger_info::LedgerInfoWithSignatures,
+    state_store::{state_key::StateKey, state_value::StateValue, ShardedStateUpdates},
+    transaction::{TransactionOutputListWithProof, TransactionToCommit, Version},
+};
+use either::Either;
+use std::sync::{Arc, RwLock};
```
JoshLind, line 19: nit: why not `use aptos_infallible::RwLock;`? You can get rid of the `expects` below.

JoshLind (approved):
Looks clean 😄 Thanks, all!

storage/aptosdb/src/fast_sync_aptos_db.rs
msmouse: :trollface:  missed this one

PR #8850, opened 2023-06-26
[ts-sdk] Allow local endpoint provider

gregnazario:
### Description
This allows local development with the TS SDK.  It isn't ideal though, because an indexer client still exists, it just won't work correctly.

### Test Plan
I've tested locally

### DIFF
--- a/ecosystem/typescript/sdk/src/utils/api-endpoints.ts
+++ b/ecosystem/typescript/sdk/src/utils/api-endpoints.ts
@@ -2,24 +2,28 @@ export const NetworkToIndexerAPI: Record<string, string> = {
   mainnet: "https://indexer.mainnet.aptoslabs.com/v1/graphql",
   testnet: "https://indexer-testnet.staging.gcp.aptosdev.com/v1/graphql",
   devnet: "https://indexer-devnet.staging.gcp.aptosdev.com/v1/graphql",
+  local: "",
 };
 
 export const NetworkToNodeAPI: Record<string, string> = {
   mainnet: "https://fullnode.mainnet.aptoslabs.com/v1",
   testnet: "https://fullnode.testnet.aptoslabs.com/v1",
   devnet: "https://fullnode.devnet.aptoslabs.com/v1",
+  local: "http://localhost:8080/v1",
 };
 
 export const NodeAPIToNetwork: Record<string, string> = {
   "https://fullnode.mainnet.aptoslabs.com/v1": "mainnet",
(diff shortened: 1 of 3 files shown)


ecosystem/typescript/sdk/src/utils/api-endpoints.ts
```diff
@@ -2,24 +2,28 @@ export const NetworkToIndexerAPI: Record<string, string> = {
...
   testnet: "https://indexer-testnet.staging.gcp.aptosdev.com/v1/graphql",
   devnet: "https://indexer-devnet.staging.gcp.aptosdev.com/v1/graphql",
+  local: "",
 };
 
 export const NetworkToNodeAPI: Record<string, string> = {
   mainnet: "https://fullnode.mainnet.aptoslabs.com/v1",
   testnet: "https://fullnode.testnet.aptoslabs.com/v1",
   devnet: "https://fullnode.devnet.aptoslabs.com/v1",
+  local: "http://localhost:8080",
```
xbtmatt, line 12: Should this be `http://localhost:8080/v1` or is it supposed to be without?

gregnazario: Yeah, I will correct it

xbtmatt (approved):

0xmaayan (commented):
nice! can we add tests to make sure it works as expected?
https://github.com/aptos-labs/aptos-core/blob/main/ecosystem/typescript/sdk/src/tests/e2e/provider.test.ts#L24

davidiw (commented):
interestingly I was able to at least run the tests without this :shrug:

gregnazario:
> interestingly I was able to at least run the tests without this :shrug: 

The tests run, but when trying to run outside the tests it doesn't.  The tests likely set the field where if it's empty and undefined it can possibly not work.

0xmaayan:
> interestingly I was able to at least run the tests without this 🤷

Yes, I think the description `This allows local development with the TS SDK` is a bit misleading, we are able to local develop but need to use the `CustomEndpoints` type, i.e `new Provider({fullnodeUrl:"<url>",indexerUrl:"<url>"}). This PR makes it easier to use local testnet with out declaring `indexerUrl`.

0xmaayan:
btw, we can remove the check here https://github.com/aptos-labs/aptos-core/pull/8850/files#diff-1ae716dab19a178b52f56cf8b1467fcbbfecb027e44320708dc7fccb3d8ed81cR115 , that would let us define only fullnode without indexer

davidiw (approved):

PR #8852, opened 2023-06-27
[do not review] execution sharding e2e

zjma:


aptos-move/block-executor/src/executor.rs
```diff
@@ -753,6 +750,10 @@ where
                     let fee_statement = output.fee_statement();
                     accumulated_fee_statement.add_fee_statement(&fee_statement);
                     self.update_sequential_txn_gas_counters(&accumulated_fee_statement);
+                    if let Some(listener) = transaction_commit_listener {
+                        output.incorporate_delta_writes(vec![]);
+                        listener.send_remote_update_for_success(idx as TxnIndex, &output);
+                    }
```
zjma, line 756: Explicitly converting VMOutput to TransactionOutput.

PR #8855, opened 2023-06-27
[dag] derive macro for enum variant conversions 

ibalajiarun:
### Description

This PR adds a derive macro that implements `From` traits for converting structs in enum to enum variants and vice-versa.

### DIFF
--- a/crates/aptos-enum-conversion-derive/tests/cases/no_fields.rs
+++ b/crates/aptos-enum-conversion-derive/tests/cases/no_fields.rs
@@ -1,0 +1,12 @@
+// Copyright © Aptos Foundation
+
+use aptos_enum_conversion_derive::EnumConversion;
+
+#[derive(EnumConversion)]
+enum Messages {
+    Test
+}
+
+fn main() {
+
+}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -660,6 +660,7 @@ dependencies = [
  "aptos-consensus-types",
  "aptos-crypto",
  "aptos-crypto-derive",
+ "aptos-enum-conversion-derive",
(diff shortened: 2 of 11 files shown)


sasha8 (approved):
Black magic. Can we add some tests?

zekun000 (changes requested):
I think the idiomatic way is to derive based on the enum instead of individual struct, I'd imagine it's like
```
#[derive(EnumMessage)]
enum DAGMessage
```
and based on the internal struct to do dispatches

ibalajiarun:
> I think the idiomatic way is to derive based on the enum instead of individual struct, I'd imagine it's like
> 
> ```
> #[derive(EnumMessage)]
> enum DAGMessage
> ```
> 
> and based on the internal struct to do dispatches

yes, this is better. The only catch is all the structs should be in the same module as the enum right?

zekun000 (approved):
I think the types just need to be in the same crate not necessarily module (not foreign types), but I could be wrong. 

another nit: the name is not very representative, I'd call it EnumConversion. the same for the crate name, it's probably better to call it aptos-enum-conversion-derive

crates/aptos-enum-conversion-derive/tests/cases/no_fields.rs
```diff
@@ -0,0 +1,12 @@
+// Copyright © Aptos Foundation
+
+use aptos_enum_conversion_derive::EnumConversion;
+
+#[derive(EnumMessage)]
```
zekun000, line 5: you need to update the tests lol

ibalajiarun: good catch.

Issue #8516, opened 2023-06-04
[Bug] nested mul_div panicked 


0xbe1:
# 🐛 Bug

## To reproduce

**Code snippet to reproduce**

This test passes

```rust
#[test]
fun test_nested_mul_div() {
    let a = math64::mul_div(1, 1, 1);
    assert!(math64::mul_div(1, a,1) == 1, 0);
}
```

This test throws exception.

```rust
#[test]
fun test_nested_mul_div() {
    assert!(math64::mul_div(1, math64::mul_div(1, 1, 1),1) == 1, 0);
}
```

error:

```
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', third_party/move/move-compiler/src/to_bytecode/context.rs:204:73
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

## Expected Behavior

nested should be fine

## System information

```
{
  "Result": {
    "build_branch": "main",
    "build_cargo_version": "cargo 1.66.1 (ad779e08b 2023-01-10)",
    "build_clean_checkout": "false",
    "build_commit_hash": "c9cb442e6f03aaa37687d30899480d40408d0719",
    "build_is_release_build": "true",
    "build_os": "macos-x86_64",
    "build_pkg_version": "2.0.0",
    "build_profile_name": "cli",
    "build_rust_channel": "1.66.1-x86_64-apple-darwin",
    "build_rust_version": "rustc 1.66.1 (90743e729 2023-01-10)",
    "build_tag": "",
    "build_time": "2023-06-01 19:50:01 +00:00",
    "build_using_tokio_unstable": "true"
  }
}
```

movekevin:
After some quick debugging, I think this is due to the way inline functions work - you can't have inline functions inside inline functions. cc @wrwg

wrwg:
> After some quick debugging, I think this is due to the way inline functions work - you can't have inline functions inside inline functions. cc @wrwg

Definitely you can have inline functions inside inline functions, it happens frequently in the functionals in the framework. This must be something different.

movekevin:
I actually got nested mul_div to work when doing this outside of the math64 module. When doing the same test inside math64, where mul_div is defined as an inline function, I got:

```
bug[ICE12001]: BYTECODE GENERATION FAILED
  ┌─ /Users/kevin/aptos-core/aptos-move/framework/aptos-stdlib/sources/math64.move:2:19
  │
2 │ module aptos_std::math64 {
[...]

PR #8867, opened 2023-06-27
Fix partial inlining bug: visitor should keep going after inlining once

brmataptos:
### Description

- This PR fixes the issue (https://github.com/aptos-labs/aptos-core/issues/8516) of move compiler not inlining calls to inlined methods in actual parameters to a call to an inlined method.

### Test Plan
- Added a test case for nested_mul problem, which no longer panics.
- Note that inlining tests already include a recursive inlining test, but I've added more.
- This also includes some tests of nested and recursive in lining in aptos-stdlib/tests/ for the original case as well as some inlining tests that need execution, not just "checker" testing.

### DIFF
--- a/third_party/move/move-compiler/src/inlining/translate.rs
+++ b/third_party/move/move-compiler/src/inlining/translate.rs
@@ -134,7 +134,11 @@ impl<'l> Inliner<'l> {
     {
         for (_, mid_, mdef) in prog.modules.iter_mut() {
             self.current_module = Some(*mid_);
-            if mode == VisitingMode::All || mdef.is_source_module {
+            let visit_module = match mode {
+                VisitingMode::All => true,
+                VisitingMode::SourceOnly => mdef.is_source_module,
+            };
+            if visit_module {
                 for (loc, fname, fdef) in mdef.functions.iter_mut() {
                     self.current_function = *fname;
                     self.current_function_loc = Some(loc);
@@ -345,7 +349,14 @@ impl<'l, 'r> SubstitutionVisitor<'l, 'r> {
                         items.push_back(sp(loc, SequenceItem_::Seq(body)));
                         Some(UnannotatedExp_::Block(items))
                     },
-                    _ => panic!("ICE expected function parameter to be a lambda"),
+                    _ => {
+                        self.inliner.env.add_diag(diag!(
+			    Inlining::Unsupported,
+			    (repl.exp.loc,
+			     "Inlined function-typed parameter currently must be a literal lambda expression")
+			));
+                        None
+                    },
                 }
             },
             _ => None,
@@ -624,14 +635,16 @@ impl<'l> Inliner<'l> {
                 .zip(mcall.type_arguments.iter())
                 .map(|(p, t)| (p.id, t.clone()))
                 .collect();
-            let (decls_for_let, bindings) = self.process_parameters(
-                call_loc,
-                fdef.signature
-                    .parameters
-                    .iter()
-                    .cloned()
-                    .zip(get_args_from_exp(&mcall.arguments)),
-            );
+            let mut inliner_visitor = OuterVisitor { inliner: self };
+            let mut inlined_args = mcall.arguments.clone();
+            Dispatcher::new(&mut inliner_visitor).exp(&mut inlined_args);
+            let mapped_params = fdef
+                .signature
+                .parameters
+                .iter()
+                .cloned()
+                .zip(get_args_from_exp(&inlined_args));
+            let (decls_for_let, bindings) = self.process_parameters(call_loc, mapped_params);
 
             // Expand the body in its own independent visitor
             self.inline_stack.push_front(global_name); // for cycle detection
--- a/aptos-move/framework/aptos-stdlib/tests/math64_tests.move
+++ b/aptos-move/framework/aptos-stdlib/tests/math64_tests.move
@@ -1,0 +1,21 @@
+#[test_only]
+module aptos_std::math64_tests {
+    use aptos_std::math64;
+
+    #[test]
+    fun test_nested_mul_div() {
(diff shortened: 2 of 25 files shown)


third_party/move/move-compiler/src/inlining/translate.rs
```diff
@@ -173,7 +177,7 @@ impl<'l, 'r> Visitor for OuterVisitor<'l, 'r> {
             UnannotatedExp_::ModuleCall(mcall) => {
                 if let Some(repl) = self.inliner.module_call(ex.exp.loc, mcall.as_mut()) {
                     ex.exp.value = repl;
-                    VisitorContinuation::Stop
+                    VisitorContinuation::Descend
```
wrwg, line 180: Something is not right with this solution. if we Descend, we traverse the tree potentially many times. As a matter of fact, the `SubstitutionVisitor` should do this job for us. (See the comment on line 210: "[The SubstitutionVisitor] recursively invokes itself if it encounters further inlined functions or lambdas.")

Basically what should happen is that the body is constructed as something along this line:

```
let (p1, ..., pn) = (arg1, ..., argn);
body_of_inlined_function
```

... and then the `SubstitutionVisitor` is called on this guy. For some reason, the 'arg' expressions above are not expanded. But there is evidence that macro expansion works for `body_of_inlined_function`, since we have many examples where one inline function calls another (e.g. `vector.move`) and in those cases the called inline functions are expanded just fine.

What could be the reason that `let (p1, ..., pn) = (arg1, ..., argn)` is not correctly expanded? Perhaps this construct is not correctly handled by the visitor logic?

wrwg: The place where the SubstitutionVisitor continues expansion is at line 235:

```
          UnannotatedExp_::ModuleCall(mcall) => {
                if let Some(repl) = self.inliner.module_call(ex.exp.loc, mcall) {
                    ex.exp.value = repl;
                }
                VisitorContinuation::Descend
            },
```

brmataptos: I agree, this solution might do some redundant work, although it seemed modest since it doesn't do much unless there is a call to an inlined function in the tree, and the tree is bounded in size.

I'll have to understand the details of the code better more to be more precise.  Hold on.

brmataptos: Since you have 3 uses of the Visitor, it's hard to change the definition, but the problem is here in visitor.rs's `Dispatcher::exp`:
```
    pub fn exp(&mut self, ex: &mut Exp) {
        self.type_(&mut ex.ty);
        if self.visitor.exp(ex) == VisitorContinuation::Stop {
            return;
        }
        self.exp_unannotated(ex.exp.loc, &mut ex.exp.value)
    }
```
so in the inlining pass (OuterVisitor) the visitor is called on the method call *before* the subexpressions have been processed.  Returning `VisitorContinuation::Stop` prevents the subexpressions being processed.  What we want for inlining is a "post-order" traversal of the AST, so that the ModuleCall is processed before it turns into something else through inlining.

Interestingly, lambdas are handled differently: they get inlined in the later substitution pass.  After inlining a lambda body into a VarCall, they reprocess the resulting expression (`SubstitutionVisitor::exp` calls `SubstitutionVisitor::var_call`, then returns VisitorContinuation::Descend), which is good as it's not clear whether substitution has happened on the lambda before then.

wrwg: But the SubstitutionVisitor which handles the lambdas does also handle further inline function calls. This is what I was pointing to with line 235: `self.inliner.module_call` handles the expansion. Otherwise inlined functions in the body of an expansion wouldn't work (and they are positively working and tested).

There is something going wrong with the arguments. Again, conceptually the following expression is constructed for  inlining and then SubstitionVisitor called on it:

```
let (p1, ..., pn) = (arg1, ..., argn); // arguments to inlined functions
body_of_inlined_function
```

Why does the current code handle inlining fine in `body_of_inlined_function` but not in `arg1,...`?

brmataptos: As discussed in Slack:

A nested inline call f1(f2(3)) in any **non-inlined function** (with both f1 and f2 inlined) fails, while putting the expression in the body of an inlined function is ok.

Incidentally, the error generated differs depending on whether the inlined function is in another module or not.  If in the same module, there is an explicit error (e.g., "IR ERROR: unbound function Self.f2") whereas the cross-module case yields the compiler crash with an unwrap() error.

I've now fixed both cases, and included the relevant tests.

davidiw (approved):
Please get @wrwg review too.

alokeutpal (approved):

third_party/move/move-compiler/src/inlining/translate.rs
```diff
@@ -345,7 +349,14 @@ impl<'l, 'r> SubstitutionVisitor<'l, 'r> {
                         items.push_back(sp(loc, SequenceItem_::Seq(body)));
                         Some(UnannotatedExp_::Block(items))
                     },
-                    _ => panic!("ICE expected function parameter to be a lambda"),
+                    _ => {
+                        self.inliner.env.add_diag(diag!(
```
wrwg, line 353: Why that? This was an ICE for a reason before (as it supposed to not happen). Do you have evidence this need to be an error like this?

wrwg (approved):

third_party/move/move-compiler/src/inlining/translate.rs
brmataptos: We should never produce an ICE, but this ICE happens if I pass a function parameter from one inlined method to another.  There's no logical reason why that shouldn't work, but implementing it is nontrivial.  That case is unsupported functionality.

PR #8869, opened 2023-06-27
Reorganize node API v1 code

banool:
## Reviewer's note
It is technically possible to split this PR somewhat up by first moving the Move code, then splitting out the context, then the entrypoint, then the rest. But this isn't a clean separation, it will add a bunch of additional intermediate busywork for me as I have to update crate names and paths multiple times. But if you find this PR too unwieldy I can do it that way.

## Description

In this PR I make the following changes:
- Move all code from `api/` to `crates/aptos-node-api`.
- Move code specific to the v1 node API into `crates/aptos-node-api/v1`.
- Pull out components that aren't specifically coupled to API v1, such as the context.
- Create a new `aptos-node-api-entrypoint` crate. This is where we setup the runtime, hook up the routes and handlers, etc.
- Move all the Move code that was used by the API tests into `aptos-move/move-examples` (in keeping with how we've been doing it for all other test for platform components).
- Update all the docs accordingly
  - I found relevant docs by checking all of the files returned by this command: `rg -l -i api --type md`

There are a few good reasons to do this:
- By separating the stuff specific to API v1 from the other stuff it paves the way for future versions of the API on the node.
- The entrypoint crate makes it possible to hook up multiple different API routes + backends without having to touch the v1 API code.
- We wanted to move all Rust code into `crates/`.
- We can more easily split up the API context library in the future. This lays the groundwork for eventually deprecating API v1 (in the far future).
- Downstream crates that depend on certain components of the node API (e.g. the context) don't have to take on all the deps of the rest of the node API.
- It makes it clear that what we formerly called the API is actually the node API, which is important because there are other APIs we offer (indexer backed).

There should be no functional changes to anything as a result of this PR.

Some of the goldenfiles get deleted as part of this PR because we don't actually run the tests that generated them (they're marked as `ignore`). As you can see, all the rest are unchanged, just direct copies.

## Test Plan
```
cd crates/aptos-node-api/v1/core
cargo test
```

We expect `node-api-compatibility-tests` to fail because of the `pull_request_target` thing. I'll change it to `pull_request` temporarily and make sure it passes. Update: I did that and it passed: https://github.com/aptos-labs/aptos-core/actions/runs/5415365329/jobs/9843892816.

I ran a local testnet and confirmed that the all pages that work with a local testnet in the explorer still work.


brianolson (approved):
I skimmed the whole thing and I believe it does what it says in the desc; a bunch of moves, some dead file deletions, a whole bunch of use/path updates. LGTM

0xmaayan (approved):
q: does this affect `node run-local-testnet` CLI command?

banool:
@0xmaayan nope it doesn't, the PR doesn't change anything in terms of function, it just reorganizes stuff.

banool:
Two stamps, I'm gonna do some more manual testing and then land this once everything is passing (except the expected failures).

banool:
Everything looks green, just need to force land. I'll do it early next week though in case something goes awry.

JoshLind:
> We wanted to move all Rust code into crates/.

FYI, @banool, I don't think this is necessarily true, e.g., different folks prefer different approaches. Personally, I think that `crates/` should only contain logically cohesive crates that are independent of higher level components/systems (e.g., `api`, `state-sync`, `mempool`, should all have sub-crates inside their own directories). I think this makes the code easier to navigate, but it's obviously subjective 😄

davidiw:
Yeah, moving this into crates seems a bit jarring... it is definitely atypical for (almost) every other core service crate

alokeutpal (approved):

banool:
I don't particularly mind either way but I think if that's the philosophy, we're not applying it particularly consistently. The node isn't the only thing in aptos-core but it's the only thing where its components get this special treatment. Why does a top level service like the indexer get put under ecosystem/ but just a component of the node get a top level directory? The answer ofc is bc it just happened like that over time and changing things now would be hard. My point is the API on the node is not that special imo and I don't see why it should clog up the root directory.

davidiw:
Not very opinionated here, but I do agree that much of the layout is a mess... I also realized that with the new cargo updates, moving around crates within a super crate (?) is much easier.... might be good to at least come to some sort of consensus here on where we go from here :)

banool:
I'll start a thread offline and see if we can come up with something.

banool:
I'm going on leave soon and this is now horribly out of date due to other changes that landed, so I am putting this back into draft mode.

PR #8861, opened 2023-06-27
[TS SDK] Use amount greater than 0 in getTokenOwnersData query

0xmaayan:
### Description
https://aptos-org.slack.com/archives/C041GRFLRL3/p1687823425097419

PR #8871, opened 2023-06-27
[TS SDK] Get token owners and current owner data queries

0xmaayan:
### Description
This is a better solution for what we had in https://github.com/aptos-labs/aptos-core/pull/8861 .

- `getTokenOwnersData` - This query returns historical owners data. Add token v2 standard support.
- `getTokenCurrentOwnerData` - This query returns the current token owner data and supports token v2 standard.

### Test Plan
tests are passing


xbtmatt (approved):
lgtm, seems like a better solution for sure!

gregnazario (approved):

PR #8872, opened 2023-06-27
[deps] Ensure that everything builds on Windows

gregnazario:
### Description
Jemalloc needs to be only built on Unix

### Test Plan
CI


msmouse (approved):
hmm I think this is needed for the benchmark only, shall it go to dev-dependencies? cc @zjma

areshand (approved):

PR #8874, opened 2023-06-27
[object] create sticky object

lightmark:
### Description
when need undeletable/sticky object, we shouldn't force users to call `create_named_object` or they don't have to create a name. what they want is a unique address.


davidiw (approved):
sticky sounds yucky

movekevin (approved):

alokeutpal (approved):

PR #8877, opened 2023-06-27
Add a view function to return address of StakePool to be created

xindingw:
### Description
It's hard to simulate creating a `StakingContract`/`DelegationPool` on testnet because of whitelist. Add a view function to return the expected address of stake pool to be created.
### Test Plan

Unit testing.

### DIFF
--- a/aptos-move/framework/aptos-framework/sources/staking_contract.move
+++ b/aptos-move/framework/aptos-framework/sources/staking_contract.move
@@ -240,7 +240,7 @@ module aptos_framework::staking_contract {
         operator: address,
         contract_creation_seed: vector<u8>,
     ): address {
-        let seed = compute_resource_account_seed(staker, operator, contract_creation_seed);
+        let seed = create_resource_account_seed(staker, operator, contract_creation_seed);
         account::create_resource_address(&staker, seed)
     }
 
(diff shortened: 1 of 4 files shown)


aptos-move/framework/aptos-framework/sources/staking_contract.move
```diff
@@ -722,6 +727,23 @@ module aptos_framework::staking_contract {
         pool_u64::update_total_coins(distribution_pool, updated_total_coins);
     }
 
+    /// Compute the seed to derive the resource account address.
+    fun compute_resource_account_seed(
```
lightmark, line 731: construct/create_

lightmark (approved):

aptos-move/framework/aptos-framework/sources/staking_contract.move
```diff
@@ -722,6 +727,23 @@ module aptos_framework::staking_contract {
         pool_u64::update_total_coins(distribution_pool, updated_total_coins);
     }
 
+    /// Create the seed to derive the resource account address.
+    fun create_resource_account_seed(
```
gerben-stavenga, line 731: create indicates you are "creating" something indicating something non pure. Perhaps compute_resource_account_seed

gerben-stavenga (approved):

PR #8885, opened 2023-06-28
[windows] Add to windows install script line to setup winget

gregnazario:
### Description
No matter what I did to get the script to work, I had to add this command that I got from winget installation.

### Test Plan
I ran this manually, but I should get someone else to confirm on a fresh machine that this works as expected.


saharct:
Hmm, interesting. I did not need to use that command across all of the tests I ran, so I'm curious as to exactly what scenario requires it. @gregnazario did you run this on a physical or virtual machine?

gregnazario:
> Hmm, interesting. I did not need to use that command across all of the tests I ran, so I'm curious as to exactly what scenario requires it. @gregnazario did you run this on a physical or virtual machine?

It was a VM, I will probably try it again on a new VM later to test it out

PR #8887, opened 2023-06-28
Cleanup deps

gregnazario:


brianolson (approved):
Yay updates! (I'm astonished BitVec reversed the order of their params!)

davidiw (approved):

msmouse:
> I'm astonished BitVec reversed the order of their params!

msmouse (approved):

gregnazario:
Had to revert the serde_yaml changes, as they caused a bunch of other issues.  We will want to look into this in the future since there are some bugfixes associated, though some whole backwards incompatible changes were made

Issue #8752, opened 2023-06-20
[Bug][Move-prover][Compiler] Wrong number of result variables in call to $1_vector_length'$1_code_PackageMetadata'


rahxephon89:
# 🐛 Bug

## To reproduce

After enabling verification for `publish_package` in the `code` module:
```rust
    spec publish_package(owner: &signer, pack: PackageMetadata, code: vector<vector<u8>>) {
        pragma verify = true;
    }
```

The following internal error is raised:

```
{
  "Error": "Move Prover failed: [internal] boogie exited with compilation errors:\n/Users/.../aptos-core/aptos-move/framework/aptos-framework/boogie.bpl(11761,4): Error: wrong number of result variables in call to $1_vector_length'$1_code_PackageMetadata': 2\n/Users/.../aptos-core/aptos-move/framework/aptos-framework/boogie.bpl(11907,4): Error: wrong number of result variables in call to $1_vector_borrow'$1_code_PackageMetadata': 2\n2 name resolution errors detected in /Users/.../aptos-core/aptos-move/framework/aptos-framework/boogie.bpl\n"
}

```

## Expected Behavior

No error is generated

rahxephon89:
Here is a minimal example that triggers the bug: 
```
module 0x42::vec_len {
    use std::vector;
    public fun test(packages: &mut vector<u64>) {
        vector::enumerate_ref(packages, |i, old| {
        });
    }
}
```

rahxephon89:
Fixed by #9954

PR #8888, opened 2023-06-28
[Spec] Update spec of code.move

UIZorrot:
### Description

This is a part of the spec work for the Aptos Framework from MoveBit. The changes are:

Add spec as much as possible in code.move;
We tired some way, but still can't bypass the vector::enum loop.

### Test Plan

aptos move prove -f code


aptos-move/framework/aptos-framework/sources/code.move
```diff
@@ -145,12 +145,14 @@ module aptos_framework::code {
         let allowed_deps = check_dependencies(addr, &pack);
 
         // Check package against conflicts
+        // ....
         let module_names = get_module_names(&pack);
-        let packages = &mut borrow_global_mut<PackageRegistry>(addr).packages;
-        let len = vector::length(packages);
+        let package_immutable = &borrow_global<PackageRegistry>(addr).packages;
```
junkil-park, line 150: @UIZorrot , is this code change necessary? or, can we avoid?

rahxephon89 (changes requested):

aptos-move/framework/aptos-framework/sources/code.move
UIZorrot: this change is suggest by teng, which could Bypass the compiler error

aptos-move/framework/aptos-framework/sources/code.spec.move
```diff
@@ -23,27 +23,32 @@ spec aptos_framework::code {
...
+        // TODO: Can't verify 'vector::enumerate' loop.
+        pragma aborts_if_is_partial;
+        let addr = signer::address_of(owner);
+        modifies global<PackageRegistry>(addr);
+        aborts_if pack.upgrade_policy.policy <= upgrade_policy_arbitrary().policy;
     }
 
     spec publish_package_txn {
-        // TODO: Calls `publish_package`.`
+        // TODO: Calls `publish_package`.`a
```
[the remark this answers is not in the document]
UIZorrot: oh its a mistake

aptos-move/framework/aptos-framework/sources/code.move
rahxephon89: Yes, this is due to a compiler bug: https://github.com/aptos-labs/aptos-core/issues/8752

rahxephon89: @movekevin @xindingw We need to make the ref immutable before passing it to `enumerate_ref` below to make the prover work. Could you give some feedbacks here? Thanks

xindingw: It makes sense. Can you add a comment to clarify the purpose?

rahxephon89: @UIZorrot Could you add a comment to explain the purpose for this code change?

alokeutpal (approved):

movekevin (approved):

aptos-move/framework/aptos-framework/sources/code.move
UIZorrot: surely

PR #8892, opened 2023-06-28
new network benchmark service

brianolson:
### Description

Make Direct Send and RPC messages with attached data blobs just to exercise the network and see how fast it can go.

### Test Plan
This is a test tool that will be exercised locally and in small datacenter clusters.

### DIFF
--- a/aptos-node/src/network.rs
+++ b/aptos-node/src/network.rs
@@ -1,6 +1,7 @@
 // Copyright © Aptos Foundation
 // SPDX-License-Identifier: Apache-2.0
 
+use crate::services::start_benchmark_service;
 use aptos_channels::{self, aptos_channel, message_queues::QueueStyle};
 use aptos_config::{
     config::{NetworkConfig, NodeConfig},
@@ -21,6 +22,7 @@ use aptos_network::{
     },
     ProtocolId,
 };
+use aptos_network_benchmark::BenchmarkMessage;
 use aptos_network_builder::builder::NetworkBuilder;
 use aptos_peer_monitoring_service_types::PeerMonitoringServiceMessage;
 use aptos_storage_service_types::StorageServiceMessage;
@@ -124,6 +126,34 @@ pub fn storage_service_network_configuration(node_config: &NodeConfig) -> Networ
     NetworkApplicationConfig::new(network_client_config, network_service_config)
 }
 
+pub fn benchmark_network_configuration(
+    node_config: &NodeConfig,
+) -> Option<NetworkApplicationConfig> {
+    let cfg = match node_config.benchmark {
+        None => return None,
+        Some(x) => x,
+    };
+    if !cfg.enabled {
+        return None;
+    }
+    let direct_send_protocols = vec![ProtocolId::BenchmarkDirectSend];
+    let rpc_protocols = vec![ProtocolId::BenchmarkRpc];
+    let network_client_config =
+        NetworkClientConfig::new(direct_send_protocols.clone(), rpc_protocols.clone());
+    let max_network_channel_size = cfg.max_network_channel_size as usize;
+    let network_service_config = NetworkServiceConfig::new(
+        direct_send_protocols,
+        rpc_protocols,
+        aptos_channel::Config::new(max_network_channel_size)
+            .queue_style(QueueStyle::FIFO)
+            .counters(&aptos_network_benchmark::PENDING_BENCHMARK_NETWORK_EVENTS),
+    );
+    Some(NetworkApplicationConfig::new(
+        network_client_config,
+        network_service_config,
+    ))
+}
+
 /// Extracts all network configs from the given node config
 fn extract_network_configs(node_config: &NodeConfig) -> Vec<NetworkConfig> {
     let mut network_configs: Vec<NetworkConfig> = node_config.full_node_networks.to_vec();
@@ -173,6 +203,7 @@ pub fn setup_networks_and_get_interfaces(
     let mut mempool_network_handles = vec![];
     let mut peer_monitoring_service_network_handles = vec![];
     let mut storage_service_network_handles = vec![];
+    let mut benchmark_handles = Vec::<ApplicationNetworkHandle<BenchmarkMessage>>::new();
     for network_config in network_configs.into_iter() {
         // Create a network runtime for the config
         let runtime = create_network_runtime(&network_config);
@@ -229,6 +260,16 @@ pub fn setup_networks_and_get_interfaces(
         );
         storage_service_network_handles.push(storage_service_network_handle);
 
+        // Register benchmark test service
+        if let Some(app_config) = benchmark_network_configuration(node_config) {
+            let benchmark_handle = register_client_and_service_with_network(
+                &mut network_builder,
+                network_id,
+                app_config,
+            );
+            benchmark_handles.push(benchmark_handle);
+        }
+
         // Build and start the network on the runtime
         network_builder.build(runtime.handle().clone());
         network_builder.start();
@@ -251,8 +292,25 @@ pub fn setup_networks_and_get_interfaces(
         mempool_network_handles,
         peer_monitoring_service_network_handles,
         storage_service_network_handles,
-        peers_and_metadata,
+        peers_and_metadata.clone(),
     );
+
+    if !benchmark_handles.is_empty() {
+        let benchmark_interfaces = create_network_interfaces(
+            benchmark_handles,
+            benchmark_network_configuration(node_config).unwrap(),
+            peers_and_metadata,
+        );
+        let benchmark_service_threads = node_config.benchmark.unwrap().benchmark_service_threads;
+        let benchmark_runtime =
+            aptos_runtimes::spawn_named_runtime("benchmark".into(), benchmark_service_threads);
+        start_benchmark_service(
+            node_config,
+            benchmark_interfaces,
+            benchmark_runtime.handle(),
+        );
+        network_runtimes.push(benchmark_runtime);
+    }
 
     (
         network_runtimes,
--- a/aptos-node/src/services.rs
+++ b/aptos-node/src/services.rs
@@ -12,6 +12,7 @@ use aptos_logger::{debug, telemetry_log_writer::TelemetryLog, LoggerFilterUpdate
 use aptos_mempool::{network::MempoolSyncMsg, MempoolClientRequest, QuorumStoreRequest};
 use aptos_mempool_notifications::MempoolNotificationListener;
 use aptos_network::application::{interface::NetworkClientInterface, storage::PeersAndMetadata};
+use aptos_network_benchmark::{run_benchmark_service, BenchmarkMessage, BenchmarkSharedState};
 use aptos_peer_monitoring_service_server::{
     network::PeerMonitoringServiceNetworkEvents, storage::StorageReader,
     PeerMonitoringServiceServer,
@@ -22,7 +23,8 @@ use aptos_time_service::TimeService;
 use aptos_types::chain_id::ChainId;
 use futures::channel::{mpsc, mpsc::Sender};
 use std::{sync::Arc, time::Instant};
-use tokio::runtime::Runtime;
+// use std::sync::RwLock;
+use tokio::runtime::{Handle, Runtime};
 
 const AC_SMP_CHANNEL_BUFFER_SIZE: usize = 1_024;
 const INTRA_NODE_CHANNEL_BUFFER_SIZE: usize = 1;
@@ -182,6 +184,26 @@ pub fn start_peer_monitoring_service(
 
     // Return the runtime
     peer_monitoring_service_runtime
+}
+
+pub fn start_benchmark_service(
+    node_config: &NodeConfig,
+    network_interfaces: ApplicationNetworkInterfaces<BenchmarkMessage>,
+    runtime: &Handle,
+) {
+    let network_client = network_interfaces.network_client;
+    // let benchmark_service_threads = node_config.benchmark.unwrap().benchmark_service_threads;
+    let shared = Arc::new(tokio::sync::RwLock::new(BenchmarkSharedState::new()));
+    runtime.spawn(run_benchmark_service(
+        // benchmark_service_threads,
+        node_config.clone(),
+        // runtime,
+        network_client,
+        network_interfaces.network_service_events,
+        TimeService::real(),
+        shared,
+    ));
+    // return runtime
 }
 
 /// Starts the telemetry service and grabs the build information
--- a/config/src/config/benchmark.rs
+++ b/config/src/config/benchmark.rs
@@ -1,0 +1,40 @@
+// Copyright © Aptos Foundation
+// SPDX-License-Identifier: Apache-2.0
+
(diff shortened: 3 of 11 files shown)


config/src/config/benchmark.rs
```diff
@@ -0,0 +1,40 @@
+// Copyright © Aptos Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+use serde::{Deserialize, Serialize};
+
+#[derive(Clone, Copy, Debug, Deserialize, PartialEq, Eq, Serialize)]
+#[serde(default, deny_unknown_fields)]
+pub struct BenchmarkConfig {
```
JoshLind, line 8: nit: maybe `NetworkBenchmarkConfig`? Also, should this live in the `network_config.rs` file?

network/benchmark/src/lib.rs
```diff
@@ -0,0 +1,605 @@
+// Copyright © Aptos Foundation
```
JoshLind, line 1: nit: full license?
```
```

network/benchmark/src/lib.rs
```diff
@@ -0,0 +1,605 @@
...
+}
+
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub struct BenchmarkDataReply {
+    pub request_counter: u64, // A monotonically increasing counter to verify responses
+    pub send_micros: i64, // micro seconds since some epoch at a moment just before this message is sent
+    pub your_send_micros: i64, // the send_micros from the previous message
+}
+
+/// Counter for pending network events to the monitoring service (server-side)
```
JoshLind, line 47: nit: `monitoring service` -> `benchmark service`?

network/benchmark/src/lib.rs
```diff
@@ -0,0 +1,605 @@
...
+        &["state"]
+    )
+    .unwrap()
+});
+
+pub fn rpc_bytes(state_label: &'static str) -> IntCounter {
+    APTOS_NETWORK_BENCHMARK_RPC_BYTES.with_label_values(&[state_label])
+}
+
+pub static APTOS_NETWORK_BENCHMARK_RPC_MICROS: Lazy<IntCounterVec> = Lazy::new(|| {
```
JoshLind, line 594: nit: why not move these into a `metrics.rs` or `counters.rs` file?

network/benchmark/src/lib.rs
```diff
@@ -0,0 +1,605 @@
...
+    pub request_counter: u64, // A monotonically increasing counter to verify responses
+    pub send_micros: i64, // micro seconds since some epoch at a moment just before this message is sent
+    pub data: Vec<u8>,    // A vector of bytes to send in the request; zero length in reply
+}
+
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub struct BenchmarkDataReply {
+    pub request_counter: u64, // A monotonically increasing counter to verify responses
+    pub send_micros: i64, // micro seconds since some epoch at a moment just before this message is sent
+    pub your_send_micros: i64, // the send_micros from the previous message
```
JoshLind, line 44: nit: maybe `request_send_micros`? I assume it's the timestamp for the corresponding request, right?

network/benchmark/src/lib.rs
```diff
@@ -0,0 +1,605 @@
...
+#[allow(clippy::large_enum_variant)]
+pub enum BenchmarkMessage {
+    DataSend(BenchmarkDataSend),
+    DataReply(BenchmarkDataReply),
+}
+
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub struct BenchmarkDataSend {
+    pub request_counter: u64, // A monotonically increasing counter to verify responses
+    pub send_micros: i64, // micro seconds since some epoch at a moment just before this message is sent
```
JoshLind, line 36: Silly question, is `some epoch` the `UNIX epoch`, or some other event?

aptos-node/src/services.rs
```diff
@@ -184,6 +185,25 @@ pub fn start_peer_monitoring_service(
...
 }
 
+pub fn start_benchmark_service(
+    node_config: &NodeConfig,
+    network_interfaces: ApplicationNetworkInterfaces<BenchmarkMessage>,
+    runtime: &Handle,
+) {
+    let network_client = network_interfaces.network_client;
+    // let benchmark_service_threads = node_config.benchmark.unwrap().benchmark_service_threads;
+    let shared = Arc::new(tokio::sync::RwLock::new(BenchmarkSharedState::new()));
```
JoshLind, line 195: Any reason we need to pass this in here?

network/benchmark/src/lib.rs
```diff
@@ -0,0 +1,605 @@
...
+    loop {
+        match network_events.next().await {
+            None => {
+                // fused stream will never return more
+                work.close();
+                return;
+            },
+            Some(x) => match work.send(x).await {
+                Ok(_) => {},
+                Err(se) => {
```
JoshLind, line 78: Silly question, what does `se` stand for? 😄

network/benchmark/src/lib.rs
```diff
@@ -0,0 +1,605 @@
...
+                request_counter: send.request_counter,
+                send_micros: time_service.now_unix_time().as_micros() as i64,
+                your_send_micros: send.send_micros,
+            };
+            let result = network_client.send_to_peer(
+                BenchmarkMessage::DataReply(reply),
+                PeerNetworkId::new(network_id, peer_id),
+            );
+            if let Err(err) = result {
+                direct_messages("rerr");
```
JoshLind, line 106: Looks like `rerr` is the metric label for the response error, right? If so, it might be worth making a constant for it. We tend to do this for services that use counters and labels, e.g., https://github.com/aptos-labs/aptos-core/blob/d62901d24ded80940c30b6dcff93d7e0dff852e6/mempool/src/counters.rs#L16

network/benchmark/src/lib.rs
```diff
@@ -0,0 +1,605 @@
...
+            };
+            if rec.request_counter == reply.request_counter {
+                let micros = receive_time - rec.send_micros;
+                direct_messages("ok");
+                direct_micros("ok", micros as u64);
+                direct_bytes("ok", rec.bytes_sent as u64);
+            } else {
+                direct_messages("late");
+                info!(
+                    "benchmark ds [{}] unk bytes in > {} micros",
```
JoshLind, line 127: Silly question: what does `unk` mean? 😄

network/benchmark/src/lib.rs
```diff
@@ -0,0 +1,605 @@
...
+    network_client: NetworkClient<BenchmarkMessage>,
+    network_requests: NetworkServiceEvents<BenchmarkMessage>,
+    time_service: TimeService,
+    shared: Arc<RwLock<BenchmarkSharedState>>,
+) {
+    let config = node_config.benchmark.unwrap();
+    let benchmark_service_threads = config.benchmark_service_threads;
+    let num_threads = match benchmark_service_threads {
+        Some(x) => x,
+        None => match std::thread::available_parallelism() {
```
JoshLind, line 258: FYI: some folks also tend to use `num_cpus: https://github.com/aptos-labs/aptos-core/blob/216c346f7ebccc6ce82a3f5ad945fff2a23a1c36/aptos-move/aptos-transaction-benchmarks/src/main.rs#L90

network/benchmark/src/lib.rs
```diff
@@ -0,0 +1,605 @@
...
+    let mut counter: u64 = rng.gen();
+
+    let mut next_blab: i64 = 0;
+
+    loop {
+        ticker.next().await;
+
+        counter += 1;
+        {
+            // tweak the random payload a little on every send
```
JoshLind, line 318: What's the reason for this? Just to make things more "fair"? 😄

network/benchmark/src/lib.rs
```diff
@@ -0,0 +1,605 @@
...
+    network_id: NetworkId,
+    peer_id: PeerId,
+    shared: Arc<RwLock<BenchmarkSharedState>>,
+) {
+    let config = node_config.benchmark.unwrap();
+    let interval = Duration::from_nanos(1_000_000_000 / config.direct_send_per_second);
+    let ticker = time_service.interval(interval);
+    futures::pin_mut!(ticker);
+    // random payload filler
+    let data_size = config.direct_send_data_size;
```
JoshLind, line 302: nit: I'd put this into a helper function?

JoshLind (commented):
Awesome, thanks @brianolson! Looks great 😄 Mostly just questions and/or nits.

aptos-node/src/network.rs
```diff
@@ -124,6 +126,34 @@ pub fn storage_service_network_configuration(node_config: &NodeConfig) -> Networ
     NetworkApplicationConfig::new(network_client_config, network_service_config)
 }
 
+pub fn benchmark_network_configuration(
```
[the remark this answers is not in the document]
brianolson: It's structurally parallel with `storage_service_network_configuration()` (the function just above), so 'benchmark' is a conceptual peer of 'storage_service'.

aptos-node/src/network.rs
```diff
@@ -255,9 +297,26 @@ pub fn setup_networks_and_get_interfaces(
         mempool_network_handles,
         peer_monitoring_service_network_handles,
         storage_service_network_handles,
-        peers_and_metadata,
+        peers_and_metadata.clone(),
     );
 
+    if !benchmark_handles.is_empty() {
```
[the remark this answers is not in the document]
brianolson: `setup_networks_and_get_interfaces` returns 5 things, a vec of Runtime and 4 items for {consensus,mempool,peer_monitoring,storage_service} but adding a 5th to that list felt like it was starting to get unwieldy so I wanted to do it without returning a 'benchmark' object and it worked. If this pattern looks good I'd be tempted to go back and make the other 4 similarly more concisely contained.

network/benchmark/src/lib.rs
brianolson: I'm surprised we don't have a lint check for that

network/benchmark/src/lib.rs
brianolson: Why not keep them where they're used?

network/benchmark/src/lib.rs
brianolson: It's probably 1970, but I've worked with a few of these where micros-since-2020 was used instead. Some variable-uint schemes store as fewer bytes if they're smaller numbers.

network/benchmark/src/lib.rs
brianolson: I wanted to make sure compression and encryption had some work to do. An early version left the buffer zero filled and it compressed away entirely.

aptos-node/src/services.rs
```diff
@@ -184,6 +185,25 @@ pub fn start_peer_monitoring_service(
     peer_monitoring_service_runtime
 }
 
+pub fn start_benchmark_service(
+    node_config: &NodeConfig,
+    network_interfaces: ApplicationNetworkInterfaces<BenchmarkMessage>,
+    runtime: &Handle,
+) {
+    let network_client = network_interfaces.network_client;
+    // let benchmark_service_threads = node_config.benchmark.unwrap().benchmark_service_threads;
```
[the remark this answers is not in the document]
brianolson: nuts. thought I had them all cleaned up. fixed.

network/benchmark/src/lib.rs
brianolson: "rerr" was premature compression for "reply_error" and has been expanded

network/framework/src/protocols/wire/handshake/v1/mod.rs
```diff
@@ -102,6 +106,8 @@ impl ProtocolId {
             ProtocolId::PeerMonitoringServiceRpc,
             ProtocolId::ConsensusRpcCompressed,
             ProtocolId::ConsensusDirectSendCompressed,
+            ProtocolId::BenchmarkRpc,
```
bchocho, line 109: nit: change the order to be consistent with the enum

network/benchmark/src/lib.rs
```diff
@@ -0,0 +1,607 @@
...
+pub enum BenchmarkMessage {
+    DataSend(BenchmarkDataSend),
+    DataReply(BenchmarkDataReply),
+}
+
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub struct BenchmarkDataSend {
+    pub request_counter: u64, // A monotonically increasing counter to verify responses
+    pub send_micros: i64, // micro seconds since some epoch at a moment just before this message is sent
+    pub data: Vec<u8>,    // A vector of bytes to send in the request; zero length in reply
```
bchocho, line 38: Is my understanding correct that we are BCS encoding the send data? I'm surprised we're getting high performance without using `#[serde(with = "serde_bytes")]` (e.g., [here](https://github.com/aptos-labs/aptos-core/blob/eef00c64113f78e585ce62bfeeaaaca30fd6fad4/consensus/src/dag/types.rs#L605C5-L605C5)). In a previous experiment we saw that the time to serde was fairly large without this.

aptos-node/src/network.rs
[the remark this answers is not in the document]
JoshLind: I see. I'd personally prefer if we kept it consistent with what we have today (and just return the 5 items for now), and then we can clean things up in the next PR? That way the code is never inconsistent across applications 😄

aptos-node/src/network.rs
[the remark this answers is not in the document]
JoshLind: Gotcha. Yeah, naming is hard 😄

network/benchmark/src/lib.rs
JoshLind: Yeah, the lint is in a bad state because we can't check for two patterns 😵‍💫

network/benchmark/src/lib.rs
JoshLind: By that logic everything could go into one file? 😄 Also, I'm in favor of code consistency. If you look at every other application, we tend to put metrics into a separate file (e.g., `counters.rs` and `metrics.rs`). It just makes it easier when code starts to grow.

JoshLind (commented):
Thanks @brianolson. LGTM. Only blocker on my side is that we should avoid calling things `Benchmark*` when the context is unclear. `NetworkBenchmark` (or similar) will go a long way in readability. 😄

network/benchmark/src/lib.rs
brianolson: It appears to do the right thing? metrics counters show the expected amount of bandwidth, if I send 1000/s of 100,000 byte messages I get 100MB/s.

brianolson:
I think I'm avoiding "NetworkBenchmark" because of it's length, would "NetTest" be okay for our network speed-test thing?

network/benchmark/src/lib.rs
bchocho: Yeah, that's weird. I'm fine keeping it this way for now, and I'll play with it when we get a forge job up.

JoshLind (approved):
LGTM, thanks @brianolson 😄 Just be sure to squash when you land, otherwise we'll get 38 commits in the main history 😃

PR #8894, opened 2023-06-28
[move-compiler-v2] Type checker: porting tests for name resolution and fixing related issues

wrwg:
This ports the tests from the v1 compiler for name resolution into the v2  compiler and fixes related issues:

- Add support for macros (`assert!`) which are expanded before type checking
- Renames and extends the builtin function representation to capture all Move builtins, not only specs
- Distinguishing between symbols defined both in spec and impl or only for one of those
- Improve error messages
- Fixes issues as needed to pass tests


third_party/move/move-model/src/builder/builtins.rs
```diff
@@ -0,0 +1,528 @@
+// Copyright (c) The Diem Core Contributors
```
wrwg, line 1: Unfortunately, even though was `git mv`ed from `spec_builtins.rs`, this history is lost after changes.

third_party/move/move-prover/boogie-backend/src/spec_translator.rs
```diff
@@ -908,7 +908,13 @@ impl<'env> SpecTranslator<'env> {
...
                 )
             },
-            Operation::MoveFunction(_, _) | Operation::BorrowGlobal(_) | Operation::Old => {
+            Operation::MoveFunction(_, _)
+            | Operation::BorrowGlobal(_)
+            | Operation::MoveTo
+            | Operation::MoveFrom
+            | Operation::Freeze
+            | Operation::Abort
+            | Operation::Old => {
```
davidiw, line 917: would this have any impact on the prover?

third_party/move/move-model/src/model.rs
```diff
@@ -2901,6 +2898,15 @@ pub struct FunctionData {
     pub(crate) transitive_closure_of_called_funs: RefCell<Option<BTreeSet<QualifiedId<FunId>>>>,
 }
 
+/// Kind of a function,
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub enum FunctionKind {
+    Regular,
+    Inline,
+    Macro,
```
davidiw, line 2906: does macro refer to a function in a parameter? it doesn't appear used, and I'm not sure how this would be instantiated in the model.

davidiw (commented):
syntactically it makes sense and we get improved compiler errors to boot?

do you think there's any possible regressions on the Aptos CLI we should be cognizant about?

aptos-move/framework/src/aptos.rs
```diff
@@ -116,7 +116,7 @@ impl ReleaseTarget {
                     landing_page_template: Some("doc_template/overview.md".to_string()),
                     references_file: Some("doc_template/references.md".to_string()),
                 }),
-                skip_fetch_latest_git_deps: false,
+                skip_fetch_latest_git_deps: true,
```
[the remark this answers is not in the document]
wrwg: During release building we should not fetch deps from github, so I think this is the right thing to do. It did not solve my specific problem, but nevertheless.

third_party/move/move-model/src/model.rs
wrwg: Good catch, we don't need this anymore as macros like `assert!` are now expanded during model construction time. Will resolve in the next PR.

third_party/move/move-prover/boogie-backend/src/spec_translator.rs
wrwg: No, since `spec_translator` translates the specification expression language subset, and those new operators belong to the implementation subset. 

One important step of the new compiler is unifying the expression language of the spec and the impl worlds. Some operators are not defined in both worlds.

wrwg (commented):
> syntactically it makes sense and we get improved compiler errors to boot?

No, this is compiler v2, and about a code path which is not used yet. 

> do you think there's any possible regressions on the Aptos CLI we should be cognizant about?

There should not be any expected regression, existing code paths should behave the same as before.

wrwg:
> > syntactically it makes sense and we get improved compiler errors to boot?
> 
> No, this is compiler v2, and about a code path which is not used yet.

But yes, we get possibly better error message in specification language type checking. Just to be clear, not in Move checking.


> 
> > do you think there's any possible regressions on the Aptos CLI we should be cognizant about?
> 
> There should not be any expected regression, existing code paths should behave the same as before.

davidiw (approved):

alokeutpal (approved):

rahxephon89 (approved):

PR #8896, opened 2023-06-29
[move-examples] Add events to marketplace example

gregnazario:
### Description
This adds marketplace token offers instead of just collection offers.  This allows users to bid on any token at any time, as well as cancel it at any time.

Built on top of https://github.com/aptos-labs/aptos-core/pull/8742

### Test Plan
TBD, CI tests are copied from Collection offer, but should test once manually

### DIFF
--- a/aptos-move/move-examples/marketplace/sources/coin_listing.move
+++ b/aptos-move/move-examples/marketplace/sources/coin_listing.move
@@ -15,10 +15,10 @@ module coin_listing {
     use std::string::String;
 
     use aptos_framework::coin::{Self, Coin};
-    use aptos_framework::event::{Self, EventHandle};
     use aptos_framework::object::{Self, ConstructorRef, Object, ObjectCore};
     use aptos_framework::timestamp;
 
+    use marketplace::events;
     use marketplace::fee_schedule::{Self, FeeSchedule};
     use marketplace::listing::{Self, Listing};
 
@@ -45,8 +45,6 @@ module coin_listing {
     struct FixedPriceListing<phantom CoinType> has key {
         /// The price to purchase the item up for listing.
         price: u64,
-        /// Purchase event -- as it is only ever executed once.
-        purchase_event: EventHandle<PurchaseEvent>,
     }
 
     #[resource_group_member(group = aptos_framework::object::ObjectGroup)]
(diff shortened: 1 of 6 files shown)


aptos-move/move-examples/marketplace/sources/token_offer.move
```diff
@@ -0,0 +1,737 @@
...
+
+        let TokenOffer {
+            fee_schedule: _,
+            item_price: _,
+            expiration_time: _,
+            delete_ref,
+            events,
+        } = move_from<TokenOffer>(token_offer_addr);
+        event::destroy_handle(events);
+        object::delete(delete_ref);
```
bowenyang007, line 352: are we going to emit another event to track what happened? e.g. bought vs canceled

aptos-move/move-examples/marketplace/sources/coin_listing.move
```diff
@@ -374,24 +340,24 @@ module coin_listing {
...
-            price: 0,
-            commission: 0,
-            royalties: 0,
-        };
-        event::emit_event(&mut purchase_event, purchase_event_data);
-        event::destroy_handle(purchase_event);
+        // TODO: Do we emit a different event for ending fixed price
+        fee_schedule::emit_purchase_event(
+            fee_schedule,
+            listing_addr,
```
rtso, line 355: When the listing gets deleted, you don't know which token the listing is for. Can we include token address here too?

aptos-move/move-examples/marketplace/sources/fee_schedule.move
```diff
@@ -240,6 +301,91 @@ module marketplace::fee_schedule {
...
+            commission,
+            royalties,
+        });
+    }
+
+    public(friend) fun emit_collection_offer_event(
+        marketplace: Object<FeeSchedule>,
+        offer: address,
+        seller: address,
+        price: u64,
```
rtso, line 348: I'm gonna assume this is price per token? Can we include # of tokens in the event?

aptos-move/move-examples/marketplace/sources/token_offer.move
```diff
@@ -0,0 +1,725 @@
...
+        coin::deposit(royalty_payee, royalties);
+
+        let fee_schedule = token_offer_obj.fee_schedule;
+        let commission_charge = fee_schedule::commission(fee_schedule, price);
+        let commission = coin::extract(&mut coins, commission_charge);
+        coin::deposit(fee_schedule::fee_address(fee_schedule), commission);
+
+        coin::deposit(seller, coins);
+
+        fee_schedule::emit_token_offer_event(
```
rtso, line 313: We probably also want to emit events when token and collection offers are canceled.

bowenyang007 (approved):

aptos-move/move-examples/marketplace/sources/fee_schedule.move
rtso: Yeah the marketplace is all non-fungible, but for collection bids, there can be multiple tokens filled. In this case, it looks like we emit 1 event per token filled in a collection bid, so token amount here will always be 1.

aptos-move/move-examples/marketplace/sources/token_offer.move
rtso: There's some data that needs to be indexed for deletes, like canceled price and token metadata.

davidiw (approved):

PR #8897, opened 2023-06-29
Add is_std_option function to StructTag

banool:
### Description
This function helps with some of the ABI codegen work I'm doing. Part of the Move stdlib so it should be uncontentious.

### Test Plan
I've used this successfully in code I'm adding later. Compile check. CI.

### DIFF
--- a/third_party/move/move-core/types/src/language_storage.rs
+++ b/third_party/move/move-core/types/src/language_storage.rs
@@ -135,6 +135,14 @@ impl StructTag {
             && self.name.as_str().eq("String")
     }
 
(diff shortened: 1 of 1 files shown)


davidiw (approved):
how are you using this?

seems like copypasta from the above, so no issue

banool:
I'm using it for the ABI based codegen stuff. For example when generating a GraphQL schema from a module ABI, I represent option as an actual option. Similarly on the value creation side I convert option from a struct with a vec inside into just an option.

banool:
One more stamp needed.

gregnazario (approved):

PR #8901, opened 2023-06-29
[storage] update the cron job time to avoid conflict

areshand:
### Description

Currently compaction and backup-verify runs at the same time. It can lead to retry of the backup-verify job due to moving of metadata files.

### DIFF
--- a/terraform/helm/fullnode/values.yaml
+++ b/terraform/helm/fullnode/values.yaml
@@ -149,7 +149,7 @@ backup_verify:
 
 backup_compaction:
   # -- The schedule for backup compaction
-  schedule: "@daily"
+  schedule: "0 6 * * *"
   resources:
     limits:
       cpu: 1


areshand:
> Shouldn't we change the configuration of the compactor and make it not interrupt all the clients? -- I think that was one of the design goals in order to guarantee all tooling runs (internal or external) are unaware of compaction?

This is different. Compaction promise compacted files exists before removing the original files. it will eventually remove the original files to reduce the number of files to be downloaded. 

I think the error happens when removing the files that backup verify assumes exits.

msmouse:
@areshand  If it can crash our job, it'll be able to disrupt external people's db-restore runs? We need to make sure no matter when we run the compaction it's not interrupting anyone.

areshand:
> @areshand If it can crash our job, it'll be able to disrupt external people's db-restore runs? We need to make sure no matter when we run the compaction it's not interrupting anyone.

you are right. I guess this is just flaky of the google command instead of timing issue

PR #8904, opened 2023-06-29
add gas fee payer by adding a new payload type

gerben-stavenga:
### Description

Add gas payer by adding a entry to RawMessageWithData however this struct is not being transmitted. So we need to add a new entry in the payload struct to transmit the intend of gas payer.


types/src/transaction/mod.rs
```diff
@@ -366,6 +368,10 @@ pub enum RawTransactionWithData {
         raw_txn: RawTransaction,
         secondary_signer_addresses: Vec<AccountAddress>,
     },
+    MultiAgentWithFeePayer {
+        raw_txn: RawTransaction,
+        secondary_signer_addresses: Vec<AccountAddress>,
```
davidiw, line 373: should this not include an additional signer called `fee_payer: Option<AccountAddress>` or maybe no option?

gerben-stavenga: It's still not clear how to get fee_payer into the system. But we can always treat it as just another agent

aptos-move/aptos-vm/src/aptos_vm_impl.rs
```diff
@@ -513,9 +516,12 @@ impl AptosVMImpl {
         txn_data: &TransactionMetadata,
         transaction_validation: &TransactionValidation,
     ) -> VMResult<()> {
-        let txn_sequence_number = txn_data.sequence_number();
+        let mut txn_sequence_number = txn_data.sequence_number();
         let txn_gas_price = txn_data.gas_unit_price();
         let txn_max_gas_units = txn_data.max_gas_amount();
+        if txn_data.with_gas_payer {
+            txn_sequence_number |= GAS_PAYER_FLAG_BIT;
+        }
```
davidiw, line 524: should this still be there?

aptos-move/aptos-vm/src/transaction_metadata.rs
```diff
@@ -17,6 +17,7 @@ pub struct TransactionMetadata {
     pub secondary_signers: Vec<AccountAddress>,
     pub secondary_authentication_keys: Vec<Vec<u8>>,
     pub sequence_number: u64,
+    pub with_gas_payer: bool,
```
davidiw, line 20: I don't understand why this isn't an explicit field.

types/src/transaction/mod.rs
davidiw: You mean `TransactionAuthenticator`? Yeah, I think we need a new one of these, this is still hacky.

aptos-move/aptos-vm/src/aptos_vm_impl.rs
gerben-stavenga: It's used to communicate with aptos-framework

aptos-move/aptos-vm/src/transaction_metadata.rs
gerben-stavenga: I think it's good to work towards an orthogonal primitives, instead of adding special cases for each new hot feature we think of. We have a authenticator for multiple signatures, either we reuse it or we implement a 3rd special cased with potential bugs.

aptos-move/aptos-vm/src/transaction_metadata.rs
```diff
@@ -41,8 +41,11 @@ impl TransactionMetadata {
                 .map(|account_auth| account_auth.authentication_key().to_vec())
                 .collect(),
             sequence_number: txn.sequence_number(),
```
gerben-stavenga, line 43: Because we don't do the sequence number bit trick we need feature flags now in this code to prevent forking

gerben-stavenga (commented):
lgtm

aptos-move/aptos-vm/src/transaction_metadata.rs
gerben-stavenga: okay if both path are surely not going on-chain then it's fine

aptos-move/aptos-vm/src/errors.rs
```diff
@@ -156,6 +156,7 @@ pub fn convert_epilogue_error(
     log_context: &AdapterLogSchema,
 ) -> Result<(), VMStatus> {
     let status = error.into_vm_status();
+    println!("{:?}", status);
```
davidiw, line 159: nit

ecosystem/indexer-grpc/indexer-grpc-parser/src/models/default_models/signatures.rs
```diff
@@ -9,12 +9,13 @@ use anyhow::{Context, Result};
...
     account_signature::Signature as AccountSignatureEnum, signature::Signature as SignatureEnum,
     AccountSignature as ProtoAccountSignature, Ed25519Signature as Ed25519SignaturePB,
-    MultiAgentSignature as ProtoMultiAgentSignature,
+    FeePayerSignature as ProtoFeePayerSignature, MultiAgentSignature as ProtoMultiAgentSignature,
     MultiEd25519Signature as ProtoMultiEd25519Signature, Signature as TransactionSignaturePB,
 };
 use field_count::FieldCount;
 use serde::{Deserialize, Serialize};
 
+// TODO: fee payer is just an extended multiagent with the last entry being the fee payer
```
davidiw, line 18: nit

sdk/src/types.rs
```diff
@@ -105,11 +105,11 @@ impl LocalAccount {
         secondary_signers: Vec<&Self>,
         builder: TransactionBuilder,
     ) -> SignedTransaction {
-        let secondary_signer_addresses = secondary_signers
+        let secondary_signer_addresses: Vec<AccountAddress> = secondary_signers
             .iter()
             .map(|signer| signer.address())
             .collect();
-        let secondary_signer_privkeys = secondary_signers
+        let secondary_signer_privkeys: Vec<&Ed25519PrivateKey> = secondary_signers
```
davidiw, line 112: probably can drop these changes

types/src/transaction/authenticator.rs
```diff
@@ -46,6 +46,14 @@ pub enum TransactionAuthenticator {
         public_key: Ed25519PublicKey,
         signature: Ed25519Signature,
     },
+    /// Optional Multi-agent transaction with a fee payer.
+    FeePayer {
+        sender: AccountAuthenticator,
```
davidiw, line 51: yup, ordering here matters, I broke bcs...

davidiw (approved):
good with the few tweaks and of course cicd passing

api/src/transactions.rs
```diff
@@ -852,7 +852,10 @@ impl TransactionsApi {
                         })?;
                 // Verify the signed transaction
                 match signed_transaction.payload() {
-                    TransactionPayload::EntryFunction(entry_function) => {
+                    TransactionPayload::MultiAgentWithFeePayer(
+                        MultisigTransactionPayload::EntryFunction(entry_function),
```
movekevin, line 856: This makes it impossible to support scripts with fee payer in the future

aptos-move/e2e-move-tests/src/tests/scripts.rs
```diff
@@ -84,7 +84,7 @@ fn read_coin(h: &MoveHarness, account: &AccountAddress) -> u64 {
 }
 
 #[test]
-fn test_two_to_two_transfer_gas_payer() {
+fn test_two_to_two_transfer_fee_payer() {
```
movekevin, line 87: Let's actually move these tests out of scripts.rs into their own file. It has been confusing to find tests as we put too many of them in here even if they're not that strongly related

gerben-stavenga: Can we do that in follow-up? I added them here as they are the only one that was using multi agent.

api/types/src/transaction.rs
```diff
@@ -1228,6 +1235,89 @@ impl
...
+}
+
+impl VerifyInput for FeePayerSignature {
+    fn verify(&self) -> anyhow::Result<()> {
+        self.sender.verify()?;
+
+        for signer in self.secondary_signers.iter() {
+            signer.verify()?;
+        }
+        self.fee_payer_signer.verify()?;
```
movekevin, line 1261: Do we need to verify that this signature matches the fee payer address here? Should at least be done in the execution flow somewhere

aptos-move/aptos-vm/src/transaction_metadata.rs
```diff
@@ -100,7 +118,7 @@ impl TransactionMetadata {
     }
 
     pub fn is_multi_agent(&self) -> bool {
-        !self.secondary_signers.is_empty()
+        !(self.secondary_signers.is_empty() && self.fee_payer.is_none())
```
movekevin, line 121: This seems weird. Why is this necessary when fee payer is a separate tx?

ecosystem/indexer-grpc/indexer-grpc-fullnode/src/convert.rs
```diff
@@ -163,6 +163,23 @@ pub fn convert_transaction_payload(
                 convert_multisig_payload(mp),
             )),
         },
+        TransactionPayload::FeePayerPayload(MultisigTransactionPayload::EntryFunctionPayload(sfp)) => {
```
movekevin, line 166: I guess we won't know if this works until it's actually run :| cc @bowenyang007

api/types/src/transaction.rs
gerben-stavenga: It's done in the prologue

aptos-move/aptos-vm/src/transaction_metadata.rs
gerben-stavenga: @davidiw what do you think is best here

types/src/transaction/mod.rs
```diff
@@ -283,6 +283,63 @@ impl RawTransaction {
...
+    /// The order and length of the secondary keys provided here have to match the order and
+    /// length of the `secondary_signers`.
+    pub fn sign_fee_payer(
+        self,
+        sender_private_key: &Ed25519PrivateKey,
+        secondary_signers: Vec<AccountAddress>,
+        secondary_private_keys: Vec<&Ed25519PrivateKey>,
+        fee_payer_address: AccountAddress,
+        fee_payer_private_key: &Ed25519PrivateKey,
+    ) -> Result<SignatureCheckedTransaction> {
```
gregnazario, line 299: You know, eventually we'll probably want to make these signatures stacked together later, rather than have all the private keys in one place (since they should never all be in the same place)

types/src/transaction/authenticator.rs
```diff
@@ -191,6 +273,33 @@ impl fmt::Display for TransactionAuthenticator {
...
+            Self::FeePayer {
+                sender,
+                secondary_signer_addresses,
+                secondary_signers,
+                fee_payer_address,
+                fee_payer_signer,
+            } => {
+                let mut sec_addrs: String = "".to_string();
+                for sec_addr in secondary_signer_addresses {
+                    sec_addrs = format!("{}\n\t\t\t{:#?},", sec_addrs, sec_addr);
```
gregnazario, line 285: probably 2 tabs was enough, but luckily 8 char tabs aren't common anymore

aptos-move/framework/aptos-framework/sources/transaction_validation.spec.move
```diff
@@ -111,42 +135,48 @@ spec aptos_framework::transaction_validation {
...
-        ensures forall i in 0..num_secondary_signers:
-            account::exists_at(secondary_signer_addresses[i])
-                && secondary_signer_public_key_hashes[i] ==
-                    account::get_authentication_key(secondary_signer_addresses[i]);
+        aborts_if !account::exists_at(gas_payer);
+        aborts_if !(fee_payer_public_key_hash == account::get_authentication_key(gas_payer));
     }
 
-    /// Abort according to the conditions.
+        /// Abort according to the conditions.
```
gregnazario, line 179: Nit: Weird tab here

aptos-move/e2e-move-tests/src/tests/scripts.rs
```diff
@@ -82,341 +80,3 @@ fn read_coin(h: &MoveHarness, account: &AccountAddress) -> u64 {
         .unwrap()
         .coin()
 }
-
-#[test]
-fn test_two_to_two_transfer_gas_payer() {
-    let mut h = MoveHarness::new_with_features(vec![FeatureFlag::GAS_PAYER_ENABLED], vec![]);
-
-    let alice = h.new_account_at(AccountAddress::from_hex_literal("0xa11ce").unwrap());
```
gregnazario, line 90: I guess we moved these?

aptos-move/aptos-vm/src/transaction_metadata.rs
gregnazario: This isn't actually checking if it's `multi-agent`, but if it's `multi-signer` or `multi-agent`.  Can you verify where it's using this function?

aptos-move/e2e-move-tests/src/tests/fee_payer.rs
```diff
@@ -0,0 +1,249 @@
...
+        .script(script)
+        .sequence_number(h.sequence_number(alice.address()))
+        .max_gas_amount(1_000_000)
+        .gas_unit_price(1)
+        .sign_fee_payer();
+
+    let output = h.executor.execute_transaction(transaction);
+    assert!(transaction_status_eq(
+        output.status(),
+        &TransactionStatus::Discard(StatusCode::UNEXPECTED_ERROR_FROM_KNOWN_MOVE_FUNCTION)
```
movekevin, line 142: Hmm this is getting UNEXPECTED_ERROR_FROM_KNOWN_MOVE_FUNCTION because prologue fails. This doesn't really look right to me at all as UNEXPECTED_ERROR_FROM_KNOWN_MOVE_FUNCTION is an error reserved for internal system (usually unexpected errors). This should return a validation error instead, similar to what I did with multisig prologue errors such as MULTISIG_TRANSACTION_NOT_FOUND

aptos-move/e2e-move-tests/src/tests/fee_payer.rs
```diff
@@ -0,0 +1,249 @@
...
+    let transaction = TransactionBuilder::new(alice.clone())
+        .fee_payer(bob.clone())
+        .entry_function(entry)
+        .sequence_number(h.sequence_number(alice.address()))
+        .max_gas_amount(1_000_000)
+        .gas_unit_price(1)
+        .sign_fee_payer();
+
+    let output = h.executor.execute_transaction(transaction);
+    // The last signer became the gas payer and thus, the execution errors with a mismatch
```
movekevin, line 174: This comment is outdated.

aptos-move/e2e-move-tests/src/tests/fee_payer.rs
```diff
@@ -0,0 +1,249 @@
+// Copyright © Aptos Foundation
```
movekevin, line 1: I think we need more tests for more cases such as:
1. Invalid gas payer signature
2. All valid signatures from senders and gas payers but signed over RawTransaction, now RawTransactionWithData
3. Gas payer account is a single signer (aka I pay for my own gas). I think most of the tests here only test when the payer is separate

types/src/transaction/authenticator.rs
```diff
@@ -108,6 +133,25 @@ impl TransactionAuthenticator {
...
                 signature,
             } => signature.verify(raw_txn, public_key),
+            Self::FeePayer {
+                sender,
+                secondary_signer_addresses,
+                secondary_signers,
+                fee_payer_address,
+                fee_payer_signer,
+            } => {
+                let message = RawTransactionWithData::new_fee_payer(
```
movekevin, line 143: Can you leave a comment here that this is similar to multiagent transactions and is necessary so that the fee payer portion cannot be "stripped" out of a transaction that has been signed by senders?

aptos-move/aptos-vm/src/transaction_metadata.rs
movekevin: I assume this is needed to know which prologue Move function to call since multiagent txs can now have separate gas payer as well, which means different prologue functions need to call for "old" multiagent txs (no gas payer) vs new. Is this necessary though as the vm logic can just look at the presence of gas payer + whether something is a multiagent tx to decide which function to call?

api/types/src/transaction.rs
movekevin: Found this in Move in fee_payer_script_prologue. I wonder if we should have a separate error here because it'd be hard to know which signature mismatches with the address - senders' or payer's. Users would see the same error for both cases and be confused

movekevin (commented):
Can you add a success test in api as well just to make sure the API portion works? E2E move test only test VM and execution. This now needs the API to work correctly as well

aptos-move/e2e-move-tests/src/tests/scripts.rs
movekevin: Yes they got moved to fee_payer. These shouldn't have been here in the first place. scripts.rs seems to getting a lot of random tests that don't really belong

aptos-move/framework/aptos-framework/sources/transaction_validation.move
```diff
@@ -44,7 +43,8 @@ module aptos_framework::transaction_validation {
     const PROLOGUE_EBAD_CHAIN_ID: u64 = 1007;
     const PROLOGUE_ESEQUENCE_NUMBER_TOO_BIG: u64 = 1008;
     const PROLOGUE_ESECONDARY_KEYS_ADDRESSES_COUNT_MISMATCH: u64 = 1009;
-    const PROLOGUE_EGAS_PAYER_ACCOUNT_MISSING: u64 = 1010;
+    const PROLOGUE_EFEE_PAYER_NOT_ENABLED: u64 = 1010;
```
movekevin, line 46: As mentioned in the test earlier, this needs to be properly mapped to a non-internal error similar to what I did for the new multisig account

types/src/transaction/mod.rs
movekevin: What do you mean by "stacked together"?

movekevin (approved):

aptos-move/e2e-move-tests/src/tests/fee_payer.rs
gerben-stavenga: I can't find a way to test for invalid signatures. As everything is private or crate private and the only way to construct signatures are valid signatures

zekun000: https://github.com/aptos-labs/aptos-core/blob/main/crates/aptos-crypto/src/ed25519/ed25519_sigs.rs#L43

PR #8908, opened 2023-06-29
[do-not-merge] Add consistent API testing

ngkuru:

### DIFF
--- a/crates/aptos-api-tester/src/main.rs
+++ b/crates/aptos-api-tester/src/main.rs
@@ -254,15 +254,6 @@ async fn probe_getaccountbalanceatversion_1() -> Result<&'static str> {
         return Ok(FAIL_WRONG_BALANCE_AT_VERSION);
     }
 
-<<<<<<< HEAD
-    // use the faucet to ensure that there exists a higher version number
-    faucet_client
-        .fund(giray.address(), 100_000_000)
-        .await
-        .context(ERROR_FAUCET_FUND)?;
-
-=======
->>>>>>> 84b40d670f (Add tests for get_account_balance_at_version)
     // ask for account balance with a higher version number
     let response = client
         .get_account_balance_at_version(giray2.address(), version + 1)
--- a/crates/aptos-api-tester/src/main.rs
+++ b/crates/aptos-api-tester/src/main.rs
@@ -1,0 +1,181 @@
+// Copyright © Aptos Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+#![forbid(unsafe_code)]
+
+use anyhow::{Context, Result};
+use aptos_api_types::U64;
+use aptos_rest_client::{Client, FaucetClient};
+use aptos_sdk::types::LocalAccount;
+use aptos_types::account_address::AccountAddress;
+use aptos_types::transaction::authenticator::AuthenticationKey;
+use once_cell::sync::Lazy;
+use rand::Rng;
+use std::str::FromStr;
+use url::Url;
+
+// global parameters (todo: make into clap)
+static HAS_FAUCET_ACCESS: bool = true;
+
+// network urls
+static DEVNET_NODE_URL: Lazy<Url> =
+    Lazy::new(|| Url::parse("https://fullnode.devnet.aptoslabs.com").unwrap());
+static DEVNET_FAUCET_URL: Lazy<Url> =
+    Lazy::new(|| Url::parse("https://faucet.devnet.aptoslabs.com").unwrap());
+static TESTNET_NODE_URL: Lazy<Url> =
+    Lazy::new(|| Url::parse("https://fullnode.testnet.aptoslabs.com").unwrap());
+
+// static accounts to use
+static TEST_ACCOUNT_1: Lazy<AccountAddress> = Lazy::new(|| {
+    AccountAddress::from_hex_literal(
+        "0x7fed760c508a8885e1438c90f81dea6940ad53b05527fadddd3851f53342d7c4",
+    )
+    .unwrap()
+});
+static TEST_ACCOUNT_2: Lazy<AccountAddress> = Lazy::new(|| {
+    AccountAddress::from_hex_literal(
+        "0xf6cee5c359839a321a08af6b8cfe4a6e439db46f4942e21b6845aa57d770efdb",
+    )
+    .unwrap()
+});
+
+// return values
+static SKIP_NO_FAUCET_ACCESS: &str = "This test requires faucet access.";
+static ERROR_CLIENT_RESPONSE: &str = "Client responded with error.";
+static ERROR_FAUCET_FUND: &str = "Funding from faucet failed.";
+static ERROR_OTHER: &str = "Something went wrong.";
+static FAIL_WRONG_AUTH_KEY: &str = "Returned wrong authentication key.";
+static FAIL_WRONG_SEQ_NUMBER: &str = "Returned wrong sequence number.";
+static FAIL_WRONG_BALANCE: &str = "Returned wrong balance.";
+static SUCCESS: &str = "success";
+
+// Calls get_account on a newly created account. Requires faucet.
+async fn probe_getaccount_1() -> Result<&'static str> {
+    // check faucet access
+    if !HAS_FAUCET_ACCESS {
+        return Ok(SKIP_NO_FAUCET_ACCESS);
+    }
+
+    // create the rest client
+    let client = Client::new(DEVNET_NODE_URL.clone());
+    let faucet_client = FaucetClient::new(DEVNET_FAUCET_URL.clone(), DEVNET_NODE_URL.clone());
+
+    // create and fund an account
+    let giray = LocalAccount::generate(&mut rand::rngs::OsRng);
+    faucet_client
+        .fund(giray.address(), 100_000_000)
+        .await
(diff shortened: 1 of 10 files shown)


crates/aptos-api-tester/src/main.rs
```diff
@@ -0,0 +1,578 @@
...
+        Err(e) => println!("{:?}", e),
+    }
+    match probe_getblockbyheight_1().await {
+        Ok(result) => println!("{:?}", result),
+        Err(e) => println!("{:?}", e),
+    }
+    match probe_getblockbyversion_1().await {
+        Ok(result) => println!("{:?}", result),
+        Err(e) => println!("{:?}", e),
+    }
```
gregnazario, line 575: Let's teach some generics and DRY.

Something you can do here is to make a function that will take in the surrounding async function and handle it.

```
async fn save_results<
  Output: std::fmt::Debug,
  Fut: futures::Future<Output = anyhow::Result<Output>>
>(fut: Fut) {
    match future.await {
      Ok(result) => log_success(),
      Err(result) => log_failure()
    }
}
```

then you can simply call for each one:
```
save_results(probe_getblockbyversion_1()).await;
```

Additionally, it would be good to look into storing results in a structured way:
1. Did it succeed or fail?
2. Why did it succeed or fail?
3. How long did it take?
4. etc. etc.

This would allow for keeping track of it all and then possibly parse it / use something else to use the results after.

crates/aptos-api-tester/src/main.rs
```diff
@@ -0,0 +1,578 @@
...
+        "0x7fed760c508a8885e1438c90f81dea6940ad53b05527fadddd3851f53342d7c4",
+    )
+    .unwrap()
+});
+static TEST_ACCOUNT_2: Lazy<AccountAddress> = Lazy::new(|| {
+    AccountAddress::from_hex_literal(
+        "0xf6cee5c359839a321a08af6b8cfe4a6e439db46f4942e21b6845aa57d770efdb",
+    )
+    .unwrap()
+});
```
gregnazario, line 42: Additionally, this is good to used fixed data for now, but when we do apply this to multiple networks, it can be harder.  May be useful to either have the tests create their own onchain context

crates/aptos-api-tester/src/main.rs
```diff
@@ -0,0 +1,578 @@
...
+
+// return values
+static SKIP_NO_FAUCET_ACCESS: &str = "This test requires faucet access.";
```
gregnazario, line 45: This actually brings up a good point that maybe just "success" and "failure" are not the only two outputs if you were to keep track of it, perhaps "skip" is another one if the test doesn't apply.  An enum can help with this on the Error, though it makes it a little less convenient then just calling `ensure!()`

sdk/src/token_client.rs
```diff
@@ -0,0 +1,178 @@
...
+                    bcs::to_bytes(&max_amount).unwrap(),
+                    bcs::to_bytes(&uri).unwrap(),
+                    bcs::to_bytes(&royalty_payee_address).unwrap(),
+                    bcs::to_bytes(&royalty_points_denominator).unwrap(),
+                    bcs::to_bytes(&royalty_points_numerator).unwrap(),
+                    bcs::to_bytes(&vec![false, false, false, false, false]).unwrap(),
+                    bcs::to_bytes(property_keys.as_ref()).unwrap(),
+                    // TODO: this is wrong
+                    bcs::to_bytes(property_values.as_ref()).unwrap(),
+                    bcs::to_bytes(property_types.as_ref()).unwrap(),
```
gregnazario, line 134: The function returns `anyhow::Result<>`, so you can change all the `unwrap`s to `?` which will make it a little more readable, and instead of panicking it will print errors at the end.

With an `unwrap`, it'll crash right then and not give you any cleanup handling

sdk/src/token_client.rs
```diff
@@ -0,0 +1,178 @@
...
+
+                ],
+            )),
+            SystemTime::now()
+                .duration_since(UNIX_EPOCH)
+                .unwrap()
```
gregnazario, line 140: Similarly, since it's `anyhow::Result`

You can actually add a contextual message too which will add a message on the error without just the message you don't really know what it is.
```suggestion
                .context("Unable to fetch System Time")?
```

sdk/src/token_client.rs
```diff
@@ -0,0 +1,178 @@
...
+        .sender(account.address())
+        .sequence_number(account.sequence_number())
+        .max_gas_amount(options.max_gas_amount)
+        .gas_unit_price(options.gas_unit_price);
+        let signed_txn = account.sign_with_transaction_builder(transaction_builder);
+        Ok(self
+            .api_client
+            .submit(&signed_txn)
+            .await
+            .context("Failed to submit transfer transaction")?
```
gregnazario, line 154: oh actually you do it here, nice

sdk/src/token_client.rs
```diff
@@ -0,0 +1,178 @@
...
+                ),
+                Identifier::new("create_collection_script").unwrap(),
+                vec![],
+                vec![
+                    bcs::to_bytes(&name).unwrap(),
+                    bcs::to_bytes(&description).unwrap(),
+                    bcs::to_bytes(&uri).unwrap(),
+                    bcs::to_bytes(&max_amount).unwrap(),
+                    bcs::to_bytes(&vec![false, false, false]).unwrap(),
+                ],
```
gregnazario, line 64: Similarly, can get rid of the unwraps with `?`

sdk/src/token_client.rs
```diff
@@ -0,0 +1,178 @@
...
+        description: &str,
+        supply: u64,
+        uri: &str,
+        max_amount: u64,
+        royalty_payee_address: &AccountAddress,
+        royalty_points_denominator: u64,
+        royalty_points_numerator: u64,
+        property_keys: Box<Vec<&str>>,
+        property_values: Box<Vec<&str>>,
+        property_types: Box<Vec<&str>>,
```
gregnazario, line 100: For now, just use hardcoded property keys and values.

sdk/src/token_client.rs
```diff
@@ -0,0 +1,178 @@
...
+    pub async fn create_token(
+        &self,
+        account: &mut LocalAccount,
+        collection_name: &str,
+        name: &str,
+        description: &str,
+        supply: u64,
+        uri: &str,
+        max_amount: u64,
+        royalty_payee_address: &AccountAddress,
```
gregnazario, line 95: `AccountAddress` implements `Copy` so you can actually just pass it in without a reference if you want.

crates/aptos-api-tester/src/main.rs
```diff
@@ -0,0 +1,250 @@
...
+    };
+
+    let expected_balance_at_version = U64(starting_receiver_balance);
+    let actual_balance_at_version = match client
+        .get_account_balance_at_version(address, version - 1)
+        .await
+    {
+        Ok(response) => response.inner().coin.value,
+        Err(e) => return TestResult::Error(e.into()),
+    };
```
gregnazario, line 142: You can cut this down with a `.map_err(|err| TestResult::Error(err.into()))?.inner().coin.value` or similar if you want.

sdk/src/token_client.rs
```diff
@@ -0,0 +1,178 @@
...
+            .api_client
+            .get_index()
+            .await
+            .context("Failed to get chain ID")?
+            .inner()
+            .chain_id;
+        let transaction_builder = TransactionBuilder::new(
+            TransactionPayload::EntryFunction(EntryFunction::new(
+                ModuleId::new(
+                    AccountAddress::from_hex_literal("0x3").unwrap(),
```
gregnazario, line 115: Can pull the `0x3` since it's common up to a lazy at the top.

sdk/src/token_client.rs
```diff
@@ -0,0 +1,178 @@
...
+                    Identifier::new("token").unwrap(),
```
gregnazario, line 116: you can use `ident_str!("token")` and make it a const at the time to clean this up a bit.

crates/aptos-api-tester/src/main.rs
```diff
@@ -0,0 +1,250 @@
...
+    result
+}
+
+/// Tests new account creation. Checks that:
+///   - account data exists
+///   - account balance reflects funded amount
+async fn test_newaccount(
+    client: &Client,
+    account: &LocalAccount,
+    amount_funded: u64,
```
xbtmatt, line 56: Not sure what context you're going to call this from- but I believe the faucets only ever send back a max amount of APT and it's not considered an error. Like if you request 10 it will still give you 1.

You might want to find the hard-coded max fixed amount and assert that the post-fund balance is the max fixed amount rather than the requested amount in the case that the requested amount is greater than the max amount that can be sent from the faucet

ngkuru: yep confirmed that 1APT is max, thanks!

crates/aptos-api-tester/src/main.rs
ngkuru: I think this returns an Err(TestResult::Error)?

gregnazario: Oh, yes you're right since it's an enum type.

The ? is only applicable to Result, so it requires reworking some other pieces then.  So, this is fine for now.

PR #8909, opened 2023-06-29
[Forge] Small fix to sanitize_forge_resource_name.

JoshLind:
### Description
Note: This is required to unblock a currently failing forge test.

This PR fixes a small issue in the `sanitize_forge_resource_name` function. Specifically, it's possible for the forge namespace to end in a `-` (after sanitization), which is not valid. This PR extends the function and adds some new tests.

### Test Plan
New unit tests.


rustielin (approved):

rustielin:
Thanks for fixing this + adding tests!

davidiw (approved):

PR #8910, opened 2023-06-29
[helm][fullnode/aptos-node] add basic PodDisruptionBudget

rustielin:
### Description

Add a pod disruption budget to validators, vfns, and PFNs, such that we try to contain the number of services that can go down at any given point when the disruptions are voluntary

PDB is stable since 1.21 so we can roll it out everywhere without gating it by a helm value for instance

### Test Plan

Apply it


rustielin:
we'll land it elsewhere, probably with VPA

PR #8911, opened 2023-06-30
[CLI][Ledger][3/n] Fix error handling

0xjinn:
### Description

Make sure there is no more panic when any ledger error appear

### Test Plan

Previously, the following would panic, now it is returning proper error:
```
➜  aptos-core git:(jin/ledger_status_code) ✗ aptos-debug account transfer --profile ledger --account devnet --amount 500
Do you want to submit a transaction for a range of [600 - 900] Octas at a gas unit price of 100 Octas? [yes/no] >
yes
{
  "Error": "Unexpected error: Error - Ledger device is locked"
}
```

### DIFF
--- a/sdk/src/types.rs
+++ b/sdk/src/types.rs
@@ -178,7 +178,7 @@ pub enum HardwareWalletType {
 }
 
 pub trait TransactionSigner {
-    fn sign_transaction(&self, txn: RawTransaction) -> SignedTransaction;
+    fn sign_transaction(&self, txn: RawTransaction) -> Result<SignedTransaction>;
 
     fn sign_with_transaction_builder(
         &mut self,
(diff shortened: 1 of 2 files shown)


sdk/src/types.rs
```diff
@@ -200,26 +200,25 @@ pub struct HardwareWalletAccount {
...
     fn sign_with_transaction_builder(
         &mut self,
         builder: TransactionBuilder,
     ) -> Result<SignedTransaction> {
         let two_minutes = Duration::from_secs(2 * 60);
-        let current_time = SystemTime::now()
-            .duration_since(UNIX_EPOCH)
-            .expect("Failed to get system time")
-            + two_minutes;
+        let current_time = SystemTime::now().duration_since(UNIX_EPOCH)? + two_minutes;
```
0xmaayan, line 221: curios, why do we add two_minutes?

0xjinn: Good question! Greg and I discussed about this, 1min might be too short, but 2 mins should be enough for people to pick up their Ledger to review and sign the transaction directly from Ledger

0xjinn:
merging into feature branch

PR #8915, opened 2023-06-30
[Spec] add spec to the math modules

junkil-park:
### Description

- Added the new test file `math8` for the Move Prover crate
- Added the abstract specs to the math modules of `aptos-stdlib`

### Test Plan
cargo test -p aptos-framework


rahxephon89 (approved):

gerben-stavenga (approved):
its a reasonable trick to just make it do on u8

PR #8917, opened 2023-06-30
[api] metrics: add some more buckets to the histograms

bchocho:
### Description

Previously the smallest bucket had a upper limit of 5ms. So the p50 in grafana was always capped below at 2.5ms even for API calls that are expected to take less time.

Also the 100-250ms bucket was large, so split it up.

### Test Plan

Run devnet fullnode locally, make sure it doesn't crash (as prometheus has some runtime checks)


perryjrandall (approved):

JoshLind (approved):
LGTM!

PR #8918, opened 2023-06-30
[vm][gas] refactor: separate out the memory usage tracker

vgao1996:
Back in time, the memory usage tracker was directly embedded into the gas meter because trait for composability did not exist. Now I'm moving it out as a separate layer, so that we have clearer separation of the logic. This does not change the behavior of the VM in any way.

### DIFF
--- a/aptos-move/aptos-memory-usage-tracker/src/lib.rs
+++ b/aptos-move/aptos-memory-usage-tracker/src/lib.rs
@@ -1,0 +1,506 @@
+// Copyright © Aptos Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+use aptos_gas::{
+    AbstractValueSize, AptosGasMeter, AptosGasParameters, Fee, FeePerGasUnit, Gas,
+    GasScalingFactor, InternalGas, NumArgs, NumBytes,
+};
+use aptos_types::{
+    account_config::CORE_CODE_ADDRESS, contract_event::ContractEvent,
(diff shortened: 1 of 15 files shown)


gerben-stavenga (approved):

vgao1996:
Replay task: https://github.com/aptos-labs/aptos-core/actions/runs/5427451381

aptos-move/aptos-memory-usage-tracker/src/lib.rs
```diff
@@ -0,0 +1,506 @@
...
+
+macro_rules! delegate_mut {
+    ($(
+        fn $fn: ident $(<$($lt: lifetime),*>)? (&mut self $(, $arg: ident : $ty: ty)* $(,)?) -> $ret_ty: ty;
+    )*) => {
+        $(fn $fn $(<$($lt)*>)? (&mut self, $($arg: $ty),*) -> $ret_ty {
+            self.base.$fn($($arg),*)
+        })*
+    };
+}
```
wlawt, line 92: Just a quick question: what'd these macros do? nit: for maintainability, it would probably be good to have some comments explaining the macro

vgao1996: Very good point. I have some additional reorganization/refactoring coming down the road, and it would be more convenient if I add the comments as part of that.

wlawt (approved):

PR #8919, opened 2023-06-30
[dag] certified node receiver handler

ibalajiarun:
### Description

This PR adds the receiver side of the Certified Node handler. A validator receives a certified node via RPC and adds it to the DAG after verifying it. It is only added if its parents exist and if they don't, there is a `TODO` to try to fetch it.

### Test Plan

New unit tests.

### DIFF
--- a/consensus/src/dag/dag_handler.rs
+++ b/consensus/src/dag/dag_handler.rs
@@ -1,5 +1,6 @@
 // Copyright © Aptos Foundation
 
+use super::{reliable_broadcast::CertifiedNodeHandler, types::TDAGMessage};
 use crate::{
     dag::{
         dag_network::RpcHandler, dag_store::Dag, reliable_broadcast::NodeBroadcastHandler,
@@ -20,6 +21,8 @@ use std::sync::Arc;
 struct NetworkHandler {
     dag_rpc_rx: aptos_channel::Receiver<Author, IncomingDAGRequest>,
     node_receiver: NodeBroadcastHandler,
+    certified_node_receiver: CertifiedNodeHandler,
+    epoch_state: Arc<EpochState>,
 }
 
 impl NetworkHandler {
@@ -31,7 +34,13 @@ impl NetworkHandler {
     ) -> Self {
         Self {
             dag_rpc_rx,
-            node_receiver: NodeBroadcastHandler::new(dag, signer, epoch_state.verifier.clone()),
+            node_receiver: NodeBroadcastHandler::new(
+                dag.clone(),
+                signer,
+                epoch_state.verifier.clone(),
+            ),
+            certified_node_receiver: CertifiedNodeHandler::new(dag),
+            epoch_state,
         }
     }
 
@@ -45,8 +54,16 @@ impl NetworkHandler {
 
     async fn process_rpc(&mut self, rpc_request: IncomingDAGRequest) -> anyhow::Result<()> {
         let dag_message: DAGMessage = TConsensusMsg::from_network_message(rpc_request.req)?;
+
         let response: anyhow::Result<DAGMessage> = match dag_message {
-            DAGMessage::NodeMsg(node) => self.node_receiver.process(node).map(|r| r.into()),
+            DAGMessage::NodeMsg(node) => node
+                .verify(&self.epoch_state.verifier)
+                .and_then(|_| self.node_receiver.process(node))
+                .map(|r| r.into()),
+            DAGMessage::CertifiedNodeMsg(node) => node
+                .verify(&self.epoch_state.verifier)
+                .and_then(|_| self.certified_node_receiver.process(node))
+                .map(|r| r.into()),
             _ => {
                 error!("unknown rpc message {:?}", dag_message);
                 Err(anyhow::anyhow!("unknown rpc message"))
--- a/consensus/src/dag/reliable_broadcast.rs
+++ b/consensus/src/dag/reliable_broadcast.rs
@@ -1,6 +1,7 @@
 // Copyright © Aptos Foundation
 // SPDX-License-Identifier: Apache-2.0
 
+use super::types::{CertifiedAck, CertifiedNode};
 use crate::{
     dag::{
         dag_network::{DAGNetworkSender, RpcHandler},
@@ -118,6 +119,7 @@ impl NodeBroadcastHandler {
 
     fn validate(&self, node: &Node) -> anyhow::Result<()> {
         let current_round = node.metadata().round();
+
         // round 0 is a special case and does not require any parents
         if current_round == 0 {
             return Ok(());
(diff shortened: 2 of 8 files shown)


consensus/src/dag/reliable_broadcast.rs
```diff
@@ -184,3 +186,53 @@ impl RpcHandler for NodeBroadcastHandler {
...
+
+impl RpcHandler for CertifiedNodeHandler {
+    type Request = CertifiedNode;
+    type Response = CertifiedAck;
+
+    fn process(&mut self, node: Self::Request) -> anyhow::Result<Self::Response> {
+        {
+            let dag_reader = self.dag.read();
+            if dag_reader.exists(&node.digest()) {
+                bail!(CertifiedNodeHandleError::NodeExists);
```
sasha8, line 216: Why error? This can happen. Maybe trace/info?

consensus/src/dag/reliable_broadcast.rs
```diff
@@ -184,3 +186,53 @@ impl RpcHandler for NodeBroadcastHandler {
...
+            }
+
+            if node.metadata().round() > 0
+                && (node.parents().len() == 0
```
sasha8, line 220: I am not sure you need this check.
If the certificate is valid then f+1 honest signed it. 
If an honest signed it then node has enough parents.

consensus/src/dag/reliable_broadcast.rs
```diff
@@ -184,3 +186,53 @@ impl RpcHandler for NodeBroadcastHandler {
...
+                    || !dag_reader.all_exists(
+                        node.parents()
+                            .iter()
+                            .map(|parent| parent.metadata().digest()),
+                    ))
+            {
+                // TODO(ibalajiarun): implement fetching logic.
+                bail!(CertifiedNodeHandleError::MissingParents);
```
sasha8, line 228: So we do not return ack in this case even though the validator got the message?
The other side will keep sending. 
Are we mixing reliable broadcast implementation with fetching?

Can we store the certificate internally and return ack instead?

consensus/src/dag/types.rs
```diff
@@ -367,10 +373,16 @@ pub struct CertifiedAck {
     epoch: u64,
 }
 
+impl CertifiedAck {
+    pub fn new(epoch: u64) -> Self {
+        Self { epoch }
```
sasha8, line 378: Why do we need the epoch here?

sasha8 (commented):
A few questions. Nothing big.

consensus/src/dag/reliable_broadcast.rs
zekun000: I don't think we should send back the ack without having all the parents if we want to guarantee that certificates mean nodes have all history.

I imagine if we plug in the dag_fetcher, it'd have a callback to send the ack if all missing parents are fetched

consensus/src/dag/reliable_broadcast.rs
zekun000: I don't think this is an error even, we probably need to send back ack still. imagine that a previous message triggered a node fetch and timeout the reliable broadcast but eventually successfully inserted. the second retry of reliable broadcast needs to receive ack

consensus/src/dag/types.rs
zekun000: every message needs epoch number to decide whether it's behind by epochs

consensus/src/dag/reliable_broadcast.rs
```diff
@@ -118,6 +119,7 @@ impl NodeBroadcastHandler {
 
     fn validate(&self, node: &Node) -> anyhow::Result<()> {
         let current_round = node.metadata().round();
+
```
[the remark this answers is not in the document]
ibalajiarun: I will follow-up in the next PR, because I need to refactor and fix some tests as well.

consensus/src/dag/dag_handler.rs
```diff
@@ -45,8 +54,16 @@ impl NetworkHandler {
 
     async fn process_rpc(&mut self, rpc_request: IncomingDAGRequest) -> anyhow::Result<()> {
         let dag_message: DAGMessage = TConsensusMsg::from_network_message(rpc_request.req)?;
+
         let response: anyhow::Result<DAGMessage> = match dag_message {
-            DAGMessage::NodeMsg(node) => self.node_receiver.process(node).map(|r| r.into()),
+            DAGMessage::NodeMsg(node) => node
+                .verify(&self.epoch_state.verifier)
```
zekun000, line 60: just a note, we may need to do verify in multi thread before passing to handler

consensus/src/dag/types.rs
```diff
@@ -288,6 +284,16 @@ impl Deref for CertifiedNode {
     }
 }
 
+impl TDAGMessage for CertifiedNode {
+    fn verify(&self, verifier: &ValidatorVerifier) -> anyhow::Result<()> {
+        let node_digest = NodeDigest::new(self.digest());
```
zekun000, line 289: we can't use the digest value, we need to calculate the digest for verification

consensus/src/dag/reliable_broadcast.rs
sasha8: Currently, Balaji is sending back the error. So the sender can choose to treat it as an ack.

consensus/src/dag/reliable_broadcast.rs
```diff
@@ -184,3 +186,46 @@ impl RpcHandler for NodeBroadcastHandler {
...
+                // TODO(ibalajiarun): implement fetching logic.
+                bail!(CertifiedNodeHandleError::MissingParents);
+            }
+        }
+
+        let mut dag_writer = self.dag.write();
+        let epoch = node.metadata().epoch();
+        dag_writer.add_node(node)?;
+
+        Ok(CertifiedAck::new(epoch))
```
sasha8, line 229: nit: to be consistent you can do Ok(CertifiedAck::new(node.metadata().epoch()))

sasha8 (approved):

consensus/src/dag/reliable_broadcast.rs
```diff
@@ -184,3 +186,46 @@ impl RpcHandler for NodeBroadcastHandler {
...
+    type Response = CertifiedAck;
+
+    fn process(&mut self, node: Self::Request) -> anyhow::Result<Self::Response> {
+        {
+            let dag_reader = self.dag.read();
+            if dag_reader.exists(&node.digest()) {
+                return Ok(CertifiedAck::new(node.metadata().epoch()));
+            }
+
+            if !dag_reader.all_exists(node.parents()) {
```
sasha8, line 219: What about round 0 now?
Do you have a special case for it in verify()?

consensus/src/dag/types.rs
ibalajiarun: In #8971

zekun000 (approved):

PR #8920, opened 2023-06-30
Create a new session id for prologue and epilogue

vusirikala:
### Description

This PR addresses a security issue raised by OtterSec related to AUIDs. This PR creates different session ids for prologue, transaction execution, and epilogue. 

Concern by Ottersec - https://aptos-org.slack.com/archives/C05ANQ8K71D/p1687885689724119


gerben-stavenga:
Can you provide a link to the details of the concern

msmouse (approved):
This is so unfortunate.. but I think it's the right approach provided the current circumstances.

igor-aptos:
This looks good,  but I'm not going to accept, as we need someone from VM to take a look. 

@msmouse how do we run replay test with this change? We shouldn't be creating tables in prologue and epilogue, but we should confirm other places also don't rely on it. 
If anything in prologue/epilogue relies on txn hash, we would need to gate this change

msmouse:
@igor-aptos  https://github.com/aptos-labs/aptos-core/actions/workflows/replay-verify.yaml <- trigger here: click "run workflow", keep "main branch" and put in a commit hash to the input box

zekun000 (approved):

vgao1996 (approved):

PR #8925, opened 2023-07-02
[move-examples] Add token offers to marketplace example

gregnazario:
### Description
Adds token offers to the marketplace

### Test Plan
Unit tests, and soon manual testing

### DIFF
--- a/aptos-move/move-examples/marketplace/sources/token_offer.move
+++ b/aptos-move/move-examples/marketplace/sources/token_offer.move
@@ -24,6 +24,7 @@ module token_offer {
     use marketplace::fee_schedule::{Self, FeeSchedule};
     use marketplace::listing::{Self, TokenV1Container};
     use aptos_token::token::TokenId;
+    use aptos_framework::aptos_account;
 
     /// No token offer defined.
     const ENO_TOKEN_OFFER: u64 = 1;
@@ -195,7 +196,7 @@ module token_offer {
     ) {
         let fee = fee_schedule::listing_fee(fee_schedule, total_to_extract);
         let fee_address = fee_schedule::fee_address(fee_schedule);
-        coin::transfer<CoinType>(purchaser, fee_address, fee);
+        aptos_account::transfer_coins<CoinType>(purchaser, fee_address, fee);
 
         let coins = coin::withdraw<CoinType>(purchaser, total_to_extract);
         move_to(offer_signer, CoinOffer { coins });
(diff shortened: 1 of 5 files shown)


aptos-move/move-examples/marketplace/sources/token_offer.move
```diff
@@ -0,0 +1,788 @@
...
+    const ENOT_TOKEN_OWNER: u64 = 3;
+    /// This is not the owner of the token offer.
+    const ENOT_OWNER: u64 = 4;
+    /// The token offer has expired.
+    const EEXPIRED: u64 = 6;
+
+    // Core data structures
+
+    #[resource_group_member(group = aptos_framework::object::ObjectGroup)]
+    /// Create a limed lifetime offer to buy a token. The token and
```
xbtmatt, line 42: Think there's a typo here 😛 supposed to be "limited or lifetime"? Not sure

Other than this lgtm!

aptos-move/move-examples/marketplace/sources/token_offer.move
```diff
@@ -0,0 +1,788 @@
...
+        let royalty_charge = price * royalty_numerator / royalty_denominator;
+        let royalties = coin::extract(&mut coins, royalty_charge);
+        coin::deposit(royalty_payee, royalties);
+
+        let fee_schedule = token_offer_obj.fee_schedule;
+        let commission_charge = fee_schedule::commission(fee_schedule, price);
+        let commission = coin::extract(&mut coins, commission_charge);
+        coin::deposit(fee_schedule::fee_address(fee_schedule), commission);
+
+        coin::deposit(seller, coins);
```
xbtmatt, line 392: If the seller isn't registered for the coin type, this will error with `ECOIN_STORE_NOT_PUBLISHED`, maybe just pass the `&signer` in and register it if it isn't already registered?

gregnazario: good callout, this can be easily fixed.  Doesn't protect well against frozen assets, but we can still handle it

aptos-move/move-examples/marketplace/sources/token_offer.move
```diff
@@ -0,0 +1,788 @@
...
+        assert!(
+            exists<CoinOffer<CoinType>>(token_offer_addr),
+            error::not_found(ENO_COIN_OFFER),
+        );
+        let coin_offer = borrow_global_mut<CoinOffer<CoinType>>(token_offer_addr);
+        let coins = coin::extract(&mut coin_offer.coins, price);
+
+        let royalty_charge = price * royalty_numerator / royalty_denominator;
+        let royalties = coin::extract(&mut coins, royalty_charge);
+        coin::deposit(royalty_payee, royalties);
```
xbtmatt, line 385: This one is a bit harder because you will possibly never have the royalty signer, but if it isn't registered for `CoinType` this function will always fail, so maybe you need to just assert `is_account_registered<CoinType>(royalty_payee)` when the offer is initially created..?

gregnazario: Easy enough, all now register accordingly

xbtmatt (approved):

rtso (approved):

PR #8933, opened 2023-07-03
[BlockSTM] Counter Fix

danielxiangzl:
### Description
Update txn gas counter when BlockSTM finishes without falling back to sequential execution to avoid double counting. Also fix one typo in sequential execution counter.


aptos-move/block-executor/src/executor.rs
```diff
@@ -505,6 +517,7 @@ where
         let mut worker_idx = 0;
 
         let mut accumulated_fee_statement = FeeStatement::zero();
+        let mut txn_fee_statements = Vec::new();
```
zekun000, line 520: nit: Vec::with_capacity(block.len())

aptos-move/block-executor/src/executor.rs
```diff
@@ -138,25 +138,33 @@ where
...
-        );
-        counters::observe_parallel_execution_txn_gas(
-            fee_statement.execution_gas_used() + fee_statement.io_gas_used(),
-            GasType::NON_STORAGE_GAS,
-        );
-        counters::observe_parallel_execution_txn_gas(
-            fee_statement.storage_fee_used(),
-            GasType::STORAGE_FEE,
-        );
+    fn update_parallel_txn_gas_counters(&self, txn_fee_statements: &mut Vec<FeeStatement>) {
```
zekun000, line 141: doesn't need to be mut?

aptos-move/block-executor/src/executor.rs
```diff
@@ -405,6 +415,7 @@ where
                         accumulated_fee_statement,
                         (txn_idx + 1) as usize,
                     );
+                    self.update_parallel_txn_gas_counters(txn_fee_statements);
```
zekun000, line 418: I'd refactor all these common code into the commit guard but probably not easy

zekun000 (approved):

igor-aptos (approved):
single_node_preformance results look good (fixed)!

PR #8947, opened 2023-07-04
[DOCS] Remove tutorial links from tutorial descriptions

alilloig:
### Description

Tutorials are linked on tutorial title and again on the tutorial description just next to the title. Smallest nitpick, but is slightly confusing when you just arrive to the dev docs, since you expect that second link to be something different.


saharct (approved):
Thanks!

PR #8948, opened 2023-07-04
add api test and ts sdk

gerben-stavenga:
### Description

1) Add test for gas fee payer in API folder
2) Add TS SDK support for gas fee payer + test

### DIFF
--- a/api/src/tests/transactions_test.rs
+++ b/api/src/tests/transactions_test.rs
@@ -276,7 +276,7 @@ async fn test_fee_payer_signed_transaction() {
         .post_bcs_txn("/transactions", body)
         .await;
 
-    let (sender, secondary_signers,fee_payer_signer) = match txn.authenticator() {
+    let (sender, _, fee_payer_signer) = match txn.authenticator() {
         TransactionAuthenticator::FeePayer {
             sender,
             secondary_signer_addresses: _,
--- a/api/src/tests/transactions_test.rs
+++ b/api/src/tests/transactions_test.rs
@@ -250,7 +250,7 @@ async fn test_multi_agent_signed_transaction() {
         .post("/transactions", resp)
         .await;
 }
-/* works when feature is enabled
+
 #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
 async fn test_fee_payer_signed_transaction() {
     let mut context = new_test_context(current_function_name!());
@@ -317,7 +317,7 @@ async fn test_fee_payer_signed_transaction() {
         .post("/transactions", resp)
         .await;
 }
-*/
+
 #[ignore]
 #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
 async fn test_multi_ed25519_signed_transaction() {
--- a/api/src/tests/transactions_test.rs
+++ b/api/src/tests/transactions_test.rs
@@ -266,7 +266,8 @@ async fn test_fee_payer_signed_transaction() {
 
     // Create a new account with a multi-agent signer
     let txn = root_account.sign_fee_payer_with_transaction_builder(
-        vec![], &secondary,
+        vec![],
+        &secondary,
         factory.create_user_account(account.public_key()),
     );
 
@@ -282,7 +283,7 @@ async fn test_fee_payer_signed_transaction() {
             secondary_signer_addresses: _,
             secondary_signers,
             fee_payer_address: _,
-            fee_payer_signer
+            fee_payer_signer,
         } => (sender, secondary_signers, fee_payer_signer),
         _ => panic!(
             "expecting TransactionAuthenticator::MultiAgent, but got: {:?}",
--- a/ecosystem/typescript/sdk/package.json
+++ b/ecosystem/typescript/sdk/package.json
@@ -53,6 +53,7 @@
   "dependencies": {
     "@noble/hashes": "1.1.3",
     "@scure/bip39": "1.1.0",
+    "aptos": "^1.12.0",
     "axios": "0.27.2",
     "form-data": "4.0.0",
     "tweetnacl": "1.0.3"
--- a/ecosystem/typescript/sdk/package.json
+++ b/ecosystem/typescript/sdk/package.json
@@ -53,7 +53,6 @@
   "dependencies": {
     "@noble/hashes": "1.1.3",
     "@scure/bip39": "1.1.0",
-    "aptos": "^1.12.0",
     "axios": "0.27.2",
     "form-data": "4.0.0",
     "tweetnacl": "1.0.3"
--- a/ecosystem/typescript/sdk/src/aptos_types/authenticator.ts
+++ b/ecosystem/typescript/sdk/src/aptos_types/authenticator.ts
@@ -20,7 +20,7 @@ export abstract class TransactionAuthenticator {
       case 2:
         return TransactionAuthenticatorMultiAgent.load(deserializer);
       case 3:
-        return TransactionAuthenticatorMultiAgent.load_fee_payer(deserializer);
+        return TransactionAuthenticatorFeePayer.load(deserializer);
       default:
         throw new Error(`Unknown variant index for TransactionAuthenticator: ${index}`);
     }
@@ -83,25 +83,15 @@ export class TransactionAuthenticatorMultiAgent extends TransactionAuthenticator
     public readonly sender: AccountAuthenticator,
     public readonly secondary_signer_addresses: Seq<AccountAddress>,
     public readonly secondary_signers: Seq<AccountAuthenticator>,
-    public readonly fee_payer?: { address: AccountAddress; authenticator: AccountAuthenticator },
   ) {
     super();
   }
 
   serialize(serializer: Serializer): void {
-    if (this.fee_payer) {
-      serializer.serializeU32AsUleb128(3);
-      this.sender.serialize(serializer);
-      serializeVector<AccountAddress>(this.secondary_signer_addresses, serializer);
-      serializeVector<AccountAuthenticator>(this.secondary_signers, serializer);
-      this.fee_payer.address.serialize(serializer);
-      this.fee_payer.authenticator.serialize(serializer);
-    } else {
-      serializer.serializeU32AsUleb128(2);
-      this.sender.serialize(serializer);
-      serializeVector<AccountAddress>(this.secondary_signer_addresses, serializer);
-      serializeVector<AccountAuthenticator>(this.secondary_signers, serializer);
-    }
+    serializer.serializeU32AsUleb128(2);
+    this.sender.serialize(serializer);
+    serializeVector<AccountAddress>(this.secondary_signer_addresses, serializer);
+    serializeVector<AccountAuthenticator>(this.secondary_signers, serializer);
   }
 
   static load(deserializer: Deserializer): TransactionAuthenticatorMultiAgent {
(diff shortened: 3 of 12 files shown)


davidiw (changes requested):
Probably should have a dedicated Authenticator just like in Rust. If we want to make this elegant, we could probably make a builder API with a story around it... something like

Alice wants to send a transaction where Bob is paying
Alice informs Bob somehow of the intent -- what does that look like?
Bob signs something and sends that signature to Alice
Alice needs to verify that Bob's signature is accurate, produces the complete transaction, and then submits it

gerben-stavenga:
> Glad that this seems to work indicating that we have a good end-to-end implementation in the core. For the SDK though, we probably should have a dedicated Authenticator just like in Rust. If we want to make this elegant, we could probably make a builder API with a story around it... something like
> 
> Alice wants to send a transaction where Bob is paying Alice informs Bob somehow of the intent -- what does that look like? Bob signs something and sends that signature to Alice Alice needs to verify that Bob's signature is accurate, produces the complete transaction, and then submits it

I don't see how this is different story to any multi-agent interaction.

ecosystem/typescript/sdk/package.json
```diff
@@ -53,6 +53,7 @@
   "dependencies": {
     "@noble/hashes": "1.1.3",
     "@scure/bip39": "1.1.0",
+    "aptos": "^1.12.0",
```
0xmaayan, line 56: hmm that doesnt make sense, we dont need the `aptos` package as a dependency in the aptos sdk

ecosystem/typescript/sdk/src/plugins/token_client.ts
```diff
@@ -13,6 +13,7 @@ import { MAX_U64_BIG_INT } from "../bcs/consts";
 import { AnyNumber, bcsToBytes, Bytes } from "../bcs";
 import { getPropertyValueRaw, PropertyMap } from "../utils/property_map_serde";
 import { Token, TokenData } from "../aptos_types/token_types";
+import { AccountAddress, AccountAuthenticator } from "../aptos_types";
```
0xmaayan, line 16: just a fyi, `token_client` works with `tokenv1` while `aptos_token` works with `tokenv2`

ecosystem/typescript/sdk/src/tests/e2e/aptos_client.test.ts
```diff
@@ -470,6 +470,109 @@ test(
...
+      property_version: `${propertyVersion}`,
+    };
+
+    // Transfer Token from Alice's Account to Bob's Account with bob paying the fee
+    await tokenClient.getCollectionData(alice.address().hex(), collectionName);
+    let aliceBalance = await tokenClient.getTokenForAccount(alice.address().hex(), tokenId);
+    expect(aliceBalance.amount).toBe("1");
+
+    let aliceBefore = undefined;
+    {
```
0xmaayan, line 535:
```suggestion
       const resources = await client.getAccountResources(alice.address());
      let accountResource = resources.find((r) => r.type === aptosCoin);
      const aliceBefore = (accountResource!.data as any).coin.value;
```

ecosystem/typescript/sdk/src/tests/e2e/aptos_client.test.ts
```diff
@@ -470,6 +470,109 @@ test(
...
+    await client.waitForTransaction(txnHash, { checkSuccess: true });
+
+    aliceBalance = await tokenClient.getTokenForAccount(alice.address().hex(), tokenId);
+    expect(aliceBalance.amount).toBe("0");
+
+    const bobBalance = await tokenClient.getTokenForAccount(bob.address().hex(), tokenId);
+    expect(bobBalance.amount).toBe("1");
+
+    // Check that Alice did not pay the fee
+    {
```
0xmaayan, line 562:
```suggestion
    
```

ecosystem/typescript/sdk/src/tests/e2e/aptos_client.test.ts
```diff
@@ -470,6 +470,109 @@ test(
...
+      const resources = await client.getAccountResources(alice.address());
+      let accountResource = resources.find((r) => r.type === aptosCoin);
+      expect((accountResource!.data as any).coin.value).toBe(aliceBefore);
+    }
+
+    {
```
0xmaayan, line 568:
```suggestion
    
```

ecosystem/typescript/sdk/src/aptos_types/transaction.ts
```diff
@@ -381,15 +383,24 @@ export class MultiAgentRawTransaction extends RawTransactionWithData {
...
   ) {
     super();
   }
 
   serialize(serializer: Serializer): void {
-    // enum variant index
-    serializer.serializeU32AsUleb128(0);
-    this.raw_txn.serialize(serializer);
-    serializeVector<TransactionArgument>(this.secondary_signer_addresses, serializer);
+    if (this.fee_payer_address === undefined) {
```
0xmaayan, line 392: nit
```suggestion
    if (!this.fee_payer_address) {
```

0xmaayan (changes requested):

ecosystem/typescript/sdk/src/tests/e2e/aptos_client.test.ts
gerben-stavenga: wouldn't that create problems with multiple definitions of the same var name?

0xmaayan: then either change the var name to be more explicit or use `let`

gerben-stavenga: i made it a closure

ecosystem/typescript/sdk/src/aptos_types/transaction.ts
```diff
@@ -381,15 +383,24 @@ export class MultiAgentRawTransaction extends RawTransactionWithData {
   constructor(
     public readonly raw_txn: RawTransaction,
     public readonly secondary_signer_addresses: Seq<AccountAddress>,
+    public readonly fee_payer_address?: AccountAddress,
```
movekevin, line 386: Can we separate this out into a different function (set Gas fee payer) to make this flow a lot more explicit? Users should be aware that setting a separate gas payer would require a different signature

ecosystem/typescript/sdk/src/plugins/token_client.ts
```diff
@@ -309,6 +310,7 @@ export class TokenClient {
     amount: AnyNumber,
     propertyVersion: AnyNumber = 0,
     extraArgs?: OptionalTransactionArgs,
+    fee_payer?: AptosAccount,
```
0xmaayan, line 313: this is a dev exp nit. 
most people don't care about the `extraArgs` so if anyone wants to use `fee_payer` would need to call this function like that (as you probably noticed when writing the test)
```
directTransferToken(..,..,..,..,..,..,..,..,undefined,bob)
```
this can be annoying for the end user. maybe a better design should be here, but for now I wonder if we could make another function `directTransferTokenWithFeePayer()` that has a required `fee_payer` argument and leave this function as is

0xmaayan (commented):
there are lint errors, please run `pnpm lint`

ecosystem/typescript/sdk/src/tests/e2e/aptos_client.test.ts
```diff
@@ -470,6 +470,103 @@ test(
...
+      token_data_id: {
+        creator: alice.address().hex(),
+        collection: collectionName,
+        name: tokenName,
+      },
+      property_version: `${propertyVersion}`,
+    };
+
+    // Transfer Token from Alice's Account to Bob's Account with bob paying the fee
+    await tokenClient.getCollectionData(alice.address().hex(), collectionName);
```
0xmaayan, line 530: nice! much cleaner

ecosystem/typescript/sdk/src/aptos_types/transaction.ts
gerben-stavenga: Done

ecosystem/typescript/sdk/src/aptos_types/transaction.ts
```diff
@@ -371,25 +371,61 @@ export abstract class RawTransactionWithData {
...
+    sender: AccountAuthenticator,
+    secondary_signers: Seq<AccountAuthenticator>,
+    fee_payer?: AccountAuthenticator,
+  ): TransactionAuthenticatorMultiAgent {
+    if (secondary_signers.length !== this.secondary_signer_addresses.length) {
+      throw new Error("Number of secondary signers does not match.");
+    }
+    let fee_pay;
+    if (this.fee_payer_address) {
+      if (!fee_payer) {
```
0xmaayan, line 406: a bit confused here, at this point, `fee_payer` is always undefined b/c of line 404?

ecosystem/typescript/sdk/src/aptos_types/transaction.ts
```diff
@@ -371,25 +371,61 @@ export abstract class RawTransactionWithData {
...
+        throw new Error("Fee payer is not set.");
+      }
+      fee_pay = { address: this.fee_payer_address, authenticator: fee_payer };
+    } else if (fee_payer) {
+        throw new Error("Fee payer is set.");
```
0xmaayan, line 411: I am a bit confused about the implementation here? looks like if we dont have `this.fee_payer_address` and/or `fee_payer` we want to throw an error? and if we have both, we want to use it as the `fee_pay`?
could we make it more clear?

gerben-stavenga: yes a tx with a fee payer must be signed with a fee payer, and tx without a fee payer must not have a fee payer signature.

0xmaayan: can we add it as a comment?

0xmaayan (approved):

ecosystem/typescript/sdk/src/aptos_types/authenticator.ts
```diff
@@ -81,15 +83,25 @@ export class TransactionAuthenticatorMultiAgent extends TransactionAuthenticator
     public readonly sender: AccountAuthenticator,
     public readonly secondary_signer_addresses: Seq<AccountAddress>,
     public readonly secondary_signers: Seq<AccountAuthenticator>,
+    public readonly fee_payer?: { address: AccountAddress; authenticator: AccountAuthenticator },
```
davidiw, line 86: Let's try to mirror the Rust code a bit more closely so that this is easier to make sense of to those that jump across code bases. Specifically, let's have a dedicated TransactionAuthenticatorFeePayer.

There's definitely nuances with the surrounding code that would benefit from a refactor, such as moving the variant number to a const / common method.

Once that's done, I think we can delay on the larger request for the end-to-end story for the multi agent and fee payer, though it will limit applications to only advanced developers.

gerben-stavenga: Mimicking the rust code, just enforces the user to pay the price of our stupid design flaws. Now they have to write duplicated specialized code for each case, instead of being able to have a single function handling all cases.

davidiw: This is a rather trivial fix on your part, I'm not sure why you protest so much, but let's dig in here a bit....

1. The TS SDK needs to establish a stronger identity, as of now, it incorporate a mixture of conventions, mostly from the upstream Rust code.
2. Your contribution is rather small and introduces yet a new convention. It breaks from trying to mimic the BCS layout and the broader Rust conventions without any other examples. Generally it is consider effective to mimic surrounding code when making contributions.
3. This PR has no concern about the rest of the SDK. This further exacerbates the concern about the lack of clear developer experience. In fact, the user has to know to look for the fee payer within a multiagent and then ask themselves what is a multiagent.
4. If you want to improve the SDK the solution isn't to make an seemingly arbitrary change deep within it. Instead, please make a holistic change that incorporates the builder framework around the base TransactionAuthenticator -- that makes it so that the developer need only know what functionality they are trying to extract. Or go one step further and make a transaction builder harness that incorporates fee payer.

gerben-stavenga: If you make each a separate type you make them incompatibile. That means you have duplicated functions in your frontend working on MultiAgentTransaction and one FeePayerTransaction and then similar Authenticator versions. So instead of allowing the same code to work on both normal multi agent and ones with additional fee payer, by just allowing one to set the fee payer, we demand people to copy their code and make a separate almost identical set of functions with the only difference a additional fee payer. Why would we want to do that?

davidiw: I don't follow. You can easily make an interface that all of these implement. Furthermore you we can make a builder pattern that hides any of this. Finally, I'd recommend reading about interfaces, inheritance, and code reuse.

There's a reason for strong typesafety and this code eschews that in a way that misguides the reader trying to understand data layouts.

While I would love to see this cleaned up in a builder interface, a one off approach to this at this layer does not seem prudent.

gregnazario: I agree with @davidiw on this one.  Let's keep the types mapped the same as Rust for now.

We have plans to rework the Typescript SDK later this quarter, and having different patterns will make it harder to keep track of / ensure we don't miss anything in this migration.

api/src/tests/transactions_test.rs
```diff
@@ -250,7 +250,7 @@ async fn test_multi_agent_signed_transaction() {
         .post("/transactions", resp)
         .await;
 }
-/* works when feature is enabled
+
 #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
 async fn test_fee_payer_signed_transaction() {
     let mut context = new_test_context(current_function_name!());
```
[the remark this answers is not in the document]
gerben-stavenga: It's largely the same thing

gerben-stavenga (commented):
I rewrote it according to David's wishes and duplicated the DirectTokenTransfer method

gregnazario (approved):

davidiw (approved):

PR #8952, opened 2023-07-05
[Spec] Add high-level specs

rahxephon89:
### Description

This PR adds some specs in the high-level requirements docs.


junkil-park (approved):
Awesome!

xindingw (approved):

PR #8957, opened 2023-07-05
[move-compiler-v2] Refining type checker and porting test cases 

wrwg:
This ports a first batch of ~100 test cases from the v1 compiler related to typing, and adapts the implementation in v2 to handles those scenarios correctly.

- Adds the concept of _constraints_ to type unification. These are basically a first simple version of (for now) builtin traits. For example, a constant `0` has type `X where Constraint::SomePrimitive(u8,u16,..,u256)`. This was needed to match the behavior of the v1 compiler regards inference related to constants, but implements in a simpler and more general way. It wasn't needed previously as in the specification language we have a unified `num` type for integers.
- Implements `&` class of operators in the checker, which weren't supported before.
- Cleans up state of the `ExpBuilder` with an `ExpBuilderMode` type instead of multiple booleans.
- Multiple smaller improvements, reporting additional errors, making error messages more clearer, etc.

Most relevant code changes are in `exp_builder.rs` and `ty.rs`.


rahxephon89 (approved):

third_party/move/move-model/src/builder/exp_builder.rs
```diff
@@ -15,30 +15,29 @@ use crate::{
         StructId, TypeParameter, TypeParameterKind,
     },
     symbol::{Symbol, SymbolPool},
-    ty::{PrimitiveType, Substitution, Type, TypeDisplayContext, Variance, BOOL_TYPE},
+    ty::{
```
gerben-stavenga, line 18: this doesn't seem to be part of compiler-v2 but a generic directory in tp/move. Is this shared code between old and new compiler?

wrwg: Indeed. The move-model contains the type checker for the specification language, which we are expanding to also cover the implementation language, according to the plan.

gerben-stavenga: So implementation language is properly contained in the specification language? With specification language containing arbitrary size integers etc.?

wrwg: Oh and to avoid misunderstanding: this is _not_ shared between old compiler and new compiler. It is shared between move prover, docgen etc. and new compiler.

gerben-stavenga:
Lgtm, but I'm still catching up on the old implementation and the new

gerben-stavenga (approved):

third_party/move/move-model/src/builder/exp_builder.rs
gerben-stavenga: Oh move-prover and new-compiler. That makes a lot more sense

gerben-stavenga: I see the design direction now. We want to make prover and compiler be handled by the same code in the 90% of logic that is common to both

gerben-stavenga: So basically

parser -> AST -> type inference and verification  are shared by prover + compiler

compiler does just the code gen from the IR-AST and type annotations.
prover does the boogie or whatever generation and the formal verification

which are the different paths.

wrwg: ChatGPT, please create design doc from this discussion ;-)

Issue #8934, opened 2023-07-03
[Bug] Compiler panics on native inline functions


georgemitenkov:
# 🐛 Bug

Currently, the Move compiler panics if native function is marked inline.
For example, consider this module:
```move
module std::test {

    public native inline fun foo();

    #[test]
    public fun bar() {
        foo()
    }
}
```

Running
```bash
$ aptos --version
aptos 2.0.1
$ aptos move test --filter test::bar
```

gives

```
thread 'main' panicked at 'ICE inline function without body', third_party/move/move-compiler/src/inlining/translate.rs:74:21
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Note that using `public inline native fun foo();` throws a compile-time error:
```
error: unexpected token
  ┌─ /Users/george/Desktop/aptos-core/aptos-move/framework/aptos-framework/sources/george.move:4:19
  │
4 │     public inline native fun foo();
  │                   ^^^^^^
  │                   │
  │                   Unexpected 'native'
  │                   Expected 'fun'

```

PR #8958, opened 2023-07-05
[move-compiler] Fix Issue 8934: display an error instead of ICE on native inline fun

brmataptos:
### Description

[move-compiler] Inlining: raise an error instead of ICE on a native inline fun declaration. (#8934)

Considered doing it in the parser, but it's easier in the inliner code.

### Test Plan
Added a test case demonstrating bug.  Should have no effect outside of this situation.

### DIFF
--- a/third_party/move/move-compiler/src/inlining/translate.rs
+++ b/third_party/move/move-compiler/src/inlining/translate.rs
@@ -65,17 +65,23 @@ pub fn run_inlining(env: &mut CompilationEnv, prog: &mut Program) {
 impl<'l> Inliner<'l> {
     fn run(&mut self, prog: &mut Program) {
         // First collect all definitions of inlined functions so we can expand them later in the AST.
-        self.visit_functions(prog, VisitingMode::All, &mut |ctx, _, fdef| {
+        self.visit_functions(prog, VisitingMode::All, &mut |ctx, fname, fdef| {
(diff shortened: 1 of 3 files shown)


third_party/move/move-compiler/src/inlining/translate.rs
```diff
@@ -65,17 +65,24 @@ pub fn run_inlining(env: &mut CompilationEnv, prog: &mut Program) {
...
                 ctx.visibilities
                     .insert(global_name, fdef.visibility.clone());
                 if fdef.inline {
-                    assert!(
-                        matches!(fdef.body.value, FunctionBody_::Defined(_)),
-                        "ICE inline function without body"
-                    );
-                    ctx.inline_defs.insert(global_name, fdef.clone());
+                    if !matches!(fdef.body.value, FunctionBody_::Defined(_)) {
+                        Inliner::<'_>::eprint_fdef(&format!("fun {} ", fname), fdef);
```
wrwg, line 75: Is this a debug print leftover?

wrwg (approved):

third_party/move/move-compiler/src/inlining/translate.rs
brmataptos: Oops, yes.

rahxephon89 (approved):

PR #8959, opened 2023-07-05
add workload mix to executor benchmark/regression tests

igor-aptos:

### DIFF
--- a/execution/executor-benchmark/src/main.rs
+++ b/execution/executor-benchmark/src/main.rs
@@ -191,8 +191,16 @@ enum Command {
 
         /// Workload (transaction type). Uses raw coin transfer if not set,
         /// and if set uses transaction-generator-lib to generate it
-        #[clap(long, value_enum, ignore_case = true)]
-        transaction_type: Option<TransactionTypeArg>,
+        #[clap(
+            long,
+            value_enum,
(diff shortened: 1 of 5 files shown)


danielxiangzl (approved):

execution/executor-benchmark/src/main.rs
```diff
@@ -191,8 +191,16 @@ enum Command {
...
         /// and if set uses transaction-generator-lib to generate it
-        #[clap(long, value_enum, ignore_case = true)]
-        transaction_type: Option<TransactionTypeArg>,
+        #[clap(
+            long,
+            value_enum,
+            num_args = 0..,
+            ignore_case = true
+        )]
+        transaction_type: Vec<TransactionTypeArg>,
```
msmouse, line 200: not `transaction_types` just in order not to change the calling scripts?

igor-aptos: I want to keep the same flags for both txn-emitter and executor-benchmark. This change uses the same flags as there. I think it is fine - when you are setting only one transaction type, which is the most common case - you use only this flag. only when you use multiple types - do you use other flags - weights and phases - which is why I put plural on those.

sitalkedia (approved):
LGTM

PR #8961, opened 2023-07-05
Storage service notifications

JoshLind:
Note: 70% of this PR is just unit tests.

### Description
This PR updates state sync to notify the storage service on successful commits. This allows the storage service to immediately serve (and advertise) new data to peers (in contrast to the current approach where the storage service must first poll the DB to look for updates). From a few (informal) experiments, this cuts end-to-end latency by about ~75ms per hop.

The PR offers two commits:
1. Add commit notifications from state sync to the storage service. This is mostly boilerplate as we already have commit notification components that exist today (i.e., for mempool and the event subscription service). Thus, we simply add another notification component and the necessary wiring between state sync and the storage service.
2. Extend the unit (and integration) tests to verify the new functionality.

Notes:
1. We keep the existing polling functionality in the storage server but drastically reduce the frequency (from every `50 ms` to `500 ms`). This is mainly to provide a fallback in case there is a bug and/or unknown race condition in the code. We can eventually remove this (if we wish), but it feels more robust to keep it in place for now.
2. We limit the queue depths (from state sync to the storage service, and from the storage service to the optimistic fetch handler) to `1`. This is because it doesn't make sense for the storage service to execute for every notification message (because it reads the latest version in the DB). Thus, if there are `X` pending notifications, the first one will refresh using the latest DB state and the next `X-1` will execute with an unchanged DB (thus, becoming a no-op and wasting the CPU).

### Test Plan

### DIFF
--- a/clippy.toml
+++ b/clippy.toml
@@ -2,7 +2,7 @@
 cognitive-complexity-threshold = 100
 # types are used for safety encoding
 type-complexity-threshold = 10000
-# manipulating complex states machines in consensus
-too-many-arguments-threshold = 13
+# The state sync driver requires a lot of wiring and channel handles
+too-many-arguments-threshold = 14
 # Reasonably large enum variants are okay
 enum-variant-size-threshold = 1000
--- a/state-sync/inter-component/storage-service-notifications/src/lib.rs
+++ b/state-sync/inter-component/storage-service-notifications/src/lib.rs
@@ -1,0 +1,181 @@
+// Copyright © Aptos Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+#![forbid(unsafe_code)]
+
+use aptos_channels::{self, aptos_channel, message_queues::QueueStyle};
+use async_trait::async_trait;
+use futures::{stream::FusedStream, Stream};
+use serde::{Deserialize, Serialize};
+use std::{
+    fmt,
+    pin::Pin,
+    task::{Context, Poll},
+};
+use thiserror::Error;
+
+// Note: we limit the queue depth to 1 because it doesn't make sense for the storage service
+// to execute for every notification (because it reads the latest version in the DB). Thus,
+// if there are X pending notifications, the first one will refresh using the latest DB and
+// the next X-1 will execute with an unchanged DB (thus, becoming a no-op and wasting the CPU).
+const STORAGE_SERVICE_NOTIFICATION_CHANNEL_SIZE: usize = 1;
+
+#[derive(Clone, Debug, Deserialize, Error, PartialEq, Eq, Serialize)]
+pub enum Error {
+    #[error("Commit notification failed: {0}")]
+    CommitNotificationError(String),
+}
+
+/// The interface between the state sync driver and the storage service, allowing the driver
+/// to notify the storage service of events (e.g., newly committed transactions).
+#[async_trait]
+pub trait StorageServiceNotificationSender: Send + Clone + Sync + 'static {
+    /// Notify the storage service of newly committed transactions
+    /// at the specified version.
+    async fn notify_new_commit(&self, highest_synced_version: u64) -> Result<(), Error>;
+}
+
+/// This method returns a (StorageServiceNotifier, StorageServiceNotificationListener) pair
+/// that can be used to allow state sync and the storage service to communicate.
+///
+/// Note: the driver should take the notifier and the storage service should take the listener.
+pub fn new_storage_service_notifier_listener_pair(
+) -> (StorageServiceNotifier, StorageServiceNotificationListener) {
+    // Create a dedicated channel for notifications
+    let (notification_sender, notification_receiver) = aptos_channel::new(
+        QueueStyle::LIFO,
+        STORAGE_SERVICE_NOTIFICATION_CHANNEL_SIZE,
+        None,
+    );
+
+    // Create a notification sender and listener
+    let storage_service_notifier = StorageServiceNotifier::new(notification_sender);
+    let storage_service_listener = StorageServiceNotificationListener::new(notification_receiver);
+
+    (storage_service_notifier, storage_service_listener)
+}
+
+/// The state sync driver component responsible for notifying the storage service
+#[derive(Clone, Debug)]
+pub struct StorageServiceNotifier {
+    notification_sender: aptos_channel::Sender<(), StorageServiceCommitNotification>,
+}
+
+impl StorageServiceNotifier {
+    fn new(
+        notification_sender: aptos_channel::Sender<(), StorageServiceCommitNotification>,
+    ) -> Self {
+        Self {
+            notification_sender,
+        }
+    }
+}
+
+#[async_trait]
+impl StorageServiceNotificationSender for StorageServiceNotifier {
+    async fn notify_new_commit(&self, highest_synced_version: u64) -> Result<(), Error> {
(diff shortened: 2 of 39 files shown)


state-sync/inter-component/storage-service-notifications/src/lib.rs
```diff
@@ -0,0 +1,178 @@
...
+
+/// This method returns a (StorageServiceNotifier, StorageServiceNotificationListener) pair
+/// that can be used to allow state sync and the storage service to communicate.
+///
+/// Note: state sync should take the notifier and the storage service should take the listener.
+pub fn new_storage_service_notifier_listener_pair(
+) -> (StorageServiceNotifier, StorageServiceNotificationListener) {
+    // Create a dedicated channel for notifications
+    let (notification_sender, notification_receiver) = aptos_channel::new(
+        QueueStyle::FIFO,
```
bchocho, line 43: Shouldn't this technically be LIFO? Maybe it doesn't matter because we don't actually use the value, but maybe the trace log will be confusing if we see old versions?

state-sync/state-sync-v2/state-sync-driver/src/notification_handlers.rs
```diff
@@ -68,24 +69,40 @@ impl CommitNotification {
...
+                "Notifying the storage service of transactions at version: {:?}, timestamp: {:?}",
+                latest_synced_version, blockchain_timestamp_usecs
             ))
         );
+        storage_service_notification_handler
+            .notify_storage_service_of_committed_transactions(latest_synced_version)
+            .await?;
+
+        // Notify mempool of the committed transactions
         let blockchain_timestamp_usecs = latest_synced_ledger_info.ledger_info().timestamp_usecs();
```
bchocho, line 99: This can be removed because it's reading the same thing?

state-sync/state-sync-v2/state-sync-driver/src/notification_handlers.rs
```diff
@@ -68,24 +69,40 @@ impl CommitNotification {
...
+        debug!(
```
bchocho, line 100: Is the log redundant? We just need to print once for both notifications?

state-sync/storage-service/server/src/lib.rs
```diff
@@ -99,45 +109,99 @@ impl<T: StorageReaderInterface> StorageServiceServer<T> {
...
+            storage_service_listener,
         }
     }
 
+    /// Spawns all continuously running utility tasks
+    async fn spawn_continuous_storage_summary_tasks(&mut self) {
+        // Create a channel to notify the optimistic fetch
+        // handler about updates to the cached storage summary.
+        let (cached_summary_update_notifier, cached_summary_update_listener) =
+            aptos_channel::new(QueueStyle::FIFO, CACHED_SUMMARY_UPDATE_CHANNEL_SIZE, None);
```
bchocho, line 121: Also LIFO?

state-sync/storage-service/server/src/lib.rs
```diff
@@ -99,45 +109,99 @@ impl<T: StorageReaderInterface> StorageServiceServer<T> {
...
+
+        // Spawn the refresher for the storage summary cache
+        self.spawn_storage_summary_refresher(cached_summary_update_notifier)
```
bchocho, line 124: Do we actually need both the storage_summary_refresher and optimistic_fetch_handler? It seems like we want an optimistic fetch if and only if the summary is refreshed, but I may be missing something.

bchocho (commented):
Made a pass through the PR

clippy.toml
```diff
@@ -2,7 +2,7 @@
 cognitive-complexity-threshold = 100
 # types are used for safety encoding
 type-complexity-threshold = 10000
-# manipulating complex states machines in consensus
-too-many-arguments-threshold = 13
+# The state sync driver requires a lot of wiring and channel handles
+too-many-arguments-threshold = 14
```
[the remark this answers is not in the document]
JoshLind: Yeah 😃 I might come back to this in the future. I think some of the code needs to be cleaned up anyway, but, it'll likely be a while. Anyway, gonna leave it for now (seeing as though it's only +1) 😄

state-sync/inter-component/storage-service-notifications/src/lib.rs
```diff
@@ -0,0 +1,178 @@
...
+// TODO: do we want to make this configurable?
+const STORAGE_SERVICE_NOTIFICATION_CHANNEL_SIZE: usize = 1;
+
+#[derive(Clone, Debug, Deserialize, Error, PartialEq, Eq, Serialize)]
+pub enum Error {
+    #[error("Commit notification failed: {0}")]
+    CommitNotificationError(String),
+}
+
+/// The interface between state sync and the storage service, allowing state sync to notify
```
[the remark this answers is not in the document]
JoshLind: > Isn't storage service part of state sync?

Yes, in the sense that it's part of the entire pipeline, but from a component perspective, they're separate. Agreed though -- I've changed the comments in this file to call out that it's from the driver (which makes more sense 😄)

state-sync/inter-component/storage-service-notifications/src/lib.rs
JoshLind: Yes, good call. 🚀  Technically, the value isn't used, but I've updated it to be `LIFO` (on both queues) so that the trace logs make sense 😄

state-sync/state-sync-v2/state-sync-driver/src/notification_handlers.rs
JoshLind: Copy/pasta, thanks 😄

state-sync/storage-service/server/src/lib.rs
JoshLind: Done, see above 😄

state-sync/state-sync-v2/state-sync-driver/src/notification_handlers.rs
JoshLind: Hmm... I think it's somewhat useful for debugging given that the `await` calls mean that there can be some time between notifications. Likewise, some notifications require an ack (e.g., mempool), which may block, and the debug statements make it obvious where the blockage/slowdown occurred. 🤔 But, I agree it's not that great. I've merged them all for now 😄 We can always revisit the logs if we encounter issues.

state-sync/storage-service/server/src/lib.rs
JoshLind: > Do we actually need both the storage_summary_refresher and optimistic_fetch_handler?

Can you elaborate? 😄 Is the question more around why we separate them into separate tasks? If so, it's because the optimistic handler can take a non-negligible amount of time to complete, and we don't want it to block the refresher (because this will slow down storage summary updates and affect data advertisements, hence the two tasks). 

> It seems like we want an optimistic fetch if and only if the summary is refreshed, but I may be missing something.

Yeah, this is partly true, but we also need the optimistic fetcher handler to run periodically so that it can perform garbage collection on expired and/or invalid requests. Which is why we still need some periodic execution (e.g., the interval).

bchocho: Makes sense. Thanks!

areshand (approved):
lgtm

bchocho (approved):
Cool! LGTM

PR #8971, opened 2023-07-05
[dag] node and certified node verification checks

ibalajiarun:
### Description

This PR adds self-contained verification checks and validation checks to `Node` and `CertifiedNode` handlers. 

### Test Plan

Unit tests

### DIFF
--- a/consensus/src/dag/types.rs
+++ b/consensus/src/dag/types.rs
@@ -85,6 +85,23 @@ pub struct NodeMetadata {
 }
 
 impl NodeMetadata {
+    #[cfg(test)]
+    pub fn new_for_test(
+        epoch: u64,
+        round: Round,
+        author: Author,
+        timestamp: u64,
+        digest: HashValue,
+    ) -> Self {
+        Self {
+            epoch,
+            round,
+            author,
+            timestamp,
+            digest,
+        }
+    }
+
     pub fn digest(&self) -> &HashValue {
         &self.digest
     }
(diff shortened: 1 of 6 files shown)


consensus/src/dag/types.rs
```diff
@@ -197,8 +239,15 @@ impl Node {
 
 impl TDAGMessage for Node {
     fn verify(&self, verifier: &ValidatorVerifier) -> anyhow::Result<()> {
+        ensure!(self.digest() == self.calculate_digest(), "invalid digest");
+
         let current_round = self.metadata().round();
 
+        if current_round == 0 {
```
zekun000, line 246: ah I think we need to decide on whether it starts at 0 or 1

zekun000 (approved):

consensus/src/dag/types.rs
```diff
@@ -197,8 +239,15 @@ impl Node {
 
 impl TDAGMessage for Node {
     fn verify(&self, verifier: &ValidatorVerifier) -> anyhow::Result<()> {
+        ensure!(self.digest() == self.calculate_digest(), "invalid digest");
```
sasha8, line 242: Does it make sense to move this check after we validate it?
To prevent DDoS. If I already signed a node for you in this round I do not need to verify and compute expensive digest.

consensus/src/dag/types.rs
```diff
@@ -281,6 +332,8 @@ impl Deref for CertifiedNode {
 
 impl TDAGMessage for CertifiedNode {
     fn verify(&self, verifier: &ValidatorVerifier) -> anyhow::Result<()> {
+        ensure!(self.digest() == self.calculate_digest(), "invalid digest");
+
```
sasha8, line 336: Same comment. If we already have this certified node, why compute digests?

ibalajiarun: We need digest to lookup the certifiednode.

sasha8: I think Zekun is changing it. Lookup by metadata.

sasha8 (approved):
Please add a future TODO regarding my comments.

PR #8973, opened 2023-07-06
[Quorum Store] metrics: give proper buckets for histogram

bchocho:
### Description

Having buckets that match well with the data allows us to have accurate histogram_quantile in grafana.

### Test Plan

Run forge, observe the charts make sense.


ibalajiarun (approved):

igor-aptos (approved):
do we want to change some other latency counters outside of QS as well?

bchocho:
We should probably do an audit of other latencies as well. But for now, merge :)

PR #8975, opened 2023-07-06
[dag] commit rule draft

zekun000:


consensus/src/dag/dag_store.rs
```diff
@@ -78,62 +96,154 @@ impl Dag {
...
+            .unwrap_or(false)
+    }
+
+    fn reachable_filter(start: HashValue) -> impl FnMut(&Arc<CertifiedNode>) -> bool {
+        let mut reachable = HashSet::from([start]);
+        move |node| {
+            for parent in node.parents() {
+                reachable.insert(*parent.metadata().digest());
+            }
+            reachable.contains(&node.digest())
```
ibalajiarun, line 190: I think you should only insert the parents if this node is reachable, otherwise you could add parents without them actually reachable.

consensus/src/dag/dag_store.rs
```diff
@@ -78,62 +96,154 @@ impl Dag {
...
+
+    pub fn reachable<'a>(
+        &self,
+        from: &'a Arc<CertifiedNode>,
+        until: Option<Round>,
+    ) -> impl Iterator<Item = &NodeStatus> {
+        let until = until.unwrap_or(self.lowest_round());
+        let mut reachable_filter = Self::reachable_filter(from.digest());
+        self.nodes_by_round
+            .range(until..=from.round())
```
ibalajiarun, line 220: not a strong opinion, but I feel like we might visit fewer nodes if we iterate via the parents.

consensus/src/dag/commit_rule.rs
```diff
@@ -0,0 +1,115 @@
...
+        epoch_state: Arc<EpochState>,
+        latest_ledger_info: LedgerInfoWithSignatures,
+        dag: Arc<RwLock<Dag>>,
+        anchor_election: Box<dyn AnchorElection>,
+        commit_sender: UnboundedSender<OrderedBlocks>,
+    ) -> Self {
+        // TODO: we need to initialize the anchor election based on the dag
+        Self {
+            epoch_state,
+            ordered_block_id: latest_ledger_info.commit_info().id(),
```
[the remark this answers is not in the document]
zekun000: it'd be the last block id from previous epoch

consensus/src/dag/dag_store.rs
zekun000: yes, I thought about it, but given it needs to have 2f+1 parents, I don't think it matters that much. iterate through vector is probably faster than going through pointer lookups too

consensus/src/dag/dag_store.rs
zekun000: good find!

consensus/src/dag/commit_rule.rs
```diff
@@ -0,0 +1,115 @@
...
+            anchor_election,
+            commit_sender,
+        }
+    }
+
+    pub fn new_node(&mut self, node: &CertifiedNode) {
+        let round = node.round();
+        while self.lowest_unordered_round <= round {
+            if let Some(direct_anchor) = self.find_first_anchor_with_enough_votes(round) {
+                let commit_anchor = self.find_first_anchor_to_commit(direct_anchor);
```
sasha8, line 47: commit_anchor -> ordered_anchor?

consensus/src/dag/commit_rule.rs
```diff
@@ -0,0 +1,115 @@
...
+        &self,
+        mut current_anchor: Arc<CertifiedNode>,
+    ) -> Arc<CertifiedNode> {
+        let dag_reader = self.dag.read();
+        let anchor_round = current_anchor.round();
+        let is_anchor = |metadata: &NodeMetadata| -> bool {
+            (metadata.round() ^ anchor_round) & 1 == 0
+                && *metadata.author() == self.anchor_election.get_anchor(metadata.round())
+        };
+        while let Some(next_anchor) = dag_reader
```
sasha8, line 89: Prev_anchor?

consensus/src/dag/commit_rule.rs
```diff
@@ -0,0 +1,115 @@
...
+        }
+    }
+
+    pub fn find_first_anchor_with_enough_votes(
+        &self,
+        target_round: Round,
+    ) -> Option<Arc<CertifiedNode>> {
+        let dag_reader = self.dag.read();
+        let mut current_round = self.lowest_unordered_round;
+        while current_round < target_round {
```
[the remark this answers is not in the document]
sasha8: I think a comment would be nice :)

sasha8 (approved):
Beautiful code!

ibalajiarun (approved):

PR #8976, opened 2023-07-06
[Forge] Add network emulation to PFN forge jobs.

JoshLind:
### Description
This PR offers the following modifications (each in their own commit):
1. Small clean ups to the `MultiRegionNetworkEmulationTest` and network emulation helpers (so that it can be reused and support network chaos for all node types).
2. Extend the existing PFN forge tests to support network chaos, and add two new PFN forge tests: (i) PFNs with max load and network chaos; and (ii) PFNs with constant TPS and network chaos.
3. Add the new PFN forge tests to the existing Github Actions workflow.

### Test Plan
Existing test infrastructure.

### DIFF
--- a/testsuite/forge/src/backend/k8s/chaos.rs
+++ b/testsuite/forge/src/backend/k8s/chaos.rs
@@ -7,6 +7,7 @@ use crate::{
 };
 use anyhow::bail;
 use aptos_logger::info;
+use aptos_sdk::{move_types::account_address::AccountAddress, types::PeerId};
 use std::process::{Command, Stdio};
 use tempfile::TempDir;
 
@@ -44,6 +45,9 @@ macro_rules! CPU_STRESS_CHAOS_TEMPLATE {
     };
 }
 
+// The node name for an address that could not be found in the swarm
+const INVALID_NODE_STRING: &str = "invalid-node";
+
 impl K8sSwarm {
     /// Injects the SwarmChaos into the specified namespace
     pub fn inject_swarm_chaos(&self, chaos: &SwarmChaos) -> Result<()> {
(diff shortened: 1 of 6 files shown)


rustielin (approved):
LGTM

Did you confirm that the chaos is properly applied by actually running these new tests?

JoshLind:
Thanks, all 😄 

> Did you confirm that the chaos is properly applied by actually running these new tests?

Yeah, I'll verify that manually before I land 🤞

JoshLind:
> Did you confirm that the chaos is properly applied by actually running these new tests?

Hmm... looks like it doesn't work. Chaos is applied between the validators, but not anywhere else 🤔 e.g., see the latencies [here](https://aptoslabs.grafana.net/d/peer_monitoring_service/peer-monitoring-service?from=1688755105329&to=1688757617829&var-Datasource=VictoriaMetrics%20Global%20%28Non-mainnet%29&var-metrics_source=All&var-chain_name=forge-1&var-cluster=gcp-forge-1&var-namespace=forge-runner-1688755348&var-kubernetes_pod_name=All&var-role=validator&var-interval=$__auto_interval_interval&orgId=1&viewPanel=47).

testsuite/forge/src/backend/k8s/chaos.rs
```diff
@@ -314,4 +263,26 @@ impl K8sSwarm {
...
+    }
+
+    /// Returns the name of the node associated with the given account address
+    fn get_node_name(&self, node: &AccountAddress) -> &str {
+        if let Some(validator) = self.validator(*node) {
+            validator.name()
+        } else if let Some(fullnode) = self.full_node(*node) {
+            fullnode.name()
+        } else {
+            // TODO: should we throw an error here instead of failing silently?
```
ibalajiarun, line 284: True, better to throw an error. Would save someone a lot of time.

JoshLind: Agreed 😄 Will follow-up to unblock this!

PR #8977, opened 2023-07-06
[TS SDK] Support contentType in request

0xmaayan:
### Description
Current implementation overrides existing `'content-type':'application/x.aptos.signed_transaction+bcs'` header when passing in custom headers. 
This PR adds `contentType` support that when submitting a transaction, the expected content type would be set.
more context here https://aptos-org.slack.com/archives/C04D0LSBVJN/p1688610023758839 
### Test Plan
tests are passing
added specific test https://github.com/aptos-labs/aptos-core/compare/add_mediaType_to_request?expand=1#diff-9187cd50043dfc9d3829abdb7a919d2497b633279daa100db09194b2d6509366R45


ecosystem/typescript/sdk/src/tests/e2e/client.test.ts
```diff
@@ -1,178 +1,235 @@
...
+        method: "GET",
+        originMethod: "test cookies",
+      });
+      expect(response.headers).toHaveProperty("set-cookie");
+    } catch (error: any) {
+      // should not get here
+      expect(true).toBe(false);
+    }
+  },
+  longTestTimeout,
```
0xjinn, line 21: Curious, why do we include a longTestTimeout on every test now?

ecosystem/typescript/sdk/src/tests/e2e/client.test.ts
```diff
@@ -1,178 +1,235 @@
...
+  async () => {
+    const bcsTxn = await getTransaction();
+    try {
+      const response = await aptosRequest({
+        url: `${NODE_URL}`,
+        method: "POST",
+        endpoint: "transactions",
+        body: bcsTxn,
+        originMethod: "test request includes all headers",
+        mediaType: "application/x.aptos.signed_transaction+bcs",
```
0xjinn, line 55: I see, so when the request is a bcs serialized txn, the content type need to be set to `application/x.aptos.signed_transaction+bcs`? Hmm not sure if I seen this in Python nor Rust sdk

ecosystem/typescript/sdk/src/tests/e2e/client.test.ts
0xmaayan: it is flaky on how much time it takes for the test to complete, if you notice we use it on most of the tests we just didn't have it on this file

ecosystem/typescript/sdk/src/tests/e2e/client.test.ts
0xmaayan: interesting.... we should have it, without it API explodes lol

0xmaayan: found it https://github.com/aptos-labs/aptos-core/blob/main/ecosystem/python/sdk/aptos_sdk/async_client.py#L245

ecosystem/typescript/sdk/src/client/core.ts
```diff
@@ -26,12 +26,14 @@ async function axiosRequest<Request, Response>(
   url: string,
   method: "GET" | "POST",
   body?: Request,
+  mediaType?: string,
```
banool, line 29: Let's call it contentType instead, since that's the actual header we're using.

banool (changes requested):

ecosystem/typescript/sdk/src/tests/e2e/client.test.ts
0xjinn: ohhh nice find!

0xjinn (approved):

ecosystem/typescript/sdk/src/client/types.ts
```diff
@@ -14,11 +14,24 @@ export type ClientConfig = {
...
 };
 
+/**
+ * The API request type
+ *
+ * @param url - the url to make the request to, i.e https://fullnode.aptoslabs.devnet.com/v1
+ * @param method - the request method "GET" | "POST"
+ * @param endpoint (optional) - the endpoint to make the request to, i.e transactions
+ * @param body (optional) - the body of the request
+ * @param mediaType (optional) - the media type, that would become the `content-type` header
```
banool, line 24: Perhaps worth mentioning what it defaults to if not specified.

banool (approved):

ecosystem/typescript/sdk/src/client/core.ts
0xmaayan: done

PR #8978, opened 2023-07-06
[CLI] release 2.0.2

0xjinn:
### Description

Notice I update the date instead of new one, because that one was never released on 6/27

### DIFF
--- a/crates/aptos/CHANGELOG.md
+++ b/crates/aptos/CHANGELOG.md
@@ -13,7 +13,7 @@ All notable changes to the Aptos CLI will be captured in this file. This project
 - init command from config subcommand is removed. Please use init from the root command.
   - Example: `aptos config init` -> `aptos init`
 ### Fixed
-- Panic issue when running `aptos move test` is fixed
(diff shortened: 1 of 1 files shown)


0xmaayan (approved):

crates/aptos/CHANGELOG.md
```diff
@@ -12,6 +12,8 @@ All notable changes to the Aptos CLI will be captured in this file. This project
 ### Removed
 - init command from config subcommand is removed. Please use init from the root command.
   - Example: `aptos config init` -> `aptos init`
+### Fixed
+- Panic issue when running `aptos move test` is fixed
```
xbtmatt, line 16: The fix for the `move test` bug is in this release?

0xjinn: Looks like it, waiting for the final signal from Move team

wrwg: This is actually the fix, perhaps we should mention this? https://github.com/aptos-labs/aptos-core/issues/8516

wrwg (approved):

PR #8983, opened 2023-07-06
[TS SDK] Set `WITH_CREDENTIALS`  to false on Indexer requests

0xmaayan:
### Description
Indexer queries fail on browser when `WITH_CREDENTIALS` is set to true

### DIFF
--- a/ecosystem/typescript/sdk/CHANGELOG.md
+++ b/ecosystem/typescript/sdk/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to the Aptos Node SDK will be captured in this file. This ch
 ## Unreleased
 
 - Support `contentType` in request so custom headers would not override required headers
+- Set `WITH_CREDENTIALS` to false on Indexer requests
 
 ## 1.13.0 (2023-07-05)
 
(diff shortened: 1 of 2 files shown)


xbtmatt:
Just curious, would the `WITH_CREDENTIALS` flag normally be inserted by our own SDK or is it something that can be messed with?

0xmaayan:
> Just curious, would the `WITH_CREDENTIALS` flag normally be inserted by our own SDK or is it something that can be inserted for a custom query

end users have the option to set the `WITH_CREDENTIALS` flag when initializing `AptosClient`, to talk with our PFN this needs to be set to true (hence the default is true). Our Indexer server has issues with this being set to true so we must set it to false

0xjinn (approved):
stamp to land.

Btw, what is this WITH_CREDENTIALS flag?

0xmaayan:
> stamp to land.
> 
> Btw, what is this WITH_CREDENTIALS flag?

is allows the browser to include credentials in the request and handle CORS

xbtmatt:
> but one can initialize `IndexerClient(node_url,{WITH_CREDENTIALS:true})` if they run their own indexer server

oh okay, and this wouldn't override that? Wouldn't they be unable to override the override though since their query goes through this function?

0xmaayan:
> > but one can initialize `IndexerClient(node_url,{WITH_CREDENTIALS:true})` if they run their own indexer server
> 
> oh okay, and this wouldn't override that? Wouldn't they be unable to override the override though since their query goes through this function?

amazing catch!

xbtmatt (approved):

PR #8985, opened 2023-07-06
[release] Add credentials to cut release branch workflow

perryjrandall:
Add credentials to allow pushing to aptos core

Test Plan: run workflow from branch

```
git add . && git commit --amend --no-edit && git push -f && gh workflow run --ref perry-fix-cut-branch -F NEW_VERSION=a.b cut-release-branch.yaml -F GIT_HASH=main
```

https://github.com/aptos-labs/aptos-core/actions/runs/5480751652/jobs/9984278547

### DIFF
--- a/.github/workflows/cut-release-branch.yaml
+++ b/.github/workflows/cut-release-branch.yaml
@@ -32,11 +32,15 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8 # pin@v3
+        with:
+          token: ${{ secrets.CUT_RELEASE_BRANCH_CREDENTIALS }}
+          fetch-depth: 0
       - name: Cut Release Branch
         run: |
           set -ex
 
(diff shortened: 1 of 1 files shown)


.github/workflows/cut-release-branch.yaml
```diff
@@ -32,11 +32,15 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8 # pin@v3
+        with:
+          token: ${{ secrets.CUT_RELEASE_BRANCH_CREDENTIALS }}
```
rustielin, line 36: does the APTOS_BOT_GH_PAT not have these creds?

Also is this an aptos bot PAT or deploy key? (if PAT, nit can we prefix the secret name with APTOS_BOT or something so we know where this cred comes from)

rustielin (approved):
LFG

.github/workflows/cut-release-branch.yaml
perryjrandall: It might, I figured it would be better to have a token per function

thepomeranian (approved):

PR #8986, opened 2023-07-06
[dag] message author and epoch checks on receiver side

ibalajiarun:
### Description

This PR adds the peer and author checks for the DAG message on the receiver side.

### DIFF
--- a/consensus/src/dag/dag_handler.rs
+++ b/consensus/src/dag/dag_handler.rs
@@ -56,7 +56,10 @@ impl NetworkHandler {
     async fn process_rpc(&mut self, rpc_request: IncomingDAGRequest) -> anyhow::Result<()> {
         let dag_message: DAGMessage = rpc_request.req.try_into()?;
 
-        if dag_message.author()? != rpc_request.sender {
+        let author = dag_message
+            .author()
+            .map_err(|_| anyhow::anyhow!("unexpected rpc message {:?}", dag_message))?;
+        if author != rpc_request.sender {
             bail!("message author and network author mismatch");
         }
 
--- a/consensus/src/dag/dag_handler.rs
+++ b/consensus/src/dag/dag_handler.rs
@@ -8,6 +8,7 @@ use crate::{
     },
     network::{IncomingDAGRequest, TConsensusMsg},
 };
+use anyhow::bail;
 use aptos_channels::aptos_channel;
 use aptos_consensus_types::common::Author;
 use aptos_infallible::RwLock;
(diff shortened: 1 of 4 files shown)


consensus/src/dag/dag_handler.rs
```diff
@@ -47,15 +48,17 @@ impl NetworkHandler {
...
             }
         }
     }
 
     async fn process_rpc(&mut self, rpc_request: IncomingDAGRequest) -> anyhow::Result<()> {
-        let dag_message: DAGMessage = TConsensusMsg::from_network_message(rpc_request.req)?;
+        let dag_message: DAGMessage = rpc_request.req.try_into()?;
 
-        // TODO: verify epoch number and author
+        if dag_message.author()? != rpc_request.sender {
```
zekun000, line 59: not every message carries the author field, this basically bails all messages that don't have author field

zekun000 (changes requested):

consensus/src/dag/dag_handler.rs
ibalajiarun: true, here, we expect NodeMsg and CertifiedNodeMsg to have the author? any other message should still be an error?

zekun000: this is the main handler that every message will go through right? if we expect all rpc requests to have an author we should have more explicit error message like "unexpected rpc request"

ibalajiarun: Okay, I made the error explicit now.

consensus/src/dag/dag_handler.rs
```diff
@@ -47,13 +48,20 @@ impl NetworkHandler {
...
     }
 
     async fn process_rpc(&mut self, rpc_request: IncomingDAGRequest) -> anyhow::Result<()> {
-        let dag_message: DAGMessage = TConsensusMsg::from_network_message(rpc_request.req)?;
+        let dag_message: DAGMessage = rpc_request.req.try_into()?;
+
+        let author = dag_message
+            .author()
+            .map_err(|_| anyhow::anyhow!("unexpected rpc message {:?}", dag_message))?;
+        if author != rpc_request.sender {
```
zekun000, line 62: we can still use ensure here

zekun000 (approved):

sasha8 (approved):

consensus/src/network.rs
```diff
@@ -583,6 +585,18 @@ impl NetworkTask {
                             warn!(error = ?e, "aptos channel closed");
                         }
                     },
+                    ConsensusMsg::DAGMessage(request) => {
```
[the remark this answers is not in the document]
zekun000: we need to wrap it with callbacks, but maybe we can just pass everything to epoch manager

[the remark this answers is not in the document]
sasha8: Yes, this is what I was thinking.

consensus/src/dag/dag_handler.rs
ibalajiarun: Yes, but ideally, we would log this here, so I kept this an if statement.

consensus/src/epoch_manager.rs
```diff
@@ -1065,6 +1065,19 @@ impl EpochManager {
...
+                let dag_message = request.req;
+
+                if dag_message.epoch == self.epoch() {
+                    // TODO: send message to DAG handler
+                    Ok(())
+                } else {
+                    monitor!(
+                        "process_different_epoch_dag_rpc",
+                        self.process_different_epoch(dag_message.epoch, peer_id)
+                    )
```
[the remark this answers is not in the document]
ibalajiarun: We dont treat QS message to decide epoch change. The block retrieval, I don't know.

PR #8987, opened 2023-07-06
[storage] split the replay_verify to multiple jobs

areshand:
### Description

Split the replay verify into multiple jobs so it can finish faster
### Test Plan

https://github.com/aptos-labs/aptos-core/actions/runs/5491816459/jobs/10008745766

### DIFF
--- a/.github/workflows/workflow-run-replay-verify.yaml
+++ b/.github/workflows/workflow-run-replay-verify.yaml
@@ -80,7 +80,12 @@ jobs:
   replay-verify:
     timeout-minutes: ${{ inputs.TIMEOUT_MINUTES || 720 }}
     runs-on: ${{ inputs.RUNS_ON }}
+    strategy:
+        matrix:
+          number: [0, 1, 2, 3, 4, 5, 6, 7] # runner number
     steps:
+      - name: Echo Runner Number
+        run: echo "Runner is ${{ matrix.number }}"
       - uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8 # pin@v3
         with:
           ref: ${{ inputs.GIT_SHA }}
@@ -101,7 +106,7 @@ jobs:
 
       - name: Run replay-verify in parallel
         shell: bash
-        run: testsuite/replay_verify.py
+        run: testsuite/replay_verify.py ${{ matrix.number }} 8 # first argument is the runner number, second argument is the total number of runners
         env:
           BUCKET: ${{ inputs.BUCKET }}
           SUB_DIR: ${{ inputs.SUB_DIR }}
--- a/testsuite/replay_verify.py
+++ b/testsuite/replay_verify.py
@@ -7,6 +7,7 @@ import os
 import subprocess
 import shutil
 import sys
+import math
 from multiprocessing import Pool, freeze_support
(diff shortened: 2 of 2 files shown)


rustielin (commented):
Are you sure the workflow run in your test plan is using your changes from this PR? looks like it's still just a single job. You will need to specify btw the git sha and workflow branch like this: 
image

Alternatively, you can test on this PR (or another canary PR) by temporarily changing this line: https://github.com/aptos-labs/aptos-core/blob/main/.github/workflows/replay-verify.yaml#L91 to a relative path like so: `uses: ./.github/workflows/workflow-run-replay-verify.yaml`

areshand:
> Are you sure the workflow run in your test plan is using your changes from this PR? looks like it's still just a single job. You will need to specify btw the git sha and workflow branch like this: image
> 
> Alternatively, you can test on this PR (or another canary PR) by temporarily changing this line: https://github.com/aptos-labs/aptos-core/blob/main/.github/workflows/replay-verify.yaml#L91 to a relative path like so: `uses: ./.github/workflows/workflow-run-replay-verify.yaml`

updated the test run using the alternative testing method. validated 8 runs are created

.github/workflows/workflow-run-replay-verify.yaml
```diff
@@ -80,7 +80,15 @@ jobs:
   replay-verify:
     timeout-minutes: ${{ inputs.TIMEOUT_MINUTES || 720 }}
     runs-on: ${{ inputs.RUNS_ON }}
+    strategy:
+        matrix:
+          number: [0, 1, 2, 3, 4, 5, 6, 7] # runner number
```
rustielin, line 85: nit: if possible, it would be nice to have some top-level config like NUM_RUNNERS or something. Otherwise, someone who wants to update this (do you expect this number of runners to be tuned?) has to do it in two places

testsuite/replay_verify.py
```diff
@@ -93,8 +94,16 @@ def main():
         "TXNS_TO_SKIP",
         "BACKUP_CONFIG_TEMPLATE_PATH",
     ]
+
     if not all(env in os.environ for env in REQUIRED_ENVS):
         raise Exception("Missing required ENV variables")
+    (runner_no, runner_cnt) = (int(sys.argv[1]), int(sys.argv[2])) if len(sys.argv) > 2 else (None, None)
```
rustielin, line 100: this is ok for now for 2 args, but any more we might want to use argparse or something

rustielin (approved):
Nice! Test plan looks good

LGTM % nits

msmouse (approved):
Thanks @rustielin  @areshand

msmouse:
Evenly divided loads might be very uneven? Let's do some test runs and see if it's an issue.. (I suspect We would like to manually partition between tasks)

.github/workflows/workflow-run-replay-verify.yaml
areshand: not found an easy way to do this. But if this turns out to be a problem. I will investigate how to auto generate the array here

rustielin:
Looks like lint is failing. Also TS SDK test failure was just fixed in main, so just update past it: https://github.com/aptos-labs/aptos-core/pull/9036

PR #8992, opened 2023-07-07
Update the python examples to use poetry install instead of poetry update

xbtmatt:
### Description

Currently the python examples have users run `poetry update`, whereas `poetry install` is only mentioned on the introduction page.

This normally wouldn't really be an issue, but lots of python users are newer and aren't familiar with `poetry`, so they don't realize they need to run `poetry install`.

The new Python SDK headers check to see what the current version of `aptos_sdk` is, so lots of people get this specific error when running the examples now:

```importlib.metadata.PackageNotFoundError: No package metadata was found for aptos-sdk```

Changing the examples to just run `poetry install` fixes this.

### DIFF
--- a/developer-docs-site/docs/tutorials/first-coin.md
+++ b/developer-docs-site/docs/tutorials/first-coin.md
@@ -69,7 +69,7 @@ Install the necessary dependencies:
 
 ```bash
 curl -sSL https://install.python-poetry.org | python3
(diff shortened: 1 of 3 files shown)


0xjinn (approved):
Nice find!

0xmaayan (approved):

PR #4788, opened 2022-10-05
[vm] Bring back vector functions gated by feature flag

wrwg:
### Description

A bug in the Move compiler leads to that `0x1::vector::foo` is mapped to the actual Move function whereas only `std::vector::foo` is mapped to the new vector opcodes. Thus users which do not use `std` have deployed code on chain which still calls into the vector functions which have been recently removed.

Before we fix the compiler, we need to add support for the legacy functions, because the bytecode is already on chain (i.e. there are calls to `0x1::vector::foo` in code on chain). Eventually we turn this off (after all users ported) with the feature flag.

This PR does the following:

1. Adds a new feature flag NO_LEGACY_VECTOR. If this is false (default) the old vector functions are again supported.
2. Copies the vector.rs module from the Move repo into our code, and modifies it such if the above flag is set, each function will abort.

There is some larger plumbing necessary to get feature flag down to natives (which will be also useful for other applications).

### Test Plan

New e2e test

- - -
This change is [Reviewable](https://reviewable.io/reviews/aptos-labs/aptos-core/4788)

PR #8996, opened 2023-07-07
[gas] introducing abstract gas algebra

vgao1996:
## Overview
This is a major refactoring of the gas metering system, introducing a zero-cost abstraction to represent gas expressions, allowing them to be handled symbolically, which is a prerequisite for tools like the incoming automated gas calibrator. (At this moment, there should be no change of behavior.)

This also comes up with some other improvements and reorganization that makes the whole system a bit less cumbersome. (We're cutting more than 1000-2000 lines boilerplate code, and still end up with a net reduction of code after duplicating a package + enabling new extension points + adding a ton of doc comments)

**Since this is a massive PR, please refer to the _Review Guide_ section for a detailed break-down!** (Unfortunately, since I'm changing the way the gas parameters are being managed, it's not that possible for me to split this into smaller PRs and migrate native functions incrementally.)

## Changes to Native Function Workflow
Previously, if you want to add a new native function, you would need to 
1. Define a new gas parameter struct
2. Build a closure that contains the gas parameter struct and pass it to your native function
3. Register the gas parameters in `aptos-gas` 

This workflow is now vastly simplified, and you just need to
1. Define the gas parameters in `aptos-gas-schedule`
2. Import your gas parameters from `aptos-gas-schedule` and use them in your native.

Now there is also a `SafeNativeBuilder` that takes care of the closure building process for you.

### `aptos-gas`
- This was split into a few smaller packages.

### `aptos-gas-algebra`
- Renamed from `gas-algebra-ext`.
- Now contains the units & type aliases that were originally defined in `aptos-gas`
- **Also home to the new types that define abstract gas algebra**

### `aptos-gas-schedule`
- Crate dedicated to defining the gas schedule + the gas feature version
- The contents were moved out from `aptos-gas`
- There is now a single unified macro to generate gas parameters for both the VM and natives
  - **What's new in this PR is that the macro now also generates a struct type for each gas parameter, which you can use in expressions**
- Instruction, txn and misc gas parameters are bundled into a struct called `VMGasParameters`, so that we don't have to always pass `AptosGasParameters` around
- I FINALLY GOT RID OF THE NOTORIOUS `* MUL` @alinush 

### `aptos-gas-schedule-updator`
- New home for the script that generates the gas schedule update proposal

### `aptos-gas-meter`
- The gas meter related traits (`AptosGasMeter`, `GasAlgebra`)
- Standard gas meter implementations (`StandardGasMeter`, `StandardGasAlgebra`)

### `aptos-native-interface`
- New crate that tries to define a more proper native function interface for the Aptos VM
- `SafeNativeContext`, `SafeNativeError` and related stuff
- `SafeNativeBuilder`: a new factory struct to simplify closure building
  - The closures now carry an `Arc<SharedData>`, which contains everything required to evaluate abstract gas expressions.

### `framework/move-stdlib`
- The natives were converted to use `SafeNativeContext` and the new gas parameters
- Notice that there is a hack to disable incremental gas metering for move-stdlib natives to maintain backward compatibility. We can get this enabled in the next gas feature version.

### `framework/src/natives`
- The natives now use the new gas parameters
- **Everyone who have written native functions, please double check the changes I made and make sure the gas formulae are correct!!!**
  - @alinush @zjma @gerben-stavenga @georgemitenkov @vusirikala 

### `framework/table-natives`
- Forked from `move-table-extension`, but uses `SafeNativeContext` and the new gas parameters
- Similar to `move-stdlib`, there is a hack to disable incremental gas metering for move-stdlib natives to maintain backward compatibility. We can get this enabled in the next gas feature version.

### `aptos-vm-types`
- I had to move `StorageGasParameters` and `ChangeSetConfig` into this crate to avoid cyclic dependency issues. @msmouse 

### Other Changes
The other changes are rather mechanical and are more of a response to the API changes I made in the crates above.

## TODOs
- [x] Docs
- [x] Fix the tests
- [x] Double check if I handled the "testing" feature flag correctly


aptos-move/aptos-gas-algebra/src/abstract_algebra.rs
```diff
@@ -0,0 +1,272 @@
...
+    fn evaluate(&self, _feature_version: u64, _env: &E) -> GasQuantity<Self::Unit> {
+        *self
+    }
+
+    fn visit(&self, visitor: &mut impl GasExpressionVisitor) {
+        visitor.quantity(*self)
+    }
+}
+
+impl<E, L, R, U> GasExpression<E> for GasAdd<L, R>
```
gerben-stavenga, line 119: I have the sinking feeling that I have a very different view on good code then everybody else here. I'm a big believer in typing and leveraging typing to maintain invariants, but typing also comes with costs. More code, more symbols/names you are polluting your mental ram with, boilerplate, restriction of functions accept them. The art of coding is to find the balance between leveraging static typing and maintaining information/invariants dynamically, such to produce efficient, readable and short/dense code.

All this code here generating a whole expression tree, where I think in the end the gas calculation is just inner_product(gas_params, size_vector). I feel we are overusing the typesystem for what is basic linear algebra. Note that in physics matrix entries often have different units, but in physics libraries you always store them as just numbers in arrays. The ability to operate on matrixes dynamically (looping) trumps the benefit of type safety that comes from encoding units in the typesystem.

gerben-stavenga: BTW. Taking the perspective of linear algebra also makes obvious a no-code solution. If we have some benchmark suite of tx's for which we know the desired cost. Let's name them t_i and c_i for i = 1...#benchmarks, then for each gas param g_j (j = 1 .. #gas params) we can calculate the gas of the transaction where we set g_j = 1 all other params 0. This gives a matrix C_{i j} for which it holds,  C.g is the vector of the gas costs of each benchmark when the parameters are set by g. So that means we want to chose g such that C.g \approx c by using something like least square.

aptos-move/framework/move-stdlib/src/natives/vector.rs
```diff
@@ -1,321 +0,0 @@
...
-        ("length", make_native_length(gas_params.length)),
-        ("push_back", make_native_push_back(gas_params.push_back)),
-        ("borrow", make_native_borrow(gas_params.borrow.clone())),
-        ("borrow_mut", make_native_borrow(gas_params.borrow)),
-        ("pop_back", make_native_pop_back(gas_params.pop_back)),
-        (
-            "destroy_empty",
-            make_native_destroy_empty(gas_params.destroy_empty),
-        ),
-        ("swap", make_native_swap(gas_params.swap)),
```
gerben-stavenga, line 317: how come you can delete this file

vgao1996: Great finding. I should've mentioned this in the review guide. Here's the full story
0. A long time ago, the vector operations only existed in the form of native functions.
1. Later, we decided to make them bytecode instructions. The compiler would translate all calls to these functions into bytecodes. However, the original native function declarations & implementations were kept. (This was still Facebook-era)
2. Before our mainnet launch, I was looking into the sanity of our native functions, and realized these legacy functions still existed in our VM, so I went ahead and got them removed. Since we were still depending on the move-stdlib crate in the public Move repo at that time, I did this through a dirty hack -- filtering out the vector natives by name. This means, even today, the vector natives are not part of the VM.
  - https://github.com/aptos-labs/aptos-core/blob/main/aptos-move/aptos-vm/src/natives.rs#L40
3. Later, I think Wolfgang discovered that due to a compiler bug, we had some code (on testnet?) that depended on these vector natives, so he tried to add them back, but for some reason I don't remember, the PR did not land. https://github.com/aptos-labs/aptos-core/pull/4788
4. A few months ago you forked the `move-stdlib` inside `aptos-core`
5. When I was working on this PR, I tried to convert the remaining natives to the safe variants, and discovered that I could not find the gas parameters for these vector natives. After some digging I realized they are not part of the VM, so I can safely delete them.

gerben-stavenga: If they are not used, I'm very happy to control-alt-delete them

aptos-move/aptos-gas-algebra/src/abstract_algebra.rs
vgao1996: > So that means we want to chose g such that C.g \approx c by using something like least square.

This is exactly the idea behind the automated gas calibration tool @wlawt and I are building ;)

Also thanks for the great write up! I think you started a very good discussion about the tradeoffs of having additional type safety. I'll chime in at little bit later and explain my views.

vgao1996: My apologies for the long write-up below, but you brought up a great topic and I think it’s worth discussing it in depth.

At a high level, I totally agree with you that one aspect of coding is to balance between achieving the desired level of security/correctness and providing straightforward user interfaces. In some rare cases you could have both, but more likely you need to sacrifice one for the other to some extent, and the choice is often determined by the nature of the project you are working with.

### On Security & Correctness

Maybe I'm being a bit too paranoid, but I would argue that **as a blockchain platform, we MUST take correctness & security in the most serious fashion, more so than almost all other commercial applications**. A mistake on our platform could not only result in people losing assets, but also hard or even impossible to revert.

Now coming back to the gas algebra (both concrete & abstract), having the unit embedded into the types of the numbers does allow us to catch certain classes of bugs at compile time. Here is one of the simplest examples

- If the gas parameters are just treated as normal integers, then it's totally possible for one to miswrite `base_cost + unit_cost * len` as `unit_cost + base_cost * len` and having this bug hiding in the system for months without being detected.
- If unit annotations are required, then the compiler would be able to see that the multiplication is invalid — you can’t multiply `gas` & `byte`. (Well, units like `gas * byte` are common in physics but they don’t make much sense in our context.) Similarly, it would prevent one from adding `base_cost` and `unit_cost`.
- Alternatively, you could see the unit annotations as a kind of redundancy — it composes a proof that your gas calculation is valid. The current approach also does it in a way with a nice property: **the code is both the instruction for execution and the proof itself, not requiring separate specs.**

One could argue that these are probably not the most serious bugs we could have, but still, I’d rather have us ruling them out completely with the help of the type system, than having to deal the aftermath one day. This is particularly appealing if you see it from the gas metering system’s perspective — there will be more and more of us interfacing with the system and one day someone will inevitably make a mistake.

You also brought up another interesting scenario: scientific computing/physics simulation. IMO this has totally different security & correctness requirements. This kind of number crunching happens on the scientist’s own computer/server and there are other ways to detect errors and redo the experiment. Things are different if you are controlling a plane, but in those cases, formal verification (even more proof) is often a hard requirement. The upcoming gas calibrator will actually take a similar approach, erasing the type info and put the numbers into matrices.

### What About User Experience & Maintainability?

Now I’m aware that this additional complexity makes it harder for clients to use the gas metering system — because the gas quantities can have different types, you would need some sort of heterogeneous collection if you want to store them conveniently. Otherwise, you would have to require the users to break down their formulae into individual terms and charge them separately:

```rust
context.charge(PARAM_A, 1)?;
context.charge(PARAM_B, len)?
..
```

My current solution to this problem is to offer such a heterogeneous data structure (which conveniently takes the form of an expression), and hide the complexities of the generation logic from the users. I would claim that this does result in a workflow is intuitive enough to follow — for each new gas parameter, you simply add a line to declare it, and then you can use it in an expression.

```rust
define_gas_params!(
  ..  
  [base: InternalGas, "base", 100],
  [per_byte: InternalGasPerByte, "per_byte", 10],
  ..
)

context.charge(BASE + PER_BYTE * len)?;
```

The user code here is actually rather dense and easy to read. The compiler will effectively expand this into (a more efficient version of) the following code

```rust
context.charge(BASE, 1)?;
context.charge(PER_BYTE , len)?
```

I do understand that some people, myself included, don’t like to be told to simply “do this, do that and then you’ll be fine”, and have the desire to fully understand the esoteric signature of an API before feeling comfortable about calling it. 

I don’t think I have a general solution to this, but at least I think it can be somewhat mitigated by 
1. Defining a clear contract between components so you can compartmentalize your mental memory and not having to keep all the boilerplate & restrictions in mind all the time
2. Having good documentation explaining not just the inter-component APIs, but also the inner workings of the system, so a curious reader could eliminate the sense of uncertainty they feel

This is also related to another issue I want to touch on: maintainability. To compensate for the user experience above, I had to sacrifice the overall complexity of the system. I do recognize it being a rather unpleasant tradeoff, and I don’t know the problem well enough to give a general solution, but at least the complexity is on the library writer’s side, not the users’. I hope some of the simplifications I made (e.g. unifying the two generation macros into one) and comments I’m planning to add will make it easier for other people to understand and maintain this part of the code. (I don’t know if it counts as a relief but we’re not at Haskell’s level of typestate programming yet…)

aptos-move/aptos-gas-algebra/src/abstract_algebra.rs
```diff
@@ -0,0 +1,273 @@
...
+    #[inline]
+    fn evaluate(&self, feature_version: u64, env: &E) -> GasQuantity<Self::Unit> {
+        (*self).evaluate(feature_version, env)
+    }
+
+    #[inline]
+    fn visit(&self, visitor: &mut impl GasExpressionVisitor) {
+        (*self).visit(visitor)
+    }
+}
```
vgao1996, line 122: I'm also planning to add a new type of node in the future, allowing tags to be attached to parts of an expression.

This has the potential to solve one problem I currently have with the gas profiler: the difficulty to break down the total storage fee into smaller items. I barely circumvented the problem by duplicating the logic, but there is huge risk that the gas meter & the profiler would go out of sync in the future, causing results to be inaccurate.

aptos-move/aptos-native-interface/src/builder.rs
```diff
@@ -0,0 +1,159 @@
...
+            }),
+            enable_incremental_gas_charging: true,
+        }
+    }
+
+    /// Controls the default incremental gas charging behavior of the natives created from this builder.
+    ///
+    /// Default: enabled.
+    pub fn set_incremental_gas_charging(&mut self, enable: bool) {
+        self.enable_incremental_gas_charging = enable;
```
zekun000, line 63: not related but it's not clear to me what this incremental gas charging means

aptos-move/aptos-native-interface/src/helpers.rs
```diff
@@ -0,0 +1,92 @@
...
+    ($args:ident, $t:ty) => {{
+        use $crate::reexports::move_vm_types::natives::function::{PartialVMError, StatusCode};
+        match $args.pop_back() {
+            Some(val) => match val.value_as::<$t>() {
+                Ok(v) => v,
+                Err(e) => return Err($crate::SafeNativeError::InvariantViolation(e)),
+            },
+            None => {
+                return Err($crate::SafeNativeError::InvariantViolation(
+                    PartialVMError::new(StatusCode::UNKNOWN_INVARIANT_VIOLATION_ERROR),
```
zekun000, line 18: this seems like nightmare to debug if it happens

aptos-move/aptos-vm-profiling/src/bins/run_move.rs
```diff
@@ -125,18 +127,27 @@ fn main() -> Result<()> {
...
         Identifier::new("Test").unwrap(),
         Identifier::new("create_signer").unwrap(),
         make_native_create_signer(),
     ));
-    natives.extend(move_table_extension::table_natives(
-        stdlib_addr,
-        move_table_extension::GasParameters::zeros(),
-    ));
+    // FIXME: Find a way to re-enalbe this.
+    //natives.extend(aptos_table_natives::table_natives(
```
zekun000, line 147: why table extension can't be enabled here?

aptos-move/aptos-vm-types/src/storage.rs
```diff
@@ -178,10 +184,55 @@ impl StoragePricingV2 {
     }
 }
 
+// No storage curve. New gas parameter representation.
+#[derive(Debug, Clone)]
+pub struct StoragePricingV3 {
```
zekun000, line 189: is this something new added in this PR or moved from somewhere else?

aptos-move/aptos-gas-algebra/src/abstract_algebra.rs
```diff
@@ -0,0 +1,273 @@
...
+///   - `gas_param<B>()`
+///   - `quantity(50)`
+///   - `mul()`
+///   - `add()`
+pub trait GasExpressionVisitor {
+    fn add(&mut self);
+
+    fn mul(&mut self);
+
+    fn gas_param<P>(&mut self);
```
zekun000, line 64: is this used anywhere? if so we should probably unify the name with `env` (from what i understand is the params)

aptos-move/aptos-gas-algebra/src/abstract_algebra.rs
```diff
@@ -0,0 +1,273 @@
...
+///   - `quantity(1)`
+///   - `quantity(2)`
+///   - `add()`
+/// - `A + B * 50`
+///   - `gas_param<A>()`
+///   - `gas_param<B>()`
+///   - `quantity(50)`
+///   - `mul()`
+///   - `add()`
+pub trait GasExpressionVisitor {
```
zekun000, line 59: is this actually used? I don't see any reference on this

zekun000 (commented):
did a first pass mainly on the abstract algebra

aptos-move/aptos-gas-algebra/src/algebra.rs
```diff
@@ -0,0 +1,56 @@
...
+
+/***************************************************************************************************
+ * Unit Conversion
+ *
+ **************************************************************************************************/
+impl ToUnit<Octa> for APT {
+    const MULTIPLIER: u64 = 1_0000_0000;
+}
+
+impl ToUnitFractional<APT> for Octa {
```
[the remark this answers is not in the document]
vgao1996: Good catch. I'll try.

aptos-move/aptos-gas-algebra/src/algebra.rs
```diff
@@ -0,0 +1,56 @@
...
+
+/// Unit of the Aptos network's native coin.
+pub enum APT {}
+
+/// Alternative unit of the Aptos network's native coin. 1 Octa = 10^-8 Aptos coins.
+pub enum Octa {}
+
+pub type Gas = GasQuantity<GasUnit>;
+
+pub type GasScalingFactor = GasQuantity<UnitDiv<InternalGasUnit, GasUnit>>;
```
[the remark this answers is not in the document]
vgao1996: Because there is an upstream gas algebra module in `third_party/move/move-core/types` defining the basics. `InternalGas` is also being referenced in the Move VM, even though it's merely used as a medium for handling native function costs.

aptos-move/aptos-gas-algebra/src/abstract_algebra.rs
```diff
@@ -0,0 +1,273 @@
...
+// ---------
+// E | &T: U
+impl<E, T> GasExpression<E> for &T
+where
+    T: GasExpression<E>,
+{
+    type Unit = T::Unit;
+
+    #[inline]
+    fn evaluate(&self, feature_version: u64, env: &E) -> GasQuantity<Self::Unit> {
```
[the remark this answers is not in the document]
vgao1996: Thought about that. Merging the feature version into env would require one to store it and some parameters in the same env struct, which is rarely the way clients organize them. Therefore it's actually cleaner to take them as separate parameters.

aptos-move/aptos-gas-algebra/src/abstract_algebra.rs
vgao1996: It's not used in this PR, but @wlawt's gas calibrator will use this interface do record abstract gas usage.

aptos-move/aptos-gas-algebra/src/abstract_algebra.rs
vgao1996: 1. Unused right now, but required for the gas calibrator & future analysis tools.
2. `P` is not the same as `env`. It is instantiated with types representing each individual gas parameters. When you add a gas parameter to `aptos-gas-schedule`, my macro generates a new type for it (e.g. LD_CONST_BASE), which you can use in gas expressions.

zekun000:
also I assume the operator priority follows arithmetic i.e mul is evaluated before add?
A + B * C would result in
```
GasAdd {
  left: A,
  right: GasMul { left: B, right: C}
} 
```

aptos-move/aptos-vm-types/src/storage.rs
vgao1996: It's newly added. Although its logic should be identical to V2 at the latest gas feature version. I created V3 so it can return abstract usage. (Notice that V2 is still returning a concrete number. It's a bit tricky to unify the two code paths.)

vgao1996:
@zekun000 you are correct about operator precedence.

aptos-move/aptos-gas-algebra/src/abstract_algebra.rs
zekun000: so it's individual gas param while env is the collection of all? it's not very clear to me what this function means for a visitor though

aptos-move/aptos-vm-profiling/src/bins/run_move.rs
vgao1996: Originally, we were using the table extension in third_party move, and that extension has special support in move_vm_test_utils that can be enabled via a cargo feature flag.

 In this PR, I forked the table extension in `aptos-move`, breaking that support. I'll fix this by making the aptos-core fork importing the relevant types (`NativeTableContext`,`TableResolver`) from the one. on third_party move, instead of making a copy of them. (We still need to fork the native functions themselves)

aptos-move/aptos-gas-algebra/src/abstract_algebra.rs
vgao1996: Ah, take a look at the doc comments I wrote for this function above. The visitor callbacks are invoked in an order following the Reverse Polish representation of the expression.

aptos-move/aptos-native-interface/src/helpers.rs
vgao1996: Thanks to @gerben-stavenga's recent work, we now include a stack trace as part of the invariant violation error, so I wouldn't worry too much about debuggability for now.

Fully agreed this API is terrible though. It was created because Move VM didn't offer good support in hte first place. The plan is to replace it with some proper user-friendly APIs in the future.

aptos-move/aptos-gas-algebra/src/abstract_algebra.rs
zekun000: yes I understand the order of invoke, but I don't know what it means to call `gas_param`

vgao1996: If you see the arithmetic expression as an AST, then `gas_param<B>` is called when we are visiting a node corresponding to gas parameter B.
```
A + B * 5
    ^ It's this B
```

aptos-move/aptos-native-interface/src/builder.rs
vgao1996: The native function API provided by the Move VM is rather primitive in terms of gas metering: it requires a native functions to calculate the total gas cost itself, and charge it after it returns. This means we are doing work before charging gas, which is not good. 

`SafeNativeContext` was initially made to address this problem by maintaining a counter of the current gas usage, and it aborts early when the accumulative usage has exceeded the remaining balance. This was neat cuz it required minimal changes to the Move VM itself (only need to expose the remaining balance). However, this is still a behavior change, so if we want to make the switch, we'll need to gate it properly.

vgao1996: Added a line of comment pointing to `SafeNativeContext::set_incremental_gas_charging`, which explains this in details.

aptos-move/framework/src/natives/cryptography/algebra/arithmetics/mod.rs
```diff
@@ -12,13 +12,13 @@ pub mod sub;
...
-        let handle_2 = safely_pop_arg!($args, u64) as usize;
-        let handle_1 = safely_pop_arg!($args, u64) as usize;
+    ($context:expr, $args:ident, $ark_typ:ty, $ark_func:expr, $gas:expr) => {{
+        let handle_2 = aptos_native_interface::safely_pop_arg!($args, u64) as usize;
+        let handle_1 = aptos_native_interface::safely_pop_arg!($args, u64) as usize;
         safe_borrow_element!($context, handle_1, $ark_typ, element_1_ptr, element_1);
         safe_borrow_element!($context, handle_2, $ark_typ, element_2_ptr, element_2);
         $context.charge($gas)?;
-        let new_element = element_1.$ark_func(element_2);
+        let new_element = $ark_func(element_1, element_2);
```
alinush, line 21: Was there a problem before, or why this change?

aptos-move/framework/src/natives/cryptography/algebra/arithmetics/scalar_mul.rs
```diff
@@ -65,20 +70,18 @@ fn ark_msm_window_size(num_entries: usize) -> usize {
...
-    cost_add: InternalGasPerArg,
-    cost_double: InternalGasPerArg,
-    num_entries: usize,
-) -> InternalGas {
-    let window_size = ark_msm_window_size(num_entries);
-    let num_windows = (255 + window_size - 1) / window_size;
-    let num_buckets = 1_usize << window_size;
-    cost_add * NumArgs::from(((num_entries + num_buckets + 1) * num_windows) as u64)
-        + cost_double * NumArgs::from((num_buckets * num_windows) as u64)
+macro_rules! ark_msm_bigint_wnaf_cost {
```
alinush, line 73: Why convert this to a macro? :-/

aptos-move/framework/src/natives/cryptography/algebra/hash_to_structure.rs
```diff
@@ -50,30 +50,34 @@ macro_rules! suite_from_ty_arg {
...
-    mapping_base: InternalGasPerArg,
-    mapping_per_byte: InternalGasPerArg,
-) -> InternalGas {
-    // DST shortening as defined in https://www.ietf.org/archive/id/draft-irtf-cfrg-hash-to-curve-16.html#name-using-dsts-longer-than-255-.
-    let dst_shortening_cost = if dst_len <= 255 {
-        InternalGas::zero()
-    } else {
-        dst_shortening_base + dst_shortening_per_byte * NumBytes::from((17 + dst_len) as u64)
-    };
+macro_rules! hash_to_bls12381gx_cost {
```
alinush, line 53: Same Q here about the conversion to macro, if the answer is different. I suspect it has something to do with the type of the abstract gas expressions.

aptos-move/aptos-gas-meter/src/traits.rs
```diff
@@ -0,0 +1,221 @@
...
+};
+use aptos_vm_types::storage::StorageGasParameters;
+use move_binary_format::errors::{Location, PartialVMResult, VMResult};
+use move_core_types::gas_algebra::{InternalGas, InternalGasUnit, NumBytes};
+use move_vm_types::gas::GasMeter as MoveGasMeter;
+use std::fmt::Debug;
+
+/// An abstraction of the part of the gas meter that manages the core algebra,
+/// defining how the gas charges should be handled.
+pub trait GasAlgebra {
```
gerben-stavenga, line 17: I don't understand the structure. If it's an algebra (as we discussed before I consider "algebras" over engineering but that's not the point here) then I dont understand what feature_version/charge_execution etc are doing in this trait. That's not algebra

aptos-move/framework/src/natives/cryptography/algebra/hash_to_structure.rs
vgao1996: > I suspect it has something to do with the type of the abstract gas expressions.

Precisely. It's possible to write this as a function, but I found it a bit too verbose for this context since you have to specify the types.

aptos-move/framework/src/natives/cryptography/algebra/arithmetics/mod.rs
alinush: Hm. Strange. Looping in @zjma since he should be made aware of this.

gerben-stavenga (approved):

aptos-move/framework/src/natives/cryptography/algebra/arithmetics/mod.rs
zjma: I tried the old macro and it builds and tests fine...
@vgao1996 here's a revert sub-PR: https://github.com/vgao1996/aptos-core/pull/2/files

zjma (commented):
Validated the gas formulas in cryptography/algebra.

aptos-move/framework/src/natives/cryptography/algebra/arithmetics/mod.rs
vgao1996: Let me double check

vgao1996: Okay, looks like it's a rust-analyzer problem. I've reverted these changes.

runtian-zhou (commented):
I guess my first question is how do we make sure this code is equivalent to the previous version? For a major refactor like this should we run a replay to make sure the code will behave the same?

vgao1996:
@runtian-zhou yes, I'll trigger a replay after the tests finish running. Some new natives got added so I had to rebase an fix those.

aptos-move/aptos-vm-types/src/storage.rs
runtian-zhou: Is it possible to move this change to a separate PR? Seems an orthogonal change?

runtian-zhou (approved):

vgao1996:
Replay task: https://github.com/aptos-labs/aptos-core/actions/runs/5626325148

@runtian-zhou I'm planning to hit the merge button once this passes. Does it sound good to you?

vgao1996:
Replay attempt 2: https://github.com/aptos-labs/aptos-core/actions/runs/5626694153

Removed an unintended change to the on-chain gas feature names.

aptos-move/aptos-gas-algebra/src/abstract_algebra.rs
```diff
@@ -0,0 +1,273 @@
...
+    type Unit;
+
+    /// Evaluates the expression within the given environment to a concrete number.
+    fn evaluate(&self, feature_version: u64, env: &E) -> GasQuantity<Self::Unit>;
+
+    /// Traverse the expression in post-order using the given visitor.
+    /// See [`GasExpressionVisitor`] for details.
+    fn visit(&self, visitor: &mut impl GasExpressionVisitor);
+
+    /// Performs a division on the unit of the expression.
```
wrwg, line 29: Sentence is unclear. "division on the unit"?

aptos-move/aptos-gas-algebra/src/abstract_algebra.rs
```diff
@@ -0,0 +1,273 @@
...
+
+/***************************************************************************************************
+ * Gas Expression & Visitor
+ *
+ **************************************************************************************************/
+/// Trait representing an abstract expression that can be used to calculate some gas amount.
+///
+/// It carries a type parameter `E`, indicating an environment in which the expression can be
+/// evaluated/materialized.
+pub trait GasExpression<E> {
```
wrwg, line 19: I think `E` might be cleaner to be `type Env`.

aptos-move/aptos-gas-algebra/src/abstract_algebra.rs
wrwg: Unclear what the `P` is needed for, specifically as it is nowhere used in this trait.

aptos-move/aptos-gas-algebra/src/abstract_algebra.rs
```diff
@@ -0,0 +1,273 @@
...
+    fn per<U>(&mut self);
+}
+
+/***************************************************************************************************
+ * Built-in Gas Expressions
+ *
+ **************************************************************************************************/
+/// Representing the addition of two gas expressions.
+#[derive(Debug, Clone)]
+pub struct GasAdd<L, R> {
```
wrwg, line 77: Why is this so super-generic and not a `GasExpression`? You are defining an arbitrary binary node with this type. This somehow can't make sense as an abstraction.

aptos-move/aptos-gas-algebra/src/abstract_algebra.rs
```diff
@@ -0,0 +1,273 @@
...
+ * Gas Expression Impl
+ *
+ **************************************************************************************************/
+// Notation:
+//   E | T: U means T is a valid gas expression with unit U under environment E.
+
+// E | T: U
+// ---------
+// E | &T: U
+impl<E, T> GasExpression<E> for &T
```
wrwg, line 107: Can't you give default impls in `GasExpression` instead of this indirection here?

wrwg (commented):
I feel the trait based implementation of Gas expressions is not ideal, though I may not see all requirements here. In general, if defining logic like this, I'd always prefer the clarity and simplicity of Rust abstract data types/enums. We are no defining here a language with a never ending expanding set of operators, but a small and well-known set. So I wonder why not just something along the line as below:

```
pub enum GasExp {
   Add(Box<GasExp>, Box<GasExp>),
   Mul(Box<GasExp>, Box<GasExp>),
   ...
   Param(Ident)
}

impl GasExp {
  fn eval<E, U>(&self, e: E) -> Quantity<U> {
     match self {
        GasExp::Add(l, r) => l.eval(e) + r.eval(e),
        ...
     }
  }
}  
```

You wouldn't really need a `visit` because you can just match over the AST.

vgao1996:
@wrwg Great question! The current design stems from the need to achieve both of the following goals at the same time:
1. Native function/VM authors write each formula exactly once, without additional annotations that duplicate the logic.
2. There should be little to no performance impact for the validator code path.
3. The gas formula representation should provide an extension point for symbolic interpretation.

The enum approach requires frequent allocation, and prevents optimizations. This is not a problem for the static representation, which is a zero-cost abstraction the compiler is capable of optimizing out.

You can find a more detailed explanation in the design doc.
[1f8e1655-7c28-49c1-8d44-240f12cacbae_Abstract_Gas_Algebra.pdf](https://github.com/aptos-labs/aptos-core/files/12137129/1f8e1655-7c28-49c1-8d44-240f12cacbae_Abstract_Gas_Algebra.pdf)

aptos-move/aptos-gas-algebra/src/abstract_algebra.rs
vgao1996: Not sure I understand this question. This blanket implementation just makes it possible to use references at locations that require (owned) gas expressions.

aptos-move/aptos-gas-algebra/src/abstract_algebra.rs
vgao1996: See my explanation for static vs dynamic.

aptos-move/aptos-gas-algebra/src/abstract_algebra.rs
vgao1996: Again, all of the methods in the visitor trait are callbacks.

You as the client implement a visitor, and I as the VM will call your methods in a particular sequence, traversing a gas expression.

Here `P` is just a function parameter that tells you (the client) which gas parameter it is currently visiting. Conceptually it is no different from
```
fn gas_param(name: &str);
```
It's just that in this static representation, we're using rust types, rather than strings to name gas parameters.

vgao1996:
@wrwg You are however correct that the static representation is harder to work with, requiring one to use the visitor pattern, and sometimes deal with heterogeneous collections. This is why we will actually have both representations in the future -- applications that are not performance-critical can benefit from the flexibility of the dynamic representation, which is built from the static one. (@wlawt's upcoming gas calibrator will be the first example)

wrwg (approved):

davidiw:
Curious why we didn't remove the MUL here and shift the maximum gas in a transaction by that amount?

PR #9002, opened 2023-07-08
[move-compiler v2] Porting remaining typing tests and refining type checker

wrwg:
This ports the remaining 200+ tests of the v1 compiler `typing` testsuite and refines the v2 compiler to handle those correctly. Major changes include:

- Introducing ordering in type unification. This is necessary to deal with conversions like `&mut T` to `&T`. The general mechanism will allow us to have also other conversions if we need down the road.
- Introducing a mechanism to define a default for type inference via constraints. If a type variable has no unique instantiation, then the default is chosen. For example, the default for integer constants is `u64`.
- Implementing the remaining expansion AST constructs from the Move parser. All expression variants are now handled.

With this, we should have a fairly feature complete and tested type checker. This doesn't do, however:

- visibility checking (private/friend)
- mutation checking (whether something which is mutated can be actually mutated)
- ability (copy, drop, ...) checking
- reference safety (borrow) checking

Those phases have been partly included in the v1 type checker. For the v2 compiler, the plan is to handle them instead on bytecode level, similar as in the bytecode verifier.


third_party/move/move-compiler-v2/tests/checking/typing/bind_unpack_references_invalid.exp
```diff
@@ -1,24 +1,24 @@
 
 Diagnostics:
-error: expected `&u64` but found `integer` in lvalue
+error: `integer` not compatible with `&u64` in lvalue
   ┌─ tests/checking/typing/bind_unpack_references_invalid.move:7:9
```
brmataptos, line 4: Comparing this with the old (compiler v1) .exp file, I think this would be clearer if we added annotations to the struct declaration/assignment let to state the types which are inferred.  For example:

`     let R { s1: S { f }, s2 } = &R { s1: S{f: 0}, s2: S{f: 1} };`
Contains sub-assignments:
`      f = &0;`
from which we infer f: &uint64.
`     s2 = &S(f:1);`
from which we infer s2 : &M::S.

So this error message would be better as:
error: `integer` not compatible with (inferred) type `&64` of lvalue f.

But if we can't distinguish inferred and declared types, this would still be better as:

error: `integer` type of rvalue not compatible with type `&u64` of lvalue in assignment.

third_party/move/move-compiler-v2/tests/checking/typing/bind_wrong_type.exp
```diff
@@ -1,18 +1,18 @@
 
 Diagnostics:
-error: expected `M::R` but found `M::S` in lvalue
+error: expected `M::S` but found `M::R` in lvalue
```
brmataptos, line 3: Again, this seems it would be clearer as:

error: assignment to lvalue of type `M::S` expected the same type but found `M::R`

third_party/move/move-compiler-v2/tests/checking/typing/block_empty_invalid.exp
```diff
@@ -1,18 +1,18 @@
 
 Diagnostics:
-error: expected `u64` but found `()` in unit expression
+error: expected `u64` but found `()` in expression
```
brmataptos, line 3: I'm concerned about the potential confusion between {} and ().  Are these the same value?  Structs are not the same as Tuples.  Oh, this is an empty expression block {} returning ().

In any case, better to say something like:

**error: expression of unit type '()' not compatible with type u64.**

Even the old error messages weren't great.  You don't expect 'u64', you expect an expression of type 'u64' here.

third_party/move/move-compiler-v2/tests/checking/typing/block_single_expr_invalid.exp
```diff
@@ -1,6 +1,6 @@
 
 Diagnostics:
-error: expected `bool` but found `integer` in constant
+error: `integer` not compatible with `bool` in constant
```
brmataptos, line 3: error: expression of type `integer` not compatible with declared type `bool` in constant.

third_party/move/move-compiler-v2/tests/checking/typing/break_any_type.move
```diff
@@ -0,0 +1,20 @@
+module 0x8675309::M {
+    struct Coin {}
+
+    fun t0() {
+        while (true) {
+            0 + break;
```
brmataptos, line 6: What happens with

const C = break;

?

brmataptos (commented):
Comments on a few error messages.

third_party/move/move-compiler-v2/tests/checking/typing/derefrence_invalid.exp
```diff
@@ -0,0 +1,79 @@
+
+Diagnostics:
+error: expected `bool` but found `u64` in name expression
+  ┌─ tests/checking/typing/derefrence_invalid.move:6:11
+  │
+6 │         (*x : bool);
```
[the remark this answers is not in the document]
wrwg: Good catch. This one wasn't easy but should be better now.

third_party/move/move-compiler-v2/tests/checking/typing/duplicate_function_parameter_names.exp
```diff
@@ -0,0 +1,25 @@
+
+Diagnostics:
+error: duplicate declaration of `x`
+  ┌─ tests/checking/typing/duplicate_function_parameter_names.move:2:21
+  │
+2 │     fun foo(x: u64, x: u64) {}
+  │                     ^
+
+error: previous declaration of `x`
```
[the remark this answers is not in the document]
wrwg: Done

wrwg (commented):
Thanks for the review! Not all things could be addressed in this PR but most of them.

third_party/move/move-compiler-v2/tests/checking/typing/bind_unpack_references_invalid.exp
brmataptos: Allowing the source code to be made so confusing isn't great for security.

I can imagine someone sharing a module with published source code for clients to scrutinize that looks ok but which does something nefarious with confusing code.

third_party/move/move-compiler-v2/tests/checking/typing/bind_wrong_type.exp
brmataptos: Ok, I filed a new issue [to track this.](https://github.com/aptos-labs/aptos-core/issues/9003).

third_party/move/move-compiler-v2/tests/checking/typing/block_empty_invalid.exp
brmataptos: How about "Expected `uint64` but found expression of type `()`."

Or maybe "value of type `()`"

wrwg: Again, this not a manually created message which you can just change like you are suggesting:

- The message the type unifier generates is `expected X but found Y`. The unifier also produces a couple of different messages, this is just one of them.
- Then this message as generated by the unifier (whatever it is) is augmented by a suffix 'in XYZ' which gives context in which the type unification failure occured.

So its `generic_type_unification_error + context_message`. 

Also please note that little of this is new but just the type checker of the specification language how it has been working since a while, with little complaints. So I did not add all this in this PR, but only adapt it.

wrwg: After reviewing the error messages, @brmataptos , I actually removed all those `in XYZ context` message suffices. The reason is that the last fix based on your feedback -- where the `^^` was not complete or on the wrong place -- actually was a bit more fundamental. Basically, when we add a variable binding with a constraint (say `x where SomeReference(y)`), and this constraint is later resolved as unsat during type unification, we need to report the error back at the point where the constraint was created -- not where we perform the unification step. This makes all those 'in lvalue', 'in constant', `in pack expression' messages potentially inaccurate. I decided to better drop them for now.

gerben-stavenga (approved):

third_party/move/move-compiler-v2/tests/checking/typing/break_any_type.move
gerben-stavenga: I guess it lets you write 

let x = if (cond) { 0 } else { break; };

okay it makes sense

wrwg: Yes, the above is a good example.

Frankly, I believe all those language design details are just taken over from Rust. I also weren't aware of those and had to consult the Move book for it.

wrwg: PS: however it must be `break` not `break;`

third_party/move/move-compiler-v2/tests/checking/typing/while_condition_invalid.exp
```diff
@@ -0,0 +1,49 @@
+
+Diagnostics:
+error: expected `bool` but found `()` in expression
```
brmataptos: I still think it would be better to say "expected `bool` but found expression of type `()`"

third_party/move/move-compiler-v2/src/bytecode_generator.rs
```diff
@@ -266,6 +267,7 @@ impl<'env> Generator<'env> {
             | Operation::MoveFrom
             | Operation::MoveTo
             | Operation::Abort
+            | Operation::Vector
```
brmataptos, line 270: Why do we need an Operation::Vector now?

third_party/move/move-model/src/ty.rs
```diff
@@ -870,58 +992,87 @@ impl Substitution {
                 }
             },
             (Type::Reference(k1, ty1), Type::Reference(k2, ty2)) => {
-                let ty = self.unify(sub_variance, ty1, ty2)?;
-                if *k1 == *k2 {
-                    return Ok(Type::Reference(*k1, Box::new(ty)));
-                } else {
+                let ty = self.unify(sub_variance, order, ty1, ty2)?;
+                let k = if variance.is_impl_variance() {
```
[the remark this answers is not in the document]
wrwg: No, this is actually intended. The unification of the types bails out (the `?`), and the intent is to reported this mismatch first, before looking at difference of `&` vs `&mut`.

third_party/move/move-model/src/ast.rs
```diff
@@ -420,6 +420,8 @@ pub enum ExpData {
     LoopCont(NodeId, bool),
     /// Assignment to a pattern. Can be a tuple pattern and a tuple expression.
     Assign(NodeId, Pattern, Exp),
+    /// Mutation of a lhs reference, as in `*lhs = rhs`.
```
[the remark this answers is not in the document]
wrwg: I think it is clear if you look at the types, as we discussed offline. Assign is pattern based and Mutate is expression based.

third_party/move/move-model/src/builder/exp_builder.rs
```diff
@@ -17,7 +17,7 @@ use crate::{
     symbol::{Symbol, SymbolPool},
     ty::{
         Constraint, PrimitiveType, ReferenceKind, Substitution, Type, TypeDisplayContext, Variance,
-        BOOL_TYPE,
+        WideningOrder, BOOL_TYPE,
```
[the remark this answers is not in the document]
wrwg: Widening is a notion in type checking. It means a conversion is possible without any information loss. It is documented where the type is defined, so I suggest to look there. I c&p it here for you:

```
/// Determines an ordering for unification. Combined with `Variance`, determines in which
/// direction automatic type conversion rules are to be applied.
#[derive(Clone, Copy, Debug, PartialEq, Eq)]
pub enum WideningOrder {
    /// The left type can be widened to the right one.
    LeftToRight,
    /// The right type can be widened to the left one.
    RightToLeft,
    /// The smallest common type into which both left and right type can be widened.
    Join,
}
```

third_party/move/move-model/src/builder/exp_builder.rs
```diff
@@ -127,18 +125,29 @@ impl<'env, 'translator, 'module_translator> ExpTranslator<'env, 'translator, 'mo
...
+            self.mode,
+            ExpTranslationMode::Spec | ExpTranslationMode::TryImplAsSpec
+        )
+    }
+
     pub fn type_variance(&self) -> Variance {
         if self.mode == ExpTranslationMode::Impl {
             // When translating Move implementation code, no variance is allowed
-            Variance::NoVariance
+            Variance::ShallowImplVariance
```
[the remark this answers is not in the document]
wrwg: Good catch, fixing...

third_party/move/move-compiler-v2/tests/checking/typing/loop_body_invalid.exp
```diff
@@ -0,0 +1,37 @@
...
+3 │         loop 0
+  │              ^
+
+error: expected `()` but found `bool` in expression
+  ┌─ tests/checking/typing/loop_body_invalid.move:7:14
+  │
+7 │         loop false
+  │              ^^^^^
+
+error: expected `()` but found `address` in value
```
brmataptos: For this and the following loop bodies, it would be better to have the scope of the complaint be the entire argument to loop.  Not sure how you're generating these, but in a situation where there is a contextual constraint on the type of a subexpression, wouldn't it be better to complain at the context level, rather than the point where the wrong value is generated?

third_party/move/move-compiler-v2/tests/checking/typing/subtype_bind_invalid.exp
```diff
@@ -0,0 +1,31 @@
+
+Diagnostics:
+error: mutability mismatch (&mut != &) in expression
+  ┌─ tests/checking/typing/subtype_bind_invalid.move:5:27
+  │
+5 │         let x: &mut u64 = &0;
```
brmataptos, line 6: Why is &0 even syntactically correct?

third_party/move/move-compiler-v2/tests/checking/typing/subtype_return_invalid.exp
```diff
@@ -0,0 +1,37 @@
+
+Diagnostics:
+error: mutability mismatch (&mut != &) in name expression
```
brmataptos: Again, mentioning context (fun body) here is more useful than the kind of expression (name expression).

Same below.

brmataptos: or maybe "fun return value" if that's possible.

third_party/move/move-compiler-v2/tests/checking/typing/type_variable_join_threaded_unpack_assign_invalid.exp
```diff
@@ -0,0 +1,7 @@
+
+Diagnostics:
+error: `integer` not compatible with `bool` in name expression
+   ┌─ tests/checking/typing/type_variable_join_threaded_unpack_assign_invalid.move:36:9
+   │
+36 │         f1
```
brmataptos: Sorry for being redundant, but it occurs to me that what we need is to do type checking bottom-up, so that "f1" is fine, but when we see the context, we can say `integer` not compatible with `bool` in sequence expression value.  or maybe you need something even term-specific.

third_party/move/move-compiler-v2/tests/checking/typing/while_body_invalid.exp
```diff
@@ -0,0 +1,37 @@
+
+Diagnostics:
+error: `integer` not compatible with `()` in constant
```
brmataptos: To clarify yet again, "in constant" is not useful here.  

Expected `()` for while loop body but found `bool` constant.

third_party/move/move-compiler-v2/tests/checking/typing/vector_mismatched_args.exp
```diff
@@ -0,0 +1,31 @@
+
+Diagnostics:
+error: `integer` not compatible with `bool` in vector expression
```
brmataptos: This error message could be improved, but it's not urgent.  Ideally you'd collect the various parameter types and mention them all at once:

error: vector arguments not compatible: `integer` not compatible with `bool`

third_party/move/move-compiler-v2/tests/checking/typing/vector_mismatched_args.exp
```diff
@@ -0,0 +1,31 @@
...
+  ┌─ tests/checking/typing/vector_mismatched_args.move:7:19
+  │
+7 │         vector[0, false];
+  │                   ^^^^^
+
+error: expected `u64` but found `u8` in vector expression
+  ┌─ tests/checking/typing/vector_mismatched_args.move:8:21
+  │
+8 │         vector[0u8, 0u64, 0u128];
```
brmataptos, line 12: This one is very confusing, and is more urgent.

third_party/move/move-compiler-v2/tests/checking/typing/while_condition_invalid.exp
wrwg: Its not always an expression. It can be a pattern or a constant as well. It depends on the context.

third_party/move/move-compiler-v2/src/bytecode_generator.rs
wrwg: Because `vector[...]` is a new expression primitive which wasn't handled here before.

third_party/move/move-compiler-v2/tests/checking/typing/lambda.exp
```diff
@@ -0,0 +1,57 @@
+
+Diagnostics:
+error: `M::reduce` is a function and not a macro
```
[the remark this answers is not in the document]
wrwg: Macro is a concept in the language. The error messages cannot explain basic concepts of the language to you. If you find the book isn't good with this, lets open a bug to fix the book.

third_party/move/move-compiler-v2/tests/checking/typing/loop_body_invalid.exp
```diff
@@ -0,0 +1,37 @@
+
+Diagnostics:
+error: `integer` not compatible with `()` in constant
```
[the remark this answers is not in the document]
wrwg: You are commenting on outdated code. The 'compatible' messages are gone since a while, the 'in constant' too.

third_party/move/move-compiler-v2/tests/checking/typing/loop_body_invalid.exp
wrwg: Technically not easily possible at the moment. Also there are tradeoffs. Sometimes it is better for the error messages to push the requirement from the context down to the leaves, sometimes not. I think more often the former is the case, which is currently happening.

third_party/move/move-compiler-v2/tests/checking/typing/subtype_return_invalid.exp
wrwg: Outdated code. Those 'in' messages are all gone because of another change you requested (rightfully), they can't be longer precise.

third_party/move/move-compiler-v2/tests/checking/typing/type_variable_join_threaded_unpack_assign_invalid.exp
wrwg: I do not agree with you, see above. Actually that an expected type is pushed from the context into the checker is a feature which is very useful and leads to better error messages in many cases.

We can also implement an alternative type checker which works bottom up and compare, but this is likely a month of work or more...

third_party/move/move-compiler-v2/tests/checking/typing/while_body_invalid.exp
wrwg: And to say it again, you are commenting on code which was outdated some time ago.

third_party/move/move-compiler-v2/tests/checking/typing/vector_mismatched_args.exp
wrwg: Outdated, this time widely sometime yesterday.

third_party/move/move-compiler-v2/tests/checking/typing/vector_mismatched_args.exp
wrwg: This must be outdated as well, because in my source, the direction is correct. BTW, instead of saying "this is confusing" you could just identify the bug and say that the types are obviously in the wrong order, and it should be `expected u8 but found u6`. Confusing in this context sounds like this would be intended. Indeed it is not.

third_party/move/move-compiler-v2/tests/checking/typing/lambda.exp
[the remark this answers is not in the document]
brmataptos: i don't know what the language is other than the implementation and the Move Book.  There is no macro in Move other than assert!.  Though the compiler implementation does parse ! in a call as "is_macro".

reduce! in the code is more likely a typo for reduce1 than an attempt at a macro use, except for Rust programmers.

Anyway, I've filed a bug.

third_party/move/move-compiler-v2/tests/checking/typing/block_empty_invalid.exp
brmataptos: Ok, I see it now.

third_party/move/move-compiler-v2/tests/checking/typing/loop_body_invalid.exp
brmataptos: Ok, well it looks better in latest commit with context removed.

brmataptos (approved):

third_party/move/move-compiler-v2/tests/checking/typing/vector_mismatched_args.exp
brmataptos: Sorry, my knowledge of Move typing is incomplete, so I didn't immediate realize that it was so simple as being the right types but in the wrong order.

third_party/move/move-compiler-v2/tests/checking/typing/while_body_invalid.exp
brmataptos: Sorry, yes.

third_party/move/move-compiler-v2/tests/checking/typing/vector_mismatched_args.exp
brmataptos: Yep.

PR #9004, opened 2023-07-09
add compiler tests for const = break

brmataptos:
### Description

Add a compiler test case for buggy code const C = break;

### Test Plan

Ran the test, see the error is caught.


third_party/move/move-compiler/transactional-tests/tests/constants/constant_break.move
```diff
@@ -0,0 +1,8 @@
+//# run
+script {
```
wrwg, line 2: No need for a transactional test, as it does not pass type checking.

brmataptos: Ok.  I was just being paranoid.

wrwg (approved):

rahxephon89 (approved):

PR #9008, opened 2023-07-10
[Spec] Ensures for block.move

UIZorrot:
### Description

This is the property-prove spec of block.move

### Test Plan

aptos move prove -f block


aptos-move/framework/aptos-framework/sources/block.spec.move
```diff
@@ -30,6 +26,9 @@ spec aptos_framework::block {
         include staking_config::StakingRewardsConfigRequirement;
 
         aborts_if false;
+
+        // Only a valid proposer or the vm is authorized to produce blocks
+        ensures proposer == @vm_reserved;
```
rahxephon89, line 31: It is not necessary to have this ensures because proposer is an input variable and it is not changed during execution. This property is expressed using the aborts_if condition.

aptos-move/framework/aptos-framework/sources/block.spec.move
```diff
@@ -52,9 +51,12 @@ spec aptos_framework::block {
         requires system_addresses::is_vm(vm);
         requires (proposer == @vm_reserved) ==> (timestamp::spec_now_microseconds() == timestamp);
         requires (proposer != @vm_reserved) ==> (timestamp::spec_now_microseconds() < timestamp);
-        requires event::counter(event_handle) == new_block_event.height;
 
         aborts_if false;
+
+        // While emitting a new block event, the number of them is equal to the current block height.
+        aborts_if event::counter(event_handle) != new_block_event.height;
```
rahxephon89, line 58: @junkil-park Should we use aborts_if or requires here?

aptos-move/framework/aptos-framework/sources/block.spec.move
```diff
@@ -71,6 +73,14 @@ spec aptos_framework::block {
         let addr = signer::address_of(aptos_framework);
         let account = global<account::Account>(addr);
         aborts_if account.guid_creation_num + 2 >= account::MAX_GUID_CREATION_NUM;
+
+        // During the module's initialization, it guarantees that the BlockResource resource move under the Aptos framework account with initial values.
```
rahxephon89, line 77: Please double check whether the spec here is overlapped with `spec schema Initialize`.

aptos-move/framework/aptos-framework/sources/block.spec.move
```diff
@@ -83,7 +93,6 @@ spec aptos_framework::block {
         aborts_if epoch_interval_microsecs <= 0;
         aborts_if exists<BlockResource>(addr);
         ensures exists<BlockResource>(addr);
-        ensures global<BlockResource>(addr).height == 0;
```
rahxephon89, line 86: Please do not remove it. Use this instead

aptos-move/framework/aptos-framework/sources/block.spec.move
```diff
@@ -118,6 +127,12 @@ spec aptos_framework::block {
         aborts_if !exists<BlockResource>(addr);
         let post block_resource = global<BlockResource>(addr);
         ensures block_resource.epoch_interval == new_epoch_interval;
+
+        // Only the Aptos framework address can execute
+        ensures addr == @aptos_framework;
+
+        // epoch_interval > 0
+        ensures block_resource.epoch_interval > 0;
```
rahxephon89, line 135: I don' think it is necessary because a data invariant `invariant epoch_interval > 0;` is already defined above.

aptos-move/framework/aptos-framework/sources/block.spec.move
UIZorrot: There is one property overlapped

aptos-move/framework/aptos-framework/sources/block.spec.move
```diff
@@ -5,10 +5,6 @@ spec aptos_framework::block {
         invariant [suspendable] chain_status::is_operating() ==> exists<BlockResource>(@aptos_framework);
     }
 
-    spec BlockResource {
```
rahxephon89, line 8: Why do you remove this invariant?

aptos-move/framework/aptos-framework/sources/block.spec.move
```diff
@@ -83,7 +85,10 @@ spec aptos_framework::block {
         aborts_if epoch_interval_microsecs <= 0;
         aborts_if exists<BlockResource>(addr);
         ensures exists<BlockResource>(addr);
+
+        // During the module's initialization, it guarantees that the BlockResource resource move under the Aptos framework account with initial values.
         ensures global<BlockResource>(addr).height == 0;
+        ensures global<BlockResource>(addr).epoch_interval > 0;
```
rahxephon89, line 91: same here, why don't you just use the data invariant above? Is there any issue with it?

aptos-move/framework/aptos-framework/sources/block.spec.move
```diff
@@ -118,6 +123,9 @@ spec aptos_framework::block {
         aborts_if !exists<BlockResource>(addr);
         let post block_resource = global<BlockResource>(addr);
         ensures block_resource.epoch_interval == new_epoch_interval;
+
+        // Only the Aptos framework address can execute
+        ensures addr == @aptos_framework;
```
rahxephon89, line 128: As we discussed last week, this is not necessary.

rahxephon89 (changes requested):

aptos-move/framework/aptos-framework/sources/block.spec.move
junkil-park: This line is not necessary because it's already required.

aptos-move/framework/aptos-framework/sources/block.spec.move
UIZorrot: it has been remove

aptos-move/framework/aptos-framework/sources/block.spec.move
UIZorrot: have removed

aptos-move/framework/aptos-framework/sources/block.spec.move
rahxephon89: Could you remove this?

aptos-move/framework/aptos-framework/sources/block.spec.move
rahxephon89: Could you clarify this?

aptos-move/framework/aptos-framework/sources/block.spec.move
rahxephon89: Could you remove this before resolving this?

rahxephon89 (changes requested):

PR #9009, opened 2023-07-10
Fungible token activities

bowenyang007:
### Description
Add fungible token activities when fungible token is deposit / withdrawn

### Test Plan
image


larry-aptos (approved):
Approved now; going to do review as follow-up

just-in-chang (approved):

PR #9011, opened 2023-07-10
Update Docker images

gedigi:
Update Docker images


geekflyer (approved):
can we add some context/reasoning on the upgrade?

gedigi:
> can we add some context/reasoning on the upgrade?

this PR generated by a github workflow that checks the latest versions of our images (base + builder) on DockerHub

PR #9014, opened 2023-07-10
[Storage Service] Several performance and scalability improvements.

JoshLind:
### Description
This PR offers a number of performance, scalability and observability improvements to the optimistic fetch handling logic in the storage service.

Each improvement is offered in its own commit:
1. Migrate the cached storage server summary from `Arc<RwLock<T>>` to `Arc<ArcSwap<T>>`. [ArcSwap](https://docs.rs/arc-swap/latest/arc_swap/) is more efficient for read-mostly scenarios, which matches that of the cached storage server summary (i.e., lots of concurrent reads, infrequent writes).
2. Migrate the active optimistic fetch map from `Arc<Mutex<HashMap<K,V>>>` to `Arc<DashMap<K, V>>,`. This reduces lock contention when handling concurrent optimistic fetch inserts and deletions. See [DashMap](https://docs.rs/dashmap/latest/dashmap/struct.DashMap.html).
3. Modify `get_peers_with_ready_optimistic_fetches` to identify ready, expired and invalid optimistic fetches in a single set pass (as opposed to multiple passes).
4. Parallelize `handle_ready_optimistic_fetches` to notify each peer of new data in a separate thread. This helps to reduce latencies when there are a lot of ready requests, as each request requires first reading from storage before sending the data to the peer. Now, we do this in parallel (as opposed to sequentially).
5. Parallelize `identify_ready_and_invalid_optimistic_fetches` to identify the ready and invalid optimistic fetches in separate threads. Similar to above, this helps to reduce latencies when there are a lot of pending requests, as each request requires reading from storage.
6. Add a new metric counter to track the number of active optimistic fetches. This is useful for monitoring optimistic fetch load across network IDs.
7. Reduce the task polling intervals in the state sync driver and data streaming service from `100ms` to `50ms`. This helps to reduce the time a received data response is pending in memory, without adding much additional overhead.

### Test Plan
New and existing test infrastructure (i.e., all modifications required updating the existing set of unit tests to ensure that the new functionality is correct).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3145,6 +3145,7 @@ dependencies = [
  "aptos-storage-service-types",
  "aptos-time-service",
  "aptos-types",
+ "arc-swap",
  "bcs 0.1.4",
  "bytes",
  "claims",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3149,6 +3149,7 @@ dependencies = [
  "bcs 0.1.4",
  "bytes",
  "claims",
+ "dashmap",
  "futures",
  "lru 0.7.8",
  "maplit",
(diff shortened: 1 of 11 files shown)


brianolson (approved):

bchocho (approved):
LGTM

PR #9015, opened 2023-07-10
[dag] reliable broadcast storage

ibalajiarun:
### Description

This PR adds storage support for reliable broadcast. The sender persists the node before broadcasting the node, and the receiver persists the signature before returning it. The GC logic is separated out: the idea is there will a periodic task that will call RB handler `gc_before_round` based on information from `dag` store.

### Test Plan

New tests.

### DIFF
--- a/consensus/src/dag/reliable_broadcast.rs
+++ b/consensus/src/dag/reliable_broadcast.rs
@@ -1,7 +1,11 @@
 // Copyright © Aptos Foundation
 // SPDX-License-Identifier: Apache-2.0
 
-use super::types::{CertifiedAck, CertifiedNode};
+use super::{
+    storage::DAGStorage,
+    types::{CertifiedAck, CertifiedNode},
+    NodeId,
+};
 use crate::{
     dag::{
         dag_network::{DAGNetworkSender, RpcHandler},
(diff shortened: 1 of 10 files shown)


consensus/src/dag/reliable_broadcast.rs
```diff
@@ -97,24 +101,50 @@ pub struct NodeBroadcastHandler {
...
+        let mut signatures_by_round_peer = BTreeMap::new();
+        storage
+            .get_node_signatures()
+            .unwrap_or_default()
+            .into_iter()
+            .for_each(|(node_id, digest_sig)| {
+                signatures_by_round_peer
+                    .entry(node_id.round())
+                    .or_insert_with(BTreeMap::new)
+                    .insert(node_id.author(), digest_sig);
```
sasha8, line 123: Clean here old epochs

consensus/src/dag/types.rs
```diff
@@ -275,6 +275,35 @@ impl TDAGMessage for Node {
     }
 }
 
+#[derive(Serialize, Deserialize, PartialEq, Debug, Eq, Hash, Clone)]
+pub struct NodeId {
+    round: Round,
+    author: Author,
```
sasha8, line 281: Add epoch for cleaning purposes. 
Nit: does it make sense to have NodeMetadata be:
{
NodeId
Payload
digrest
}
?

zekun000 (commented):
there's two things I feel a bit weird
- we only store the signature but not the node, it's fine that we broadcast certified node but seems wasteful
- we don't have any bound on what round/how many rounds we accepted, for example we need to guarantee not signing again after min_round

ibalajiarun:
> there's two things I feel a bit weird
> 
> * we only store the signature but not the node, it's fine that we broadcast certified node but seems wasteful
> * we don't have any bound on what round/how many rounds we accepted, for example we need to guarantee not signing again after min_round

for 1, we could store the node during initial broadcast, but we still need to fetch during certified node broadcast. Sasha's prototype was sending twice too and it was performant, so we can revisit this later if it becomes a problem?
for 2, we look at the DAG store to make sure the parent round is at least equal to min_round of dag store. Also, the idea is to gc based on the DAG store min_round, so we wont sign again.

sasha8 (approved):

zekun000 (approved):

PR #9016, opened 2023-07-10
[marketplace-example] Add listing types to events

gregnazario:
### Description
Add listing type to all listing events


aptos-move/move-examples/marketplace/sources/coin_listing.move
```diff
@@ -39,6 +39,8 @@ module coin_listing {
     const ENOT_SELLER: u64 = 6;
 
     // Core data structures
+    const FIXED_PRICE_TYPE: u8 = 1;
+    const AUCTION_TYPE: u8 = 2;
```
bowenyang007, line 43: tbh this is better as a string. Having an int enum is confused, it's also prone to errors.

gregnazario: Sure, this does have a non-negligible overhead though.

rtso (approved):

bowenyang007 (approved):

PR #9018, opened 2023-07-10
[TS SDK] Higher throughput reads

0xmaayan:
### Description

This PR adds support for http2 to achieve higher throughput reads. In a node environment, it depends on what http module the client uses whereas in a browser env it is up to the browser and the BE server to negotiate http2 connection.

We use [got](https://github.com/sindresorhus/got) client  in a node environment as a client that supports http2. `got` doesn't support browser env, and can't be built as part of the main SDK package. Therefore, we use a `aptos-client` package, that exports files based on the environment. For browser downstream packages, the `index.browser.ts` file will be used and that includes the `axios` client, and for node downstream packages, the `index.node.ts` file will be used and that includes the `got` client.

_note: Once approved, we will publish the `aptos-client` package and add it as a dependency in the aptos SDK package, therefore this PR will have minor changes_

### Test Plan
- [x] tests are passing
- [x] Cookies on web
 first request
Screenshot 2023-07-10 at 12 28 22 PM
seconds request
Screenshot 2023-07-10 at 12 28 53 PM

- [x] Cookies on node

first request
Screenshot 2023-07-10 at 12 31 13 PM

second request
Screenshot 2023-07-10 at 12 31 38 PM

requests are coming from `got`
Pasted Graphic

- [x] React app
  - [x] Fullnode
  - [x] Indexer
- [x] Nextjs app
  - [x] Fullnode
  - [x] Indexer
- [x] Node app
  - [x] Fullnode
  - [x] Indexer
- [x] Explorer is working as expected


alokeutpal (approved):

ecosystem/typescript/sdk/packages/aptosClient/src/index.node.ts
```diff
@@ -0,0 +1,107 @@
...
+    delete response.request.options.headers["content-type"];
+  }
+}
+
+/**
+ * got supports only - string | number | boolean | null | undefined
+ * in the searchParam props,
+ * so if we have bigint type, convert it to Number
+ */
+function convertBigIntToNumber(obj: any): any {
```
gregnazario, line 92: What happens if we need a value > the largest number?

ecosystem/typescript/sdk/packages/aptosClient/src/index.node.ts
```diff
@@ -0,0 +1,107 @@
+import got, { OptionsOfJSONResponseBody, RequestError, Response } from "got";
+import { CookieJar } from "./cookieJar";
+import { AptosClientRequest, AptosClientResponse } from "./types";
+
+const cookieJar = new CookieJar();
+
+export default async function aptosClient<Res>(requestOptions: AptosClientRequest): Promise<AptosClientResponse<Res>> {
```
gregnazario, line 7: maybe just a comment here that explains what `aptosClient` does, since there are a lot of client names in the SDK

gregnazario (approved):
Seems simple enough, basically just replace Axios

ecosystem/typescript/sdk/packages/aptosClient/src/index.node.ts
0xmaayan: yea.. so largest Number == 9007199254740991, think it is big enough...?

gregnazario: Is it used for the purposes of a counter or something just for management of the connection, or for information that's transferred on the wire associated with smart contract data?

ecosystem/typescript/sdk/packages/aptosClient/src/index.node.ts
0xmaayan: haha sure

ecosystem/typescript/sdk/packages/aptosClient/src/index.node.ts
0xmaayan: it is used in as a `query param` - so querying something like [ledger version](https://github.com/aptos-labs/aptos-core/blob/main/ecosystem/typescript/sdk/src/providers/aptos_client.ts#L140) or [event type creation number](https://github.com/aptos-labs/aptos-core/blob/main/ecosystem/typescript/sdk/src/providers/aptos_client.ts#L335) or in [pagination](https://github.com/aptos-labs/aptos-core/blob/main/ecosystem/typescript/sdk/src/providers/aptos_client.ts#L47) or [transaction version](https://github.com/aptos-labs/aptos-core/blob/main/ecosystem/typescript/sdk/src/providers/aptos_client.ts#L548) where `AnyNumber` can be a `number` or `BigInt` type

0xmaayan: or we use `string`
image

gregnazario: Let's convert to string then.

ecosystem/typescript/sdk/packages/aptosClient/src/index.node.ts
```diff
@@ -0,0 +1,106 @@
...
+    responseType: "json",
+    headers,
+    hooks: {
+      beforeRequest: [
+        (options) => {
+          const cookies = cookieJar.getCookies(new URL(options.url!));
+
+          if (cookies?.length > 0 && options.headers) {
+            /* eslint-disable no-param-reassign */
+            options.headers.cookie = cookies.map((cookie: any) => `${cookie.name}=${cookie.value}`).join("; ");
```
0xjinn, line 24: Do we grab the cookie associated with an request url, and then assign it to the current request? Is this to ensure the cookie associated with the url is always included in the headers?

0xjinn (approved):

ecosystem/typescript/sdk/packages/aptosClient/src/index.node.ts
0xmaayan: yes. this is just our own implementation of a cookieJar. it is an existing code in our current sdk, where axios doesnt support cookies in a node env, we have our own cookie implementation. this is the same issue here, where got doesn't support cookies out of the box so we use our implementation. The need for cookie in the request is b/c sdk depends on the cookie based sticky routing that happens in the LB

davidiw (commented):
nice work!

The new aptos-client probably belongs in the typescript folder not in the top-level ecosystem folder.

davidiw:
Would be great to do a performance comparison by querying a local node or a non-rate limited upstream fullnode.

0xmaayan:
> Would be great to do a performance comparison by querying a local node or a non-rate limited upstream fullnode.

I believe local node (localhost) doesnt support http2 and we still don't have http2 set up on a LB https://aptos-org.slack.com/archives/C03EP0XM99D/p1689368546475929?thread_ts=1689214615.104159&cid=C03EP0XM99D

any suggestions on how we can test it?

davidiw:
The local node does, you just have to tell your http client to only use http2 and not http1. For example, in Python

https://github.com/aptos-labs/aptos-core/blob/main/ecosystem/python/sdk/aptos_sdk/async_client.py#L57

http2=True
http1=False

I believe forces http2

but its been a little while, you can check the http response to confirm

0xmaayan:
> The local node does, you just have to tell your http client to only use http2 and not http1. For example, in Python
> 
> https://github.com/aptos-labs/aptos-core/blob/main/ecosystem/python/sdk/aptos_sdk/async_client.py#L57
> 
> http2=True http1=False
> 
> I believe forces http2
> 
> but its been a little while, you can check the http response to confirm

yes, the new got client uses http2, but if the server doesn't support it it uses http1
```
Note:

ALPN negotiation will have place in order to determine if the server actually supports HTTP/2. If it doesn't, HTTP/1.1 will be used.
```

I'll do some tests

davidiw:
Hahaha, I am surprised there's no perf difference though... I wonder if it is due to your network.

0xmaayan:
> Hahaha, I am surprised there's no perf difference though... I wonder if it is due to your network.

looks like with http1 it just dies..? there are no results to compare to lol
my network has download speed of ~23.38 mbps and upload speed of ~8.40 mbps

davidiw (approved):

PR #9021, opened 2023-07-10
[Storage Service] Surface renames and refactors.

JoshLind:
Note: (i) nothing should really change in this PR; and (ii) this PR is built on: https://github.com/aptos-labs/aptos-core/pull/9014.

### Description
This PR offers a small number of surface changes (i.e., variable renames and utility function refactors). Specifically, it offers the following (each in their own commit):
1. Rename `max_optimistic_fetch_period` to `max_optimistic_fetch_period_ms`.
2. Rename `OPTIMISTIC_FETCH_VERSION_DELTA` to `OPTIMISTIC_FETCH_VERSION_LAG` and increase the version lag (to tolerate higher TPS networks and timeouts).
3. Add a simple `from_mock` utility method to the `TimeService`.
4. Move several utility functions into the `utils.rs` files. This is so they can be re-used in the future.
5. Alphabetically order the test `utils.rs` file. This should help make it easier to navigate now that the file has grown.

### Test Plan
Existing test infrastructure.

### DIFF
--- a/config/src/config/state_sync_config.rs
+++ b/config/src/config/state_sync_config.rs
@@ -155,7 +155,7 @@ pub struct StorageServiceConfig {
     /// Maximum number of bytes to send per network message
     pub max_network_chunk_bytes: u64,
     /// Maximum period (ms) of pending optimistic fetch requests
-    pub max_optimistic_fetch_period: u64,
+    pub max_optimistic_fetch_period_ms: u64,
     /// Maximum number of state keys and values per chunk
     pub max_state_chunk_size: u64,
     /// Maximum number of transactions per chunk
@@ -179,7 +179,7 @@ impl Default for StorageServiceConfig {
             max_lru_cache_size: 500, // At ~0.6MiB per chunk, this should take no more than 0.5GiB
             max_network_channel_size: 4000,
             max_network_chunk_bytes: MAX_MESSAGE_SIZE as u64,
-            max_optimistic_fetch_period: 5000, // 5 seconds
+            max_optimistic_fetch_period_ms: 5000, // 5 seconds
             max_state_chunk_size: MAX_STATE_CHUNK_SIZE,
             max_transaction_chunk_size: MAX_TRANSACTION_CHUNK_SIZE,
             max_transaction_output_chunk_size: MAX_TRANSACTION_OUTPUT_CHUNK_SIZE,
--- a/state-sync/storage-service/server/src/optimistic_fetch.rs
+++ b/state-sync/storage-service/server/src/optimistic_fetch.rs
@@ -3,13 +3,12 @@
 
 use crate::{
     error::Error,
-    handler::Handler,
     metrics,
     metrics::{increment_counter, OPTIMISTIC_FETCH_EXPIRE},
     moderator::RequestModerator,
     network::ResponseSender,
     storage::StorageReaderInterface,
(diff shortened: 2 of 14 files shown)


state-sync/storage-service/server/src/optimistic_fetch.rs
```diff
@@ -248,7 +251,7 @@ async fn handle_ready_optimistic_fetches<T: StorageReaderInterface>(
                     let optimistic_fetch_request = optimistic_fetch.request.clone();
 
                     // Notify the peer of the new data
-                    if let Err(error) = notify_peer_of_new_data(
+                    if let Err(error) = utils::notify_peer_of_new_data(
```
davidiw, line 254: why move this though if it is only used here?

davidiw (approved):

state-sync/storage-service/server/src/tests/mock.rs
```diff
@@ -213,14 +213,6 @@ fn get_random_network_id() -> NetworkId {
     }
 }
 
-/// Initializes the Aptos logger for tests
-fn initialize_logger() {
```
brianolson, line 217: why move this away? is it going to be referenced somewhere else additionally? Wait for that code change to do this move?

brianolson (commented):
several things move away from where they're used, but don't seem to need to be lifted (e.g. so that other code could also refer to them)

state-sync/storage-service/server/src/optimistic_fetch.rs
JoshLind: Aah, this is used in a future PR, e.g., see here: https://github.com/aptos-labs/aptos-core/pull/9104 😄

state-sync/storage-service/server/src/tests/mock.rs
JoshLind: Aah, see this: https://github.com/aptos-labs/aptos-core/pull/9104. It's the motivation for most of the moves. The PR is already huge, so I didn't want to bloat it further with util changes 😄

config/src/config/state_sync_config.rs
```diff
@@ -155,7 +155,7 @@ pub struct StorageServiceConfig {
     /// Maximum number of bytes to send per network message
     pub max_network_chunk_bytes: u64,
     /// Maximum period (ms) of pending optimistic fetch requests
-    pub max_optimistic_fetch_period: u64,
+    pub max_optimistic_fetch_period_ms: u64,
```
aching, line 158: Does anyone use this config? Will the rename break anything here?

state-sync/storage-service/types/src/responses.rs
```diff
@@ -30,10 +30,10 @@ use std::{
 };
 use thiserror::Error;
 
-/// The version delta we'll tolerate when considering if a peer is eligible
+/// The version lag we'll tolerate when considering if a peer is eligible
 /// to handle an optimistic fetch for new data. This value is set assuming
-/// 5k TPS for a 5 second delay, which should be more than enough.
-pub const OPTIMISTIC_FETCH_VERSION_DELTA: u64 = 25000;
+/// 5k TPS for a 10 second delay, which should be more than enough.
```
bchocho, line 35: For now ;)

bchocho (approved):
LGTM

PR #9023, opened 2023-07-10
[CLI][Ledger] Add hardware wallet options

0xjinn:
### Description

NOTE: This will be merging into ledger feature branch instead of main

This is similar to the private key option we had. Basically create a struct to better manage the `derivation path` or `derivation index` option. There might be case that user do not have profile initialized, but wanted to go-ahead and use their ledger for submitting the transaction

Next Step:
Add this option for txn signing

### Test Plan
```
<!-- Please provide us with clear details for verifying that your changes work. -->
➜  aptos-core git:(feature_branch/cli_ledger_integration) ✗ aptos-debug init --derivation-index 15 --profile deri15
Configuring for profile deri15

Account 4b19d281ce22da0f39ac3d75d8e1aeef27c430d8830dc659d106990eb70cbd6f doesn't exist, creating it and funding it with 100000000 Octas
Account 4b19d281ce22da0f39ac3d75d8e1aeef27c430d8830dc659d106990eb70cbd6f funded successfully

---
Aptos CLI is now set up for account 4b19d281ce22da0f39ac3d75d8e1aeef27c430d8830dc659d106990eb70cbd6f as profile deri15!  Run `aptos --help` for more information about commands
{
  "Result": "Success"
}
```

Result on the config
```
  deri15:
    public_key: "0x7e0e012edd9d932b2079973457abd44bdc485d93af1843f75bef89c2c28f4dc5"
    account: 4b19d281ce22da0f39ac3d75d8e1aeef27c430d8830dc659d106990eb70cbd6f
    rest_url: "https://fullnode.devnet.aptoslabs.com"
    faucet_url: "https://faucet.devnet.aptoslabs.com"
    derivation_path: "m/44'/637'/15'/0'/0'"
```

### DIFF
--- a/crates/aptos/src/common/init.rs
+++ b/crates/aptos/src/common/init.rs
@@ -1,6 +1,7 @@
 // Copyright © Aptos Foundation
 // SPDX-License-Identifier: Apache-2.0
 
+use super::types::HardwareWalletOptions;
 use crate::{
     account::key_rotation::lookup_address,
     common::{
@@ -56,6 +57,9 @@ pub struct InitTool {
     /// and opened. Otherwise CLI would not be able to open the connection to your account
     #[clap(long)]
     pub ledger: bool,
+
+    #[clap(flatten)]
+    pub(crate) hardware_wallet_options: HardwareWalletOptions,
 
     #[clap(flatten)]
     pub rng_args: RngArgs,
@@ -138,8 +142,13 @@ impl CliCommand<()> for InitTool {
             Network::Custom => self.custom_network(&mut profile_config)?,
         }
 
-        // Fetch the top 5 (index 0-4) accounts from Ledger
-        let derivation_path = if self.ledger {
+        // Check if any ledger flag is set
+        let derivation_path = if let Some(deri_path) =
+            self.hardware_wallet_options.extract_derivation_path()?
+        {
+            Some(deri_path)
+        } else if self.ledger {
+            // Fetch the top 5 (index 0-4) accounts from Ledger
             let account_map = aptos_ledger::fetch_batch_accounts(Some(0..5))?;
             eprintln!(
                 "Please choose an index from the following {} ledger accounts, or choose an arbitrary index that you want to use:",
@@ -173,7 +182,7 @@ impl CliCommand<()> for InitTool {
         profile_config.derivation_path = derivation_path.clone();
 
         // Private key
-        let private_key = if self.ledger {
+        let private_key = if self.ledger || self.hardware_wallet_options.is_hardware_wallet() {
             // Private key stays in ledger
             None
         } else {
@@ -208,7 +217,7 @@ impl CliCommand<()> for InitTool {
         };
 
         // Public key
-        let public_key = if self.ledger {
+        let public_key = if self.ledger || self.hardware_wallet_options.is_hardware_wallet() {
             let pub_key = match aptos_ledger::get_public_key(
                 derivation_path
                     .ok_or(CliError::UnexpectedError(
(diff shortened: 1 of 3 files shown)


crates/aptos/src/common/types.rs
```diff
@@ -738,6 +738,40 @@ pub trait ParsePrivateKey {
...
+    /// Make sure your wallet is unblocked and have Aptos opened
+    #[clap(long)]
+    pub derivation_path: Option<String>,
+
+    /// Index of your account in hardware wallet
+    ///
+    /// This is the simpler version of derivation path e.g format - [0]
+    /// we will translate this index into [m/44'/637'/0'/0'/0]
+    #[clap(long)]
+    pub derivation_index: Option<String>,
```
0xjinn, line 755: TBH, I feel like these names are too long. Open for better naming suggestion

crates/aptos/src/common/init.rs
```diff
@@ -208,7 +217,7 @@ impl CliCommand<()> for InitTool {
         };
 
         // Public key
-        let public_key = if self.ledger {
+        let public_key = if self.ledger || self.hardware_wallet_options.is_hardware_wallet() {
```
0xmaayan, line 220: nit: global function for `if self.ledger || self.hardware_wallet_options.is_hardware_wallet()` ?

crates/aptos/src/common/types.rs
```diff
@@ -738,6 +738,40 @@ pub trait ParsePrivateKey {
     }
 }
 
+#[derive(Debug, Default, Parser)]
+pub struct HardwareWalletOptions {
```
0xmaayan, line 742: is this type for any hardware wallet or only ledger? if it is the latest, should probably name it `LedgerWalletOption`

0xmaayan (approved):
some nits, but LGTM

crates/aptos/src/common/init.rs
```diff
@@ -1,6 +1,7 @@
 // Copyright © Aptos Foundation
 // SPDX-License-Identifier: Apache-2.0
 
+use super::types::HardwareWalletOptions;
```
gregnazario, line 4: Probably should use `crate::common::types` it'll combine below

crates/aptos/src/common/init.rs
gregnazario: Yeah, you can put this function on `self`. `fn is_hardware_wallet(&self) -> bool`

crates/aptos/src/common/types.rs
gregnazario: These shouldn't be used particularly often, so I say it's fine.  I think the real question is whether these apply to hardware wallets directly that are not Ledger.  Would be cool to see eventually something like Keystone or otherwise.

gregnazario (approved):

crates/aptos/src/common/types.rs
0xjinn: Keystone?

I did some research, looks like most of the hardware wallet does uses derivation path. But yeah, lets keep these for now for general hardware wallet. And we can made adjustment when we support more than 1 HD wallet

gregnazario: They pretty much all use derivation paths and secret phrases so I think we should be good.

Don't worry about keystone for now :)

crates/aptos/src/common/types.rs
0xjinn: Yeah, I am actually planning it for Hardware overall, I believe all HD wallet should have derivation path

PR #9024, opened 2023-07-10
vertical pod autoscaling for pfn

abbyxion:
### Description
add vertical pod autoscaling for public fullnode, reference:
[reference1](https://cloud.google.com/kubernetes-engine/docs/concepts/verticalpodautoscaler)
[reference2](https://cloud.google.com/kubernetes-engine/docs/how-to/vertical-pod-autoscaling#gcloud_1)
### Test Plan
on local workplace, 
1. manually set the resource request to be extremely low:
```
resources:
  requests:
    cpu: 0.01
    memory: 0.01Gi
  
```
2. set the pod number to be 3, in local workplace, run following commands:
```
terraform apply
k describe pod pfn1-aptos-fullnode-0
```

it shows requests as follows, the same as what we set initially
```
Requests:
  cpu: 10m
  memory: 10737418240m
```

3. check the actual usage of this pod by running `k top pod pfn1-aptos-fullnode-0`, shows:

```
NAME                    CPU(cores)   MEMORY(bytes)   
pfn1-aptos-fullnode-0   234m         2582Mi  

```
it exceeds the resource we initially request, meaning that vpa is taking effect

the vpa detail:
```
NAME                  MODE   CPU    MEM          PROVIDED   AGE
pfn1-aptos-fullnode   Auto   775m   3260022784   True       38m
```

4. from the events of pod:
```
Warning  MultiplePodDisruptionBudgets  30m (x2 over 30m)  controllermanager        Pod "aptos"/"pfn1-aptos-fullnode-0" matches multiple PodDisruptionBudgets.  Chose "pfn2-aptos-fullnode-fullnode" arbitrarily.
```
so we know that MultiplePodDisruptionBudgets is also taking effect

### DIFF
--- a/terraform/fullnode/gcp/kubernetes.tf
+++ b/terraform/fullnode/gcp/kubernetes.tf
@@ -79,7 +79,8 @@ resource "helm_release" "fullnode" {
 
   values = [
     jsonencode({
-      imageTag     = var.image_tag
+      imageTag = var.image_tag
+      deploy = 1
       manageImages = var.manage_via_tf # if we're managing the entire deployment via terraform, override the images as well
       chain = {
         era  = var.era
@@ -124,6 +125,7 @@ resource "helm_release" "fullnode" {
           "iam.gke.io/gcp-service-account" = google_service_account.backup.email
         }
       }
+      verticalPodAutoscalingEnabled = var.enable_vertical_pod_autoscaling
     }),
     jsonencode(var.fullnode_helm_values),
     jsonencode(var.fullnode_helm_values_list == {} ? {} : var.fullnode_helm_values_list[count.index]),
--- a/terraform/fullnode/gcp/kubernetes.tf
+++ b/terraform/fullnode/gcp/kubernetes.tf
@@ -80,7 +80,6 @@ resource "helm_release" "fullnode" {
   values = [
     jsonencode({
       imageTag = var.image_tag
-      deploy = 1
       manageImages = var.manage_via_tf # if we're managing the entire deployment via terraform, override the images as well
       chain = {
         era  = var.era
@@ -126,6 +125,10 @@ resource "helm_release" "fullnode" {
         }
       }
       verticalPodAutoscalingEnabled = var.enable_vertical_pod_autoscaling
+      minAllowedForVpa = {
+        cpu = var.min_allowed_cpu_for_vertical_pod_autoscaling
+        memory = var.min_allowed_memory_for_vertical_pod_autoscaling
+      }
     }),
     jsonencode(var.fullnode_helm_values),
     jsonencode(var.fullnode_helm_values_list == {} ? {} : var.fullnode_helm_values_list[count.index]),
--- a/terraform/fullnode/gcp/kubernetes.tf
+++ b/terraform/fullnode/gcp/kubernetes.tf
@@ -129,6 +129,10 @@ resource "helm_release" "fullnode" {
         cpu = var.min_allowed_cpu_for_vertical_pod_autoscaling
         memory = var.min_allowed_memory_for_vertical_pod_autoscaling
       }
+      maxAllowedForVpa = {
+        cpu = var.max_allowed_cpu_for_vertical_pod_autoscaling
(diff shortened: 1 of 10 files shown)


terraform/fullnode/gcp/kubernetes.tf
```diff
@@ -79,7 +79,8 @@ resource "helm_release" "fullnode" {
 
   values = [
     jsonencode({
-      imageTag     = var.image_tag
+      imageTag = var.image_tag
+      deploy = 1
```
sherry-x, line 83: is this a testing leftover? remove?

terraform/fullnode/gcp/variables.tf
```diff
@@ -199,3 +199,8 @@ variable "manage_via_tf" {
   description = "Whether to manage the aptos-node k8s workload via Terraform. If set to false, the helm_release resource will still be created and updated when values change, but it may not be updated on every apply"
   default     = true
 }
+
+variable "enable_vertical_pod_autoscaling" {
+  description = "vertical pod autoscaling"
```
sherry-x, line 204: a bit more description here on what does "vertical pod autoscaling" means? :D

terraform/helm/fullnode/templates/fullnode.yaml
```diff
@@ -23,6 +23,7 @@ spec:
         seccomp.security.alpha.kubernetes.io/pod: runtime/default
         prometheus.io/scrape: "true"
         prometheus.io/port: "9101"
+        autoscaling.k8s.io/vertical-pod-autoscaler-enabled: "true"
```
sherry-x, line 26: this would hardcode "true" as the option here and bypass the setting with the variable

abbyxion: this is to enable vpa for statefulset, vpa implementation itself is wrapped by if condition of the variable, but yeah it's better to use variable here as well

rustielin (approved):
synced offline about the test plan, LGTM!

terraform/fullnode/gcp/variables.tf
```diff
@@ -199,3 +199,20 @@ variable "manage_via_tf" {
   description = "Whether to manage the aptos-node k8s workload via Terraform. If set to false, the helm_release resource will still be created and updated when values change, but it may not be updated on every apply"
   default     = true
 }
+
+variable "enable_vertical_pod_autoscaling" {
+  description = "vertical pod autoscaling to automatically adjust resource requests"
```
perryjrandall, line 204: nit: please capitalize / add more description to match the above descriptions

perryjrandall (commented):
Have we tested the behavior with a sudden influx of load?

Say for instance we have relatively low traffic, then all of a sudden an NFT mint or the launch of a new app sends 10 or 100x traffic our way how do we respond to such a spike

Its ok not to handle it completely, but we might want to know the performance characteristic and what tradeoff we're making for reliability

Specifically this seems like a good use of MultidimPodAutoscaler since we probably want a combination of response to incoming requests (horizontal) and response to resource utilization (vertical)

abbyxion:
> Have we tested the behavior with a sudden influx of load?
> 
> Say for instance we have relatively low traffic, then all of a sudden an NFT mint or the launch of a new app sends 10 or 100x traffic our way how do we respond to such a spike
> 
> Its ok not to handle it completely, but we might want to know the performance characteristic and what tradeoff we're making for reliability
> 
> Specifically this seems like a good use of MultidimPodAutoscaler since we probably want a combination of response to incoming requests (horizontal) and response to resource utilization (vertical)

vpa is not a recommended way to handle traffic surge from the[ google guidance](https://cloud.google.com/architecture/best-practices-for-running-cost-effective-kubernetes-applications-on-gke#vertical_pod_autoscaler), and from [here](https://aptos-org.slack.com/archives/C0330TM48F4/p1689181522660049?thread_ts=1689113238.031689&cid=C0330TM48F4) we can not do HPA or multi either

abbyxion:
> 

we verify that the vpa will kick in when the resource allocation is not enough for the load

sherry-x (approved):

PR #9030, opened 2023-07-11
[fa/token] allow owner to burn

lightmark:
### Description

RFC: 
1. allow owner to burn owned token (but not object)
2. allow owner to burn owned fa and even purge the whole `fungible store`
### Test Plan
ut


davidiw (changes requested):
would be good to get these ideas into the AIP and then we evaluate all the options...

* the burn method makes the data unrecoverable, which may permanently break an account...
* the move from account makes data recoverable without having to introduce nuanced aspects around how to manage burning of fungible assets.

:thinking:

davidiw:
As we discussed internally, we prefer to have a burn address that we can move objectionable content to...

PR #9032, opened 2023-07-11
[cli] Update homebrew instructions

gregnazario:
### Description
Remove the duplicated homebrew file, and add a reference to where it is in the homebrew repo


crates/aptos/homebrew/README.md
```diff
@@ -43,16 +47,18 @@ aptos --help
 
 Note: This guide is for developers who are trying to update the Aptos homebrew formula.
 
+You can get the latest formula here: https://github.com/Homebrew/homebrew-core/blob/master/Formula/aptos.rb
+
 Copy the `aptos.rb` file to your `homebrew` `formula` directory. For example, on macOS with an M1, this will likely be:
```
0xjinn, line 52: I think this has been updated on m2 mac. Everything now is sitting under `/opt/homebrew/Library/Homebrew`

crates/aptos/homebrew/README.md
```diff
@@ -43,16 +47,18 @@ aptos --help
...
 
 ```bash
 /opt/homebrew/Library/Taps/homebrew/homebrew-core/Formula
```
0xjinn, line 55: I think this might have changed for m2 mac. I dont see any formula folder anymore. Everything seems to be sitting under `/opt/homebrew/Library/Homebrew`

gregnazario: You have to create a local tap afaik.  But, also it varies from system to system and based on configuration.  So, we can skip this for now.

0xjinn: Ahh ok, i guess thats why I have no /Tap folder when I was looking at this yesterday.

0xjinn (approved):

xbtmatt (approved):

PR #9037, opened 2023-07-11
[forge stable fix] reduce success criteria for ha proxy, increase for long running

igor-aptos:


testsuite/forge-cli/src/main.rs
```diff
@@ -237,8 +237,14 @@ fn main() -> Result<()> {
...
         CliCommand::Test(ref test_cmd) => {
+            let ha_proxy = if let TestCommand::K8sSwarm(k8s) = test_cmd {
+                k8s.enable_haproxy
+            } else {
+                false
+            };
+
             // Identify the test suite to run
-            let mut test_suite = get_test_suite(suite_name, duration)?;
+            let mut test_suite = get_test_suite(suite_name, duration, ha_proxy)?;
```
perryjrandall, line 247: adding this feels kinda gross

is there another way to adjust the criteria?

alternatively can we pass the command object itself, that might be generally more useful

alternatively can we override / modify the criteria afterwards?

perryjrandall (approved):

testsuite/forge-cli/src/main.rs
igor-aptos: I don't want to adjust later (i.e like x% for ha proxy), as it can depend on the test.

I can pass test command, though not sure what other fields could be useful, but it might look cleaner :P

perryjrandall: Yeah at the least it couples the haproxy logic to the spot where its being used instead of every point along the way

lgtm

sitalkedia (approved):

PR #9038, opened 2023-07-11
[Python SDK] change library for sync client: httpX -> requests

bchocho:
### Description

Observed two issues in httpX using the sync client, both related to latency:
(1) It does not properly reuse connections on POST, incurring the full TLS handshake time ~100ms.
(2) It seems to delay ~100ms sending of the POST body.

Combined, this was showing ~200ms additional time when measuring latency. These problems don't show up with requests.

Updated with `poetry lock` as well.

### Test Plan

Run `poetry install` `poetry install --sync` `make test` and `make examples`.
Also confirmed manually that latency is reduced when running an e2e measurement script.


ecosystem/python/sdk/pyproject.toml
```diff
@@ -11,9 +11,9 @@ keywords = ["web3", "sdk", "aptos", "blockchain"]
 
 [tool.poetry.dependencies]
 h2 = "^4.1.0"
-httpx = "^0.23.0"
```
bchocho, line 14: Hmm, looks like this is used by async_client. Will add it back. But interesting that the tests still pass?

bchocho: Added it back. Confirmed that things would break, but only after running `poetry install --sync`

0xjinn:
Hey @bchocho do we know how many ms is reduced after switching to `requests` instead of httpx?

And do we need to update `async_client` as well?

bchocho:
@0xjinn for submitting a transaction (and presumably all post calls), the reduction is ~200ms (from my laptop to the aptoslabs devnet).

I haven't had a chance to measure `async_client`. I don't believe `requests` has an async version out of the box.

bchocho:
Actually it's a bit unclear whether it's closer to ~100ms or ~200ms. On my local machine, the current post time is ~120ms. On a cloud runner job the post time is ~220ms. Once this lands and we deploy the version, we can run the e2e job and see the change there.

Note, this isn't observed at localhost. Whatever the bug, it seems to be affected by RTT a lot -- which makes me believe the client is waiting for some kind of ack before sending data.

0xjinn (approved):
LGTM, stamp to unblock - will let @davidiw the python expert to take a final look

nit: Btw, can we update CHANGELOG to include this changes? :D

bchocho:
Added to changelog. Would love to push out 0.6.3 with this change soon!

davidiw (approved):
last release of the sync client ... after this we finally remove it in the 0.7.0 :sob:

PR #9042, opened 2023-07-11
[dag] node fetch handler at receiver

ibalajiarun:
### Description

This PR adds the receiver-side logic for handling a remote node fetch request. For simplicity, a node will return an error if it is unable to fulfill the whole request. There should be some peer in the system that should be able to satisfy this request. 

### Test Plan

 unit test

### DIFF
--- a/consensus/src/dag/dag_fetcher.rs
+++ b/consensus/src/dag/dag_fetcher.rs
@@ -15,7 +15,7 @@ use aptos_consensus_types::common::Author;
 use aptos_infallible::RwLock;
 use aptos_logger::error;
 use aptos_types::epoch_state::EpochState;
-use std::{sync::Arc, time::Duration};
+use std::{ops::Deref, sync::Arc, time::Duration};
 use thiserror::Error as ThisError;
 use tokio::sync::{
     mpsc::{Receiver, Sender},
@@ -91,20 +91,9 @@ impl DagFetcher {
                     local_request.notify();
                     continue;
                 }
-
-                let (lowest_round, bitmask, missing_nodes) =
-                    if let Some((lowest_round, bitmask, missing_nodes)) = dag_reader.bitmask() {
-                        (lowest_round, bitmask, missing_nodes)
-                    } else {
-                        error!("Incomplete round not found, but fetch request received");
-                        continue;
-                    };
-
                 RemoteFetchRequest::new(
                     local_request.node().metadata().clone(),
-                    lowest_round,
-                    bitmask,
-                    missing_nodes,
+                    dag_reader.bitmask(local_request.node().round()),
                 )
             };
             let network_request = DAGMessage::from(remote_request.clone()).into_network_message();
@@ -117,22 +106,21 @@ impl DagFetcher {
                 .and_then(|response| response.verify(&remote_request, &self.epoch_state.verifier))
             {
                 let ceritified_nodes = response.certified_nodes();
-                if ceritified_nodes.len() != remote_request.missing_count() {
-                    error!(
-                        "expected {} nodes, received {}",
-                        remote_request.missing_count(),
-                        ceritified_nodes.len()
-                    );
-                    continue;
-                }
                 // TODO: support chunk response or fallback to state sync
-                let mut dag_writer = self.dag.write();
-                for node in ceritified_nodes {
-                    if let Err(e) = dag_writer.add_node(node) {
-                        error!("Failed to add node {}", e);
+                {
+                    let mut dag_writer = self.dag.write();
+                    for node in ceritified_nodes {
+                        if let Err(e) = dag_writer.add_node(node) {
+                            error!("Failed to add node {}", e);
+                        }
                     }
                 }
-                local_request.notify();
+
+                if self.dag.read().all_exists(local_request.node().parents()) {
+                    local_request.notify();
+                } else {
+                    // TODO: implement retry logic
+                }
             }
         }
     }
@@ -140,8 +128,10 @@ impl DagFetcher {
 
 #[derive(Debug, ThisError)]
 pub enum FetchHandleError {
-    #[error("not enough nodes to satisfy request")]
-    NotEnoughNodes,
+    #[error("target node is not present")]
+    TargetNotPresent,
+    #[error("causal parents not present")]
+    NotPresent,
 }
 
 pub struct FetchHandler {
@@ -161,16 +151,25 @@ impl RpcHandler for FetchHandler {
     fn process(&mut self, message: Self::Request) -> anyhow::Result<Self::Response> {
         let dag_reader = self.dag.read();
 
-        let nodes = dag_reader.get_missing_nodes(message.start_round(), message.exists_bitmask());
-        // If this peer cannot satisfy the request, return an error.
-        ensure!(
-            nodes.len() == message.missing_count(),
-            FetchHandleError::NotEnoughNodes
-        );
+        let target_node = dag_reader
+            .get_node(message.target())
+            .ok_or(FetchHandleError::TargetNotPresent)?;
+
+        let certified_nodes: Vec<_> = dag_reader
+            .reachable(
+                &target_node,
+                Some(message.exists_bitmask().first_round()),
+                |_| true,
+            )
+            .skip(1) // Skip target node
+            .map(|node_status| node_status.as_node().clone().deref().clone())
+            .collect();
+
+        ensure!(certified_nodes.len() > 0, FetchHandleError::NotPresent);
 
         Ok(FetchResponse::new(
             message.target().epoch(),
-            nodes.iter().map(|n| n.as_ref().clone()).collect(),
+            certified_nodes,
         ))
     }
 }
--- a/consensus/src/dag/dag_fetcher.rs
+++ b/consensus/src/dag/dag_fetcher.rs
@@ -91,8 +91,16 @@ impl DagFetcher {
                     local_request.notify();
                     continue;
                 }
+
+                let target = local_request.node();
                 RemoteFetchRequest::new(
-                    local_request.node().metadata().clone(),
+                    target.metadata().epoch(),
+                    local_request
+                        .node()
+                        .parents()
+                        .iter()
+                        .map(|node| node.metadata().clone())
+                        .collect(),
                     dag_reader.bitmask(local_request.node().round()),
                 )
             };
@@ -128,10 +136,8 @@ impl DagFetcher {
 
 #[derive(Debug, ThisError)]
 pub enum FetchHandleError {
-    #[error("target node is not present")]
-    TargetNotPresent,
-    #[error("causal parents not present")]
-    NotPresent,
+    #[error("parents are missing")]
+    ParentsMissing,
 }
 
 pub struct FetchHandler {
@@ -151,24 +157,25 @@ impl RpcHandler for FetchHandler {
     fn process(&mut self, message: Self::Request) -> anyhow::Result<Self::Response> {
         let dag_reader = self.dag.read();
 
-        let target_node = dag_reader
-            .get_node(message.target())
-            .ok_or(FetchHandleError::TargetNotPresent)?;
+        ensure!(
+            message
+                .parents()
+                .iter()
+                .all(|metadata| dag_reader.get_node(metadata).is_some()),
+            FetchHandleError::ParentsMissing
+        );
 
         let certified_nodes: Vec<_> = dag_reader
-            .reachable(
-                &target_node,
+            .reachable_from_parents(
+                message.parents(),
                 Some(message.exists_bitmask().first_round()),
                 |_| true,
             )
-            .skip(1) // Skip target node
             .map(|node_status| node_status.as_node().clone().deref().clone())
             .collect();
 
-        ensure!(certified_nodes.len() > 0, FetchHandleError::NotPresent);
-
         Ok(FetchResponse::new(
-            message.target().epoch(),
+            message.epoch(),
             certified_nodes,
         ))
     }
--- a/consensus/src/dag/dag_fetcher.rs
+++ b/consensus/src/dag/dag_fetcher.rs
@@ -174,6 +174,8 @@ impl RpcHandler for FetchRequestHandler {
             .map(|node_status| node_status.as_node().clone().deref().clone())
             .collect();
 
+        // TODO: decide if the response is too big and act accordingly.
+
         Ok(FetchResponse::new(
             message.epoch(),
             certified_nodes,
--- a/consensus/src/dag/dag_fetcher.rs
+++ b/consensus/src/dag/dag_fetcher.rs
@@ -89,7 +89,7 @@ impl DagFetcher {
                 let dag_reader = self.dag.read();
 
                 let missing_parents: Vec<NodeMetadata> =
-                    dag_reader.filter_missing(local_request.node().parents_metadata_iter());
+                    dag_reader.filter_missing(local_request.node().parents_metadata_iter()).cloned().collect();
 
                 if missing_parents.is_empty() {
                     local_request.notify();
--- a/consensus/src/dag/dag_fetcher.rs
+++ b/consensus/src/dag/dag_fetcher.rs
@@ -88,8 +88,10 @@ impl DagFetcher {
             let remote_request = {
                 let dag_reader = self.dag.read();
 
-                let missing_parents: Vec<NodeMetadata> =
-                    dag_reader.filter_missing(local_request.node().parents_metadata_iter()).cloned().collect();
+                let missing_parents: Vec<NodeMetadata> = dag_reader
+                    .filter_missing(local_request.node().parents_metadata())
+                    .cloned()
+                    .collect();
 
                 if missing_parents.is_empty() {
                     local_request.notify();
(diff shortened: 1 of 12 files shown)


consensus/src/dag/dag_fetcher.rs
```diff
@@ -88,10 +91,20 @@ impl DagFetcher {
                     local_request.notify();
                     continue;
                 }
+
+                let (lowest_round, bitmask, missing_nodes) =
+                    if let Some((lowest_round, bitmask, missing_nodes)) = dag_reader.bitmask() {
```
sasha8, line 96: You are not passing local_request.node()?

consensus/src/dag/dag_store.rs
```diff
@@ -139,8 +173,39 @@ impl Dag {
...
+
+    pub fn bitmask(&self) -> Option<(Round, Vec<Vec<bool>>, usize)> {
+        let lowest_round = match self.lowest_incomplete_round() {
+            Some(lowest_round) => lowest_round,
+            None => return None,
+        };
+
+        let mut missing_count = 0;
+        let bitmask = self
+            .nodes_by_round
```
sasha8, line 193: I think you are not requesting enough.
What if you need to fetch patents in round r, but you have not received nodes in round r-1 yet? (you will not have (r-1, Vec) in the map).

consensus/src/dag/dag_fetcher.rs
sasha8: See comments below, I think it is a bug.

sasha8 (commented):
Two issues:
1. You do not check that the certificates in the response match the requested bitmask.
2. You create bitmask by the dag status only. It is possible that you do not ask for the parents if the target is in a future round.
(See comments)

consensus/src/dag/dag_fetcher.rs
```diff
@@ -103,17 +116,61 @@ impl DagFetcher {
                 .and_then(FetchResponse::try_from)
                 .and_then(|response| response.verify(&remote_request, &self.epoch_state.verifier))
             {
+                let ceritified_nodes = response.certified_nodes();
+                if ceritified_nodes.len() != remote_request.missing_count() {
```
zekun000, line 120: this should be written in the verify function above

consensus/src/dag/dag_fetcher.rs
```diff
@@ -103,17 +116,61 @@ impl DagFetcher {
...
-                    for node in rounds {
-                        if let Err(e) = dag_writer.add_node(node) {
-                            error!("Failed to add node {}", e);
-                        }
+                for node in ceritified_nodes {
+                    if let Err(e) = dag_writer.add_node(node) {
+                        error!("Failed to add node {}", e);
                     }
                 }
                 local_request.notify();
```
zekun000, line 135: we probably should re-check if all parents exists before notify

consensus/src/dag/dag_fetcher.rs
```diff
@@ -103,17 +116,61 @@ impl DagFetcher {
...
+    type Response = FetchResponse;
+
+    fn process(&mut self, message: Self::Request) -> anyhow::Result<Self::Response> {
+        let dag_reader = self.dag.read();
+
+        let nodes = dag_reader.get_missing_nodes(message.start_round(), message.exists_bitmask());
+        // If this peer cannot satisfy the request, return an error.
+        ensure!(
+            nodes.len() == message.missing_count(),
+            FetchHandleError::NotEnoughNodes
```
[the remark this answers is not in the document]
sasha8: I mean, if you can satisfy part of the request, why not send it?

consensus/src/dag/dag_fetcher.rs
sasha8: Probably need to also pass the bitmask and verify that the return certificates match the bitmask.

zekun000: the bitmask is there, the verify function takes in the request itself

consensus/src/dag/dag_store.rs
```diff
@@ -119,6 +119,40 @@ impl Dag {
         self.nodes_by_digest.get(digest).cloned()
     }
 
+    pub fn get_missing_nodes(
+        &self,
+        start_round: Round,
+        bitmask: &[Vec<bool>],
```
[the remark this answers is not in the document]
ibalajiarun: Linter wanted me to do this.

consensus/src/dag/dag_store.rs
```diff
@@ -253,8 +254,32 @@ impl Dag {
...
+            }
+        }
+        None
+    }
+
+    pub fn bitmask(&self, target_round: Round) -> DagSnapshotBitmask {
+        let lowest_round = match self.lowest_incomplete_round() {
+            Some(round) => round,
+            None => {
+                return DagSnapshotBitmask::new(self.highest_round() + 1, vec![vec![
```
sasha8, line 270: I think we still have a problem. What r is the latest round we locally have. but the target round is r+3?

consensus/src/dag/dag_fetcher.rs
```diff
@@ -103,17 +105,71 @@ impl DagFetcher {
...
+impl RpcHandler for FetchHandler {
+    type Request = RemoteFetchRequest;
+    type Response = FetchResponse;
+
+    fn process(&mut self, message: Self::Request) -> anyhow::Result<Self::Response> {
+        let dag_reader = self.dag.read();
+
+        let target_node = dag_reader
+            .get_node(message.target())
+            .ok_or(FetchHandleError::TargetNotPresent)?;
```
sasha8, line 156: I think this is expected behavior when we fetch the parents of a certified node. It is possible that only a byzantine node will have the target. We can either:
1. Persist the node after we sign (today we only persist the signature).
2. Send parents metadata as part of the fetch request.

sasha8 (commented):
I think we still have a few issues (see comments):
1. I think we still have a bug in the bitmask creation logic.
2. Currently, it is not guaranteed that honest parties will have the target.

consensus/src/dag/dag_store.rs
ibalajiarun: That is fine. The bitmask only needs to record what exists in the local dag, and everything in future and not in bitmask doesn't exist? `DagSnapshotBitmask::has()` takes care of this.

consensus/src/dag/types.rs
```diff
@@ -503,39 +503,78 @@ impl BroadcastStatus for CertificateAckState {
...
+            "target node round should be strictly higher than the last bitmark round"
+        );
+
+        ensure!(
+            self.exists_bitmask
+                .bitmask
+                .iter()
+                .all(|round| round.len() == verifier.len()),
+            "invalid bitmask: each round length is not equal to validator count"
+        );
```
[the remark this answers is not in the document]
ibalajiarun: no. currently, I don't include future rounds if they are all `false`

consensus/src/dag/dag_store.rs
sasha8: Yes, I think you are right.
But then why do you return all false in self.highest_round() + 1? Cannot you just return an empty vector?

consensus/src/dag/types.rs
```diff
@@ -503,39 +503,78 @@ impl BroadcastStatus for CertificateAckState {
 #[derive(Serialize, Deserialize, Clone, Debug)]
 pub struct RemoteFetchRequest {
     target: NodeMetadata,
-    start_round: Round,
-    exists_bitmask: Vec<Vec<bool>>,
+    exists_bitmask: DagSnapshotBitmask,
```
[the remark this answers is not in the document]
zekun000: we can just make the target a Vec<NodeMetadata> and only put the missing parents there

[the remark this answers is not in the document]
ibalajiarun: I think we should send all parents metadata to check the voting power of the parents on the receiver-side, just as a sanity check. Otherwise, any validator can draft RemoteFetchRequests with one parent and make a validator work more, send more data and consume bandwidth.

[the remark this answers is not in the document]
zekun000: I think that's orthogonal problem to solve (waste resource/bandwidth), I can always send a remote fetch request from any valid node to achieve the same effect

consensus/src/dag/types.rs
```diff
@@ -480,20 +480,26 @@ impl BroadcastStatus for CertificateAckState {
 /// if a node exist at [start_round + index][validator_index].
 #[derive(Serialize, Deserialize, Clone, Debug)]
 pub struct RemoteFetchRequest {
-    target: NodeMetadata,
+    epoch: u64,
+    parents: Vec<NodeMetadata>,
```
sasha8, line 484: Change name to targets?

consensus/src/dag/dag_store.rs
```diff
@@ -216,14 +216,15 @@ impl Dag {
 
     pub fn reachable(
         &self,
-        from: &Arc<CertifiedNode>,
+        initial: Vec<HashValue>,
```
sasha8, line 219: "pub fn reachable" remove pub.

consensus/src/dag/dag_store.rs
```diff
@@ -232,6 +233,29 @@ impl Dag {
...
+        &self,
+        parents: &[NodeMetadata],
+        until: Option<Round>,
+        filter: impl Fn(&NodeStatus) -> bool,
+    ) -> impl Iterator<Item = &NodeStatus> {
+        self.reachable(
+            parents.iter().map(|p| *p.digest()).collect(),
+            parents[0].round(),
+            until,
+            filter,
```
sasha8, line 255: Instead of two helper functions, maybe just pass Vec of targets to reachable() and have this code there to create the initial?

consensus/src/dag/dag_fetcher.rs
```diff
@@ -127,49 +141,47 @@ impl DagFetcher {
...
-        let target_node = dag_reader
-            .get_node(message.target())
-            .ok_or(FetchHandleError::TargetNotPresent)?;
+        ensure!(
+            message
+                .parents()
+                .iter()
+                .all(|metadata| dag_reader.get_node(metadata).is_some()),
+            FetchRequestHandleError::ParentsMissing
+        );
```
sasha8, line 172: Add a comment explaining why we must have them in honest behavior.

consensus/src/dag/dag_fetcher.rs
```diff
@@ -91,8 +91,22 @@ impl DagFetcher {
...
+                        .node()
+                        .parents()
+                        .iter()
+                        .filter_map(|node| {
+                            if dag_reader.get_node(node.metadata()).is_some() {
+                                None
+                            } else {
+                                Some(node.metadata().clone())
+                            }
+                        })
```
sasha8, line 108: Can you merge the all_exists() check (from above) here?

sasha8 (approved):

consensus/src/dag/dag_fetcher.rs
zekun000: we probably should just return the missing ones from `all_exists` function?

consensus/src/dag/dag_fetcher.rs
```diff
@@ -103,17 +119,69 @@ impl DagFetcher {
...
+    type Response = FetchResponse;
+
+    fn process(&mut self, message: Self::Request) -> anyhow::Result<Self::Response> {
+        let dag_reader = self.dag.read();
+
+        ensure!(
+            message
+                .parents()
+                .iter()
+                .all(|metadata| dag_reader.get_node(metadata).is_some()),
```
zekun000, line 170: looks like redundant to all_exists? probably can take impl Iterator<&NodeMetadata> and having a function to return this from certified_node.parents too

consensus/src/dag/dag_fetcher.rs
```diff
@@ -103,17 +119,69 @@ impl DagFetcher {
...
+            FetchRequestHandleError::ParentsMissing
+        );
+
+        let certified_nodes: Vec<_> = dag_reader
+            .reachable_from_parents(
+                message.parents(),
+                Some(message.exists_bitmask().first_round()),
+                |_| true,
+            )
+            .map(|node_status| node_status.as_node().clone().deref().clone())
```
zekun000, line 180: this looks ugly lol we're trying to clone a CertifiedNode out? can't we just do as_ref().clone()

consensus/src/dag/dag_store.rs
zekun000: the filter seems very general, I wonder should we just pass which NodeStatus we want to accept here instead of a general Fn -> bool

consensus/src/dag/dag_fetcher.rs
```diff
@@ -103,17 +119,69 @@ impl DagFetcher {
...
+            .collect();
+
+        // TODO: decide if the response is too big and act accordingly.
```
zekun000, line 183: apply has filter based on the bitmask

consensus/src/dag/dag_fetcher.rs
ibalajiarun: all_exists is used in many places and doesn't need to clone nodemetadata everywhere. I can introduce a new fn.

consensus/src/dag/dag_store.rs
```diff
@@ -215,19 +227,22 @@ impl Dag {
 
     pub fn reachable(
         &self,
-        from: &Arc<CertifiedNode>,
+        targets: &[NodeMetadata],
         until: Option<Round>,
+        filter: impl Fn(&NodeStatus) -> bool,
```
zekun000, line 232: instead of passing a pattern match I think we can simply pass a bool indicating whether to only include ordered or not

consensus/src/dag/types.rs
```diff
@@ -218,6 +218,10 @@ impl Node {
         &self.parents
     }
 
+    pub fn parents_metadata_iter(&self) -> impl Iterator<Item = &NodeMetadata> {
```
zekun000, line 221: nit: I'd just call it parents_metadata

consensus/src/dag/types.rs
```diff
@@ -480,40 +484,80 @@ impl BroadcastStatus for CertificateAckState {
...
     }
 
     pub fn verify(
         self,
         _request: &RemoteFetchRequest,
-        _validator_verifier: &ValidatorVerifier,
+        validator_verifier: &ValidatorVerifier,
     ) -> anyhow::Result<Self> {
-        todo!("verification");
+        ensure!(
```
zekun000, line 553: we should also verify it matches the bitmask from the request?

zekun000 (approved):

consensus/src/dag/dag_store.rs
ibalajiarun: but this gives more flexibility? One can also filter based on other node fields as well -- not saying its needed but good to have? I will add a TODO to clean it up if we end up not using it for anything else.

consensus/src/dag/types.rs
```diff
@@ -480,40 +484,80 @@ impl BroadcastStatus for CertificateAckState {
...
+            self.certified_nodes
+                .iter()
+                .all(|node| node.verify(validator_verifier).is_ok()),
```
zekun000, line 556: as a general nit, we should start to expose useful information for debug. for example this doesn't expose what the actual error is, it's better to use `try_fold` so that the first error would be returned or it'll return Ok in the end

consensus/src/dag/dag_fetcher.rs
```diff
@@ -163,21 +166,34 @@ impl RpcHandler for FetchRequestHandler {
     fn process(&mut self, message: Self::Request) -> anyhow::Result<Self::Response> {
         let dag_reader = self.dag.read();
 
+        // `Certified Node`: In the good case, there should exist at least one honest validator that
```
sasha8, line 169: f+1 honest, no?

consensus/src/dag/dag_fetcher.rs
```diff
@@ -163,21 +166,34 @@ impl RpcHandler for FetchRequestHandler {
...
+                self.author_to_index
+                    .get(arc_node.author())
+                    .and_then(|author_idx| {
+                        if !message.exists_bitmask().has(arc_node.round(), *author_idx) {
+                            Some(arc_node.as_ref().clone())
+                        } else {
+                            None
+                        }
+                    })
+            })
```
sasha8, line 196: Can you add the filter_map as an argument to reachable?

sasha8 (approved):

consensus/src/dag/dag_fetcher.rs
ibalajiarun: I think that would mean I need to move self.author_to_index and that means I would need to clone it. 

Or I need to make reachable resolve the author to index and pass it to the filter. 

Not sure whichever is readable.

consensus/src/dag/dag_fetcher.rs
ibalajiarun: I was thinking f is byzantine, f is under a network partition during fetch and another f did not sign. wrong?

sasha8: Then you mean one honest that has the data in very 2f+1 quorum/

PR #9049, opened 2023-07-11
Profiling tool

yunuseozer:
### Description
How it works: /server/mod.rs has the endpoints and calls the corresponding functions from different files:

- profiling.rs: reads the HTML content of index.html as text and returns response in HTML format
- cpu_profiling.rs: starts the CPU profiling using pprof-rs crate, sleeps for the specified amount of time (profiling length), dumps the result into "flamegraph.svg", and profiling stops when it goes out of scope (according to crate documentation)
- cpu_flamegraph.rs: reads "flamegraph.svg" into a string, and returns the string in SVG format.
- memory_profiling.rs: starts memory profiling by setting "prof.active" = True, waits, and sets it = False to stop. Then calls the jeprof.py script, which runs jeprof to generate both txt and svg results from the generated heap files. Can work on doing the python part in this file with Rust.
- memory_svg.rs: reads and returns memory profiling svg
- memory_text.rs: reads and returns memory profiling txt
- thread_dump.rs: generates a thread dump using rstack-self crate and dumps the result to thread_dump.txt. The implementation here is far from ideal as rstack-self requires me to start a dummy process for tracing as apparently applications can't trace themselves. To do that, I have a dummy crate called aptos-iuchild, which does nothing apart from this. However, I think it's possible to make this much cleaner when we create a separate profiling crate. 
- thread_dump_result.rs: reads and returns thread_dump.txt
- thread_dump_cpu-experiment.rs: one of the many versions of the code I've been using to generate many thread dumps and combine them in a file for Off-CPU profiling. This commit currently doesn't have all the thing needed to use that as I'm still thinking about the best implementation.

the returned API responses (like "555" in many of them are irrelevant)


crates/aptos-inspection-service/src/server/memory_profiling.rs
```diff
@@ -0,0 +1,60 @@
...
+            println!("{}", result);
+            if result != 0 {
+                panic!("Failed to deactivate jemalloc profiling");
+            }
+        });
+
+        handle.join().unwrap();
+    }
+    let output = Command::new("python3")
+        .arg("/home/yunusozer/aptos-core/crates/aptos-inspection-service/src/server/memory_profile/jeprof.py")
```
brianolson, line 47: since `jeprof.py` is just two calls to subprocess, and this is one external sub command, may as well just do the two external subcommands here and delete `jeprof.py` ?

crates/aptos-inspection-service/src/server/memory_svg.rs
```diff
@@ -0,0 +1,13 @@
+// Copyright © Aptos Foundation
+
+use crate::server::utils::CONTENT_TYPE_SVG;
+use hyper::{Body, StatusCode};
+use std::fs;
+
+pub fn handle_memory_svg_request() -> (StatusCode, Body, String) {
+    let content =
+        fs::read_to_string("./crates/aptos-inspection-service/src/server/memory_profile/heap.svg")
```
brianolson, line 9: how does that file get there, and why does it go there and not `.` or `/tmp` ?

crates/aptos-inspection-service/src/server/cpu_profiling.rs
```diff
@@ -0,0 +1,24 @@
+// Copyright © Aptos Foundation
+
+use crate::server::utils::CONTENT_TYPE_JSON;
+use hyper::{Body, StatusCode};
+use std::{fs::File, path::Path, thread, time};
+
+pub fn handle_cpu_profiling_request() -> (StatusCode, Body, String) {
+    let guard = pprof::ProfilerGuard::new(100).unwrap();
+    let five_secs = time::Duration::from_millis(5000);
```
brianolson, line 9: does this mean we gather CPU sampling profile for 5 seconds? An option to run longer would be nice.

crates/aptos-inspection-service/src/server/memory_profiling.rs
yunuseozer: Yes, for some reason it was erroring when I tried to do that so I just did it this way to temporarily unblock, but it's on my todo list. Thanks!

crates/aptos-inspection-service/src/server/memory_svg.rs
yunuseozer: You're right, the file locations are to-be-changed as well. Will either store them in `/tmp` or another folder depending on if we want to store previous results long-term.

crates/aptos-inspection-service/src/server/cpu_profiling.rs
yunuseozer: 5 secs were for demo purposes, but will either determine a more ideal duration or make it an input.

crates/aptos-inspection-service/src/server/jeprof.py
```diff
@@ -0,0 +1,16 @@
+import subprocess
```
areshand, line 1: if possible, let's rewrite it in rust and put it in the aptos-profiler crater as well

crates/aptos-iuchild/src/main.rs
```diff
@@ -0,0 +1,18 @@
+// Copyright © Aptos Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+extern crate rstack_self;
```
areshand, line 4: move this to the profiler crate

crates/aptos-inspection-service/src/server/mod.rs
```diff
@@ -132,6 +143,41 @@ async fn serve_requests(
             // Exposes the system and build information
             system_information::handle_system_information_request(node_config)
         },
+        CPU_PROFILING_PATH => {
```
areshand, line 146: Ideally, we should be able to start a separate profiling service independent of inspection service with its own dedicated URL mapping.

grao1991 (commented):
Please break this PR into smaller PRs, otherwise it's hard to review.

crates/aptos-inspection-service/src/server/profiling.rs
```diff
@@ -16,116 +16,58 @@ pub fn handle_profiling_request() -> (StatusCode, Body, String) {
...
-        let file = File::create("./flamegraph.svg").unwrap();
-        report.flamegraph(file).unwrap();
-
-        println!("report: {:?}", &report);
-    };
+    //Call aptos-profiling cpu profiling 
+    let config = ProfilerConfig::new_with_defaults();
+    let handler = ProfilerHandler::new(config);
+    let cpu_profiler = handler.get_cpu_profiler();
+    cpu_profiler.start_profiling();
```
areshand, line 28: if the start_profiling returns error, we should return a different status code? similarly for the rest of functions

crates/aptos-inspection-service/src/server/profiling.rs
```diff
@@ -16,116 +16,58 @@ pub fn handle_profiling_request() -> (StatusCode, Body, String) {
...
+    memory_profiler.start_profiling();
 
-    (StatusCode::OK, Body::from("555"), CONTENT_TYPE_TEXT.into())
+    (StatusCode::OK, Body::from("Success"), CONTENT_TYPE_TEXT.into())
 }
 
 //TODO: use aptos-profiler crater
 pub fn handle_cpu_flamegraph_request() -> (StatusCode, Body, String) {
-    let content = fs::read_to_string("./flamegraph.svg").expect("Failed to read input");
+    let config = ProfilerConfig::new_with_defaults();
```
areshand, line 45: update the comment above

crates/aptos-profiler/src/memory_profiler.rs
```diff
@@ -2,29 +2,92 @@
...
+            println!("Command executed successfully. Output:\n{}", stdout);
+        } else {
+            let stderr = String::from_utf8_lossy(&output.stderr);
+            println!("Command failed. Error:\n{}", stderr);
+        }
+        Ok(())
     }
-    // End profiling
+
     fn end_profiling(&self) -> Result<()> {
```
areshand, line 77: we should be able to stop memory profiling before the duration expires?

crates/aptos-profiler/src/memory_profiler.rs
```diff
@@ -2,29 +2,92 @@
...
+    
+                println!("{}", result);
+                if result != 0 {
+                    panic!("Failed to deactivate jemalloc profiling");
+                }
+            });
+    
+            handle.join().unwrap();
+        }
+        let output = Command::new("python3")
```
areshand, line 62: should we run this command only when we want to expose the results?

crates/aptos-profiler/src/memory_profiler.rs
```diff
@@ -2,29 +2,92 @@
...
     }
 }
 
 impl Profiler for MemProfiler {
     fn start_profiling(&self) -> Result<()> {
-        unimplemented!()
+        unsafe {
+            let mut prof_active: bool = true;
+    
+            let result = jemalloc_sys::mallctl(
```
areshand, line 28: does it work if only change the environment variable to start profiling? or we have to use unsafe here?

crates/aptos-profiler/src/memory_profiler.rs
```diff
@@ -2,29 +2,92 @@
...
+                b"prof.active\0".as_ptr() as *const _,
+                std::ptr::null_mut(),
+                std::ptr::null_mut(),
+                &mut prof_active as *mut _ as *mut _,
+                std::mem::size_of::<bool>(),
+            );
+    
+            println!("{}", result);
+            if result != 0 {
+                panic!("Failed to activate jemalloc profiling");
```
areshand, line 38: return error and don't panic. we don't want profiling errors causing node exit

crates/aptos-profiler/src/memory_profiler.rs
```diff
@@ -2,29 +2,92 @@
...
+    
+            let result = jemalloc_sys::mallctl(
+                b"prof.active\0".as_ptr() as *const _,
+                std::ptr::null_mut(),
+                std::ptr::null_mut(),
+                &mut prof_active as *mut _ as *mut _,
+                std::mem::size_of::<bool>(),
+            );
+    
+            println!("{}", result);
```
areshand, line 36: use info!() for logging?

PR #9053, opened 2023-07-12
[Python][SDK] Make string representation of AccountAddress conform to AIP-40

banool:
### Description
This PR adds the `to_standard_string` function which formats AccountAddress instances as strings in a way that is compliant with AIP 40.

See also https://github.com/aptos-labs/aptos-core/pull/8727.

### Test Plan
```
make test
```
```
make examples
poetry run python -m examples.multisig
```

### DIFF
--- a/ecosystem/python/sdk/aptos_sdk/account_address.py
+++ b/ecosystem/python/sdk/aptos_sdk/account_address.py
@@ -25,6 +25,8 @@ class AccountAddress:
     def __init__(self, address: bytes):
         self.address = address
 
+        # Note: The correctness of is_special relies on this check. Be careful when
+        # changing anything about this function.
         if len(address) != AccountAddress.LENGTH:
             raise Exception("Expected address of length 32")
 
@@ -36,7 +38,48 @@ class AccountAddress:
     def __str__(self):
         return self.hex()
 
+    def to_standard_string(self):
+        """
+        Represent an account address in a way that is compliant with the v1 address
+        standard. The standard is defined as part of AIP-40, read more here:
+        https://github.com/aptos-foundation/AIPs/blob/main/aips/aip-40.md
+
+        In short, all special addresses SHOULD be represented in SHORT form, e.g.
+
+        0x1
+
+        All other addresses MUST be represented in LONG form, e.g.
+
+        0x002098630cfad4734812fa37dc18d9b8d59242feabe49259e26318d468a99584
+
+        For an explanation of what defines a "special" address, see `is_special`.
+
+        All string representations of addresses MUST be prefixed with 0x.
+        """
+        suffix = self.address.hex()
+        if self.is_special():
+            suffix = suffix.lstrip("0") or "0"
+        return f"0x{suffix}"
+
+    def is_special(self):
+        """
+        Returns whether the address is a "special" address. Addresses are considered
+        special if the first 63 characters of the hex string are zero. In other words,
+        an address is special if the first 31 bytes are zero and the last byte is
+        smaller than than `0b10000` (16). In other words, special is defined as an address
+        that matches the following regex: `^0x0{63}[0-9a-f]$`. In short form this means
+        the addresses in the range from `0x0` to `0xf` (inclusive) are special.
+
+        For more details see the v1 address standard defined as part of AIP-40:
+        https://github.com/aptos-foundation/AIPs/blob/main/aips/aip-40.md
+        """
+        return all(b == 0 for b in self.address[:-1]) and self.address[-1] < 0b10000
+
     def hex(self) -> str:
+        """
+        NOTE: Prefer to_standard_string for representing instances of AccountAddress
+        as strings.
+        """
         return f"0x{self.address.hex()}"
 
     @staticmethod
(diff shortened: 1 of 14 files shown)


ecosystem/python/sdk/aptos_sdk/account_address.py
```diff
@@ -167,3 +210,97 @@ def test_token(self):
...
+            "0x1",
+        )
+
+        # Test special address: 0x4
+        self.assertEqual(
+            AccountAddress.from_hex(
+                "0x0000000000000000000000000000000000000000000000000000000000000004"
+            ).to_standard_string(),
+            "0x4",
+        )
```
0xjinn, line 237: nit: These two seems a bit redundant

ecosystem/python/sdk/aptos_sdk/account_address.py
```diff
@@ -167,3 +210,97 @@ def test_token(self):
...
+            "0xd",
+        )
+
+        # Test non-special address from long:
+        # 0x0000000000000000000000000000000000000000000000000000000000000010
+        self.assertEqual(
+            AccountAddress.from_hex(
+                "0x0000000000000000000000000000000000000000000000000000000000000010"
+            ).to_standard_string(),
+            "0x0000000000000000000000000000000000000000000000000000000000000010",
```
0xjinn, line 259: nit: For all of these testing, I would assign the addr string to a var. Kinda hard to eyeball them, especially the one with 0s

0xjinn (approved):
Couple nits, overall looks good!

ecosystem/python/sdk/aptos_sdk/account_address.py
```diff
@@ -167,3 +210,97 @@ def test_token(self):
...
+        # Test special address: 0x0
+        self.assertEqual(
+            AccountAddress.from_hex(
+                "0x0000000000000000000000000000000000000000000000000000000000000000"
+            ).to_standard_string(),
+            "0x0",
+        )
+
+        # Test special address: 0x1
+        self.assertEqual(
```
0xmaayan, line 224: > In other words, an address is special if the first 31 bytes are zero and the last byte is smaller than than `0b10000` (16)

does it mean that `0x0000000000000000000000000000000000000000000000000000000000000015` would become `0x15`? 
and `0x0000000000000000000000000000000000000000000000000000000000000016` would be `0x0000000000000000000000000000000000000000000000000000000000000016`?

xbtmatt: I believe the special addresses are just `0x0` to `0xf`

Anything that's a single hex char

davidiw (changes requested):
Thanks for taking the initiative here. I'd rather just eliminate the legacy code and do a migration to the "standard" string.

ecosystem/python/sdk/aptos_sdk/account_address.py
```diff
@@ -25,6 +25,8 @@ class AccountAddress:
     def __init__(self, address: bytes):
         self.address = address
 
+        # Note: The correctness of is_special relies on this check. Be careful when
+        # changing anything about this function.
```
davidiw, line 29: I don't understand the usefulness of this comment. If it isn't 32-bytes it isn't a valid bytes-based address.

ecosystem/python/sdk/aptos_sdk/account_address.py
```diff
@@ -36,7 +38,48 @@ def __eq__(self, other: object) -> bool:
...
+        that matches the following regex: `^0x0{63}[0-9a-f]$`. In short form this means
+        the addresses in the range from `0x0` to `0xf` (inclusive) are special.
+
+        For more details see the v1 address standard defined as part of AIP-40:
+        https://github.com/aptos-foundation/AIPs/blob/main/aips/aip-40.md
+        """
+        return all(b == 0 for b in self.address[:-1]) and self.address[-1] < 0b10000
+
     def hex(self) -> str:
+        """
```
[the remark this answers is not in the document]
banool: That would be a breaking change, I'm not trying to break anything in this PR.

ecosystem/python/sdk/aptos_sdk/account_address.py
banool: That's right ^

ecosystem/python/sdk/aptos_sdk/account_address.py
banool: We went from 16 to 32 at one point, changes happen, I'd rather be defensive. I don't mind too much though, I can remove it if you want.

ecosystem/python/sdk/aptos_sdk/account_address.py
banool: Fair call!

ecosystem/python/sdk/aptos_sdk/account_address.py
banool: Eh, the more cases the better, it's redundant until it's not, I see 0x1 being somehow different and I wanna test for it specifically.

banool:
Alright works for me, given the Python SDK is still 0.x.x. I'll just make breaking changes.

ecosystem/python/sdk/aptos_sdk/account_address.py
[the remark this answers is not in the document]
banool: Disregard, we'll just break it! We're at 0.x.x after all.

banool:
Just addressed the feedback. I made it that there is only `__str__`, let me know what you think.

ecosystem/python/sdk/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "aptos-sdk"
-version = "0.6.3"
+version = "0.7.0"
```
davidiw, line 3: I haven't published 0.7.0 so no need to bump this just update the change log

ecosystem/python/sdk/aptos_sdk/account_address.py
```diff
@@ -34,10 +34,41 @@ def __eq__(self, other: object) -> bool:
...
+        that matches the following regex: `^0x0{63}[0-9a-f]$`. In short form this means
+        the addresses in the range from `0x0` to `0xf` (inclusive) are special.
+
+        For more details see the v1 address standard defined as part of AIP-40:
+        https://github.com/aptos-foundation/AIPs/blob/main/aips/aip-40.md
+        """
+        return all(b == 0 for b in self.address[:-1]) and self.address[-1] < 0b10000
 
     @staticmethod
     def from_hex(address: str) -> AccountAddress:
```
davidiw, line 74: probably should relabel this `from_str` at some point

davidiw (approved):
thanks!

ecosystem/python/sdk/aptos_sdk/account_address.py
banool: I'll do it now while we're in the business of making breaking changes

PR #9055, opened 2023-07-12
[Governance][Move] Add a specific function in aptos governance to change feature flags

movekevin:


perryjrandall (approved):

aptos-move/framework/aptos-framework/sources/aptos_governance.move
```diff
@@ -542,6 +542,13 @@ module aptos_framework::aptos_governance {
         reconfiguration::reconfigure();
     }
 
+    /// Update feature flags and also trigger reconfiguration.
+    public fun toggle_features(aptos_framework: &signer, enable: vector<u64>, disable: vector<u64>) {
+        system_addresses::assert_aptos_framework(aptos_framework);
+        features::change_feature_flags(aptos_framework, enable, disable);
+        reconfiguration::reconfigure();
```
xindingw, line 549: Why it has to reconfigure?

movekevin: Because of forced syncing. Without reconfiguration, state sync, when processing transactions in chunks spanning across blocks, might have stale data as it doesn't reset the VM until the chunk of txs is fully processed. This means if the feature enables a new feature in the VM and is done in a transaction in the middle of a chunk, the txs are this should be on the new VM behavior but would instead be on the old one when re-executed in state sync. This would lead to a divergent result from actual execution as part of a block.

davidiw: this is just to save a few lines of code in the upgrade script?

davidiw (approved):

aptos-move/framework/aptos-framework/sources/aptos_governance.move
movekevin: Yes that can accidentally be forgotten

Issue #6922, opened 2023-03-04
[Feature Request] Support type annotations for lambda parameters


wrwg:
Currently it is not possible to provide type annotations for lambda parameters. Instead of `|x: T| e` one has to write `|x| { let x:T = x; e }`. Type annotations are required to help the inference algorithm in some occasions.

PR #9056, opened 2023-07-12
[Move examples] Add a fungible-asset-based reward pool for distributing epoch-based rewards in multiple tokens

movekevin:
### Description
/// An example module that manages rewards for multiple tokens on a per-epoch basis. Rewards can be added for multiple
/// tokens by anyone for any epoch but only friended modules can increase/decrease claimer shares.
///
/// This module is designed to be integrated into a complete system that manages epochs and rewards.
///
/// The flow works as below:
/// 1. A rewards pool is created with a set of reward tokens (fungible assets). If coins are to be used as rewards,
/// developers can use the coin_wrapper module from move-examples/swap to convert coins into fungible assets.
/// 2. Anyone can add rewards to the pool for any epoch for multiple tokens by calling add_rewards.
/// 3. Friended modules can increase/decrease claimer shares for current epoch by calling increase_allocation and
/// decrease_allocation.
/// 4. Claimers can claim their rewards in all tokens for any epoch that has ended by calling claim_rewards, which
/// return a vector of all the rewards. Claiming also removes the claimer's shares from that epoch's rewards as their
/// rewards have all been claimed.
///
/// Although claimers have to be signers, this module can be easily modified to support objects (e.g. NFTs) as claimers.

### Test Plan
Unit tests


aptos-move/move-examples/rewards_pool/sources/rewards_pool.move
```diff
@@ -0,0 +1,211 @@
...
+///
+/// The flow works as below:
+/// 1. A rewards pool is created with a set of reward tokens (fungible assets). If coins are to be used as rewards,
+/// developers can use the coin_wrapper module from move-examples/swap to convert coins into fungible assets.
+/// 2. Anyone can add rewards to the pool for any epoch for multiple tokens by calling add_rewards.
+/// 3. Friended modules can increase/decrease claimer shares for current epoch by calling increase_allocation and
+/// decrease_allocation.
+/// 4. Claimers can claim their rewards for any epoch that has ended by calling claim_rewards.
+///
+/// Note that due to integer division, there can be rounding errors that leave 1 unit of reward left for each token in
```
lightmark, line 14: how does this happen, because you don't update the pool?

aptos-move/move-examples/rewards_pool/sources/rewards_pool.move
```diff
@@ -0,0 +1,211 @@
...
+        /// We need to keep the fungible store's extend ref to be able to transfer rewards from it during claiming.
+        store_extend_ref: ExtendRef,
+    }
+
+    #[resource_group_member(group = aptos_framework::object::ObjectGroup)]
+    struct RewardsPool has key {
+        /// Track allocation per address per epoch. This is used to calculate the rewards for each address.
+        epoch_allocations: SmartTable<u64, Pool>,
+        epoch_rewards: SmartTable<Object<Metadata>, EpochRewards>,
+        reward_tokens: vector<Object<Metadata>>,
```
lightmark, line 49: why not smart?

aptos-move/move-examples/rewards_pool/sources/rewards_pool.move
```diff
@@ -0,0 +1,211 @@
...
+        rewards_pool: Object<RewardsPool>,
+        amount: u64,
+    ) acquires RewardsPool {
+        let epoch_allocations = &mut unchecked_mut_rewards_pool_data(&rewards_pool).epoch_allocations;
+        let current_epoch = epoch::now();
+        let pool = smart_table::borrow_mut(epoch_allocations, current_epoch);
+        pool_u64::redeem_shares(pool, claimer, (amount as u128));
+
+        // Delete the epoch allocation pool if there are no remaining claimers.
+        if (pool_u64::total_coins(pool) == 0) {
```
lightmark, line 161: Does `pool_u64::total_coins(pool) == 0` always equal to `pool_u64::total_shares(pool) == 0`? I would suggest to add a function in pool module called `is_empty` and assert there.

aptos-move/move-examples/rewards_pool/sources/rewards_pool.move
```diff
@@ -0,0 +1,211 @@
...
+        rewards_pool: Object<RewardsPool>,
+        reward_token: Object<Metadata>,
+        epoch: u64,
+    ): FungibleAsset acquires RewardsPool {
+        assert!(epoch < epoch::now(), EREWARDS_CANNOT_BE_CLAIMED_FOR_CURRENT_EPOCH);
+        let reward = rewards(claimer, rewards_pool, reward_token, epoch);
+        assert!(reward > 0, ENO_REWARDS_TO_CLAIM);
+        let reward_store = smart_table::borrow(&safe_rewards_pool_data(&rewards_pool).epoch_rewards, reward_token);
+        let store_signer = &object::generate_signer_for_extending(&reward_store.store_extend_ref);
+        fungible_asset::withdraw(store_signer, reward_store.store, reward)
```
lightmark, line 117: It seems you never update the shares/rewards in the EpochRewards.amount for claim_rewards. Also the shares in epoch_allocation should be updated by calling `decrease_allocation` correspondingly. If you update, the pool can be cleaned.

movekevin: Its not correct to decrease allocation after one claim as users can be claiming a single token each time instead of all tokens. With that said though, may be this can just only support claiming all tokens at once 🤔

aptos-move/move-examples/rewards_pool/sources/rewards_pool.move
movekevin: Yes its the same because total coins is never changed. Its always equal to total shares

aptos-move/move-examples/rewards_pool/sources/rewards_pool.move
movekevin: Not iterable. And this is not meant to support thousands of rewards tokens

aptos-move/move-examples/rewards_pool/sources/rewards_pool.move
movekevin: Yes because no redemption per a comment below. Im experimenting with only allowing all tokens to be claimed at once so we can clean up the pool

aptos-move/move-examples/rewards_pool/sources/epoch.move
```diff
@@ -0,0 +1,31 @@
...
+
+    /// The epoch duration is fixed at 1 week (in seconds).
+    const EPOCH_DURATION: u64 = 604800;
+
+    #[view]
+    public fun now(): u64 {
+        to_epoch(timestamp::now_seconds())
+    }
+
+    public inline fun duration(): u64 {
```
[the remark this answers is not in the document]
movekevin: Its a view function. No reason for it to be private if it can be useful for reading data on chain

aptos-move/move-examples/rewards_pool/sources/rewards_pool.move
lightmark: I think this should be matched with 48. smart+smart or simple_map+vector

aptos-move/move-examples/rewards_pool/sources/rewards_pool.move
lightmark: I think now I get it.. a little bit overcomplicated...

aptos-move/move-examples/rewards_pool/sources/rewards_pool.move
lightmark: total coins = total shares? sry I don't get this part.

aptos-move/move-examples/rewards_pool/sources/rewards_pool.move
movekevin: How would you have done it more simply?

aptos-move/move-examples/rewards_pool/sources/rewards_pool.move
movekevin: Why? If the data is not expected to be large, why does this need to use a smart vector?

aptos-move/move-examples/rewards_pool/sources/rewards_pool.move
lightmark: No, I didn't mean it can be done simply if you just have one pool and support withdraw separately. As you said, withdraw all at once... I mean, withdraw separately may be over complicated and not that important to have

aptos-move/move-examples/rewards_pool/sources/rewards_pool.move
lightmark: I mean,..., it's not large then why there is a smart table... vector + smart_table seems a little inconsistent.

aptos-move/move-examples/rewards_pool/sources/rewards_pool.move
movekevin: Updated code to be simpler and more consistent

aptos-move/move-examples/rewards_pool/sources/rewards_pool.move
```diff
@@ -0,0 +1,288 @@
...
+        /// Pool representing the claimer shares in this epoch.
+        claimer_pool: Pool,
+    }
+
+    /// Data regarding the store object for a specific reward token.
+    struct RewardStore has store {
+        /// The fungible store for this reward token.
+        store: Object<FungibleStore>,
+        /// We need to keep the fungible store's extend ref to be able to transfer rewards from it during claiming.
+        store_extend_ref: ExtendRef,
```
lightmark, line 50: finicky disclamer: do you think it is better pattern to warp the extend_ref and put it into the same `Object<FungibleStore>`? The current pattern has 2 hops.

aptos-move/move-examples/rewards_pool/sources/rewards_pool.move
```diff
@@ -0,0 +1,288 @@
...
+            let reward_store = simple_map::borrow(&rewards_data.reward_stores, &reward_token);
+            if (reward == 0) {
+                vector::push_back(
+                    &mut rewards,
+                    fungible_asset::zero(fungible_asset::store_metadata(reward_store.store)),
+                );
+            } else {
+                // Withdraw the reward from the corresponding store.
+                let store_signer = &object::generate_signer_for_extending(&reward_store.store_extend_ref);
+                vector::push_back(&mut rewards, fungible_asset::withdraw(store_signer, reward_store.store, reward));
```
lightmark, line 172: Now I wonder maybe we should allow withdraw 0.

aptos-move/move-examples/rewards_pool/sources/rewards_pool.move
```diff
@@ -0,0 +1,288 @@
...
+            let reward_token = fungible_asset::metadata_from_asset(&fa);
+            assert!(simple_map::contains_key(reward_stores, &reward_token), EREWARD_TOKEN_NOT_SUPPORTED);
+
+            // Deposit the rewards into the corresponding store.
+            let reward_store = simple_map::borrow(reward_stores, &reward_token);
+            fungible_asset::deposit(reward_store.store, fa);
+
+            // Update total amount of rewards for this token for this epoch.
+            let total_amounts = &mut epoch_rewards_or_default(&mut rewards_data.epoch_rewards, epoch).total_amounts;
+            if (simple_map::contains_key(total_amounts, &reward_token)) {
```
lightmark, line 208: missing borrow_mut_with_default in simple_map

lightmark (approved):

aptos-move/move-examples/rewards_pool/sources/rewards_pool.move
movekevin: May be we can introduce a withdraw_safe function. I still think disallowing withdrawing 0 is cleaner. There's usually much confusion when a function call is successful but doesn't do/return anything

aptos-move/move-examples/rewards_pool/sources/rewards_pool.move
movekevin: Yep. It's slightly tricky because the syntax in simple_map is very strange to avoid requiring droppable values (relatively to smart_table). May be @xindingw can look into this as he's working on improving simple_map separately anyway :)

aptos-move/move-examples/rewards_pool/sources/rewards_pool.move
```diff
@@ -0,0 +1,288 @@
...
+    public fun create(reward_tokens: vector<Object<Metadata>>): Object<RewardsPool> {
+        // The owner of the object doesn't matter as there are no owner-based permissions.
+        // If developers want to be extra cautious here, they can make the owner @0x0.
+        // Here the reward pool also doesn't keep an ExtendRef so there would be no way to obtain its signer.
+        let rewards_pool_constructor_ref = &object::create_object(@rewards_pool);
+        let rewards_pool_signer = &object::generate_signer(rewards_pool_constructor_ref);
+        let rewards_pool_addr = signer::address_of(rewards_pool_signer);
+        let reward_stores = simple_map::new();
+        vector::for_each(reward_tokens, |reward_token| {
+            let reward_token: Object<Metadata> = reward_token;
```
davidiw, line 74: is this a deficiency in inline functions?

davidiw (approved):

aptos-move/move-examples/rewards_pool/sources/rewards_pool.move
movekevin: Yep. It doesn't infer types correctly sometimes. There's a bug tracking this https://github.com/aptos-labs/aptos-core/issues/6922 but @wrwg mentioned this would not be fixed in compiler v1

PR #9063, opened 2023-07-12
[Draft] Adding `deferred_read` function for aggregators

vusirikala:
### Description

This PR adds `deferred_read` function for aggregators. The `deferred_read` function outputs an AggregatorSnapshot.

### DIFF
--- a/aptos-move/framework/aptos-framework/sources/aggregator/aggregator.move
+++ b/aptos-move/framework/aptos-framework/sources/aggregator/aggregator.move
@@ -29,6 +29,15 @@ module aptos_framework::aggregator {
         limit: u128,
     }
 
+    /// Represents snapshot of an aggregator at a particular point in time.
+    /// In storage, the value represents the materialized value of the aggregator
+    /// when the snapshot was created.
+    /// During the block execution, the value represents an identifier pointing to the
+    /// Rust `AggregatorSnapshot` which cointains additional details related.
+    struct AggregatorSnapshot has store {
+        value: u128
+    }
+
     /// Returns `limit` exceeding which aggregator overflows.
     public fun limit(aggregator: &Aggregator): u128 {
         aggregator.limit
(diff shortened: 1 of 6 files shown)


aptos-move/framework/aptos-framework/sources/aggregator/aggregator.move
```diff
@@ -29,6 +29,15 @@ module aptos_framework::aggregator {
         limit: u128,
     }
 
+    /// Represents snapshot of an aggregator at a particular point in time.
```
georgemitenkov, line 32: Maybe let's add a new module `aggregator_v2` already? All these beautiful new things can go there.

aptos-move/framework/aptos-framework/sources/aggregator/aggregator.move
```diff
@@ -29,6 +29,15 @@ module aptos_framework::aggregator {
...
+    /// In storage, the value represents the materialized value of the aggregator
+    /// when the snapshot was created.
+    /// During the block execution, the value represents an identifier pointing to the
+    /// Rust `AggregatorSnapshot` which cointains additional details related.
+    struct AggregatorSnapshot has store {
```
georgemitenkov, line 37: What about `AggregatorSnapshot` destruction?

vusirikala: I added only `store` capability because aggregator only had `store` and is destroyed using a native function `destroy`. 
Are we using `destroy` native function instead of the `drop` capability because we want to store the list of `destroyed_aggregators` in `AggregatorData`?

aptos-move/framework/aptos-framework/sources/aggregator/aggregator.move
vusirikala: To distinguish between Aggregator V1 and Aggregator V2, I saw that you are using `enum { Legacy, Ephemeral }` now. After adding Aggregator V2, are we going to remove the codebase for Aggregator V1? Or do we have to retain the code for compatibility reasons?

georgemitenkov: We have to keep it at least for some time because supply is Aggregator V1

aptos-move/framework/aptos-framework/sources/aggregator/aggregator.move
georgemitenkov: Before we were using destroy because it was a state item. Now we might not need it, as we discussed on Thu. It doesn't matter for this PR, let's just not forget this :)

PR #9066, opened 2023-07-12
[Forge] Reduce swarm sizes for various forge jobs.

JoshLind:
### Description
This PR reduces the swarm sizes used by the various forge jobs (to help reduce forge contention and costs). Specifically, the PR contains the following commits:
1. [Land blocking]: Reduce `compat` and `framework_upgrade` validator counts (from 5 to 4, i.e., 3f+1, where f=1); and (ii) reduce `realistic_env_max_load` validators (from 20 to 7, i.e., f=2) and VFNs (from 10 to 5).
2. [PFN]: Reduce the PFN validators (from 20 to 7, i.e., f=2), VFNs (from 10 to 7), and PFNs (from 10 to 7).
3. [Specialized]: Split the `specialized` forge workflow into two separate workflows: (i) a PFN workflow that runs every 2 days; and (ii) a state sync workflow that runs every 3 days.
4. [Unstable] Reduce the `unstable` forge workflow validators (from 20/15/10 to 7, i.e., f=2).
5. [Stable] For the `stable` jobs, we only reduce a few of the swarms sizes such that we still maintain a good variety of tests, i.e., there are currently 12 `stable` jobs, and we're now testing a wide variety of validators counts (i.e., 20->7) and VFN counts (10->7).

### Test Plan
Existing test infrastructure.

### DIFF
--- a/.github/workflows/forge-stable.yaml
+++ b/.github/workflows/forge-stable.yaml
@@ -120,7 +120,7 @@ jobs:
       IMAGE_TAG: ${{ needs.determine-test-metadata.outputs.IMAGE_TAG }}
       FORGE_NAMESPACE: forge-realistic-env-max-load-long-${{ needs.determine-test-metadata.outputs.IMAGE_TAG }}
       FORGE_RUNNER_DURATION_SECS: 7200
-      FORGE_TEST_SUITE: realistic_env_max_load
+      FORGE_TEST_SUITE: realistic_env_max_load_large
       POST_TO_SLACK: true
 
   run-forge-realistic-env-load-sweep:
--- a/testsuite/forge-cli/src/main.rs
+++ b/testsuite/forge-cli/src/main.rs
@@ -1880,8 +1880,8 @@ fn pfn_const_tps(
     add_network_emulation: bool,
 ) -> ForgeConfig {
     ForgeConfig::default()
-        .with_initial_validator_count(NonZeroUsize::new(20).unwrap())
-        .with_initial_fullnode_count(10)
+        .with_initial_validator_count(NonZeroUsize::new(7).unwrap())
+        .with_initial_fullnode_count(7)
         .with_emit_job(EmitJobRequest::default().mode(EmitJobMode::ConstTps { tps: 100 }))
         .add_network_test(PFNPerformance::new(add_cpu_chaos, add_network_emulation))
         .with_genesis_helm_config_fn(Arc::new(|helm_values| {
@@ -1924,8 +1924,8 @@ fn pfn_performance(
 
     // Create the forge config
     ForgeConfig::default()
-        .with_initial_validator_count(NonZeroUsize::new(20).unwrap())
-        .with_initial_fullnode_count(10)
+        .with_initial_validator_count(NonZeroUsize::new(7).unwrap())
+        .with_initial_fullnode_count(7)
         .add_network_test(PFNPerformance::new(add_cpu_chaos, add_network_emulation))
         .with_genesis_helm_config_fn(Arc::new(|helm_values| {
             // Require frequent epoch changes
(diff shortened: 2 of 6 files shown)


perryjrandall (approved):
💸

sionescu (approved):

rustielin (approved):
LGTM, thanks for driving this @JoshLind !! 
We might have to adjust the success criteria for the performance related tests iiuc given the lesser number of VFNs we're submitting traffic through?

testsuite/forge-cli/src/main.rs
```diff
@@ -1526,13 +1518,13 @@ fn realistic_env_max_load_test(duration: Duration, test_cmd: &TestCommand) -> Fo
 
 fn realistic_network_tuned_for_throughput_test() -> ForgeConfig {
     ForgeConfig::default()
-        .with_initial_validator_count(NonZeroUsize::new(12).unwrap())
+        .with_initial_validator_count(NonZeroUsize::new(7).unwrap())
```
igor-aptos, line 1521: I would be more comfortable with this and above change if we add weekly jobs with more nodes.  Should workflow be able to modify number of validators externally?

testsuite/forge-cli/src/main.rs
```diff
@@ -1472,8 +1464,8 @@ fn realistic_env_max_load_test(duration: Duration, test_cmd: &TestCommand) -> Fo
     let duration_secs = duration.as_secs();
     let long_running = duration_secs >= 2400;
     ForgeConfig::default()
-        .with_initial_validator_count(NonZeroUsize::new(20).unwrap())
-        .with_initial_fullnode_count(10)
+        .with_initial_validator_count(NonZeroUsize::new(7).unwrap())
```
igor-aptos, line 1467: Argh we use this for PRs as well as for long running test/ha proxy

testsuite/forge-cli/src/main.rs
```diff
@@ -787,8 +787,8 @@ fn consensus_stress_test() -> ForgeConfig {
 
 fn realistic_env_load_sweep_test() -> ForgeConfig {
     ForgeConfig::default()
-        .with_initial_validator_count(NonZeroUsize::new(20).unwrap())
-        .with_initial_fullnode_count(10)
+        .with_initial_validator_count(NonZeroUsize::new(7).unwrap())
```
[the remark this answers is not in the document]
JoshLind: Okay, rolled back these changes and left them unmodified 😄

testsuite/forge-cli/src/main.rs
JoshLind: > I would be more comfortable with this and above change if we add weekly jobs with more nodes.
Gotcha, this makes sense to me 😄 I'm gonna roll back these changes (to help move the PR forward), but setting up some weekly jobs with more nodes makes sense. That way, we still get signal on large swarms, but without the cost hit of running them daily.

> Should workflow be able to modify number of validators externally?
Yeah, we can probably configure this if we want. Or just add new test flavours that specify different swarm sizes? e.g., small, medium, large swarms?

.github/workflows/forge-stable.yaml
```diff
@@ -210,7 +210,7 @@ jobs:
       FORGE_NAMESPACE: forge-haproxy-${{ needs.determine-test-metadata.outputs.IMAGE_TAG }}
       FORGE_RUNNER_DURATION_SECS: 600
       FORGE_ENABLE_HAPROXY: true
-      FORGE_TEST_SUITE: realistic_env_max_load
+      FORGE_TEST_SUITE: realistic_env_max_load_large
```
igor-aptos, line 213: I'm fine with this not being large

igor-aptos (approved):

.github/workflows/forge-stable.yaml
JoshLind: Awesome, updated, thanks! 😄

PR #9070, opened 2023-07-13
Refactor aptos CLI doc structure

0xjinn:


0xjinn:
Use https://github.com/aptos-labs/aptos-core/pull/9072 instead

PR #9076, opened 2023-07-13
[Indexer] Add transactions by address indexing

bowenyang007:
### Description
It's nice to have a dedicated table to get all transactions that touch an address instead of a view. I'm adding this to the coin_processor because the coin processor already needs to process every transaction. We don't want this to be in default processor currently because we've had instances where default processor takes forever to run migrations and it stalls explorer and wallet.

What qualifies as a transaction associated with an address: 
1. account signed or is part of a multi sig / multi agent
2. account resources / events are directly modified
3. account owns an object where resources / events are modified (only 1 degree)

I called out a bunch of TODOs here. Notably what we're currently unable to do: 
* Recursively find owner of an object
* Find an owner of an object that was deleted
* Find owner of a table, recursively or not

P.S. This PR also fixes a bug where we're not actually recording object deletes. 

### Test Plan
image

### DIFF
--- a/crates/indexer/src/models/coin_models/account_transactions.rs
+++ b/crates/indexer/src/models/coin_models/account_transactions.rs
@@ -1,0 +1,136 @@
+// Copyright © Aptos Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+// This is required because a diesel macro makes clippy sad
+#![allow(clippy::extra_unused_lifetimes)]
+#![allow(clippy::unused_unit)]
+
+use crate::{
+    models::{
+        token_models::v2_token_utils::ObjectWithMetadata, user_transactions::UserTransaction,
+    },
+    schema::account_transactions,
+    util::standardize_address,
+};
+use aptos_api_types::{DeleteResource, Event, Transaction, WriteResource, WriteSetChange};
+use field_count::FieldCount;
+use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
+
+pub type AccountTransactionPK = (String, i64);
+
+#[derive(Debug, Deserialize, FieldCount, Identifiable, Insertable, Serialize)]
+#[diesel(primary_key(account_address, transaction_version))]
+#[diesel(table_name = account_transactions)]
+pub struct AccountTransaction {
+    pub transaction_version: i64,
+    pub account_address: String,
+}
+
+impl AccountTransaction {
+    /// This table will record every transaction that touch an account which could be
+    /// a user account, an object, or a resource account.
+    /// We will consider all transactions that modify a resource or event associated with a particular account.
+    /// We will do 1 level of redirection for now (e.g. if it's an object, we will record the owner as account address).
+    /// We will also consider transactions that the account signed or is part of a multi sig / multi agent.
+    /// TODO: recursively find the parent account of an object
+    /// TODO: include table items in the detection path
+    pub fn from_transaction(
+        transaction: &Transaction,
+    ) -> anyhow::Result<HashMap<AccountTransactionPK, Self>> {
+        let (events, wscs, signatures, txn_version) = match transaction {
+            Transaction::UserTransaction(inner) => (
+                &inner.events,
+                &inner.info.changes,
+                UserTransaction::get_signatures(inner, inner.info.version.0 as i64, 0),
+                inner.info.version.0 as i64,
+            ),
+            Transaction::GenesisTransaction(inner) => (
+                &inner.events,
+                &inner.info.changes,
+                vec![],
+                inner.info.version.0 as i64,
+            ),
+            Transaction::BlockMetadataTransaction(inner) => (
+                &inner.events,
+                &inner.info.changes,
+                vec![],
+                inner.info.version.0 as i64,
+            ),
+            _ => {
+                return Ok(HashMap::new());
+            },
+        };
+        let mut account_transactions = HashMap::new();
+        for sig in &signatures {
+            account_transactions.insert((sig.signer.clone(), txn_version), Self {
+                transaction_version: txn_version,
+                account_address: sig.signer.clone(),
+            });
+        }
+        for event in events {
+            account_transactions.extend(Self::from_event(event, txn_version));
+        }
+        for wsc in wscs {
+            match wsc {
+                WriteSetChange::DeleteResource(res) => {
+                    account_transactions.extend(Self::from_delete_resource(res, txn_version)?);
+                },
+                WriteSetChange::WriteResource(res) => {
+                    account_transactions.extend(Self::from_write_resource(res, txn_version)?);
+                },
+                _ => {},
+            }
+        }
+        Ok(account_transactions)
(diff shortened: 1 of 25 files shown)


crates/indexer/src/models/coin_models/account_transactions.rs
```diff
@@ -0,0 +1,136 @@
...
+    /// a user account, an object, or a resource account.
+    /// We will consider all transactions that modify a resource or event associated with a particular account.
+    /// We will do 1 level of redirection for now (e.g. if it's an object, we will record the owner as account address).
+    /// We will also consider transactions that the account signed or is part of a multi sig / multi agent.
+    /// TODO: recursively find the parent account of an object
+    /// TODO: include table items in the detection path
+    pub fn from_transaction(
+        transaction: &Transaction,
+    ) -> anyhow::Result<HashMap<AccountTransactionPK, Self>> {
+        let (events, wscs, signatures, txn_version) = match transaction {
```
movekevin, line 41: This looks like a fair amount of common code. Is there a common library to call to do this?

crates/indexer/src/models/v2_objects.rs
```diff
@@ -120,35 +108,42 @@ impl Object {
     /// This should never really happen since it's very difficult to delete the entire resource group
```
movekevin, line 108: This is no longer true. Unnamed token objects can be burnt and deleted very thoroughly. For most use cases using aptos_token, we'll see clean deletions so let's make sure we handle this case well (can be in a separate PR though).

crates/indexer/src/models/v2_objects.rs
```diff
@@ -157,4 +152,44 @@ impl Object {
             Ok(None)
         }
     }
+
+    /// This is actually not great because object owner can change. The best we can do now though
```
movekevin, line 156: What's the performance implication for doing this multiple times in the same processing logic? Is this cached somewhere? This can potentially lead to a lot more quota/capacity consumed when running processors on the client side against a grpc streaming service

crates/indexer/src/models/v2_objects.rs
```diff
@@ -157,4 +152,44 @@ impl Object {
...
+    fn get_object_owner(
```
movekevin, line 157: I wonder if this can lead to some performance/failure issue for the processor if the data is stored in remote databases (e.g. in Hive). If a processor has to query the object owner multiple times, this means multiple remote calls

crates/indexer/src/models/v2_objects.rs
```diff
@@ -157,4 +152,44 @@ impl Object {
...
+        conn: &mut PgPoolConnection,
+        object_address: &str,
+    ) -> anyhow::Result<CurrentObject> {
+        let mut retried = 0;
+        while retried < QUERY_RETRIES {
+            retried += 1;
+            match CurrentObjectQuery::get_by_address(object_address, conn) {
+                Ok(res) => {
+                    return Ok(CurrentObject {
```
movekevin, line 166: I thought we also planned to do this for table items?

bowenyang007: Did you read the summary lol. We can't do this for table items performantly.

crates/indexer/src/models/v2_objects.rs
bowenyang007: We don't delete that often so it should be fine for now.

crates/indexer/src/models/v2_objects.rs
bowenyang007: Yea I've seen this already so I'll change the comments.

crates/indexer/src/models/coin_models/account_transactions.rs
bowenyang007: This will need to be cleaned up but for now I don't have time to refactor all that code.

larry-aptos (approved):

crates/indexer/src/models/coin_models/account_transactions.rs
```diff
@@ -0,0 +1,136 @@
...
+    pub account_address: String,
+}
+
+impl AccountTransaction {
+    /// This table will record every transaction that touch an account which could be
+    /// a user account, an object, or a resource account.
+    /// We will consider all transactions that modify a resource or event associated with a particular account.
+    /// We will do 1 level of redirection for now (e.g. if it's an object, we will record the owner as account address).
+    /// We will also consider transactions that the account signed or is part of a multi sig / multi agent.
+    /// TODO: recursively find the parent account of an object
```
rtso, line 36: Should this be done in post-processing instead of in the processor?

crates/indexer/src/models/v2_objects.rs
```diff
@@ -117,38 +105,55 @@ impl Object {
...
+    /// using KV store
+    pub fn from_delete_resource(
         delete_resource: &DeleteResource,
         txn_version: i64,
         write_set_change_index: i64,
+        object_mapping: &HashMap<CurrentObjectPK, CurrentObject>,
+        conn: &mut PgPoolConnection,
     ) -> anyhow::Result<Option<(Self, CurrentObject)>> {
-        if delete_resource.resource.to_string() == "0x1::object::ObjectCore" {
+        if delete_resource.resource.to_string() == "0x1::object::ObjectGroup" {
```
rtso, line 118: What's object group and why do we need to change this?

crates/indexer/src/processors/default_processor.rs
```diff
@@ -524,12 +526,54 @@ impl TransactionProcessor for DefaultTransactionProcessor {
         }
 
         // TODO, merge this loop with above
+        // Moving object handling here because we need a single object
```
rtso, line 529: What's the difference between this new way and the old way of getting the objects? They look like they're doing the same thing.

rtso (approved):

crates/indexer/src/models/coin_models/account_transactions.rs
bowenyang007: we might not have a choice lol.

crates/indexer/src/models/v2_objects.rs
bowenyang007: the old thing wasn't working hahaha. I assuemd it would be ObjectCore but actually when it's deleted it's ObjectGroup. It was just buggy before.

crates/indexer/src/processors/default_processor.rs
bowenyang007: They are. If you notice though that I'm passing all_objects into the from_delete_resource function. That's the main difference. I mean I could've made it work as before as well but the right thing to do is to use a general forloop to avoid the duplicate work so I just refactored this at the same time.

crates/indexer/src/models/v2_objects.rs
rtso: that's weird that the resource type is different for writes and deletes.

movekevin (approved):

PR #9079, opened 2023-07-13
Add a vector::find function that returns index of element matching a predicate

movekevin:
### Description
vector::find as a more useful generic version of vector::index_of

### Test Plan
Unit tests

### DIFF
--- a/aptos-move/framework/move-stdlib/sources/vector.move
+++ b/aptos-move/framework/move-stdlib/sources/vector.move
@@ -176,6 +176,26 @@ module std::vector {
         pragma intrinsic = true;
     }
 
+    /// Return `(true, i)` if there's an element that matches the predicate. If there are multiple elements that match
+    /// the predicate, only the index of the first one is returned.
+    /// Otherwise, returns `(false, 0)`.
+    public inline fun find<Element>(v: &vector<Element>, f: |&Element|bool): (bool, u64) {
+        let find = false;
+        let found_index = 0;
+        let i = 0;
+        let len = length(v);
+        while (i < len) {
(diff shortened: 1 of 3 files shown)


aptos-move/framework/aptos-framework/sources/code.move
```diff
@@ -90,6 +90,9 @@ module aptos_framework::code {
     /// Creating a package with incompatible upgrade policy is disabled.
     const EINCOMPATIBLE_POLICY_DISABLED: u64 = 0x8;
 
+    /// Cannot find specified package.
+    const EPACKAGE_NOT_FOUND: u64 = 0x9;
```
gerben-stavenga, line 94: why is this in this pr?

aptos-move/framework/move-stdlib/sources/vector.move
```diff
@@ -176,6 +176,25 @@ module std::vector {
...
+    /// Otherwise, returns `(false, 0)`.
+    public inline fun find<Element>(v: &vector<Element>, f: |&Element|bool): (bool, u64) {
+        let find = false;
+        let found_index = 0;
+        let i = 0;
+        let len = length(v);
+        while (i < len) {
+            if (f(borrow(v, i))) {
+                find = true;
+                found_index = i;
```
gerben-stavenga, line 190: maybe a comment that restrictions of inline prevents a return

aptos-move/framework/aptos-framework/sources/code.move
movekevin: Removed

gerben-stavenga (approved):

lightmark (approved):

Issue #6936, opened 2023-03-06
[Feature Request] Incremental package upload


wrwg:
# 🚀 Feature Request

Incremental population of packages on chain should be supported. A set of modules can be provided on package deployment to either upgrade or add new modules. Since module deletion is not not supported anyway today, this is a conservative extension of existing behavior.

## Motivation

The 64k transaction size limit severely restricts builders with larger Move packages. Packages have to be split introducing security issues because friend visibility is not longer there. 

With this features, builders can upload a larger package in multiple steps. They can also later upgrade the package on a more granular level, e.g. on module base, which safes gas.

wrwg:
@alnoki indeed the scenario you are describing would not work with what I'm proposing. You can only chunk on the level of whole modules with the proposal here. Actually, it must be strongly connected components w.r.t. the friend relations, because Move currently does not allow dangling friend references. 

On the other hand, the solution I'm proposing would allow a smart deployer (tool) to automatically compute the delta (that is the modules actually changed) on package upgrade, and could this way drastically reduce gas cost for repeated upgrade.

Thoughts?

movekevin:
Arbitrarily cutting a package's bytecode into several chunks and uploading them on-chain does intuitively sound messy. This can mean that there's no validation that can protect the data from being corrupted, which developers would only realize upon uploading all the chunks and trying to call code::deploy.

@wrwg If I understand correctly, you're suggesting a more semantic breakdown of the packages where devs can upload module one by one for example (as long as they're in the right dependency order)?

wrwg:
> Arbitrarily cutting a package's bytecode into several chunks and uploading them on-chain does intuitively sound messy. 

Perhaps a generic mechanism for staging data would be an interesting feature for the aptos stdlib.

> @wrwg If I understand correctly, you're suggesting a more semantic breakdown of the packages where devs can upload module one by one for example (as long as they're in the right dependency order)?

Yes, though it is not just the dependency order which can go wrong. For example, you may have A and B in the package, now you compile A and B with some new stuff in A, but only update B on chain. Then bytecode verification will fail. All situations I can think of should be protected by the bytecode verifier. 

In general, we can extend the CLI to compute a delta to the package on chain. This computed subset of uploaded modules should never produce an error. In addition, we can add options for chunking.

alnoki:
> > Arbitrarily cutting a package's bytecode into several chunks and uploading them on-chain does intuitively sound messy.
> 
[...]

PR #9080, opened 2023-07-13
[Move] Add a function to add source code after publish

movekevin:
### Description
Package size always doubles with source code and it's really easy to exceed the transaction limit of 64 KB. Deployers can publish code without the source but they currently can add them later. This adds a function that allowing adding the source code separately.

### Test Plan
Manual tests

### DIFF
--- a/aptos-move/framework/aptos-framework/sources/code.move
+++ b/aptos-move/framework/aptos-framework/sources/code.move
@@ -90,6 +90,12 @@ module aptos_framework::code {
     /// Creating a package with incompatible upgrade policy is disabled.
     const EINCOMPATIBLE_POLICY_DISABLED: u64 = 0x8;
 
+    /// Cannot find specified package.
+    const EPACKAGE_NOT_FOUND: u64 = 0x9;
+
+    /// Only the source code for existing modules can be updated.
+    const EMODULE_NOT_FOUND: u64 = 0xA;
+
     /// Whether unconditional code upgrade with no compatibility check is allowed. This
     /// publication mode should only be used for modules which aren't shared with user others.
     /// The developer is responsible for not breaking memory layout of any resources he already
@@ -186,6 +192,35 @@ module aptos_framework::code {
     public entry fun publish_package_txn(owner: &signer, metadata_serialized: vector<u8>, code: vector<vector<u8>>)
     acquires PackageRegistry {
         publish_package(owner, util::from_bytes<PackageMetadata>(metadata_serialized), code)
+    }
+
+    /// Allowing the owner of a package to upload source code after the package was published.
+    /// The source code will be mapped to existing list of modules in PackageRegistry by name. New modules and missing
+    /// ones will be ignored.
+    public entry fun add_source_code(
+        owner: &signer,
+        package_name: String,
+        modules: vector<vector<u8>>,
+    ) acquires PackageRegistry {
+        let package_registry = borrow_global_mut<PackageRegistry>(signer::address_of(owner));
+        let (found, i) = vector::find(&package_registry.packages, |p| {
+            let p: &PackageMetadata = p;
+            p.name == package_name
+        });
+        assert!(found, error::invalid_argument(EPACKAGE_NOT_FOUND));
+        let package_metadata = vector::borrow_mut(&mut package_registry.packages, i);
+        vector::for_each(modules, |serialized_module| {
+            let m = util::from_bytes<ModuleMetadata>(serialized_module);
+            let source = m.source;
+            let name = m.name;
+            let (found, index) = vector::find(&package_metadata.modules, |m| {
+                let m: &ModuleMetadata = m;
+                m.name == name
+            });
+            assert!(found, error::invalid_argument(EMODULE_NOT_FOUND));
+            let module_metadata = vector::borrow_mut(&mut package_metadata.modules, index);
+            module_metadata.source = source;
+        });
     }
 
     // Helpers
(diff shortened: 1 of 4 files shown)


aptos-move/framework/aptos-framework/sources/code.move
```diff
@@ -188,6 +191,36 @@ module aptos_framework::code {
...
+        let package_metadata = vector::borrow_mut(&mut package_registry.packages, i);
+        vector::for_each(modules, |serialized_module| {
+            let m = util::from_bytes<ModuleMetadata>(serialized_module);
+            let source = m.source;
+            let name = m.name;
+            let (found, index) = vector::find(&package_metadata.modules, |m| {
+                let m: &ModuleMetadata = m;
+                m.name == name
+            });
+            if (found) {
```
gerben-stavenga, line 217: Should this not be an assert? It's strange to just silently drop module updates that aren't found.

movekevin: Good point. I was thinking it's more symmetric as this allows missing modules as well. But assertion would make it more clear there are extra modules involved

gerben-stavenga (approved):

aptos-move/framework/aptos-framework/sources/code.move
```diff
@@ -188,6 +194,35 @@ module aptos_framework::code {
         publish_package(owner, util::from_bytes<PackageMetadata>(metadata_serialized), code)
     }
 
+    /// Allowing the owner of a package to upload source code after the package was published.
```
wrwg, line 197: Please use complete english sentences. "Allows the owner..."

wrwg (commented):
How do you usually test this? Right now there seems nothing in place. Add an e2e-test together with the other code publishing tests?

davidiw:
Is this really better than building a small library / module that can support chunked module publishing?

wrwg:
> 

We do not want to do chunked module publishing, we want to this. https://github.com/aptos-labs/aptos-core/issues/6936

Until its done this seems to be a low hanging fruit.

Not this is already discussed on slack as well...

davidiw:
> > 
> 
> We do not want to do chunked module publishing, we want to this. #6936
> 
> Until its done this seems to be a low hanging fruit.
> 
> Note this is already discussed on slack as well...

Have we confirmed that this change will sufficiently help developers?

Also carrying on the pros and cons of chunking in that issue.

alnoki:
@movekevin @davidiw is there anyway to add a check on source code listings to verify that it compiles to the bytecode?

This disallows the scenario where bytecode is rug code but source code listings are valid

wrwg:
> @movekevin @davidiw is there anyway to add a check on source code listings to verify that it compiles to the bytecode?
> 
> This disallows the scenario where bytecode is rug code but source code listings are valid

This isn't possible that easily, and even with the current model (without source update), someone can easily fake the sources. However, the CLI has an option to download a package and verify it locally.

movekevin:
> > @movekevin @davidiw is there anyway to add a check on source code listings to verify that it compiles to the bytecode?
> > This disallows the scenario where bytecode is rug code but source code listings are valid
> 
> This isn't possible that easily, and even with the current model (without source update), someone can easily fake the sources. However, the CLI has an option to download a package and verify it locally.

We’re planning to add CLI support for this

> > We do not want to do chunked module publishing, we want to this. #6936
> > Until its done this seems to be a low hanging fruit.
> > Note this is already discussed on slack as well...
> 
> Have we confirmed that this change will sufficiently help developers?
> 
> Also carrying on the pros and cons of chunking in that issue.

Yes. Ive tested this manually. This is also useful for cases where people accidentally skip including  artifacts and are now stuck because their package is immutable

lightmark:
tbh if we never verify the code and bytecode are consistent on-chain, why do we store code on-chain contributing to the state root hash.

wrwg:
> tbh if we never verify the code and bytecode are consistent on-chain, why do we store code on-chain contributing to the state root hash.

Verification can be easily done by downloading the package from chain and compile it locally. Its just not easily possible to do at deployment time.

lightmark:
> > tbh if we never verify the code and bytecode are consistent on-chain, why do we store code on-chain contributing to the state root hash.
> 
> Verification can be easily done by downloading the package from chain and compile it locally. Its just not easily possible to do at deployment time.

I get that. What I tried to stress there is, now the blockchain has been used as a storage for unverified code unnecessarily. If the code is not verified on-chain, we don't have to store it onchain, which get authenticated by the root hash unnecessarily. Especially if someone uploads inconsistent/garbage/rug code.

alnoki:
> What I tried to stress there is, now the blockchain has been used as a storage for unverified code unnecessarily.

@lightmark I agree that this is a design flaw. I suggest that only verified code be stored

wrwg:
> > What I tried to stress there is, now the blockchain has been used as a storage for unverified code unnecessarily.
> 
> @lightmark I agree that this is a design flaw. I suggest that only verified code be stored

The feature is rather popular as it allows user to browse the code they are executing. Its used by the explorer. The package system is also able to actually compile against sources stored on chain, instead of compiling against github, which is way more insecure. We can get to verification at some point, but it is not a top priority. As said, you can with a single CLI command verify whether the package you see onchain is consistent w.r.t. bytecode/source. Also note that "verification" ala Etherscan is inherently centralized, so not really much better.

davidiw:
gonna close out as we now provide some examples of at least how to chunk across modules and metadata. Still needs more work to actually support chunking a single module...

PR #9084, opened 2023-07-13
Add PubSub Protobufs

just-in-chang:
# Process

1. Copy and paste relevant protobuf files
2. Install buf `brew install bufbuild/buf/buf`
3. Install plugins `cargo install protoc-gen-prost; cargo install protoc-gen-prost-serde; cargo install protoc-gen-prost-crate; cargo install protoc-gen-tonic`
4. Generate files `buf generate`

Fails to compile for generated `xx.serde.rs` files


crates/aptos-protos/Cargo.toml
```diff
@@ -12,5 +12,6 @@ edition = "2021"
 [dependencies]
 pbjson = "0.4.0"
 prost = "0.11.3"
+prost-types = "0.11.3"
```
larry-aptos, line 15: try `0.11.1`

just-in-chang: Still broken 😦

larry-aptos:
also... looks like `https://github.com/yoshidan/google-cloud-rust/tree/main/pubsub` is doing similar. maybe switch to it without importing the proto?

just-in-chang:
> also... looks like `https://github.com/yoshidan/google-cloud-rust/tree/main/pubsub` is doing similar. maybe switch to it without importing the proto?

`https://github.com/yoshidan/google-cloud-rust/tree/main/googleapis` will probably just use this instead

PR #9085, opened 2023-07-13
[Spec] Add more high-level specs

rahxephon89:
### Description

This PR adds more high-level specs and fixes some existing spec issues.

### DIFF
--- a/aptos-move/framework/aptos-framework/sources/aptos_account.spec.move
+++ b/aptos-move/framework/aptos-framework/sources/aptos_account.spec.move
@@ -38,7 +38,7 @@ spec aptos_framework::aptos_account {
 
         aborts_if exists<coin::CoinStore<AptosCoin>>(to) && global<coin::CoinStore<AptosCoin>>(to).frozen;
         ensures exists<aptos_framework::account::Account>(to);
-        ensures exists<AptosCoin>(to);
+        ensures exists<coin::CoinStore<AptosCoin>>(to);
     }
 
     spec assert_account_exists(addr: address) {
--- a/aptos-move/framework/aptos-framework/sources/aptos_governance.spec.move
+++ b/aptos-move/framework/aptos-framework/sources/aptos_governance.spec.move
@@ -499,7 +499,6 @@ spec aptos_framework::aptos_governance {
         let signer_cap = simple_map::spec_get(governance_responsibility.signer_caps, signer_address);
         let addr = signer_cap.account;
         ensures signer::address_of(result) == addr;
-        ensures table::spec_get(voting_forum.proposals, proposal_id).is_resolved == true;
     }
 
     /// Address @aptos_framework must exist ApprovedExecutionHashes and GovernanceProposal.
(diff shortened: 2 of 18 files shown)


aptos-move/framework/aptos-framework/sources/aptos_account.spec.move
```diff
@@ -38,7 +38,7 @@ spec aptos_framework::aptos_account {
 
         aborts_if exists<coin::CoinStore<AptosCoin>>(to) && global<coin::CoinStore<AptosCoin>>(to).frozen;
         ensures exists<aptos_framework::account::Account>(to);
-        ensures exists<AptosCoin>(to);
```
junkil-park, line 41: Did Prover prove this previously?

aptos-move/framework/aptos-framework/sources/aptos_governance.spec.move
```diff
@@ -499,7 +499,6 @@ spec aptos_framework::aptos_governance {
         let signer_cap = simple_map::spec_get(governance_responsibility.signer_caps, signer_address);
         let addr = signer_cap.account;
         ensures signer::address_of(result) == addr;
-        ensures table::spec_get(voting_forum.proposals, proposal_id).is_resolved == true;
```
junkil-park, line 502: Why is this removed?

aptos-move/framework/aptos-framework/sources/event.spec.move
```diff
@@ -7,7 +7,7 @@ spec aptos_framework::event {
     spec emit_event {
         pragma opaque;
         aborts_if [abstract] false;
-        ensures handle_ref.counter == old(handle_ref.counter) + 1;
+        ensures handle_ref.counter < MAX_U64 ==> handle_ref.counter == old(handle_ref.counter) + 1;
```
junkil-park, line 10: Do we need to add the condition `handle_ref.counter < MAX_U64` although it can be proved without it?

junkil-park (approved):
Thanks a lot! I left a few comments.

aptos-move/framework/aptos-framework/sources/aptos_governance.spec.move
rahxephon89: this is duplicated with an ensures at line 488.

aptos-move/framework/aptos-framework/sources/event.spec.move
rahxephon89: Yes, there is an assume clause in the `emit_event` function. Without this condition, there would be an inconsistency issue.

aptos-move/framework/aptos-framework/sources/transaction_validation.spec.move
```diff
@@ -175,6 +175,7 @@ spec aptos_framework::transaction_validation {
 
         aborts_if !account::exists_at(gas_payer);
         aborts_if !(fee_payer_public_key_hash == account::get_authentication_key(gas_payer));
+        aborts_if !features::spec_fee_payer_enabled();
```
gerben-stavenga, line 178: how come the prover accepted this before without this

rahxephon89: Because there is a pragma above set to 120s which means only when the global time out is set to greater or equal to 120s, this function will be verified. Currently the timeout is set to 80s by default and the verification is skipped

davidiw (approved):

Issue #8167, opened 2023-05-11
[SEV] Convert transaction bytes to human readable format.


lbmeiyi:
sev action item from Several invariant violation triggered on devnet https://aptoslabs.pagerduty.com/incidents/Q3CREMRZ8GKCXT

junkil-park:
Submitted a PR for this: https://aptos-org.slack.com/archives/C03N9HNSUB1/p1689279594627749

PR #9087, opened 2023-07-13
[aptos-debugger] Added the bcs-txn-decoder

junkil-park:
### Description
* Added the bcs-exn-decoder to the aptos-debugger
* Added `README.md` to explain how to use it

Resolves https://github.com/aptos-labs/aptos-core/issues/8167.

### Test Plan
`cargo run -p aptos-debugger --bin bcs-txn-decoder`


runtian-zhou (commented):
Was wondering if we should just move the reading part as an argument to the original debugger so that we don't need two separate binaries?

gerben-stavenga (approved):

aptos-move/aptos-vm/README.md
```diff
@@ -7,7 +7,7 @@ custom_edit_url: https://github.com/aptos-labs/aptos-core/edit/main/language/mov
 # MoveVM Runtime
 
 The MoveVM runtime is the verification and execution engine for the Move
-bytecode format. The runtime is imported and loaded in 2 modes:
+bytecode format. The runtime is imported and loaded in 2 mod0es:
```
rahxephon89, line 10: Is this a typo?

junkil-park: Oops, it's fixed now. Thanks!

rahxephon89 (approved):

junkil-park:
> Was wondering if we should just move the reading part as an argument to the original debugger so that we don't need two separate binaries?

Right, I tried that initially, but it didn't nicely fit into the current CLI of `aptos-debugger`. Could this PR go as it is, and be merged into the `aptos-debugger` later when we refactor the CLI?

PR #9088, opened 2023-07-13
[TS SDK] Clean up tests

0xmaayan:
### Description
We can't really test this function now as we need to change the `seed` and then precompile the module to be published with the new resource account address before everytest, obviously this is not feasible when running ts sdk tests.
more context https://aptos-org.slack.com/archives/C03N9HNSUB1/p1689278043871009


gregnazario (approved):

lightmark (approved):
:(

0xjinn (approved):

PR #9090, opened 2023-07-13
[gha] run ts sdk tests on latest main best effort

rustielin:
### Description

Run TS SDK tests against a local testnet spun up via docker against the latest docker image built on main. A new composite action called `get-latest-docker-image-tag` gets the latest image built on a given branch. It checks out the given branch and uses the existing `find_latest_image.py` script to crawl the git history and check which commits have images built.

### Test Plan

CI run on push: https://github.com/aptos-labs/aptos-core/actions/runs/5548754559/jobs/10132092825?pr=9090

### DIFF
--- a/.github/workflows/ts-sdk-e2e-tests.yaml
+++ b/.github/workflows/ts-sdk-e2e-tests.yaml
@@ -32,7 +32,7 @@ jobs:
           required-permission: write
           comment-not-permitted: Sorry, you don't have permission to trigger this workflow.
 
-  run-tests-devnet:
+  run-tests-main-branch:
     needs: [permission-check]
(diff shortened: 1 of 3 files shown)


.github/workflows/ts-sdk-e2e-tests.yaml
```diff
@@ -13,6 +13,7 @@ on:
   push:
     branches:
       - main
+      - rustielin/ts-sdk-main
```
0xmaayan, line 16: prob need to remove =)

0xmaayan (approved):
Thank you!

gregnazario (approved):

PR #9096, opened 2023-07-14
gather /metrics to /tmp

brianolson:
### Description

Add "gather_metrics" mode to forge. Useful for local swarm. Grab /metrics from each node at beginning and of some run time, save data to files in /tmp/

### Test Plan

This is developer test code. I've been using it. It's useful to me.


perryjrandall (approved):
Looks awesome

rustielin (approved):
LGTM

Should we just do this for every test workload? or make it an option?

brianolson:
I built "gather_metrics" to be : 1. gather metrics; 2. sleep 60 seconds; 3. gather metrics.

Now I think about this more and I should take advantage of forge's composable units and make a 'gather metrics' thing and a 'sleep 60' thing, and we could insert 'gather metrics' into a bunch of other workflows.

perryjrandall:
> I built "gather_metrics" to be : 1. gather metrics; 2. sleep 60 seconds; 3. gather metrics.
> 
> Now I think about this more and I should take advantage of forge's composable units and make a 'gather metrics' thing and a 'sleep 60' thing, and we could insert 'gather metrics' into a bunch of other workflows.

Yeah like adding this on the swarm interface, would be neat to be able to say "swarm get metrics" which could be implemented locally or remote

brianolson:
It's modularized now so you can do things like
```
ForgeConfig::default()
        .add_network_test(GatherMetrics)
        .add_network_test(Delay::new(60))
        .add_network_test(GatherMetrics)
```

PR #9098, opened 2023-07-14
Move example of a decentralized lottery based on the drand VRF

alinush:
### Description

This PR consists of an example showing how to:
1. Verify randomness from the [`drand`](https://drand.love) randomness beacon
2. Create a decentralized lottery whose winner is picked based on future randomness generated by `drand`

**WARNING:** The decentralized lottery example has **NOT** been audited and is likely to have security bugs. Proceed with caution. In particular, one important security concern is accounting for the clock drift between the Aptos validators and the `drand` beacon.


aptos-move/move-examples/drand/sources/drand.move
```diff
@@ -0,0 +1,136 @@
+/// The drand randomness beacon outputs a piece of randomness `r_i` for every round `i` such that anybody can verify
```
jillxuu, line 1: should we add this drand contract to the framework instead?

aptos-move/move-examples/drand/sources/drand.move
```diff
@@ -0,0 +1,136 @@
...
+        let sig = extract(&mut deserialize<G1, FormatG1Compr>(&bytes));
+        let msg_hash = hash_to<G1, HashG1XmdSha256SswuRo>(&DRAND_DST, &round_number_to_bytes(round));
+        assert!(eq(&pairing<G1, G2, Gt>(&msg_hash, &pk), &pairing<G1, G2, Gt>(&sig, &one<G2>())), 1);
+        option::some(Randomness {
+            bytes: sha3_256(bytes)
+        })
+    }
+
+    /// Returns a number in $[0, max)$ given some drand (verified) `randomness` and a `Type`.
+    /// Note: This is a one-shot API that consumes the `randomness`.
```
gerben-stavenga, line 71: i dont see a 'Type'

aptos-move/move-examples/drand/sources/drand.move
```diff
@@ -0,0 +1,136 @@
...
+    /// A randomness object, created from a successfully-verified piece of drand randomness.
+    /// This object can be converted into a uniform random integer.
+    struct Randomness has drop {
+        bytes: vector<u8>
+    }
+
+    /// Checks if the randomness in `bytes` verifies for the specified `round`.
+    /// If it verifies, returns the actual randomness, which is a hash function applied over `bytes`.
+    public fun verify_and_extract_randomness(
+        bytes: vector<u8>,
```
gerben-stavenga, line 58: why not call this signature? it's confusing with the fact you are returning "bytes" as  a field of randomness but it's certainly another "bytes"

aptos-move/move-examples/drand/sources/drand.move
```diff
@@ -0,0 +1,136 @@
...
+        let max_128 = (max as u128);
+
+        // Ugh, we have to manually deserialize this into a u64 (8 chunks of 8 bits each)
+        while (!vector::is_empty(&last_128_bits)) {
+            let byte = vector::pop_back(&mut last_128_bits);
+            num = num << 8;
+            num = num + (byte as u128);
+        };
+
+        ((num % max_128) as u64)
```
gerben-stavenga, line 90: technically not uniform

aptos-move/move-examples/drand/sources/drand.move
```diff
@@ -0,0 +1,136 @@
...
+        let msg_hash = hash_to<G1, HashG1XmdSha256SswuRo>(&DRAND_DST, &round_number_to_bytes(round));
+        assert!(eq(&pairing<G1, G2, Gt>(&msg_hash, &pk), &pairing<G1, G2, Gt>(&sig, &one<G2>())), 1);
+        option::some(Randomness {
+            bytes: sha3_256(bytes)
+        })
+    }
+
+    /// Returns a number in $[0, max)$ given some drand (verified) `randomness` and a `Type`.
+    /// Note: This is a one-shot API that consumes the `randomness`.
+    public fun random_number(randomness: Randomness, max: u64): u64 {
```
gerben-stavenga, line 72: probably you want to add a sequence number, now you only get one random number. I think there are plenty of of cases where you'd want to generate more random numbers

gerben-stavenga: Or perhaps &mut Randomness and update the randomness according to some reasonable pseudo-random algo

aptos-move/move-examples/drand/sources/drand.move
alinush: Oops, will fix.

aptos-move/move-examples/drand/sources/drand.move
alinush: I think it would be a bit premature: there's many other randomness beacons we (could) support and it's unclear what the criteria should be for adding one to the framework. An example works fine, I think.

aptos-move/move-examples/drand/sources/drand.move
alinush: Fixed.

aptos-move/move-examples/drand/sources/drand.move
alinush: Yes. But this is just a specific example for a lottery where only one random # is needed. Trying to get this out there for now. More complicated usecases require a fancier `Randomness` interface. See [AIP-41](https://github.com/aptos-foundation/AIPs/blob/main/aips/aip-41.md) for such a proposal.

aptos-move/move-examples/drand/sources/drand.move
alinush: Yes, indeed. In fact, I should keep`num` 256 bits rather than 128 in order to minimize the bias...

alinush: AFAICT, the bias is computationally-indistiguinshable, but let me know if you disagree. @zjma and I analyzed this a few months ago.

gerben-stavenga: the relative error is bounded by 2^-64. I agree it's unlikely we even have security guarantees that the cryptofgraphic hash functions won't have non-uniformity errors of such size. What I mostly meant is that some comment about the error analysis and compared with reasonable bounds on errors that it's very acceptable

gerben-stavenga (approved):

zjma (approved):

PR #9101, opened 2023-07-14
NFT Metadata Crawler

just-in-chang:
### Description

- Set up database migrations
- Define database connection pool
- Define skeleton functions for parsing flow

Fixes EI-249


ecosystem/nft-metadata-crawler/nft-metadata-crawler-parser/migrations/2023-06-27-001532_create_tables/up.sql
```diff
@@ -0,0 +1,14 @@
...
+  cdn_json_uri VARCHAR,
+  cdn_image_uri VARCHAR,
+  cdn_animation_uri VARCHAR,
+  image_resizer_retry_count INT NOT NULL,
+  json_parser_retry_count INT NOT NULL,
+  last_updated TIMESTAMP NOT NULL DEFAULT NOW()
+);
+
+CREATE INDEX nft_raw_image_uri ON nft_metadata_crawler_uris (raw_image_uri);
+CREATE INDEX nft_raw_animation_uri ON nft_metadata_crawler_uris (raw_animation_uri);
```
bowenyang007, line 14: add IF NOT EXISTS. It helps.

ecosystem/nft-metadata-crawler/nft-metadata-crawler-parser/src/db.rs
```diff
@@ -0,0 +1,35 @@
...
+    upsert::excluded,
+    ExpressionMethods, PgConnection, RunQueryDsl, SelectableHelper,
+};
+use std::error::Error;
+
+pub fn upsert_uris(
+    conn: &mut PooledConnection<ConnectionManager<PgConnection>>,
+    entry: NFTMetadataCrawlerURIs,
+) -> Result<NFTMetadataCrawlerURIs, Box<dyn Error + Send + Sync>> {
+    use schema::nft_metadata_crawler_uris::dsl::*;
```
bowenyang007, line 15: Please add some logging in case it fails.

ecosystem/nft-metadata-crawler/nft-metadata-crawler-parser/src/lib.rs
```diff
@@ -0,0 +1,19 @@
...
+pub mod parser;
+pub mod schema;
+
+use diesel::{
+    pg::PgConnection,
+    r2d2::{ConnectionManager, Pool},
+};
+
+pub fn establish_connection_pool() -> Pool<ConnectionManager<PgConnection>> {
+    let database_url = std::env::var("DATABASE_URL").expect("DATABASE_URL must be set");
```
bowenyang007, line 14: Let's use a yaml file b/c that's what we're using everywhere else.

ecosystem/nft-metadata-crawler/nft-metadata-crawler-parser/src/main.rs
```diff
@@ -0,0 +1,13 @@
...
+
+use diesel::{
+    r2d2::{ConnectionManager, Pool},
+    PgConnection,
+};
+use nft_metadata_crawler_parser::establish_connection_pool;
+
+#[tokio::main]
+async fn main() {
+    println!("Starting parser");
```
bowenyang007, line 11: Use tracing library 
https://github.com/aptos-labs/aptos-core/blob/main/ecosystem/indexer-grpc/indexer-grpc-parser/src/worker.rs#L35

ecosystem/nft-metadata-crawler/nft-metadata-crawler-parser/src/models.rs
```diff
@@ -0,0 +1,18 @@
+// Copyright © Aptos Foundation
+
+use diesel::prelude::*;
+
+#[derive(Clone, Insertable, Queryable, Selectable)]
+#[diesel(table_name = crate::schema::nft_metadata_crawler_uris)]
+#[diesel(check_for_backend(diesel::pg::Pg))]
```
bowenyang007, line 7: Let's standardize. 
https://github.com/aptos-labs/aptos-core/blob/main/ecosystem/indexer-grpc/indexer-grpc-parser/src/models/coin_models/coin_infos.rs#L15-L29

ecosystem/nft-metadata-crawler/nft-metadata-crawler-parser/src/models.rs
```diff
@@ -0,0 +1,18 @@
...
+pub struct NFTMetadataCrawlerURIs {
+    pub token_uri: String,
+    pub raw_image_uri: Option<String>,
+    pub raw_animation_uri: Option<String>,
+    pub cdn_json_uri: Option<String>,
+    pub cdn_image_uri: Option<String>,
+    pub cdn_animation_uri: Option<String>,
+    pub image_resizer_retry_count: i32,
+    pub json_parser_retry_count: i32,
+    pub last_updated: chrono::NaiveDateTime,
```
bowenyang007, line 17: Better to let postgres set this directly instead of us passing in a timestamp. This has implications on data replication which might or might not be relevant in this case but good practice.

bowenyang007: It's a little bit more tedious b/c you have to setup a different query struct, but not that bad. 
https://github.com/aptos-labs/aptos-core/blob/main/ecosystem/indexer-grpc/indexer-grpc-parser/src/models/coin_models/coin_infos.rs#L31-L46

ecosystem/nft-metadata-crawler/nft-metadata-crawler-parser/src/parser.rs
```diff
@@ -0,0 +1,194 @@
...
+use serde_json::Value;
+use std::error::Error;
+
+// Stuct that represents a parser for a single entry from queue
+pub struct Parser<'a> {
+    pub entry: NFTMetadataCrawlerEntry,
+    model: NFTMetadataCrawlerURIs,
+    format: ImageFormat,
+    bucket: String,
+    ts: &'a dyn TokenSource,
```
bowenyang007, line 24: why?

ecosystem/nft-metadata-crawler/nft-metadata-crawler-parser/src/parser.rs
```diff
@@ -0,0 +1,194 @@
...
+                image_resizer_retry_count: 0,
+                json_parser_retry_count: 0,
+                last_updated: Utc::now().naive_utc(),
+            },
+            entry: e,
+            format: ImageFormat::Jpeg,
+            bucket: b,
+            ts: t,
+            force: f,
+        }
```
bowenyang007, line 47: don't be lazy with variable names.

ecosystem/nft-metadata-crawler/nft-metadata-crawler-parser/src/parser.rs
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+    pub async fn parse(
+        &mut self,
+        conn: &mut PooledConnection<ConnectionManager<PgConnection>>,
+    ) -> Result<(), Box<dyn Error + Send + Sync>> {
```
bowenyang007, line 53: Generally I try to simplify function signatures. Not sure why we need a box here when we can either panic or convert the error. Use anyhow for example.

ecosystem/nft-metadata-crawler/nft-metadata-crawler-parser/src/parser.rs
```diff
@@ -0,0 +1,194 @@
...
+        // Deduplication
+        if nft_metadata_crawler_uris::table
+            .find(&self.entry.token_uri)
+            .first::<NFTMetadataCrawlerURIs>(conn)
+            .is_ok()
+        {
+            if self.force {
+                self.log("Found URIs entry but forcing parse");
```
bowenyang007, line 61: tracing.

ecosystem/nft-metadata-crawler/nft-metadata-crawler-parser/src/parser.rs
```diff
@@ -0,0 +1,194 @@
...
+        &mut self,
+        conn: &mut PooledConnection<ConnectionManager<PgConnection>>,
+    ) -> Result<(), Box<dyn Error + Send + Sync>> {
+        // Deduplication
+        if nft_metadata_crawler_uris::table
+            .find(&self.entry.token_uri)
+            .first::<NFTMetadataCrawlerURIs>(conn)
+            .is_ok()
+        {
+            if self.force {
```
bowenyang007, line 60: If force, why check the table?

ecosystem/nft-metadata-crawler/nft-metadata-crawler-parser/src/parser.rs
```diff
@@ -0,0 +1,194 @@
...
+            },
+            entry: e,
+            format: ImageFormat::Jpeg,
+            bucket: b,
+            ts: t,
+            force: f,
+        }
+    }
+
+    pub async fn parse(
```
bowenyang007, line 50: This function needs to be split into several other functions. If you want clean code it might be worth creating a separate utils file.

ecosystem/nft-metadata-crawler/nft-metadata-crawler-utils/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+name = "nft-metadata-crawler-utils"
```
bowenyang007, line 2: What's the point of splitting out into 2 packages?

bowenyang007 (commented):
Added some comments. Code structure can be improved. I didn't go too deep into the logic yet, mostly commented on structure.

ecosystem/nft-metadata-crawler/nft-metadata-crawler-parser/src/models.rs
just-in-chang: Can we use this? https://github.com/aptos-labs/aptos-core/blob/main/ecosystem/indexer-grpc/indexer-grpc-parser/migrations/00000000000000_diesel_initial_setup/up.sql#L19

ecosystem/nft-metadata-crawler/nft-metadata-crawler-utils/Cargo.toml
just-in-chang: Originally had all GCS and pubsub logic in a separate package so they can be used by uri retriever, parser, and any other processors

ecosystem/nft-metadata-crawler/nft-metadata-crawler-parser/src/parser.rs
just-in-chang: To repeatedly get GCP service account auth bearer token in case it expires

bowenyang007: No i'm asking why the lifetime variable?

ecosystem/nft-metadata-crawler/nft-metadata-crawler-parser/src/db.rs
```diff
@@ -0,0 +1,41 @@
+// Copyright © Aptos Foundation
+
+use crate::{models::NFTMetadataCrawlerURIs, schema};
+use diesel::{
+    r2d2::{ConnectionManager, PooledConnection},
+    upsert::excluded,
+    ExpressionMethods, PgConnection, RunQueryDsl,
+};
+use tracing::{error, info};
```
larry-aptos, line 9: i didn't see you setup the tracing-subscriber. do you plan to do in next pr?

ecosystem/nft-metadata-crawler/nft-metadata-crawler-utils/Cargo.toml
bowenyang007: But then you're calling this crawler-utils. I'd say that if it's customize code to crawler, merge them together b/c this is a single service. If you want to make some generic utils, maybe rename this and move this out. Currently I don't see your code as generic enough for other use cases so better to just combine.

bowenyang007 (commented):
Please add some comments in all the functions and if needed within the functions.

ecosystem/nft-metadata-crawler/nft-metadata-crawler-utils/src/gcs.rs
```diff
@@ -0,0 +1,93 @@
...
+    let filename = format!("json_{}.json", id);
+    let url = format!(
+        "https://storage.googleapis.com/upload/storage/v1/b/{}/o?uploadType=media&name={}",
+        bucket, filename
+    );
+    let json_string = json.to_string();
+
+    let res = client
+        .post(url)
+        .bearer_auth(ts.token().await?.access_token)
```
larry-aptos, line 28: what'll happen if access token expires? do we refresh it?

ecosystem/nft-metadata-crawler/nft-metadata-crawler-parser/src/main.rs
```diff
@@ -0,0 +1,44 @@
...
+}
+
+#[tokio::main]
+async fn main() {
+    info!("Starting Parser");
+    let args = <ServerArgs as clap::Parser>::parse();
+    let config =
+        load_config_from_yaml::<ParserConfig>(args.config_path).expect("Unable to load config");
+    let pool = establish_connection_pool(config.database_url.clone());
+    let conn = pool.get().unwrap();
```
larry-aptos, line 31: for better performance, we want to pass the pool around and get the connection when we start to process the db, otherwise, the threads may share the single connection.

ecosystem/nft-metadata-crawler/nft-metadata-crawler-parser/src/parser.rs
```diff
@@ -0,0 +1,289 @@
...
+    gcs::{write_image_to_gcs, write_json_to_gcs},
+    NFTMetadataCrawlerEntry,
+};
+use serde_json::Value;
+use tracing::{error, info};
+
+// Stuct that represents a parser for a single entry from queue
+pub struct Parser<'a> {
+    pub entry: NFTMetadataCrawlerEntry,
+    model: NFTMetadataCrawlerURIs,
```
larry-aptos, line 23: If `Parser` represents a single entity in the queue, can we rename it to something like `ParsingEntity/Unity/Entry` or `ProcessingEntity/Unity/Entry`

ecosystem/nft-metadata-crawler/nft-metadata-crawler-utils/src/gcs.rs
just-in-chang: Using the package that automatically takes the credentials and generates a new time on use. Changed this function ^

ecosystem/nft-metadata-crawler/nft-metadata-crawler-parser/src/main.rs
just-in-chang: Yep this was just a stub. Replaced it

ecosystem/nft-metadata-crawler/nft-metadata-crawler-parser/src/parser.rs
```diff
@@ -0,0 +1,289 @@
...
+                        },
+                        Err(e) => error!(
+                            last_transaction_version = self.entry.last_transaction_version,
+                            "{}",
+                            e.to_string()
+                        ),
+                    }
+                },
+                Err(e) => {
+                    // Do not increment retry count for animation
```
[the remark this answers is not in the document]
just-in-chang: Because for most NFTs we don't expect there to be an animation. On top of that, we directly save the raw animation (don't perform optimizations). Should I still have a counter for that?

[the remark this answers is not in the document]
just-in-chang: Dealing with this next PR

ecosystem/nft-metadata-crawler/nft-metadata-crawler-parser/src/parser.rs
```diff
@@ -0,0 +1,289 @@
...
+                },
+            };
+        }
+    }
+
+    fn handle_uri_parser_jsons(&mut self) -> String {
+        match Self::parse_uri(self.entry.token_uri.clone()) {
+            Ok(u) => u,
+            Err(_) => self.entry.token_uri.clone(),
+        }
```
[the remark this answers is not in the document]
just-in-chang: Will deal with next PR

ecosystem/nft-metadata-crawler/nft-metadata-crawler-parser/migrations/2023-06-27-001532_create_tables/up.sql
```diff
@@ -0,0 +1,15 @@
...
+  token_uri VARCHAR UNIQUE PRIMARY KEY NOT NULL,
+  raw_image_uri VARCHAR,
+  raw_animation_uri VARCHAR,
+  cdn_json_uri VARCHAR,
+  cdn_image_uri VARCHAR,
+  cdn_animation_uri VARCHAR,
+  json_parser_retry_count INT NOT NULL,
+  image_optimizer_retry_count INT NOT NULL,
+  animation_optimizer_retry_count INT NOT NULL,
+  last_updated TIMESTAMP NOT NULL DEFAULT NOW()
```
bowenyang007, line 11: now that I think about it, can we make this inserted_at and add an index on this similar to other tables in indexer? We might want to ingest into bigquery so it'd be nice to follow the same pattern.

ecosystem/nft-metadata-crawler/nft-metadata-crawler-parser/src/lib.rs
```diff
@@ -0,0 +1,45 @@
...
+use diesel::{
+    pg::PgConnection,
+    r2d2::{ConnectionManager, Pool},
+};
+use hyper::header;
+use reqwest::Client;
+
+/**
+ * Establishes a connection pool to Postgres
+ */
```
bowenyang007, line 16: /// ? Idk that's what we do everywhere else so...

ecosystem/nft-metadata-crawler/nft-metadata-crawler-parser/src/models/nft_metadata_crawler_uris_query.rs
```diff
@@ -0,0 +1,64 @@
...
+            .find(token_uri)
+            .first::<NFTMetadataCrawlerURIsQuery>(conn)
+            .optional()
+    }
+
+    pub fn get_by_raw_image_uri(
+        raw_image_uri: Option<String>,
+        conn: &mut PooledConnection<ConnectionManager<PgConnection>>,
+    ) -> diesel::QueryResult<Option<Self>> {
+        if raw_image_uri.is_none() {
```
bowenyang007, line 41: why do you allow passing optional? better to just enforce that it needs to be non null and in the main thread return early or something.

ecosystem/nft-metadata-crawler/nft-metadata-crawler-parser/src/models/nft_metadata_crawler_uris_query.rs
```diff
@@ -0,0 +1,64 @@
...
+    pub cdn_image_uri: Option<String>,
+    pub cdn_animation_uri: Option<String>,
+    pub json_parser_retry_count: i32,
+    pub image_optimizer_retry_count: i32,
+    pub animation_optimizer_retry_count: i32,
+    pub last_updated: chrono::NaiveDateTime,
+}
+
+impl NFTMetadataCrawlerURIsQuery {
+    pub fn get_by_token_uri(
```
bowenyang007, line 27: I dont' see callsites here so are we doing the retry thing? I would just in case there's a temporary connection error or something.

ecosystem/nft-metadata-crawler/nft-metadata-crawler-utils/src/lib.rs
```diff
@@ -0,0 +1,54 @@
...
+    pub token_data_id: String,
+    pub token_uri: String,
+    pub last_transaction_version: i32,
+    pub last_transaction_timestamp: chrono::NaiveDateTime,
+    pub last_updated: chrono::NaiveDateTime,
+    pub force: bool,
+}
+
+impl NFTMetadataCrawlerEntry {
+    pub fn new(s: String) -> anyhow::Result<Self> {
```
bowenyang007, line 26: Hmm it might be helpful to explain what's going on. What is S and why are we doing splits?

ecosystem/nft-metadata-crawler/nft-metadata-crawler-utils/src/pubsub.rs
```diff
@@ -0,0 +1,86 @@
+// Copyright © Aptos Foundation
+
+use google_cloud_googleapis::pubsub::v1::{
+    publisher_client::PublisherClient, subscriber_client::SubscriberClient, AcknowledgeRequest,
+    PublishRequest, PublishResponse, PubsubMessage, PullRequest, PullResponse,
+};
+use tonic::{metadata::MetadataValue, transport::Channel, Request, Response};
+
+#[allow(deprecated)]
```
bowenyang007, line 9: why?

ecosystem/nft-metadata-crawler/nft-metadata-crawler-utils/src/pubsub.rs
```diff
@@ -0,0 +1,86 @@
...
+pub async fn publish_uris(
```
bowenyang007, line 10: Let's add a comment on what this function does. Like instead of just saying "it publishes uris" we should explain in more detail what it does, e.g. what is the message thing and why is it an array.

bowenyang007 (approved):
looks good. left some comments.

ecosystem/nft-metadata-crawler/nft-metadata-crawler-parser/src/models/nft_metadata_crawler_uris_query.rs
just-in-chang: The `raw_image_uri` field is an `Option<String>` in the model, and the JSON parsing function also returns an `Option<String>` for it. Makes things flow better in the main thread IMO I think? so I can just do something like:
```
let raw_image_uri: Option<String>, raw_animation_uri: Option<String>  = parseJSON()

model.set_raw_image_uri(raw_image_uri)
model.set_raw_animation_uri(raw_animation_uri)

if force_flag || get_by_raw_image_uri(raw_image_uri) {
  ... do image optimization stuff
}

if force_flag || get_by_raw_animation_uri(raw_animation_uri) {
  ... do animation optimization stuff
}
```

bowenyang007: But you can just return early before this function right? IMO this is not a great pattern b/c then you have to explain that if it's null it's an no-op.

larry-aptos (approved):

PR #9110, opened 2023-07-16
Python SDK updates

davidiw:
* Kill sync client
* Manually update all packages and move to Python 3.8.1
* As a result, found a lot of new lints and fixed them, yay
* The new version of httpx claims it fixes a performance bug, wonder if that was related to the performance situation in the latency tests
* Updated all examples to use async client
* Improved token client to use events to discover minted tokens
* Add chunked module support proof it works: https://fullnode.testnet.aptoslabs.com/v1/accounts/0xfa3911d7715238b2e3bd5b26b6a35e11ffa16cff318bc11471e84eccee8bd291/modules


movekevin (approved):

ecosystem/python/sdk/CHANGELOG.md
```diff
@@ -2,9 +2,14 @@
 
 All notable changes to the Aptos Python SDK will be captured in this file. This changelog is written by hand for now.
 
-**Note:** Until we cut a 1.0.0 release, the Aptos Python SDK does not follow semantic versioning.
+## 0.7.0
```
0xmaayan, line 5: maybe do a major version? seems like major breaking changes

davidiw: Haha, I was having a conversation with @banool about this. The reason why we remain in version major version zero is so we can be obnoxious like this.

https://semver.org/#spec-item-4

0xmaayan: haha ok, well at some point we would need to officially release a production version, guess we first need to agree that there is an official aptos developed and maintained python sdk.

gregnazario (approved):

ecosystem/python/sdk/CHANGELOG.md
davidiw: Nothing official for now, we should revisit the doc you produced and evaluate what more work needs to be done. As much as I would like to be silly here, I think the two criteria are 1) is it fully fledged according to your doc and 2) our commitment level to its state.

PR #9112, opened 2023-07-17
[framework] fix named addresses

davidiw:
because the frameworks share a common address, one can get away with using the named address incorrectly. this fixes many of those situations. this is a no-op at the bytecode level.


movekevin (approved):

wrwg (approved):
We should add this to the list of things a linter should warn about.

PR #9114, opened 2023-07-17
[Spec] update spec for staking_proxy.move

UIZorrot:
### Description

We finished the basic properties specification of staking_proxy.move

### Test Plan
aptos move prove -f staking_proxy


aptos-move/framework/aptos-framework/sources/staking_proxy.spec.move
```diff
@@ -25,14 +26,54 @@ spec aptos_framework::staking_proxy {
     }
 
     spec set_staking_contract_operator(owner: &signer, old_operator: address, new_operator: address) {
-        use aptos_framework::staking_contract::{Store};
-        use aptos_framework::simple_map;
         // TODO: Verify timeout and can't verify `staking_contract::switch_operator`.
+        pragma verify = true;
```
rahxephon89, line 30: It is not necessary

aptos-move/framework/aptos-framework/sources/staking_proxy.spec.move
```diff
@@ -25,14 +26,54 @@ spec aptos_framework::staking_proxy {
...
         pragma aborts_if_is_partial;
```
rahxephon89, line 31: Could you leave a comment indicating why it is still partial?

UIZorrot: This function related to switch_operator, which contained a vector enum while loop and can't not be verified. Meanwhile it will cause an timeout

aptos-move/framework/aptos-framework/boogie.bpl
rahxephon89, line 1: Please remove this file

rahxephon89 (approved):

PR #9116, opened 2023-07-17
[docs] Add the upcoming aptos book for better tracking

davidiw:
this is definitely a work in progress
* Need to add links to the source code
* Source code needs to be audited to make sure it actually aligns with the chapter / section discussing it
* Objects and digital assets are naked
* Need to kick off advanced topics sections quickly too

The book doesn't need to have all the content, but it would be great to have an entire outline.

If you feel very strongly that this needs to be of high quality before actually pushing, just let me know. We can iterate in the branch `david-book`

### DIFF
--- a/aptos-move/move-examples/fight_the_baddies/basics/Move.toml
+++ b/aptos-move/move-examples/fight_the_baddies/basics/Move.toml
@@ -1,0 +1,11 @@
+[package]
+name = 'FightTheBaddies'
+version = '0.0.1'
+
+[addresses]
+fight_the_baddies = '_'
+
+[dependencies.MoveStdlib]
+git = 'https://github.com/aptos-labs/aptos-core.git'
+rev = 'framework-mainnet'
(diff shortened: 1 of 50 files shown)


xbtmatt:
Not sure if this discussion should go here, but my general thoughts on this are that this is actually structured pretty clearly, with a few things out of place here and there but overall *much* better than the existing content we have. There is a very clear, linear path for learning how to develop on Aptos with this, and it's very useful to have as an opening introduction to point people towards.

The disparate nature of the current docs where lots of things are in random places is much more suitable for reference material and examples, not necessarily learning and understanding.

I wasn't sure how the book would work with the existing docs, but after reading through it again I think this would give us a much clearer distinction to work off of in terms of how to structure existing content, because we could take a strictly technical approach with the existing docs, focusing on goal-oriented examples or reference material.

Overall I think this isn't that far from being content that we could publish, partly because I think a long-form, contextualized tutorial focused on learning is sorely needed, but also because it can be built upon later with more advanced concepts.

PR #9123, opened 2023-07-17
[release] Promote the performance profile to release

perryjrandall:
We have tested the performance profile for a while and are ready to share the benefits

The major change here is enabling LTO (link time optimization, thin variant)

tldr its a whole binary optimization technique that increases compile time but produces significant benefits

https://doc.rust-lang.org/cargo/reference/profiles.html

This has been baking in devnet / testnet for several releases, no issues

Test Plan: builds on PR pass

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -717,37 +717,50 @@ move-vm-test-utils = { path = "third_party/move/move-vm/test-utils", features = 
     "table-extension",
 ] }
 move-vm-types = { path = "third_party/move/move-vm/types" }
+### Cargo profiles
+# See here for reference https://doc.rust-lang.org/cargo/reference/profiles.html
 
+# Release build
+# Major difference here is LTO which give a significant performance gain
 [profile.release]
+# Include debuginfo
 debug = true
+# Panic on overflow
 overflow-checks = true
-
-# The performance build is not currently recommended
-# for production deployments. It has not been widely tested.
-[profile.performance]
-inherits = "release"
-opt-level = 3
-debug = true
-overflow-checks = true
+# Perform thin lto, achieving most of the benefit with less compile time
 lto = "thin"
+# Do not split crates into multiple compiled units
 codegen-units = 1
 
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -645,28 +645,42 @@ move-vm-test-utils = { path = "third_party/move/move-vm/test-utils", features =
...
+# This is ostenisbly the same as release minus lto
+[profile.correctness]
+inherits = "release"
+lto = false
+
+# Performance build
+# Previously the LTO build, duplicate until we remove references to this
 [profile.performance]
 inherits = "release"
-opt-level = 3
```
sherry-x, line 656: missing this one in release profile?

perryjrandall: This is actually unnecessary because it is default in the release profile which we inherit from

https://doc.rust-lang.org/cargo/reference/profiles.html#release

sherry-x (approved):

Cargo.toml
```diff
@@ -645,28 +645,42 @@ move-vm-test-utils = { path = "third_party/move/move-vm/test-utils", features =
...
+# Do not split crates into multiple compiled units
+codegen-units = 1
 
-# The performance build is not currently recommended
-# for production deployments. It has not been widely tested.
+# Correctness build, used for forge / testing
+# This is ostenisbly the same as release minus lto
+[profile.correctness]
+inherits = "release"
+lto = false
```
igor-aptos, line 667: let's remove for now, and add if we see the need?

igor-aptos (approved):

Cargo.toml
perryjrandall: I like the way u roll

movekevin (approved):

Cargo.toml
perryjrandall: I'm going to keep this in case, but we dont need to use it, useful for debugging non lto failures

PR #9124, opened 2023-07-17
[Forge] Tweak HAProxy TPS requirement.

JoshLind:
### Description
This PR tweaks the TPS requirement for the HAProxy test (from `4700` to `4600`). This is because we recently reduced the number of machines for this test and have seen at least one ["unwanted"](https://github.com/aptos-labs/aptos-core/actions/runs/5573747214/jobs/10181487488) failure, i.e.:

```
TPS requirement for inner traffic failed. Average TPS 4654, minimum TPS requirement 4700.
Full stats: committed: 4654 txn/s, submitted: 4656 txn/s, expired: 2 txn/s, latency: 7720 ms,
(p50: 6600 ms, p90: 11100 ms, p99: 26200 ms), latency samples: 2746095
```

### Test Plan
Existing test infrastructure.

### DIFF
--- a/testsuite/forge-cli/src/main.rs
+++ b/testsuite/forge-cli/src/main.rs
@@ -1498,7 +1498,7 @@ fn realistic_env_max_load_test(
                 .init_gas_price_multiplier(20),
             inner_success_criteria: SuccessCriteria::new(
                 if ha_proxy {
-                    4700
+                    4600
                 } else if long_running {
                     5500
                 } else {


perryjrandall (approved):
you say HAProxy I say haha proxy

sherry-x:
Isn't reducing number of machine suppose to make this result better instead of worse?

JoshLind:
> Isn't reducing number of machine suppose to make this result better instead of worse?

In this case, I don't think so -- we went from 10 VFNs to 5 VFNs, which means much less resources to handle incoming transactions (i.e., the transaction submission rate is reduced) 😄 The test failure actually shows this: `submitted: 4656 txn/s` which suggests we're bottlenecked on submission, not consensus, etc. Does that make sense?

perryjrandall:
> > Isn't reducing number of machine suppose to make this result better instead of worse?
> 
> In this case, I don't think so -- we went from 10 VFNs to 5 VFNs, which means much less resources to handle incoming transactions (i.e., the transaction submission rate is reduced) 😄 The test failure actually shows this: `committed: 4654 txn/s, submitted: 4656 txn/s` which suggests we're bottlenecked on submission, not consensus, etc. Does that make sense?

In that case how should we work around that bottleneck, it seems strange if we're measuring throughput to measure a bottlenecked value on a configuration?

JoshLind:
> In that case how should we work around that bottleneck, it seems strange if we're measuring throughput to measure a bottlenecked value on a configuration?

Hmm... I don't think we need to worry too much 😄 IIUC, the purpose of this test is to: (i) have a test that ensures HAProxy works; and (ii) see how HAProxy affects the network. And, in that sense, the test has met its purpose 😄 If someone wants to remove the bottleneck, they can simply test with a larger swarm and/or a better HAProxy configuration. But, until then, we don't lose anything (i.e., we still have several other tests that test for maximum throughput, and larger swarms). Does that make sense?

FWIW, @igor-aptos and I discussed having weekly stable jobs that run all these tests with larger swarms. This is probably something we'll do, so it'll also help to give us additional HAProxy signal.

igor-aptos:
That's not a correct interpretation of stats.

In that mode,  txn emitter sends transactions so that there is a constant number of transactions pending. Faster and more transactions go through, it will submit more.  

This means throughput is reduced.

JoshLind:
> In that mode, txn emitter sends transactions so that there is a constant number of transactions pending. Faster and more transactions go through, it will submit more.

Aah, gotcha. So the client is sending less transactions because it detects that the blockchain has lower throughput? Makes sense to me, thanks! 😄

igor-aptos (approved):

PR #9127, opened 2023-07-17
[Execution][Config] Add the newly added execution config to onchain config registry

sitalkedia:
### Description

This was introduced with transaction shuffling feature, however, we missed to add it to the onchain config registry. 

### Test Plan

Added an smoke test and verified that the shuffling is enabled after the governance proposal.

### DIFF
--- a/consensus/src/transaction_shuffler.rs
+++ b/consensus/src/transaction_shuffler.rs
@@ -6,7 +6,7 @@ use aptos_logger::info;
 use aptos_types::{
     on_chain_config::{
         TransactionShufflerType,
-        TransactionShufflerType::{NoShuffling, SenderAwareV1},
+        TransactionShufflerType::{DeprecatedSenderAwareV1, NoShuffling, SenderAwareV2},
     },
     transaction::SignedTransaction,
 };
@@ -31,7 +31,8 @@ pub fn create_transaction_shuffler(
 ) -> Arc<dyn TransactionShuffler> {
     match shuffler_type {
         NoShuffling => Arc::new(NoOpShuffler {}),
-        SenderAwareV1(confict_window_size) => {
+        DeprecatedSenderAwareV1(_) => Arc::new(NoOpShuffler {}),
+        SenderAwareV2(confict_window_size) => {
             info!(
                 "Using sender aware transaction shuffling with conflict window size {}",
                 confict_window_size
--- a/consensus/src/transaction_shuffler.rs
+++ b/consensus/src/transaction_shuffler.rs
@@ -30,8 +30,14 @@ pub fn create_transaction_shuffler(
     shuffler_type: TransactionShufflerType,
 ) -> Arc<dyn TransactionShuffler> {
     match shuffler_type {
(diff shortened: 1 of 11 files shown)


bchocho (approved):

zekun000 (commented):
there's a few extra things we should do
- add error logs if it doesn't read from on-chain config https://github.com/aptos-labs/aptos-core/blob/main/consensus/src/epoch_manager.rs#L807
- revert this to unwrap_or_default and update the default value https://github.com/aptos-labs/aptos-core/blob/5065e5d65db680d4769141e3216b3c5ef835cad5/consensus/src/epoch_manager.rs#L815
- Add config values properly in genesis instead of using default value https://github.com/aptos-labs/aptos-core/blob/659720530ff2939a37f3c3c799ede6908858c72f/crates/aptos-genesis/src/builder.rs#L613

sitalkedia:
> there's a few extra things we should do
> 
> * add error logs if it doesn't read from on-chain config https://github.com/aptos-labs/aptos-core/blob/main/consensus/src/epoch_manager.rs#L807

- This would have caused spurious error logs when this config was added because the on-chain resource didn't exist but it should be okay to add it now.

zekun000 (approved):

consensus/src/transaction_shuffler.rs
```diff
@@ -30,8 +30,15 @@ pub fn create_transaction_shuffler(
...
-        SenderAwareV1(confict_window_size) => {
+        NoShuffling => {
+            info!("Using no-op transaction shuffling");
+            Arc::new(NoOpShuffler {})
+        },
+        DeprecatedSenderAwareV1(_) => {
+            info!("Using no-op sender aware shuffling v1");
+            Arc::new(NoOpShuffler {})
+        },
+        SenderAwareV2(confict_window_size) => {
```
ibalajiarun, line 41: typo:
```suggestion
        SenderAwareV2(conflict_window_size) => {
```

crates/aptos/src/genesis/mod.rs
```diff
@@ -295,7 +297,9 @@ pub fn fetch_genesis_info(git_options: GitOptions) -> CliTypedResult<GenesisInfo
             employee_vesting_start: layout.employee_vesting_start,
             employee_vesting_period_duration: layout.employee_vesting_period_duration,
             consensus_config: OnChainConsensusConfig::default(),
-            execution_config: OnChainExecutionConfig::default(),
+            execution_config: OnChainExecutionConfig::V1(ExecutionConfigV1 {
+                transaction_shuffler_type: TransactionShufflerType::SenderAwareV2(32),
+            }),
```
ibalajiarun, line 302: Should this be revert this back to being the `default()`? For anyone wanting to test other onchain execution configs in forge, they should be aware to change here.

sitalkedia: @ibalajiarun - Good point. The issue is in Forge, there is not a good way to override various configs.. I will look to see if we can do it without overriding it here but lmk if you have any suggestion.

PR #9129, opened 2023-07-17
[forge tests] put VFNs within emulated networks

bchocho:
### Description

Today, in the land-blocking test, only validators are in emulated networks. This means VFNs:
* Have very low latency connection with validators. This is probably realistic, but we may want to test with some more latency. We can revisit if we need smaller latencies later.
* Have very low latency connection with each other. This means a state sync path from VN1 -> VFN1 -> VFN2 can be (much) faster than VN1 -> VN2 -> VFN2, which likely doesn't happen in practice.

Instead, we now add the VFNs into the emulated networks, making sure they're in the same region as their VN.

In addition, fixed the chunking so a 7 node network will be chunked into 4 networks as (2,2,2,1). Previously the chunking was (4,1,1,1).

### Test Plan

Run forge, observe that the VFN <-> VN RTT is ~100ms, and via logs observe that the nodes are chunked in a (2,2,2,1) configuration.

### DIFF
--- a/testsuite/testcases/src/multi_region_network_test.rs
+++ b/testsuite/testcases/src/multi_region_network_test.rs
@@ -34,31 +34,28 @@ fn get_link_stats_table() -> BTreeMap<String, BTreeMap<String, (u64, f64)>> {
     stats_table
 }
 
-/// Chunks the given set of peers into the number of specified groups
-pub(crate) fn chunk_peers(peers: Vec<Vec<PeerId>>, num_groups: usize) -> Vec<Vec<PeerId>> {
-    // Chunk the peers into exact groups
-    let approx_chunk_size = peers.len() / num_groups;
-    let chunks = peers.chunks_exact(approx_chunk_size);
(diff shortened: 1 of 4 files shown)


igor-aptos (approved):

testsuite/testcases/src/multi_region_network_test.rs
```diff
@@ -229,8 +230,21 @@ impl MultiRegionNetworkEmulationTest {
     /// the fullnodes).
     fn create_netem_chaos(&self, swarm: &mut dyn Swarm) -> SwarmNetEm {
         let all_validators = swarm.validators().map(|v| v.peer_id()).collect::<Vec<_>>();
+        let all_vfns = swarm.full_nodes().map(|v| v.peer_id()).collect::<Vec<_>>();
+        assert!(all_validators.len() >= all_vfns.len());
+
+        let all_pairs: Vec<_> = all_validators
+            .iter()
+            .zip_longest(all_vfns)
```
JoshLind, line 238: Fancy 😄

JoshLind (approved):

PR #9130, opened 2023-07-17
[Examples] Account authentication key rotation with the CLI, Typescript and Python SDKs

xbtmatt:
### Description

Added examples of how to rotate an account's authentication key with an accompanying simple page in the docs that covers how to run the code and the relevant code snippets.

The intention here is to very briefly explain how to do things, not to explain all of the concepts involved in the process. This will be used later in the resource accounts tutorial, for example.

### Test Plan

Run the examples in the docs page by copying and pasting the commands

### DIFF
--- a/developer-docs-site/docs/guides/resource-accounts/key-rotation.md
+++ b/developer-docs-site/docs/guides/resource-accounts/key-rotation.md
@@ -14,9 +14,9 @@ In this guide, we show examples for how to rotate an account's authentication ke
 
 Here are the installation links for the SDKs we will cover in this example:
 
-* [Aptos CLI](../../tools/install-cli/)
+* [Aptos CLI](../../tools/install-cli)
 * [Typescript SDK](../../sdks/ts-sdk/index)
-* [Python SDK](../../sdks/python-sdk/index)
+* [Python SDK](../../sdks/python-sdk)
 
 :::warning
 Some of the following examples use private keys. Do not share your private keys with anyone.
@@ -92,15 +92,15 @@ cd ~/aptos-core/ecosystem/typescript/sdk/examples/typescript-esm
 pnpm install && pnpm rotate_key
 ```
 ```shell title="rotate_key.ts output"
-Account          Address           Auth Key           Private Key     
--------------------------------------------------------------------
-Alice            0x8dcc...7dbe    '0x8dcc...7dbe'    '0x1cec...cc88'
-Bob              0x36eb...9b6c     0x36eb...9b6c      0x9d7c...0610   
- 
+Account            Address             Auth Key             Private Key          Public Key         
+------------------------------------------------------------------------------------------------
+Alice              0x213d...031013    '0x213d...031013'    '0x00a4...b2887b'    '0x859e...08d2a9'
+Bob                0x1c06...ac3bb3     0x1c06...ac3bb3      0xf2be...9486aa      0xbbc1...abb808    
+
 ...rotating...
 
-Alice            0x8dcc...7dbe    '0x36eb...9b6c'    '0x9d7c...0610'
-Bob              0x36eb...9b6c     0x36eb...9b6c      0x9d7c...0610   
+Alice              0x213d...031013    '0x1c06...ac3bb3'    '0xf2be...9486aa'    '0xbbc1...abb808'
+Bob                0x1c06...ac3bb3     0x1c06...ac3bb3      0xf2be...9486aa      0xbbc1...abb808 
 ```
   </TabItem>
   <TabItem value="python" label="Python">
@@ -112,16 +112,16 @@ Bob              0x36eb...9b6c     0x36eb...9b6c      0x9d7c...0610
 cd ~/aptos-core/ecosystem/python/sdk
 poetry install && poetry run python -m examples.rotate-key
 ```
-```shell title="rotate_key.ts output"
-Account            Address             Auth Key             Private Key        
-------------------------------------------------------------------------
-Alice              0x3e58...2b6243    '0x3e58...2b6243'    '0xb927...f4264d'    
-Bob                0xe39d...adf91f     0xe39d...adf91f      0xb2cd...a5c415    
+```shell title="rotate_key.py output"
+Account            Address             Auth Key             Private Key          Public Key         
+------------------------------------------------------------------------------------------------
+Alice              0x213d...031013    '0x213d...031013'    '0x00a4...b2887b'    '0x859e...08d2a9'
+Bob                0x1c06...ac3bb3     0x1c06...ac3bb3      0xf2be...9486aa      0xbbc1...abb808    
 
 ...rotating...
 
-Alice              0x3e58...2b6243    '0xe39d...adf91f'    '0xb2cd...a5c415'    
-Bob                0xe39d...adf91f     0xe39d...adf91f      0xb2cd...a5c415  
+Alice              0x213d...031013    '0x1c06...ac3bb3'    '0xf2be...9486aa'    '0xbbc1...abb808'
+Bob                0x1c06...ac3bb3     0x1c06...ac3bb3      0xf2be...9486aa      0xbbc1...abb808 
 ```
 
   </TabItem>
--- a/developer-docs-site/docs/guides/resource-accounts/key-rotation.md
+++ b/developer-docs-site/docs/guides/resource-accounts/key-rotation.md
@@ -10,14 +10,13 @@ Aptos Move accounts have authentication keys that are separate from their public
 
 After rotating the key, the controlling account can sign transactions for the initial account.
 
-In this guide, we show examples for how to rotate an account's authentication key using the various Aptos SDKs.
+In this guide, we show examples for how to rotate an account's authentication key using a few of the various Aptos SDKs.
 
-If you haven't installed the SDKs, you can do so at the following links:
+Here are the installation links for the SDKs we will cover in this example:
 
 * [Aptos CLI](../../tools/install-cli/)
 * [Typescript SDK](../../sdks/ts-sdk/index)
 * [Python SDK](../../sdks/python-sdk/index)
-* [Rust SDK](../../sdks/rust-sdk/index)
 
 :::warning
 Some of the following examples use private keys. Do not share your private keys with anyone.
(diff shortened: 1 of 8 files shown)


ecosystem/typescript/sdk/examples/typescript-esm/rotate_key.ts
```diff
@@ -0,0 +1,56 @@
...
+        account.authKey(),
+        HexString.fromUint8Array(account.signingKey.secretKey)
+    ];
+
+    return vals.map(v => {
+        return truncate(v).padEnd(WIDTH)
+    }).join(' ');
+}
+
+const WIDTH = 16;
```
0xmaayan, line 22: let's move it to the top of the file

ecosystem/typescript/sdk/examples/typescript-esm/rotate_key.ts
```diff
@@ -0,0 +1,56 @@
+import { AptosAccount, FaucetClient, Network, Provider, HexString } from 'aptos';
```
0xmaayan, line 1: curios why decided to use examples/typescript-esm and not examples/typescript

ecosystem/typescript/sdk/examples/typescript-esm/rotate_key.ts
```diff
@@ -0,0 +1,56 @@
...
+
+export const NODE_URL = "https://fullnode.devnet.aptoslabs.com";
```
0xmaayan, line 3:
```suggestion
const NODE_URL = process.env.APTOS_NODE_URL || "https://fullnode.devnet.aptoslabs.com";
const FAUCET_URL = process.env.APTOS_FAUCET_URL || "https://faucet.devnet.aptoslabs.com";
```

ecosystem/typescript/sdk/examples/typescript-esm/rotate_key.ts
```diff
@@ -0,0 +1,56 @@
...
+
+    return vals.map(v => {
+        return truncate(v).padEnd(WIDTH)
+    }).join(' ');
+}
+
+const WIDTH = 16;
+
+(async() => {
+    const provider = new Provider(Network.DEVNET);
```
0xmaayan, line 25:
```suggestion
    const provider = new Provider({fullnodeUrl: NODE_URL});
```

ecosystem/typescript/sdk/examples/typescript-esm/rotate_key.ts
xbtmatt: I wasn't sure which one to use, since I thought you said we would be slowly moving to esm. But I may have misunderstood. Also I kind of just wanted to organize mine in a different folder since I'm probably going to be adding a bunch of examples over time 😅

should I use the other one?

0xmaayan: ohh no right or wrong decision, was just curios, feel free to use whatever you like. I might revisit this whole examples folder later and think on what to do with it - don't think we need all those different folders.

developer-docs-site/docs/guides/resource-accounts/key-rotation.md
```diff
@@ -0,0 +1,133 @@
...
+---
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+Aptos Move accounts have authentication keys that are separate from their public address. This means you can rotate an account's private key to give control of it to another account without changing the initial account's public address.
+
+After rotating the key, the controlling account can sign transactions for the initial account.
+
+In this guide, we show examples for how to rotate an account's authentication key using the various Aptos SDKs.
```
0xjinn, line 13: This mention various Aptos SDK, are we planning to add it to Rust SDK as well? Maybe we can remove line 20 for now, and follow up with Rust SDK later on when needed :)

developer-docs-site/docs/guides/resource-accounts/key-rotation.md
```diff
@@ -0,0 +1,133 @@
...
+
+```shell title="Initialize two test profiles on devnet"
+echo "devnet" | aptos init --profile rotate-1
+echo "devnet" | aptos init --profile rotate-2
+```
+```shell title="Rotate the authentication key for rotate-1 to rotate-2's authentication key"
+aptos account rotate-key --profile rotate-1 --new-private-key <ROTATE_2_PRIVATE_KEY>
+```
+:::tip Where do I get the private key for a profile?
+Public, private, and authentication keys for Aptos CLI profiles are stored in `~/.aptos/config.yaml`.
```
0xjinn, line 38: I think `~/.aptos/config.yaml` is only correct when config is set to "global". Should we mention this?

ecosystem/python/sdk/examples/rotate-key.py
```diff
@@ -0,0 +1,105 @@
...
+from .common import FAUCET_URL, NODE_URL
+
+WIDTH = 19
+
+def truncate(address: str) -> str:
+    return address[0:6] + '...' + address[-6:]
+
+def format_account_info(account: Account) -> str:
+    vals = [str(account.address()), account.auth_key(), account.private_key.hex()]
+    return ''.join([truncate(v).ljust(WIDTH, ' ') for v in vals])
```
0xjinn, line 30: Why do we want to truncate address, auth_key, and private_key instead of showing the actual values? I might prefer seeing the whole addr, but that probably just me :)

ecosystem/python/sdk/examples/rotate-key.py
```diff
@@ -0,0 +1,105 @@
...
+
+async def rotate_auth_key_ed_25519_payload(
+    rest_client: RestClient,
+    from_account: Account,
+    to_account: Account
+) -> TransactionPayload:
+    rotation_proof_challenge = RotationProofChallenge(
+        sequence_number=await rest_client.account_sequence_number(from_account.address()),
+        originator=from_account.address(),
+        current_auth_key=from_account.address(),
```
0xjinn, line 40: Is there a `auth_key` method or params in Account? Maybe we should use that. 

IIRC I think when the account first created, the `auth_key` and address are the same, but if this is already rotated before, it probably don't match.

developer-docs-site/docs/guides/resource-accounts/key-rotation.md
xbtmatt: Yea, shoot I forgot about this. Are they stored locally in `config.yaml` if you initialize to a directory?

ecosystem/python/sdk/examples/rotate-key.py
xbtmatt: That's just so it's easy to view in the docs, otherwise the strings are too long to all fit on one line to see the changes. Run the docs locally and go to the key rotation page and you'll see what I mean! If it's confusing that way still lmk

developer-docs-site/docs/guides/resource-accounts/key-rotation.md
0xjinn: Yep, it's still in config.yaml, but just root directory will be different.

```
workspace: <current_dir>/.aptos/config.yaml
global: ~/.aptos/config.yaml
```

ecosystem/python/sdk/examples/rotate-key.py
xbtmatt: Good catch!! There is an `auth_key()` method, and it derives it from the private key, so I'll change it to that, yeah!

This is actually code from https://github.com/aptos-labs/aptos-core/blob/main/ecosystem/python/sdk/aptos_sdk/account.py#L144

and

https://github.com/aptos-labs/aptos-core/blob/main/ecosystem/python/sdk/examples/multisig.py#L194

I'll update those in a new PR and tag/attribute you in it!

developer-docs-site/docs/guides/resource-accounts/key-rotation.md
xbtmatt: <current_dir> being wherever you initialized it, right?

ecosystem/typescript/sdk/examples/typescript-esm/rotate_key.ts
```diff
@@ -0,0 +1,58 @@
+import { AptosAccount, FaucetClient, Network, Provider, HexString } from 'aptos';
+
+const NETWORK = Network.DEVNET;
+const FAUCET_URL = `https://faucet.${NETWORK}.aptoslabs.com`;
```
gregnazario, line 4: We should probably at some point make faucet client also use network cc @0xmaayan

ecosystem/typescript/sdk/examples/typescript-esm/rotate_key.ts
```diff
@@ -0,0 +1,58 @@
...
+        account.address(),
+        account.authKey(),
+        HexString.fromUint8Array(account.signingKey.secretKey),
+        HexString.fromUint8Array(account.signingKey.publicKey)
+    ];
+
+    return vals.map(v => {
+        return truncate(v).padEnd(WIDTH)
+    }).join(' ');
+}
```
gregnazario, line 23: We should at some point at least have these be common for examples

ecosystem/python/sdk/examples/rotate-key.py
```diff
@@ -0,0 +1,109 @@
...
+
+    signed_transaction = await rest_client.create_bcs_signed_transaction(
+        alice, payload
+    ) # <:!:rotate_key
+
+    tx_hash = await rest_client.submit_bcs_transaction(signed_transaction)
+    await rest_client.wait_for_transaction(tx_hash)
+
+    new_account_info = await rest_client.account(alice.address())
+    assert AccountAddress.from_hex(new_account_info['authentication_key']) == bob.address(), "Authentication key doesn't match Bob's address"
```
gregnazario, line 101: This is a little confusing here, maybe put that you're rotating Alice to Bob, and that the new authentication key should be the same as the authentication key of Bob's rather than the address.

developer-docs-site/docs/guides/resource-accounts/key-rotation.md
```diff
@@ -0,0 +1,128 @@
...
+Some of the following examples use private keys. Do not share your private keys with anyone.
+:::
+
+## How to rotate an account's authentication key
+<Tabs groupId="examples">
+  <TabItem value="CLI" label="CLI">
+
+```shell title="Initialize two test profiles on devnet"
+echo "devnet" | aptos init --profile rotate-1
+echo "devnet" | aptos init --profile rotate-2
```
gregnazario, line 31: `--network devnet --assume-yes` also works

developer-docs-site/docs/guides/resource-accounts/key-rotation.md
```diff
@@ -0,0 +1,128 @@
...
+```
+```shell title="Rotate the authentication key for rotate-1 to rotate-2's authentication key"
+aptos account rotate-key --profile rotate-1 --new-private-key <ROTATE_2_PRIVATE_KEY>
+```
+:::info Where do I get the private key for a profile?
+Public, private, and authentication keys for Aptos CLI profiles are stored in `~/.aptos/config.yaml` if your config is set to `Global` and `<local_directory>/.aptos/config.yaml` if it's set to `Workspace`.
```
gregnazario, line 37: auth key and address are a little confusing since their interchangeable until they're not

gregnazario (approved):
We may need to explain auth key vs address and when they match and when they don't

0xjinn (approved):

PR #9131, opened 2023-07-17
[ts-sdk] Support null and undefined as absense of value in option

gregnazario:
### Description
The previous implementation supported `undefined`.  It occurred that it may want to also support `null` for an empty option.

### Test Plan
Unit test


ecosystem/typescript/sdk/src/transaction_builder/builder_utils.ts
```diff
@@ -161,7 +161,7 @@ function serializeStruct(argVal: any, argType: TypeTag, serializer: Serializer,
 function serializeOption(argVal: any, argType: TypeTag, serializer: Serializer, depth: number) {
   // For option, we determine if it's empty or not empty first
   // empty option is nothing, we specifically check for undefined to prevent fuzzy matching
-  if (argVal === undefined) {
+  if (argVal === undefined || argVal === null) {
```
0xmaayan, line 164: might be easier
```suggestion
  if (!argVal) {
```

0xmaayan (changes requested):
please remove yarn-error.log and package-lock.json files

ecosystem/typescript/sdk/src/transaction_builder/builder_utils.ts
gregnazario: We have be very careful about things like this, and probably should audit them.

This argval could be `""` or `0`, which are both falsey.

0xmaayan (approved):

ecosystem/typescript/sdk/src/transaction_builder/builder_utils.ts
hardsetting: +1 for being explicit

hardsetting (approved):
LGTM

PR #9132, opened 2023-07-17
[smart_vec] add add_all, to_vector

lightmark:
### Description

`to_vector` basically will be used by view functions to show the vector content atomically. No other better way to do this even including pagination because it could keep changing frequently.

### Test Plan
ut


aptos-move/framework/aptos-stdlib/sources/data_structures/big_vector.move
```diff
@@ -262,6 +264,19 @@ module aptos_std::big_vector {
         exist
     }
 
+    /// Convert the big vector to a native vector.
+    /// Disclaimer: This function is costly. Use it at your own discretion.
```
movekevin, line 268: More details on this would be nice regarding why it's costly. We should also mention this is most useful for view functions, where gas is not charged if called via the fullnode API.

aptos-move/framework/aptos-stdlib/sources/data_structures/smart_vector.move
```diff
@@ -116,6 +116,25 @@ module aptos_std::smart_vector {
         destroy_empty(other);
     }
 
+    /// Add multiple values to the vector at once.
+    public fun add_all<T: store>(v: &mut SmartVector<T>, vals: vector<T>) {
+        vector::for_each(vals, |val| {
+            push_back(v, val);
```
movekevin, line 122: nit: This can fit on previous line:
```
vector::for_each(vals, |val| push_back(v, val));
```

movekevin (approved):

gerben-stavenga (approved):

PR #9133, opened 2023-07-17
Igor reordering test

sitalkedia:

### DIFF
--- a/testsuite/smoke-test/src/aptos_cli/validator.rs
+++ b/testsuite/smoke-test/src/aptos_cli/validator.rs
@@ -353,7 +353,6 @@ async fn test_onchain_config_change() {
 }
 
 #[tokio::test]
-#[ignore]
 // This test is ignored because it is very long running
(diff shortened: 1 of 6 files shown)


testsuite/smoke-test/src/aptos_cli/validator.rs
```diff
@@ -340,6 +352,179 @@ async fn test_onchain_config_change() {
...
+        .submit_batch_bcs(&txns)
+        .await
+        .unwrap()
+        .into_inner();
+    println!("result: {:?}", result);
+    tokio::time::sleep(Duration::from_secs(3)).await;
+
+    wait_for_all_nodes_to_catchup(&clients, Duration::from_secs(30))
+        .await
+        .unwrap();
```
igor-aptos, line 480: can you replace these two with:

for txn in &txns {
    clients[0].1.wait_for_signed_transaction(txn).await.unwrap();
}

PR #9135, opened 2023-07-18
[Release] Pump execution config version

danielxiangzl:
### Description
Pump to execution on chain config version from v2 to v3 to be consistent with the previous release.


bchocho (approved):
The values look good. But I don't know how exactly this fits into the release process

PR #9137, opened 2023-07-18
[forge tests] also colocate vfns and vns for the pfn tests

bchocho:
### Description

For the PFN tests, also colocate VFNs and VNs, and add the PFNs as a single colocated group.

### Test Plan

Ensure the PFNs have the correct index (larger than the VFNs), via forge.

### DIFF
--- a/testsuite/forge-cli/src/main.rs
+++ b/testsuite/forge-cli/src/main.rs
@@ -1891,7 +1891,7 @@ fn pfn_const_tps(
         .with_initial_validator_count(NonZeroUsize::new(7).unwrap())
         .with_initial_fullnode_count(7)
         .with_emit_job(EmitJobRequest::default().mode(EmitJobMode::ConstTps { tps: 100 }))
-        .add_network_test(PFNPerformance::new(add_cpu_chaos, add_network_emulation))
+        .add_network_test(PFNPerformance::new(2, add_cpu_chaos, add_network_emulation))
         .with_genesis_helm_config_fn(Arc::new(|helm_values| {
             // Require frequent epoch changes
             helm_values["chain"]["epoch_duration_secs"] = 300.into();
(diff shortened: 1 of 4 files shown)


testsuite/forge-cli/src/main.rs
```diff
@@ -1899,7 +1899,7 @@ fn pfn_const_tps(
         .with_initial_validator_count(NonZeroUsize::new(7).unwrap())
         .with_initial_fullnode_count(7)
         .with_emit_job(EmitJobRequest::default().mode(EmitJobMode::ConstTps { tps: 100 }))
-        .add_network_test(PFNPerformance::new(add_cpu_chaos, add_network_emulation))
+        .add_network_test(PFNPerformance::new(2, add_cpu_chaos, add_network_emulation))
```
JoshLind, line 1902: Why are we reducing this? Do you require it for more accurate mempool experiments? If so, could we make this configurable? It's helpful on the state sync side because it allows us to test more cases.

JoshLind (commented):
Looks good! Only concern is reducing the number of PFNs for all tests 🤔

JoshLind (approved):

testsuite/testcases/src/public_fullnode_performance.rs
```diff
@@ -80,6 +82,31 @@ impl PFNPerformance {
...
+    /// Gathers and shuffles all peer IDs in the swarm, colocating VFNs with their validator
+    fn gather_and_shuffle_peer_ids_with_colocation(
+        &self,
+        swarm: &mut dyn Swarm,
+    ) -> Vec<Vec<AccountAddress>> {
+        // Identify the validators and fullnodes in the swarm
+        let validator_peer_ids = swarm.validators().map(|v| v.peer_id()).collect::<Vec<_>>();
+        let fullnode_peer_ids = swarm.full_nodes().map(|v| v.peer_id()).collect::<Vec<_>>();
+        let (vfn_peer_ids, pfn_peer_ids) =
+            fullnode_peer_ids.split_at(fullnode_peer_ids.len() - self.num_pfns as usize);
```
ibalajiarun, line 95: Discussed offline with Brian: all the PFNs have index set to 0, so this split will not work.

ibalajiarun (approved):
LGTM

PR #9138, opened 2023-07-18
[smart_table] add add_all, to_simple_map

lightmark:
### Description
as title

to_simple_map would be used for view functions mostly to get a snapshot of this table.
Not supposed to be called for a large table.
### Test Plan
ut

### DIFF
--- a/aptos-move/framework/aptos-stdlib/sources/data_structures/smart_table.move
+++ b/aptos-move/framework/aptos-stdlib/sources/data_structures/smart_table.move
@@ -11,6 +11,8 @@ module aptos_std::smart_table {
     use aptos_std::table_with_length::{Self, TableWithLength};
     use aptos_std::type_info::size_of_val;
     use aptos_std::math64::max;
+    use aptos_std::simple_map::SimpleMap;
+    use aptos_std::simple_map;
 
     /// Key not found in the smart table
     const ENOT_FOUND: u64 = 1;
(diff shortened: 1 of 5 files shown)


aptos-move/framework/aptos-stdlib/sources/data_structures/smart_table.move
```diff
@@ -134,6 +142,44 @@ module aptos_std::smart_table {
         }
     }
 
+    /// Add multiple key/value pairs to the smart table. The keys must not already exist.
+    public fun add_all<K, V>(
+        table: &mut SmartTable<K, V>,
+        keys: vector<K>,
+        values: vector<V>,
+    ) {
+        assert!(vector::length(&keys) == vector::length(&values), error::invalid_argument(EMISMATCHED_LENGTHS));
```
movekevin, line 151: Is this needed? vector::zip has its own check too

aptos-move/framework/aptos-stdlib/sources/data_structures/smart_table.move
```diff
@@ -134,6 +142,44 @@ module aptos_std::smart_table {
...
+        vector::zip(keys, values, |key, value| {
+            add(table, key, value);
```
movekevin, line 153: nit: can be all one line on line 152

aptos-move/framework/aptos-stdlib/sources/data_structures/smart_table.move
```diff
@@ -134,6 +142,44 @@ module aptos_std::smart_table {
...
+        vector::for_each_ref(entries, |e|{
+            let entry: &Entry<K, V> = e;
+            vector::push_back(&mut keys, entry.key);
+            vector::push_back(&mut values, entry.value);
+        });
+        (keys, values)
+    }
+
+    /// Convert a smart table to a simple_map.
+    /// Disclaimer: This function may be costly. Use it at your own discretion.
```
movekevin, line 169: I'd expand this a bit more to explain why it can be costly. We can also mention that this should only be used for view functions ideally, which is the most useful use case

movekevin (approved):

aptos-move/framework/aptos-stdlib/sources/data_structures/smart_table.move
```diff
@@ -26,6 +28,8 @@ module aptos_std::smart_table {
     const EINVALID_TARGET_BUCKET_SIZE: u64 = 6;
     /// Invalid target bucket size.
     const EEXCEED_MAX_BUCKET_SIZE: u64 = 7;
+    /// Lengths of keys and values do not match
+    const EMISMATCHED_LENGTHS: u64 = 8;
```
brmataptos, line 32: You added this here, but don't seem to use it.

lightmark: oh yeah, I deleted the usage. thanks

PR #9139, opened 2023-07-18
[cached-packages] Cached-packages build step skipping

wrwg:
The cached-packages build step generates multiple artifacts from the Move framework sources: a release bundle for genesis, Rust SDKs, and documentation. This allows to make the cached-packages build step optional:

- When env var`SKIP_FRAMEWORK_BUILD` set, this step will be skipped. People may put this into there profile if preferred.
- Build works without this step by checking the only remaining ephemeral artifact (`generated/head.mrb`) into the repo. The other artifacts were already in the repo.
- The script `./scripts/rust_lint.sh` will perform this build step regardless of the env var, and is also run on CI in `--check` mode. This makes it impossible to merge an inconsistent artifact state into the repo.


zekun000 (approved):
update the pr title?

davidiw (approved):

PR #9141, opened 2023-07-18
[dag] add tests for order rule

zekun000:
Add a simple proptest for the order rule

### DIFF
--- a/consensus/src/dag/anchor_election.rs
+++ b/consensus/src/dag/anchor_election.rs
@@ -8,3 +8,21 @@ pub trait AnchorElection {
 
     fn commit(&mut self, round: Round);
 }
+
+pub struct RoundRobinAnchorElection {
+    validators: Vec<Author>,
+}
+
+impl RoundRobinAnchorElection {
+    pub fn new(validators: Vec<Author>) -> Self {
+        Self { validators }
+    }
+}
+
+impl AnchorElection for RoundRobinAnchorElection {
+    fn get_anchor(&self, round: Round) -> Author {
+        self.validators[(round / 2) as usize % self.validators.len()]
+    }
+
+    fn commit(&mut self, _round: Round) {}
+}
--- a/consensus/src/dag/order_rule.rs
+++ b/consensus/src/dag/order_rule.rs
@@ -2,14 +2,14 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use super::dag_store::NodeStatus;
-use crate::{
-    dag::{anchor_election::AnchorElection, dag_store::Dag, types::NodeMetadata, CertifiedNode},
-    experimental::buffer_manager::OrderedBlocks,
+use crate::dag::{
+    anchor_election::AnchorElection, dag_store::Dag, types::NodeMetadata, CertifiedNode,
 };
 use aptos_consensus_types::common::Round;
 use aptos_crypto::HashValue;
 use aptos_infallible::RwLock;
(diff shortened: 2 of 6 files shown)


consensus/src/dag/order_rule.rs
```diff
@@ -105,6 +107,8 @@ impl OrderRule {
         };
         while let Some(prev_anchor) = dag_reader
             .reachable(&current_anchor, Some(self.lowest_unordered_anchor_round))
+            // skip the current anchor itself
+            .skip(1)
```
sasha8, line 111: :)

sasha8 (approved):

consensus/src/dag/anchor_election.rs
```diff
@@ -8,3 +8,21 @@ pub trait AnchorElection {
...
+
+impl RoundRobinAnchorElection {
+    pub fn new(validators: Vec<Author>) -> Self {
+        Self { validators }
+    }
+}
+
+impl AnchorElection for RoundRobinAnchorElection {
+    fn get_anchor(&self, round: Round) -> Author {
+        self.validators[round as usize % self.validators.len()]
```
ibalajiarun, line 24: This is not actually round-robin, because its skipping validators. Works in the good case with Shoal, but not if an anchor doesn't direct commit?

consensus/src/dag/tests/order_rule_tests.rs
```diff
@@ -0,0 +1,142 @@
...
+use aptos_types::{
+    aggregate_signature::AggregateSignature, epoch_state::EpochState,
+    validator_verifier::random_validator_verifier,
+};
+use futures_channel::mpsc::{unbounded, UnboundedReceiver};
+use proptest::prelude::*;
+use std::sync::Arc;
+
+/// Generate a virtual dag that first layer represents round
+/// second layer represents nodes
```
ibalajiarun, line 26: need to fix this comment? I only see two layers.

consensus/src/dag/tests/order_rule_tests.rs
```diff
@@ -0,0 +1,142 @@
...
+        .take(f)
+        .chain(std::iter::repeat(true).take(num_validators - f))
+        .collect();
+    proptest::collection::vec(
+        proptest::collection::vec(Just(bitmask).prop_shuffle(), num_validators),
+        round as usize,
+    )
+}
+
+/// Generate `num_per` random permutations of how nodes are processed by the order rule
```
ibalajiarun, line 43: possible elaborate this comment? IIUC, this basically picks X node indices out of a total Y nodes and these indices are used as the order in which the node is added to order rule, correct?

ibalajiarun (commented):
Pretty neat!

consensus/src/dag/anchor_election.rs
zekun000: you mean round robin with step of 2?

consensus/src/dag/tests/order_rule_tests.rs
zekun000: it's three nested vector :)

consensus/src/dag/anchor_election.rs
ibalajiarun: yes

consensus/src/dag/tests/order_rule_tests.rs
zekun000: it's just a permutation of the order of all Y nodes being added

ibalajiarun (approved):

PR #9142, opened 2023-07-18
[framework] removing this resulted in incompatible framework revisions

davidiw:
Apparently if an application and its dependencies import aptos stdlib with and without this address set, it would break compilation, example:

{
  "Error": "Move compilation failed: Unable to resolve packages for
package 'Tsunami': While resolving dependency 'Switchboard' in package 'Tsunami': Unable to resolve package depen
dency 'Switchboard': While resolving dependency 'AptosStdlib' in package 'Switchboard': Unable to resolve package dependency 'AptosStdlib': Conflicting dependencies found: package '
AptosStdlib' conflicts with 'AptosStdlib'"
}

Adding it back and everything magically works...

### DIFF
--- a/aptos-move/framework/aptos-stdlib/Move.toml
+++ b/aptos-move/framework/aptos-stdlib/Move.toml
@@ -5,6 +5,7 @@ version = "1.0.0"
 [addresses]
 std = "0x1"
 aptos_std = "0x1"
+aptos_framework = "0x1"
 Extensions = "0x1" # For Prover to instantiate `{{Ext}}` in prelude.
 
 [dependencies]


movekevin (approved):

aalok-t:
Unable to reproduce the bug. For instance, consider [`aptos-move/framework/aptos-framework`](https://github.com/aptos-labs/aptos-core/blob/2ea28a76d5761ff5e862edf5018c9712dc809f14/aptos-move/framework/aptos-framework/Move.toml) where the address `aptos_framework = “0x1”` occurs. Removing it would cause a compilation error (as it should) because files like [`aptos-framework/sources/configs/execution_config.move`](https://github.com/aptos-labs/aptos-core/blob/2ea28a76d5761ff5e862edf5018c9712dc809f14/aptos-move/framework/aptos-framework/sources/configs/execution_config.move#L3) require it. A number of files such as [`aptos-move/framework/aptos-token/Move.toml`](https://github.com/aptos-labs/aptos-core/blob/2ea28a76d5761ff5e862edf5018c9712dc809f14/aptos-move/framework/aptos-token/Move.toml) use this address unnecessarily and removing it caused no compilation errors.

PR #9143, opened 2023-07-18
[Network] Add latency metrics for message serialization/deserialization.

JoshLind:
### Description
This PR adds latency metrics for message serialization/deserialization to the network stack. This should help improve observability when handling large message sizes. The PR also does some quick ad-hoc cleanups over the nearby code.

### Test Plan
Existing test infrastructure.

### DIFF
--- a/network/src/counters.rs
+++ b/network/src/counters.rs
@@ -6,7 +6,7 @@ use crate::protocols::wire::handshake::v1::ProtocolId;
 use aptos_config::network_id::NetworkContext;
 use aptos_metrics_core::{
     register_histogram_vec, register_int_counter_vec, register_int_gauge, register_int_gauge_vec,
-    Histogram, HistogramVec, IntCounter, IntCounterVec, IntGauge, IntGaugeVec,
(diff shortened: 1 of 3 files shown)


bchocho (approved):
LGTM

brianolson (approved):

network/src/counters.rs
```diff
@@ -557,3 +561,20 @@ pub fn network_application_outbound_traffic(
         ])
         .observe(size as f64);
 }
+
+/// Time it takes to perform message serialization and deserialization
+pub static NETWORK_APPLICATION_SERIALIZATION_METRIC: Lazy<HistogramVec> = Lazy::new(|| {
+    register_histogram_vec!(
```
grao1991, line 567: The default buckets might not be a good choice here.

JoshLind: Yeah -- I did wonder about that. 😄 From some forge runs, they looked okay. I can send out a separate PR if we feel we need more accuracy. I'm less worried about precision and more about identifying bad behaviours (i.e., if things start to blow up).

PR #9146, opened 2023-07-18
[api] Add consistent API testing

ngkuru:
### Description

This PR introduces consistent API testing where we run a consistent load through the network mimicking various user flows. See [proposal](https://www.notion.so/aptoslabs/API-Flow-Testing-Proposal-ae3424ce6e15433fa69f7c9dd99db037) for more high level information.

Changes:
* Created `aptos-api-tester` crate which runs 4 user flows:
  * Account creation
  * Coin transfer
  * NFT transfer
  * Module publishing, and outputs results to stdout.
* Added `token-client` to the SDK with the methods needed to run [Your First NFT](https://aptos.dev/tutorials/your-first-nft).
  * mirroring the [typescript SDK](https://github.com/aptos-labs/aptos-core/blob/main/ecosystem/typescript/sdk/src/plugins/token_client.ts) and the rust [coin client](https://github.com/aptos-labs/aptos-core/blob/main/sdk/src/coin_client.rs) by using [existing builders](https://github.com/aptos-labs/aptos-core/blob/main/aptos-move/framework/cached-packages/src/aptos_token_sdk_builder.rs)
* Added `PartialEq` to `aptos_rest_client::Account` for testing.
* Added the crate to the `tools` docker image to run with GCP Cloud Run.

Changes to follow (next PR):
* Build push-style metrics logging in the crate.

### Test Plan
The crate is manually run and will be scheduled to run periodically in Cloud Run.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -19,6 +19,7 @@
 !**/*.mrb
 !**/*.errmap
 !config/src/config/test_data
+!aptos-move/
 !aptos-move/aptos-release-builder/data/release.yaml
 !aptos-move/aptos-release-builder/data/proposals/*
 !aptos-move/framework/
--- a/crates/aptos-api-tester/src/main.rs
+++ b/crates/aptos-api-tester/src/main.rs
@@ -254,15 +254,6 @@ async fn probe_getaccountbalanceatversion_1() -> Result<&'static str> {
         return Ok(FAIL_WRONG_BALANCE_AT_VERSION);
     }
 
-<<<<<<< HEAD
-    // use the faucet to ensure that there exists a higher version number
-    faucet_client
-        .fund(giray.address(), 100_000_000)
-        .await
-        .context(ERROR_FAUCET_FUND)?;
-
-=======
->>>>>>> 84b40d670f (Add tests for get_account_balance_at_version)
     // ask for account balance with a higher version number
     let response = client
         .get_account_balance_at_version(giray2.address(), version + 1)
--- a/crates/aptos-api-tester/src/main.rs
+++ b/crates/aptos-api-tester/src/main.rs
@@ -1,0 +1,181 @@
+// Copyright © Aptos Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+#![forbid(unsafe_code)]
+
+use anyhow::{Context, Result};
+use aptos_api_types::U64;
+use aptos_rest_client::{Client, FaucetClient};
+use aptos_sdk::types::LocalAccount;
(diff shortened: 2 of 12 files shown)


crates/aptos-api-tester/src/main.rs
```diff
@@ -0,0 +1,541 @@
...
+    let message = "test message";
+    set_message(client, account, message).await?;
+
+    // check that the message is sent
+    let expected_message = message.to_string();
+    let actual_message = match get_message(client, account.address()).await {
+        Some(message) => message,
+        None => {
+            return Err(TestFailure::Error(anyhow!(
+                "module interaction isn't reflected"
```
gregnazario, line 465: Are these supposed to be generic types?  Do these messages end up in metrics or just logs?  If they're supposed to be typed messages that can be searched later, it may be easier to put them in constants.

crates/aptos-api-tester/src/main.rs
```diff
@@ -0,0 +1,541 @@
...
+    // this test is critical to pass for the next tests
+    if handle_result("new account", test_newaccount(&client, &giray, 100_000_000))
+        .await
+        .is_err()
+    {
+        return Err(anyhow!("returning early because new account test failed"));
+    }
+
+    // Flow 1: Coin transfer
+    let _ = handle_result(
```
gregnazario, line 503: Curious why are all the test results thrown away?

.dockerignore
```diff
@@ -19,6 +19,7 @@
 !**/*.mrb
 !**/*.errmap
 !config/src/config/test_data
+!aptos-move/
```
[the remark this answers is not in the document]
ngkuru: this is for adding the move example to the docker image

crates/aptos-api-tester/src/main.rs
```diff
@@ -0,0 +1,541 @@
...
+    let actual_balance = client
+        .get_account_balance(account.address())
+        .await?
+        .inner()
+        .coin
+        .value;
+
+    if expected_balance != actual_balance {
+        return Err(TestFailure::Fail("wrong balance"));
+    }
```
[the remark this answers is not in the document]
ngkuru: can I add this to the next PR (logger & metrics)? this PR doesn't have the logger and I think it's a better fit

crates/aptos-api-tester/src/main.rs
ngkuru: this is because we need the return value for dependencies so it is returned but most of them aren't used

gregnazario (approved):

PR #9148, opened 2023-07-18
[dag] rpc with fallback implementation

ibalajiarun:
### Description

This PR implements the `send_rpc_with_fallback`.

### DIFF
--- a/consensus/src/dag/dag_fetcher.rs
+++ b/consensus/src/dag/dag_fetcher.rs
@@ -1,7 +1,7 @@
 // Copyright © Aptos Foundation
 // SPDX-License-Identifier: Apache-2.0
 
-use super::{types::NodeMetadata, RpcHandler};
+use super::{dag_network::RpcWithFallback, RpcHandler, types::NodeMetadata};
 use crate::{
     dag::{
         dag_network::DAGNetworkSender,
@@ -14,8 +14,10 @@ use anyhow::ensure;
 use aptos_consensus_types::common::Author;
 use aptos_infallible::RwLock;
 use aptos_logger::error;
+use aptos_time_service::TimeService;
 use aptos_types::epoch_state::EpochState;
-use std::{collections::HashMap, sync::Arc, time::Duration};
+use futures::StreamExt;
+use std::{sync::Arc, time::Duration, collections::HashMap};
 use thiserror::Error as ThisError;
 use tokio::sync::{
     mpsc::{Receiver, Sender},
@@ -61,6 +63,7 @@ struct DagFetcher {
     network: Arc<dyn DAGNetworkSender>,
     dag: Arc<RwLock<Dag>>,
     request_rx: Receiver<LocalFetchRequest>,
+    time_service: TimeService,
 }
 
 impl DagFetcher {
@@ -68,6 +71,7 @@ impl DagFetcher {
         epoch_state: Arc<EpochState>,
         network: Arc<dyn DAGNetworkSender>,
         dag: Arc<RwLock<Dag>>,
+        time_service: TimeService,
     ) -> (Self, Sender<LocalFetchRequest>) {
         let (request_tx, request_rx) = tokio::sync::mpsc::channel(16);
         (
@@ -76,6 +80,7 @@ impl DagFetcher {
                 network,
                 dag,
                 request_rx,
+                time_service,
             },
             request_tx,
         )
@@ -105,36 +110,41 @@ impl DagFetcher {
                     dag_reader.bitmask(local_request.node().round()),
                 )
             };
+
             let network_request = DAGMessage::from(remote_request.clone()).into_network_message();
-            if let Ok(response) = self
-                .network
-                .send_rpc_with_fallbacks(responders, network_request, Duration::from_secs(1))
-                .await
-                .and_then(DAGMessage::try_from)
-                .and_then(FetchResponse::try_from)
-                .and_then(|response| response.verify(&remote_request, &self.epoch_state.verifier))
-            {
-                let ceritified_nodes = response.certified_nodes();
-                // TODO: support chunk response or fallback to state sync
+            let mut rpc = RpcWithFallback::new(
+                responders,
+                network_request,
+                Duration::from_secs(1),
+                self.network.clone(),
+                self.time_service.clone(),
+            );
+            while let Some(response) = rpc.next().await {
+                if let Ok(response) = response
+                    .and_then(DAGMessage::try_from)
+                    .and_then(FetchResponse::try_from)
+                    .and_then(|response| {
+                        response.verify(&remote_request, &self.epoch_state.verifier)
+                    })
                 {
-                    let mut dag_writer = self.dag.write();
-                    for node in ceritified_nodes {
-                        if let Err(e) = dag_writer.add_node(node) {
-                            error!("Failed to add node {}", e);
+                    let certified_nodes = response.certified_nodes();
+                    // TODO: support chunk response or fallback to state sync
+                    {
+                        let mut dag_writer = self.dag.write();
+                        for node in certified_nodes {
+                            if let Err(e) = dag_writer.add_node(node) {
+                                error!("Failed to add node {}", e);
+                            }
                         }
                     }
-                }
-
-                if self
-                    .dag
-                    .read()
-                    .all_exists(local_request.node().parents_metadata())
-                {
-                    local_request.notify();
-                } else {
-                    // TODO: implement retry logic
+                    
+                    if self.dag.read().all_exists(local_request.node().parents_metadata()) {
+                        local_request.notify();
+                        break;
+                    }
                 }
             }
+            // TODO retry
         }
     }
 }
--- a/consensus/src/dag/dag_fetcher.rs
+++ b/consensus/src/dag/dag_fetcher.rs
@@ -111,6 +111,7 @@ impl DagFetcher {
             let mut rpc = RpcWithFallback::new(
                 responders,
                 remote_request.clone().into(),
+                Duration::from_millis(500),
                 Duration::from_secs(1),
                 self.network.clone(),
                 self.time_service.clone(),
--- a/consensus/src/dag/dag_fetcher.rs
+++ b/consensus/src/dag/dag_fetcher.rs
@@ -1,14 +1,11 @@
 // Copyright © Aptos Foundation
 // SPDX-License-Identifier: Apache-2.0
 
-use super::{dag_network::RpcWithFallback, RpcHandler, types::NodeMetadata};
-use crate::{
-    dag::{
-        dag_network::DAGNetworkSender,
-        dag_store::Dag,
-        types::{CertifiedNode, DAGMessage, FetchResponse, Node, RemoteFetchRequest},
-    },
-    network::TConsensusMsg,
+use super::{dag_network::RpcWithFallback, types::NodeMetadata, RpcHandler};
+use crate::dag::{
+    dag_network::DAGNetworkSender,
+    dag_store::Dag,
+    types::{CertifiedNode, FetchResponse, Node, RemoteFetchRequest},
 };
 use anyhow::ensure;
 use aptos_consensus_types::common::Author;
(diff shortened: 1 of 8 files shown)


consensus/src/dag/dag_network.rs
```diff
@@ -28,5 +39,136 @@ pub trait DAGNetworkSender: Send + Sync {
...
+    }
+}
+
+impl Stream for RpcWithFallback {
+    type Item = anyhow::Result<ConsensusMsg>;
+
+    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        let send_rpc = move |sender: Arc<dyn DAGNetworkSender>, peer, message, timeout| async move {
+            sender.send_rpc(peer, message, timeout).await
+        };
```
sasha8, line 116: Why do you need this wrapper?

consensus/src/dag/dag_network.rs
```diff
@@ -28,5 +39,136 @@ pub trait DAGNetworkSender: Send + Sync {
...
+
+        let timeout = matches!(self.interval.as_mut().poll_next(cx), Poll::Ready(_));
+
+        if self.futures.is_empty() || timeout {
```
sasha8, line 120: What if timeout expires but futures has more ready futures that were not pooled yet?

consensus/src/dag/dag_network.rs
```diff
@@ -28,5 +39,136 @@ pub trait DAGNetworkSender: Send + Sync {
...
+        }
+
+        let result = futures::ready!(self.futures.as_mut().poll_next(cx));
+        Poll::Ready(result)
+    }
+}
+
+impl FusedStream for RpcWithFallback {
+    fn is_terminated(&self) -> bool {
+        self.futures.is_empty()
```
sasha8, line 143: I am confused. If it is empty you might want to send_rpc again.
Are not you supposed to check self.terminated?

consensus/src/dag/dag_network.rs
```diff
@@ -28,5 +39,136 @@ pub trait DAGNetworkSender: Send + Sync {
...
+        )
+    }
+}
+
+pub struct RpcWithFallback {
+    responders: Responders,
+    message: ConsensusMsg,
+    timeout: Duration,
+
+    terminated: bool,
```
sasha8, line 81: You are not using this.
Where is the logic for the fetcher to kill the RPC when data is received?

sasha8 (commented):
A few initial comments

consensus/src/dag/dag_network.rs
```diff
@@ -28,5 +39,136 @@ pub trait DAGNetworkSender: Send + Sync {
...
+        Some(
+            self.peers.split_off(
+                self.peers
+                    .len()
+                    .checked_sub(count as usize)
+                    .unwrap_or(0),
+            ),
+        )
+    }
+}
```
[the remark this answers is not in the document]
ibalajiarun: I shuffle the validators first in line 52. do we need to also shuffle each iteration?

consensus/src/dag/dag_network.rs
ibalajiarun: I have a for loop below, so I need to clone each iteration and move the variables into the future. This wrapper makes it readable?

consensus/src/dag/dag_network.rs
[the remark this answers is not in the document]
sasha8: After a few iterations, peers will become an empty vector.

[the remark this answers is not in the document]
ibalajiarun: in that case, the RPC should technically fail, right? and the upper level logic should decide to retry?

I agree I don't return an explicit error (need to check if that is possible with streams)

consensus/src/dag/dag_network.rs
ibalajiarun: are you saying I should poll the future first and then poll the timeout? sure

consensus/src/dag/dag_network.rs
```diff
@@ -28,5 +39,136 @@ pub trait DAGNetworkSender: Send + Sync {
...
+                    let future = Box::pin(send_rpc(
+                        self.sender.clone(),
+                        peer,
+                        self.message.clone(),
+                        self.timeout,
+                    ));
+                    self.futures.push(future);
+                }
+            } else if self.futures.is_empty() {
+                return Poll::Ready(None);
```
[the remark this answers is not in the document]
ibalajiarun: I think at this point the caller (in this case fetcher) should decide to retry and how.

consensus/src/dag/dag_network.rs
ibalajiarun: I think I need to set the terminated first and check here.

consensus/src/dag/dag_network.rs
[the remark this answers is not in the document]
sasha8: What if the timeout was just to aggressive?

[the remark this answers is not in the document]
ibalajiarun: I think the caller should decide that too. I think this matters for fetch in response to a Node request. We can keep retrying but the malicious node is never going to respond, so at some point the caller needs to timeout. This way there is periodic termination and caller can then choose to retry based on the request (certified node or node). 

We can do things like the caller can retry twice for node but 5 times for certified node with increased timeouts.

consensus/src/dag/dag_fetcher.rs
```diff
@@ -105,36 +110,41 @@ impl DagFetcher {
...
+            let mut rpc = RpcWithFallback::new(
+                responders,
+                network_request,
+                Duration::from_secs(1),
+                self.network.clone(),
+                self.time_service.clone(),
+            );
+            while let Some(response) = rpc.next().await {
+                if let Ok(response) = response
+                    .and_then(DAGMessage::try_from)
```
zekun000, line 124: not related but i remember we discussed this can be handled by the trait i.e. make the interface takes DAGMessage instead of ConsensusMessage

consensus/src/dag/dag_network.rs
```diff
@@ -28,5 +39,136 @@ pub trait DAGNetworkSender: Send + Sync {
...
+    }
+}
+
+impl Iterator for ExponentialNumberGenerator {
+    type Item = u32;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        let result = self.current;
+        if self.current < self.max_limit {
+            self.current = (self.current * self.factor).min(self.max_limit)
```
zekun000, line 169: it's purely nit, but * can overflow here

consensus/src/dag/dag_network.rs
```diff
@@ -28,5 +39,136 @@ pub trait DAGNetworkSender: Send + Sync {
...
+            interval: Box::pin(time_service.interval(timeout)),
+        }
+    }
+}
+
+impl Stream for RpcWithFallback {
+    type Item = anyhow::Result<ConsensusMsg>;
+
+    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        let send_rpc = move |sender: Arc<dyn DAGNetworkSender>, peer, message, timeout| async move {
```
zekun000, line 114: I'd probably have a static async function instead of this

consensus/src/dag/dag_network.rs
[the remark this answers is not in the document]
zekun000: I "think" if this fails, we should bail the fetch request and move on. the retry would happen on the sender side, either everyone already move forward then we need to state sync or it blocks someone's reliable broadcast and they'll retry

consensus/src/dag/dag_network.rs
ibalajiarun: I think it was working previously, because poll_next makes sure that `futures`` is always non-empty befor returning. If it `futures` is empty, then it means it is terminated. It won't work now because we check the futures first and return proactively.

consensus/src/dag/dag_network.rs
sasha8: BTW, why do you need this timeout? cannot we rely on the timeout inside the RPC?

consensus/src/dag/dag_network.rs
```diff
@@ -17,16 +28,161 @@ pub trait DAGNetworkSender: Send + Sync {
...
+                for peer in peers {
+                    let future = Box::pin(send_rpc(
+                        self.sender.clone(),
+                        peer,
+                        self.message.clone(),
+                        self.timeout,
+                    ));
+                    self.futures.push(future);
+                }
+            } else if self.futures.is_empty() {
```
sasha8, line 141: Related to the comment above, if we cannot rely on the timeouts in the RPC,  then futures might be pending forever and this called path will never be called.

sasha8 (commented):
Probably minor.

consensus/src/dag/dag_network.rs
```diff
@@ -17,16 +28,161 @@ pub trait DAGNetworkSender: Send + Sync {
...
+    type Item = u32;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        let result = self.current;
+        if self.current < self.max_limit {
+            self.current = self
+                .current
+                .checked_mul(self.factor)
+                .unwrap_or(self.max_limit)
+                .min(self.max_limit)
```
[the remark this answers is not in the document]
ibalajiarun: they both serve different purpose. `min(max_limit)` makes sure the value doesn't exceed the max_limit (say we limit to only enqueuing 4 additional peers max). `checked_mul().unwrap_or()` prevent overflows if someone sets a very big max_limit equal to MAX_INT for example.

consensus/src/dag/dag_network.rs
ibalajiarun: Good point/catch. I meant to split these into two timeouts. We should give a longer timeout for RPC to return and a shorter timeout for queuing more RPCs. The idea is in the best case we should expect a reply within the shorter timeout, but wait till the longer timeout to actually consider the RPC timed out. And, we will use the shorter timeout to queue more requests to improve the chances of getting a reply. I don't remember where but I read this is a way to reduce tail latencies. QS Batch requester does something similar.

consensus/src/dag/dag_network.rs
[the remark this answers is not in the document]
sasha8: I see. But I think you may have a corner case in which you will return self.max_limit even though you overflow with less.

consensus/src/dag/dag_fetcher.rs
```diff
@@ -105,36 +107,45 @@ impl DagFetcher {
...
+                        .dag
+                        .read()
+                        .all_exists(local_request.node().parents_metadata())
+                    {
+                        local_request.notify();
+                        break;
+                    }
                 }
             }
+            // TODO retry
```
zekun000, line 148: we should remove this todo since at this point it's on the other side to retry

consensus/src/dag/dag_network.rs
```diff
@@ -17,16 +28,163 @@ pub trait DAGNetworkSender: Send + Sync {
...
     async fn send_rpc_with_fallbacks(
         &self,
         responders: Vec<Author>,
-        message: ConsensusMsg,
-        timeout: Duration,
-    ) -> anyhow::Result<ConsensusMsg>;
+        message: DAGMessage,
+        retry_interval: Duration,
+        rpc_timeout: Duration,
+    ) -> RpcWithFallback;
```
zekun000, line 43: I don't think we use this in the code?

consensus/src/dag/dag_network.rs
zekun000: yes, we want two separate timeout for the request itself and sending next request

consensus/src/dag/dag_network.rs
```diff
@@ -17,16 +28,163 @@ pub trait DAGNetworkSender: Send + Sync {
...
+
+impl Iterator for ExponentialNumberGenerator {
+    type Item = u32;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        let result = self.current;
+        if self.current < self.max_limit {
+            self.current = self
+                .current
+                .checked_mul(self.factor)
```
zekun000, line 183: lol I was wondering if i should comment on this overflow on previous pass

zekun000 (approved):

sasha8 (approved):

PR #9153, opened 2023-07-18
[Network] Parallelize message serialization tasks.

JoshLind:
### Description
This PR updates the networking code to perform message serialization in parallel. Specifically:

1. We use `for_each_concurrent` to spawn up to a fixed number of threads to perform message serialization in parallel (per application). Now, whenever an application (e.g., consensus or mempool) sends a direct send message (to one or multiple peers), the message is sent to this task, to be serialized and then sent to the peer manager.
    
    The benefit of this is: (i) serialization is moved off the application thread, allowing the application thread to return to other work earlier; (ii) serialization is done using a blocking thread (which makes more sense, because it does not yield and may be expensive with large messages); and (iii) serialization can be performed in parallel, because the serialization task works using multiple threads.

2. We also update RPC requests to use blocking threads to perform serialization and deserialization. Note, RPC requests cannot make use of the new serialization task because they are blocking by nature, and thus applications (e.g., state sync), must already account for this fact (i.e., perform and handle RPC requests on different threads at the application layer).

Note: it's interesting to point out that the recent PR that added support for [parallel message deserialization](https://github.com/aptos-labs/aptos-core/pull/8762) only did so for: (i) incoming direct send messages; and (ii) incoming RPC requests. Deserialization of RPC responses are handled by case (2) above.

This PR offers two commits:
1. Perform message serialization in parallel.
2. Refactor some of the task logic into methods.

### Test Plan
Existing test infrastructure.

### DIFF
--- a/network/src/protocols/network/mod.rs
+++ b/network/src/protocols/network/mod.rs
@@ -8,7 +8,7 @@ pub use crate::protocols::rpc::error::RpcError;
 use crate::{
     error::NetworkError,
     peer_manager::{
-        ConnectionNotification, ConnectionRequestSender, PeerManagerNotification,
+        ConnectionNotification, ConnectionRequestSender, PeerManagerError, PeerManagerNotification,
         PeerManagerRequestSender,
     },
     transport::ConnectionMetadata,
@@ -33,7 +33,7 @@ pub trait Message: DeserializeOwned + Serialize {}
 impl<T: DeserializeOwned + Serialize> Message for T {}
 
 // TODO: do we want to make this configurable?
-const MAX_DESERIALIZATION_QUEUE_SIZE_PER_PEER: usize = 50;
+const MAX_SERIALIZATION_QUEUE_SIZE_PER_APPLICATION: usize = 500;
 
 /// Events received by network clients in a validator
 ///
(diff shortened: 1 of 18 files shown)


network/src/protocols/network/mod.rs
```diff
@@ -311,34 +337,117 @@ impl<TMessage> FusedStream for NetworkEvents<TMessage> {
 /// interface they need.
 ///
 /// Provide Protobuf wrapper over `[peer_manager::PeerManagerRequestSender]`
-#[derive(Clone, Debug)]
+#[derive(Clone)]
 pub struct NetworkSender<TMessage> {
     peer_mgr_reqs_tx: PeerManagerRequestSender,
     connection_reqs_tx: ConnectionRequestSender,
-    _marker: PhantomData<TMessage>,
+    serializing_message_sender: aptos_channel::Sender<(), MessageSerializationRequest<TMessage>>,
```
brianolson, line 344: I'm pretty sure that adding another queue in the sending path isn't what we want. I'd vote more strongly against this now if I weren't planning on replacing it in a couple months.

JoshLind: Hmm... I'm not sure there's a way of parallelizing the sends without introducing another queue (without a huge code refactor). I'm gonna close this for now and we can reopen it if we see the need (e.g., for performance).

PR #9156, opened 2023-07-19
[ts-sdk] Fix parsing Objects with longer generics or structs

gregnazario:
### Description
Previously, anything longer than `0x1::object::Object<T>` e.g. even `0x1::object::Object<T1>` would fail parsing with `Invalid type tag.`. Now, it will just consume anything in the `< >` as it all rolls up to a single address

### Test Plan
See unit tests, I also plan on testing with some sort of wallet integration, but tbd


0xmaayan (approved):
please update changelog so we can get it out

xbtmatt (approved):

PR #9157, opened 2023-07-19
[Faucet] Make it possible to configure separate max amount for requests that bypass checkers

banool:
### Description
This adds support to the faucet for configuring a separate maximum funding amount per request for requests that bypass the eligibility checks. This way certain CI use cases that can already bypass ratelimits don't have to hit the faucet 100 times to get the APT they need.

More context: https://aptos-org.slack.com/archives/C03EP0XM99D/p1689724734273109.

### Test Plan
CI. I also ran the tests locally to double check.

I think the CI isn't properly running the new tests, probably due to the same issue we had with the TS SDK CI. I will look into that first.


perryjrandall (approved):

kent-white (approved):
Tyty

PR #9167, opened 2023-07-19
reduce the runner number in the replay verify

areshand:
### Description
8 runners demands too much resource that can run out of resource. reduce it to 2 for now

### DIFF
--- a/.github/workflows/workflow-run-replay-verify.yaml
+++ b/.github/workflows/workflow-run-replay-verify.yaml
@@ -82,7 +82,7 @@ jobs:
     runs-on: ${{ inputs.RUNS_ON }}
     strategy:
         matrix:
-          number: [0, 1, 2, 3, 4, 5, 6, 7] # runner number
+          number: [0, 1] # runner number
     steps:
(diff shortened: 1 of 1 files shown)


gerben-stavenga (approved):

sionescu (approved):

.github/workflows/workflow-run-replay-verify.yaml
```diff
@@ -82,7 +82,7 @@ jobs:
     runs-on: ${{ inputs.RUNS_ON }}
     strategy:
         matrix:
-          number: [0, 1, 2, 3, 4, 5, 6, 7] # runner number
+          number: [0, 1] # runner number
```
perryjrandall, line 85: Is there a way to make this more configurable?

perryjrandall (approved):

.github/workflows/workflow-run-replay-verify.yaml
areshand: there should be. But I am not an expert on github config and didn't figured out

PR #9168, opened 2023-07-19
[CLI] fix vector args in CLI

0xjinn:
### Description

CLI view function vector arg is broken due to the nested vector PR couple months ago. When you attempt to view a function using vector, it will return the following error

`"Error": "API error: API error Error(InvalidInput): parse arguments[0] failed`

Because you have to implicitly convert u64-256 to string.

This PR fix the vector args problem. 
NOTE: This only fix 1 depth level vector

### Test Plan

```
2023-07-19 10:29:25,967 - INFO - These tests passed:
2023-07-19 10:29:25,967 - INFO - test_metrics_accessible
2023-07-19 10:29:25,967 - INFO - test_init
2023-07-19 10:29:25,967 - INFO - test_config_show_profiles
2023-07-19 10:29:25,967 - INFO - test_account_fund_with_faucet
2023-07-19 10:29:25,967 - INFO - test_account_create
2023-07-19 10:29:25,967 - INFO - test_account_lookup_address
2023-07-19 10:29:25,967 - INFO - test_aptos_header_included
2023-07-19 10:29:25,967 - INFO - test_move_compile
2023-07-19 10:29:25,968 - INFO - test_move_compile_script
2023-07-19 10:29:25,968 - INFO - test_move_publish
2023-07-19 10:29:25,968 - INFO - test_move_run
2023-07-19 10:29:25,968 - INFO - test_move_view
2023-07-19 10:29:25,968 - INFO - All tests passed!
```

### DIFF
--- a/aptos-move/move-examples/cli-e2e-tests/common/sources/cli_e2e_tests.move
+++ b/aptos-move/move-examples/cli-e2e-tests/common/sources/cli_e2e_tests.move
@@ -267,6 +267,21 @@ module addr::cli_e2e_tests {
         (num64, num128, num256)
     }
 
+    #[view]
+    public fun test_vector(
+        input_u8: vector<u8>,
+        input_u16: vector<u16>,
+        input_u32: vector<u32>,
+        input_u64: vector<u64>,
+        input_u128: vector<u128>,
+        input_u256: vector<u256>,
+        input_addr: vector<address>,
+        input_bool: vector<bool>,
+        input_string: vector<String>,
+    ): (vector<u8>, vector<u16>, vector<u32>, vector<u64>, vector<u128>, vector<u256>, vector<address>, vector<bool>, vector<String>) {
+        (input_u8, input_u16, input_u32, input_u64, input_u128, input_u256, input_addr, input_bool, input_string)
+    }
+
     inline fun get_hero(creator: &address, collection: &String, name: &String): (Object<Hero>, &Hero) {
         let token_address = token::create_token_address(
             creator,
--- a/crates/aptos/src/move_tool/mod.rs
+++ b/crates/aptos/src/move_tool/mod.rs
@@ -1531,8 +1531,33 @@ impl ArgWithType {
                 _ => serde_json::to_value(bcs::from_bytes::<T>(&self.arg)?)
                     .map_err(|err| CliError::UnexpectedError(err.to_string())),
             },
-            1 => serde_json::to_value(bcs::from_bytes::<Vec<T>>(&self.arg)?)
-                .map_err(|err| CliError::UnexpectedError(err.to_string())),
+            1 => match self._ty.clone() {
+                FunctionArgType::U64 => {
+                    let u64_vector: Vec<u64> = bcs::from_bytes::<Vec<u64>>(&self.arg)?;
+                    let string_vector: Vec<String> =
+                        u64_vector.iter().map(ToString::to_string).collect();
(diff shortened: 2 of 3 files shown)


aptos-move/move-examples/cli-e2e-tests/common/sources/cli_e2e_tests.move
```diff
@@ -267,6 +267,21 @@ module addr::cli_e2e_tests {
...
+        input_u32: vector<u32>,
+        input_u64: vector<u64>,
+        input_u128: vector<u128>,
+        input_u256: vector<u256>,
+        input_addr: vector<address>,
+        input_bool: vector<bool>,
+        input_string: vector<String>,
+    ): (vector<u8>, vector<u16>, vector<u32>, vector<u64>, vector<u128>, vector<u256>, vector<address>, vector<bool>, vector<String>) {
+        (input_u8, input_u16, input_u32, input_u64, input_u128, input_u256, input_addr, input_bool, input_string)
+    }
```
gregnazario, line 283: What about nested nested vectors?

Basically what I'm seeing right now is that the code that was previously here is basically being added in for single nested vectors

gregnazario (commented):
Approving to unblock this, but let's think about ways we can handle the nesting and all more gracefully

gregnazario (approved):

crates/aptos/src/move_tool/mod.rs
```diff
@@ -1531,8 +1531,33 @@ impl ArgWithType {
...
+                FunctionArgType::U256 => {
+                    let u256_vector: Vec<U256> = bcs::from_bytes::<Vec<U256>>(&self.arg)?;
+                    let string_vector: Vec<String> =
+                        u256_vector.iter().map(ToString::to_string).collect();
+                    serde_json::to_value(string_vector)
+                        .map_err(|err| CliError::UnexpectedError(err.to_string()))
+                },
+                FunctionArgType::Raw => serde_json::to_value(&self.arg)
+                    .map_err(|err| CliError::UnexpectedError(err.to_string())),
+                _ => serde_json::to_value(bcs::from_bytes::<Vec<T>>(&self.arg)?)
```
xbtmatt, line 1558: I can't really understand why the move tests work on `u16` and `u32`, etc when there's no case that matches their type. How come `u64, u128, u256` need to be converted into string vectors first but the others don't?

aptos-move/move-examples/cli-e2e-tests/common/sources/cli_e2e_tests.move
0xjinn: Good question. I was trying to write a recursive func to handle all depth level but failed lol

Since there isn't that much demand on nested vector for u64 - u256. Lets get these into our next release and bandage it. I will follow up with a nested vector fix.

crates/aptos/src/move_tool/mod.rs
xbtmatt: but other than that looks like a good start to me and it works :)

0xjinn: @xbtmatt Yep I was actually wondering the same thing. Quoting the response I got from Greg "Numbers in JSON only are supported up to u32.  The larger ones have to be converted to a string to be supported"

You will also notice that when you run `aptos move view` and the func returns u64/u128/u256, the value returned is actually a `string` as well. e.g `"123"` instead of `123`

xbtmatt (approved):

PR #9172, opened 2023-07-19
[Example] Added the guild/member token

junkil-park:
### Description
Guild and Member token example using Token V2. This module allows the following:
* `admin` adds guild masters to the whitelist
* a whitelisted guild master mints a guild token
* a guild master with a guild token mints a member token for a user
* a guild master burns (or revokes) a member token

For simplicity, this module does not implement the following because these are already shown in the Ambassador example (https://github.com/aptos-labs/aptos-core/blob/main/aptos-move/move-examples/token_objects/ambassador/sources/ambassador.move):
* `admin` mutates a guild token's onchain property (e.g., guild level)
* guild master mutates a member token's onchain property (e.g., member level)

TODO:
* Typescript/indexer example for guild/member tokens

### Test Plan
`aptos move test`

### DIFF
--- a/aptos-move/move-examples/token_objects/guild/Move.toml
+++ b/aptos-move/move-examples/token_objects/guild/Move.toml
@@ -1,0 +1,10 @@
+[package]
+name = 'guild'
+version = '1.0.0'
+
+[addresses]
+guild = "_"
+
+[dependencies]
+AptosFramework = { local = "../../../framework/aptos-framework" }
+AptosTokenObjects = { local = "../../../framework/aptos-token-objects" }
--- a/aptos-move/move-examples/token_objects/guild/sources/guild.move
+++ b/aptos-move/move-examples/token_objects/guild/sources/guild.move
@@ -1,0 +1,320 @@
+/// This module implements the guild and member tokens.
+/// The flow works as follows:
+/// 1. When the admin (i.e., the contract owner) publishes this module,
+///   - a guild collection manager object is created,
+///   - a guild collection is created with the signer of the guild collection manager, and
+///   - a whitelist for guild masters is created.
+/// 2. The admin adds guild masters to the whitelist.
+/// 3. A whitelisted guild master calls `mint_guild` to mint a guild token,
+///    and creates a member collection for the guild.
+/// 4. A guild master with a guild token calls `mint_member` to mint
+///    a member token for a user who is a guild member.
+/// 5. A guild master burns (or revokes) a member token.
+module guild::guild {
+    use std::option;
+    use std::signer;
+    use std::string::{Self, String};
+    use aptos_framework::object::{Self, Object};
+    use aptos_std::smart_vector::{Self, SmartVector};
+    use aptos_token_objects::collection;
+    use aptos_token_objects::token;
+
+    /// The provided signer is not the admin
+    const ENOT_ADMIN: u64 = 1;
+    /// The provided signer is not the owner
+    const ENOT_OWNER: u64 = 2;
+    /// The provided signer is not a guild master
+    const ENOT_GUILD_MASTER: u64 = 3;
+
+    /// The guild token collection name
+    const GUILD_COLLECTION_NAME: vector<u8> = b"Guild Collection Name";
+    /// The guild collection description
+    const GUILD_COLLECTION_DESCRIPTION: vector<u8> = b"Guild Collection Description";
+    /// The guild collection URI
+    const GUILD_COLLECTION_URI: vector<u8> = b"https://guild.collection.uri";
+
+    /// Published under the contract owner's account.
+    struct Config has key {
+        /// Whitelist of guild masters.
+        whitelist: SmartVector<address>,
+        /// `extend_ref` of the guild collection manager object. Used to obtain its signer.
+        extend_ref: object::ExtendRef,
+    }
+
+    #[resource_group_member(group = aptos_framework::object::ObjectGroup)]
+    /// Guild token
+    struct GuildToken has key {
+        /// Used to get the signer of the token
+        extend_ref: object::ExtendRef,
+        /// Member collection name
+        member_collection_name: String,
+    }
+
+    #[resource_group_member(group = aptos_framework::object::ObjectGroup)]
+    /// Member token
+    struct MemberToken has key {
+        /// Belonging guild
+        guild: Object<GuildToken>,
+        /// Used to burn
+        burn_ref: token::BurnRef,
+    }
+
+    /// Initializes the module, creating the manager object, the guild token collection and the whitelist.
+    fun init_module(sender: &signer) acquires Config {
+        // Create the guild collection manager object to use it to autonomously
+        // manage the guild collection (e.g., create the collection and mint tokens).
+        let constructor_ref = object::create_object(signer::address_of(sender));
+        let extend_ref = object::generate_extend_ref(&constructor_ref);
+
+        // Publish the config resource.
+        move_to(sender, Config { whitelist: smart_vector::new(), extend_ref});
+
+        // Create the guild collection.
+        create_guild_collection(&guild_collection_manager_signer());
+    }
+
+    #[view]
+    /// Returns the guild token address by name
+    public fun guild_token_address(guild_token_name: String): address acquires Config {
+        token::create_token_address(&guild_collection_manager_address(), &string::utf8(GUILD_COLLECTION_NAME), &guild_token_name)
+    }
+
+    #[view]
+    /// Returns the guild token address by name
+    public fun member_token_address(guild_token: Object<GuildToken>, member_token_name: String): address acquires GuildToken {
+        let guild_token_addr = object::object_address(&guild_token);
+        let member_collection_name = &borrow_global<GuildToken>(guild_token_addr).member_collection_name;
+        token::create_token_address(&guild_token_addr, member_collection_name, &member_token_name)
+    }
+
+    /// Adds a guild master to the whitelist. This function allows the admin to add a guild master
+    /// to the whitelist.
+    public entry fun whitelist_guild_master(admin: &signer, guild_master: address) acquires Config {
+        assert!(signer::address_of(admin) == guild_collection_manager_owner(), ENOT_ADMIN);
+        let config = borrow_global_mut<Config>(@guild);
+        smart_vector::push_back(&mut config.whitelist, guild_master);
+    }
+
+    /// Mints a guild token, and creates a new associated member collection.
+    /// This function allows a whitelisted guild master to mint a new guild token.
+    public entry fun mint_guild(
+        guild_master: &signer,
+        description: String,
+        name: String,
+        uri: String,
+        member_collection_name: String,
+        member_collection_description: String,
+        member_collection_uri: String,
+    ) acquires Config {
+        // Checks if the guild master is whitelisted.
+        let guild_master_addr = signer::address_of(guild_master);
+        assert!(is_whitelisted(guild_master_addr), ENOT_GUILD_MASTER);
+
+        // The collection name is used to locate the collection object and to create a new token object.
+        let collection = string::utf8(GUILD_COLLECTION_NAME);
+        // Creates the guild token, and get the constructor ref of the token. The constructor ref
+        // is used to generate the refs of the token.
+        // TODO: Switch to `create_token` once it is available.
+        let constructor_ref = token::create_named_token(
+            &guild_collection_manager_signer(),
+            collection,
+            description,
+            name,
+            option::none(),
+            uri,
+        );
+
+        // Generates the object signer and the refs. The refs are used to manage the token.
+        let object_signer = object::generate_signer(&constructor_ref);
+        let extend_ref = object::generate_extend_ref(&constructor_ref);
+
+        // Transfers the token to the guild master.
+        let transfer_ref = object::generate_transfer_ref(&constructor_ref);
+        let linear_transfer_ref = object::generate_linear_transfer_ref(&transfer_ref);
+        object::transfer_with_ref(linear_transfer_ref, guild_master_addr);
+
+
+        // Publishes the GuildToken resource with the refs.
+        let guild_token = GuildToken {
+            extend_ref,
+            member_collection_name,
+        };
+        move_to(&object_signer, guild_token);
+
+        // Creates a member collection which is associated to the guild token.
+        create_member_collection(&object_signer, member_collection_name, member_collection_description, member_collection_uri);
+    }
+
+    /// Mints a member token. This function mints a new member token and transfers it to the
+    /// `receiver` address.
+    public entry fun mint_member(
+        guild_master: &signer,
+        guild_token: Object<GuildToken>,
+        description: String,
+        name: String,
+        uri: String,
+        receiver: address,
+    ) acquires GuildToken {
+        // Checks if the guild master is the owner of the guild token.
+        assert!(object::owner(guild_token) == signer::address_of(guild_master), ENOT_OWNER);
+
+        let guild = borrow_global<GuildToken>(object::object_address(&guild_token));
+        let guild_token_object_signer = object::generate_signer_for_extending(&guild.extend_ref);
+        // Creates the member token, and get the constructor ref of the token. The constructor ref
+        // is used to generate the refs of the token.
+        let constructor_ref = token::create_named_token(
+            &guild_token_object_signer,
+            guild.member_collection_name,
+            description,
+            name,
+            option::none(),
+            uri,
+        );
+
+        // Generates the object signer and the refs. The refs are used to manage the token.
+        let object_signer = object::generate_signer(&constructor_ref);
+        let burn_ref = token::generate_burn_ref(&constructor_ref);
+        let transfer_ref = object::generate_transfer_ref(&constructor_ref);
+
+        // Transfers the token to the `soul_bound_to` address
+        let linear_transfer_ref = object::generate_linear_transfer_ref(&transfer_ref);
+        object::transfer_with_ref(linear_transfer_ref, receiver);
+
+        // Publishes the MemberToken resource with the refs.
+        let member_token = MemberToken {
+            guild: guild_token,
+            burn_ref,
+        };
+        move_to(&object_signer, member_token);
+    }
+
+    /// Burns a member token.
+    public entry fun burn_member(
+        guild_master: &signer,
+        token: Object<MemberToken>,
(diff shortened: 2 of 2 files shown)


aptos-move/move-examples/token_objects/guild/sources/guild.move
```diff
@@ -0,0 +1,309 @@
...
+    /// The guild token collection name
+    const GUILD_COLLECTION_NAME: vector<u8> = b"Guild Collection Name";
+    /// The guild collection description
+    const GUILD_COLLECTION_DESCRIPTION: vector<u8> = b"Guild Collection Description";
+    /// The guild collection URI
+    const GUILD_COLLECTION_URI: vector<u8> = b"https://guild.collection.uri";
+
+    /// Whitelisted guild masters
+    struct Config has key {
+        whitelist: vector<address>,
```
movekevin, line 30: This should use SmartVector instead of vector if the list of whitelisted guild masters can get large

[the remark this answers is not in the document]
movekevin: Usually a centralized list is better for one off whitelist. Creating objects just to store a cap is inefficient storage wise and just costs a lot more gas

aptos-move/move-examples/token_objects/guild/sources/guild.move
```diff
@@ -0,0 +1,309 @@
...
+    use std::option;
+    use std::signer;
+    use std::string::{Self, String};
+    use std::vector;
+    use aptos_framework::object::{Self, Object};
+    use aptos_token_objects::collection;
+    use aptos_token_objects::token;
+
+    /// The token does not exist
+    const ETOKEN_DOES_NOT_EXIST: u64 = 1;
```
movekevin, line 13: Doesn't seem to be used

aptos-move/move-examples/token_objects/guild/sources/guild.move
```diff
@@ -0,0 +1,309 @@
+/// This module implements the guild token (non-fungible token) including the
+/// functions create the collection and the guild tokens.
```
movekevin, line 2: Can you add more details about the flow and what are the different account roles here?

aptos-move/move-examples/token_objects/guild/sources/guild.move
```diff
@@ -0,0 +1,309 @@
...
+        /// Belonging guild
+        guild: Object<GuildToken>,
+        /// Used to burn
+        burn_ref: token::BurnRef,
+    }
+
+    /// Initializes the module, creating the manager object, the guild token collection and the whitelist.
+    fun init_module(sender: &signer) acquires Manager {
+        // Create the manager object to manage the guild collection.
+        create_manager(sender);
```
movekevin, line 60: I'm not quite sure how and why a separate resource is needed for management. It also doesn't seem like the manager does anything really and the extend ref is just used to mint tokens. In that case, why not store the extend ref in the collection object? Check this code out as an example: https://github.com/aptos-labs/aptos-core/blob/main/aptos-move/move-examples/rewards_pool/sources/rewards_pool.move#L48

aptos-move/move-examples/token_objects/guild/sources/guild.move
```diff
@@ -0,0 +1,309 @@
...
+        // Create a collection for guild tokens.
+        create_guild_collection(&manager_signer());
+        // Create a whitelist
+        move_to(sender, Config { whitelist: vector::empty() });
+    }
+
+    #[view]
+    /// Returns the guild token address by name
+    public fun guild_token_address(guild_token_name: String): address acquires Manager {
```
movekevin, line 69: What is the intended use case for this?

aptos-move/move-examples/token_objects/guild/sources/guild.move
```diff
@@ -0,0 +1,309 @@
...
+    public fun member_token_address(guild_token: Object<GuildToken>, member_token_name: String): address acquires GuildToken {
+        let guild_token_addr = object::object_address(&guild_token);
+        let member_collection_name = &borrow_global<GuildToken>(copy guild_token_addr).member_collection;
+        token::create_token_address(&guild_token_addr, member_collection_name, &member_token_name)
+    }
+
+    /// Adds a guild master to the whitelist. This function allows the admin to add a guild master
+    /// to the whitelist.
+    public entry fun add_guild_master(admin: &signer, guild_master: address) acquires Config {
+        assert!(signer::address_of(admin) == @token_objects, ENOT_ADMIN);
```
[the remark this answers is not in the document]
movekevin: +1. very weird that this doesn't use the whitelist

aptos-move/move-examples/token_objects/guild/sources/guild.move
```diff
@@ -0,0 +1,309 @@
...
+    /// Returns the guild token address by name
+    public fun member_token_address(guild_token: Object<GuildToken>, member_token_name: String): address acquires GuildToken {
+        let guild_token_addr = object::object_address(&guild_token);
+        let member_collection_name = &borrow_global<GuildToken>(copy guild_token_addr).member_collection;
+        token::create_token_address(&guild_token_addr, member_collection_name, &member_token_name)
+    }
+
+    /// Adds a guild master to the whitelist. This function allows the admin to add a guild master
+    /// to the whitelist.
+    public entry fun add_guild_master(admin: &signer, guild_master: address) acquires Config {
```
movekevin, line 83: Name is confusing here. This is adding to the whitelist. How about whitelist_guild_master?

aptos-move/move-examples/token_objects/guild/sources/guild.move
```diff
@@ -0,0 +1,309 @@
...
+    ) acquires Config, Manager {
+        // Checks if the guild master is whitelisted.
+        let guild_master_addr = signer::address_of(guild_master);
+        assert!(is_whitelisted(guild_master_addr), ENOT_GUILD_MASTER);
+
+        // The collection name is used to locate the collection object and to create a new token object.
+        let collection = string::utf8(GUILD_COLLECTION_NAME);
+        // Creates the guild token, and get the constructor ref of the token. The constructor ref
+        // is used to generate the refs of the token.
+        let constructor_ref = token::create_named_token(
```
movekevin, line 108: Can you leave a TODO here to switch to create_token (not enabled in mainnet yet) so this can be parallelizable.

aptos-move/move-examples/token_objects/guild/sources/guild.move
```diff
@@ -0,0 +1,309 @@
...
+            option::none(),
+            uri,
+        );
+
+        // Generates the object signer and the refs. The refs are used to manage the token.
+        let object_signer = object::generate_signer(&constructor_ref);
+        let transfer_ref = object::generate_transfer_ref(&constructor_ref);
+        let extend_ref = object::generate_extend_ref(&constructor_ref);
+
+        // Transfers the token to the `soul_bound_to` address
```
movekevin, line 122: This is unnecessary. I would just use the manager_signer directly for transferring

aptos-move/move-examples/token_objects/guild/sources/guild.move
```diff
@@ -0,0 +1,309 @@
...
+        object::transfer_with_ref(linear_transfer_ref, guild_master_addr);
+
+        // Publishes the GuildToken resource with the refs.
+        let guild_token = GuildToken {
+            extend_ref,
+            member_collection: member_collection_name,
+        };
+        move_to(&object_signer, guild_token);
+
+        // Creates a member collection which is associated to the guild token.
```
movekevin, line 133: What are the tradeoffs between this model where guilds are in one collection and members are in another. Vs each guild is one collection?

aptos-move/move-examples/token_objects/guild/Move.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = 'guild'
+version = '1.0.0'
+
+[addresses]
+token_objects = "_"
```
movekevin, line 6: Can you name this guild instead? Let's keep separate names for separate packages.

aptos-move/move-examples/token_objects/guild/sources/guild.move
```diff
@@ -0,0 +1,309 @@
...
+    /// Returns the guild token address by name
+    public fun guild_token_address(guild_token_name: String): address acquires Manager {
+        token::create_token_address(&manager_address(), &string::utf8(GUILD_COLLECTION_NAME), &guild_token_name)
+    }
+
+    #[view]
+    /// Returns the guild token address by name
+    public fun member_token_address(guild_token: Object<GuildToken>, member_token_name: String): address acquires GuildToken {
+        let guild_token_addr = object::object_address(&guild_token);
+        let member_collection_name = &borrow_global<GuildToken>(copy guild_token_addr).member_collection;
```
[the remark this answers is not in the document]
junkil-park: Removed `copy`, which is not wrong here, but not necessary. It already implicitly copies.

aptos-move/move-examples/token_objects/guild/sources/guild.move
```diff
@@ -0,0 +1,309 @@
...
+            description,
+            name,
+            option::none(),
+            uri,
+        );
+    }
+
+    inline fun is_whitelisted(guild_master: address): bool {
+        let whitelist = &borrow_global<Config>(@token_objects).whitelist;
+        vector::contains(whitelist, &guild_master)
```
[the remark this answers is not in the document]
junkil-park: Yes, it can. I chose `vector` for simplicity. For a large list, it's recommended to use `table`. Please see a more realistic whitelist example: https://github.com/aptos-labs/aptos-core/pull/8484/files/d48a9a73ab3f1c2154e8f4711edf507ca2d92ae4..0b54959f4393c51bc0fbe511f84def33d384a445#diff-13c7dc164d67b29d8d16d95acc90548d4850a0f2cffca51f9805e7b0f8f5a033R38

aptos-move/move-examples/token_objects/guild/sources/guild.move
```diff
@@ -0,0 +1,309 @@
...
+
+    /// Adds a guild master to the whitelist. This function allows the admin to add a guild master
+    /// to the whitelist.
+    public entry fun add_guild_master(admin: &signer, guild_master: address) acquires Config {
+        assert!(signer::address_of(admin) == @token_objects, ENOT_ADMIN);
+        let config = borrow_global_mut<Config>(@token_objects);
+        vector::push_back(&mut config.whitelist, guild_master);
+    }
+
+    /// Mints an guild token, and creates a new associated member collection.
```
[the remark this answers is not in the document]
junkil-park: Fixed!

aptos-move/move-examples/token_objects/guild/sources/guild.move
```diff
@@ -0,0 +1,309 @@
...
+        extend_ref: object::ExtendRef,
+    }
+
+    #[resource_group_member(group = aptos_framework::object::ObjectGroup)]
+    /// Guild token
+    struct GuildToken has key {
+        /// Used to get the signer of the token
+        extend_ref: object::ExtendRef,
+        /// Member collection name
+        member_collection: String,
```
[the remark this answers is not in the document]
junkil-park: Done!

aptos-move/move-examples/token_objects/guild/sources/guild.move
```diff
@@ -0,0 +1,309 @@
...
+
+        // Publishes the MemberToken resource with the refs.
+        let member_token = MemberToken {
+            guild: guild_token,
+            burn_ref,
+        };
+        move_to(&object_signer, member_token);
+    }
+
+    /// Burns an member token.
```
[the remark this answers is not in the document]
junkil-park: Fixed!

aptos-move/move-examples/token_objects/guild/sources/guild.move
```diff
@@ -0,0 +1,309 @@
...
+            extend_ref,
+            member_collection: member_collection_name,
+        };
+        move_to(&object_signer, guild_token);
+
+        // Creates a member collection which is associated to the guild token.
+        create_member_collection(&object_signer, member_collection_name, member_collection_description, member_collection_uri);
+    }
+
+    /// Mints an member token. This function mints a new member token and transfers it to the
```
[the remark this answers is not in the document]
junkil-park: Fixed!

aptos-move/move-examples/token_objects/guild/sources/guild.move
junkil-park: Ok. I changed `vector` to `SmartVector` here. When should one use `SmartTable` vs. `SmartVector`?

aptos-move/move-examples/token_objects/guild/sources/guild.move
junkil-park: Done.

aptos-move/move-examples/token_objects/guild/sources/guild.move
junkil-park: Done.

aptos-move/move-examples/token_objects/guild/sources/guild.move
junkil-park: Removed!

aptos-move/move-examples/token_objects/guild/Move.toml
junkil-park: Done.

aptos-move/move-examples/token_objects/guild/sources/guild.move
```diff
@@ -0,0 +1,309 @@
...
+    /// to the whitelist.
+    public entry fun add_guild_master(admin: &signer, guild_master: address) acquires Config {
+        assert!(signer::address_of(admin) == @token_objects, ENOT_ADMIN);
+        let config = borrow_global_mut<Config>(@token_objects);
+        vector::push_back(&mut config.whitelist, guild_master);
+    }
+
+    /// Mints an guild token, and creates a new associated member collection.
+    /// This function allows a whitelisted guild master to mint a new guild token.
+    public entry fun mint_guild(
```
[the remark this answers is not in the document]
junkil-park: > Or can the game just create a guild directly with the admin account?

Yes, we can. And, it would make things much simpler. There would be no need for "whitelist" and "manager object".

aptos-move/move-examples/token_objects/guild/sources/guild.move
junkil-park: This is for a user (guild master) to locate their guild token because `mint_member` requires a guild token as an argument.

aptos-move/move-examples/token_objects/guild/sources/guild.move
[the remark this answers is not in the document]
junkil-park: That's a good point! I skipped it for simplicity and implemented the simplest whitelist in this demo example. For a better reference of the whitelist implementation, one can refer to https://github.com/aptos-labs/aptos-core/pull/8484.

aptos-move/move-examples/token_objects/guild/sources/guild.move
junkil-park: Removed.

aptos-move/move-examples/token_objects/guild/sources/guild.move
junkil-park: Done.

aptos-move/move-examples/token_objects/guild/sources/guild.move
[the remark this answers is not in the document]
junkil-park: Updated to check if the admin is the owner of the manager object instead of the static named address.

aptos-move/move-examples/token_objects/guild/sources/guild.move
junkil-park: In the current model, there is globally one guild collection, and there are many member collections (one member collection per guild).

Alternatively, there could be a different model, where there could be globally one guild collection and one global member collection. @movekevin, is this model what you're referring to?

The former vs. latter comes down to decentralized member collections vs. the centralized member collection. And the tradeoff may be in the parallelizability and fine-grained controllability (e.g., each member collection could be controlled/managed differently).

aptos-move/move-examples/token_objects/guild/sources/guild.move
junkil-park: First of all, let me clarify what the "manager" is for. Here, the manager is a dummy object to use as an autonomous account instead of a resource account. The admin (i.e., contract owner) creates the manager and lets it own the guild collection so that it's possible to mint guild tokens without the admin involved (in other words, without the admin's signature).

The `Manager` resource containing the `extend_ref` of the manager object is published under the module address so that later, the module locates/accesses the `extend_ref` and obtains the signer of the manger object (to use it in minting guild tokens). Alternatively, `extend_ref` could be stored in `Config,` which is also published under the module address. In this way, we could just reduce the number of resources to publish. Moreover, the `extend_ref` of the manager object could also be stored under the guild collection object address. It seems to me that all of these are possible and where to store the `extend_ref` doesn't seem to matter much. A more important question may be: if the pattern to use a dummy object as an autonomous account (signer) instead of a resource account looks ok or not?

aptos-move/move-examples/token_objects/guild/sources/guild.move
movekevin: SmartTable/SmartVector should be used when you expect to be a lot of items for better gas efficiency.

aptos-move/move-examples/token_objects/guild/sources/guild.move
```diff
@@ -0,0 +1,321 @@
...
+
+    /// Initializes the module, creating the manager object, the guild token collection and the whitelist.
+    fun init_module(sender: &signer) acquires Manager {
+        // Create the guild collection manager object to use it to autonomously
+        // manage the guild collection (e.g., create the collection and mint tokens).
+        create_guild_collection_manager(sender);
+        // Create the guild collection
+        create_guild_collection(&guild_collection_manager_signer());
+        // Create a whitelist
+        move_to(sender, Config { whitelist: smart_vector::empty() });
```
movekevin, line 73: Can this be combined with the Manager struct? Both are stored on the deployer (sender)'s account anyway

junkil-park: Updated. `Manager` struct is merged into `Config`.

movekevin (approved):

PR #9173, opened 2023-07-19
[Forge] Increase TPS threshold for land blocking after enabling txn shuffling

sitalkedia:
### Description

With txn shuffling enabled, we are getting 10-15% TPS boost on Forge. All my run with shuffling could achieve more than 6k TPS. So bumping up the threshold to 5.5k for land blocking. Will bump other runs as well once we have some datapoints for them.

### Test Plan
Run Forge

### DIFF
--- a/testsuite/forge-cli/src/main.rs
+++ b/testsuite/forge-cli/src/main.rs
@@ -1496,15 +1496,7 @@ fn realistic_env_max_load_test(
                     mempool_backlog: 40000,
                 })
                 .init_gas_price_multiplier(20),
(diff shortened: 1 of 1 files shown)


igor-aptos:
In a few days, we can check the minimum on humio and increase further if needed

igor-aptos (approved):

bchocho (approved):

sitalkedia:
> In a few days, we can check the minimum on humio and increase further if needed

Do you have the link to monitor this?

PR #9180, opened 2023-07-19
[Prover] Upgraded Boogie and adopted the new datatype syntax

junkil-park:
### Description
Boogie has changed the syntax for datatype since version 2.15.9. This PR makes Prover adopt the new syntax, and upgrade the boogie and z3 versions. ~~This change causes some specs to timeout, which needs to be investigated.~~

This resolves https://github.com/move-language/move/issues/626.

The new syntax is:
#### datatype declaration
old syntax
```
type {:datatype} $Range;
function {:constructor} $Range(lb: int, ub: int): $Range;
```
new syntax
```
datatype $Range {
    $Range(lb: int, ub: int)
}
```

#### field access
old syntax
```
lb#$Range(r)
```
new syntax
```
r->lb
```

#### is operator
old syntax
```
is#$Global(l#$Mutation(m))
```
new syntax
```
(m->l) is $Global
```

### Test Plan
`cargo test -p move-prover`
`cargo test -p aptos-framework`


aptos-move/framework/cached-packages/generated/head.mrb
rahxephon89, line 1: What is this file?

third_party/move/move-prover/boogie-backend/src/options.rs
```diff
@@ -15,7 +15,6 @@ const DEFAULT_BOOGIE_FLAGS: &[&str] = &[
     "-printVerifiedProceduresCount:0",
     "-printModel:1",
     "-enhancedErrorMessages:1",
-    "-monomorphize",
```
rahxephon89, line 18: could you clarify more on the reason to remove this option?

third_party/move/move-prover/boogie-backend/src/options.rs
```diff
@@ -221,7 +220,7 @@ impl BoogieOptions {
             add(&[&format!("-proverOpt:PROVER_PATH={}", &self.z3_exe)]);
         }
         if self.use_array_theory {
-            add(&["-useArrayTheory"]);
+            add(&["-useArrayAxioms"]);
```
rahxephon89, line 223: Is there a reference to introduce these flag changes in boogie?

third_party/move/move-prover/boogie-backend/src/options.rs
junkil-park: First of all, this flag `-monomorphize` does not exist any more, but seem to be merged into the `typeEncoding` option. And, `monomorphic` is by default.
```
/typeEncoding:<t>
                Encoding of types when generating VC of a polymorphic program:
                   m = monomorphic (default)
                   p = predicates
                   a = arguments
                Boogie automatically detects monomorphic programs and enables
                monomorphic VC generation, thereby overriding the above option.
                If the latter two options are used, then arrays are handled via axioms.
```

third_party/move/move-prover/boogie-backend/src/options.rs
junkil-park: https://github.com/boogie-org/boogie/pull/732

third_party/move/move-prover/boogie-backend/src/options.rs
junkil-park: The reference for this change in Boogie: https://github.com/boogie-org/boogie/pull/732

aptos-move/framework/cached-packages/generated/head.mrb
junkil-park: It was generated when I ran `cargo test -p aptos-framework` today. The reference seems to be https://github.com/aptos-labs/aptos-core/pull/9139

wrwg (commented):
Looks great! Its concerning that this upgrade creates so many more timeouts. To really understand the impact on timing behavior, I think we need to have first restored the benchmark stuff in move-prover/lab and run systematic comparison, so we have data like this (as an example): https://github.com/aptos-labs/aptos-core/blob/a1ae60836f987cacb6ed3f731e585df7c684c67e/third_party/move/move-prover/lab/data/new-boogie/fun_by_fun.svg

davidiw (commented):
you introduced a few todos, should we resolve those first or move forward with this as is?
I will say the syntax is a lot nicer to read.

wrwg:
> you introduced a few todos, should we resolve those first or move forward with this as is? I will say the syntax is a lot nicer to read.

I have asked to postpone landing this PR until we better understand the performance impact. We skipped a few versions of Boogie, and this upgrade creates new unexpected timeouts. For that we are re-activating an older benchmarking infrastructure (`move-prover/lab`) for benchmarking. I believe Junkil is currently working on this.

junkil-park:
Here is the PR to revive the Prover lab: https://github.com/aptos-labs/aptos-core/pull/9518

third_party/move/move-prover/boogie-backend/src/options.rs
rahxephon89: @junkil-park As discussed in https://github.com/boogie-org/boogie/discussions/769, `-useArrayAxioms` should be set by default apart from removing `-useArrayTheory`. And thus the option `use_array_theory` may also be removed. cc @wrwg

wrwg: Yes, the naming is confusing. 'useArrayTheory' == SMT arrays, 'useArrayAxioms' boogie arrays, which is our default.

wrwg:
Curious: with passing the right `userArrayAxioms` option, do we still see a regression in the performance?

rahxephon89:
> Curious: with passing the right `userArrayAxioms` option, do we still see a regression in the performance?

For regular checking, the regression seems to be negligible. However, there is a regression when doing inconsistency checking.

junkil-park:
> Curious: with passing the right `userArrayAxioms` option, do we still see a regression in the performance?

@wrwg , previously, there were many timeouts without using the right option after upgrading Boogie. However, with the right option (`useArrayAxioms`), all the timeouts are gone!

As shown in the lab result (with `diem-framework`), there is a slight regression, but it looks better than before. I am checking the regression with the `aptos-framework`.

Like @rahxephon89 said, the new boogie doesn't pass the inconsistency check (running Prover with unlimited time bound). Prover hangs. The reason is not clear yet. Teng is investigating this.

wrwg (approved):
Great that at least the performance issue with regular execution is solved! For the problem with consistency checks we can iterate, so lets land this.

third_party/move/move-prover/boogie-backend/src/options.rs
rahxephon89, line 1: @junkil-park We need to set `num_instances` to 2 to pass the CI test for regular checking.

junkil-park: Done. Thanks!

rahxephon89 (approved):

PR #9181, opened 2023-07-19
Enable unit testing of events in Move

davidiw:
Often times developer of smart contracts have a need to evaluate their events are correctly emitting. This becomes more obvious as some developers build utilities that are entirely embedded within Move and make assumptions about the runtimes behaviors, such that what is expressed internally is the same as externally.

As a result, event testing can be very complicated as events are currently unreadable in Move.

This adds a test-only method for obtaining the data emitted by an event.
* Events can be queried by their handle and the sequence number for the event.
* The public move function returns an option where it is none if no event exists in memory and some if it does.
* Only events emitted in the current session can be queried.

Some design decisions:
* Querying the event and determining if it exist ideally would be a single function that returns an Option<T>. Unfortunately this is rather complicated to implement and so it was decided to split these into two functions.
* For testing, events could have been stored within the adapter layer for efficient querying without adding new code to core Move. Instead the core VMs event cache can be queried directly resulting in an additional NativeContext function.
* This is only available for testing, but there's no (clear) reason why it couldn't be enabled in production, but also no clear application for doing so.


aptos-move/framework/src/natives/event.rs
```diff
@@ -83,15 +130,30 @@ pub fn make_all(
...
+                (),
+                timed_features.clone(),
+                features.clone(),
+                native_emitted_internal,
+            ),
+        ),
+        (
+            "emitted_event_internal",
+            make_safe_native((), timed_features, features, native_emitted_event_internal),
         ),
```
vgao1996, line 155: Weird. Why has the [check-vm-features](https://github.com/aptos-labs/aptos-core/actions/runs/5605042252/jobs/10253708618?pr=9181#logs) test succeed even though you are not feature gating these natives here...

alnoki:
@davidiw First of all this is a substantial unlock for developers, thank you for looking into this

If an event is successfully looked up inside a test context (e.g. `event::emitted_event()` returns `some`), will running the same lookup call return the same event again, or will it return `none`?

I advocate for returning the same event each time, rather than "consuming" the event in the first call

aptos-move/framework/aptos-framework/sources/event.move
```diff
@@ -54,4 +54,23 @@ module aptos_framework::event {
     public fun destroy_handle<T: drop + store>(handle: EventHandle<T>) {
         EventHandle<T> { counter: _, guid: _ } = handle;
     }
+
+    #[test_only]
+    use std::option::{Self, Option};
+
+    #[test_only]
+    public fun emitted_event<T: drop + store>(handle: &EventHandle<T>, sequence_number: u64): Option<T> {
```
alnoki, line 62: Cant this be named something more descriptive like `get_emitted_event()` ?

aptos-move/framework/aptos-framework/sources/event.move
```diff
@@ -54,4 +54,23 @@ module aptos_framework::event {
...
+        } else {
+            option::none()
+        }
+    }
+
+    #[test_only]
+    native fun emitted_internal<T: drop + store>(id: vector<u8>, sequence_number: u64): bool;
+
+    #[test_only]
+    native fun emitted_event_internal<T: drop + store>(id: vector<u8>, sequence_number: u64): T;
```
vgao1996, line 75: > Unfortunately this is rather complicated to implement and so it was decided to split these into two functions.

Okay, it's a bit complicated, but not that hard. Here's how you return an `Option<T>` in a native function.
```
Value::struct_(Struct::pack(Value::vector_for_testing_only([event_value])))
```
Totally understand the frustration though. Hopefully one day I'll have time to implement some user-friendly value APIs for the values...

aptos-move/framework/aptos-framework/sources/event.move
```diff
@@ -54,4 +54,23 @@ module aptos_framework::event {
...
+    #[test_only]
+    public fun emitted_event<T: drop + store>(handle: &EventHandle<T>, sequence_number: u64): Option<T> {
+        let id = bcs::to_bytes(&handle.guid);
+        if (emitted_internal<T>(id, sequence_number)) {
+            option::some(emitted_event_internal<T>(id, sequence_number))
+        } else {
+            option::none()
+        }
+    }
+
```
alnoki, line 70: Can this also include a function like `get_emitted_event_unchecked()` that assumes some, and either unwraps it or asserts it is some?

aptos-move/framework/aptos-framework/sources/event.move
```diff
@@ -54,4 +54,23 @@ module aptos_framework::event {
...
+    #[test_only]
+    native fun emitted_internal<T: drop + store>(id: vector<u8>, sequence_number: u64): bool;
```
alnoki, line 72: How about a public test-only wrapper around this `was_event_emitted()`?

aptos-move/framework/aptos-framework/sources/event.move
davidiw: oooh I was so close... I had only `Value::vector...`

vgao1996:
I'm also wondering if we should have a way to return a list of all events

davidiw:
> @davidiw First of all this is a substantial unlock for developers, thank you for looking into this
> 
> If an event is successfully looked up inside a test context (e.g. `event::emitted_event()` returns `some`), will running the same lookup call return the same event again, or will it return `none`?
> 
> I advocate for returning the same event each time, rather than "consuming" the event in the first call

You can call it multiple times. If you look, I bumped the sequence number up to the next event, so it should be none. Maybe I can add a comment to the test.

aptos-move/framework/aptos-framework/sources/event.move
davidiw: you mean so you don't have to write the check and assert yourself?

davidiw:
> I'm also wondering if we should have a way to return a list of all events

I think that might be useful for debugging, but I wanted to leave some impact opportunity for the future generation.

aptos-move/framework/aptos-framework/sources/account.move
```diff
@@ -1309,4 +1309,20 @@ module aptos_framework::account {
...
+    fun test_events(account: &signer) acquires Account {
+        let addr = signer::address_of(account);
+        create_account_unchecked(addr);
+        register_coin<FakeCoin>(addr);
+
+        let eventhandle = &borrow_global<Account>(addr).coin_register_events;
+        let counter = event::counter(eventhandle);
+        let event = CoinRegisterEvent { type_info: type_info::type_of<FakeCoin>() };
+        assert!(event::emitted_event(eventhandle, counter - 1) == option::some(event), 0);
+        assert!(event::emitted_event(eventhandle, counter) == option::none(), 0);
```
alnoki, line 1326: Can you add in a test here that asserts the same return from the getter when it is called twice?

aptos-move/framework/aptos-framework/sources/event.move
vgao1996: I'm not surprised. `vector_for_testing_only` is the only API that allows you to create a vector of complex values. I marked it as "for_testing_only" because it's not checking the layouts of the values, making it possible for one to create invalid values.

I think for now it's okay to use it, considering the natives you are adding are only run locally. However we really should come up with a proper API that offers this functionality. In fact, let me see if I can come up with one.

aptos-move/framework/aptos-framework/sources/event.move
alnoki: Sure, the same functionality could be achieved within each unit context by checking the result of `event::emitted_event()` as some/none,  but I figure that would be a common function call that could be abstracted into a helper

Not super necessary, just requiring less function calls for test writers

aptos-move/framework/aptos-framework/sources/account.move
alnoki: e.g. 

```move
        let event = CoinRegisterEvent { type_info: type_info::type_of<FakeCoin>() };
        assert!(event::emitted_event(eventhandle, counter - 1) == option::some(event), 0);
        let event_2 = CoinRegisterEvent { type_info: type_info::type_of<FakeCoin>() };
        assert!(event::emitted_event(eventhandle, counter - 1) == option::some(event_2), 0);
```

aptos-move/aptos-vm/src/natives.rs
```diff
@@ -76,7 +76,8 @@ pub fn assert_no_test_natives(err_msg: &str) {
                     && func_name.as_str() == "generate_keys_internal"
                 || module_name.as_str() == "bls12381" && func_name.as_str() == "sign_internal"
                 || module_name.as_str() == "bls12381"
-                    && func_name.as_str() == "generate_proof_of_possession_internal")
+                    && func_name.as_str() == "generate_proof_of_possession_internal"
+                || module_name.as_str() == "event" && func_name.as_str() == "emitted")
```
vgao1996, line 80: Oh it's because you have the wrong name here. Mind changing `== "emitted"` to `.contains("emitted")`?

aptos-move/framework/aptos-framework/sources/event.move
vgao1996: Okay, that layout check can be a bit hard to write. Let's just use `vector_for_testing_only` for now.

aptos-move/framework/src/natives/event.rs
vgao1996: @davidiw once you fix the check, you'll need feature gate these natives here via `#[cfg(feature = "testing")]`.

aptos-move/framework/aptos-framework/sources/account.move
```diff
@@ -1309,4 +1309,20 @@ module aptos_framework::account {
...
+    struct FakeCoin { }
+
+    #[test(account = @0x1234)]
+    fun test_events(account: &signer) acquires Account {
+        let addr = signer::address_of(account);
+        create_account_unchecked(addr);
+        register_coin<FakeCoin>(addr);
+
+        let eventhandle = &borrow_global<Account>(addr).coin_register_events;
+        let counter = event::counter(eventhandle);
```
wrwg, line 1323: The need for the counter appears unfortunate to me. You would never ever have to remember it apart of for this case. Also if the event is emitted somewhere deep inside some call chain, how to even get the counter?

Why not return a vector of all events? Then you can write something like:

```
assert!(contains(event:emitted_events(handle), |e| ...e...))
```

Checking for the counter becomes a special case.

wrwg: > The need for the counter appears unfortunate to me. You would never ever have to remember it apart of for this case. Also if the event is emitted somewhere deep inside some call chain, how to even get the counter?
> 
> Why not return a vector of all events? Then you can write something like:
> 
> ```
> assert!(contains(event:emitted_events(handle), |e| ...e...))
> ```
> 
> Checking for the counter becomes a special case.

Another problem: both with your and mine suggestion event needs to be `copy` which usually it isn't. Damn it we do not have `Option<&T>` or `vector<&T>`. If we would have fulll lambdas this could be mitigated: `event:emitted(|e: &E| ...)`

davidiw: These are intended for unit tests. So having explicit knowledge seems relevant. That's also why I didn't expose the native function, that way we can always change it to be more expressive as you suggest, add more accessors for all events, and get an easy bump.

Regardless of where we land here, I think we'll want to have the access by sequence number.

wrwg: I don't understand why this explicit knowledge of the sequence number would be relevant. For events v2 were are in the process of removing those, so why would anybody care?

Also, how can you know the sequence number if the event is emitted somewhere deep inside the logic you are testing? Lets say your test function calls A calls B calls C which emits the event. How would you know the number? The code you are testing doesn't remember the number as it doesn't care about it. How to get it up to the test?

The issue have having events emitted somewhere quite indirectly deep inside code is well-known from the time we implemented event verification for the Move prover.

Last not least, testing based on sequence number is not compostional. Suppose I test a function which calls A and then B. B emits an event and I test it. A today does not emit an event. But now I change the code so A also does emit an even. Now your test fails because the sequence number shifted. Not good.

davidiw: I've added a new core function that returns all (seq_num, events)

aptos-move/framework/aptos-framework/sources/event.move
davidiw: I think the desire for "get_" prefix is a personal choice.

aptos-move/framework/aptos-framework/sources/event.move
davidiw: Inclined to leave the current set and go from there?

aptos-move/framework/src/natives/event.rs
davidiw: addressed

aptos-move/framework/aptos-framework/sources/event.move
davidiw: addressed

aptos-move/framework/aptos-framework/sources/event.move
davidiw: addressed

aptos-move/framework/aptos-framework/sources/account.move
davidiw: addressed

aptos-move/aptos-vm/src/natives.rs
davidiw: addressed

davidiw:
@wrwg makes a good point. We have intention to eliminate sequence numbers. If we start training people to think about sequence numbers here, we're creating unnecessary challenges for us and them later. Removed the sequence number APIs.

wrwg:
> @wrwg makes a good point. We have intention to eliminate sequence numbers. If we start training people to think about sequence numbers here, we're creating unnecessary challenges for us and them later. Removed the sequence number APIs.

Great!

For further motivation, I expect we more often find tests as below than tests which deal with sequence numbers:

```
#[test]
fun test_withdraw() {
  let test_sender = new_sender();
  system_under_test::withdraw(sender, 5267);
  <<verify state>>
  assert!(contains(event::emitted<WithdrawEvent>(..), |e| e.sender == test_sender && e.amount == 5267);
}
```

We have seen many of those kind of invariants when specifying and verifying events with the Move Prover for Diem. Based on this experience, the `contains` test is the right (and perhaps the only) way how those events should be accessed from tests. We encoded this model also as the semantics of the Move prover with its `emits` specification clause.

aptos-move/framework/aptos-framework/sources/event.move
```diff
@@ -54,4 +54,18 @@ module aptos_framework::event {
...
         EventHandle<T> { counter: _, guid: _ } = handle;
     }
+
+    #[test_only]
+    public fun emitted_events<T: drop + store>(handle: &EventHandle<T>): vector<T> {
+        emitted_events_internal(bcs::to_bytes(&handle.guid))
+    }
+
+    #[test_only]
+    public fun was_event_emitted<T: drop + store>(handle: &EventHandle<T>, msg: &T): bool {
```
lightmark, line 64: assert_emission?

third_party/move/move-vm/runtime/src/data_cache.rs
```diff
@@ -295,4 +295,18 @@ impl<'r> TransactionDataCache<'r> {
         let ty_layout = loader.type_to_type_layout(&ty)?;
         Ok(self.event_data.push((guid, seq_num, ty, ty_layout, val)))
     }
+
+    pub(crate) fn emitted_events(
+        &self,
+        guid: Vec<u8>,
+        ty: Type,
```
lightmark, line 302: Is this necessary?

movekevin (approved):

aptos-move/framework/aptos-framework/sources/event.move
davidiw: assert is typically tied to an `assert` statement that explodes if false. here we want the user space to be able to dictate the behavior if the event exists or not.

aptos-move/framework/aptos-framework/sources/event.move
davidiw: This doesn't fit in with the model we landed on.

aptos-move/framework/aptos-framework/sources/event.move
davidiw: This doesn't fit in with the model we landed on.

movekevin:
> > > @wrwg makes a good point. We have intention to eliminate sequence numbers. If we start training people to think about sequence numbers here, we're creating unnecessary challenges for us and them later. Removed the sequence number APIs.
> > 
> > 
> > Great!
> > For further motivation, I expect we more often find tests as below than tests which deal with sequence numbers:
> > ```
> > #[test]
> > fun test_withdraw() {
> >   let test_sender = new_sender();
> >   system_under_test::withdraw(sender, 5267);
> >   <<verify state>>
> >   assert!(contains(event::emitted<WithdrawEvent>(..), |e| e.sender == test_sender && e.amount == 5267);
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > We have seen many of those kind of invariants when specifying and verifying events with the Move Prover for Diem. Based on this experience, the `contains` test is the right (and perhaps the only) way how those events should be accessed from tests. We encoded this model also as the semantics of the Move prover with its `emits` specification clause.
> 
> We'll still need to evaluate how one would query for events. For example, in this model, one needs access to the event handle, so it likely is that a test-only accessor would need to be added to the module that emitted the events.
> 
> I think the same case will be true for next gen events, but it is better in that it only needs access to the struct, whereas, in this model, it needs access to the eventhandle which could be stored non-deterministically in global storage.

Everything would be so much easier with event v2

third_party/move/move-vm/runtime/src/data_cache.rs
lightmark: could we? I feel it is impossible since guid matches, ty must match too.

lightmark (approved):

third_party/move/move-vm/runtime/src/data_cache.rs
```diff
@@ -295,4 +295,14 @@ impl<'r> TransactionDataCache<'r> {
...
+
+    pub(crate) fn emitted_events(&self, guid: Vec<u8>, ty: Type) -> PartialVMResult<Vec<Value>> {
+        let mut events = vec![];
+        for event in self.event_data.iter() {
+            if event.0 == guid && event.2 == ty {
+                events.push(event.4.copy_value()?);
+            }
+        }
+        Ok(events)
+    }
```
vgao1996, line 307: No need to do it in this PR, but we really should move the event buffer out of the Move VM and re implement it as a native extension. Events being part of the Move VM is a legacy of the past we haven't got the time to clean up.

lightmark: I don't mind if DW can finish that in this PR 💯

third_party/move/move-vm/runtime/src/data_cache.rs
davidiw: actually let me push back again:
* The Move contract API makes this impossible, since it expects an event handle and therefore we have some bizarre invariant if we hit this
* The Move core code API has no expectations that the types are consistent and so we create a new application specific behavior by adding this check. I could rewrite the code quite a bit to be more advanced, but I'm not inclined to add the complexity

alnoki:
@davidiw seeing as this is test-only functionality is it possible to get this included in the `mainnet` branch for inclusion in downstream Move package dependencies?

davidiw:
> @davidiw seeing as this is test-only functionality is it possible to get this included in the `mainnet` branch for inclusion in downstream Move package dependencies?

Unlikely due to effort -- as it requires some gnarly changes to many pieces.

PR #9183, opened 2023-07-20
Make Display impl of AccountAddress conform to AIP-40

banool:
### Stack
- Previous in stack: https://github.com/aptos-labs/aptos-core/pull/9186
- Next in stack: N/A

### Description
This PR makes the Display impl of AccountAddress use to_standard_string and updates a few relevant locations in response to this new impl. The motivation behind this change is to make it that when people use AccountAddress, they get standard compliant behavior by default.

As it turns out, there are many places throughout the codebase where people implement their own flavor of the standard, e.g. `format!("0x{}", addr.short_str_lossless())`. As such this PR updates a lot of those locations to just `format!("{}", addr)`, which does almost the same thing (only difference is it won't format non special addresses in the SHORT format).

The impacts of this change:
- The representation of addresses in all structs using the Display impl of account addresses has changed. This affects how a lot of Move VM structs (for example) are represented in debug mode / when serialized. You can see the changes by looking at the updated goldens for those tests. Account addresses in generated scripts are different too.
- The address in the config generated by `aptos init` has the address in a standard-compliant format.
- Accordingly, all of those sites have been updated to use `AccountAddress:from_str` rather than `AccountAddress::from_hex_literal`. I could split this out into its own PR / commit.
- The address returned by `aptos config show-profiles --profile default` conforms to the standard.
- Same with `aptos node show-validator-set` and every other subcommand in the CLI that outputs an address.
- Validator config tooling:

Before:
```
$ rg address
public-keys.yaml
2:account_address: 0b02b46b26d96166bc6887b9aa2eaf7b1ffdcd73c1d609ba28c4d6abefe26d07

validator-identity.yaml
2:account_address: 0b02b46b26d96166bc6887b9aa2eaf7b1ffdcd73c1d609ba28c4d6abefe26d07

private-keys.yaml
2:account_address: 0b02b46b26d96166bc6887b9aa2eaf7b1ffdcd73c1d609ba28c4d6abefe26d07

validator-full-node-identity.yaml
2:account_address: 0b02b46b26d96166bc6887b9aa2eaf7b1ffdcd73c1d609ba28c4d6abefe26d07
```

After:
```
$ rg address
public-keys.yaml
2:account_address: "0x0900d5c2e257870dd4bdda3856b7f182d26040f434f7445796f3101cc73c3924"

validator-identity.yaml
2:account_address: "0x0900d5c2e257870dd4bdda3856b7f182d26040f434f7445796f3101cc73c3924"

private-keys.yaml
2:account_address: "0x0900d5c2e257870dd4bdda3856b7f182d26040f434f7445796f3101cc73c3924"

validator-full-node-identity.yaml
2:account_address: "0x0900d5c2e257870dd4bdda3856b7f182d26040f434f7445796f3101cc73c3924"
```

With effort any of the above could be left unchanged by explicitly using different functions, but I've tested that each of the above changes has compatibility with the existing ecosystem.

Things intentionally not impacted:
- Any storage of addresses by the node.
- How addresses are returned by the v1 node API.
- How addresses are returned by any existing indexer API.

### Test Plan
Manually running the CLI E2E tests:
```
DOCKER_DEFAULT_PLATFORM=linux/amd64 poetry run python main.py -d --base-network testnet --test-cli-path ~/a/core/target/debug/aptos
```

Confirming that the new CLI works with an older config.
```
cd /tmp/testing
aptos init
~/aptos-core/target/debug/aptos show-profiles
```

Confirming that a node from before this change happily starts with a set of configs generated by the new CLI (once I remove config sections that the new CLI generates that the node in the existing CLI doesn't know about, but not fixing any address stuff).
```
cd /tmp/node
~/aptos-core/target/debug/aptos node run-local-testnet
aptos node run-local-testnet
```


banool:
Currently on hold since the impact of this change is somewhat vast. Lots of tests to fix for starters, and even once those are fixed, it might require some serious thought to make sure no change in behavior resulting from this PR will break anything.

PR #9186, opened 2023-07-20
Make AccountAddress from_str conform to AIP-40, add from_str_strict

banool:
### Stack
- Previous in stack: https://github.com/aptos-labs/aptos-core/pull/8727
- Next in stack: https://github.com/aptos-labs/aptos-core/pull/9183

### Description
The `from_str` implementation of `AccountAddress` does not currently conform to AIP-40. Upstream users often do something like this currently, since even before AIP-40 was defined, they were reaching for a parsing function akin to what the standard defines:
```
if str.starts_with("0x") {
    AccountAddress::from_hex_literal(str)
        .map(Some)
        .map_err(|err| {
            CliError::CommandArgumentError(format!("Failed to parse AccountAddress {}", err))
        })
} else if let Ok(account_address) = AccountAddress::from_str(str) {
    Ok(Some(account_address))
}
```

After this PR this is not necessary and they can just use `from_str` immediately. The new parsing corresponds to the "relaxed" parsing described in the AIP.

For strict parsing (as per the spec), a `from_str_strict` method has been added. I considered just adding another `from_str` method with a different trait, which I think is normally how Rust would prefer you do it, but given FromStr is a very common import, it'd likely mean you'd have to import both FromStr and a hypothetical FromStrStrict trait, which means you'd have to fully qualify which function you're using, which would be a pain.

One thing that this PR doesn't do currently that the reference implementation (v2 TS SDK) does is return specific error codes for each parsing failure. I intend to add this assuming I can do it in a non breaking way, but I want to make sure we agree this is a good approach for AccountAddress in Rust first.

### Test Plan
```
cd third_party/move/move-core/types
cargo test
```


third_party/move/move-core/types/src/account_address.rs
```diff
@@ -336,8 +350,7 @@ impl fmt::Display for AccountAddressParseError {
     fn fmt(&self, f: &mut fmt::Formatter) -> std::fmt::Result {
         write!(
             f,
-            "Unable to parse AccountAddress (must be hex string of length {})",
-            AccountAddress::LENGTH
+            "Unable to parse AccountAddress (must be hex string that conforms to AIP-40)",
```
gregnazario, line 353: I don't think this is a very use friendly message tbh.  This could be provided to an end retail user when passed through several layers.  They should probably not know or care about AIP-40.

Can we perhaps explain simply what might need to be fixed?

third_party/move/move-core/types/src/account_address.rs
```diff
@@ -276,19 +280,29 @@ impl TryFrom<String> for AccountAddress {
...
+            if s.is_empty() {
+                return Err(AccountAddressParseError);
+            }
+            AccountAddress::from_hex_literal(&format!("0x{}", s))
         } else {
-            Self::from_hex(s)
+            if s.len() == 2 {
+                return Err(AccountAddressParseError);
+            }
+            AccountAddress::from_hex_literal(s)
```
gregnazario, line 305: I think from_hex_literal should be handling the length 2.

third_party/move/move-core/types/src/account_address.rs
```diff
@@ -276,19 +280,29 @@ impl TryFrom<String> for AccountAddress {
...
+    /// string representations of addresses are accepted (to maximize compatibility).
+    /// This is allowed by AIP-40 at the discretion of the implementation.
     fn from_str(s: &str) -> Result<Self, AccountAddressParseError> {
-        // Accept 0xADDRESS or ADDRESS
-        if let Ok(address) = AccountAddress::from_hex_literal(s) {
-            Ok(address)
+        if !s.starts_with("0x") {
+            if s.is_empty() {
+                return Err(AccountAddressParseError);
+            }
```
gregnazario, line 299: Similarly for this I think?

third_party/move/move-core/types/src/account_address.rs
banool: Yeah fair call. I would actually prefer to include a URL to the relevant section of the AIP, what do you think about that?

third_party/move/move-core/types/src/account_address.rs
banool: It doesn't actually (if I remember correctly, from when I wrote the tests), and I don't want to change `from_hex_literal`.

third_party/move/move-core/types/src/account_address.rs
banool: It doesn't actually (if I remember correctly, from when I wrote the tests), and I don't want to change `from_hex_literal`.

third_party/move/move-core/types/src/account_address.rs
davidiw: I'd probably just suggest we state it is an invalid address and maybe print out the hex version of what we saw... Otherwise you'll need more explicit parsing errors that enumerate the various different cases.

davidiw (commented):
might it be good to have a complete issue that proposes the changes you plan to make?

for example we have a lot of comments about what to use or how to use this, I think it would be better to just say these are all deprecated and by the end of the stack you will only have 1 or 2 output functions and 1 or 2 input functions.

Output:
* Most readable form (shortening for special cases)
* Full form (no shortening)
Input:
* Permissive -- allows reading all types
* Rigid -- only reads the allowed output types

Then conforming to that interface will improve the code base ... or so I think...

banool:
@davidiw Apologies for not keeping you in the loop but we've done actually exactly what you suggest. In the v2 TS SDK there are now `fromString` and `fromStringRelaxed` functions that correspond to the rigid and permissive parsing you suggest. Similarly `toString` is most readable form and `toStringLong` is the long form unconditionally. 

If you check out https://github.com/aptos-foundation/AIPs/blob/main/aips/aip-40.md#acceptable-input-formats you'll see that this parsing behavior has been specified more clearly. It also links to those reference implementations (where the reference implementation is the v2 TS SDK now) that I mentioned.

I'll put this PR into draft until I can make it align with all this. My plan is to make `from_str` the permissive version and then have a `from_str_strict`, since `from_str` already exists and can't be made more strict without breaking backwards compatibility.

banool:
As for errors explaining the different parsing failures, that's what I do in the v2 TS SDK. I can do the same thing here with a custom error type for from_str.

banool:
I just pushed some changes to comments for `from_str` and added a `from_str_strict` method. I discuss these a bit in the summary of the PR and pose some questions re error codes.

davidiw (approved):

0xmaayan (approved):

gregnazario (approved):

PR #9192, opened 2023-07-20
[State Sync] Enable intelligent syncing mode.

JoshLind:
### Description
This PR enables "intelligent" syncing mode for all nodes (by default). Intelligent syncing mode executes transactions or applies outputs, depending on whichever is cheaper (per chunk).

### Test Plan
Existing test infrastructure.

### DIFF
--- a/config/src/config/state_sync_config.rs
+++ b/config/src/config/state_sync_config.rs
@@ -122,9 +122,9 @@ pub struct StateSyncDriverConfig {
 impl Default for StateSyncDriverConfig {
     fn default() -> Self {
         Self {
-            bootstrapping_mode: BootstrappingMode::ApplyTransactionOutputsFromGenesis,
+            bootstrapping_mode: BootstrappingMode::ExecuteOrApplyFromGenesis,
             commit_notification_timeout_ms: 5000,
-            continuous_syncing_mode: ContinuousSyncingMode::ApplyTransactionOutputs,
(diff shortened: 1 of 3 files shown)


config/src/config/state_sync_config.rs
```diff
@@ -122,9 +122,9 @@ pub struct StateSyncDriverConfig {
 impl Default for StateSyncDriverConfig {
     fn default() -> Self {
         Self {
-            bootstrapping_mode: BootstrappingMode::ApplyTransactionOutputsFromGenesis,
+            bootstrapping_mode: BootstrappingMode::ExecuteOrApplyFromGenesis,
             commit_notification_timeout_ms: 5000,
-            continuous_syncing_mode: ContinuousSyncingMode::ApplyTransactionOutputs,
+            continuous_syncing_mode: ContinuousSyncingMode::ExecuteTransactionsOrApplyOutputs,
```
igor-aptos, line 127: should Intelligent or AutoSelect be in the name? :)

igor-aptos (approved):

config/src/config/state_sync_config.rs
JoshLind: Aah, I'm still so-so on the term `intelligent` 😄 Also, these configs are currently pretty technical 😄, e.g., `fast sync` is actually called `DownloadLatestStates` in the config. When we write the docs, we can use the informal names (e.g., https://aptos.dev/guides/state-sync#fast-syncing), but we can always follow up once we think the informal names make sense (long term).

PR #9194, opened 2023-07-20
add --analyze-mode=max-tps to analyze-validator-performance CLI

igor-aptos:

### DIFF
--- a/crates/aptos/src/node/mod.rs
+++ b/crates/aptos/src/node/mod.rs
@@ -1390,6 +1390,8 @@ pub enum AnalyzeMode {
     /// For each epoch summarize how many validators were in
     /// each of the reliability buckets.
     NetworkHealthOverTime,
+    /// Max TPS
+    MaxTps,
 }
 
 #[async_trait]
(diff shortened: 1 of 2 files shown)


bowenyang007:
What's the purpose of calculating tps this way? To validate or is this expected to be the source of truth. I don't think that having on chain analytics as part of the node is the right direction. In fact at some point we should clean up as much of this as possible from telemetry. Telemetry should be only tracking offchain metrics.

igor-aptos:
@bowenyang007 this is a CLI anyone externally/ operators/ we can use to get the source of truth based on onchain information, provided by the fullnode

I.e. external person can point CLI to their fullnode and confirm number he sees on the explorer. 
I've used this to find a small bug in the number explorer was showing. 

This is not to be used by analytics in the pipelines,  and we can deprecated using CLI there at all,  if we are comfortable in maintaining a correct calculation there

crates/aptos/src/node/mod.rs
```diff
@@ -1440,6 +1444,20 @@ impl CliCommand<()> for AnalyzeValidatorPerformance {
                     true,
                 );
             }
+            if print_max_tps {
+                for (num_blocks_for_max_tps, max_tps) in &epoch_stats.max_tps_per_block_interval {
+                    println!(
+                        "In {}epoch {}: during {} consecutive blocks, peak of {} TPS, ending on version: {}, {} txns over {}s",
```
sitalkedia, line 1450: nit space before "epoch"

sitalkedia (approved):
LGTM

bowenyang007:
Got ya, this is just doing the math from storage. What if data got pruned? Or is our assumption that we should keep at least past X days, long enough to do this math.

crates/aptos/src/node/mod.rs
igor-aptos: that's intentional - if it is a full epoch - it is replaced with empty string. if it is partial epoch , it is replaced with "partial ". no space means we don't end up with double space

igor-aptos:
@bowenyang007 - if things got pruned CLI will say that. 

We do expect some full nodes to not have pruned recent history, so that we can analyze things from Rest endpoint

bchocho (approved):
LGTM

PR #9195, opened 2023-07-20
[indexer] handle edge case of self transfer token v2

bowenyang007:
### Description
If there is a self transfer of token v2, we currently soft delete the token from the user's account. However, this should just be a noop

Need to backfill testnet from transaction version 514946350 and mainnet from transaction version 169213813

### DIFF
--- a/crates/indexer/src/models/token_models/v2_token_ownerships.rs
+++ b/crates/indexer/src/models/token_models/v2_token_ownerships.rs
@@ -164,6 +164,10 @@ impl TokenOwnershipV2 {
 
         // check if token was transferred
         if let Some((event_index, transfer_event)) = &metadata.transfer_event {
+            // If it's a self transfer then skip
+            if transfer_event.get_to_address() == transfer_event.get_from_address() {
+                return Ok(Some((ownership, current_ownership, None, None)));
+            }
             Ok(Some((
                 ownership,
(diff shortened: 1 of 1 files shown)


rtso (approved):

crates/indexer/src/models/token_models/v2_token_ownerships.rs
```diff
@@ -164,6 +164,10 @@ impl TokenOwnershipV2 {
 
         // check if token was transferred
         if let Some((event_index, transfer_event)) = &metadata.transfer_event {
+            // If it's a self transfer then skip
+            if transfer_event.get_to_address() == transfer_event.get_from_address() {
```
movekevin, line 168: Hmm it seems a bit hacky that this needs a special case. Shouldn't this still work if the to address is still this account? I.e. It seems like the code is asymmetric and deleting ownership for from but don't really update it for to?

bowenyang007: It's the ordering right? If I move this code before the previous section, it'll be fine. However, IMO explicitly calling out this edge case is the safer route. I mean there really aren't that many cases in a transfer: transfer to self, transfer to someone else.

movekevin (approved):

PR #9196, opened 2023-07-20
[Move examples] Ensure no self transfer in knight token module.

movekevin:
### Description
Current code self-transfers if the creator is also the recipient which is unnecessary. Also fix some other nits (import ordering, etc.)

### Test Plan
Unit tests

### DIFF
--- a/api/src/tests/objects.rs
+++ b/api/src/tests/objects.rs
@@ -25,7 +25,7 @@ async fn test_gen_object() {
     let user_addr = user.address();
 
     // Publish packages
-    let named_addresses = vec![("token_objects".to_string(), user_addr)];
+    let named_addresses = vec![("hero".to_string(), user_addr)];
     let txn = futures::executor::block_on(async move {
         let path = PathBuf::from(std::env!("CARGO_MANIFEST_DIR"))
             .join("../aptos-move/move-examples/token_objects/hero");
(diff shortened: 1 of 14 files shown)


bowenyang007 (approved):

xbtmatt:
The unit test compilation will fail because `std::signer` is imported with a test only annotation still

```rust
use std::signer;

// ...

#[test_only]
use std::signer;
```

https://github.com/aptos-labs/aptos-core/blob/e7b54a90182c730ad8ee9a2256edf6d359097547/aptos-move/move-examples/token_objects/knight/sources/knight.move#L222

movekevin:
> The unit test compilation will fail because `std::signer` is imported with a test only annotation still
> 
> ```rust
> use std::signer;
> 
> // ...
> 
> #[test_only]
> use std::signer;
> ```
> 
> https://github.com/aptos-labs/aptos-core/blob/e7b54a90182c730ad8ee9a2256edf6d359097547/aptos-move/move-examples/token_objects/knight/sources/knight.move#L222
> 
> Also note it doesn't fail in ci because it's not in `move_unit_tests.rs`:
> 
> ```rust
>     run_tests_for_pkg("token_objects/hero", named_address.clone());
>     run_tests_for_pkg("token_objects/token_lockup", named_address.clone());
>     run_tests_for_pkg("token_objects/ambassador/move", named_address);
> ```

Updated tests

xbtmatt (approved):

PR #9203, opened 2023-07-20
[exp] fix arg passing for forge gha

rustielin:
### Description

Use named args for `workflow_dispatch_forge`, which was previously broken.

### Test Plan

New unit test for this, and run `exp`

### DIFF
--- a/testsuite/exp
+++ b/testsuite/exp
@@ -193,7 +193,6 @@ def workflow_dispatch_forge(
     git_sha: str,
     duration: int = 480,
     test_suite: str = FORGE_DEFAULT_TEST_SUITE,
-    cluster_name: str = FORGE_DEFAULT_CLUSTER_NAME,
     dry_run: bool = False,
     wait: bool = False,
 ) -> bool:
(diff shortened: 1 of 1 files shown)


testsuite/exp
```diff
@@ -214,7 +213,7 @@ def workflow_dispatch_forge(
         "--field",
         f"FORGE_TEST_SUITE={test_suite}",
         "--field",
-        f"FORGE_CLUSTER_NAME={cluster_name}",
+        f"FORGE_CLUSTER_NAME={FORGE_DEFAULT_CLUSTER_NAME}",
```
perryjrandall, line 216: Can we make this an argument instead?

testsuite/exp
```diff
@@ -380,9 +379,10 @@ def run(
             shell,
             new_exp_git_branch,
             git_sha,
-            forge_runner_duration_secs,
-            forge_test_suite,
-            dry_run,
+            duration=forge_runner_duration_secs,
```
perryjrandall, line 382: If you're using kwargs here.... why not make them all kwargs?

Why not replace the args with an "options" struct

perryjrandall (approved):

bchocho (approved):
unblocking

testsuite/exp
rustielin: sure, done. I expect most will not set this though, as we want folks to use the dedicated experimental cluster. This might be useful for the the multi-region karmada forge that @ibalajiarun runs though.

perryjrandall: Nice, you can also do it in a future PR to unblock this now

PR #9213, opened 2023-07-20
[NFT Metadata Crawler] URI Parser

just-in-chang:
### Description

Logic for URI Parser implementation

Fixes EI-250

### DIFF
--- a/ecosystem/nft-metadata-crawler-parser/src/utils/uri_parser.rs
+++ b/ecosystem/nft-metadata-crawler-parser/src/utils/uri_parser.rs
@@ -1,11 +1,39 @@
 // Copyright © Aptos Foundation
 
+use regex::Regex;
+use std::env;
+use url::Url;
+
 pub struct URIParser;
 
 impl URIParser {
     /// Attempts to parse IPFS URI to use dedicated gateway.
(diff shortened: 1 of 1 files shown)


ecosystem/nft-metadata-crawler/nft-metadata-crawler-parser/src/utils/uri_parser.rs
```diff
@@ -1,13 +1,51 @@
...
      * Returns the original URI if parsing fails.
      */
-    pub fn parse(_uri: String) -> String {
-        todo!();
+    pub fn parse(uri: String) -> String {
+        match Self::parse_uri(uri.clone()) {
+            Ok(parsed_uri) => parsed_uri,
+            Err(_) => uri,
+        }
+    }
```
bowenyang007, line 19: I don't think that you should return the original uri if it fails. Why not just return the error and handle it explicitly somewhere else.

jjleng: yeah, this is a util function. It should not swallow errors

bowenyang007 (approved):

jjleng (approved):

Issue #9092, opened 2023-07-14
[Prover] Issue with the inconsistency checker


junkil-park:
This spec introduces an inconsistency: 
https://github.com/aptos-labs/aptos-core/blob/c1b453440b3d2f2c01b0299e31090fb584b1b6e7/aptos-move/framework/aptos-framework/sources/event.spec.move#L10

However, the daily inconsistency check doesn't catch it:
https://github.com/aptos-labs/aptos-core/actions/workflows/prover-daily-test.yaml

PR #9218, opened 2023-07-20
[Spec] Fix spec in the event module to avoid inconsistency issue

rahxephon89:
### Description

This PR fixes a spec in the event module avoid the inconsistency issue described in #9092.


movekevin (approved):

junkil-park (approved):
Thank you!

PR #9219, opened 2023-07-21
Remove aggregator_enabled flag

vusirikala:
### Description

Remove `aggregator_enabled` flag.

### DIFF
--- a/aptos-move/aptos-aggregator/src/aggregator_extension.rs
+++ b/aptos-move/aptos-aggregator/src/aggregator_extension.rs
@@ -299,19 +299,12 @@ pub struct AggregatorData {
 impl AggregatorData {
     /// Returns a mutable reference to an aggregator with `id` and a `limit`.
     /// If transaction that is currently executing did not initialize it, a new aggregator instance is created.
-    /// The state of the new aggregator instance depends on the `aggregator_enabled` flag.
-    /// If the `aggregator_enabled` flag is true, the new aggregator instance
-    /// is initialized with zero and in a delta state.
-    /// If the `aggregator_enabled` flag is false, the new aggregator instance
-    /// is initialized in the Data state with its latest value.
     /// Note: when we say "aggregator instance" here we refer to Rust struct and
     /// not to the Move aggregator.
     pub fn get_aggregator(
         &mut self,
         id: AggregatorID,
         limit: u128,
-        resolver: &dyn TableResolver,
-        aggregator_enabled: bool,
     ) -> PartialVMResult<&mut Aggregator> {
         let aggregator = self.aggregators.entry(id).or_insert(Aggregator {
             value: 0,
@@ -319,10 +312,6 @@ impl AggregatorData {
             limit,
             history: Some(History::new()),
         });
-
-        if !aggregator_enabled {
-            aggregator.read_and_materialize(resolver, &id)?;
-        }
         Ok(aggregator)
     }
 
@@ -401,7 +390,7 @@ mod test {
         let mut aggregator_data = AggregatorData::default();
 
         let aggregator = aggregator_data
-            .get_aggregator(aggregator_id_for_test(300), 700, &*TEST_RESOLVER, true)
+            .get_aggregator(aggregator_id_for_test(300), 700)
             .expect("Get aggregator failed");
         assert_err!(aggregator.read_and_materialize(&*TEST_RESOLVER, &aggregator_id_for_test(700)));
     }
@@ -412,7 +401,7 @@ mod test {
         aggregator_data.create_new_aggregator(aggregator_id_for_test(200), 200);
 
         let aggregator = aggregator_data
-            .get_aggregator(aggregator_id_for_test(200), 200, &*TEST_RESOLVER, true)
+            .get_aggregator(aggregator_id_for_test(200), 200)
             .expect("Get aggregator failed");
         assert_ok!(aggregator.add(100));
         assert_ok!(aggregator.read_and_materialize(&*TEST_RESOLVER, &aggregator_id_for_test(200)));
@@ -432,7 +421,7 @@ mod test {
         let mut aggregator_data = AggregatorData::default();
 
         let aggregator = aggregator_data
-            .get_aggregator(aggregator_id_for_test(500), 500, &fake_resolver, false)
+            .get_aggregator(aggregator_id_for_test(500), 500)
             .expect("Get aggregator failed");
         assert_eq!(aggregator.state, AggregatorState::Data);
         assert_eq!(aggregator.value, 150);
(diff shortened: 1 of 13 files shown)


aptos-move/aptos-aggregator/src/aggregator_extension.rs
```diff
@@ -299,30 +299,19 @@ pub struct AggregatorData {
...
-        aggregator_enabled: bool,
     ) -> PartialVMResult<&mut Aggregator> {
         let aggregator = self.aggregators.entry(id).or_insert(Aggregator {
             value: 0,
             state: AggregatorState::PositiveDelta,
             limit,
             history: Some(History::new()),
         });
-
-        if !aggregator_enabled {
```
movekevin, line 323: Can this break replay?

vusirikala: We have earlier added this flag but always set this flag to true in the execution. So, removing this flag won't break backward compatibility.

igor-aptos (approved):

georgemitenkov:
Why not just reversing the PR? :)

aptos-move/aptos-aggregator/src/aggregator_extension.rs
```diff
@@ -432,7 +421,7 @@ mod test {
         let mut aggregator_data = AggregatorData::default();
 
         let aggregator = aggregator_data
-            .get_aggregator(aggregator_id_for_test(500), 500, &fake_resolver, false)
+            .get_aggregator(aggregator_id_for_test(500), 500)
```
georgemitenkov, line 424: Cannot comment on above, but this test makes no sense anymore if we remove the flag? Let's remove it as well? I think there are a few tests like that as well which are duplicates without the flag.

aptos-move/aptos-aggregator/src/aggregator_extension.rs
```diff
@@ -455,12 +444,7 @@ mod test {
     #[test]
     fn test_unknown_aggregator_disabled_fail() {
         let mut aggregator_data = AggregatorData::default();
-        assert_err!(aggregator_data.get_aggregator(
-            aggregator_id_for_test(200),
-            200,
-            &*TEST_RESOLVER,
-            false,
-        ));
+        assert_err!(aggregator_data.get_aggregator(aggregator_id_for_test(200), 200,));
```
georgemitenkov, line 447: nit: no trailing comma, but actually test should be removed :)

aptos-move/aptos-vm/src/aptos_vm.rs
```diff
@@ -1373,7 +1372,6 @@ impl AptosVM {
                     &simulation_vm.0.as_move_resolver(state_view),
                     txn,
                     &log_context,
-                    false,
```
georgemitenkov, line 1376: How does the simulation work now?  Let's fix this because otherwise we have some stale code that talks about "aggregator enabled" which is a non-existent thing.

aptos-move/framework/src/natives/aggregator_natives/context.rs
```diff
@@ -82,7 +76,6 @@ impl<'a> NativeAggregatorContext<'a> {
                 AggregatorState::Data => AggregatorChange::Write(value),
                 AggregatorState::PositiveDelta => {
                     // Aggregator state can be a delta only if aggregators are enabled.
-                    assert!(self.aggregator_enabled);
```
georgemitenkov, line 85: The comment doesn't make sense anymore.

aptos-move/framework/src/natives/aggregator_natives/context.rs
```diff
@@ -91,7 +84,6 @@ impl<'a> NativeAggregatorContext<'a> {
                 },
                 AggregatorState::NegativeDelta => {
                     // Aggregator state can be a delta only if aggregators are enabled.
-                    assert!(self.aggregator_enabled);
```
georgemitenkov, line 94: ditto

aptos-move/framework/src/natives/aggregator_natives/context.rs
```diff
@@ -209,53 +201,28 @@ mod test {
...
-            context.resolver,
-            aggregator_enabled,
-        ));
-        assert_ok!(aggregator_data.get_aggregator(
-            aggregator_id_for_test(200),
-            200,
-            context.resolver,
-            aggregator_enabled,
-        ));
+        assert_ok!(aggregator_data.get_aggregator(aggregator_id_for_test(100), 100,));
```
georgemitenkov, line 212: nit: trailing comma, otherwise not sure if the test logic still makes sense, but we can revisit later.

georgemitenkov (changes requested):
LGTM, requesting changes to make sure:
1. Tests and comments do not mention "enabled/disabled" things. Otherwise it will be super-confusing for anyone from the outside.
2. Let's make sure simulation works (or at least fails with a reasonable message)

aptos-move/aptos-aggregator/src/aggregator_extension.rs
vusirikala: Removed unnecessary test cases.

aptos-move/aptos-aggregator/src/aggregator_extension.rs
vusirikala: Done.

aptos-move/framework/src/natives/aggregator_natives/context.rs
vusirikala: Done.

aptos-move/framework/src/natives/aggregator_natives/context.rs
vusirikala: Done.

aptos-move/framework/src/natives/aggregator_natives/context.rs
vusirikala: Done.

aptos-move/aptos-vm/src/aptos_vm.rs
georgemitenkov: I think we panicked in case of failures there😄 the problem right now is that we re-execute but because there is no flag we will keep failing again, and again, etc.

georgemitenkov (approved):
LGTM, thanks!

Please let's also make sure simulation works (i.e. no re-execution on failure), just error or something like that should be ok, current use cases do not fail anyway.

aptos-move/aptos-vm/src/aptos_vm.rs
vusirikala: So, what's the solution now?

PR #9220, opened 2023-07-21
[dag] add basic test for order rule

zekun000:
Write an example of dag to test basic ordering, and update the voting power check to support f+1 besides 2f+1

### DIFF
--- a/consensus/src/dag/dag_store.rs
+++ b/consensus/src/dag/dag_store.rs
@@ -183,7 +183,7 @@ impl Dag {
                             .any(|cert| cert.metadata() == metadata)
                     })
                     .map(|node_status| node_status.as_node().author());
-                validator_verifier.check_voting_power(votes).is_ok()
+                validator_verifier.check_voting_power(votes, false).is_ok()
             })
             .unwrap_or(false)
     }
@@ -251,6 +251,7 @@ impl Dag {
             .check_voting_power(
                 self.get_round_iter(round)?
                     .map(|node_status| node_status.as_node().metadata().author()),
+                true,
             )
             .is_ok()
         {
--- a/consensus/src/dag/order_rule.rs
+++ b/consensus/src/dag/order_rule.rs
@@ -135,13 +135,14 @@ impl OrderRule {
         self.lowest_unordered_anchor_round = anchor.round() + 1;
 
         let mut dag_writer = self.dag.write();
-        let ordered_nodes: Vec<_> = dag_writer
+        let mut ordered_nodes: Vec<_> = dag_writer
             .reachable_mut(&anchor, None)
             .map(|node_status| {
                 node_status.mark_as_ordered();
                 node_status.as_node().clone()
             })
             .collect();
+        ordered_nodes.reverse();
         if let Err(e) = self.ordered_nodes_sender.unbounded_send(ordered_nodes) {
             error!("Failed to send ordered nodes {:?}", e);
         }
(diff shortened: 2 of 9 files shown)


types/src/validator_verifier.rs
```diff
@@ -321,6 +321,7 @@ impl ValidatorVerifier {
     pub fn check_voting_power<'a>(
         &self,
         authors: impl Iterator<Item = &'a AccountAddress>,
+        check_quorum: bool,
```
ibalajiarun, line 324: nit: f+1 and 2f+1 are both quorums, no?

may be `check_strong_quorum` or `check_supermajority_quorum`?

consensus/src/dag/dag_store.rs
```diff
@@ -183,7 +183,7 @@ impl Dag {
                             .any(|cert| cert.metadata() == metadata)
                     })
                     .map(|node_status| node_status.as_node().author());
-                validator_verifier.check_voting_power(votes).is_ok()
+                validator_verifier.check_voting_power(votes, false).is_ok()
```
ibalajiarun, line 186: have a TODO to make this an onchain config? we were discussing this could be used for deterministically skipping anchors?

consensus/src/dag/order_rule.rs
```diff
@@ -135,13 +135,14 @@ impl OrderRule {
...
         let mut dag_writer = self.dag.write();
-        let ordered_nodes: Vec<_> = dag_writer
+        let mut ordered_nodes: Vec<_> = dag_writer
             .reachable_mut(&anchor, None)
             .map(|node_status| {
                 node_status.mark_as_ordered();
                 node_status.as_node().clone()
             })
             .collect();
+        ordered_nodes.reverse();
```
ibalajiarun, line 145: why not use rev() on the iter?

consensus/src/dag/tests/order_rule_tests.rs
```diff
@@ -187,3 +187,83 @@ proptest! {
...
+        // anchor (1, 0) has 1 votes, anchor (3, 1) has 2 votes and a path to (1, 0)
+        vec![(1, 0)],
+        // anchor (2, 1) has 3 votes
+        vec![(1, 2), (1, 1), (2, 1)],
+        // anchor (3, 1) has 2 votes
+        vec![(1, 3), (2, 2), (2, 0), (3, 1)],
+        // anchor (4, 2) has 3 votes
+        vec![(3, 3), (3, 2), (3, 0), (4, 2)],
+        // anchor (5, 2) has 3 votes
+        vec![(4, 1), (4, 0), (5, 2)],
```
ibalajiarun, line 256: same round authors is reversed. its deterministic but not intuitive. can we add a comment somewhere in order_rule that this is the expected order.

consensus/src/dag/dag_store.rs
zekun000: later :)

consensus/src/dag/order_rule.rs
zekun000: this is not a double iterator so it doesn't have rev()

consensus/src/dag/tests/order_rule_tests.rs
zekun000: yeah it's because we reverse the order

types/src/validator_verifier.rs
zekun000: I don't think f+1 is a "quorum" but I might be wrong

ibalajiarun: @sasha8 what do you think?

ibalajiarun (approved):

types/src/validator_verifier.rs
ibalajiarun: nvm, just saw the comment in the other PR.

PR #9227, opened 2023-07-21
[cached packages] Put .mrb back into OUT_DIR (target)

wrwg:
Too much trouble checking it in, we may come back with a different solution over time.


davidiw (approved):
It was a worthwhile effort.

I did experience the pain but the exporting of the skip command seemed to work okayish.

gregnazario (approved):

wrwg:
> It was a worthwhile effort.
> 
> I did experience the pain but the exporting of the skip command seemed to work okayish.

It did not actually work because this uses a file in the target dir and if you thats not build then things fail... I added this SKIP thing mostly to for hacking around the well-known compiler bootstrapping problem into which this design ever now and then runs. Moreover, it is rebuilding frequently if you switch branches or if the feature set changes, because the target dir is invalidated.

However, my personal machine is fast enough so I can live with rebuilding this.

Issue #8875, opened 2023-06-27
[Bug] error out in compilation when encounter an unknown annotation/attribute


lightmark:
# 🚀 Feature Request

Compiler ignores unknown annotations but actually it should just error out. Otherwise, typo must be introduced while the author doesn't know. For example -> https://github.com/aptos-labs/aptos-core/pull/8866 for examples.

## Motivation
`#[testonly]` in  https://github.com/aptos-labs/aptos-core/pull/8866 for examples.

brmataptos:
list of allowed attributes is implicitly in the constants defined [here](https://github.com/aptos-labs/aptos-core/blob/564f66ad09f3fe26cfa10e63c834c2de69a7ca31/third_party/move/move-compiler/src/shared/mod.rs#L518).   

Oops, [extended_checks.rs](https://github.com/aptos-labs/aptos-core/blob/c8e0581f9b8a8aa1888f118c881a8073d1e5172a/aptos-move/framework/src/extended_checks.rs#L24) also has a few known attributes ("legacy_entry_fun", "view", "group", "scope", "resource_group", "resource_group_member", "group"), with an odd variety of constant names associated with them.

We’re also missing #[deprecated{

----
More from Slack discussion:

Kevin Hoang
Another thing to note is we need a good error message in case users already use some custom attributes

Wolfgang Grieskamp
The cli needs a flag to skip those checks

Kevin Hoang
We might also consider giving a harsher warning for variants of test_only such as testonly as those are the most risky ones (accidentally releasing test code)

brmataptos:
Specific Requirements:
- mechanism to register and check known attribute names from various code
- register attributes
  - third_party/move/move-compiler/src/shared/mod.rs#L518
  - aptos-move/framework/src/extended_checks.rs#L24
  - #[deprecated]
- good error message in case users use custom attributes
  - CLI needs a flag to skip those checks
  - maybe a flag to allow registering more attributes
- also consider harsher warning for variants of test_only such as testonly
- tests

brmataptos:
Wolfgang is apparently greatly proposed to a flag registering more attributes, so we'll just have a flag to skip checks.

wrwg:
Yes, I do think we should leave user registerable attributes to the compiler v2 and language extensions. Specifically, I propose to do attributes similar as in C#/Java, so they can be declared in the language, together with typing of their parameters, as in:

```
#[define_attribute] public fun my_attribute(param1: u64, param2: String)
```

[...]

Issue #9228, opened 2023-07-21
[Feature Request] [compiler] Refactor: Separate configuration BuildConfig from code


brmataptos:
# 🚀 Feature Request

Unfortunately, there are some Serialize/Deserialize traits on the currently available structures like BuildConfig (defined in third_party/move/tools/move-package/src/lib.rs) which also implements the entry points from Aptos to third_party BuildConfig::compile_package* , which make it hard to pass in something smart.

We should separate the configuration BuildConfig from the code currently in BuildConfig impl so we can have that code accept config which is non-serializable to make it possible to pass non-serializable config/plugins through the Aptos/third_party compile interface.

Once this is done, language aspects like known_attributes can be passed more easily from Aptos into the compiler..

wrwg:
`BuildConfig` is a command line option data structure which uses clap.. It should not (and must not, by design) have behavior attached. If you want to pass in behavior don’t use BuiltConfig but add a new parameter.

```
#[derive(Debug, Parser, Clone, Serialize, Deserialize, Eq, PartialEq, PartialOrd, Default)]
#[clap(author, version, about)]
pub struct BuildConfig {
    /// Compile in 'dev' mode. The 'dev-addresses' and 'dev-dependencies' fields will be used if
    /// this flag is set. This flag is useful for development of packages that expose named
    /// addresses that are not set to a specific value.
    #[clap(name = "dev-mode", short = 'd', long = "dev", global = true)] <---- command line option
    pub dev_mode: bool,
   ...
```

brmataptos:
As an alternative, there could be an additional operand to BuildConfig operations, perhaps DialectConfig, which can be defined more flexibly.

brmataptos:
Here is an alternative approach: https://github.com/aptos-labs/aptos-core/issues/9615

brmataptos:
I note 2 main Problems with this PR:

1. too many files changed
  - config option added in a lot of places
  - move-package/tests outputs include these config values
  - changing ModelConfig warnings to generate errors isn't enough
    - We have warnings in move-compiler/ code used by V2:
      - parser/lexer.rs: Syntax::InvalidDocComment,
      - expansion/translate.rs: NameResolution::DeprecatedModule, Declarations::UnknownAttribute, UnusedItem::Alias,

2. test outputs change
  - PR changes so that we don't run bytecode validation on a test if
    there are errors, matching CLI compiler behavior.
  - PR adds a message when exiting to give visibility into exit/no-exit
[...]

PR #9229, opened 2023-07-21
Thoroughly fix Issue 8875: Warn in compilation when encounter an unknown attribute

brmataptos:
### Description

Fix issue https://github.com/aptos-labs/aptos-core/issues/8875 by adding a check for unknown attributes when parsing.

To handle different language variants (Aptos, third_party, EVM, Async)
we create and pass around a set of known attributes.  While this may
seem a bit unnecessary, it pioneers the code path needed to pass
language extensions in the future.

Ultimately (https://github.com/aptos-labs/aptos-core/issues/9228)
we should separate the configuration BuildConfig from the
code in BuildConfig so we can have that code accept "language
extension" config which is non-serializable to make it possible to
pass non-serializable config/plugins through the Aptos/third_party
compile interface.  This would allow known_attributes to be passed
by reference (and be a HashSet) to be more efficient.

### More details

We thoroughly fix this issue for the moment.  This involves:
- Issuing a warning on unknown attributes when compiling Move source code.
- Distinguishing the set of known attributes based on Move dialect (base, Aptos, EVM, Async)
- Making sure the dialect is selected for every one of the many different paths into the compiler
- A temporary hack to avoid warning on unknown attributes used in 0x1:aptos_std module sources
- Fixes to the attributes for those aptos stdlib sources so they can eventually be deployed and the hack removed
- Fix to Move.toml files for internal tests to point to the local stdlib rather than pulling from network

Note that a quicker hack might have fixed this issue, and this code change is pretty involved.

However, in addition to its immediate effect, it also illustrates all the paths into the compiler (search for "known_attributes" and "skip_attribute_checks"), pioneering a path which can be used in the future to pass a "MoveDialect" container to make it cleaner to support multiple dialects through the shared third-party/ repo.

See https://pastebin.com/TH1nwNXs for some notes I took while enumerating paths through the code.

### Commits:
- cda4ae67d2 [compiler] test cases and stdlib source code updates for attributes checks PR
- c123a0f2eb [compiler] Fix issue-8875 thoroughly: error for unknown attributes, add --skip-attribute-check flag
- 0dac829773 [compiler] Hack to move-compiler to avoid warning about unknown attributes in aptos_std=0x1. This avoids surprising warning on currently deployed library code.
- 31d69ee2b4 [third party/.../test, compiler] Fix Move.toml files to point to local files instead of network to avoid tests using old stdlib.  Oddly, this requires fixes to capitalization of std.  Fix the resolution warning to make it clear that std needs to be uncapitalized.
- dfb800a8e0 (HEAD -> brm-issue-8875, origin/brm-issue-8875) [aptos stdlib] Now that we have verified that the move-compiler Aptos stdlib hack works, fix stdlib to use #[test_only] on test helper functions instead of #[testonly].  Leave struct attributes since it's a deployment problem to make it disappear.

### Test Plan

Added more attribute examples and test cases.


aptos-move/framework/src/extended_checks.rs
```diff
@@ -30,6 +36,30 @@ const RESOURCE_GROUP_NAME: &str = "group";
 const RESOURCE_GROUP_SCOPE: &str = "scope";
 const VIEW_FUN_ATTRIBUTE: &str = "view";
 
+// top-level attribute names, only.
+pub fn get_all_attribute_names_set() -> &'static BTreeSet<String> {
+    const ALL_ATTRIBUTE_NAMES: [&str; 4] = [
```
wrwg, line 41: nit: you could use crate [maplit](https://docs.rs/maplit/latest/maplit/), which would allow you to do do
```
static KNOWN_ATTRIBUTES_SET: Lazy<BTreeSet<String>> = Lazy::new(|| {
    btreeset!{LEGACY_ENTRY_FUN_ATTRIBUTE, RESOURCE_GROUP, ...}
});
```

Or you just do:

```
static KNOWN_ATTRIBUTES_SET: Lazy<BTreeSet<String>> = Lazy::new(|| {
    ALL_ATTRIBUTE_NAMES.into_iter().collect()
});
```

aptos-move/framework/src/built_package.rs
```diff
@@ -68,6 +68,10 @@ pub struct BuildOptions {
     pub skip_fetch_latest_git_deps: bool,
     #[clap(long)]
     pub bytecode_version: Option<u32>,
+    #[clap(long)]
+    pub skip_attribute_checks: bool,
+    #[clap(skip)]
+    pub known_attributes: BTreeSet<String>,
```
wrwg, line 74: I don't think this need to be configurable, since the BuildConfig is only for Aptos, and in Aptos we wont use command line options to define allowed attributes. Looking at existing languages with attributes (Java/C#/Python come to mind), none of them have attributes defined this way. I'm worried we are creating technical debt here by adding this option which we now have to maintain forever. Can we remove it and hardwire the set of known attributes? I also wouldn't mind if, for everything else then Aptos (e.g. evm or async flavor), we just use --skip-attribute-checks.

third_party/move/move-compiler/src/command_line/compiler.rs
```diff
@@ -95,6 +97,7 @@ impl<'a> Compiler<'a> {
     pub fn from_package_paths<Paths: Into<Symbol>, NamedAddress: Into<Symbol>>(
         targets: Vec<PackagePaths<Paths, NamedAddress>>,
         deps: Vec<PackagePaths<Paths, NamedAddress>>,
+        known_attributes: &BTreeSet<String>,
```
wrwg, line 100: Do we really need to change this API here? I don't see a use case where known_attributes would be passed into the compiler like this.

third_party/move/move-compiler/src/expansion/translate.rs
```diff
@@ -574,12 +578,38 @@ fn unique_attributes(
             | E::Attribute_::Assigned(n, _)
             | E::Attribute_::Parameterized(n, _) => *n,
         };
-        let name_ = match known_attributes::KnownAttribute::resolve(sym) {
-            None => E::AttributeName_::Unknown(sym),
+        let name_ = match KnownAttribute::resolve(sym) {
+            None => {
+                let flags = &context.env.flags();
```
wrwg, line 583: It would have been simpler to do this in extended_checks, plus fitting well to compiler v2, because extended_checks is using the same AST/env than that one. That is, in the new compiler, we can just run extended_checks after regular type checking.

third_party/move/move-compiler/tests/move_check/parser/attribute_placement.exp
```diff
@@ -0,0 +1,84 @@
+error[E02015]: invalid attribute
```
wrwg, line 1: These tests are expected to pass.

third_party/move/move-compiler/tests/move_check/parser/attribute_variants.exp
```diff
@@ -0,0 +1,60 @@
+error[E02015]: invalid attribute
```
wrwg, line 1: These too I think.

third_party/move/move-compiler/tests/move_check/parser/duplicate_attributes.exp
```diff
@@ -1,3 +1,15 @@
+error[E02015]: invalid attribute
```
wrwg, line 1: Expected different errors here I think

aptos-move/framework/src/built_package.rs
brmataptos: The desire is not to have it be configurable by the user, but to make it configurable per Move dialect (base, Aptos, EVM, Async).  As we discussed, this may be overkill for the current use case, but can serve as a guide for some more interesting dialect changes later.

third_party/move/move-compiler/src/command_line/compiler.rs
brmataptos: The problem is that there are a huge number of paths into the compiler today.  Maybe they shouldn't all be there, but I've traced them all (see https://pastebin.com/TH1nwNXs for some of my notes).

third_party/move/move-compiler/tests/move_check/parser/attribute_placement.exp
brmataptos: I copied the same check to the /skip_attribute_checks/ directory where you can see the old result.

third_party/move/move-compiler/tests/move_check/parser/attribute_variants.exp
brmataptos: I copied the same check to the /skip_attribute_checks/ directory where you can see the old result.

third_party/move/move-compiler/tests/move_check/parser/duplicate_attributes.exp
brmataptos: I copied the same check to the /skip_attribute_checks/ directory where you can see the old result.

aptos-move/framework/src/extended_checks.rs
brmataptos: I missed this earlier.  I'll try it if/when I make more changes.

third_party/move/move-compiler/src/expansion/translate.rs
brmataptos: Simpler, yes.  But this was more educational.

aptos-move/framework/aptos-stdlib/sources/string_utils.move
```diff
@@ -107,7 +107,7 @@ module aptos_std::string_utils {
         native_format_list(&b"a = {} b = {} c = {}", &l);
     }
 
-    #[testonly]
+    /// #[test_only]
```
davidiw, line 110: ? why the comment out?

third_party/move/evm/move-to-yul/src/lib.rs
```diff
@@ -102,6 +107,8 @@ pub fn run_to_abi_metadata<W: WriteColor>(
 ) -> anyhow::Result<Vec<Metadata>> {
     // Run the model builder.
     let addrs = parse_addresses_from_options(options.named_address_mapping.clone())?;
+    let flags = Flags::empty().set_flavor("async");
```
davidiw, line 110: I guess there's no enum for flavors?

third_party/move/move-compiler/src/expansion/translate.rs
```diff
@@ -384,6 +394,32 @@ fn set_sender_address(
     })
 }
 
+// This is a hack to recognize APTOS StdLib to avoid warnings on some old errors.
+// This will be removed after library attributes are cleaned up.
+// (See https://github.com/aptos-labs/aptos-core/issues/9410)
+fn module_is_in_aptos_stdlib(module_address: Option<Spanned<Address>>) -> bool {
```
davidiw, line 400: maybe we should just relabel this to "stdlib" but I guess this just exists temporarily anyway...

davidiw (approved):
I suspect @wrwg should be the definitive approver here, but it looks good to me. Modulo some tiny feedback on commenting out some code.

aptos-move/framework/aptos-stdlib/sources/string_utils.move
brmataptos: We can't actually use #[test_only] on this struct because it will make the module incompatible for upgrades (according to what @wrwg tells me, anyway).  So instead I put it in a comment that will pass through to docs.  Leaving the typo '#[testonly]' attribute could be safe, but might encourage users to keep misspelling, so it's deleted.

third_party/move/evm/move-to-yul/src/lib.rs
brmataptos: There is not.  I can't boil the ocean.

third_party/move/move-compiler/src/expansion/translate.rs
brmataptos: Thx.

aptos-move/framework/src/extended_checks.rs
wrwg: I still think you should use at least my 2nd suggestion to make the initialization of KNOWN_ATTRIBUTES_SET more idiomatic and shorter.

third_party/move/move-compiler/tests/move_check/parser/aptos_stdlib_attributes.move
```diff
@@ -0,0 +1,8 @@
+module aptos_std::M {
+    #[a, a(x = 0)]
+    fun foo() {}
+
+    #[testonly]
```
wrwg, line 5: This test case is really hard to understand, I think you are testing the warning suppression here. Perhaps name the module `aptos_std::module_with_suppressed_warnings`. BTW newer Move naming conventions have modules in small letters.

third_party/move/tools/move-cli/tests/build_tests/dependency_chain/args.exp
```diff
@@ -2,3 +2,9 @@ Command `build -v`:
 INCLUDING DEPENDENCY Bar
 INCLUDING DEPENDENCY Foo
 BUILDING A
+warning[W02016]: unknown attribute
+  ┌─ ./sources/A.move:1:3
+  │
+1 │ #[evm_contract] // for passing evm test flavor
```
wrwg, line 8: Should this be added to the evm flavor?

wrwg (commented):
This looks great and very close to land, just a few remaining nits and questions.

wrwg:
Checking, what is the state of this PR? Are we still planning on landing this?

brmataptos:
Yes.

third_party/move/tools/move-cli/tests/build_tests/dependency_chain/args.exp
brmataptos: #[evm_contract] is already in the evm flavor, but this test is run without evm flavor so it should yield a warning, according to my understanding.

I'm pretty sure we don't want any attributes to enable async/evm flavor if it's not selected by flag.  There is an arch= option in Move.toml files which in one test case selects "async" and I think it could select "evm", which will likely alter the compilation flavor, but since we're not supporting these flavors I don't think it's worth making sure it works.

movekevin (approved):

wrwg (approved):

PR #9234, opened 2023-07-21
add chunking to grpc to avoid large response is rejected by the client.

larry-aptos:
### Description
* add chunking to grpc to avoid large response is rejected by the client.

### DIFF
--- a/ecosystem/indexer-grpc/indexer-grpc-fullnode/src/stream_coordinator.rs
+++ b/ecosystem/indexer-grpc/indexer-grpc-fullnode/src/stream_coordinator.rs
@@ -8,6 +8,7 @@ use crate::{
 };
 use aptos_api::context::Context;
 use aptos_api_types::{AsConverter, Transaction as APITransaction, TransactionOnChainData};
+use aptos_indexer_grpc_utils::{chunk_transactions, constants::MESSAGE_SIZE_LIMIT};
 use aptos_logger::{error, info, sample, sample::SampleRate};
 use aptos_protos::{
     internal::fullnode::v1::{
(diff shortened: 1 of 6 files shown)


bowenyang007 (approved):

ecosystem/indexer-grpc/indexer-grpc-utils/src/lib.rs
```diff
@@ -60,3 +62,73 @@ pub fn time_diff_since_pb_timestamp_in_secs(timestamp: &Timestamp) -> f64 {
...
+            }),
+            ..Transaction::default()
+        };
+        // Create a vec with 10 transactions.
+        let transactions = vec![t.clone(); 10];
+        assert!(t.encoded_len() > 1);
+        let chunked_transactions = chunk_transactions(transactions, 1);
+        assert_eq!(chunked_transactions.len(), 10);
+    }
+    #[test]
```
bowenyang007, line 113: extra line

ecosystem/indexer-grpc/indexer-grpc-fullnode/src/stream_coordinator.rs
```diff
@@ -24,6 +25,8 @@ use tokio::sync::mpsc;
 use tonic::Status;
 
 type EndVersion = u64;
+// Limit the message size to 15MB. By default the downstream can receive up to 15MB.
+const MESSAGE_SIZE_LIMIT: usize = 1024 * 1024 * 15;
```
bowenyang007, line 29: let's use only 1 constant.

bowenyang007 (approved):

bowenyang007:
Btw we should add in docs somewhere that the max size is 80MB. Downstream I would want to keep a buffer and let's say I want 100 batches. That would be 8GB of memory and maybe not practical, so this is an important piece of data.

ecosystem/indexer-grpc/indexer-grpc-utils/src/lib.rs
larry-aptos: done.

ecosystem/indexer-grpc/indexer-grpc-fullnode/src/stream_coordinator.rs
larry-aptos: done.

PR #9235, opened 2023-07-21
[Helm] Fix templating of aptos-vector-log-agent

sionescu:
### Description

The lack of a terminator means that Helm was effectively generating a single resource with duplicated fields, and K8s does not signal an error when applying such a malformed resource, which means that latter fields (in this case the loki-credentials secret) overrode the earlier ones.

Example:

```
apiVersion: v1
kind: Secret
metadata:
  name: humio-credentials
  namespace: vector
apiVersion: v1
kind: Secret
metadata:
  name: loki-credentials
  namespace: vector
```

### Test Plan

Deployed manually to cluster `aptos-devtestnet/aptos-pfn`.


rustielin (approved):

perryjrandall (approved):
nice catch

PR #9242, opened 2023-07-22
[dag] persist ordered anchor for recovery

zekun000:

### DIFF
--- a/types/src/validator_verifier.rs
+++ b/types/src/validator_verifier.rs
@@ -243,7 +243,7 @@ impl ValidatorVerifier {
             pub_keys.push(validator.public_key());
         }
         // Verify the quorum voting power of the authors
-        self.check_voting_power(authors.iter())?;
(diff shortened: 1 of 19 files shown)


sasha8 (approved):
LGTM

types/src/validator_verifier.rs
```diff
@@ -321,6 +321,7 @@ impl ValidatorVerifier {
     pub fn check_voting_power<'a>(
         &self,
         authors: impl Iterator<Item = &'a AccountAddress>,
+        check_quorum: bool,
```
sasha8, line 324: I'm not too fond of the name. Maybe "super-majority quorum"?

sasha8 (approved):
Some of the code is going to prod the next release, right? I tried to read carefully these parts and they are straightforward, but just wonder if we need more tests there?

ibalajiarun (approved):

PR #9246, opened 2023-07-22
Update external-resources.md

cryptomolot:
Added article about basic security methods on Aptos Fullnode.


davidiw (commented):
I skimmed through the article, it might be useful to explain a bit about what you're doing in it. For example, your public fullnode doesn't seem particularly useful except to allow other public fullnodes to connect. That is you don't seem to be exposing your API service.

cryptomolot:
> I skimmed through the article, it might be useful to explain a bit about what you're doing in it. For example, your public fullnode doesn't seem particularly useful except to allow other public fullnodes to connect. That is you don't seem to be exposing your API service.

thanks I will think how to improve

davidiw:
Gonna close out, feel free to re-open / update when it makes sense.

PR #9260, opened 2023-07-24
[Rust] Upgrade to Rust 1.71.0

JoshLind:
### Description
This PR updates the `aptos-core` repository to use the [latest version](https://github.com/rust-lang/rust/blob/master/RELEASES.md) of the rust toolchain.

### Test Plan
Existing test infrastructure.


brianolson (approved):

execution/executor-benchmark/src/lib.rs
```diff
@@ -487,6 +487,49 @@ fn add_accounts_impl<V>(
     );
 }
 
+struct GasMesurement {
```
gregnazario, line 490: I know this probably isn't you but it's spelled wrong

gregnazario (commented):
Can we test this build on Windows?

0xjinn:
I know we just upgraded to `1.7` recently. I wonder what is the need to get to 1.71?

Also, we will most likely need to upgrade CLI homebrew formula to match this new version `1.71` - https://github.com/Homebrew/homebrew-core/blob/master/Formula/aptos.rb

I can open a PR in homebrew if needed 😄

execution/executor-benchmark/src/lib.rs
JoshLind: Yeah, this wasn't me. I didn't want to litter the PR with other fixes, hence why I've left it as is 😄

JoshLind:
> I wonder what is the need to get to 1.71?

@0xjinn, we're hoping that the upgrade will help address some of the spurious compiler panics we've seen in relation to MIR (e.g., https://github.com/rust-lang/rust/issues/112637).

> I can open a PR in homebrew if needed 😄

Thanks @0xjinn 😄 Once this lands, we can also update the CLI docs (if we update the CLI like we did last time?)

JoshLind:
> Can we test this build on Windows?

@gregnazario, do we have a process for this? e.g., a CI/CD job/label?

ibalajiarun (approved):
Docker builder changes LGTM

PR #9264, opened 2023-07-24
Use 8 runners for replay-verify

areshand:
### Description
increase runner of runner to speed up replay-verify

### Test Plan
existing tests


sionescu:
Can you add a short explanation as to why we're going back to 8 runners ?

areshand:
> Can you add a short explanation as to why we're going back to 8 runners ?

added

sionescu (approved):

perryjrandall (approved):
This test plan is not enough but accpeting to unblock

PR #9267, opened 2023-07-24
additional partitioner timers

zjma:


execution/block-partitioner/src/sharded_block_partitioner/mod.rs
```diff
@@ -324,15 +325,26 @@ impl ShardedBlockPartitioner {
         }
 
         // First round, we filter all transactions with cross-shard dependencies
+        let timer = BLOCK_PARTITIONING_MISC_TIMERS_SECONDS
+            .with_label_values(&["partition_by_senders"])
+            .start_timer();
         let mut txns_to_partition = self.partition_by_senders(transactions);
+        let _duration = timer.stop_and_record();
```
sitalkedia, line 332: nit: I think if you just drop the timer, it will automatically take care of recording the timings.

sitalkedia (approved):

execution/block-partitioner/src/sharded_block_partitioner/mod.rs
zjma: I'm not sure what's the order of dropping things and some drop can be expensive, so explicitly stopping as soon as possible.

zjma: and keeping the duration is just for easy debugging :)

manudhundi (approved):

PR #9289, opened 2023-07-24
Olsen/pangu sdk

olsenbudanur:
### Description
Creating a really light Rust wrapper around the python Pangu CLI for using Pangu from Rust.
### Test Plan
All the commands function locally through the tests in the sdk lib.


testsuite/pangu-sdk/src/lib.rs
```diff
@@ -0,0 +1,829 @@
...
+    fn pangu_directory() -> String {
+        if let Ok(pangu_dir) = env::var("PANGU_DIR") {
+            pangu_dir
+        } else {
+            panic!("Failed to retrieve PANGU_DIR environment variable.");
+        }
+    }
+}
+
+#[cfg(test)]
```
rustielin, line 563: Since these wont work out of the box in an isolated setup, you can throw these behind a feature flag. I did something similar for the `indexer-grpc` integration tests, which require docker. https://github.com/aptos-labs/aptos-core/blob/main/ecosystem/indexer-grpc/indexer-grpc-integration-tests/src/lib.rs#L4-L7 By throwing it behind a feature flag, CI will not pick it up by default

You can then run them via: `cargo nextest run --features integration-tests --package aptos-pangu-sdk`

testsuite/pangu-sdk/src/lib.rs
```diff
@@ -0,0 +1,829 @@
...
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_create_testnet() {
+        env::set_var(
+            "PANGU_DIR",
+            "/Users/olsenbudanur/Desktop/aptos-repos/aptos-core/testsuite",
```
rustielin, line 571: maybe just assert that `PANGU_DIR` env var is set? or derive it as a relative path from this dir? e.g `../../`

PR #9290, opened 2023-07-24
[Aptos Node] Bump aptos-node crate to 1.6

JoshLind:
### Description
This PR bumps the `aptos-node` crate version to 1.6.

### Test Plan
Existing test infrastructure.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2378,7 +2378,7 @@ dependencies = [
 
 [[package]]
 name = "aptos-node"
-version = "1.5.0"
+version = "1.6.0"
 dependencies = [
  "anyhow",
  "aptos-api",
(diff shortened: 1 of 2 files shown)


perryjrandall (approved):

davidiw:
woops, this is a different branch lol

davidiw (approved):

JoshLind:
> but this really doesn't make sense... one should be like 1.6 and the other 1.7, no? or at least a 1.6.0 and a 1.6.1

Yeah, this is all a bit adhoc at the moment and ideally we find a way to automate this. Until then, I think it's probably worth only bumping the version on each major release (otherwise, minor releases will be painful -- we're already forgetting to do this 😢).

Regarding `1.6.0` vs `1.7.0` , I can update the `main` branch to be `1.7.0`, but we typically only bump after each cut, so it feels a little odd. Either way, if folks have a more automated and/or scheduled way of doing this, I'm game 😄

perryjrandall:
The best way might be to set this to some non version in between and have a test on each release branch that fails if the version isnt set to the tag

PR #9295, opened 2023-07-25
[rb] Generic Reliable Broadcast module

ibalajiarun:
### Description

This PR extracts the reliable broadcast sender implementation out of consensus and makes it its own crate. The idea is to make this generic and modular such that various applications can use it when required.

### Test Plan

Unit tests


danielxiangzl (approved):
LGTM. Thank you boss!

zekun000 (approved):

PR #9297, opened 2023-07-25
Memoize account modules

styner32:
### Description
In typescript sdk, when [generate a transaction](https://github.com/aptos-labs/aptos-core/blob/main/ecosystem/typescript/sdk/src/providers/aptos_client.ts#L303), it fetches [account modules](https://github.com/aptos-labs/aptos-core/blob/aptos-node-v1.5.1/ecosystem/typescript/sdk/src/transaction_builder/builder.ts#L344), but the memoization is on the instance level, so it will not reused for the next transaction.

Adding the memoization on client level, so it can be shared in the same client.

### Test Plan
Add a unit test.


ecosystem/typescript/sdk/src/tests/unit/account_module.test.ts
```diff
@@ -0,0 +1,68 @@
+import { AptosClient } from "../../providers";
+import { longTestTimeout, NODE_URL } from "./test_helper.test";
+import * as aptosClient from "../../client";
+import { MoveFunctionVisibility } from "../../generated";
+
+jest.mock("../../client", () => {
+  return {
+    __esModule: true,
```
0xmaayan, line 8: is this needed?

ecosystem/typescript/sdk/src/tests/unit/account_module.test.ts
```diff
@@ -0,0 +1,68 @@
...
+        statusText: "OK",
+        data: accountModulesResponse,
+        url: "/fullnode/v1/accounts/0x1/modules",
+        headers: {},
+      });
+
+      const accountModules = await client.getAccountModules(moduleAddress);
+      expect(accountModules).toEqual(accountModulesResponse);
+
+      expect(getSpy).toBeCalledTimes(1);
```
0xmaayan, line 57: this can be removed? as we only need to expect is called one time once

ecosystem/typescript/sdk/src/tests/unit/account_module.test.ts
styner32: Yes. it will complain if I don't add it

e.g.
```
TypeError: Cannot redefine property: get
        at Function.defineProperty (<anonymous>)

      42 |       ];
      43 |
    > 44 |       const getSpy = jest.spyOn(aptosClient, "get").mockResolvedValue({
```

I wanted to mock http calls with [nock](https://github.com/nock/nock), but didn't want to introduce a new dependency for this simple change.

styner32:
thanks for reviewing. addressed all comments!

0xmaayan (approved):

ecosystem/typescript/sdk/src/providers/aptos_client.ts
```diff
@@ -135,6 +136,7 @@ export class AptosClient {
    * which is JSON representation of a module
```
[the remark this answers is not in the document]
styner32: done!

PR #9301, opened 2023-07-25
Add native functions for Aggregator V2

vusirikala:
### Description

Add `create_aggregator`, `try_add`, `try_sub`, `read`, `destroy` native functions for Aggregator V2.


aptos-move/framework/src/natives/aggregator_natives/aggregator_v2.rs
```diff
@@ -0,0 +1,169 @@
...
+
+    // Extract information from aggregator struct reference.
+    let (id, _limit) = aggregator_info(&safely_pop_arg!(args, StructRef))?;
+
+    // Get aggregator data.
+    let aggregator_context = context.extensions().get::<NativeAggregatorContext>();
+    let mut aggregator_data = aggregator_context.aggregator_data.borrow_mut();
+
+    // Actually remove the aggregator.
+    aggregator_data.remove_aggregator(id);
```
igor-aptos, line 145: just to check, will this work when AggregatorSnapshot is in the picture? i.e. do we need anything from the aggregator that was cloned from?

igor-aptos (approved):

aptos-move/e2e-move-tests/src/aggregator_v2.rs
```diff
@@ -0,0 +1,193 @@
+// Copyright © Aptos Foundation
```
georgemitenkov, line 1: I think we can reduce duplication between v1 and v2, can be done later in other PR though

aptos-move/aptos-aggregator/src/aggregator_extension.rs
```diff
@@ -239,6 +243,14 @@ impl Aggregator {
         Ok(())
     }
 
+    pub fn try_add(&mut self, value: u128) -> PartialVMResult<()> {
```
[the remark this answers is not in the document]
igor-aptos: should we rename add/sub to try_add/try_sub thought? :)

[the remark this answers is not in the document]
georgemitenkov: Or just this, yes

aptos-move/framework/src/natives/aggregator_natives/helpers_v2.rs
```diff
@@ -0,0 +1,35 @@
...
+    let field_ref = aggregator.borrow_field(index)?.value_as::<Reference>()?;
+    field_ref.read_ref()
+}
+
+/// Returns ID and a limit of aggrgegator based on a reference to `Aggregator` Move struct.
+pub(crate) fn aggregator_info(aggregator: &StructRef) -> PartialVMResult<(AggregatorID, u128)> {
+    let (value, limit) = get_aggregator_fields(aggregator)?;
+    assert!(
+        value & 0xFFFFFFFFFFFFFFFF == 0,
+        "value in aggregator exceeds u64::MAX"
```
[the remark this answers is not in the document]
vusirikala: Done.

aptos-move/framework/src/natives/aggregator_natives/helpers_v2.rs
```diff
@@ -0,0 +1,35 @@
...
+    );
+    Ok((AggregatorID::ephemeral(value as u64), limit))
+}
+
+/// Given a reference to `Aggregator` Move struct, returns a tuple of its
+/// fields: (`handle`, `key`, `limit`).
```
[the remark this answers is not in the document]
vusirikala: Done.

aptos-move/framework/src/natives/aggregator_natives/helpers_v2.rs
```diff
@@ -0,0 +1,35 @@
...
+pub(crate) fn get_aggregator_field(aggregator: &StructRef, index: usize) -> PartialVMResult<Value> {
+    let field_ref = aggregator.borrow_field(index)?.value_as::<Reference>()?;
+    field_ref.read_ref()
+}
+
+/// Returns ID and a limit of aggrgegator based on a reference to `Aggregator` Move struct.
+pub(crate) fn aggregator_info(aggregator: &StructRef) -> PartialVMResult<(AggregatorID, u128)> {
+    let (value, limit) = get_aggregator_fields(aggregator)?;
+    assert!(
+        value & 0xFFFFFFFFFFFFFFFF == 0,
```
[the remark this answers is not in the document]
vusirikala: No specific reason. Could that too.

aptos-move/framework/src/natives/aggregator_natives/aggregator_v2.rs
```diff
@@ -0,0 +1,169 @@
...
+fn native_create_aggregator(
+    context: &mut SafeNativeContext,
+    _ty_args: Vec<Type>,
+    mut args: VecDeque<Value>,
+) -> SafeNativeResult<SmallVec<[Value; 1]>> {
+    debug_assert_eq!(args.len(), 1);
+
+    context.charge(AGGREGATOR_V2_CREATE_AGGREGATOR_BASE)?;
+
+    // Extract fields: `limit` of the new aggregator and a `phantom_handle` of
```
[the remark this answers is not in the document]
vusirikala: Done.

aptos-move/framework/src/natives/aggregator_natives/aggregator_v2.rs
```diff
@@ -0,0 +1,169 @@
...
+
+    // Get the current aggregator data.
+    let aggregator_context = context.extensions().get::<NativeAggregatorContext>();
+    let mut aggregator_data = aggregator_context.aggregator_data.borrow_mut();
+
+    let id = AggregatorID::ephemeral(aggregator_data.generate_id());
+    aggregator_data.create_new_aggregator(id, limit);
+
+    Ok(smallvec![Value::struct_(Struct::pack(vec![
+        Value::u128(0),
```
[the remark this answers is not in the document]
vusirikala: Won't the value of the newly generated aggregator be 0? We replace the value with identifier at the end of the transaction before storing in MVHashmap right?

aptos-move/framework/aptos-framework/sources/aggregator_v2/aggregator_v2.move
```diff
@@ -33,6 +34,9 @@ module aptos_framework::aggregator_v2 {
 
     public native fun deferred_read(aggregator: &Aggregator): AggregatorSnapshot<u128>;
```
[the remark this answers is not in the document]
vusirikala: It's not implemented in this PR. I will implement aggregator snapshots stuff in another PR.

aptos-move/aptos-aggregator/src/aggregator_extension.rs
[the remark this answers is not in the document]
vusirikala: Yeah. I didn't know how try_add and add would be different. So, for this PR I didn't focus on the implementation of try_add. I just added the native functions. 
I can replace the existing `add` and `sub` with `try_add` and `try_sub` if you think that's more appropriate.

aptos-move/aptos-gas/src/aptos_framework.rs
```diff
@@ -0,0 +1,206 @@
+// Copyright © Aptos Foundation
```
gelash, line 1: What's happening with this file? I see the gas-related changes above, but this seems like a new file (and somewhat maybe auto-generated)? should it be like this or is it some rebase issue?

aptos-move/e2e-move-tests/src/aggregator_v2.rs
```diff
@@ -0,0 +1,184 @@
...
+        str::parse("0x1::aggregator_v2_test::try_sub").unwrap(),
+        vec![],
+        vec![
+            bcs::to_bytes(&index).unwrap(),
+            bcs::to_bytes(&value).unwrap(),
+        ],
+    )
+}
+
+pub fn try_sub_add(
```
gelash, line 90: What is this method?

gelash (commented):
Let's rebase it on aggregator v2 branch?

aptos-move/aptos-gas/src/aptos_framework.rs
vusirikala: I think this file was removed by Victor in the main branch. And for some reason, the file still existed on this branch. I removed it now.

aptos-move/e2e-move-tests/src/aggregator_v2.rs
vusirikala: This is just for the e2e move tests. The transaction first subtracts and then adds some value.

aptos-move/aptos-aggregator/src/aggregator_extension.rs
```diff
@@ -301,9 +305,18 @@ pub struct AggregatorData {
     destroyed_aggregators: BTreeSet<AggregatorID>,
     // All aggregator instances that exist in the current transaction.
     aggregators: BTreeMap<AggregatorID, Aggregator>,
+    // Counter for generating identifiers for AggregatorSnapshots.
```
georgemitenkov, line 308: "... and Aggregators."?

aptos-move/aptos-aggregator/src/delta_change_set.rs
```diff
@@ -553,7 +553,6 @@ mod test {
             })
         );
     }
-
```
georgemitenkov, line 556: Removing this change will save a file :)

aptos-move/aptos-aggregator/src/resolver.rs
```diff
@@ -73,10 +73,6 @@ pub mod test_utils {
             Ok(self.0.get(state_key).cloned())
         }
 
-        fn is_genesis(&self) -> bool {
```
georgemitenkov, line 76: I think you need to rebase? `is_genesis` has been removed.

aptos-move/e2e-move-tests/src/aggregator_v2.rs
georgemitenkov: @vusirikala let's reduce duplication here, e.g. `initialize` function can be shared with the initial aggregator file, otherwise it is hard to sync changes in the future, see my comment for `initialize`.

aptos-move/e2e-move-tests/src/aggregator_v2.rs
```diff
@@ -0,0 +1,184 @@
+// Copyright © Aptos Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::{assert_success, harness::MoveHarness};
+use aptos_language_e2e_tests::account::Account;
+use aptos_types::{account_address::AccountAddress, transaction::SignedTransaction};
+use std::path::PathBuf;
+
+pub fn initialize(path: PathBuf) -> (MoveHarness, Account) {
+    let mut harness = MoveHarness::new();
```
georgemitenkov, line 10: If we rebase on main, parallel execution is disabled in e2e tests by default. I have fixed that for the old aggregator file in main. But if have 2 initialisations this becomes error prone, so having a common APIs seems better to me, at least for code initialisation.

aptos-move/e2e-move-tests/src/tests/aggregator_v2.rs
```diff
@@ -0,0 +1,158 @@
...
+    },
+    assert_abort, assert_success,
+    tests::common,
+    MoveHarness,
+};
+use aptos_language_e2e_tests::account::Account;
+use aptos_types::transaction::SignedTransaction;
+
+fn setup() -> (MoveHarness, Account) {
+    initialize(common::test_dir_path("aggregator.data/pack"))
```
georgemitenkov, line 17: Btw your set-up builds both V1 and V2 Move files for both V1 and V2 tests, so x2 times more than needed, I think... maybe having it in a separate package is better, or you can merge tests alternatively.

aptos-move/e2e-move-tests/src/tests/aggregator_v2.rs
```diff
@@ -0,0 +1,158 @@
...
+}
+
+#[test]
+fn test_aggregators_e2e() {
```
georgemitenkov, line 21: This is the same as old aggregator tests, let's not duplicate test code. If you want to use the different files for V1 and V2, let's put the common code somewhere, e.g. you can have an array of function pointers you want to test, or use macros, etc.

aptos-move/e2e-tests/src/executor.rs
```diff
@@ -805,7 +805,6 @@ impl FakeExecutor {
                     println!("Should error, but ignoring for now... {}", err);
                 }
             }
-
```
georgemitenkov, line 808: why this change?

aptos-move/framework/src/natives/aggregator_natives/context.rs
```diff
@@ -14,6 +14,8 @@ use std::{
     collections::{btree_map, BTreeMap},
 };
 
+pub type TxnIndex = u32;
```
georgemitenkov, line 17: Where is this used?

aptos-move/framework/src/natives/aggregator_natives/helpers_v2.rs
```diff
@@ -0,0 +1,51 @@
...
+    let field_ref = aggregator.borrow_field(index)?.value_as::<Reference>()?;
+    field_ref.read_ref()
+}
+
+/// Returns ID and a limit of aggrgegator based on a reference to `Aggregator` Move struct.
+pub(crate) fn aggregator_info_u128(
+    aggregator: &StructRef,
+) -> PartialVMResult<(AggregatorID, u128)> {
+    let (value, limit) = get_aggregator_fields_u128(aggregator)?;
+    assert!(
```
georgemitenkov, line 24: Assertion seems ok, but given it is used by natives I would rather have an `Err` here. @gelash what do you think?

aptos-move/framework/src/natives/aggregator_natives/helpers_v2.rs
```diff
@@ -0,0 +1,51 @@
...
+
+/// Returns ID and a limit of aggrgegator based on a reference to `Aggregator` Move struct.
+pub(crate) fn aggregator_info_u64(aggregator: &StructRef) -> PartialVMResult<(AggregatorID, u64)> {
+    let (value, limit) = get_aggregator_fields_u64(aggregator)?;
+    Ok((AggregatorID::ephemeral(value), limit))
+}
+
+/// Given a reference to `Aggregator` Move struct, returns a tuple of its
+/// fields: (`value`, `limit`).
+pub fn get_aggregator_fields_u128(aggregator: &StructRef) -> PartialVMResult<(u128, u128)> {
```
georgemitenkov, line 39: This seems identical to `u64` case. Can we avoid it by making generic or using macros?

aptos-move/framework/src/natives/aggregator_natives/aggregator_v2.rs
```diff
@@ -0,0 +1,202 @@
...
+    let aggregator_context = context.extensions().get::<NativeAggregatorContext>();
+    let mut aggregator_data = aggregator_context.aggregator_data.borrow_mut();
+    let id = AggregatorID::ephemeral(aggregator_data.generate_id());
+
+    match ty_args[0] {
+        Type::U128 => {
+            let limit = safely_pop_arg!(args, u128);
+            aggregator_data.create_new_aggregator(id, limit);
+            Ok(smallvec![Value::struct_(Struct::pack(vec![
+                Value::u128(0),
```
georgemitenkov, line 47: I don't think this is correct: why we write 0 there? If I call `add` on this `StructRef` your code would interpret zero as an identifier.

aptos-move/framework/src/natives/aggregator_natives/aggregator_v2.rs
```diff
@@ -0,0 +1,202 @@
...
+        },
+        Type::U64 => {
+            let limit = safely_pop_arg!(args, u64);
+            aggregator_data.create_new_aggregator(id, limit as u128);
+            Ok(smallvec![Value::struct_(Struct::pack(vec![
+                Value::u64(0),
+                Value::u64(limit),
+            ]))])
+        },
+        _ => Err(PartialVMError::new(StatusCode::ABORTED)
```
georgemitenkov, line 59: Let's use `SafeNativeError`. I know that I myself put `PartialVMError`s in aggregator but it was before safe errors appeared and also I was stupid :) We shouldn't do that haha

aptos-move/framework/src/natives/aggregator_natives/aggregator_v2.rs
```diff
@@ -0,0 +1,202 @@
...
+            Ok(smallvec![Value::struct_(Struct::pack(vec![
+                Value::u128(0),
+                Value::u128(limit),
+            ]))])
+        },
+        Type::U64 => {
+            let limit = safely_pop_arg!(args, u64);
+            aggregator_data.create_new_aggregator(id, limit as u128);
+            Ok(smallvec![Value::struct_(Struct::pack(vec![
+                Value::u64(0),
```
georgemitenkov, line 55: Same as above

aptos-move/framework/src/natives/aggregator_natives/aggregator_v2.rs
```diff
@@ -0,0 +1,202 @@
...
+                Value::u64(limit),
+            ]))])
+        },
+        _ => Err(PartialVMError::new(StatusCode::ABORTED)
+            .with_message("Unsupported type supplied to aggregator".to_string())
+            .with_sub_status(0x02_0005)
```
georgemitenkov, line 61: Can we have this as a constant somewhere?

georgemitenkov: Also, do we have a corresponding error const in the Move file?

aptos-move/framework/src/natives/aggregator_natives/aggregator_v2.rs
```diff
@@ -0,0 +1,202 @@
...
+            // Get aggregator information and a value to add.
+            let value = safely_pop_arg!(args, u128);
+            let (id, limit) = aggregator_info_u128(&safely_pop_arg!(args, StructRef))?;
+            let aggregator = aggregator_data.get_aggregator(id, limit)?;
+            Ok(smallvec![Value::bool(aggregator.try_add(value).is_ok())])
+        },
+        Type::U64 => {
+            // Get aggregator information and a value to add.
+            let value = safely_pop_arg!(args, u64);
+            let (id, limit) = aggregator_info_u64(&safely_pop_arg!(args, StructRef))?;
```
georgemitenkov, line 91: If we return limit as u128, we can avoid casts, then we can factor out the code a lot, e.g. using a simple macro. Cause this seems identical?

aptos-move/framework/src/natives/aggregator_natives/aggregator_v2.rs
```diff
@@ -0,0 +1,202 @@
...
+            Ok(smallvec![Value::u128(value)])
+        },
+        Type::U64 => {
+            let (id, limit) = aggregator_info_u64(&safely_pop_arg!(args, StructRef))?;
+            let aggregator = aggregator_data.get_aggregator(id, limit as u128)?;
+            let value = aggregator.read_and_materialize(aggregator_context.resolver, &id)?;
+            if value > u64::MAX as u128 {
+                return Err(PartialVMError::new(StatusCode::ABORTED)
+                    .with_message("Aggregator<u64>::read() output exceeds u64::MAX".to_string())
+                    .with_sub_status(0x02_0001)
```
georgemitenkov, line 172: What is `0x02_0001`? Can we make this a constant? Does it exists in the Move file?

aptos-move/framework/src/natives/aggregator_natives/aggregator_v2.rs
```diff
@@ -0,0 +1,202 @@
...
+                    .into());
+            }
+            Ok(smallvec![Value::u64(value as u64)])
+        },
+        _ => Err(PartialVMError::new(StatusCode::ABORTED)
```
georgemitenkov, line 177: This last case is repeated a lot :) Can we avoid that?

georgemitenkov (changes requested):

aptos-move/framework/src/natives/aggregator_natives/aggregator_v2.rs
georgemitenkov: Can we have tests for this behaviour? e.g. I create, and add and check the result.

aptos-move/framework/src/natives/aggregator_natives/aggregator_v2.rs
igor-aptos: Aggregator is now a generic struct as well, and limit is also u64 inside. 

that makes the API to creating snapshots cleaner, and the only place we need to deal with casting is here. 

this pattern repeats below as well, we can either do a match to just give us a limit in u128:

```
let (value, limit) = match ty_args[0] { 
   ...
};
let aggregator = aggregator_data.get_aggregator(id, limit)?;
Ok(smallvec![Value::bool(aggregator.try_add(value).is_ok())])
```

that might be a bit cleaner.

But as it is repeated across functions, maybe we should just have utility functions pop_arg_into_u128 and pop_arg_aggregator_into_limit_u128:

```
let value = pop_arg_into_u128(args, ty_args[0]);
let limit = pop_arg_aggregator_into_limit_u128(args, ty_args[0]);
let aggregator = aggregator_data.get_aggregator(id, limit)?;
Ok(smallvec![Value::bool(aggregator.try_add(value).is_ok())])

```

aptos-move/framework/src/natives/aggregator_natives/helpers_v2.rs
igor-aptos: should just

```
  pub fn get_aggregator_fields<T>(aggregator: &StructRef) -> PartialVMResult<(T, T)> {
    let value = get_aggregator_field(aggregator, VALUE_FIELD_INDEX)?.value_as::<T>()?;
    let limit = get_aggregator_field(aggregator, LIMIT_FIELD_INDEX)?.value_as::<T>()?;
    Ok((value, limit))
  }
  
```
work?

what does value_as do if it is given a type different from the one stored? does it convert it/cast it, or fails, or?

igor-aptos (changes requested):

aptos-move/framework/src/natives/aggregator_natives/aggregator_v2.rs
vusirikala: Done.

aptos-move/aptos-aggregator/src/resolver.rs
vusirikala: Done.

aptos-move/aptos-aggregator/src/aggregator_extension.rs
vusirikala: Done.

aptos-move/e2e-move-tests/src/aggregator_v2.rs
vusirikala: @georgemitenkov aggregator.rs and aggregator_v2.rs may look the same. But aggregator.vs is initializing aggregator v1 and storing in AggregatorStore. aggregator_v2.rs is initializing aggregator from aggregator v2 module and storing in AggregatorStore. They work differently.

aptos-move/e2e-move-tests/src/aggregator_v2.rs
vusirikala: I applied that parallel execution change to aggregator_v2 as well. The initialize functions in aggregator.vs and aggregator_v2.rs run different entry functions (0x1::aggregator_test::initialize, 0x1::aggregator_v2_test::initialize) and they use different feature flags. So, it would get messy to combine them.

aptos-move/e2e-tests/src/executor.rs
vusirikala: May be rust lint.

aptos-move/e2e-move-tests/src/tests/aggregator_v2.rs
vusirikala: I agree the code looks unfortunately duplicate. But the function they call are actually different (aggregator.rs calls functions from aggregator module and aggregator_v2.rs calls functions from aggregator_v2 module). Using function pointers and macros for all of these seems to make the code less readable. We will deprecate the aggregator v1 code soon anyways right?

aptos-move/framework/src/natives/aggregator_natives/helpers_v2.rs
vusirikala: Implemented your code just now. It gives the following error, so I reverted back to the old code. 
```
the trait `move_vm_types::values::VMValueCast<T>` is not implemented for `move_vm_types::values::Value`
```

aptos-move/framework/src/natives/aggregator_natives/aggregator_v2.rs
vusirikala: Done.

aptos-move/framework/src/natives/aggregator_natives/aggregator_v2.rs
vusirikala: Done.

aptos-move/framework/src/natives/aggregator_natives/aggregator_v2.rs
vusirikala: With a match statement on an enum, you need to cover all the variants. The value can take several types. So, we do need an `_ => {something}` code.

aptos-move/framework/src/natives/aggregator_natives/helpers_v2.rs
vusirikala: I added a new error code 0x2_0008 and returning this error code for this.

aptos-move/framework/src/natives/aggregator_natives/aggregator_v2.rs
vusirikala: > Can we have tests for this behaviour? e.g. I create, and add and check the result.
Unfortunately, end to end tests still don't work at this point as there are missing blocks in our code (for example, materialization is still not done for aggregator v2).

aptos-move/framework/src/natives/aggregator_natives/aggregator_v2.rs
vusirikala: I tried to used your ideas and abstract out the common code in try_add, try_sub and put it in a separate function. Please let me know if this looks good.

aptos-move/framework/src/natives/aggregator_natives/aggregator_v2.rs
vusirikala: Done.

aptos-move/framework/src/natives/aggregator_natives/aggregator_v2.rs
vusirikala: > yeah, I assume we need to put here "id", but also create an aggregator with initial value 0 (i.e. in Data state) in the context? I see that you are only creating id, but not putting it in the map?

Done.

aptos-move/framework/aptos-framework/sources/aggregator_v2/aggregator_v2.move
```diff
@@ -9,16 +9,21 @@ module aptos_framework::aggregator_v2 {
...
     const EAGGREGATOR_SNAPSHOTS_NOT_ENABLED: u64 = 6;
 
+    /// The generic type supplied to the aggregator is not supported.
+    const EUNSUPPORTED_AGGREGATOR_TYPE: u64 = 7;
+
+    /// The aggregator identifier is out of bounds (above U64::MAX).
+    const EINVALID_AGGREGATOR_IDENTIFIER: u64 = 8;
+
+    /// The aggregator<u64> read is out of bounds (above U64::MAX).
+    const EREAD_OUT_OF_BOUNDS: u64 = 9;
```
igor-aptos, line 25: these are not something users should ever see. not sure if they should go here, but for now let's just add a comment: "should never happen, indicates issue in the internal implementation" or something

aptos-move/framework/src/natives/aggregator_natives/aggregator_v2.rs
```diff
@@ -39,6 +50,148 @@ fn is_string_type(context: &SafeNativeContext, type_arg: &Type) -> SafeNativeRes
     Ok(false)
 }
 
+/// Given the list of native function arguments, returns a tuple of its
+/// fields: (`aggregator id`, `value`, `limit`).
+pub fn get_aggregator_fields(
```
igor-aptos, line 55: it is a bit awkward to take both arguments together, what if we have a different function with different arguments. i.e. having pair of "pop_value_of_type" and "pop_aggregator_fields_of_type" seems better.

but if you want to keep it as is, you need to make it clear in the comment what it is:

/// Given the list of two arguments (value: T, agg: Aggregator<T>) to a native function, returns a tuple of its fields: (`value`, `aggregator id`, `limit`).

(and I would definitely keep the order of what is returned same as what is passed (i.e. value going before aggregator fields

aptos-move/framework/src/natives/aggregator_natives/aggregator_v2.rs
```diff
@@ -39,6 +50,148 @@ fn is_string_type(context: &SafeNativeContext, type_arg: &Type) -> SafeNativeRes
...
+            ]))])
+        },
+        Type::U64 => {
+            let limit = safely_pop_arg!(args, u64);
+            aggregator_data.create_new_aggregator(aggregator_id, limit as u128);
+            Ok(smallvec![Value::struct_(Struct::pack(vec![
+                Value::u64(id),
+                Value::u64(limit),
+            ]))])
+        },
```
igor-aptos, line 112: if you had pop_value_of_type function separately, as I suggested above, this can become:

```
    let limit = pop_value_of_type(args, ty_args[0]);
    aggregator_data.create_new_aggregator(aggregator_id, limit);
    match ty_args[0] {
        Type::U128 => Ok(smallvec![Value::struct_(Struct::pack(vec![
            Value::u128(id as u128),
            Value::u128(limit),
        ]))]),
        Type::U64 => Ok(smallvec![Value::struct_(Struct::pack(vec![
             Value::u64(id),
             Value::u64(limit as u64),
        ]))]),
        _ => Err(SafeNativeError::Abort {
            abort_code: EUNSUPPORTED_AGGREGATOR_TYPE,
        }),
    }

```

aptos-move/framework/src/natives/aggregator_natives/aggregator_v2.rs
```diff
@@ -39,6 +50,148 @@ fn is_string_type(context: &SafeNativeContext, type_arg: &Type) -> SafeNativeRes
...
+                return Err(SafeNativeError::Abort {
+                    abort_code: EREAD_OUT_OF_BOUNDS,
+                });
+            }
+            Ok(smallvec![Value::u64(value as u64)])
+        },
+        _ => Err(SafeNativeError::Abort {
+            abort_code: EUNSUPPORTED_AGGREGATOR_TYPE,
+        }),
+    }
```
igor-aptos, line 192:
```
    // Extract information from aggregator struct reference.
    let (id, limit) = pop_aggregator_fields_of_type(args, ty_args[0])?;
    let aggregator = aggregator_data.get_aggregator(id, limit)?;
    let value = aggregator.read_and_materialize(aggregator_context.resolver, &id)?;
           
    if value > limit {
        return Err(SafeNativeError::Abort {
             abort_code: EREAD_OUT_OF_BOUNDS,
        });
    }

    match ty_args[0] {
        Type::U128 => {
            Ok(smallvec![Value::u128(value)])
        },
        Type::U64 => {
            // this should never happen, but since we check the limit above - this should really never happen
            if value > u64::MAX as u128 {
                return Err(SafeNativeError::Abort {
                    abort_code: EREAD_OUT_OF_BOUNDS,
                });
            }
            Ok(smallvec![Value::u64(value as u64)])
        },
        _ => Err(SafeNativeError::Abort {
            abort_code: EUNSUPPORTED_AGGREGATOR_TYPE,
        }),
    }
```

igor-aptos (approved):
looks good. minor style comments/comments on what seems cleaner to me.

you can decide if you prefer how it is now, or if you like my suggestions - and land.

aptos-move/e2e-move-tests/src/tests/aggregator_v2.data/pack/sources/aggregator_v2_test.move
```diff
@@ -0,0 +1,126 @@
...
+    }
+    
+    /// Initializes a fake resource which holds aggregators.
+    public entry fun initialize(account: &signer) {
+        let aggregators = table::new();
+        let store = AggregatorStore { aggregators };
+        move_to(account, store);
+    }
+
+    public entry fun verify_copy_snapshot() {
```
georgemitenkov, line 22: Seems like this and other `verify_` functions are never called?

aptos-move/e2e-move-tests/src/tests/aggregator_v2.rs
georgemitenkov: Hmm, maybe. I do not like copying cause this way we can copy some mistakes done in V1 and then we would need to fix both😂. Let's add at least a TODO?

aptos-move/framework/aptos-framework/sources/aggregator_v2/aggregator_v2.move
```diff
@@ -9,16 +9,21 @@ module aptos_framework::aggregator_v2 {
     /// The value of aggregator underflows (goes below zero). Raised by uncoditional sub() call
     const EAGGREGATOR_UNDERFLOW: u64 = 2;
 
-    /// Element type of the aggregator not supported. Raised by create_aggregator() call,
-    /// if Element type is not supported (currently u64 or u128).
-    const EAGGREGATOR_ELEMENT_TYPE_NOT_SUPPORTED: u64 = 3;
-
     /// The generic type supplied to the aggregator snapshot is not supported.
     const EUNSUPPORTED_AGGREGATOR_SNAPSHOT_TYPE: u64 = 5;
```
georgemitenkov, line 13: Why 5?

aptos-move/framework/aptos-framework/sources/aggregator_v2/aggregator_v2.move
georgemitenkov: @igor-aptos @vusirikala I think the comment here for `EREAD_OUT_OF_BOUNDS` is a bit misleading? We use it for converting `aggregator<u128>` --> `snapshot<u64>`, no? Then it is a cast error and can be user-facing. Why `EREAD_OUT_OF_BOUNDS` would ever happen for `aggregator<u64>`?

aptos-move/framework/src/natives/aggregator_natives/aggregator_v2.rs
```diff
@@ -39,6 +50,148 @@ fn is_string_type(context: &SafeNativeContext, type_arg: &Type) -> SafeNativeRes
...
+            let value = aggregator.read_and_materialize(aggregator_context.resolver, &id)?;
+            Ok(smallvec![Value::u128(value)])
+        },
+        Type::U64 => {
+            let (id, limit) = aggregator_value_as_u64(&safely_pop_arg!(args, StructRef))?;
+            let aggregator = aggregator_data.get_aggregator(id, limit as u128)?;
+            let value = aggregator.read_and_materialize(aggregator_context.resolver, &id)?;
+            if value > u64::MAX as u128 {
+                return Err(SafeNativeError::Abort {
+                    abort_code: EREAD_OUT_OF_BOUNDS,
```
georgemitenkov, line 184: You can use `SafeNativeError::InvariantViolation(...)` here.

aptos-move/framework/src/natives/aggregator_natives/helpers_v2.rs
```diff
@@ -1,11 +1,55 @@
...
+
+/// Returns ID and a limit of aggrgegator based on a reference to `Aggregator` Move struct.
+pub(crate) fn aggregator_value_as_u128(
+    aggregator: &StructRef,
+) -> PartialVMResult<(AggregatorID, u128)> {
+    let (value, limit) = get_aggregator_fields_u128(aggregator)?;
+    if value > u64::MAX as u128 {
+        return Err(PartialVMError::new(StatusCode::ABORTED)
+            .with_message("Aggregator identifier is too small".to_string())
+            .with_sub_status(EINVALID_AGGREGATOR_IDENTIFIER));
```
georgemitenkov, line 25: You can use `SafeNativeError::InvariantViolation(...)` here as well? Or simply invariant violation PartialVMResult.

georgemitenkov (approved):

aptos-move/framework/aptos-framework/sources/aggregator_v2/aggregator_v2.move
vusirikala: @georgemitenkov This error is not for casting. I raise this error during the normal expensive read of an Aggregator<u64> when the return value is above u64::MAX. Ideally, this error code should never be raised if our code runs correctly. As we generally prefer raising errors instead of making assertions, I raised this error.

aptos-move/e2e-move-tests/src/tests/aggregator_v2.data/pack/sources/aggregator_v2_test.move
vusirikala: They are called in the `test_copy_snapshot` test.

aptos-move/framework/aptos-framework/sources/aggregator_v2/aggregator_v2.move
vusirikala: I'm using 3 and 4 in the validate_history function when an overflow/underflow is raised in the original execution, but should not have been raised when the speculative_start_value is replaced with a new value.

georgemitenkov: Why validate_history would have a different error? An overflow is an overflow I no matter where it comes from I think or maybe I am missing something

aptos-move/framework/aptos-framework/sources/aggregator_v2/aggregator_v2.move
georgemitenkov: Yep yep, was confused at the start :) we should use InvariantViolation then for these things

aptos-move/e2e-move-tests/src/tests/aggregator_v2.data/pack/sources/aggregator_v2_test.move
georgemitenkov: Cannot see the test in the diff?

georgemitenkov: Found it :)

aptos-move/framework/src/natives/aggregator_natives/helpers_v2.rs
vusirikala: Done.

aptos-move/framework/src/natives/aggregator_natives/aggregator_v2.rs
vusirikala: Done.

aptos-move/framework/aptos-framework/sources/aggregator_v2/aggregator_v2.move
```diff
@@ -9,16 +9,21 @@ module aptos_framework::aggregator_v2 {
...
 
     /// The aggregator snapshots feature flag is not enabled.
     const EAGGREGATOR_SNAPSHOTS_NOT_ENABLED: u64 = 6;
 
+    /// The generic type supplied to the aggregator is not supported.
+    const EUNSUPPORTED_AGGREGATOR_TYPE: u64 = 7;
+
+    /// The aggregator identifier is out of bounds (above U64::MAX).
+    const EINVALID_AGGREGATOR_IDENTIFIER: u64 = 8;
```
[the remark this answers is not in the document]
vusirikala: Done.

aptos-move/framework/src/natives/aggregator_natives/aggregator_v2.rs
vusirikala: Done.

aptos-move/framework/src/natives/aggregator_natives/aggregator_v2.rs
vusirikala: Done.

aptos-move/framework/src/natives/aggregator_natives/aggregator_v2.rs
vusirikala: Done.

Issue #9150, opened 2023-07-18
[Feature Request] Get rid of the onchain configuration registry


sitalkedia:
# 🚀 Feature Request

 Get rid of the onchain configuration registry

## Motivation
We have a configuration registry - https://github.com/aptos-labs/aptos-core/blob/361c6e0a3835df9f11865043e6d45dba446b2ac1/types/src/on_chain_config/mod.rs#L82, that tracks all the onchain configs we register. On an epoch change, we read the registry and read all the config values from the registry - https://github.com/aptos-labs/aptos-core/blob/0f8b9d1c492696d9bcf885c03fe54160a8f66718/state-sync/inter-component/event-notifications/src/lib.rs#L268. This is converted to an `OnChainConfigPayload` - https://github.com/aptos-labs/aptos-core/blob/0f8b9d1c492696d9bcf885c03fe54160a8f66718/state-sync/inter-component/event-notifications/src/lib.rs#L312. 

This payload is used by the epoch manager to parse various configs - https://github.com/aptos-labs/aptos-core/blob/5065e5d65db680d4769141e3216b3c5ef835cad5/consensus/src/epoch_manager.rs#L804. If the configuration is not found, it also falls back to the default. 

The issue with maintaining a separate registry is that this registry can easily get out of the sync and can create on-chain config mismatch issue. In fact, there is no need for maintaining a registry at all. 

**Describe the solution you'd like**

- Instead of passing serialized bytes in the on-chain config payload in https://github.com/aptos-labs/aptos-core/blob/361c6e0a3835df9f11865043e6d45dba446b2ac1/types/src/on_chain_config/mod.rs#L93, we can pass the DB state view at a particular version - https://github.com/aptos-labs/aptos-core/blob/0f8b9d1c492696d9bcf885c03fe54160a8f66718/state-sync/inter-component/event-notifications/src/lib.rs#L284.

- On the epoch manager/mempool side, when we read the payload, we can use the db state view along with configuration path to read a particular configuration value and fallback to default otherwise. 

- [Option] If we see a particular configuration is being read repeatedly from storage, we can add a caching layer to the on chain config payload.
- We also probably need to rename the on chain config payload to something like onchain config view as it doesn't contain the actual payload.

PR #9303, opened 2023-07-25
[Configuration] Remove the need for maintaining an on-chain config registry

sitalkedia:
### Description

See https://github.com/aptos-labs/aptos-core/issues/9150 for more details

### Test Plan

Existing Unit tests.

Tested it by running `test_onchain_shuffling_change`, which toggles the shuffling flag and ensures that the txn shuffling actually takes place.

### DIFF
--- a/state-sync/inter-component/event-notifications/src/lib.rs
+++ b/state-sync/inter-component/event-notifications/src/lib.rs
@@ -3,12 +3,12 @@
 // SPDX-License-Identifier: Apache-2.0
 
 #![forbid(unsafe_code)]
-
+use anyhow::Result;
 use aptos_channels::{aptos_channel, message_queues::QueueStyle};
 use aptos_id_generator::{IdGenerator, U64IdGenerator};
 use aptos_infallible::RwLock;
 use aptos_state_view::account_with_state_view::AsAccountWithStateView;
-use aptos_storage_interface::{state_view::DbStateViewAtVersion, DbReaderWriter};
+use aptos_storage_interface::{state_view::DbStateViewAtVersion, DbReader, DbReaderWriter};
 use aptos_types::{
     account_config::CORE_CODE_ADDRESS,
     account_view::AccountView,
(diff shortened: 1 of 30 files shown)


zekun000 (commented):
looks reasonable to me, cc @JoshLind to take a look too

vm-validator/Cargo.toml
```diff
@@ -14,6 +14,7 @@ rust-version = { workspace = true }
 
 [dependencies]
 anyhow = { workspace = true }
+aptos-event-notifications = { workspace = true }
```
JoshLind, line 17: Silly question: do we need to add these? They look unused below, right?

state-sync/inter-component/event-notifications/src/lib.rs
```diff
@@ -391,6 +376,34 @@ impl ReconfigSubscription {
     }
 }
 
+#[derive(Clone)]
+pub struct DbBackedOnChainConfig {
```
JoshLind, line 380: Can we avoid putting this here? Feels like it would be better in `types/src/on_chain_config/mod.rs`, no?

Also, it would avoid the need for us to import `anyhow` into state sync (which I've tried to avoid because we have our own notion of error and result). 😄

JoshLind (commented):
Thanks @sitalkedia. LGTM -- only concern is the location of `DbBackedOnChainConfig`.

state-sync/inter-component/event-notifications/src/lib.rs
sitalkedia: @JoshLind - We can't put the `DbBackedOnChainConfig` in `types` crate as that will create circular dependency. This depends on DBReader, which depends on the `types`.  The `types` crate is a very thin crate with minimal dependency and probably we need to keep it that way..

JoshLind: Discussed offline -- my primary concern is that it feels odd for the on chain config logic to be split up. But, will unblock the PR for now and we can add a comment (e.g., for a nice onboarding task 😄).

JoshLind (approved):
Thanks @sitalkedia!

zekun000 (approved):

sitalkedia:
@zekun000 - Any thoughts on how we should go about testing it?

sitalkedia:
@JoshLind , @zekun000 - I tested this change by running the smoke test which toggles the shuffling flag and ensures it takes effect. Any other testing we need to do before landing this change?

sitalkedia:
Discussed with @JoshLind and @zekun000 offline, the current testing is good and we can land this change. But as a followup, we might want to add an endpoint to inspect the on-chain config to be able inspect it easily.

PR #9309, opened 2023-07-25
[Mempool] optimize fullnode broadcast hops for latency

bchocho:
### Description

* Increase single fullnode max throughput from 4K TPS to 6K TPS (max batch size 200 -> 300, scheduled every 50 ms)
* Increase throughput when broadcast RTT is large, by increasing the number of outstanding requests. E.g., previously an RTT of 500 ms with 2 outstanding requests, meant only 2 requests could be made every 500 ms.

### Test Plan

Run forge with PFNs and network emulation, observe that `Avg Insertion-to-Broadcast-Batched` on PFNs drops significantly, from 1-2 s for some PFNs to < 200 ms.

### DIFF
--- a/config/src/config/mempool_config.rs
+++ b/config/src/config/mempool_config.rs
@@ -60,11 +60,11 @@ impl Default for MempoolConfig {
         MempoolConfig {
             shared_mempool_tick_interval_ms: 50,
             shared_mempool_backoff_interval_ms: 30_000,
-            shared_mempool_batch_size: 200,
+            shared_mempool_batch_size: 300,
             shared_mempool_max_batch_bytes: MAX_APPLICATION_MESSAGE_SIZE as u64,
             shared_mempool_ack_timeout_ms: 2_000,
             shared_mempool_max_concurrent_inbound_syncs: 4,
-            max_broadcasts_per_peer: 2,
+            max_broadcasts_per_peer: 20,
             max_network_channel_size: 1024,
             mempool_snapshot_interval_secs: 180,
             capacity: 2_000_000,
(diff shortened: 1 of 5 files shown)


testsuite/forge-cli/src/main.rs
```diff
@@ -500,7 +500,7 @@ fn single_test_suite(
     let single_test_suite = match test_name {
         // Land-blocking tests to be run on every PR:
         "land_blocking" => land_blocking_test_suite(duration), // to remove land_blocking, superseeded by the below
-        "realistic_env_max_load" => realistic_env_max_load_test(duration, test_cmd, 7, 5),
+        "realistic_env_max_load" => pfn_performance(duration, true, true),
```
JoshLind, line 503: Does this need to be rolled back? 😄

JoshLind (approved):
Config changes are the best changes 😄

testsuite/forge-cli/src/main.rs
```diff
@@ -500,7 +500,7 @@ fn single_test_suite(
     let single_test_suite = match test_name {
         // Land-blocking tests to be run on every PR:
         "land_blocking" => land_blocking_test_suite(duration), // to remove land_blocking, superseeded by the below
-        "realistic_env_max_load" => realistic_env_max_load_test(duration, test_cmd, 7, 5),
+        "realistic_env_max_load" => pfn_const_tps(duration, true, true),
```
sitalkedia, line 503: You probably want to revert this change before landing?

config/src/config/mempool_config.rs
```diff
@@ -103,19 +103,25 @@ impl ConfigOptimizer for MempoolConfig {
 
         // Change the default configs for VFNs
         let mut modified_config = false;
+        if node_type.is_validator() {
+            // Set the max_broadcasts_per_peer to 2 (default is 20)
+            if local_mempool_config_yaml["max_broadcasts_per_peer"].is_null() {
```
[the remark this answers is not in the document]
bchocho: This isn't used anymore because there is no mempool broadcast between validators with Quorum Store. For the time being, I want to preserve behavior as much as possible just in case we rollback Quorum Store.

testsuite/forge-cli/src/main.rs
bchocho: Done!

bchocho: Oops, now really reverted.

config/src/config/mempool_config.rs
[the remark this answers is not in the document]
sitalkedia: Can you add a comment to clarify this?

sitalkedia (approved):
LGTM

PR #9310, opened 2023-07-25
[consensusdb][bug-fix] Add Dag Vote to column family

ibalajiarun:
### Description

This PR adds the Dag Vote schema to the column family so that the column is created if it doesn't exist.

### Test Plan

Added a unit test to catch the bug


ibalajiarun:
Duplicate of #9242

PR #9312, opened 2023-07-26
[dag] Handle certified node in DAG Driver

ibalajiarun:
### Description

This PR removes the `CertifiedNodeHandler` and instead implements the `RpcHandler` on DagDriver to handle `CertifiedNode`. 

Also, adds a method to trigger a new round from special case current round 0 from Dag Handler.

### Test Plan

Existing tests fixed.

### DIFF
--- a/consensus/src/dag/dag_driver.rs
+++ b/consensus/src/dag/dag_driver.rs
@@ -1,7 +1,11 @@
 // Copyright © Aptos Foundation
 // SPDX-License-Identifier: Apache-2.0
 
-use super::{storage::DAGStorage, types::CertifiedAck, RpcHandler};
+use super::{
+    storage::DAGStorage,
+    types::{CertifiedAck, DAGMessage},
+    RpcHandler,
+};
 use crate::{
     dag::{
         dag_store::Dag,
@@ -20,8 +24,8 @@ use futures::{
     FutureExt,
 };
 use std::sync::Arc;
-use tokio_retry::strategy::ExponentialBackoff;
 use thiserror::Error as ThisError;
+use tokio_retry::strategy::ExponentialBackoff;
 
 #[derive(Debug, ThisError)]
 pub enum DagDriverError {
@@ -67,9 +71,12 @@ impl DagDriver {
     }
 
     pub fn try_enter_new_round(&mut self) {
+        // In case of a new epoch, kickstart building the DAG by entering the next round
+        // without any parents.
         if self.current_round == 0 {
             self.enter_new_round(vec![]);
         }
+        // TODO: add logic to handle building DAG from the middle, etc.
     }
 
     pub fn add_node(&mut self, node: CertifiedNode) -> anyhow::Result<()> {
--- a/consensus/src/dag/dag_driver.rs
+++ b/consensus/src/dag/dag_driver.rs
@@ -66,6 +66,12 @@ impl DagDriver {
         }
     }
 
+    pub fn try_enter_new_round(&mut self) {
+        if self.current_round == 0 {
+            self.enter_new_round(vec![]);
+        }
+    }
+
     pub fn add_node(&mut self, node: CertifiedNode) -> anyhow::Result<()> {
         let mut dag_writer = self.dag.write();
         let round = node.metadata().round();
--- a/consensus/src/dag/dag_driver.rs
+++ b/consensus/src/dag/dag_driver.rs
@@ -1,7 +1,7 @@
 // Copyright © Aptos Foundation
 // SPDX-License-Identifier: Apache-2.0
 
-use super::{storage::DAGStorage, types::DAGMessage};
+use super::{storage::DAGStorage, types::CertifiedAck, RpcHandler};
 use crate::{
     dag::{
         dag_store::Dag,
@@ -10,6 +10,7 @@ use crate::{
     state_replication::PayloadClient,
     util::time_service::TimeService,
 };
+use anyhow::bail;
 use aptos_consensus_types::common::{Author, Payload};
 use aptos_infallible::RwLock;
 use aptos_reliable_broadcast::ReliableBroadcast;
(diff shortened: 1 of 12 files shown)


consensus/src/dag/dag_driver.rs
```diff
@@ -114,3 +131,22 @@ impl DagDriver {
...
+impl RpcHandler for DagDriver {
+    type Request = CertifiedNode;
+    type Response = CertifiedAck;
+
+    fn process(&mut self, node: Self::Request) -> anyhow::Result<Self::Response> {
+        let epoch = node.metadata().epoch();
+        {
+            let dag_reader = self.dag.read();
+            if dag_reader.exists(node.metadata()) {
+                return Ok(CertifiedAck::new(node.metadata().epoch()));
```
sasha8, line 144: node.metadata().epoch()) -> epoch

consensus/src/dag/dag_driver.rs
```diff
@@ -114,3 +131,22 @@ impl DagDriver {
...
+            }
+        }
+
+        self.add_node(node)?;
```
sasha8, line 148: I still don't really convinced that in case we do not have the parents we should rely on the sender to retransmit. Instead of storing it as pending. 
Remind me why we chose this path?

consensus/src/dag/dag_handler.rs
```diff
@@ -37,22 +38,40 @@ impl NetworkHandler {
...
                 epoch_state.clone(),
+                storage.clone(),
+            ),
+            dag_driver: DagDriver::new(
+                signer.author(),
+                epoch_state.clone(),
+                dag.clone(),
+                payload_client,
+                rb,
+                1,
```
sasha8, line 63: Why cannot you hard code this? Why would anyone call it with a round different than 1?
Moreover, you handle the 0 case, but what if someone calls it with 17? this can cause issues, right?

ibalajiarun: We can't start from 1 at all times because 1..16 can be GCed in all nodes and we can only start from round 17. This will change as we add state-sync logic and integrate to epoch manager.

consensus/src/dag/dag_driver.rs
ibalajiarun: We will send the node to dag fetcher, so we are not dropping it.

sasha8: They we should send back Ack, right?

ibalajiarun: One reason is that it might be misleading because its not actually in the DAG (not sure if it matters much) 
Another reason is, in DAG fetcher, we also plan to try to fetch once from all nodes and give up eventually, so we might need retransmission.

consensus/src/dag/dag_handler.rs
sasha8: I see, good point. But in this case you should not check parents for nodes in round 17.

consensus/src/dag/dag_driver.rs
```diff
@@ -57,21 +65,30 @@ impl DagDriver {
         }
     }
 
+    pub fn try_enter_new_round(&mut self) {
```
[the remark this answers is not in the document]
zekun000: i assume this is mainly for jumpstart, we need to be more general than this for restart but probably not in this pr

consensus/src/dag/dag_driver.rs
zekun000: we should only send back Ack if it's successfully added to the DAG, if the fetch failed we rely on the sender to retransmit for retry

consensus/src/dag/dag_driver.rs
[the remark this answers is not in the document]
ibalajiarun: yes, that is correct.

consensus/src/dag/types.rs
```diff
@@ -250,8 +250,8 @@ impl TDAGMessage for Node {
 
         let current_round = self.metadata().round();
 
-        if current_round == 0 {
-            ensure!(self.parents().is_empty(), "invalid parents for round 0");
+        if current_round == 1 {
```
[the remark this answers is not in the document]
zekun000: we need to verify the round is not 0 or >=1 otherwise the prev_round = current_round -1 would just underflow

consensus/src/dag/dag_handler.rs
ibalajiarun: Yes, I will add a TODO

consensus/src/dag/dag_handler.rs
```diff
@@ -37,22 +41,44 @@ impl NetworkHandler {
         signer: ValidatorSigner,
         epoch_state: Arc<EpochState>,
         storage: Arc<dyn DAGStorage>,
+        payload_client: Arc<dyn PayloadClient>,
+        _dag_network_sender: Arc<dyn DAGNetworkSender>,
+        rb_network_sender: Arc<dyn RBNetworkSender<DAGMessage>>,
+        time_service: Arc<dyn TimeService>,
+        aptos_time_service: aptos_time_service::TimeService,
```
ibalajiarun, line 48: Having two time services is ugly :(

ibalajiarun:
* main
  * **PR #9312** <a href="https://app.graphite.dev/github/pr/aptos-labs/aptos-core/9312" target="_blank">Graphite</a>  👈
          * **PR #9455** <a href="https://app.graphite.dev/github/pr/aptos-labs/aptos-core/9455" target="_blank">Graphite</a> 
            * **PR #9456** <a href="https://app.graphite.dev/github/pr/aptos-labs/aptos-core/9456" target="_blank">Graphite</a> 
              * **PR #9457** <a href="https://app.graphite.dev/github/pr/aptos-labs/aptos-core/9457" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/aptos-labs/aptos-core/9312?utm_source=stack-comment).

zekun000 (approved):

consensus/src/dag/dag_handler.rs
ibalajiarun: Yes, the question is which one is preferred. I see you initially used consensus::TimeService but because RB became its own crate so I had to the aptos_time_service crate. I see some implementation differences between both.

zekun000: I think we should unify on the aptos_time_service, we can merge them on tech debt week

sasha8 (approved):

PR #9314, opened 2023-07-26
Add vector-load-test move example

banool:
### Description
Motivation in the README.

### Test Plan
```
aptos move publish --assume-yes --named-addresses addr=`yq .profiles.default.account < .aptos/config.yaml`
```
```
aptos move run --assume-yes --function-id `yq .profiles.default.account < .aptos/config.yaml`::vector_load_test::create --args u64:1000
```
```
aptos move run --assume-yes --function-id `yq .profiles.default.account < .aptos/config.yaml`::vector_load_test::update --args address:0x4c13422e2565b602481f729f26f1e3cd5af93f601b513e30567b5cd79a66bc7b u64:501 u8:67 u8:123 u8:244
```


banool:
Blockchain team wanted this for txn emitter btw.

igor-aptos (approved):
What length(s) should we test it with? Which one you used on testnet?

davidiw (changes requested):
Overall seems fine, but would prefer that any examples come with tests and inclusion in move unit tests. That way code is verified to compile.

Also it isn't super clear to me what this is demonstrating to users or where else this is being used.

banool:
@igor-aptos, please let me know if this module will be of particular value for the txn emitter (the original reason I made this PR) and I'll action the feedback. Otherwise I have other work to focus on right now.

banool:
Igor says this is still handy for benchmarking, he'll take over and put this in the right place.

PR #9317, opened 2023-07-26
Update base-gas.md

cryptomolot:
replaced the broken links. I hope I had the right logic (moving files to a different gas_schedule folder). 

I am also not sure about numbers of some parametres (they are different compared to new transaction.rs ; if you want I can update this with new numbers too.


vgao1996 (approved):

saharct (approved):
Thanks!

PR #9321, opened 2023-07-26
[Forge] Add dependencies between GHA workflow jobs.

JoshLind:
### Description
This PR updates the forge workflows to add dependencies between jobs. Specifically, we add a dependency between each forge test such that only one test should be running at a time. Additionally, the tests should always run, even if previous tests fail, e.g., see [here](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idif). This should help reduce spikes in the number of machines forge requires.

I've also updated the GHA workflows to start at different times, so that we hopefully scatter the jobs one after the other based on the rough end-to-end job time and scheduling, e.g., forge stable starts at 6am UTC (11pm PDT), and forge unstable starts at 12pm UTC (5am PDT), allowing ~6 hours for the stable job to complete. It won't be perfect, but we can fine tune things later to minimize overlap :)

 
### Test Plan
Existing test infrastructure.

### DIFF
--- a/.github/workflows/forge-pfn.yaml
+++ b/.github/workflows/forge-pfn.yaml
@@ -21,7 +21,7 @@ on:
         type: string
         description: The git SHA1 to checkout. This affects the Forge test runner that is used. If not specified, the latest main will be used
   schedule:
-    - cron: "0 8 */2 * *" # The main branch cadence. This runs every other day at 8am UTC.
+    - cron: "0 3 */2 * *" # The main branch cadence. This runs every other day at 3am UTC.
   pull_request:
     paths:
       - ".github/workflows/forge-pfn.yaml"
@@ -50,7 +50,7 @@ jobs:
         id: determine-test-branch
         run: |
           if [[ "${{ github.event_name }}" == "schedule" ]]; then
-            if [[ "${{ github.event.schedule }}" == "0 8 */2 * *" ]]; then
+            if [[ "${{ github.event.schedule }}" == "0 3 */2 * *" ]]; then
               echo "Branch: main"
               echo "BRANCH=main" >> $GITHUB_OUTPUT
             else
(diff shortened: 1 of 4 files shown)


rustielin (approved):

.github/workflows/forge-pfn.yaml
```diff
@@ -126,7 +126,7 @@ jobs:
   # Measures PFN latencies with a constant TPS (with network chaos)
   run-forge-pfn-const-tps-network-chaos:
     if: ${{ github.event_name != 'pull_request' }}
```
ibalajiarun, line 128:
```suggestion
    if: ${{ github.event_name != 'pull_request' && always() }}
```

I think we want the next job to start regardless of success criteria. (same for all)

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-not-requiring-successful-dependent-jobs

JoshLind: @ibalajiarun, I did wonder about that, but is that not logically equivalent to the current thing? e.g., `if X && True == X`?

JoshLind: Aah, found this: https://github.com/actions/runner/issues/491#issuecomment-660122693. You're correct 😄 What a weird world.

.github/workflows/forge-pfn.yaml
```diff
@@ -50,7 +50,7 @@ jobs:
         id: determine-test-branch
         run: |
           if [[ "${{ github.event_name }}" == "schedule" ]]; then
-            if [[ "${{ github.event.schedule }}" == "0 8 */2 * *" ]]; then
+            if [[ "${{ github.event.schedule }}" == "0 3 */2 * *" ]]; then
```
ibalajiarun, line 53: Somebody can miss changing this easily 😄

ibalajiarun (approved):
LGTM.

.github/workflows/forge-pfn.yaml
JoshLind: Yeah, it's a bit annoying 😄 I've done it at least once. We should probably refactor all of these files to have a single cron string defined, or just get rid of the check altogether?

PR #9324, opened 2023-07-26
[Rust][SDK] update rust sdk doc

0xjinn:


xbtmatt:
How come we're having them apply the patch as opposed to us doing it directly to the SDK? Just trying to understand- not sure I do

0xjinn:
> How come we're having them apply the patch as opposed to us doing it directly to the SDK? Just trying to understand- not sure I do

@xbtmatt  Good question! Greg and I was looking at the rust doc, and we found this - https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html

> Remember that [patch] is applicable transitively but can only be defined at the top level so we consumers of my-library have to repeat the [patch] section if necessary.

banool (approved):

gregnazario (approved):

saharct (approved):

PR #9333, opened 2023-07-27
[rb] Support for backoff based retry

ibalajiarun:
### Description

This PR adds backoff ability for RPC calls made for RB broadcast calls. 

### Test Plan

Existing unit tests modified

### DIFF
--- a/crates/reliable-broadcast/src/lib.rs
+++ b/crates/reliable-broadcast/src/lib.rs
@@ -56,19 +56,20 @@ where
         let receivers: Vec<_> = self.validators.clone();
         let network_sender = self.network_sender.clone();
         let time_service = self.time_service.clone();
-        let mut backoff_policies: HashMap<Author, TBackoff> = self.validators
+        let mut backoff_policies: HashMap<Author, TBackoff> = self
+            .validators
             .iter()
             .cloned()
             .map(|author| (author, self.backoff_policy.clone()))
             .collect();
         async move {
             let mut fut = FuturesUnordered::new();
-            let send_message = |receiver, message, sleep_duration: Duration| {
+            let send_message = |receiver, message, sleep_duration: Option<Duration>| {
                 let network_sender = network_sender.clone();
                 let time_service = time_service.clone();
                 async move {
-                    if !sleep_duration.is_zero() {
-                        time_service.sleep(sleep_duration).await;
+                    if let Some(duration) = sleep_duration {
+                        time_service.sleep(duration).await;
                     }
                     (
                         receiver,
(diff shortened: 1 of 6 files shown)


crates/reliable-broadcast/src/lib.rs
```diff
@@ -44,11 +55,21 @@ where
     ) -> impl Future<Output = S::Aggregated> {
         let receivers: Vec<_> = self.validators.clone();
         let network_sender = self.network_sender.clone();
+        let time_service = self.time_service.clone();
+        let mut backoff_policies: HashMap<Author, TBackoff> = self.validators
```
danielxiangzl, line 59: So this allows different backoff policies for different nodes, but here it uses the same policy for all nodes? Should we support different policies, or update the policy?

danielxiangzl (approved):

crates/reliable-broadcast/src/lib.rs
ibalajiarun: We will have the same policy for different peers. TBackoff is both a policy definition and an implementation - it gives the next interval to wait for each peer, so we need a copy for each one of them.

crates/reliable-broadcast/src/lib.rs
```diff
@@ -44,11 +55,21 @@ where
...
+            .map(|author| (author, self.backoff_policy.clone()))
+            .collect();
         async move {
             let mut fut = FuturesUnordered::new();
-            let send_message = |receiver, message| {
+            let send_message = |receiver, message, sleep_duration: Duration| {
                 let network_sender = network_sender.clone();
+                let time_service = time_service.clone();
                 async move {
+                    if !sleep_duration.is_zero() {
```
zekun000, line 70: nit: if this is only used for start, use Option<Duration> seems more readable

crates/reliable-broadcast/src/lib.rs
sasha8: Why do you have a map if all have the same policy?

crates/reliable-broadcast/src/lib.rs
```diff
@@ -44,11 +55,21 @@ where
...
+                        time_service.sleep(sleep_duration).await;
```
sasha8, line 71: Will this block the outside while let Some((receiver, result)) = fut.next() ?

sasha8 (approved):

crates/reliable-broadcast/src/lib.rs
ibalajiarun: because each one maintains the next interval to wait for. we need one for each peer. 

Otherwise, if we have two peers and one exponential backoff, then both of them have to retried, then for first peer we will use interval*factor but for second peer interval*factor*factor which is not what we want.

sasha8: Right, I am stupid, ignore me.

PR #9337, opened 2023-07-27
[Move prover] Some fixes in the prover and the specs

### LINKED ISSUE #9092: [Prover] Issue with the inconsistency checker

rahxephon89:
### Description

This PR does the following things:

- add the flag `--unconditional-abort-as-inconsistency` and `--disallow-global-timeout-to-be-overwritten` to `aptos move prove`;
- add a normal prover test to the daily prover test (in CI) with enough verification time so that all verification targets will be checked;
- add an inconsistency test to the daily prover test;
- fix some inconsistency issues in the specs;
- comment out problematic specs which needs further investigation.

close #9092

### DIFF
--- a/.github/workflows/prover-daily-test.yaml
+++ b/.github/workflows/prover-daily-test.yaml
@@ -38,5 +38,7 @@ jobs:
           echo "/home/runner/bin" | tee -a $GITHUB_PATH
           echo "/home/runner/.dotnet" | tee -a $GITHUB_PATH
           echo "/home/runner/.dotnet/tools" | tee -a $GITHUB_PATH
-      - run: MVP_TEST_INCONSISTENCY=1 cargo test -p aptos-move-examples --release -- --include-ignored prover
-      - run: MVP_TEST_INCONSISTENCY=1 cargo test -p aptos-framework --release -- --include-ignored prover
+      - run: MVP_TEST_DISALLOW_TIMEOUT_OVERWRITE=1 MVP_TEST_VC_TIMEOUT=1200 cargo test -p aptos-move-examples --release -- --include-ignored prover
+      - run: MVP_TEST_DISALLOW_TIMEOUT_OVERWRITE=1 MVP_TEST_VC_TIMEOUT=1200 cargo test -p aptos-framework --release -- --include-ignored prover
+      - run: MVP_TEST_DISALLOW_TIMEOUT_OVERWRITE=1 MVP_TEST_VC_TIMEOUT=1200 MVP_TEST_INCONSISTENCY=1 cargo test -p aptos-move-examples --release -- --include-ignored prover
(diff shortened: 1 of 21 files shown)


.github/workflows/prover-daily-test.yaml
```diff
@@ -38,5 +38,7 @@ jobs:
           echo "/home/runner/bin" | tee -a $GITHUB_PATH
           echo "/home/runner/.dotnet" | tee -a $GITHUB_PATH
           echo "/home/runner/.dotnet/tools" | tee -a $GITHUB_PATH
-      - run: MVP_TEST_INCONSISTENCY=1 cargo test -p aptos-move-examples --release -- --include-ignored prover
-      - run: MVP_TEST_INCONSISTENCY=1 cargo test -p aptos-framework --release -- --include-ignored prover
+      - run: MVP_TEST_DISALLOW_TIMEOUT_OVERWRITE=1 MVP_TEST_VC_TIMEOUT=1200 cargo test -p aptos-move-examples --release -- --include-ignored prover
```
junkil-park, line 41: Does this `--disable-global-timeout-overwrite` force Prover to the global timeout limit, ignoring the local ones?

rahxephon89: yes

junkil-park (approved):

junkil-park (approved):
Great work! Thank you!

aptos-move/framework/aptos-framework/sources/staking_contract.spec.move
```diff
@@ -20,8 +20,8 @@ spec aptos_framework::staking_contract {
 
     /// Staking_contract exists the stacker/operator pair.
     spec staking_contract_amounts(staker: address, operator: address): (u64, u64, u64) {
-        // TODO: set because of timeout (property proved)
-        pragma verify_duration_estimate = 1000;
+        // TODO: set because of timeout
```
wrwg, line 23: About those TODOs as it apparently is not properly proved? Here and elsewhere

wrwg (approved):

aptos-move/framework/aptos-framework/sources/staking_contract.spec.move
rahxephon89: MoveBit claimed that these properties could be proved with longer verification time but I found that they will timeout when being verified along with other specs. So I changed the comments and will investigate them more.

wrwg: But the comment is still saying here and there that timeout is the problem.

rahxephon89: Yes, the spec itself is not problematic but timeout is the issue. They cannot be proved even if the verification time is set to 1000 so I changed the comment.

PR #9341, opened 2023-07-27
[TS SDK] Set default denominator to 1

0xmaayan:

### DIFF
--- a/ecosystem/typescript/sdk/src/plugins/aptos_token.ts
+++ b/ecosystem/typescript/sdk/src/plugins/aptos_token.ts
@@ -130,7 +130,7 @@ export class AptosToken {
         options?.tokensBurnableByCreator ?? true,
         options?.tokensFreezableByCreator ?? true,
         options?.royaltyNumerator ?? 0,
-        options?.royaltyDenominator ?? 0,
+        options?.royaltyDenominator ?? 1,
       ],
       extraArgs,
     );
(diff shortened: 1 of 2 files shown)


ecosystem/typescript/sdk/src/plugins/aptos_token.ts
```diff
@@ -130,7 +130,7 @@ export class AptosToken {
         options?.tokensBurnableByCreator ?? true,
         options?.tokensFreezableByCreator ?? true,
         options?.royaltyNumerator ?? 0,
-        options?.royaltyDenominator ?? 0,
+        options?.royaltyDenominator ?? 1,
```
gregnazario, line 133: 1 let's you get past with 0 for default, I guess you have to change the denominator otherwise, it'll get blocked by limits on royalties I'm assuming, 100% isn't allowed?

gregnazario (approved):
Consider 100 perhaps, though we'd have to document the default behavior

ecosystem/typescript/sdk/src/plugins/aptos_token.ts
0xmaayan: https://aptos-org.slack.com/archives/C03N83P7QUC/p1690474951903069?thread_ts=1690474688.293299&cid=C03N83P7QUC

banool (approved):

PR #9344, opened 2023-07-27
[bytecode verifier] improve signature checker v2

vgao1996:
This fixes a counting bug in the script path and introduces a pseudo metering mechanism that aborts on programs that are too complex to analyze by this pass.

The budget of the meter is configured to be proportional to the size of the module/script being analyzed, clamping the time complexity of this pass to `O(n)`.


vgao1996:
Module verify task: https://github.com/aptos-labs/aptos-core/actions/runs/5744353154

Let's see if the current limit is good for the existing modules

gerben-stavenga (approved):

third_party/move/move-bytecode-verifier/src/signature_v2.rs
```diff
@@ -21,9 +22,271 @@ use std::{
...
+    }
+
+    fn meter_struct_instantiation(
+        &self,
+        struct_inst_idx: StructDefInstantiationIndex,
+    ) -> PartialVMResult<()> {
+        let struct_inst = self
+            .resolver
+            .struct_instantiation_at(struct_inst_idx)
+            .expect("struct instantiation should exist");
```
davidiw, line 117: expects are problematic in the case someone can construct a variant that would fail here. Would really rather we emit an error that this is not a valid module.

third_party/move/move-bytecode-verifier/src/signature_v2.rs
```diff
@@ -21,9 +22,271 @@ use std::{
...
+        self.meter_signature(struct_inst.type_parameters)
+    }
+
+    fn meter_struct_instantiations(&self) -> PartialVMResult<()> {
+        for struct_inst_idx in 0..self
+            .resolver
+            .struct_instantiations()
+            .expect("struct instantiations should exist")
```
davidiw, line 125: same here...

third_party/move/move-core/types/src/vm_status.rs
```diff
@@ -701,11 +701,12 @@ pub enum StatusCode {
     MAX_FIELD_DEFINITIONS_REACHED = 1121,
     // Reserved error code for future use
```
davidiw, line 702: feels like this comment should be shifted

third_party/move/move-vm/integration-tests/src/tests/regression_tests.rs
```diff
@@ -79,7 +79,7 @@ fn script_large_ty() {
 
     let script_code = &mut decompiled_script.code.code;
     script_code.clear();
-    let num_vecpacks = 1000;
+    let num_vecpacks = 1;
```
davidiw, line 82: why the adjustment?

third_party/move/move-vm/runtime/src/loader.rs
```diff
@@ -777,10 +778,14 @@ impl Loader {
 
     // Script verification steps.
     // See `verify_module()` for module verification steps.
-    fn verify_script(&self, script: &CompiledScript) -> VMResult<()> {
+    fn verify_script(&self, script: &CompiledScript, script_size: usize) -> VMResult<()> {
```
davidiw, line 781: Should we instead consider adding length to CompiledScript? It seems dangerous to decompose these.

third_party/move/move-vm/runtime/src/loader.rs
```diff
@@ -1014,19 +1019,20 @@ impl Loader {
...
+        modules: &[(CompiledModule, usize)],
         data_store: &mut TransactionDataCache,
     ) -> VMResult<()> {
         fail::fail_point!("verifier-failpoint-1", |_| { Ok(()) });
 
-        let mut bundle_unverified: BTreeSet<_> = modules.iter().map(|m| m.self_id()).collect();
+        let mut bundle_unverified: BTreeSet<_> = modules.iter().map(|(m, _)| m.self_id()).collect();
         let mut bundle_verified = BTreeMap::new();
-        for module in modules {
+        for (module, module_size) in modules {
```
davidiw, line 1029: same, might be better to add size to the CompiledModule

davidiw (commented):
Logic seems fine.

third_party/move/move-bytecode-verifier/src/signature_v2.rs
```diff
@@ -21,9 +22,271 @@ use std::{
 };
 use typed_arena::Arena;
 
+/***************************************************************************************************
+ * Pseudo Metering
+ *
+ **************************************************************************************************/
```
wrwg, line 28: I'm wondering how this relates to the existing metering approach: https://github.com/aptos-labs/aptos-core/blob/main/third_party/move/move-bytecode-verifier/src/meter.rs. What's the reason this rolls up its own?

vgao1996:
@wrwg This is because the actual metering logic here is extremely simple -- you keep subtracting some size from a variable representing your budget, and it is the only thing this has in common with `meter.rs`. I found it awkward to only use `trait Meter` as a simple counter, while not needing the other APIs (e.g. scope, transfer etc.).

I can see the point of having some common meter construct for the whole bytecode verifier though. Maybe that's your original intention with `meter.rs`?

third_party/move/move-bytecode-verifier/src/signature_v2.rs
vgao1996: @wrwg This is because the actual metering logic here is extremely simple -- you keep subtracting some size from a variable representing your budget, and it is the only thing this has in common with meter.rs. I found it awkward to only use trait Meter as a simple counter, while not needing the other APIs (e.g. scope, transfer etc.).

I can see the point of having some common meter construct for the whole bytecode verifier though. Maybe that's your original intention with meter.rs?

third_party/move/move-vm/integration-tests/src/tests/regression_tests.rs
vgao1996: It's because
1. This test only needs 1 vec pack to test the per-type limit
2. Having many vec packs with a large type will trigger the new metering mechanism, raising a different error code

third_party/move/move-vm/runtime/src/loader.rs
vgao1996: Yeah this is what I also found tricky. I also hope there's an easy way to obtain the size of a `CompiledModule/Script`, but the problem is that such size is not always well defined. `CompiledModule` by its current definition, is basically a low level AST of a Move module, and by because of this, (serialized) size is naturally doesn't belong to this struct and it could be undefined in many use cases (e.g. a test creating a module in memory).

third_party/move/move-bytecode-verifier/src/signature_v2.rs
wrwg: The meter type is already used in various components (absint-based but also type safety). Please check out the existing examples. Since its an `&mut impl Meter` when passed through, it supports zero abstraction cost if needed. Also it already has the error logic you are re-implementing here. You wont need scope et. al; this is for composing functions costs in a module (and currently has no effect IIRC).

Perhaps the concern here is that then you need to align the metering units with the remaining components? This is indeed a certain challenge. Typically this is done by assigning a cost as in

```
  const TYPE_NODE_COST: u128 = 30;
  ...
  fn charge_ty(&mut self, meter: &mut impl Meter, ty: &SignatureToken) -> PartialVMResult<()> {
        meter.add_items(
            Scope::Function,
            TYPE_NODE_COST,
            ty.preorder_traversal().count(),
        )
    }
```

This is a bit more cumbersome to calibrate but it makes it easier to give a reasonable bound to execution as a whole (plus brings us closer to a gas model).

vgao1996: I think I'm starting to see the bigger picture here. Is my understanding correct that you want us to have a unified meter interface (`trait Meter`) that is used by all metering components? Let's also discuss our long-term vision for metering.

> Perhaps the concern here is that then you need to align the metering units with the remaining components?

This should not be a major problem. Although I do have one question: do you think we should have a shared meter budget for all verifier passes? (My current implementation sets a separate budget for the signature checker.)

wrwg: > I think I'm starting to see the bigger picture here. Is my understanding correct that you want us to have a unified meter interface (`trait Meter`) that is used by all metering components? Let's also discuss our long-term vision for metering.
> 
> > Perhaps the concern here is that then you need to align the metering units with the remaining components?
> 
> This should not be a major problem. Although I do have one question: do you think we should have a shared meter budget for all verifier passes? (My current implementation sets a separate budget for the signature checker.)

Yes, that is the goal and already happening in multiple places in the existing code.

The goal is to have some global cost limit which we can set (similar as in gas). This requires callibration towards actual execution cost, which is currently a bit awkward and happening over the multiple constants in the code. The hope is that we can transition this eventually into gas charging. One `impl Meter` could then connect to the gas meter infra in the future..

vgao1996: Sounds good. I'll switch over to `trait Meter`.

> Although I do have one question: do you think we should have a shared meter budget for all verifier passes? (My current implementation sets a separate budget for the signature checker.)

Still, what your take on this? My feeling is that if we want proper metering eventually then we should just have a global budget?

PR #9349, opened 2023-07-27
[api] Add threads and timestamps to api tests

ngkuru:
### Description

This PR builds threads and run ID on top of the consistent API testing introduced [here](https://github.com/aptos-labs/aptos-core/pull/9146). See [proposal](https://www.notion.so/aptoslabs/API-Flow-Testing-Proposal-ae3424ce6e15433fa69f7c9dd99db037) for more high level information.

Changes:
* Refactored tests into dedicated file `tests.rs`.
* Switched to creating accounts for every tests instead of using the same and created pre test set up routines in `testsetups.rs`.
* Added run ID `start_time` to metrics.

Changes to follow (next PR):
* Increase resiliency for node consistency (try a few times if we don't get a correct answer).
* Emit latency metrics for individual steps of each test.

### Test Plan

Can view real time data from the cloud run job on the [dashboard](https://aptoslabs.grafana.net/d/f4fa3a4b-12aa-4804-8a2f-35264ca388f2/api-tester?orgId=1&from=now-3h&to=now).


crates/aptos-api-tester/src/main.rs
```diff
@@ -46,77 +44,63 @@ async fn handle_result<Fut: Future<Output = Result<(), TestFailure>>>(
...
-        network_type,
-        test_module(&client, &mut giray),
+        network_name,
+        wrapper_publishmodule(network_name),
     )
     .await;
 
+    handle_newaccount.await?;
+    handle_cointransfer.await?;
+    handle_nfttransfer.await?;
```
gregnazario, line 113: You can also do a join_all, which I believe will ensure they all happen in parallel. 
I think this might spawn them in tasks, but run them in serial currently.  How do the start times look? https://docs.rs/futures/latest/futures/future/fn.join_all.html

crates/aptos-api-tester/src/main.rs
```diff
@@ -126,21 +110,9 @@ async fn main() -> Result<()> {
...
-    // test flows on devnet
-    let _ = test_flows(
-        NetworkName::Devnet,
-        Client::new(DEVNET_NODE_URL.clone()),
-        FaucetClient::new(DEVNET_FAUCET_URL.clone(), DEVNET_NODE_URL.clone()),
-    )
-    .await;
+    // test flows
+    let _ = test_flows(NetworkName::Testnet).await;
+    let _ = test_flows(NetworkName::Devnet).await;
```
gregnazario, line 126: similarly you could join these together, because right now it runs Testnet, then it runs devnet.  Though separation in this case is probably good

crates/aptos-api-tester/src/wrappers.rs
```diff
@@ -0,0 +1,70 @@
...
+    let faucet_client = get_faucet_client(network_name);
+
+    // create and fund account
+    let account = create_and_fund_account(&faucet_client).await?;
+
+    // run test
+    test_newaccount(&client, &account, 100_000_000).await
+}
+
+pub async fn wrapper_cointransfer(network_name: NetworkName) -> Result<(), TestFailure> {
```
gregnazario, line 25: Wrapper might be better as a different word, since this is the test setup and all, but I don't have a good suggestion right now

crates/aptos-api-tester/src/main.rs
ngkuru: ok changed this but no effect because we still await for publish_module before we await for the handles and publish_module finishes last

crates/aptos-api-tester/src/tests.rs
```diff
@@ -0,0 +1,488 @@
...
+
+    // build module
+    let package = BuiltPackage::build(move_dir, options)?;
+    let blobs = package.extract_code();
+    let metadata = package.extract_metadata()?;
+
+    // create payload
+    let payload: aptos_types::transaction::TransactionPayload =
+        EntryFunctionCall::CodePublishPackageTxn {
+            metadata_serialized: bcs::to_bytes(&metadata).expect("PackageMetadata has BCS"),
```
gregnazario, line 381:
```suggestion
            metadata_serialized: bcs::to_bytes(&metadata).expect("PackageMetadata should deserialize"),
```

gregnazario (approved):

PR #9352, opened 2023-07-27
[move] kill the option in return type of type-to-layout conversion

lightmark:
### Description
None is unnecessary and I will run replay test.

### Test Plan
replay test and unit test

### DIFF
--- a/third_party/move/move-vm/runtime/src/native_functions.rs
+++ b/third_party/move/move-vm/runtime/src/native_functions.rs
@@ -164,23 +164,12 @@ impl<'a, 'b, 'c> NativeContext<'a, 'b, 'c> {
         self.resolver.loader().type_to_type_tag(ty)
     }
 
-    pub fn type_to_type_layout(&self, ty: &Type) -> PartialVMResult<Option<MoveTypeLayout>> {
-        match self.resolver.type_to_type_layout(ty) {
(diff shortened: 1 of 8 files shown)


lightmark:
replay test
https://github.com/aptos-labs/aptos-core/actions/runs/5686061683

third_party/move/move-vm/runtime/src/native_functions.rs
```diff
@@ -164,23 +164,12 @@ impl<'a, 'b, 'c> NativeContext<'a, 'b, 'c> {
         self.resolver.loader().type_to_type_tag(ty)
     }
 
-    pub fn type_to_type_layout(&self, ty: &Type) -> PartialVMResult<Option<MoveTypeLayout>> {
-        match self.resolver.type_to_type_layout(ty) {
-            Ok(ty_layout) => Ok(Some(ty_layout)),
-            Err(e) if e.major_status().status_type() == StatusType::InvariantViolation => Err(e),
```
wrwg, line 170: I guess this treatment might be related with the older but now removed feature of native structs. Once those structs where supported, and they would not have a type layout. They are gone, so this change is good.

wrwg (approved):

movekevin (approved):

PR #9356, opened 2023-07-28
Implement estimate gas fee method for Python SDK along with simulate_…

ikapeykin:
…transaction method

### Description
Let's allow dev community estimating gas fee with specific method. More details:
https://github.com/web3engineering/LayerZero-python-aptos/pull/10/files
https://github.com/web3engineering/LayerZero-python-aptos/issues/9 (discussion from Discord)

### Test Plan
Already tested there: https://github.com/web3engineering/LayerZero-python-aptos/pull/10/files

Tested locally on my side. Transaction hash: 0xe1609d8ebc2816e5c03f92bc2e51bee101e6036c65f7017e2e3f5b4a7f2f7761
(I used theaptosbridge for this)


bobu4 (commented):
More laconic way is to add estimate_gas_usage bool argument with default value False in simulate transaction function. If this is True pass dict with required params set to True, if false pass empty dict to params.

ikapeykin:
> No need to create new function with the same feature if there is a function which could be expand with new feature.

Agree, will update it as well

ikapeykin:
Thanks again @bobu4, it works really nice :)

davidiw (approved):
cool!

I might have used kwargs instead, but this whole thing will need to be re-worked at some point to support more variants of transactions when they're available.

ikapeykin:
Hi @0xmaayan! Could you take a look on this PR? I'd to merge it and start using there https://github.com/web3engineering/LayerZero-python-aptos

ikapeykin:
> Are you looking for an SDK update bump?

Yeah, could be nice

PR #9357, opened 2023-07-28
[TS] Added a typescript example for the gas fee payer

junkil-park:
### Description
Added a gas fee payer example based on the existing e2e test scenario.

### Test Plan
`pnpm run gas_fee_payer`


ecosystem/typescript/sdk/examples/typescript/gas_fee_payer.ts
```diff
@@ -0,0 +1,153 @@
...
+  console.log(`Alice's balance: ${await getBalance(alice)} octas`)
+  console.log(`Bob's balance: ${await getBalance(bob)} octas`)
+  await waitForEnter();
+
+
+  // Transfer Token from Alice's Account to Bob's Account with bob paying the fee
+  console.log("\n=== Alice sent the token to Bob while Bob paied the fee ===");
+  const aliceBefore = await getBalance(alice);
+  const bobBefore = await getBalance(bob);
+  const txnHash = await tokenClient.directTransferTokenWithFeePayer(
```
movekevin, line 125: It's a good start to use directTransferTokenWithFeePayer which supports a separate gas payer. But we should add another type of tx here where you have to construct a tx with a gas payer also to show people how to do it. directTransferTokenWithFeePayer is a function only used for token v1 and not a common one either. Most people using gas payer would use it for a different transaction

gerben-stavenga: yes directtransfer was my example on how to construct the tx. You should look in that to see how to make a general tx.

gerben-stavenga (approved):

ecosystem/typescript/sdk/examples/typescript/gas_fee_payer.ts
junkil-park: Added another use of fee payer transaction (`createTokenWithFeePayer`). It builds, signs, and submits a raw transaction with a fee payer. One should be able to follow this flow for their own type of transactions.

ecosystem/typescript/sdk/examples/typescript/gas_fee_payer.ts
```diff
@@ -0,0 +1,228 @@
...
+}
+
+const getBalance = async (account: AptosAccount) => {
+  const resources = await client.getAccountResources(account.address().hex());
+  const aptosCoin = "0x1::coin::CoinStore<0x1::aptos_coin::AptosCoin>";
+  let accountResource = resources.find((r) => r.type === aptosCoin);
+  return BigInt((accountResource!.data as any).coin.value);
+};
+
+async function createTokenWithFeePayer(feePayer: AptosAccount, account: AptosAccount, collectionName: string, name: string, description: string, supply: number, uri: string, max: BCS.AnyNumber, royalty_payee_address?: MaybeHexString, royalty_points_denominator?: number, royalty_points_numerator?: number, property_keys?: Array<string>, property_values?: Array<string>, property_types?: Array<string>, extraArgs?: OptionalTransactionArgs): Promise<string> {
```
movekevin, line 51: This looks great already. But I'd add some comments at the top here to explain the overall flow. May be something along the like of:

To create a transaction with a different account paying for gas, we need to:
1. Create a fee payer raw transaction (TxnBuilderTypes.FeePayerRawTransaction) and specify which account will be paying for gas.
2. Two signatures are needed - one from the account sending the transaction and one from the separate gas payer. In this example, the sender will sign first, followed by the gas payer. However, the reverse order should also work. The signatures can be generated the same way they are for normal transactions.
3. Once we have two signatures, the transaction can be sent, and gas fees will be deducted from the gas payer account instead.

ecosystem/typescript/sdk/examples/typescript/gas_fee_payer.ts
```diff
@@ -0,0 +1,228 @@
...
+  const accountSignature = new TxnBuilderTypes.Ed25519Signature(
+    account.signBuffer(TransactionBuilder.getSigningMessage(feePayerTxn)).toUint8Array(),
+  );
+  // Construct the account authenticator.
+  const accountAuthenticator = new TxnBuilderTypes.AccountAuthenticatorEd25519(
+    new TxnBuilderTypes.Ed25519PublicKey(account.signingKey.publicKey),
+    accountSignature,
+  );
+  // `feePayer` signs the fee payer raw transaction.
+  const feePayerSignature = new TxnBuilderTypes.Ed25519Signature(
```
movekevin, line 89: I wonder if this flow works out of the box with wallets. cc @0xmaayan @kent-white do wallets/wallet connector currently support signing multi-agent transactions?

0xmaayan: I just talked to @hardsetting about it, if I understand correctly it should be supported, but defer to @hardsetting

0xmaayan: from the wallet adapter side, I believe it supports only a payload input so it assumes wallet knows how to handle it to generate and submit a multi transaction - might need to do some adapter changes?

ecosystem/typescript/sdk/examples/typescript/gas_fee_payer.ts
junkil-park: Sounds good. I just added the description to the example.

ecosystem/typescript/sdk/examples/typescript/gas_fee_payer.ts
junkil-park: I don't have the context. What do you mean by the wallet supporting the gas fee payment flow? What is a use case scenario?

movekevin: The wallet + wallet adapter need to support these specific lines
```
const accountSignature = new TxnBuilderTypes.Ed25519Signature(
    account.signBuffer(TransactionBuilder.getSigningMessage(feePayerTxn)).toUint8Array(),
  );
```

@0xmaayan I'm a bit rusty on what the wallet adapter interface looks like. Is there a way to sign something like this? You can also give us some code pointers as well and we can go dig into it

0xmaayan: the adapter is just a layer to pass data/info to the wallet. It does have functions like `signMessage` / `signTransaction` but at the end it depends on the wallet implementation
https://github.com/aptos-labs/aptos-wallet-adapter/blob/main/packages/wallet-adapter-core/src/WalletCore.ts#L329-L372

movekevin: Sounds good. @junkil-park I think the best next step here is to test this out in the browser and see if wallet connection works. Perhaps you can do some simple tweakings around Mayaan's e2e dapp guide https://aptos.dev/tutorials/build-e2e-dapp/create-a-smart-contract to load your code in the browser and test out the wallet connection part.

movekevin (approved):

PR #9361, opened 2023-07-28
[compiler-v2] Use livevar analysis to compute copy/move instructions

wrwg:
Live variables is a standard compiler construction data flow analysis. It computes the set of variables which are alive (being used in subsequent reachable code) before and after each program point. The implementation from the prover uses our dataflow analysis framework and is [here](https://github.com/aptos-labs/aptos-core/blob/206f529c0c9d8488e27d2e50297178f0caf429a5/third_party/move/move-prover/bytecode/src/livevar_analysis.rs#L381). 

In this PR we create the first processing step in the bytecode pipeline for the V2 compiler with the new module `pipeline/livevar_analysis_step.rs` which forwards logical work to the existing analysis. This is then used primarily to compute copy/move information correctly which is a pre-requisite for ability and borrow safety analysis.


brmataptos (changes requested):
Also: Not sure what to do with this, but the BTreeMap initialization in livevar_analysis.rs (outside of this review) is pretty pessimal, as it is sequential and will rebalance the tree on every insert.  Apparently if you create a BTreeMap directly from an iterator it saves the elements in a vector, sorts it (hopefully efficiently if it's already sorted), then creates the BTree more efficiently.  I don't have numbers, but maybe be careful about anti-performance patterns like that.

third_party/move/move-vm/runtime/src/loader.rs
```diff
@@ -754,6 +754,7 @@ impl Loader {
             Ok(script) => script,
             Err(err) => {
                 let msg = format!("[VM] deserializer for script returned error: {:?}", err);
+                println!("foo: {}", msg);
```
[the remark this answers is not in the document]
wrwg: Debug print is already removed (outdated code).

third_party/move/move-compiler-v2/src/file_format_generator/function_generator.rs
```diff
@@ -710,11 +780,36 @@ impl<'env> FunctionContext<'env> {
...
     pub fn is_copyable(&self, temp: TempIndex) -> bool {
         self.module
             .env
             .type_abilities(self.temp_type(temp), &self.type_parameters)
             .has_ability(FF::Ability::Copy)
     }
+
+    /// Determine whether the temporary is alive (used) in the reachable code after this point.
+    pub fn is_alive_after(&self, code_offset: FF::CodeOffset, temp: TempIndex) -> bool {
+        let LiveVarAnnotation(map) = self
```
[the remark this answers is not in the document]
wrwg: It is nearly zero cost since a reference to the inner map is returned. But you are right about the abstraction, in fact there was already a function `get_live_var_info_at` for the annotation type. Done.

third_party/move/move-compiler-v2/src/file_format_generator/function_generator.rs
```diff
@@ -132,17 +134,18 @@ impl<'a> FunctionGenerator<'a> {
         // Walk the bytecode
         let bytecode = ctx.fun.get_bytecode();
         for i in 0..bytecode.len() {
+            let code_offset = i as FF::CodeOffset;
             if i + 1 < bytecode.len() {
                 let bc = &bytecode[i];
                 let next_bc = &bytecode[i + 1];
-                self.gen_bytecode(ctx, &bytecode[i], Some(next_bc));
+                self.gen_bytecode(ctx, code_offset, &bytecode[i], Some(next_bc));
```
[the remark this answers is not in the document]
wrwg: Yes, fully agree. I was thinking of doing this earlier, but worked around until now. Done.

third_party/move/move-model/src/model.rs
```diff
@@ -3390,22 +3390,23 @@ impl<'env> FunctionEnv<'env> {
...
-                // where <num> seems to be generated non-deterministically.
-                // Substitute this by a deterministic name which the backend accepts.
-                let clean_ident = if ident.contains("%#") {
-                    format!("tmp#${}", idx)
-                } else {
-                    ident
-                };
-                return Some(self.module_env.env.symbol_pool.make(clean_ident.as_str()));
+        // Try to obtain user name
+        if let Some(source_map) = &self.module_env.data.source_map {
```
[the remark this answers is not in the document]
wrwg: Its for a consistent experience of naming. If there is no source map available, always name temps using `$t<n>`. Before it was some hybrid mix. Eventually the frontend still needs to generate the source map, which is work to do.

third_party/move/move-prover/bytecode/src/function_target_pipeline.rs
```diff
@@ -337,11 +337,6 @@ impl FunctionTargetPipeline {
...
             let fun_env = env.get_function(fun_id);
             for callee in fun_env.get_called_functions().expect("called functions") {
-                assert!(
-                    nodes.contains_key(callee),
-                    "{}",
-                    env.get_function(*callee).get_full_name_str()
-                );
                 let dst_idx = nodes
                     .get(callee)
                     .expect("callee is not in function targets");
```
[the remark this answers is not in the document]
wrwg: Yes that is an ICE. Its like an `assert!` but idiomatic also to do this way.

rahxephon89 (approved):
LGTM, thanks!

third_party/move/move-compiler-v2/src/file_format_generator/function_generator.rs
```diff
@@ -637,17 +667,25 @@ impl<'a> FunctionGenerator<'a> {
...
-    fn abstract_flush_stack(&mut self, ctx: &FunctionContext, top: usize) {
+    /// Flush the abstract stack, ensuring that all values on the stack are stored in locals, if
+    /// they are still alive.
+    fn abstract_flush_stack(&mut self, ctx: &BytecodeContext, top: usize) {
+        let fun_ctx = ctx.fun_ctx;
         while self.stack.len() > top {
             let temp = self.stack.pop().unwrap();
-            let local = self.temp_to_local(ctx, temp);
-            self.emit(FF::Bytecode::StLoc(local));
+            if ctx.is_alive_before(temp) || ctx.is_alive_after(temp) || self.pinned.contains(&temp)
```
brmataptos, line 676: Why are you testing is_alive_before here?

Also, It seems that you only need to save a pinned local if it has been changed on the stack.  But abstract_push_result() already flushes a pinned temp.

We may need some targeted test cases that make this situation easier to see.

brmataptos (changes requested):

third_party/move/move-compiler-v2/src/file_format_generator/function_generator.rs
wrwg: When abstract_flush_stack is called, we are interested in the pre state. That is because the flushing will be essentially inserted _before_ the next instruction. This is different to when we use `is_alive_after` above: here we are interested in what is alive _after_ this given program point. FWIW this has been discovered by testing; before I would use the after-set.

Regards your 2nd point: There are temps on the stack which may in fact have never been stored into any local (as this is a stack machine -- in the ideal case we need to store as little as possible). Those need to be saved here. You are right we could optimize the case if the temp on the stack is loaded from a non-pinned local. I suggest we keep this optimization for later. Not really need the most optimal code here -- doing the liveness here so we can do meaningful ability and borrow analysis, which require the right placement of copy and move as pre-requisite.

Testing: please note this is quite extensively tested via the transactional tests [here](https://github.com/aptos-labs/aptos-core/tree/main/third_party/move/move-compiler-v2/transactional-tests/tests), which you do not see as part of this change. In fact I debugged quite a bit based on those tests.

wrwg: Regards the abstract stack, there is now also a section in the design doc: https://www.notion.so/aptoslabs/Aptos-Move-Compiler-V2-High-Level-Design-6ab5e5ffcd574c0693795e6fdefc1331?pvs=4#fc7f5dc9114a4a6097eeb75cdc2d34ff

brmataptos: I think you should generate tight code based on the information you have available here, rather than being sloppy and assuming it will be fixed later.  We have the liveness results, let's use them effectively.

I'm afraid that abstract_flush_stack() is saving temps from the stack into locals in too many cases.  It might help to add comments for the pre/post conditions for a few of these shared operations, since the state of the function generator is a bit inconsistent as it is called in a few places at different points in generating code for a given input bytecode. 

The function is defined here: https://github.com/aptos-labs/aptos-core/blob/47803cd9d7ca46de87d9cfbbdd24678a59dbd144/third_party/move/move-compiler-v2/src/file_format_generator/function_generator.rs#L672.
and the overly-generous test is here:
https://github.com/aptos-labs/aptos-core/blob/47803cd9d7ca46de87d9cfbbdd24678a59dbd144/third_party/move/move-compiler-v2/src/file_format_generator/function_generator.rs#L676

(1) is_alive_before/after: See [function_generator.rs:152-156](https://github.com/aptos-labs/aptos-core/blob/47803cd9d7ca46de87d9cfbbdd24678a59dbd144/third_party/move/move-compiler-v2/src/file_format_generator/function_generator.rs#L152) (your branch) where we see:
```
                self.gen_bytecode(&bytecode_ctx, &bytecode[i], Some(next_bc));
                if !bc.is_branch() && matches!(next_bc, Bytecode::Label(..)) {
                    // At block boundaries without a preceding branch, need to flush stack
                    // TODO: to avoid this, we should use the CFG for code generation.
                    self.abstract_flush_stack(&bytecode_ctx, 0);
```
So in this particular case, **after** the bytecode is generated, we carefully store temps including those which are alive **before** the generated bytecode (even if not live **after** the bytecode).  In other cases (e.g., [in abstract_push_args()](https://github.com/aptos-labs/aptos-core/blob/47803cd9d7ca46de87d9cfbbdd24678a59dbd144/third_party/move/move-compiler-v2/src/file_format_generator/function_generator.rs#L656)) the flush is presumably happening **before** the current bytecode.  There's also a call [in abstract_push_result()](https://github.com/aptos-labs/aptos-core/blob/47803cd9d7ca46de87d9cfbbdd24678a59dbd144/third_party/move/move-compiler-v2/src/file_format_generator/function_generator.rs#L698) which also seems to happen **after** the bytecode is generated.

The current code test on line 676 (`if ctx.is_alive_before(temp) || ctx.is_alive_after(temp) || self.pinned.contains(&temp)`)  is conservative, and safe, in that it saves a superset of the required temps that works for both cases.  **But it would be better to differentiate the 2 cases (before and after the bytecode) so that spurious temp saves are not done.**

(2) is_pinned: The current code seems to ignore the lifetime of a pinned variable and consider it pinned for the rest of the method.  **Fine, though it would be good to mention it somewhere.**  Even with this conservative assumption, a pinned temp only needs to be saved if it has not already been copied to its home local.

There are only 2 locations where stack.push() is called.  (i) In abstract_push_args(), the temp is saved to a local before being pushed on the stack.  This case could presumably happen when pushing args for a Borrow operation, in which case the value would not yet have been saved, and should probably be saved to its home location immediately rather than waiting for abstract_flush_stack() to do it.

(ii) In abstract_push_result(), a temp is pushed, and if it is pinned, then it is saved right away, so there is no need to copy it later when abstract_flush_stack() is called and pops that temp.

So the is_pinned case is redundant in case (ii), and in case (i) it could be saved immediately.

(3) I don't see bytecode output for the transactional tests, so they aren't as useful to understand what is happening.

brmataptos: Btw, since the current code is *safe*, I'd be ok with accepting this PR if we add an issue to go back and clarify the "stack temp needs to be stored in local" property in a latter PR.

brmataptos: Ok.

brmataptos (approved):
Looks ok.

PR #9366, opened 2023-07-28
[ts-sdk] Export more types

gregnazario:
### Description
There were a lot of types that were not being exported, and many users asked for them to be exported.  Any types that are not exported here are exported under `TransactionBuilderTypes`

Additionally, I cleaned up a few interfaces that were either not using the already defined types, or were less permissive than other places that took in the same values.

### DIFF
--- a/ecosystem/typescript/sdk/src/index.ts
+++ b/ecosystem/typescript/sdk/src/index.ts
@@ -14,4 +14,4 @@ export * from "./utils/api-endpoints";
 export * from "./utils/hex_string";
 export * from "./utils/hd-key";
 export * from "./utils/property_map_serde";
-export {APTOS_COIN } from "./utils/misc"
+export { APTOS_COIN } from "./utils/misc";
--- a/ecosystem/typescript/sdk/src/index.ts
+++ b/ecosystem/typescript/sdk/src/index.ts
@@ -4,17 +4,14 @@
 export * from "./account";
 export * from "./providers";
 export * as BCS from "./bcs";
-export * from "./utils/hex_string";
 export * from "./plugins";
 export * from "./transaction_builder";
 export * from "./transactions";
 export * as TokenTypes from "./aptos_types/token_types";
 export * as Types from "./generated/index";
 export * from "./client";
-export { derivePath } from "./utils/hd-key";
-export {
-  deserializePropertyMap,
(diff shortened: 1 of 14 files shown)


ecosystem/typescript/sdk/src/index.ts
```diff
@@ -4,17 +4,14 @@
...
 export * as Types from "./generated/index";
 export * from "./client";
-export { derivePath } from "./utils/hd-key";
-export {
-  deserializePropertyMap,
-  deserializeValueBasedOnTypeTag,
-  getPropertyValueRaw,
-} from "./utils/property_map_serde";
-export { Network, CustomEndpoints } from "./utils/api-endpoints";
+export * from "./utils/api-endpoints";
```
0xmaayan, line 13: why not `export * from "./utils"`?

ecosystem/typescript/sdk/src/providers/index.ts
```diff
@@ -1,3 +1,18 @@
...
+
+/**
+ * Controls the number of results that are returned and the starting position of those results.
+ * limit specifies the maximum number of items or records to return in a query result.
+ * offset parameter specifies the starting position of the query result within the set of data.
+ * For example, if you want to retrieve records 11-20,
+ * you would set the offset parameter to 10 (i.e., the index of the first record to retrieve is 10)
+ * and the limit parameter to 10 (i.e., the number of records to retrieve is 10))
+ */
+export interface PaginationArgs {
```
0xmaayan, line 15: can probably remove it from here? as we now have `IndexerPaginationArgs`

0xmaayan (approved):
noice!

ecosystem/typescript/sdk/src/providers/index.ts
gregnazario: Yes, let me fix that.

gregnazario: For future versions, let's make one that's common, and do the name conversion for the user.

ecosystem/typescript/sdk/src/index.ts
gregnazario: There's some circular dependency

xbtmatt (approved):

PR #9368, opened 2023-07-28
Aggregator v2 and Snapshots Implementation

vusirikala:
### Description

Implementing the Aggregator v2 and Snapshots feature in aggregator extension.

### DIFF
--- a/aptos-move/aptos-aggregator/src/aggregator_change_set.rs
+++ b/aptos-move/aptos-aggregator/src/aggregator_change_set.rs
@@ -1,0 +1,507 @@
+// Copyright © Aptos Foundation
+
+use crate::{
+    aggregator_extension::{
+        validate_history, AggregatorID, AggregatorState, DeltaHistory, SpeculativeStartValue,
+    },
+    bounded_math::{abort_error, addition_deltavalue, ok_overflow, ok_underflow},
+};
+use aptos_state_view::StateView;
+use move_binary_format::errors::{PartialVMError, PartialVMResult};
+use move_core_types::vm_status::{StatusCode, VMStatus};
+
+/// When merging aggregator changes of two transactions,
+/// unable to merge the histories of the transactions.
+pub(crate) const EMERGE_HISTORIES: u64 = 0x02_0008;
+
+/// Represents a single aggregator change.
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub struct AggregatorChange {
+    pub max_value: u128,
+    pub state: AggregatorState,
+    // `base_aggregator` is None for Aggregators.
+    // `base_aggregator` is Some(id) for AggregatorSnapshots.
+    pub base_aggregator: Option<AggregatorID>,
+}
+
+impl AggregatorChange {
+    /// TODO: Implement this.
+    pub fn materialize_aggregator(
+        self,
+        _state_view: &dyn StateView,
+        _aggregator_id: &AggregatorID,
+    ) -> anyhow::Result<AggregatorChange, VMStatus> {
+        Ok(self)
+    }
+
+    /// Returns the result of applying the AggregatorState to the base value
+    /// Returns error if postcondition is not satisfied.
+    pub fn apply_aggregator_change_to(&self, base: u128) -> PartialVMResult<u128> {
+        match self.state {
+            AggregatorState::Data { value } => Ok(value),
+            AggregatorState::Delta { delta, .. } => {
+                // First, validate if the current delta operation can be applied to the base.
+                validate_history(base, self.max_value, &self.state)?;
+                Ok(addition_deltavalue(base, delta, self.max_value)?)
+            },
+        }
+    }
+
+    /// Applies self on top of previous delta, merging them together. Note
+    /// that the strict ordering here is crucial for catching overflows
+    /// correctly.
+    ///
+    /// TODO: What happens if the base aggregator is not none?
+    pub fn merge_with_previous_aggregator_change(
+        &mut self,
+        previous_change: AggregatorChange,
+    ) -> PartialVMResult<()> {
+        assert_eq!(
+            self.max_value, previous_change.max_value,
+            "Cannot merge aggregator changes with different max_values",
+        );
+        // When the previous aggregator change is a snapshot, we cannot merge it with the current.
+        if previous_change.base_aggregator.is_some() {
+            return Err(
+                PartialVMError::new(StatusCode::UNKNOWN_INVARIANT_VIOLATION_ERROR)
+                    .with_message("Cannot merge aggregator snapshots".to_string()),
+            );
+        }
+        match self.state {
+            AggregatorState::Data { .. } => {
+                // If the current state is Data, then merging with previous state won't change anything.
+                return Ok(());
+            },
+            AggregatorState::Delta { delta, history, .. } => {
+                match previous_change.state {
+                    AggregatorState::Data { value } => {
+                        // When prev_state is Data { value }, and current state is Delta { delta }, merging them into Data { value + delta }
+                        validate_history(value, self.max_value, &self.state)?;
+                        self.state = AggregatorState::Data {
+                            value: addition_deltavalue(value, delta, self.max_value)?,
+                        };
+                    },
+                    AggregatorState::Delta {
+                        speculative_start_value: _,
+                        delta: prev_delta,
+                        history: prev_history,
+                    } => {
+                        let new_delta = prev_delta.add(&delta, self.max_value)?;
+
+                        let new_min_overflow = {
+                            let adjusted_min_overflow_positive_delta =
+                                history.min_overflow_positive_delta.map_or(
+                                    Ok(None),
+                                    // Return Result<Option<u128>>. we want to have None on overflow,
+                                    // and to fail the merging on underflow
+                                    |min_overflow_positive_delta| {
+                                        ok_overflow(addition_deltavalue(
+                                            min_overflow_positive_delta,
+                                            prev_delta,
+                                            self.max_value,
+                                        ))
+                                    },
+                                )?;
+
+                            match (
+                                adjusted_min_overflow_positive_delta,
+                                prev_history.min_overflow_positive_delta,
+                            ) {
+                                (Some(a), Some(b)) => Some(u128::min(a, b)),
+                                (a, b) => a.or(b),
+                            }
+                        };
+
+                        let new_max_underflow = {
+                            let adjusted_max_underflow_negative_delta =
+                                history.max_underflow_negative_delta.map_or(
+                                    Ok(None),
+                                    // Return Result<Option<u128>>. we want to have None on overflow,
+                                    // and to fail the merging on underflow
+                                    |max_underflow_negative_delta| {
+                                        ok_overflow(addition_deltavalue(
+                                            max_underflow_negative_delta,
+                                            prev_delta.minus(),
+                                            self.max_value,
+                                        ))
+                                    },
+                                )?;
+
+                            match (
+                                adjusted_max_underflow_negative_delta,
+                                prev_history.max_underflow_negative_delta,
+                            ) {
+                                (Some(a), Some(b)) => Some(u128::min(a, b)),
+                                (a, b) => a.or(b),
+                            }
+                        };
+
+                        // new_max_achieved = max(prev_max_achieved, max_achieved + prev_delta)
+                        // When adjusting max_achieved, if underflow - than the other is bigger,
+                        // but if overflow - we fail the merge, as we cannot successfully achieve delta larger than max_value.
+                        let new_max_achieved = ok_underflow(addition_deltavalue(
+                            history.max_achieved_positive_delta,
+                            prev_delta,
+                            self.max_value,
+                        ))?
+                        .map_or(prev_history.max_achieved_positive_delta, |value| {
+                            u128::max(prev_history.max_achieved_positive_delta, value)
+                        });
+
+                        // new_min_achieved = max(prev_min_achieved, min_achieved - prev_delta)
+                        let new_min_achieved = ok_underflow(addition_deltavalue(
+                            history.min_achieved_negative_delta,
+                            prev_delta.minus(),
+                            self.max_value,
+                        ))?
+                        .map_or(prev_history.min_achieved_negative_delta, |value| {
+                            u128::max(prev_history.min_achieved_negative_delta, value)
+                        });
+
+                        if (new_min_overflow.is_some()
+                            && new_min_overflow.unwrap() <= new_max_achieved)
+                            || (new_max_underflow.is_some()
+                                && new_max_underflow.unwrap() <= new_min_achieved)
+                        {
+                            return Err(abort_error(
+                                "Unable to merge aggregator change histories",
+                                EMERGE_HISTORIES,
+                            ));
+                        };
+
+                        self.state = AggregatorState::Delta {
+                            speculative_start_value: SpeculativeStartValue::Unset,
+                            delta: new_delta,
+                            history: DeltaHistory {
+                                max_achieved_positive_delta: new_max_achieved,
+                                min_achieved_negative_delta: new_min_achieved,
+                                min_overflow_positive_delta: new_min_overflow,
+                                max_underflow_negative_delta: new_max_underflow,
+                            },
+                        };
+                    },
+                }
+            },
+        }
+        Ok(())
+    }
+
+    /// Applies next aggregator change on top of self, merging two changes together. This is a reverse
+    /// of `merge_with_previous_aggregator_change`.
+    pub fn merge_with_next_aggregator_change(
+        &mut self,
+        next_change: AggregatorChange,
+    ) -> PartialVMResult<()> {
+        // Now self follows the other delta.
+        let mut prev_change = next_change;
+        std::mem::swap(self, &mut prev_change);
+
+        // Perform the merge.
+        self.merge_with_previous_aggregator_change(prev_change)?;
+        Ok(())
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use crate::bounded_math::DeltaValue;
+    use claims::{assert_err, assert_ok};
+
+    #[test]
+    fn test_merge_aggregator_data_into_delta() {
+        let aggregator_change1 = AggregatorChange {
+            max_value: 100,
+            state: AggregatorState::Data { value: 20 },
+            base_aggregator: None,
+        };
+
+        let mut aggregator_change2 = AggregatorChange {
+            max_value: 100,
+            state: AggregatorState::Delta {
+                speculative_start_value: SpeculativeStartValue::AggregatedValue(10),
+                delta: DeltaValue::Positive(10),
+                history: DeltaHistory {
+                    max_achieved_positive_delta: 50,
+                    min_achieved_negative_delta: 5,
+                    min_overflow_positive_delta: None,
+                    max_underflow_negative_delta: None,
+                },
+            },
+            base_aggregator: None,
+        };
+        let mut aggregator_change3 = AggregatorChange {
+            max_value: 100,
+            state: AggregatorState::Delta {
+                speculative_start_value: SpeculativeStartValue::AggregatedValue(40),
+                delta: DeltaValue::Positive(10),
+                history: DeltaHistory {
+                    max_achieved_positive_delta: 50,
+                    min_achieved_negative_delta: 35,
+                    min_overflow_positive_delta: None,
+                    max_underflow_negative_delta: None,
+                },
+            },
+            base_aggregator: None,
+        };
+
+        assert_ok!(aggregator_change2.merge_with_previous_aggregator_change(aggregator_change1));
+        assert_eq!(aggregator_change2, AggregatorChange {
+            max_value: 100,
+            state: AggregatorState::Data { value: 30 },
+            base_aggregator: None,
+        });
+        assert_err!(aggregator_change3.merge_with_previous_aggregator_change(aggregator_change2));
+    }
+
+    #[test]
+    fn test_merge_data_into_data() {
+        let aggregator_change1 = AggregatorChange {
+            max_value: 100,
+            state: AggregatorState::Data { value: 20 },
+            base_aggregator: None,
+        };
+
+        let mut aggregator_change2 = AggregatorChange {
+            max_value: 100,
+            state: AggregatorState::Data { value: 50 },
+            base_aggregator: None,
+        };
+
+        let mut aggregator_change3 = AggregatorChange {
+            max_value: 100,
+            state: AggregatorState::Data { value: 70 },
+            base_aggregator: None,
+        };
+
+        assert_ok!(aggregator_change2.merge_with_previous_aggregator_change(aggregator_change1));
+        assert_eq!(aggregator_change2, AggregatorChange {
+            max_value: 100,
+            state: AggregatorState::Data { value: 50 },
+            base_aggregator: None,
+        });
+        assert_ok!(aggregator_change3.merge_with_previous_aggregator_change(aggregator_change2));
+        assert_eq!(aggregator_change3, AggregatorChange {
+            max_value: 100,
+            state: AggregatorState::Data { value: 70 },
+            base_aggregator: None,
+        });
+    }
+
+    #[test]
+    fn test_merge_delta_into_delta() {
+        let aggregator_change1 = AggregatorChange {
+            max_value: 100,
+            state: AggregatorState::Delta {
+                speculative_start_value: SpeculativeStartValue::AggregatedValue(20),
+                delta: DeltaValue::Positive(10),
+                history: DeltaHistory {
+                    max_achieved_positive_delta: 30,
+                    min_achieved_negative_delta: 15,
+                    min_overflow_positive_delta: Some(90),
+                    max_underflow_negative_delta: Some(25),
+                },
+            },
+            base_aggregator: None,
+        };
+        let mut aggregator_change2 = AggregatorChange {
+            max_value: 100,
+            state: AggregatorState::Delta {
+                speculative_start_value: SpeculativeStartValue::AggregatedValue(40),
+                delta: DeltaValue::Positive(20),
+                history: DeltaHistory {
+                    max_achieved_positive_delta: 25,
+                    min_achieved_negative_delta: 20,
+                    min_overflow_positive_delta: Some(95),
+                    max_underflow_negative_delta: Some(45),
+                },
+            },
+            base_aggregator: None,
+        };
+        assert_ok!(aggregator_change2.merge_with_previous_aggregator_change(aggregator_change1));
+        assert_eq!(aggregator_change2, AggregatorChange {
+            max_value: 100,
+            state: AggregatorState::Delta {
+                speculative_start_value: SpeculativeStartValue::Unset,
+                delta: DeltaValue::Positive(30),
+                history: DeltaHistory {
+                    max_achieved_positive_delta: 35,
+                    min_achieved_negative_delta: 15,
+                    min_overflow_positive_delta: Some(90),
+                    max_underflow_negative_delta: Some(25),
+                },
+            },
+            base_aggregator: None,
+        });
+    }
+
+    #[test]
+    fn test_merge_delta_into_delta2() {
+        let aggregator_change1 = AggregatorChange {
+            max_value: 100,
+            state: AggregatorState::Delta {
+                speculative_start_value: SpeculativeStartValue::AggregatedValue(70),
+                delta: DeltaValue::Negative(40),
+                history: DeltaHistory {
+                    max_achieved_positive_delta: 20,
+                    min_achieved_negative_delta: 60,
+                    min_overflow_positive_delta: Some(40),
+                    max_underflow_negative_delta: Some(80),
+                },
+            },
+            base_aggregator: None,
+        };
+        let mut aggregator_change2 = AggregatorChange {
+            max_value: 100,
+            state: AggregatorState::Delta {
+                speculative_start_value: SpeculativeStartValue::AggregatedValue(60),
+                delta: DeltaValue::Negative(20),
+                history: DeltaHistory {
+                    max_achieved_positive_delta: 35,
+                    min_achieved_negative_delta: 20,
+                    min_overflow_positive_delta: Some(85),
+                    max_underflow_negative_delta: Some(75),
+                },
+            },
+            base_aggregator: None,
+        };
+        assert_ok!(aggregator_change2.merge_with_previous_aggregator_change(aggregator_change1));
+        assert_eq!(aggregator_change2, AggregatorChange {
+            max_value: 100,
+            state: AggregatorState::Delta {
+                speculative_start_value: SpeculativeStartValue::Unset,
+                delta: DeltaValue::Negative(60),
+                history: DeltaHistory {
+                    max_achieved_positive_delta: 20,
+                    min_achieved_negative_delta: 60,
+                    min_overflow_positive_delta: Some(40),
+                    max_underflow_negative_delta: Some(80),
+                },
+            },
+            base_aggregator: None,
+        });
+        let mut aggregator_change3 = AggregatorChange {
+            max_value: 100,
+            state: AggregatorState::Delta {
+                speculative_start_value: SpeculativeStartValue::AggregatedValue(80),
+                delta: DeltaValue::Positive(5),
+                history: DeltaHistory {
+                    max_achieved_positive_delta: 5,
+                    min_achieved_negative_delta: 5,
+                    min_overflow_positive_delta: Some(91),
+                    max_underflow_negative_delta: Some(95),
+                },
+            },
+            base_aggregator: None,
+        };
+        assert_ok!(aggregator_change3.merge_with_previous_aggregator_change(aggregator_change2));
+        assert_eq!(aggregator_change3, AggregatorChange {
+            max_value: 100,
+            state: AggregatorState::Delta {
+                speculative_start_value: SpeculativeStartValue::Unset,
+                delta: DeltaValue::Negative(55),
+                history: DeltaHistory {
+                    max_achieved_positive_delta: 20,
+                    min_achieved_negative_delta: 65,
+                    min_overflow_positive_delta: Some(31),
+                    max_underflow_negative_delta: Some(80),
+                },
+            },
+            base_aggregator: None,
+        });
+    }
+
+    #[test]
+    fn test_merge_delta_into_delta3() {
+        let aggregator_change1 = AggregatorChange {
+            max_value: 100,
+            state: AggregatorState::Delta {
+                speculative_start_value: SpeculativeStartValue::AggregatedValue(70),
+                delta: DeltaValue::Positive(20),
+                history: DeltaHistory {
+                    max_achieved_positive_delta: 20,
+                    min_achieved_negative_delta: 60,
+                    min_overflow_positive_delta: None,
+                    max_underflow_negative_delta: None,
+                },
+            },
+            base_aggregator: None,
+        };
+        let mut aggregator_change2 = AggregatorChange {
+            max_value: 100,
+            state: AggregatorState::Delta {
+                speculative_start_value: SpeculativeStartValue::AggregatedValue(10),
+                delta: DeltaValue::Negative(5),
+                history: DeltaHistory {
+                    max_achieved_positive_delta: 10,
+                    min_achieved_negative_delta: 5,
+                    min_overflow_positive_delta: Some(95),
+                    max_underflow_negative_delta: None,
+                },
+            },
+            base_aggregator: None,
+        };
+        assert_ok!(aggregator_change2.merge_with_previous_aggregator_change(aggregator_change1));
+        assert_eq!(aggregator_change2, AggregatorChange {
+            max_value: 100,
+            state: AggregatorState::Delta {
+                speculative_start_value: SpeculativeStartValue::Unset,
+                delta: DeltaValue::Positive(15),
+                history: DeltaHistory {
+                    max_achieved_positive_delta: 30,
+                    min_achieved_negative_delta: 60,
+                    min_overflow_positive_delta: None,
+                    max_underflow_negative_delta: None,
+                },
+            },
+            base_aggregator: None,
+        });
+    }
+
+    #[test]
+    fn test_merge_delta_into_delta4() {
+        let aggregator_change1 = AggregatorChange {
+            max_value: 100,
+            state: AggregatorState::Delta {
+                speculative_start_value: SpeculativeStartValue::AggregatedValue(70),
+                delta: DeltaValue::Negative(20),
+                history: DeltaHistory {
+                    max_achieved_positive_delta: 20,
+                    min_achieved_negative_delta: 60,
+                    min_overflow_positive_delta: None,
+                    max_underflow_negative_delta: None,
+                },
+            },
+            base_aggregator: None,
+        };
+        let mut aggregator_change2 = AggregatorChange {
+            max_value: 100,
+            state: AggregatorState::Delta {
+                speculative_start_value: SpeculativeStartValue::LastCommittedValue(70),
+                delta: DeltaValue::Positive(5),
+                history: DeltaHistory {
+                    max_achieved_positive_delta: 10,
+                    min_achieved_negative_delta: 5,
+                    min_overflow_positive_delta: None,
+                    max_underflow_negative_delta: Some(90),
+                },
+            },
+            base_aggregator: None,
+        };
+        assert_ok!(aggregator_change2.merge_with_previous_aggregator_change(aggregator_change1));
+        assert_eq!(aggregator_change2, AggregatorChange {
+            max_value: 100,
+            state: AggregatorState::Delta {
+                speculative_start_value: SpeculativeStartValue::Unset,
+                delta: DeltaValue::Negative(15),
+                history: DeltaHistory {
+                    max_achieved_positive_delta: 20,
+                    min_achieved_negative_delta: 60,
+                    min_overflow_positive_delta: None,
+                    max_underflow_negative_delta: None,
+                },
+            },
+            base_aggregator: None,
+        });
+    }
+}
--- a/aptos-move/aptos-aggregator/src/aggregator_change_set.rs
+++ b/aptos-move/aptos-aggregator/src/aggregator_change_set.rs
@@ -1,4 +1,5 @@
 // Copyright © Aptos Foundation
+// SPDX-License-Identifier: Apache-2.0
 
 use crate::{
     aggregator_extension::{
@@ -47,10 +48,8 @@ impl AggregatorChange {
         }
     }
 
-    /// Applies self on top of previous delta, merging them together. Note
-    /// that the strict ordering here is crucial for catching overflows
-    /// correctly.
-    ///
+    /// Applies self on top of previous change, merging them together. Note
+    /// that the strict ordering here is crucial for catching overflows correctly.
     /// TODO: What happens if the base aggregator is not none?
     pub fn merge_with_previous_aggregator_change(
         &mut self,
--- a/aptos-move/aptos-aggregator/src/aggregator_change_set.rs
+++ b/aptos-move/aptos-aggregator/src/aggregator_change_set.rs
@@ -3,47 +3,47 @@
 
 use crate::{
     aggregator_extension::{
-        validate_history, AggregatorID, AggregatorState, DeltaHistory, SpeculativeStartValue,
+        AggregatorID, DerivedFormula,
     },
-    bounded_math::{abort_error, addition_deltavalue, ok_overflow, ok_underflow},
+    bounded_math::{SignedU128, code_invariant_error, BoundedMath, expect_ok}, delta_math::{DeltaHistory, merge_data_and_delta, merge_two_deltas},
 };
-use aptos_state_view::StateView;
-use move_binary_format::errors::{PartialVMError, PartialVMResult};
-use move_core_types::vm_status::{StatusCode, VMStatus};
+use move_binary_format::errors::PartialVMResult;
 
-/// When merging aggregator changes of two transactions,
-/// unable to merge the histories of the transactions.
-pub(crate) const EMERGE_HISTORIES: u64 = 0x02_0008;
-
-/// Represents a single aggregator change.
-#[derive(Copy, Clone, Debug, PartialEq, Eq)]
-pub struct AggregatorChange {
-    pub max_value: u128,
-    pub state: AggregatorState,
-    // `base_aggregator` is None for Aggregators.
-    // `base_aggregator` is Some(id) for AggregatorSnapshots.
-    pub base_aggregator: Option<AggregatorID>,
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub enum AggregatorChange {
+    Data {
+        value: u128,
+    },
+    StringData {
+        value: Vec<u8>,
+    },
+    AggregatorDelta {
+        delta: SignedU128,
+        max_value: u128,
+        history: DeltaHistory,
+    },
+    SnapshotDelta {
+        delta: SignedU128,
+        base_aggregator: AggregatorID,
+        formula: DerivedFormula,
+    }
 }
 
 impl AggregatorChange {
-    /// TODO: Implement this.
-    pub fn materialize_aggregator(
-        self,
-        _state_view: &dyn StateView,
-        _aggregator_id: &AggregatorID,
-    ) -> anyhow::Result<AggregatorChange, VMStatus> {
-        Ok(self)
-    }
 
     /// Returns the result of applying the AggregatorState to the base value
-    /// Returns error if postcondition is not satisfied.
+    /// Returns error if the aggregator behavior based on the
+    /// speculative_base_value isn't consistent with the provided base value.
     pub fn apply_aggregator_change_to(&self, base: u128) -> PartialVMResult<u128> {
-        match self.state {
-            AggregatorState::Data { value } => Ok(value),
-            AggregatorState::Delta { delta, .. } => {
-                // First, validate if the current delta operation can be applied to the base.
-                validate_history(base, self.max_value, &self.state)?;
-                Ok(addition_deltavalue(base, delta, self.max_value)?)
+        match self {
+            AggregatorChange::Data { value } => Ok(*value),
+            AggregatorChange::StringData { .. } => Err(code_invariant_error("There should be no base value for StringData")),
+            AggregatorChange::AggregatorDelta { delta, history, max_value } => {
+                merge_data_and_delta(base, delta, history, *max_value)
+            },
+            AggregatorChange::SnapshotDelta { delta, .. } => {
+                // History validation should already be done before this call.
+                expect_ok(BoundedMath::new(u128::MAX).unsigned_add_delta(base, delta))
             },
         }
     }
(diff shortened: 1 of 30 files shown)


igor-aptos (approved):

aptos-move/aptos-aggregator/src/aggregator_extension.rs
```diff
@@ -134,20 +175,36 @@ pub struct Aggregator {
...
                     unreachable!("history is not tracked when aggregator knows its value")
                 },
             }
         }
     }
 
+    /// Records overflows in history. Should be called after an addition is unsuccessful
+    /// to record its side-effects.
+    fn record_overflow(&mut self, value: ExtendedU128) {
+        if let Some(history) = self.history.as_mut() {
```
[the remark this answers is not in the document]
vusirikala: When a new aggregator is created, we assign a Data state to it, and set History to None. 
Currently, we record overflows only when the aggregator is in Data state. So, it should be okay even if we don't do this check. I just did the check for the sake of completeness.

aptos-move/aptos-aggregator/src/aggregator_extension.rs
```diff
@@ -94,24 +94,58 @@ impl AggregatorID {
...
+        self.min_overflow_positive = match self.min_overflow_positive {
+            Some(min) => match value {
+                Some(v) => Some(u128::min(min, v)),
+                None => Some(min),
+            },
+            None => value,
+        }
+    }
+
+    fn record_underflow_negative(&mut self, value: Option<u128>) {
```
gelash, line 141: ditto the comments for record_overflow

aptos-move/aptos-aggregator/src/aggregator_extension.rs
```diff
@@ -177,7 +227,14 @@ impl Aggregator {
             },
             AggregatorState::PositiveDelta => {
                 // If positive delta, add directly but also record the state.
-                self.value = addition(self.value, value, self.limit)?;
+                self.value = addition(self.value, value, self.limit).map_err(|err| {
+                    if self.value < u128::MAX - value {
+                        self.record_overflow(Some(self.value + value));
+                    } else {
```
gelash, line 233: This branch isn't needed.

aptos-move/aptos-aggregator/src/aggregator_extension.rs
```diff
@@ -189,14 +246,15 @@ impl Aggregator {
                 if self.value <= value {
                     self.value = subtraction(value, self.value)?;
                     self.state = AggregatorState::PositiveDelta;
+                    if self.value > self.limit {
```
gelash, line 249: I think this is a bug? tell me if I'm wrong:
it seems like there was a silent assumption before that the value won't break the limit (@georgemitenkov ?), but we should never allow the value to be set above the limit, right? Right now, we would record overflow, but actually allow overflow to happen.

If the above is right, then we would have to perform subtraction, check the result vs the limit, if it's below the limit set it, otherwise, don't set it, record overflow, and return an error from try_add?

gelash (changes requested):
If the thing about overflow is true - let's definitely add a test that would fail before fixing (i.e. make sure the value isn't set above the limit).

gelash:
let's also make sure underflow doesn't have similar issues.

aptos-move/aptos-aggregator/src/aggregator_extension.rs
```diff
@@ -94,24 +94,58 @@ impl AggregatorID {
...
     }
 
-    fn record_negative(&mut self, value: u128) {
-        self.min_negative = u128::max(self.min_negative, value);
+    fn record_success_negative(&mut self, value: u128) {
+        self.min_achieved_negative = u128::max(self.min_achieved_negative, value);
+    }
+
+    fn record_overflow_positive(&mut self, value: Option<u128>) {
+        self.min_overflow_positive = match self.min_overflow_positive {
```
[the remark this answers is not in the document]
vusirikala: Here, the problem is if `self.min_overflow_positive` is None, then `self.min_overflow_positive` stays None. The map statement won't change it. I updated the statement now.

aptos-move/aptos-aggregator/src/aggregator_extension.rs
```diff
@@ -94,24 +94,58 @@ impl AggregatorID {
...
+        self.max_achieved_positive = u128::max(self.max_achieved_positive, value);
     }
 
-    fn record_negative(&mut self, value: u128) {
-        self.min_negative = u128::max(self.min_negative, value);
+    fn record_success_negative(&mut self, value: u128) {
+        self.min_achieved_negative = u128::max(self.min_achieved_negative, value);
+    }
+
+    fn record_overflow_positive(&mut self, value: Option<u128>) {
```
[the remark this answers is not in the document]
vusirikala: Nice observation. Fixed it.

aptos-move/aptos-aggregator/src/aggregator_extension.rs
```diff
@@ -189,14 +246,15 @@ impl Aggregator {
                 if self.value <= value {
                     self.value = subtraction(value, self.value)?;
                     self.state = AggregatorState::PositiveDelta;
+                    if self.value > self.limit {
+                        self.record_overflow(Some(self.value));
+                    }
                 } else {
                     self.value = subtraction(self.value, value)?;
```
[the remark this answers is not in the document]
vusirikala: If self.value is -10, value is 100, limit is 50, then subtraction(value, self.value) will succeed, but there is still an overflow. We were not catching that overflow earlier. So, I added this if statement to catch the overflow and record it.

aptos-move/aptos-aggregator/src/aggregator_extension.rs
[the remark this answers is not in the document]
gelash: good catch, then we should use 
https://doc.rust-lang.org/stable/std/option/enum.Option.html#method.map_or
to be more rusty

aptos-move/aptos-aggregator/src/aggregator_extension.rs
vusirikala: You are right. I updated the self.value and then raised an overflow. I should have raised an overflow before updating self.value. Fixing it.

aptos-move/aptos-aggregator/src/aggregator_extension.rs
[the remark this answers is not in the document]
gelash: I think you are responding to the wrong comment?
in this if branch, value is lower than self value and we have ?; in the end, but if ? were to ever fire, we wouldn't record anything. But it will actually never fire, I think, because we are subtracting a smaller value from a larger value.

The other if branch, you correctly recorded the overflow, but you didn't actually disallow the overflow to happen.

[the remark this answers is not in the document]
gelash: oh I saw your comment there, so maybe I misunderstand here. What I was proposing though was this:self.value = subtraction(self.value, value).expect("Subtraction of smaller value from larger value must succeed")

aptos-move/aptos-aggregator/src/aggregator_extension.rs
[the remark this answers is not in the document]
vusirikala: Updated the comment in `unreachable!("")`.

aptos-move/aptos-aggregator/src/aggregator_extension.rs
```diff
@@ -177,7 +221,12 @@ impl Aggregator {
             },
             AggregatorState::PositiveDelta => {
                 // If positive delta, add directly but also record the state.
-                self.value = addition(self.value, value, self.limit)?;
+                self.value = addition(self.value, value, self.limit).map_err(|err| {
+                    if self.value < u128::MAX - value {
```
georgemitenkov, line 225: Maybe let's add a comment why we do not have to record if we overflow u128::MAX?

aptos-move/aptos-aggregator/src/aggregator_extension.rs
```diff
@@ -187,16 +236,21 @@ impl Aggregator {
                 //     1. X <= Y: then the result is +(Y-X)
                 //     2. X  > Y: then the result is -(X-Y)
                 if self.value <= value {
+                    if value - self.value > self.limit {
+                        self.record_overflow(self.value);
```
georgemitenkov, line 240: Shouldn't the overflow value be the difference?

aptos-move/aptos-aggregator/src/aggregator_extension.rs
```diff
@@ -218,13 +272,29 @@ impl Aggregator {
...
+                    // This case doesn't result in an underflow.
+                    // So we are not calling record_underflow here.
                     self.value = subtraction(self.value, value)?;
                 } else {
                     // Check that we can subtract in general: we don't want to
                     // allow -10000 when limit is 10.
                     // TODO: maybe `subtraction` should also know about the limit?
-                    subtraction(self.limit, value)?;
-
+                    subtraction(self.limit, value).map_err(|err| {
```
georgemitenkov, line 282: I don't think we need this anymore. With your check below it holds:
```
value - self.value > self.limit  ==> value > self.limit
```

aptos-move/aptos-aggregator/src/aggregator_extension.rs
```diff
@@ -218,13 +272,29 @@ impl Aggregator {
...
+                    if value - self.value > self.limit {
+                        self.record_underflow(value - self.value);
+                        return Err(abort_error(
+                            format!(
+                                "underflow occurred when subtracting {} from {}",
+                                value, self.value
+                            ),
+                            ESUB_UNDERFLOW,
+                        ));
+                    }
```
georgemitenkov, line 297: let's expect on subtraction one line below that it cannot fail since self.value < value

aptos-move/aptos-aggregator/src/aggregator_extension.rs
```diff
@@ -488,17 +577,127 @@ mod test {
...
+        assert_eq!(
+            aggregator.history.as_ref().unwrap().max_achieved_positive,
+            300
+        );
+        assert_eq!(
+            aggregator.history.as_ref().unwrap().min_overflow_positive,
+            Some(800)
+        );
+
+        // +300 + 400 > 600!x
```
georgemitenkov, line 606: nit: no x in the end

aptos-move/aptos-aggregator/src/delta_change_set.rs
```diff
@@ -16,18 +16,22 @@ use move_binary_format::errors::{Location, PartialVMError, PartialVMResult};
 use std::collections::{btree_map::Entry, BTreeMap};
 
 /// When `Addition` operation overflows the `limit`.
-const EADD_OVERFLOW: u64 = 0x02_0001;
+pub const EADD_OVERFLOW: u64 = 0x02_0001;
```
georgemitenkov, line 19: can be pub(crate) i think?

aptos-move/aptos-aggregator/src/delta_change_set.rs
```diff
@@ -123,8 +140,10 @@ impl DeltaOp {
         // was +99 at some point. Now, if we merge some `d1` which is +2 with `d2`, we get
         // the result is +5. However, it should not have happened because `d2` should hit
         // +2+99 > 100 at some point in history and fail.
-        let shifted_max_positive = self.shifted_max_positive_by(&previous_delta)?;
-        let shifted_min_negative = self.shifted_min_negative_by(&previous_delta)?;
+        let shifted_max_achieved_positive =
```
georgemitenkov, line 143: Cannot comment on the function declaration, how do overflows change as we merge deltas? e.g.:
d1 is +10 and overflows on +300 and d2 is +10 and overflows on +400, we result is +20 and overflows on +300

georgemitenkov: This is actually important I think?

georgemitenkov (dismissed):

aptos-move/aptos-aggregator/src/aggregator_extension.rs
```diff
@@ -177,7 +221,12 @@ impl Aggregator {
             },
             AggregatorState::PositiveDelta => {
                 // If positive delta, add directly but also record the state.
-                self.value = addition(self.value, value, self.limit)?;
+                self.value = addition(self.value, value, self.limit).map_err(|err| {
+                    if self.value < u128::MAX - value {
+                        self.record_overflow(self.value + value);
```
georgemitenkov, line 226: I think we should record limit here?

aptos-move/aptos-aggregator/src/aggregator_extension.rs
vusirikala: Done.

aptos-move/aptos-aggregator/src/delta_change_set.rs
vusirikala: Done.

aptos-move/aptos-aggregator/src/aggregator_extension.rs
vusirikala: What do you mean by recoding limit? `self.value + value` is the value of delta for which the overflow has occurred right?

aptos-move/aptos-aggregator/src/aggregator_extension.rs
vusirikala: Done.

aptos-move/aptos-aggregator/src/aggregator_extension.rs
vusirikala: Done.

aptos-move/aptos-aggregator/src/aggregator_extension.rs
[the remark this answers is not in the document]
vusirikala: Done.

aptos-move/aptos-aggregator/src/aggregator_extension.rs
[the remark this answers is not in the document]
vusirikala: Done.

aptos-move/aptos-aggregator/src/delta_change_set.rs
vusirikala: As we aggregator delta from previous transactions, we don't change the overflow/underflow bounds of the current transaction right?

zekun000: I have the same question, it looks to me we should do a max/min of the four bounds? since squash is used for multiple sessions from the same txn iirc

zekun000: I was discussing with @gelash on this, it looks both used on squash and read, but I think the overflow bounds are important for both cases.
for reads we need to validate after shifting if it triggers overflow or not. for squash we need to take the max bounds I think

gelash: The 4 values stored in the delta with (0, max_value) essentially determine the acceptable pre-image for which the behavior is intact.

Lots of things are missing here:
(1) we need to shift all 4 values of the new delta based on the previous value. this is because the "origin" for the current delta is now shifted.
(2) we then need to take min / max of the shifted bounds and the previous bounds, i.e. if the shifted overflew after adding 103 and the previous overflew after adding 101, then we store 101 in the min_added overflow and so on.
We then need to confirm that the new merged bounds are consistent, i.e. that max_added_no_overflow < max_added_overflow (similarly for the underflow side). 
The reason for this last part is to not assume that the speculative value used for "prediction" inside the session that generated the DeltaOp isn't the same (I believe if it was the same we could prove / assert that the bounds would end up consistent) - let's add a comment about this aspect too.
 
When we generate the new 4 values (if all is successful), We can also assert that the combined delta update value falls in between.

We can rewrite the shift functions as we need to, e.g. those can also do the min/max (but not assign until all the checks pass), right now it seems we do the max when we assign at the end of the function. And in general, feel free to refactor the utility functions, there should be improvement opportunities esp. since the flow and logic is evolving.

gelash (changes requested):
Going in a good direction, but still requires some work.

aptos-move/aptos-aggregator/src/delta_change_set.rs
```diff
@@ -14,22 +14,26 @@ use aptos_types::{
 };
 use move_binary_format::errors::{Location, PartialVMError, PartialVMResult};
 
-/// When `Addition` operation overflows the `limit`.
-const EADD_OVERFLOW: u64 = 0x02_0001;
+/// When `Addition` operation overflows the `max_value`.
```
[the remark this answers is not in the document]
vusirikala: Which comment are you referring to? I couldn't find any comment of the form "every delta is is a stake key (...) and an operation".

aptos-move/framework/src/natives/aggregator_natives/aggregator_v2.rs
```diff
@@ -39,6 +50,194 @@ fn is_string_type(context: &SafeNativeContext, type_arg: &Type) -> SafeNativeRes
...
+            Ok(smallvec![Value::struct_(Struct::pack(vec![Value::u64(
+                id
+            )]))])
+        },
+        _ => Err(SafeNativeError::Abort {
+            abort_code: EUNSUPPORTED_AGGREGATOR_SNAPSHOT_TYPE,
+        }),
+    }
+}
+
```
igor-aptos, line 240: just to note - you'll need to change the functions below to use aggregator data after aggregators are enabled. (but you can leave that for separate PR)

igor-aptos (changes requested):

igor-aptos (approved):
looks good now, can you re-run and recheck if tests are still passing, I haven't tested those?

aptos-move/aptos-aggregator/src/bounded_math.rs
```diff
@@ -0,0 +1,214 @@
...
+            (DeltaValue::Negative(v1), DeltaValue::Positive(v2)) => update_different_sign!(v2, v1),
+            (DeltaValue::Negative(v1), DeltaValue::Negative(v2)) => {
+                DeltaValue::Negative(addition(*v1, *v2, max_value).map_err(|err| {
+                    if err == BoundedMathError::Overflow {
+                        BoundedMathError::Underflow
+                    } else {
+                        err
+                    }
+                })?)
+            },
```
igor-aptos, line 143: let's create a function to negate_error(BoundedMathResult), and invert the other side too, i.e:

```
fn negate_error<T>(result: BoundedMathResult<T>): BoundedMathResult<T> {
  result.map_err(|err| {
    match err {
      BoundedMathError::Overflow => BoundedMathError::Underflow,
      BoundedMathError::Underflow => BoundedMathError::Overflow,
    }
  }
}

```

igor-aptos (approved):
thanks, looks good to me!

aptos-move/aptos-aggregator/src/delta_change_set.rs
```diff
@@ -14,22 +14,26 @@ use aptos_types::{
...
+    max_achieved_positive: u128,
     /// Smallest negative delta seen during execution.
-    min_negative: u128,
-    /// Postcondition: delta overflows on exceeding this limit or going below
+    min_achieved_negative: u128,
+    /// Smallest positive delta observed during execution that resulted in an overflow.
+    min_overflow_positive: Option<u128>,
+    /// Maximum negative delta observed during execution that result in an underflow.
+    max_underflow_negative: Option<u128>,
+    /// Postcondition: delta overflows on exceeding this max_value or going below
```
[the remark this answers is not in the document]
vusirikala: Done.

aptos-move/e2e-move-tests/src/aggregator_v2.rs
```diff
@@ -230,3 +231,68 @@ pub fn verify_string_snapshot_concat(
...
+
+pub fn try_add_and_read_snapshot_u128(
+    harness: &mut MoveHarness,
+    account: &Account,
+    index: u64,
+    value: u128,
+) -> SignedTransaction {
+    harness.create_entry_function(
+        account,
+        str::parse("0x1::aggregator_v2_test::try_add_and_read_snapshot").unwrap(),
```
georgemitenkov, line 291: I can only see `try_add_snapshot` in the module?

aptos-move/e2e-move-tests/src/tests/aggregator_v2.rs
```diff
@@ -188,3 +189,21 @@ fn test_aggregator_materialize_overflow() {
...
+fn test_aggregator_snapshot() {
+    let (mut h, acc) = setup();
+    let txn1 = new(&mut h, &acc, 0, 400);
+    let txn2 = snapshot(&mut h, &acc, 0);
+    let txn3 = snapshot_with_u64_limit(&mut h, &acc, 0);
+    let txn4 = read_snapshot_u128(&mut h, &acc, 0);
+    let txn5 = read_snapshot_u64(&mut h, &acc, 0);
+    let txn6 = try_add_and_read_snapshot_u128(&mut h, &acc, 0, 100);
+
+    assert_success!(h.run(txn1));
```
georgemitenkov, line 203: How does this work? If we do not have pieces connected together?

For the future: once we connect end-to-end we should run these txns as a block, not one by one to test

aptos-move/mvhashmap/src/versioned_aggregators.rs
```diff
@@ -483,11 +485,21 @@ mod test {
 
     // For compactness, in tests where the Delta contents do not matter.
     fn test_delta() -> DeltaOp {
-        DeltaOp::new(DeltaUpdate::Plus(30), 1000, 30, 0)
+        DeltaOp::new(DeltaValue::Positive(30), 1000, DeltaHistory {
```
georgemitenkov, line 488: Let's add a constructor for delta history so that we do not have to add None everywhere

third_party/move/extensions/move-table-extension/src/lib.rs
```diff
@@ -42,7 +42,7 @@ use std::{
 /// The representation of a table handle. This is created from truncating a sha3-256 based
 /// hash over a transaction hash provided by the environment and a table creation counter
 /// local to the transaction.
-#[derive(Copy, Clone, Debug, PartialOrd, Ord, PartialEq, Eq)]
+#[derive(Copy, Clone, Debug, Hash, PartialOrd, Ord, PartialEq, Eq)]
```
georgemitenkov, line 45: I think it might be better to switch to aptos_table_extension::TableHandle and derive Hash there

aptos-move/aptos-aggregator/src/aggregator_change_set.rs
```diff
@@ -0,0 +1,507 @@
...
+    },
+    bounded_math::{abort_error, addition_deltavalue, ok_overflow, ok_underflow},
+};
+use aptos_state_view::StateView;
+use move_binary_format::errors::{PartialVMError, PartialVMResult};
+use move_core_types::vm_status::{StatusCode, VMStatus};
+
+/// When merging aggregator changes of two transactions,
+/// unable to merge the histories of the transactions.
+pub(crate) const EMERGE_HISTORIES: u64 = 0x02_0008;
```
georgemitenkov, line 15: We can keep this for now, but this error should never be propagated to the user. I think I mentioned during one of the syncs about aggregator errors, and this is somewhat what I meant (with what @igor-aptos also partially added in bounded math)

```
enum AggregatorError {
  UnmergableHistories,
  ....
}
```
so we can use this over VMStatus. But let's revisit this later👍

aptos-move/aptos-aggregator/src/aggregator_change_set.rs
```diff
@@ -0,0 +1,507 @@
...
+    // `base_aggregator` is None for Aggregators.
+    // `base_aggregator` is Some(id) for AggregatorSnapshots.
+    pub base_aggregator: Option<AggregatorID>,
+}
+
+impl AggregatorChange {
+    /// TODO: Implement this.
+    pub fn materialize_aggregator(
+        self,
+        _state_view: &dyn StateView,
```
georgemitenkov, line 31: dyn --> impl

aptos-move/aptos-aggregator/src/aggregator_change_set.rs
```diff
@@ -0,0 +1,507 @@
...
+        _aggregator_id: &AggregatorID,
+    ) -> anyhow::Result<AggregatorChange, VMStatus> {
+        Ok(self)
```
georgemitenkov, line 34: why not `unimplemented!()`?

aptos-move/aptos-aggregator/src/aggregator_change_set.rs
```diff
@@ -0,0 +1,507 @@
...
+
+    /// Returns the result of applying the AggregatorState to the base value
+    /// Returns error if postcondition is not satisfied.
+    pub fn apply_aggregator_change_to(&self, base: u128) -> PartialVMResult<u128> {
+        match self.state {
+            AggregatorState::Data { value } => Ok(value),
+            AggregatorState::Delta { delta, .. } => {
+                // First, validate if the current delta operation can be applied to the base.
+                validate_history(base, self.max_value, &self.state)?;
+                Ok(addition_deltavalue(base, delta, self.max_value)?)
```
georgemitenkov, line 45: I think mentioned above, calling it `apply` or similar would make more sense :)

aptos-move/aptos-aggregator/src/aggregator_change_set.rs
```diff
@@ -0,0 +1,507 @@
...
+    pub fn merge_with_previous_aggregator_change(
+        &mut self,
+        previous_change: AggregatorChange,
+    ) -> PartialVMResult<()> {
+        assert_eq!(
+            self.max_value, previous_change.max_value,
+            "Cannot merge aggregator changes with different max_values",
+        );
+        // When the previous aggregator change is a snapshot, we cannot merge it with the current.
+        if previous_change.base_aggregator.is_some() {
```
georgemitenkov, line 64: We can factor out `base_aggregator.is_some()` as `is_snapshot()`

georgemitenkov (commented):
Nice! I started having a look, pushing some minor comments for now

aptos-move/aptos-aggregator/src/aggregator_change_set.rs
igor-aptos: in the following diff, snapshot might also not have a base aggregator - for example if we call aggregator_v2::create_snapshot(value)

aptos-move/aptos-aggregator/src/bounded_math.rs
```diff
@@ -0,0 +1,215 @@
...
+                DeltaValue::Negative(negate_error(addition(*v1, *v2, max_value))?)
+            },
+        })
+    }
+}
+
+/// Implements base + value
+pub fn addition_deltavalue(
+    base: u128,
+    value: DeltaValue,
```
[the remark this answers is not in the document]
igor-aptos: how about we call it what it is - SignedU128 ? :)

we can then in our structs have: 
delta: SignedU128,

aptos-move/aptos-aggregator/src/aggregator_extension.rs
```diff
@@ -96,206 +127,442 @@ impl AggregatorID {
...
-    pub fn try_add(&mut self, value: u128) -> PartialVMResult<()> {
+    /// Returns error if transaction is in invalid state, and should be re-executed.
+    /// Returns true if addition succeeded, and false if it would overflow.
+    pub fn try_add(
+        &mut self,
+        resolver: &dyn AggregatorResolver,
+        input: u128,
+    ) -> PartialVMResult<bool> {
+        if input > self.max_value {
+            // we do not have to record the overflow.
```
georgemitenkov, line 327: nit: We

aptos-move/aptos-aggregator/src/aggregator_extension.rs
```diff
@@ -96,206 +127,442 @@ impl AggregatorID {
...
+                    let overflow_value = expect_ok(ok_overflow(addition_deltavalue(
+                        input,
+                        delta,
+                        self.max_value,
+                    )))?;
+
+                    // if value overflowed, we don't need to record it
+                    if let Some(overflow_value) = overflow_value {
+                        self.get_mut_history()
+                            .unwrap()
```
georgemitenkov, line 366: nit: expect?

aptos-move/aptos-aggregator/src/aggregator_extension.rs
```diff
@@ -96,206 +127,442 @@ impl AggregatorID {
...
+                ))?;
+
+                if addition(cur_value, input, self.max_value).is_err() {
+                    let overflow_value = expect_ok(ok_overflow(addition_deltavalue(
+                        input,
+                        delta,
+                        self.max_value,
+                    )))?;
+
+                    // if value overflowed, we don't need to record it
```
georgemitenkov, line 363: nit: . missing, capitalize

aptos-move/aptos-aggregator/src/aggregator_extension.rs
```diff
@@ -96,206 +127,442 @@ impl AggregatorID {
...
 
     /// Implements logic for subtracting from an aggregator.
-    pub fn try_sub(&mut self, value: u128) -> PartialVMResult<()> {
+    pub fn try_sub(
+        &mut self,
+        resolver: &dyn AggregatorResolver,
+        input: u128,
+    ) -> PartialVMResult<bool> {
+        if input > self.max_value {
+            // we do not have to record the underflow.
```
georgemitenkov, line 392: nit: duplicated comment

aptos-move/aptos-aggregator/src/aggregator_extension.rs
```diff
@@ -96,206 +127,442 @@ impl AggregatorID {
...
-    /// result, the aggregator knows it value (i.e. its state changes to
-    /// `Data`).
-    pub fn read_and_materialize(
+    /// Implements logic for doing a "cheap read" of an aggregator.
+    /// This means that we query the MVHashmap for the last committed value
+    /// of the aggregator, and store it in the `speculative_start_value` with
+    /// `LastCommittedValue` variant.
+    pub fn read_last_committed_aggregator_value(
         &mut self,
         resolver: &dyn AggregatorResolver,
```
georgemitenkov, line 455: let's have this impl?

aptos-move/aptos-aggregator/src/aggregator_extension.rs
```diff
@@ -96,206 +127,442 @@ impl AggregatorID {
...
+                if speculative_start_value != SpeculativeStartValue::Unset {
+                    return Ok(addition_deltavalue(
+                        speculative_start_value.get_value(),
+                        delta,
+                        self.max_value,
+                    )?);
+                }
+                // Otherwise, we have to go to storage and read the value.
+                let value_from_storage = resolver
+                    .resolve_aggregator_value(&self.id, AggregatorReadMode::Speculative)
```
georgemitenkov, line 477: reading the last committed is not speculative? Should it be Mode::Precise here?

georgemitenkov: Nvm, I see the second function. To me committed sounds pretty non-speculative 😂 and most recent sounds speculative instead

aptos-move/aptos-aggregator/src/aggregator_extension.rs
```diff
@@ -96,206 +127,442 @@ impl AggregatorID {
...
+                        extension_error(format!(
+                            "Could not find the value of the aggregator: {}",
+                            e
+                        ))
+                    })?;
+
+                // Validate history and apply the delta.
+                validate_history(value_from_storage, self.max_value, &self.state)?;
+                self.state = AggregatorState::Delta {
+                    speculative_start_value: SpeculativeStartValue::AggregatedValue(
```
georgemitenkov, line 548: Let's add constructors

aptos-move/aptos-aggregator/src/bounded_math.rs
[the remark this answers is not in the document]
igor-aptos: seems like sharding also is adding similar class - https://github.com/aptos-labs/aptos-core/pull/9882/files#diff-af089d9a125f77b0177a0d2722c9e46d98aaf4d95650386195606481e59c7409

suggesting we should call it generically (SignedU128) and be able to use it in both places

aptos-move/aptos-aggregator/src/bounded_math.rs
```diff
@@ -0,0 +1,215 @@
...
+
+pub fn ok_underflow<T>(value: BoundedMathResult<T>) -> BoundedMathResult<Option<T>> {
+    match value {
+        Ok(value) => Ok(Some(value)),
+        Err(BoundedMathError::Overflow) => Err(BoundedMathError::Overflow),
+        Err(BoundedMathError::Underflow) => Ok(None),
+    }
+}
+
+pub fn expect_ok<T>(value: BoundedMathResult<T>) -> PartialVMResult<T> {
```
gelash, line 68: I thought this method is never used in the speculative context, but I do see that it just map the bounded math error. In this case, error! here is wrong, as it will log errors for cases that happened during speculative execution (i.e. with wrong speculative value). If we want to log, what we need to use is the speculative logging crate, but that needs transaction index. So it'd have to be up in the chain of calls, as long as we can figure out an aggregator error happened, we can log way up the call chain? @igor-aptos @vusirikala 
  
If this was intended to only be used when the operations are supposed to succeed, then I'm not sure about the uses below (I see uses in try_add, try_sub methods). I think it's the same question for me to understand the difference of how ECODE_INVARIANT_BROKEN is used.

igor-aptos: this is called for cases that should never happen, if code is valid, even in speculative context.

For example, we are maintaining that speculative_start_value + delta is always in the [0, max_value] range. We check that before updating either of the two. 
If that is broken at any point, ECODE_INVARIANT_BROKEN will be thrown.

gelash: I think it was the use at line e.g. 357 that confused me, adding the new input, but now I think I figured, ok_overflow actually is okay with an overflow, and then we check we are okay :)

aptos-move/aptos-aggregator/src/aggregator_change_set.rs
```diff
@@ -0,0 +1,507 @@
...
+                            match (
+                                adjusted_min_overflow_positive_delta,
+                                prev_history.min_overflow_positive_delta,
+                            ) {
+                                (Some(a), Some(b)) => Some(u128::min(a, b)),
+                                (a, b) => a.or(b),
+                            }
+                        };
+
+                        let new_max_underflow = {
```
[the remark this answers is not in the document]
gelash: and also basic unit tests.
Should this be a part of boundedmath? @igor-aptos

aptos-move/aptos-aggregator/src/aggregator_change_set.rs
```diff
@@ -0,0 +1,507 @@
...
+        }
+        match self.state {
+            AggregatorState::Data { .. } => {
+                // If the current state is Data, then merging with previous state won't change anything.
+                return Ok(());
+            },
+            AggregatorState::Delta { delta, history, .. } => {
+                match previous_change.state {
+                    AggregatorState::Data { value } => {
+                        // When prev_state is Data { value }, and current state is Delta { delta }, merging them into Data { value + delta }
```
gelash, line 78: nit: some comment lines seem to short, while some others seem too long (like this one)

aptos-move/aptos-aggregator/src/aggregator_change_set.rs
```diff
@@ -0,0 +1,507 @@
...
+                                prev_history.max_underflow_negative_delta,
+                            ) {
+                                (Some(a), Some(b)) => Some(u128::min(a, b)),
+                                (a, b) => a.or(b),
+                            }
+                        };
+
+                        // new_max_achieved = max(prev_max_achieved, max_achieved + prev_delta)
+                        // When adjusting max_achieved, if underflow - than the other is bigger,
+                        // but if overflow - we fail the merge, as we cannot successfully achieve delta larger than max_value.
```
gelash, line 141: nit: and this one (too long? - much longer than code around it.

aptos-move/aptos-aggregator/src/aggregator_change_set.rs
```diff
@@ -0,0 +1,507 @@
...
+                        let new_max_achieved = ok_underflow(addition_deltavalue(
```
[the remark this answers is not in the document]
gelash: Indeed, also for me same comment as above: can we expose this at a library level.

aptos-move/aptos-aggregator/src/aggregator_change_set.rs
```diff
@@ -0,0 +1,507 @@
...
+            base_aggregator: None,
+        };
+
+        assert_ok!(aggregator_change2.merge_with_previous_aggregator_change(aggregator_change1));
+        assert_eq!(aggregator_change2, AggregatorChange {
+            max_value: 100,
+            state: AggregatorState::Data { value: 30 },
+            base_aggregator: None,
+        });
+        assert_err!(aggregator_change3.merge_with_previous_aggregator_change(aggregator_change2));
```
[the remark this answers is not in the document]
gelash: hm, isn't it +40, max +80, and min -5 which is under the limit so can't be achieved? Ah, @georgemitenkov, note that Aggregator2 changed after merging.

However, the test merges delta with previous data, name merge data into delta is ambiguous (into can be both directions)

aptos-move/aptos-aggregator/src/aggregator_extension.rs
```diff
@@ -13,22 +16,50 @@ use move_binary_format::errors::{PartialVMError, PartialVMResult};
 use move_core_types::account_address::AccountAddress;
 use std::collections::{BTreeMap, BTreeSet};
 
+/// Describes how the `speculative_start_value` in
```
gelash, line 19: nit: these two comment lines can be merged.
also maybe "was obtained" instead of "is obtained" (also in comments below)

aptos-move/aptos-aggregator/src/aggregator_extension.rs
```diff
@@ -13,22 +16,50 @@ use move_binary_format::errors::{PartialVMError, PartialVMResult};
...
+/// `AggregatorState` is obtained.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum SpeculativeStartValue {
+    // The speculative_start_value is assigned to 0.
+    Unset,
+    // The speculative_start_value is obtained by reading
+    // the last committed value of the aggregator from MVHashmap.
+    LastCommittedValue(u128),
+    // The speculative_start_value is obtained by aggregating all
```
gelash, line 28: No need to say anything about MVHashMap here (separate abstraction), we can instead say "by performing a read procedure on the aggregator, which involves aggregating deltas present at the read time"

aptos-move/aptos-aggregator/src/aggregator_extension.rs
```diff
@@ -13,22 +16,50 @@ use move_binary_format::errors::{PartialVMError, PartialVMResult};
...
+    // the previous deltas of the aggregator from MVHashmap that
+    // were present at the fetch time.
+    AggregatedValue(u128),
+}
+
+impl SpeculativeStartValue {
+    pub fn get_value(&self) -> u128 {
+        match self {
+            SpeculativeStartValue::Unset => 0,
```
gelash, line 37: This can be dangerous, I believe this method should be returning Result<u128> and have the caller explicitly expect, to make sure it's not unset when it's supposed to be set.

aptos-move/aptos-aggregator/src/aggregator_change_set.rs
```diff
@@ -0,0 +1,507 @@
...
+            base_aggregator: None,
+        };
+        let mut aggregator_change2 = AggregatorChange {
+            max_value: 100,
+            state: AggregatorState::Delta {
+                speculative_start_value: SpeculativeStartValue::AggregatedValue(60),
+                delta: DeltaValue::Negative(20),
+                history: DeltaHistory {
+                    max_achieved_positive_delta: 35,
+                    min_achieved_negative_delta: 20,
```
gelash, line 361: isn't it better to test when 20+40 != 60, i.e. wouldn't it be choosing between 60 and 60 here.

aptos-move/aptos-aggregator/src/aggregator_extension.rs
```diff
@@ -96,206 +127,442 @@ impl AggregatorID {
 ///
 /// TODO: while we support tracking of the history, it is not yet fully used on
 /// executor side because we don't know how to throw errors.
```
gelash, line 129: We need to update the picture above to include the overflow bounds.
We probably don't need the TODO, as we can't avoid having to resolve it as a part of current PRs (and hard to remember to remove the TODO). It'd make sense to keep it if it'd be unaddressed.

aptos-move/aptos-aggregator/src/aggregator_extension.rs
```diff
@@ -96,206 +127,442 @@ impl AggregatorID {
...
+    pub min_overflow_positive_delta: Option<u128>,
+    // `max_underflow_negative_delta` is None in two possible cases:
+    // 1. No underflow occured in the try_add/try_sub functions throughout the
+    // transaction execution.
+    // 2. The only underflows that occured in the try_add/try_sub functions in
+    // this transaction execution are with delta that drops below -limit.
+    pub max_underflow_negative_delta: Option<u128>,
+}
+
+impl Default for DeltaHistory {
```
gelash, line 148: Can't this be derived for DeltaHistory?

aptos-move/aptos-aggregator/src/aggregator_extension.rs
```diff
@@ -96,206 +127,442 @@ impl AggregatorID {
...
+            DeltaValue::Negative(value) => {
+                self.min_achieved_negative_delta =
+                    u128::max(self.min_achieved_negative_delta, value)
+            },
+        }
     }
 
-    fn record_negative(&mut self, value: u128) {
-        self.min_negative = u128::max(self.min_negative, value);
+    /// Records overflows in history. Should be called after an addition is unsuccessful
```
gelash, line 179: This and the next one an be a single method, i.e. record_failure to mirror record_success, on &mut u128, and the caller can apply it to min_overflow_positive_delta and max_underflow_negative_delta members.

aptos-move/aptos-aggregator/src/aggregator_extension.rs
```diff
@@ -96,206 +127,442 @@ impl AggregatorID {
...
+    max_value: u128,
+    state: &AggregatorState,
+) -> PartialVMResult<()> {
+    if let AggregatorState::Delta {
+        speculative_start_value,
+        delta: _,
+        history,
+    } = state
+    {
+        // We need to make sure the following 4 conditions are satisified.
```
gelash, line 231: great comment.

aptos-move/aptos-aggregator/src/aggregator_extension.rs
```diff
@@ -96,206 +127,442 @@ impl AggregatorID {
...
+
+        if history.min_overflow_positive_delta.is_some()
+            && base_value <= max_value - history.min_overflow_positive_delta.unwrap()
+        {
+            return Err(abort_error(
+                format!("Overflow was expected when setting the aggreagator start value to {}. Previous speculative start value = {:?}, Min overflow delta = {}, Max value = {}", base_value, speculative_start_value, history.min_overflow_positive_delta.unwrap(), max_value),
+                EEXPECTED_OVERFLOW,
+            ));
+        }
+        if history.max_underflow_negative_delta.is_some()
```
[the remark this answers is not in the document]
gelash: #[stable(feature = "is_some_and", since = "1.70.0")]
impressive being up-to-date, @georgemitenkov

gelash (commented):
first batch of comments.

aptos-move/aptos-aggregator/src/bounded_math.rs
igor-aptos: expect_ok(ok_overflow()) asserts there is no underflow

igor-aptos:
I128 is 128 bits. SignedU128 is 129 bits. 

We are not using i128 exactly because we need the extra bit, if we didn't - we would just use i128.
So I think naming is appropriate. We could call it I129 if you prefer, but it is slightly different, as 0 has two values, and min is one off.


For math, it doesn't make sense to have limit inside those values, as it needs to be always the same. We can have 
struct BoundedMath{max_value: u128}, with ask utility functions being its members though. 

But let's land as is, for this part, and then follow up with refactoring

aptos-move/aptos-aggregator/src/aggregator_extension.rs
```diff
@@ -323,20 +592,23 @@ impl AggregatorData {
...
-        limit: u128,
+        max_value: u128,
     ) -> PartialVMResult<&mut Aggregator> {
         let aggregator = self.aggregators.entry(id).or_insert(Aggregator {
-            value: 0,
-            state: AggregatorState::PositiveDelta,
-            limit,
-            history: Some(History::new()),
+            id,
+            state: AggregatorState::Delta {
```
gelash, line 606: this could also be a special default_delta constructor.

aptos-move/aptos-aggregator/src/aggregator_extension.rs
```diff
@@ -346,22 +618,22 @@ impl AggregatorData {
...
+            state: AggregatorState::Data { value: 0 },
+            max_value,
         };
         self.aggregators.insert(id, aggregator);
         self.new_aggregators.insert(id);
     }
 
     /// If aggregator has been used in this transaction, it is removed. Otherwise,
     /// it is marked for deletion.
+    /// TODO: Should we return an error if aggregator id is v2?
```
gelash, line 636: @georgemitenkov where is this method even used? I only see uses in tests? (then we can gate it w. cfg too)

aptos-move/aptos-aggregator/src/aggregator_extension.rs
```diff
@@ -376,6 +648,39 @@ impl AggregatorData {
...
+                history: _,
+            } => Ok(addition_deltavalue(
+                speculative_start_value.get_value(),
+                delta,
+                snapshot.max_value,
+            )?),
+        }
+    }
+
     pub fn generate_id(&mut self) -> u64 {
```
gelash, line 684: should we also connect to the resolver? (since we did connect read interfaces)

aptos-move/aptos-aggregator/src/aggregator_extension.rs
```diff
@@ -376,6 +648,39 @@ impl AggregatorData {
         }
     }
 
+    pub fn snapshot(&mut self, id: &AggregatorID) -> u64 {
+        let new_id = self.generate_id();
+        let snapshot_id = AggregatorID::ephemeral(new_id);
+        let aggregator = self.aggregators.get(id).expect("Aggregator doesn't exist");
```
gelash, line 654: Should / can we have fallback here as well (and similar places), i.e. return the CODE_ERROR like in expect_ok (maybe customized for ID lookups), since at the end of the day on the executor side we can fall back to aggregator-less execution eventually, so we would be tolerate an ID related bug where look-ups failed. @igor-aptos - what's your thoughts too

aptos-move/aptos-aggregator/src/aggregator_extension.rs
```diff
@@ -96,206 +127,442 @@ impl AggregatorID {
...
-            .expect("History should be set for validation");
+    #[allow(dead_code)]
+    fn get_value(&self) -> BoundedMathResult<u128> {
+        match self.state {
+            AggregatorState::Data { value } => Ok(value),
+            AggregatorState::Delta {
+                speculative_start_value,
+                delta,
+                history: _,
+            } => addition_deltavalue(speculative_start_value.get_value(), delta, self.max_value),
```
gelash, line 298: connecting to the conversation re: (.get value of speculative value and read snapshot example) - .get_value here should probably not be able to return "last committed" value, as that is not registered as a read and validated.

gelash (changes requested):
I think we have a serious logic bug - we can't have anything other than try_add try_sub outcome depend on the LastCommitted speculative value - as those aren't validated (the try_add try_sub behavior is)

aptos-move/aptos-aggregator/src/aggregator_change_set.rs
```diff
@@ -0,0 +1,507 @@
...
+                        self.state = AggregatorState::Data {
+                            value: addition_deltavalue(value, delta, self.max_value)?,
+                        };
+                    },
+                    AggregatorState::Delta {
+                        speculative_start_value: _,
+                        delta: prev_delta,
+                        history: prev_history,
+                    } => {
+                        let new_delta = prev_delta.add(&delta, self.max_value)?;
```
gelash, line 89: after the code below is re-structured, let's also do it in a way that we re-use code from here (everything is based on prev_delta, prev_history, self delta and self history & max value) to also do merging for DeltaOps? I'll add a comment there as well

aptos-move/aptos-aggregator/src/delta_change_set.rs
```diff
@@ -64,54 +51,53 @@ impl DeltaOp {
...
-            // new maximum value is M+V provided addition do no overflow.
-            DeltaUpdate::Plus(value) => addition(value, self.max_positive, self.limit),
-            // Suppose that maximum value seen is +M and we shift by -V this time.
-            // If M >= V, the result is +(M-V). Otherwise, `self` should have never
-            // reached any positive value. By convention, we use 0 for the latter
-            // case. Also, we can reuse `subtraction` which throws an error when M < V,
-            // simply mapping the error to 0.
-            DeltaUpdate::Minus(value) => Ok(subtraction(self.max_positive, value).unwrap_or(0)),
-        }
+    fn shifted_max_achieved_positive_delta_by(&self, delta: &DeltaOp) -> PartialVMResult<u128> {
```
gelash, line 68: Let's re-use the code for all this and below (shifting, merging) things from AggregatorChange merging?(merge_with_previous_aggregator_change), I added a comment there as well.
@igor-aptos might be related to the discussion of what types to keep (e.g. if we unset speculative value, maybe we can get rid of DeltaOp altogether - at least for V2? - may end up ugly though to have an enum of AggregatorChange or old DeltaOp, so I'm actually fine as is as long as we share the code)

aptos-move/aptos-aggregator/src/delta_change_set.rs
```diff
@@ -272,37 +208,50 @@ pub fn serialize(value: &u128) -> Vec<u8> {
 }
 
 // Helper for tests, #[cfg(test)] doesn't work for cross-crate.
-pub fn delta_sub(v: u128, limit: u128) -> DeltaOp {
-    DeltaOp::new(DeltaUpdate::Minus(v), limit, 0, v)
+pub fn delta_sub(v: u128, max_value: u128) -> DeltaOp {
+    DeltaOp::new(DeltaValue::Negative(v), max_value, DeltaHistory {
```
gelash, line 212: DeltaHistory constructor

aptos-move/aptos-aggregator/src/delta_change_set.rs
```diff
@@ -272,37 +208,50 @@ pub fn serialize(value: &u128) -> Vec<u8> {
...
+        min_overflow_positive_delta: None,
+        max_underflow_negative_delta: None,
+    })
 }
 
 // Helper for tests, #[cfg(test)] doesn't work for cross-crate.
-pub fn delta_add(v: u128, limit: u128) -> DeltaOp {
-    DeltaOp::new(DeltaUpdate::Plus(v), limit, v, 0)
+pub fn delta_add(v: u128, max_value: u128) -> DeltaOp {
+    DeltaOp::new(DeltaValue::Positive(v), max_value, DeltaHistory {
```
gelash, line 222: Ditto Deltahistory constructor

aptos-move/aptos-aggregator/src/delta_change_set.rs
```diff
@@ -213,53 +175,27 @@ impl DeltaOp {
...
         match self.update {
-            DeltaUpdate::Plus(value) => {
+            DeltaValue::Positive(value) => {
                 write!(
                     f,
                     "+{} ensures 0 <= result <= {}, range [-{}, {}]",
-                    value, self.limit, self.min_negative, self.max_positive
+                    value,
+                    self.max_value,
+                    self.history.min_achieved_negative_delta,
```
gelash, line 187: Let's implement Debug for history, and {:?} history directly here, also will make sure it shows the new fields.

aptos-move/aptos-aggregator/src/delta_change_set.rs
```diff
@@ -213,53 +175,27 @@ impl DeltaOp {
...
+                    self.history.max_achieved_positive_delta
                 )
             },
-            DeltaUpdate::Minus(value) => {
+            DeltaValue::Negative(value) => {
                 write!(
                     f,
                     "-{} ensures 0 <= result <= {}, range [-{}, {}]",
-                    value, self.limit, self.min_negative, self.max_positive
+                    value,
```
gelash, line 195: ditto formatting history and not missing new fields.

aptos-move/aptos-aggregator/src/delta_change_set.rs
```diff
@@ -64,54 +51,53 @@ impl DeltaOp {
...
+            delta.update.minus(),
+            self.max_value,
+        ))
+        .map(|result| result.unwrap_or(0))?)
     }
 
     /// Applies self on top of previous delta, merging them together. Note
     /// that the strict ordering here is crucial for catching overflows
     /// correctly.
     pub fn merge_with_previous_delta(&mut self, previous_delta: DeltaOp) -> PartialVMResult<()> {
```
gelash, line 98: ditto merging re-use with AggregatorChange

gelash (commented):
more comments (another file) :)

aptos-move/aptos-vm-types/src/change_set.rs
```diff
@@ -185,66 +195,115 @@ impl VMChangeSet {
...
+            resource_write_set,
+            module_write_set,
+            aggregator_v1_write_set,
+            aggregator_v1_delta_set: HashMap::new(),
+            aggregator_v2_change_set,
+            events,
+        })
+    }
+
+    pub fn try_materialize_aggregator_v2_changes(
```
gelash, line 255: I started a discussion about this (not sure how to handle txn simulation the best way), but let's delete this method for now and do it in separate PRs.

aptos-move/aptos-vm-types/src/output.rs
```diff
@@ -74,12 +74,17 @@ impl VMOutput {
...
+                && self.change_set().aggregator_v2_change_set().is_empty())
+        {
             return Ok(self);
         }
 
         let (change_set, fee_statement, status) = self.unpack_with_fee_statement();
-        let materialized_change_set = change_set.try_materialize(state_view)?;
+        let materialized_change_set = change_set
+            .try_materialize_aggregator_v1_delta_set(state_view)?
+            .try_materialize_aggregator_v2_changes(state_view)?;
```
gelash, line 87: when we return the actual output from block executor, we need all V2 changes cleared and the write outputs to be updated / replaced properly, at least that's the current expectation in the code. this isn't accomplished atm, so let's delete the try_materialize_v2_changes and revisit /address after this PR lands (as a part of materialization effort)

third_party/move/extensions/move-table-extension/src/lib.rs
igor-aptos: leaving for separate PR

igor-aptos (commented):
I've addressed all comments (I think), except for the constructors.

It seems better to me to be verbose, and make sure values are used correctly, than to potentially swap order of values (like the 4 values in delta history

but we can revisit on that in later PRs

aptos-move/aptos-aggregator/src/aggregator_change_set.rs
```diff
@@ -0,0 +1,320 @@
...
+    },
+    StringData {
+        value: Vec<u8>,
+    },
+    AggregatorDelta {
+        delta: SignedU128,
+        max_value: u128,
+        history: DeltaHistory,
+    },
+    SnapshotDelta {
```
gelash, line 25: nit: should we add a comment for SnapshotDelta, to convey the delta is from the base_aggregator (and the correctness of the delta is checked by base aggregator)

aptos-move/aptos-aggregator/src/aggregator_change_set.rs
```diff
@@ -0,0 +1,507 @@
...
+        std::mem::swap(self, &mut prev_change);
+
+        // Perform the merge.
+        self.merge_with_previous_aggregator_change(prev_change)?;
+        Ok(())
+    }
+}
+
+#[cfg(test)]
+mod test {
```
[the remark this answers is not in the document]
gelash: We can compile the list of "hardening" tasks that we do after code complete

aptos-move/aptos-aggregator/src/aggregator_extension.rs
```diff
@@ -66,236 +57,344 @@ impl AggregatorID {
...
-pub struct History {
-    pub max_positive: u128,
-    pub min_negative: u128,
+/// Describes how the `speculative_start_value` in `AggregatorState` was obtained.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum SpeculativeStartValue {
+    // The speculative_start_value is assigned to 0.
+    Unset,
+    // The speculative_start_value was obtained by reading
+    // the last committed value of the aggregator from MVHashmap.
```
gelash, line 66: maybe let's add a BIG warning here or in the struct comments, that the LastCommittedValue should only be used for branch prediction purposes (try_add / try_sub), or else with parallel execution we may get incorrect / non-deterministic results.

aptos-move/aptos-aggregator/src/aggregator_extension.rs
```diff
@@ -66,236 +57,344 @@ impl AggregatorID {
...
 
-impl History {
-    fn new() -> Self {
-        History {
-            max_positive: 0,
-            min_negative: 0,
+impl SpeculativeStartValue {
+    pub fn get_any_value(&self) -> PartialVMResult<u128> {
+        match self {
+            SpeculativeStartValue::Unset => Err(
```
gelash, line 77: The comment says that "speculative start value is assigned to 0", but I feel I do like not returning 0 here better. Should we change the comment maybe?

aptos-move/aptos-aggregator/src/aggregator_extension.rs
```diff
@@ -66,236 +57,344 @@ impl AggregatorID {
...
 
-        // Record side-effects of addition in history.
-        self.record();
+    /// Implements logic for doing a "cheap read" of an aggregator.
+    /// Reads the last committed value of the aggregator that's known at the
+    /// time of the call, and as such, can be computed efficiently (i.e. no
+    /// need to consider any speculative state, deltas, etc)
+    /// This method has a sideffect, of updating `speculative_start_value` with
+    /// `LastCommittedValue` variant.
+    /// `get_any_value()` is guaranteed to succeed after this call.
```
gelash, line 311: Should we add the invariant also, i.e. aggregator can't be a non-trivial delta state (pre-condition to initialize Delta state basically).

aptos-move/aptos-aggregator/src/bounded_math.rs
```diff
@@ -0,0 +1,252 @@
...
+pub enum SignedU128 {
+    Positive(u128),
+    Negative(u128),
+}
+
+impl PartialEq for SignedU128 {
+    fn eq(&self, other: &Self) -> bool {
+        match (self, other) {
+            (Self::Positive(v1), Self::Positive(v2)) => v1 == v2,
+            (Self::Negative(v1), Self::Negative(v2)) => v1 == v2,
```
gelash, line 162: nit: we can use | to combine the pairs of arms that have the same handler?

aptos-move/aptos-vm-types/src/change_set.rs
```diff
@@ -293,16 +352,16 @@ impl VMChangeSet {
         }
 
         // Next, squash write ops.
-        for (key, additional_write_op) in additional_aggregator_write_set {
-            match aggregator_write_set.entry(key) {
+        for (key, additional_write_op) in additional_aggregator_v1_write_set {
```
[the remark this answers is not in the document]
gelash: I think we should @igor-aptos @vusirikala

gelash (approved):
🚢

aptos-move/aptos-aggregator/src/aggregator_extension.rs
igor-aptos: It's not the branch prediction that's important, it is that all consequences leaking to the caller need be in the change set as appropriate restrictions. will add some comments

aptos-move/aptos-vm-types/src/change_set.rs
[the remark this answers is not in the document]
igor-aptos: added unimplemented!() to not forget

aptos-move/framework/src/natives/aggregator_natives/context.rs
```diff
@@ -66,93 +63,64 @@ impl<'a> NativeAggregatorContext<'a> {
...
 
     fn get_test_resolver() -> AggregatorStore {
         let mut state_view = AggregatorStore::default();
-        state_view.set_from_id(aggregator_id_for_test(500), 150);
-        state_view.set_from_id(aggregator_id_for_test(600), 100);
-        state_view.set_from_id(aggregator_id_for_test(700), 200);
+        state_view.set_from_id(aggregator_v1_id_for_test(500), 150);
+        state_view.set_from_id(aggregator_v1_id_for_test(600), 100);
+        state_view.set_from_id(aggregator_v1_id_for_test(700), 200);
+        state_view.set_from_id(AggregatorID::ephemeral(900), 300);
```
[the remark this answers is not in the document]
igor-aptos: not sure if these are correctly added, I don't see create/get/remove for them

PR #9370, opened 2023-07-28
[aggregator_v2] add utility add()/sub() methods

igor-aptos:
and change narrower deferred read to also throw an exception, instead of returning an option.

In general, it allows simpler usage when we know it cannot overflow/underflow - https://github.com/igor-aptos/aptos-core/pull/5

### DIFF
--- a/aptos-move/framework/aptos-framework/sources/aggregator_v2/aggregator_v2.move
+++ b/aptos-move/framework/aptos-framework/sources/aggregator_v2/aggregator_v2.move
@@ -1,6 +1,17 @@
 /// This module provides an interface for aggregators (version 2).
 module aptos_framework::aggregator_v2 {
-    use std::option::{Self, Option};
+    use std::error;
+
+    /// The value of aggregator overflows. Raised by uncoditional add() call
+    const EAGGREGATOR_OVERFLOW: u64 = 1;
+
+    /// The value of aggregator underflows (goes below zero). Raised by uncoditional sub call
+    const EAGGREGATOR_UNDERFLOW: u64 = 2;
+
+    /// Tried casting into a narrower type (i.e. u64), but aggregator range of valid values
+    /// cannot fit (i.e. limit exceeds type::MAX).
+    /// Raised by native code (i.e. inside snapshot_with_u64_limit())
+    const EAGGREGATOR_LIMIT_ABOVE_CAST_MAX: u64 = 2;
 
     /// Represents an integer which supports parallel additions and subtractions
     /// across multiple transactions. See the module description for more details.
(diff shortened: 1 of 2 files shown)


aptos-move/framework/aptos-framework/sources/aggregator_v2/aggregator_v2.move
```diff
@@ -21,22 +31,34 @@ module aptos_framework::aggregator_v2 {
...
+        assert!(try_add(aggregator, value), error::out_of_range(EAGGREGATOR_OVERFLOW));
+    }
+
     /// Subtracts `value` from aggregator.
-    /// Returns `true` if the subtraction succeeded and `false` if it tried going below 0.
+    /// If subtraction would result in a negative value, `false` is returned, and aggregator value is left unchanged.
     public native fun try_sub(aggregator: &mut Aggregator, value: u128): bool;
 
+    // Adds `value` to aggregator, uncoditionally.
+    // If addition would exceed the limit, EAGGREGATOR_OVERFLOW exception will be thrown
```
georgemitenkov, line 48: should be EAGGREGATOR_UNDERFLOW

georgemitenkov (approved):

aptos-move/framework/aptos-framework/sources/aggregator_v2/aggregator_v2.move
```diff
@@ -1,6 +1,16 @@
...
+
+    /// The value of aggregator overflows. Raised by uncoditional add() call
+    const EAGGREGATOR_OVERFLOW: u64 = 1;
+
+    /// The value of aggregator underflows (goes below zero). Raised by uncoditional sub call
+    const EAGGREGATOR_UNDERFLOW: u64 = 2;
+
+    /// Tried casting into a narrower type (i.e. u64), but aggregator range of valid values
+    /// cannot fit (i.e. limit exceeds type::MAX)
+    const EAGGREGATOR_LIMIT_ABOVE_CAST_MAX: u64 = 2;
```
georgemitenkov, line 13: Maybe add a comment that this is raised by native code?

aptos-move/framework/aptos-framework/sources/aggregator_v2/aggregator_v2.move
```diff
@@ -21,22 +31,34 @@ module aptos_framework::aggregator_v2 {
...
 
     public native fun deferred_read(aggregator: &Aggregator): AggregatorSnapshot<u128>;
 
     // Do automatic conversion to u64, if all possible values of aggregator fit it (i.e. limit is <= u64::MAX)
-    // If limit of the aggregator exceeds u64::MAX, this will return None)
-    // This doesn't check if actual value can be converted.
-    public native fun deferred_read_convert_u64(aggregator: &Aggregator): Option<AggregatorSnapshot<u64>>;
+    // If limit of the aggregator exceeds u64::MAX, this will throw an exception.
+    // This doesn't check if actual value can be converted, only if it can always be converted, based on the limit.
+    public native fun deferred_read_u64(aggregator: &Aggregator): AggregatorSnapshot<u64>;
```
georgemitenkov, line 61: I am wondering if aggregator is always an int, should the casting be done by having some map/cast API on AggregatorSnapshot directly? So we first deferred read and then map/cast

igor-aptos: Aggregator snapshot doesn't have a limit any more? And casting of snapshot would suggest it's based on value, instead of the limit

aptos-move/framework/aptos-framework/sources/aggregator_v2/aggregator_v2.move
movekevin: I'm curious - what do people see as the external error if this happens? Do they get this nice error or something more opaque?

aptos-move/framework/aptos-framework/sources/aggregator_v2/aggregator_v2.move
```diff
@@ -21,22 +31,34 @@ module aptos_framework::aggregator_v2 {
...
+
     /// Returns a value stored in this aggregator.
     public native fun read(aggregator: &Aggregator): u128;
 
     public native fun deferred_read(aggregator: &Aggregator): AggregatorSnapshot<u128>;
 
     // Do automatic conversion to u64, if all possible values of aggregator fit it (i.e. limit is <= u64::MAX)
-    // If limit of the aggregator exceeds u64::MAX, this will return None)
-    // This doesn't check if actual value can be converted.
-    public native fun deferred_read_convert_u64(aggregator: &Aggregator): Option<AggregatorSnapshot<u64>>;
```
movekevin, line 39: Renaming a public function will break compatibility. May be add this as a separate function?

movekevin (commented):
What are the diffs between these and try_add/try_sub beyond safety?

aptos-move/framework/aptos-framework/sources/aggregator_v2/aggregator_v2.move
igor-aptos: This is all on aggregators_v2 branch,  so we can make breaking changes while working on it

igor-aptos:
Difference is only in add throwing exception,  while try_add returning boolean.  So just for convenience,  and why these are implemented as simple wrappers in Move. 

For what will be shown to users,  not sure,  will check. Is there somewhere a message can be added @movekevin ?

aptos-move/framework/aptos-framework/sources/aggregator_v2/aggregator_v2.move
movekevin: I see

gelash (approved):
The fact that casting is (and must be) based on the limit is an important point. Wonder how we can emphasize more clearly for the user perspective that it isn't a dynamic cast based on value. I like that the error is very clear about that. Couldn't find a good suggestion for method call yet.
snapshot_u64_limit?.. idk

igor-aptos:
snapshot_with_u64_limit ? 

for exceptions, since backing code is not there yet, will test later, once that is implemented

Issue #9164, opened 2023-07-19
[Feature Request] more control over the conflict pattern when running executor-benchmark


zjma:
# 🚀 Feature Request

Currently we can run `executor-benchmark` with transactions uniformly sampled from a finite account pool using the commands as follow, where in-block conflict level can be roughly controlled by `num_accounts / block_size`.
```
cargo run --p aptos-executor-benchmark -- create-db --data-dir /tmp/some-db --num-accounts 1000
cargo run --p aptos-executor-benchmark --block-size 100-- run-executor --main-signer-accounts 1000 --data-dir /tmp/some-db --checkpoint-dir /tmp/some-checkpoint --blocks 99
```

We can also run `executor-benchmark` with guaranteed non-conflicting transactions using the commands as follow, where the account pool size needs to be 3 times as large as the block size for the [BurnAndRecycleSampler](https://github.com/aptos-labs/aptos-core/blob/0b6859f036719bd33932ba262f7692e5fc18c1ae/crates/transaction-generator-lib/src/p2p_transaction_generator.rs#L76) to work.
```
cargo run --p aptos-executor-benchmark -- create-db --data-dir /tmp/some-db --num-accounts 300
cargo run -package aptos-executor-benchmark -- --transactions-per-sender 1 --block-size 100 run-executor --transaction-type non-conflicting-coin-transfer --main-signer-accounts 300 --data-dir /tmp/some-db --checkpoint-dir /tmp/some-ck --blocks 99
```

But they are not good enough: can we define and have finer control over the conflict level (@sitalkedia to add more details)? can we also have a way to sample non-uniformly from the account pool, just for simulating the "hot-spot" scenario (e.g., 20% of the accounts are used 80% of the time)

## Motivation

Sharded block execution can be sensitive to the conflict pattern. Being able to do what's described above is necessary for benchmarking any sharded executor implementation.

PR #9371, opened 2023-07-28
[Executor-benchmark] Command to generate connected groups of random txns

manudhundi:
Command example:
cargo run -p aptos-executor-benchmark -- --block-size 50 --independent-tx-grps-in-a-block 10 run-executor --main-signer-accounts 1000 --data-dir /tmp/some-db --checkpoint-dir /tmp/some-checkpoint --blocks 2

Note: For a good random workload we would want 'num_accounts_per_grp' to be much greater than 'num_txns_per_grp'.
      The command enforces 'num_accounts_per_grp' > 'num_txns_per_grp'.

### DIFF
--- a/execution/executor-benchmark/src/main.rs
+++ b/execution/executor-benchmark/src/main.rs
@@ -124,10 +124,10 @@ struct Opt {
     transactions_per_sender: usize,
 
     /// 0 implies random TX generation; if non-zero, then 'transactions_per_sender is ignored
-    /// 'independent_tx_grps_in_a_block' should be less than 'block_size'
-    /// 'independent_tx_grps_in_a_block' should be less than 'num_main_signer_accounts / 2'
+    /// 'connected_tx_grps' should be less than 'block_size'
+    /// 'connected_tx_grps' should be less than 'num_main_signer_accounts / 2'
     #[clap(long, default_value_t = 0)]
-    independent_tx_grps_in_a_block: usize,
+    connected_tx_grps: usize,
 
     #[clap(long)]
     concurrency_level: Option<usize>,
@@ -287,7 +287,7 @@ where
                 blocks,
                 transaction_mix,
                 opt.transactions_per_sender,
-                opt.independent_tx_grps_in_a_block,
+                opt.connected_tx_grps,
                 main_signer_accounts,
                 additional_dst_pool_accounts,
                 data_dir,
--- a/execution/executor-benchmark/src/main.rs
+++ b/execution/executor-benchmark/src/main.rs
@@ -125,7 +125,6 @@ struct Opt {
 
     /// 0 implies random TX generation; if non-zero, then 'transactions_per_sender is ignored
     /// 'connected_tx_grps' should be less than 'block_size'
-    /// 'connected_tx_grps' should be less than 'num_main_signer_accounts / 2'
     #[clap(long, default_value_t = 0)]
     connected_tx_grps: usize,
 
--- a/execution/executor-benchmark/src/main.rs
+++ b/execution/executor-benchmark/src/main.rs
@@ -123,6 +123,12 @@ struct Opt {
     #[clap(long, default_value_t = 5)]
     transactions_per_sender: usize,
 
+    /// 0 implies random TX generation; if non-zero, then 'transactions_per_sender is ignored
+    /// 'independent_tx_grps_in_a_block' should be less than 'block_size'
+    /// 'independent_tx_grps_in_a_block' should be less than 'num_main_signer_accounts / 2'
+    #[clap(long, default_value_t = 0)]
+    independent_tx_grps_in_a_block: usize,
+
     #[clap(long)]
     concurrency_level: Option<usize>,
 
@@ -281,6 +287,7 @@ where
                 blocks,
                 transaction_mix,
                 opt.transactions_per_sender,
+                opt.independent_tx_grps_in_a_block,
                 main_signer_accounts,
                 additional_dst_pool_accounts,
                 data_dir,
--- a/execution/executor-benchmark/src/transaction_generator.rs
+++ b/execution/executor-benchmark/src/transaction_generator.rs
@@ -269,9 +269,15 @@ impl TransactionGenerator {
         block_size: usize,
         num_transfer_blocks: usize,
         transactions_per_sender: usize,
+        independent_tx_grps_in_a_block: usize,
     ) {
         assert!(self.block_sender.is_some());
-        self.gen_transfer_transactions(block_size, num_transfer_blocks, transactions_per_sender);
+        self.gen_transfer_transactions(
+            block_size,
+            num_transfer_blocks,
+            transactions_per_sender,
+            independent_tx_grps_in_a_block,
+        );
     }
 
     pub fn run_workload(
@@ -405,7 +411,7 @@ impl TransactionGenerator {
     }
 
     /// Generates transactions for random pairs of accounts.
-    pub fn gen_transfer_transactions(
+    pub fn gen_random_transfer_transactions(
         &mut self,
         block_size: usize,
         num_blocks: usize,
@@ -438,6 +444,63 @@ impl TransactionGenerator {
             if let Some(sender) = &self.block_sender {
                 sender.send(transactions).unwrap();
             }
+        }
+    }
+
+    /// Generates connected groups of transactions for random pairs of accounts.
+    pub fn gen_independent_grps_of_connected_random_transfer_transactions(
+        &mut self,
+        block_size: usize,
+        num_blocks: usize,
+        independent_tx_grps_in_a_block: usize,
+    ) {
+        let num_accounts_per_grp = self.main_signer_accounts.as_ref().unwrap().accounts.len()
+            / independent_tx_grps_in_a_block;
+        // TODO: handle when block_size isn't divisible by independent_tx_grps_in_a_block; an easy
+        //       way to do this is to just generate a few more transactions in the last group
+        let num_txns_per_grp = block_size / independent_tx_grps_in_a_block;
+
+        if num_txns_per_grp >= num_accounts_per_grp {
+            panic!("For the desired workload we want num_accounts_per_grp ({}) > num_txns_per_grp ({})", num_accounts_per_grp, num_txns_per_grp);
+        }
+        for _ in 0..num_blocks {
+            let mut transactions: Vec<_> = (0..independent_tx_grps_in_a_block)
+                .flat_map(|grp_idx| {
+                    self.main_signer_accounts
+                        .as_mut()
+                        .unwrap()
+                        .get_connected_random_transfers(
+                            &self.transaction_factory,
+                            num_txns_per_grp,
+                            grp_idx * num_accounts_per_grp,
+                            (grp_idx + 1) * num_accounts_per_grp - 1,
+                        )
+                })
+                .collect();
+            transactions.push(Transaction::StateCheckpoint(HashValue::random()));
+            self.version += transactions.len() as Version;
+
+            if let Some(sender) = &self.block_sender {
+                sender.send(transactions).unwrap();
+            }
+        }
+    }
+
+    pub fn gen_transfer_transactions(
+        &mut self,
+        block_size: usize,
+        num_blocks: usize,
+        transactions_per_sender: usize,
+        independent_tx_grps_in_a_block: usize,
+    ) {
+        if independent_tx_grps_in_a_block > 0 {
+            self.gen_independent_grps_of_connected_random_transfer_transactions(
+                block_size,
+                num_blocks,
+                independent_tx_grps_in_a_block,
+            );
+        } else {
+            self.gen_random_transfer_transactions(block_size, num_blocks, transactions_per_sender);
         }
     }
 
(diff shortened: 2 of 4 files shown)


manudhundi:
#9164

execution/executor-benchmark/src/transaction_generator.rs
```diff
@@ -405,7 +411,7 @@ impl TransactionGenerator {
     }
 
     /// Generates transactions for random pairs of accounts.
-    pub fn gen_transfer_transactions(
```
sitalkedia, line 408: @manudhundi - Sorry about the confusion but This is old way of generating txns, which should be deprecated. Can you add this functionality to the  transaction generator in `run_workload` function ?

execution/executor-benchmark/src/transaction_generator.rs
```diff
@@ -441,6 +447,63 @@ impl TransactionGenerator {
         }
     }
 
+    /// Generates connected groups of transactions for random pairs of accounts.
+    pub fn gen_independent_grps_of_connected_random_transfer_transactions(
```
sitalkedia, line 451: Function name too long

execution/executor-benchmark/src/transaction_generator.rs
```diff
@@ -441,6 +447,63 @@ impl TransactionGenerator {
         }
     }
 
+    /// Generates connected groups of transactions for random pairs of accounts.
```
sitalkedia, line 450: This could use a comment explaining what a connected group of txn is.

execution/executor-benchmark/src/transaction_generator.rs
sitalkedia: May be we can add a new sampling mode here https://github.com/aptos-labs/aptos-core/blob/4cc1c40a1b5b27c71b79a976458b27f3a7629541/crates/transaction-generator-lib/src/p2p_transaction_generator.rs#L25. @zjma thoughts?

execution/executor-benchmark/src/transaction_generator.rs
```diff
@@ -441,6 +447,63 @@ impl TransactionGenerator {
...
+    pub fn gen_independent_grps_of_connected_random_transfer_transactions(
+        &mut self,
+        block_size: usize,
+        num_blocks: usize,
+        independent_tx_grps_in_a_block: usize,
```
sitalkedia, line 455: just rename to `independent_txn_grps`

execution/executor-benchmark/src/transaction_generator.rs
zjma: IMO we should stop messing with `TransactionGenerator`s for controlling in-block conflict, because [its API](https://github.com/aptos-labs/aptos-core/blob/991c7ed1a8375f8f799583b628d5a8974656ef35/crates/transaction-generator-lib/src/lib.rs#L85-L91) is not designed for this: e.g. non-conflicting block generating was trivial, but we made it much more complicated by `BurnAndRecycle` sampling inside TransactionGenerator, and outside also ensuring unique senders as well as correct pool size... (remember why pool size has to be 3x+ of block size?)
 
We simply need a trait `BlockGenerator` with API `fn generator_block(size: usize) -> Vec<Transaction>`.
And we can have instances
```
block_generators::P2PDefault {
  accounts: Vec<Account>,
}

block_generators::P2PAllConnected { // connected transfer graph
  accounts: Vec<Account>,
}

///  Can leverage P2PAllConnected;
/// get non-conflicting block for free when num_non_conflicting_groups=block_size
block_generators::P2PNonConflictGroup {
  accounts: Vec<Account>,
  num_non_conflicting_groups: usize,
}

block_generators::P2PHotspot {
  accounts: Vec<Account>,
  hotspot: f32, //if 0.8, then 80% accounts will be picked with probability 20%, the other 20% get 80%
}

// other patterns/shapes
```

execution/executor-benchmark/src/transaction_generator.rs
manudhundi: renamed it as 'connected_tx_grps'.. Using 'connected_..' instead of 'independent..' because independent_grp does not clearly indicate whether txns within the same group are independent or independent across the groups

execution/executor-benchmark/src/transaction_generator.rs
manudhundi: As discussed offline, adding the functionality in `TransactionGenerator` used by `run_workload` requires non-trivial amount of refactoring. So adding it here (in `run_transfer`) for now. We will get a refactoring in the future perhaps

@zjma trait `BlockGenerator` is a good idea, considered this as part of the above [refactoring](https://github.com/aptos-labs/aptos-core/commit/719953622c9d0236ed427cb26edc7b892c88318c). But, to `run_workload` with `BlockGenerator` we will have to refactor `TransactionGenerator` and also add `BlockGenerator` for non P2P cases.

execution/executor-benchmark/src/main.rs
```diff
@@ -120,6 +120,12 @@ struct Opt {
     #[clap(long, default_value_t = 5)]
     transactions_per_sender: usize,
 
+    /// 0 implies random TX generation; if non-zero, then 'transactions_per_sender is ignored
+    /// 'connected_tx_grps' should be less than 'block_size'
```
sitalkedia, line 124: Can you enforce these checks somewhere?

execution/executor-benchmark/src/transaction_generator.rs
```diff
@@ -441,6 +451,118 @@ impl TransactionGenerator {
...
+        //       way to do this is to just generate a few more transactions in the last group
+        let num_txns_per_grp = block_size / connected_tx_grps;
+
+        if num_txns_per_grp >= num_accounts_per_grp {
+            panic!("For the desired workload we want num_accounts_per_grp ({}) > num_txns_per_grp ({})", num_accounts_per_grp, num_txns_per_grp);
+        }
+
+        (0..connected_tx_grps)
+            .flat_map(|grp_idx| {
+                let accounts_st_idx = grp_idx * num_accounts_per_grp;
```
sitalkedia, line 471: nit: `accounts_st_idx` -> `accounts_start_idx`

execution/executor-benchmark/src/transaction_generator.rs
```diff
@@ -441,6 +451,118 @@ impl TransactionGenerator {
...
+                    vec![unused_indices.pop().unwrap(), unused_indices.pop().unwrap()];
+                let mut transfer_indices: Vec<(_, _)> = vec![(used_indices[0], used_indices[1])];
+
+                for _ in 1..num_txns_per_grp {
+                    // index1 is always from used_indices, so that all the txns are connected
+                    let mut index1 = used_indices[rng.gen_range(0, used_indices.len())];
+
+                    // index2 is either from used_indices or unused_indices
+                    let mut index2;
+                    let rnd = rng.gen_range(0, used_indices.len() + unused_indices.len());
```
sitalkedia, line 485: I am not sure if this will result in a desired random distribution. As the used indices increases, the probability of choosing the used index will grow larger. May be we should just flip a coin and decide if to pick from used vs unused?

execution/executor-benchmark/src/transaction_generator.rs
```diff
@@ -441,6 +451,118 @@ impl TransactionGenerator {
...
+                    if rnd < used_indices.len() {
+                        index2 = used_indices[rnd];
+                    } else {
+                        // unused_indices is shuffled already, so last element is random
+                        index2 = unused_indices.pop().unwrap();
+                        used_indices.push(index2);
+                    }
+
+                    if rng.gen_range(0, 2) == 0 {
```
sitalkedia, line 494: You can use rng.gen() to get a boolean instead.

execution/executor-benchmark/src/transaction_generator.rs
```diff
@@ -441,6 +451,118 @@ impl TransactionGenerator {
...
+                let accounts_end_idx = accounts_st_idx + num_accounts_per_grp - 1;
+                let mut unused_indices: Vec<_> = (accounts_st_idx..=accounts_end_idx).collect();
+                unused_indices.shuffle(rng);
+                let mut used_indices: Vec<_> =
+                    vec![unused_indices.pop().unwrap(), unused_indices.pop().unwrap()];
+                let mut transfer_indices: Vec<(_, _)> = vec![(used_indices[0], used_indices[1])];
+
+                for _ in 1..num_txns_per_grp {
+                    // index1 is always from used_indices, so that all the txns are connected
+                    let mut index1 = used_indices[rng.gen_range(0, used_indices.len())];
```
sitalkedia, line 481: Can you add a comment in the function header clearly explaining the logic for connected component generation?

execution/executor-benchmark/src/transaction_generator.rs
```diff
@@ -483,3 +605,64 @@ impl TransactionGenerator {
         self.block_sender.take().unwrap();
     }
 }
+
+#[test]
+fn test_get_connected_grps_transfer_indices() {
```
sitalkedia, line 610: Great work adding some tests for this.

execution/executor-benchmark/src/transaction_generator.rs
manudhundi: done

execution/executor-benchmark/src/transaction_generator.rs
manudhundi: I see your point. By flipping a coin we decide whether we have one side or both sides connected with equal probabilities. Will add the logic

execution/executor-benchmark/src/main.rs
manudhundi: Done.

execution/executor-benchmark/src/transaction_generator.rs
```diff
@@ -441,6 +451,124 @@ impl TransactionGenerator {
...
+        } else if connected_tx_grps > block_size {
+            panic!(
+                "connected_tx_grps ({}) > block_size ({}) cannot guarantee at least 1 txn per grp",
+                connected_tx_grps, block_size
+            );
+        }
+
+        (0..connected_tx_grps)
+            .flat_map(|grp_idx| {
+                let accounts_start_idx = grp_idx * num_accounts_per_grp;
```
zjma, line 478: optional, but it would be even better if the account grouping is also random for every block...

zjma (approved):

execution/executor-benchmark/src/transaction_generator.rs
manudhundi: sorry missed this comment, adding it now

execution/executor-benchmark/src/transaction_generator.rs
manudhundi: technically, I was thinking, what does random mean here ? I think even the `unused_indices.shuffle(rng);`is probably not needed.

If we are given a pool of addresses, they are random by default (since the address values don't matter for individual txns I suppose). Only after we touch an address (use it as a sender or receiver), the usage of the rest may have to be randomized.

execution/executor-benchmark/src/transaction_generator.rs
manudhundi: done

execution/executor-benchmark/src/transaction_generator.rs
zjma: ideally no recognizable pattern other than the specified constraints.

I think the current implementation is almost pattern-free. One example of the pattern (under some params):
- if accounts[0] appears, it always comes very early in the block.
- if accounts[n-1] appears, it always comes very late in the block.
- a transfer of accounts[0] -> accounts[n-1] is impossible.

manudhundi: But aren't accounts[0] and accounts[n-1] themselves random to begin with ?
From execution benchmark perspective, is there a difference if transfer from accounts[0] -> accounts[n-1] not possible vs transfer from accounts[0] -> accounts[1] not possible ? Values of accounts[0], accounts[1], accounts[n-1] are all random. What might matter is how there are connected wrt txns, or how many times an account sends/receives.

zjma: Right, For implementing connected components the pattern doesn't matter. But we still want to avoid it if possible - otherwise we have the small risk of slowly overfitting our partitioner to this pattern.

zjma: I think a way to avoid this pattern is to shuffle the account list at the beginning of each block generation. Can be implemented as generating a random permutation p=[5,2,0,1,3,4...] of all account indices, and replace accounts[I] with accounts[p[I]]

zjma: Still, this is optional and we can do it later.

sitalkedia: @zjma - Actually this is a very good point and I think we need to resolve this.  So in the current implementation, let's say we have a block size of 10 and number of connected component is 5 - this pattern ensures that all conflicting txns are close to each other. For example, txn1 won't ever conflict with txn10, which will lead us to optimize partitioning for this pattern specifically. @manudhundi - To avoid this problem, how about shuffling the txn indices at the end before returning from this function?

manudhundi: @sitalkedia you are saying we shuffle at the end of the block where conflicting txns are not close to each other.
@zjma you are saying we shuffle account addresses at the beginning to avoid risks of overfitting the partitioner.

Will do both

sitalkedia (approved):
LGTM

PR #9383, opened 2023-07-31
[Spec] Ensures of reconfiguration.move

UIZorrot:
### Description

This is the extra ensures the condition of reconfiguration.move module.
The High-level properties of this module are 100% complete.
(property 6 is already solved by the global invariant)

### Test Plan
Aptos move prove -f reconfig


aptos-move/framework/aptos-framework/sources/reconfiguration.spec.move
```diff
@@ -58,33 +72,49 @@ spec aptos_framework::reconfiguration {
...
         pragma verify_duration_estimate = 120; // TODO: set because of timeout (property proved)
 
         requires exists<stake::ValidatorFees>(@aptos_framework);
+        requires exists<CoinInfo<AptosCoin>>(@aptos_framework);
 
         include transaction_fee::RequiresCollectedFeesPerValueLeqBlockAptosSupply;
-        include features::spec_periodical_reward_rate_decrease_enabled() ==> staking_config::StakingRewardsConfigEnabledRequirement;
-        include features::spec_collect_and_distribute_gas_fees_enabled() ==> aptos_coin::ExistsAptosCoin;
-
+        include staking_config::StakingRewardsConfigRequirement;
```
rahxephon89, line 95: why do you change this?

aptos-move/framework/aptos-framework/sources/reconfiguration.spec.move
```diff
@@ -58,33 +72,49 @@ spec aptos_framework::reconfiguration {
...
         aborts_if false;
+
+        // The ensure conditions of the reconfigure function are not fully written, because there is a new cycle in it,
+        // but its existing ensure conditions satisfy hp.
         let success = !(chain_status::is_genesis() || timestamp::spec_now_microseconds() == 0 || !reconfiguration_enabled())
             && timestamp::spec_now_microseconds() != global<Configuration>(@aptos_framework).last_reconfiguration_time;
+        // property 3: Synchronization of NewEpochEvent counter with configuration epoch.
         ensures success ==> global<Configuration>(@aptos_framework).epoch == old(global<Configuration>(@aptos_framework).epoch) + 1;
+        ensures (success && event::counter<NewEpochEvent>(old(global<Configuration>(@aptos_framework)).events) <
```
rahxephon89, line 104: Please remove the property on event, this is proved because of inconsistency

rahxephon89 (changes requested):

aptos-move/framework/aptos-framework/sources/reconfiguration.spec.move
UIZorrot: It should be a error in conflict solve, it has been added back

rahxephon89: It has not been added back from what I see here

aptos-move/framework/aptos-framework/sources/reconfiguration.spec.move
rahxephon89: Could you double check whether it is actually removed?

aptos-move/framework/aptos-framework/sources/reconfiguration.spec.move
rahxephon89: `include staking_config::StakingRewardsConfigRequirement;` is not necessary. Please remove it

aptos-move/framework/aptos-framework/sources/reconfiguration.spec.move
UIZorrot: I removed the         ensures (success && event::counter<NewEpochEvent>(old(global<Configuration>(@aptos_framework)).events) <
            MAX_U64) ==>
            event::counter<NewEpochEvent>(global<Configuration>(@aptos_framework).events) ==
                event::counter<NewEpochEvent>(old(global<Configuration>(@aptos_framework)).events) + 1;

UIZorrot: Do you means this item should be keep as :

aptos-move/framework/aptos-framework/sources/reconfiguration.spec.move
UIZorrot: removed

aptos-move/framework/aptos-framework/sources/reconfiguration.spec.move
rahxephon89: No, current status looks good to me

rahxephon89 (approved):

PR #9385, opened 2023-07-31
[GHA] Skip expensive tests on docs only changes.

JoshLind:
### Description
This PR updates our GHA to avoid running the expensive code related tests on changes that are docs only. Essentially the PR:
- Adds a new file change determinator job using this [action](https://github.com/marketplace/actions/skip-duplicate-actions). Right now, I'm just checking for docs changes (e.g., `*.md` or only changes to `/developer-docs-site`).
- If the job identifies that only docs have changed:
  - We skip running expensive and not required tests (e.g., performance tests, consensus tests, etc.)
  - We still run expensive and required tests, but they essentially become a no-op (e.g., smoke tests, rust unit tests, linters, etc.). This way, all required jobs still pass but they're empty.

Note: in the PR, I've moved somethings into new files (e.g., `rust_lint`, `rust_unit_tests` , `rust_smoke_tests`, etc.). This is mainly so that it's easy to skip these actions as units (instead of having ugly if-else branches in our workflows).

Once this PR lands, I'll follow up by: (i) removing the old "required tests"; (ii) updating the "required test" list to include these; and (iii) extending the checks to avoid skipping the other expensive tests.

### Test Plan
Existing test infrastructure and manual runs.

### DIFF
--- a/.github/actions/file-change-determinator/action.yaml
+++ b/.github/actions/file-change-determinator/action.yaml
@@ -1,0 +1,16 @@
+name: File Change Determinator
+description: Runs the file change determinator (to identify which files changed in a pull request)
+outputs:
+  only_docs_changed:
+    description: "Returns true if only docs files were changed in the pull request"
+    value: ${{ steps.doc_change_determinator.outputs.should_skip }}
+
+runs:
+  using: composite
+  steps:
+    # Run the docs only change determinator
+    - id: doc_change_determinator
+      continue-on-error: true # Avoid skipping any checks if this job fails (see: https://github.com/fkirc/skip-duplicate-actions/issues/301)
+      uses: fkirc/skip-duplicate-actions@v5
+      with:
+        paths_ignore: '["**/*.md", "developer-docs-site/**"]'
--- a/.github/actions/general-lints/action.yaml
+++ b/.github/actions/general-lints/action.yaml
@@ -1,0 +1,56 @@
+name: General Lints
+description: Runs all general lints. This includes all linters except rust and docs lints.
+inputs:
+  GIT_CREDENTIALS:
+    description: "Optional credentials to pass to git. Useful if you need to pull private repos for dependencies"
+    required: false
+
+runs:
+  using: composite
+  steps:
+    # Checkout the repository
+    - uses: actions/checkout@v3
+
+    # Install shellcheck and run it on the dev_setup.sh script
+    - name: Run shell lints
+      run: |
+        sudo apt-get install shellcheck --assume-yes --no-install-recommends
+        shellcheck scripts/dev_setup.sh
+      shell: bash
+
+    # Run the python setup and run the python SDK linters
+    - name: Run python setup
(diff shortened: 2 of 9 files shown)


.github/actions/file-change-determinator/action.yaml
```diff
@@ -0,0 +1,17 @@
...
+runs:
+  using: composite
+  steps:
+    # Run the docs only change determinator
+    - id: doc_change_determinator
+      continue-on-error: true # Avoid skipping any checks if this job fails (see: https://github.com/fkirc/skip-duplicate-actions/issues/301)
+      uses: fkirc/skip-duplicate-actions@v5
+      with:
+        #paths_ignore: '["**/*.md", "developer-docs-site/**"]'
+        paths_ignore: '["**/*.md", "developer-docs-site/**", "**/*.yaml"]' # TODO: Remove yaml support. This is for testing!
```
JoshLind, line 17: I'll remove this before landing 😄

.github/actions/general-lints/action.yaml
```diff
@@ -0,0 +1,56 @@
+name: General Lints
```
JoshLind, line 1: Note: these new files are all just copy/pasted snippets from the existing "required" workflows. The idea is to move the existing "required" workflow steps into these files so we can skip the entire file/flow if we need to 😄

.github/workflows/execution-performance.yaml
```diff
@@ -8,15 +8,15 @@ on:
 
 jobs:
   execution-performance:
-    uses: aptos-labs/aptos-core/.github/workflows/workflow-run-execution-performance.yaml@main
+    uses: ./.github/workflows/workflow-run-execution-performance.yaml #aptos-labs/aptos-core/.github/workflows/workflow-run-execution-performance.yaml@main
```
JoshLind, line 11: I'll also roll these types of changes back on landing. They're just to make sure I'm testing the actual actions (that aren't on the main branch yet 😄)

.github/workflows/execution-performance.yaml
```diff
@@ -8,15 +8,15 @@ on:
...
 
   spot-runner-execution-performance:
-    uses: aptos-labs/aptos-core/.github/workflows/workflow-run-execution-performance.yaml@main
+    uses: ./.github/workflows/workflow-run-execution-performance.yaml  #aptos-labs/aptos-core/.github/workflows/workflow-run-execution-performance.yaml@main
     secrets: inherit
     with:
       GIT_SHA: ${{ github.event.pull_request.head.sha || github.sha }}
-      RUNNER_NAME: spot-runner
\ No newline at end of file
+      RUNNER_NAME: spot-runner
```
ibalajiarun: need to rebase I believe.

JoshLind: Aah, yes, thanks @ibalajiarun 😄 Let me rebase!

gedigi (approved):

sionescu (approved):

PR #9388, opened 2023-07-31
Gerben apply to v1.5

gerben-stavenga:
### Description

Apply patch to v1.5 public repo for historical consistency


types/src/on_chain_config/timed_features.rs
```diff
@@ -58,6 +60,8 @@ impl TimedFeatureFlag {
         match (self, chain_id) {
             (DisableInvariantViolationCheckInSwapLoc, TESTNET) => NOT_YET_SPECIFIED,
             (DisableInvariantViolationCheckInSwapLoc, MAINNET) => NOT_YET_SPECIFIED,
+            (LimitTypeTagSize, TESTNET) => 1689966000000,
```
perryjrandall, line 63: I think you need to remove these

gerben-stavenga: This is historic right? So this is to the best of the our knowledge what was final release of v1.5

perryjrandall (approved):
accepting to unblock

types/src/on_chain_config/timed_features.rs
perryjrandall: In that case maybe we can include a human readable timestamp

davidiw (approved):

PR #9394, opened 2023-07-31
[State Sync] Ensure fast sync doesn't exit early.

JoshLind:
WIP.


JoshLind:
Closing this out as it should hopefully not be needed.

PR #9395, opened 2023-07-31
[TS SDK] Improve `IndexerClient` to support tokenv2 sorting and new queries

0xmaayan:
### Description
Per @kent-white's request in https://github.com/aptos-labs/explorer/pull/578

- Export indexer types `export * from "./indexer/generated/types";`
- Support for token v2 activities
- Aggregate query for token v2 activities
- Support for sorting indexer queries
- Support for get owned tokens by token_data_id
### Test Plan
ts sdk indexer tests are passing

### DIFF
--- a/ecosystem/typescript/sdk/src/providers/indexer.ts
+++ b/ecosystem/typescript/sdk/src/providers/indexer.ts
@@ -20,6 +20,7 @@ import {
   GetCollectionDataQuery,
   GetCollectionsWithOwnedTokensQuery,
   GetTokenCurrentOwnerDataQuery,
+  GetOwnedTokensByTokenDataQuery,
 } from "../indexer/generated/operations";
 import {
   GetAccountTokensCount,
@@ -41,9 +42,18 @@ import {
   GetCollectionData,
   GetCollectionsWithOwnedTokens,
   GetTokenCurrentOwnerData,
+  GetOwnedTokensByTokenData,
 } from "../indexer/generated/queries";
 import { ClientConfig, post } from "../client";
 import { ApiError } from "./aptos_client";
+import {
+  Current_Collections_V2_Order_By,
+  Current_Collection_Ownership_V2_View_Order_By,
+  Current_Token_Datas_V2_Order_By,
+  Current_Token_Ownerships_V2_Order_By,
+  InputMaybe,
+  Token_Activities_V2_Order_By,
+} from "../indexer/generated/types";
 
 /**
  * Controls the number of results that are returned and the starting position of those results.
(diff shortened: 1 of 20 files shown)


kent-white (approved):
noice

ecosystem/typescript/sdk/src/providers/indexer.ts
```diff
@@ -146,31 +168,52 @@ export class IndexerClient {
   }
 
   /**
-   * Queries a token activities by token id hash
+   * Queries a token activities by token address (token data id)
```
xbtmatt, line 171: Maybe to be a little clearer: `* Queries a token activities by token address or token data id`

ecosystem/typescript/sdk/src/providers/indexer.ts
```diff
@@ -451,6 +419,45 @@ export class IndexerClient {
...
         offset: extraArgs?.options?.offset,
         limit: extraArgs?.options?.limit,
+        order_by: extraArgs?.orderBy,
+      },
+    };
+    return this.queryIndexer(graphqlQuery);
+  }
+
+  /**
+   * Queries account's current owned tokens by the token address (token data id).
```
xbtmatt, line 429: same here, (and maybe anywhere else..?) `token address or token data id`

ecosystem/typescript/sdk/src/providers/indexer.ts
0xmaayan: yea... so token data id is token address, indexer uses token_data_id key but token address might be more understandable for users..?

xbtmatt: it's either the token_data_id hash or the token address, the former for v1 and latter for v2, right? that was my understanding

0xmaayan: `token_data_id` is still a thing in v2 https://cloud.hasura.io/public/graphiql?endpoint=https%3A%2F%2Findexer.mainnet.aptoslabs.com%2Fv1%2Fgraphql&query=query+MyQuery+%7B%0A++token_activities_v2%28%0A++++where%3A+%7Btoken_data_id%3A+%7B_eq%3A+%220x5a8e6dd0531df0b68603d0fe6fd57924b34fc3bb7d26c74cfd4955d809ef30cb%22%7D%7D%0A++++distinct_on%3A+transaction_version%0A++%29+%7B%0A++++token_data_id%0A++++token_standard%0A++++from_address%0A++++to_address%0A++%7D%0A%7D%0A

xbtmatt: yeah, but it's just a naming convention to make v1 and v2 queries coalesce, the value for v2 is just the token address, whereas with token v1 the value is the hash of the token name + creator + collection name, and it's a hash only used for the indexer I believe (at least I haven't seen it anywhere else)

Sorry if you already knew this, don't mean to explain it to you if you did, but just saying because some people might not understand that the field for the query works for both v1 and v2

0xmaayan: I see what you mean! yes I was just thinking about sdk vs indexer naming - I think what you are saying makes sense, will update the code! thanks!

xbtmatt: Yeah some of the names are very confusing for the indexer. At first glance you'd think all the `_v2` queries are specifically for token v2, but they're not, they're just more robust versions of the "v1" queries

That's the main reason I suggest it, because then it's a little clearer to people what the purpose of the "_v2" queries are

xbtmatt (approved):

ecosystem/typescript/sdk/src/providers/indexer.ts
0xmaayan: yes, I think the idea was that all `_v2` queries can handle v1 and v2 data, and all v1 queries can only handle v1 data

PR #9396, opened 2023-07-31
[ts-sdk] Fix default for coin client transfer and register on default

gregnazario:
### Description
Change default in the SDK to transfer with register.  Can still be turned off by specifying false

### DIFF
--- a/ecosystem/typescript/sdk/src/plugins/coin_client.ts
+++ b/ecosystem/typescript/sdk/src/plugins/coin_client.ts
@@ -87,7 +87,12 @@ export class CoinClient {
 
     // If we should create the receiver account if it doesn't exist on-chain,
     // use the `0x1::aptos_account::transfer` function.
-    const func = extraArgs?.createReceiverIfMissing ? "0x1::aptos_account::transfer_coins" : "0x1::coin::transfer";
+    let func: string;
+    if (extraArgs?.createReceiverIfMissing === undefined) {
+      func = "0x1::aptos_account::transfer_coins";
+    } else {
+      func = extraArgs?.createReceiverIfMissing ? "0x1::aptos_account::transfer_coins" : "0x1::coin::transfer";
+    }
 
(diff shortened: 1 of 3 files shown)


ecosystem/typescript/sdk/src/plugins/coin_client.ts
```diff
@@ -87,7 +87,12 @@ export class CoinClient {
 
     // If we should create the receiver account if it doesn't exist on-chain,
     // use the `0x1::aptos_account::transfer` function.
-    const func = extraArgs?.createReceiverIfMissing ? "0x1::aptos_account::transfer_coins" : "0x1::coin::transfer";
+    let func: string;
+    if (extraArgs?.createReceiverIfMissing === undefined) {
```
0xmaayan, line 91: do we have tests to verify this logic?

banool (commented):
This is a breaking change, can we mark it as such? Also we should make it possible to opt in to the old behaviour still for those who know they don't need the new function.

gregnazario:
> This is a breaking change, can we mark it as such? Also we should make it possible to opt in to the old behaviour still for those who know they don't need the new function.

You can always provide `false` to use the old default.

0xmaayan:
> > This is a breaking change, can we mark it as such? Also we should make it possible to opt in to the old behaviour still for those who know they don't need the new function.
> 
> You can always provide `false` to use the old default.

we can have it `false` by default

gregnazario:
> > > This is a breaking change, can we mark it as such? Also we should make it possible to opt in to the old behaviour still for those who know they don't need the new function.
> > 
> > 
> > You can always provide `false` to use the old default.
> 
> we can have it `false` by default

The previous default was `false` because of `falsyness`

davidiw (approved):

ecosystem/typescript/sdk/src/plugins/coin_client.ts
gregnazario: Test added

PR #9397, opened 2023-07-31
Re-enable dedup and shuffling as defaults (for test networks)

bchocho:
### Description

Introduce a new `Missing` version for `OnChainExecutionConfig` that will maintain backwards compatibility for before the config was registered. Use this when the config is missing from epoch manager.

Otherwise, use `default_for_genesis` for all new networks e.g., forge, devnet.

### Test Plan

Observe that e2e tests use both shuffle and dedup.

### DIFF
--- a/types/src/on_chain_config/execution_config.rs
+++ b/types/src/on_chain_config/execution_config.rs
@@ -41,6 +41,14 @@ impl OnChainExecutionConfig {
             OnChainExecutionConfig::V3(config) => config.transaction_deduper_type.clone(),
         }
     }
+
+    // TODO: Remove and replace all calls with default once shuffler is enabled in all networks
+    pub fn default_for_test() -> Self {
+        OnChainExecutionConfig::V3(ExecutionConfigV3 {
+            transaction_shuffler_type: TransactionShufflerType::SenderAwareV2(32),
+            ..Default::default()
+        })
+    }
 }
 
 /// This is used when on-chain config is not initialized.
--- a/types/src/on_chain_config/execution_config.rs
+++ b/types/src/on_chain_config/execution_config.rs
@@ -11,6 +11,9 @@ pub enum OnChainExecutionConfig {
     V1(ExecutionConfigV1),
     V2(ExecutionConfigV2),
     V3(ExecutionConfigV3),
+    /// To maintain backwards compatibility on replay, we must ensure that any new features resolve
+    /// to previous behavior (before OnChainExecutionConfig was registered) in case of Missing.
+    Missing,
 }
 
 /// The public interface that exposes all values with safe fallback.
@@ -18,6 +21,7 @@ impl OnChainExecutionConfig {
     /// The type of the transaction shuffler being used.
     pub fn transaction_shuffler_type(&self) -> TransactionShufflerType {
         match &self {
+            OnChainExecutionConfig::Missing => TransactionShufflerType::NoShuffling,
             OnChainExecutionConfig::V1(config) => config.transaction_shuffler_type.clone(),
             OnChainExecutionConfig::V2(config) => config.transaction_shuffler_type.clone(),
             OnChainExecutionConfig::V3(config) => config.transaction_shuffler_type.clone(),
@@ -27,6 +31,7 @@ impl OnChainExecutionConfig {
     /// The per-block gas limit being used.
     pub fn block_gas_limit(&self) -> Option<u64> {
         match &self {
+            OnChainExecutionConfig::Missing => None,
             OnChainExecutionConfig::V1(_config) => None,
             OnChainExecutionConfig::V2(config) => config.block_gas_limit,
             OnChainExecutionConfig::V3(config) => config.block_gas_limit,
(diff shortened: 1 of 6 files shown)


sitalkedia (approved):
LGTM

types/src/on_chain_config/execution_config.rs
```diff
@@ -106,7 +106,7 @@ pub struct ExecutionConfigV3 {
 impl Default for ExecutionConfigV3 {
     fn default() -> Self {
         Self {
-            transaction_shuffler_type: TransactionShufflerType::NoShuffling,
+            transaction_shuffler_type: TransactionShufflerType::SenderAwareV2(32),
             block_gas_limit: None,
```
sitalkedia, line 110: @danielxiangzl - Should we enable block gas limit by default for Forge and other testings?

types/src/on_chain_config/execution_config.rs
```diff
@@ -106,7 +106,7 @@ pub struct ExecutionConfigV3 {
 impl Default for ExecutionConfigV3 {
```
igor-aptos, line 106: if we want to permanently have different things, I would have :
`default_for_genesis`
`default_if_missing`
basically `default_for_genesis`  is used for any new network. `default_if_missing`  is the backward compatible thing (needed for reply for example), to have old behavior if flag is missing

and I would remove Default macro completely, so people don't explicitly pick the one they want.

currently it is not critical, but if at any point in the future we add anything in the execution config that affects transaction output (not the order/which ones get on chain), we would need to have a separation, as replay will start with missing config

types/src/on_chain_config/execution_config.rs
danielxiangzl: Yes, but only after the execution on-chain config registry is fixed. I still find it hacky to set the config values via default for test networks. Maybe do what Igor suggested and set the block gas limit to be 35000 for genesis?

sitalkedia: This will be used only for Forge and existing integration tests, so I say we enable gas limit so that our tests exercise the block gas limit before rolling this out. For testnet/mainnet, this is no-op as they are still using the V1.

bchocho: Actually I checked and we are on V3 in mainnet and testnet. But it still shouldn't matter, right?

danielxiangzl: If the execution on-chain config registry is not fixed, and main/testnet reads config via default (V3), then it will be a problem. So just to double check, is it fixed already?

sitalkedia: How are we on V3 in mainnet? If so, I think this is a dangerous change to make..

bchocho: The governance proposal for v1.5 was V3

sitalkedia: Since the registry fix is not rolled out yet, I think we should not enable shuffling by default in V3 - if we are already in V3 in mainnet. The reason for that being if by any chance we roll out this change without the registry fix, then we break the network.

types/src/on_chain_config/execution_config.rs
```diff
@@ -11,13 +11,17 @@ pub enum OnChainExecutionConfig {
     V1(ExecutionConfigV1),
     V2(ExecutionConfigV2),
     V3(ExecutionConfigV3),
+    /// To maintain backwards compatibility on replay, we must ensure that any new features resolve
+    /// to previous behavior (before OnChainExecutionConfig was registered) in case of Missing.
+    Missing,
```
sitalkedia, line 16: It is pretty easy to make a mistake and add the next version `V4` after `V3` and before `Missing` and thus breaking backward compatibility. Can you add a comment here to add V4 after `Missing` ?

types/src/on_chain_config/execution_config.rs
```diff
@@ -106,8 +118,8 @@ pub struct ExecutionConfigV3 {
 impl Default for ExecutionConfigV3 {
     fn default() -> Self {
```
sitalkedia, line 119: Is this being used anywhere? If so, better to change them to `default_for_genesis` or `default_if_missing` and get rid of this?

sitalkedia: Nevermind, just realized, this is called from `default_for_genesis` function, but still may be it will be less confusing to remove `Default` trait for `ExecutionConfigV3` and return the desired value from `default_for_genesis` function..

igor-aptos (approved):
Thanks!

types/src/on_chain_config/execution_config.rs
igor-aptos: Agreed with this,  maybe even just inline there

sitalkedia (approved):
Thanks for fixing the config mess @bchocho 🙌

types/src/on_chain_config/execution_config.rs
igor-aptos: Let's add V4 below Missing, and return no shuffling in v3? Then we don't need new shuffling enum either

PR #9398, opened 2023-07-31
[CLI] update local testnet validator initial stake 1 to 10

0xjinn:
### Description

I need to increase this initial validator stake amount from 1 to 10. When it was 1, I cannot join a validator set locally, because it will always violate EVOTING_POWER_INCREASE_EXCEEDS_LIMIT

After increase to 10, I can join this validator set with an initial stake of 1, and 1/10 is less than power increase limit of 50%. 

### Test Plan

Double check and make sure the stake is updated in local testnet

```
➜  aptos-core git:(main) ✗ aptos node show-validator-set --profile=local
{
  "Result": {
    "scheme": "BLS12381",
    "active_validators": [
      {
        "account_address": "5879e6a0adab3b89e7a10ef1a9de4eedbe5d666b51ee2cdb24d73f490e58ff61",
        "consensus_voting_power": 10,
        "config": {
          "consensus_public_key": "0x8c4a898fe746045e79a8e6f2ce2b3c7ad3d88880148208d2a306b45481cc62e17790b75a03e1a1e7117866834d6f2033",
          "validator_network_addresses": [
            "/ip4/0.0.0.0/tcp/6180/noise-ik/0x657dac6c023b39b53b8ae5fdc13a1e1222dacde98165324e16086a6766821628/handshake/0"
          ],
          "fullnode_network_addresses": [
            "/ip4/0.0.0.0/tcp/6181/noise-ik/0x61f5a3ffc4bd744a898e865e23d05fdf52658225721dbf8adb12bb913af5ab4b/handshake/0"
          ],
          "validator_index": 0
        }
      }
    ],
    "pending_inactive": [],
    "pending_active": [],
    "total_voting_power": 10,
    "total_joining_power": 0
  }
}
```


0xmaayan (approved):

gregnazario (approved):
We'll have to see if something will fail in testing because of a change in the genesis

0xjinn:
@gregnazario Yeah, I was checking for dependency on that initial stake of 1, but didn't find anything. So I think we should be good there, but we'll see.

PR #9406, opened 2023-08-01
Add aptos move generate schema command

banool:
### Stack
- Previous in stack: https://github.com/aptos-labs/aptos-core/pull/9405
- Next in stack: https://github.com/aptos-labs/aptos-core/pull/9407

### Description
This PR adds `aptos move generate schema`, which lets you generate a GraphQL schema with types that correspond to the structs in a Move module.

### Test Plan
Tested manually for now. If we like the PR I can add a CLI E2E test.

From inside a Move package directory:
```
aptos move generate schema --named-addresses addr=0x3
```

You get a schema like this: https://gist.github.com/banool/cbfbb0956c155ef018ef7da353a43323.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -246,6 +246,7 @@ dependencies = [
  "aptos-ledger",
  "aptos-logger",
  "aptos-move-debugger",
+ "aptos-move-graphql-schema",
  "aptos-network-checker",
  "aptos-node",
  "aptos-protos 1.1.2",
--- a/crates/aptos/Cargo.toml
+++ b/crates/aptos/Cargo.toml
@@ -35,6 +35,7 @@ aptos-keygen = { workspace = true }
 aptos-ledger = { workspace = true }
 aptos-logger = { workspace = true }
 aptos-move-debugger = { workspace = true }
+aptos-move-graphql-schema = { workspace = true }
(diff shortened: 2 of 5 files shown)


davidiw:
Do we have an issue or design doc associated with this? How is this going to be used, where? How globally useful is this? E.g., how does this apply to TS, Python, and other SDKs? With an emphasis first on Rust, it isn't clear to me.

Ideally this would also line up with killing off the aptos-sdk-builder...

Also curious whether or not this could also be applied to BCS, since that would be a precursor to eliminate that and supporting function calls.

I think this is probably all going in the right direction, but just in case it isn't...

banool:
So I have a rather long list of docs on this topic here: https://www.notion.so/aptoslabs/Leveraging-ABIs-Index-ba0e20ffb39640f6b155d2233bebaeda. This tool is just the realization of the "generated reader" described there.

The tool works for Python, Rust, and TS, and probably any language with a decent GraphQL based code generator. This doc contains a guide for using this toolset for Rust, Python, and TS: https://www.notion.so/aptoslabs/Leveraging-ABIs-Prototype-Demonstration-with-v1-resource-representation-e8f09131626d4f6da9bdb76c4829a7f9.

Issue #9350, opened 2023-07-27
[Bug] dev_setup.sh installs the wrong version of solc for Mac Arm64


brmataptos:
# 🐛 Bug

Script was changed to use Arm64 version of Solidity installed from brew.  Unfortunately, the newer version installed breaks tests in third-party/evm/.  The Amd64 version may be slower, since it runs through emulation, but tests work with it.  This prevents local building of nextest unit tests on developer laptop.

Getting the old version of the Arm64 version from brew requires a lot of work, and it doesn't seem worthwhile for now.

PR #9412, opened 2023-08-01
Update dev_setup.sh to get right SOLC version for Mac Arm64

brmataptos:
Changed to use Amd64 version of Solidity, since the correct version is not available precompiled for Arm64. The Amd64 version may be slower, since it runs through emulation, but tests work with it.

### Description

Just use the AMD64 version of SOLC_BIN instead of installing the wrong version through Homebrew (which breaks the Move EVM tests).  Fixes https://github.com/aptos-labs/aptos-core/issues/9350

### Test Plan

Check out and run the script on a fresh Arm64 Mac.  Perhaps also on another machine, to ensure no anticipated breakage.


brianolson:
I don't like it. ARM Mac has been out over 2.5 years and we should have native versions of tools by now.

brmataptos:
To provide the needed functionality to build/test current code we need these older versions of Z3 and Solidity, which don't have standard Homebrew versions.  We can easily set up a private repo, but that's a big distraction for me.  If you want to do that, feel free.

This is urgent to fix: I just wasted days of work trying to debug tests before realizing that they only work with the old version of Solidity.  Each new Move SWE we get will keep bumping into these poorly tested dev_script changes.

brmataptos:
Btw, an alternative workaround solution is to avoid needing Solidity by deprecating and/or deleting the EVM and Async Move code, which has been discussed, and Wolfgang feels strongly about keeping them.  While I don't care about those particular variants, I feel that keeping some variant dialects around to exercise "flavor" code paths helps to make sure we don't break those paths.

brmataptos:
> I don't like it. ARM Mac has been out over 2.5 years and we should have native versions of tools by now.

I've filed a bug to request a better fix in https://github.com/aptos-labs/aptos-core/issues/9478 but can we please have this workaround for now?

JoshLind (approved):
Unblocking this for now 😄 Seems reasonable to me given the constraints. Thanks @brianrmurphy!

vgao1996 (approved):
Sounds like a reasonable workaround for now. 

Just to clarify, is the arm64 binary built by brew totally broken, or is it producing results that are different the old versions, making it incompatible with the EVM tests? In the latter case maybe we should come back and fix the tests.

@wrwg @runtian-zhou @rahxephon89 @junkil-park let's also discuss how we'll want to maintain the move-on-evm project going forward.

junkil-park:
By the way, is the move-on-evm the only project that needs the Solidity compiler? Or, do we have any other places in `aptos-core` where `solc` is used?

brmataptos:
> Just to clarify, is the arm64 binary built by brew totally broken, or is it producing results that are different the old versions, making it incompatible with the EVM tests? In the latter case maybe we should come back and fix the tests.

None of the EVM tests were building successfully.  Wolfgang wants to keep EVM, but at some point more work may be needed to fix however we're interacting with the Solidity.  It is only reachable in some #[cfg(feature = "evm-backend")] in build_plan.rs, but is actually only used in code under /third_party/move/evm/.

aalok-t (commented):
Ran this on arm64 and works well. Thanks for this workaround.

brmataptos:
Btw, you can see the previous issue by running (from the CI build scripts):

`cargo nextest run --profile ci --locked --workspace --exclude smoke-test --exclude aptos-testcases --retries 3 --no-fail-fast `

in aptos-core/.  If the evm tests work, then the fix works.

PR #9414, opened 2023-08-01
[event v2] e2e module event

lightmark:
### Description
This PR did the following change:
1. add ContractEvent::V1(). Due to `Deref` of ContractEvent that simplify a lot previous work as a tech debt and as a result, I have to pay the price now.
2. add the API in move module to emit event v2.
3. Add in extended_checks that saves event attributes to metadata.
4. add native functions for the move api with runtime checks of `#[event]` by step 3.
5. refactored move-vm to remove all events code. Instead we created an event native context to store both v1 and v2 events.
6. add https://github.com/aptos-labs/aptos-core/pull/9507 and add logic to limit T to struct defined the same module.
7. add event_validation to `validate_publish_request` implementing the same constraint as 6.
8. add MODULE_EVENT feature.
9. add `event.move` to move-examples.
10. add event test to e2e-move-test on top of 9.

### Test Plan
move-unit-test
attribute-test

manually test with local testnet.
1. test compiler error out if `T` is not defined in the same module or `T` does not have `#[event]`.
2. turn off compiler check code, and check `event_validation` works the same as compiler check.


third_party/move/tools/move-cli/src/sandbox/utils/mod.rs
```diff
@@ -151,21 +151,10 @@ fn print_struct_diff_with_indent(
...
-    if !events.is_empty() {
-        println!("Emitted {:?} events:", events.len());
-        // TODO: better event printing
-        for (event_key, event_sequence_number, _event_type, event_data) in events {
-            println!(
-                "Emitted {:?} as the {}th event to stream {:?}",
-                event_data, event_sequence_number, event_key
-            )
-        }
-    }
```
msmouse, line 168: Events intentionally removed from the "sandbox" (which I don't know what it is)? Is it okay?

types/src/contract_event.rs
```diff
@@ -93,6 +129,16 @@ pub struct ContractEventV0 {
...
 
+/// Entry produced via a call to the `emit` builtin.
+#[derive(Hash, Clone, Eq, PartialEq, Serialize, Deserialize, CryptoHasher)]
+pub struct ContractEventV1 {
+    /// The type of the data
+    type_tag: TypeTag,
+    /// The data payload of the event
+    #[serde(with = "serde_bytes")]
+    event_data: Vec<u8>,
+}
```
msmouse: nit: I think it's slightly better to move this declaration down to right before its own implementation.

types/src/contract_event.rs
```diff
@@ -129,87 +175,150 @@ impl ContractEventV0 {
...
 impl TryFrom<&ContractEvent> for NewEpochEvent {
     type Error = Error;
 
     fn try_from(event: &ContractEvent) -> Result<Self> {
-        if event.type_tag != TypeTag::Struct(Box::new(Self::struct_tag())) {
-            anyhow::bail!("Expected NewEpochEvent")
+        match event {
+            ContractEvent::V0(event) => {
+                if event.type_tag != TypeTag::Struct(Box::new(Self::struct_tag())) {
+                    anyhow::bail!("Expected NewEpochEvent")
```
msmouse, line 255: nit: import `bail!()`?

types/src/contract_event.rs
```diff
@@ -129,87 +175,150 @@ impl ContractEventV0 {
...
-            )
-        } else if let Ok(payload) = DepositEvent::try_from(self) {
-            write!(
-                f,
-                "ContractEvent {{ key: {}, index: {:?}, type: {:?}, event_data: {:?} }}",
-                self.key, self.sequence_number, self.type_tag, payload,
-            )
-        } else {
-            write!(f, "{:?}", self)
+        if let ContractEvent::V0(v0) = self {
```
msmouse: nit: the `try_from()`'s match `V0` anyway, so this seems redundant?

types/src/contract_event.rs
```diff
@@ -129,87 +175,150 @@ impl ContractEventV0 {
...
+        write!(f, "{:?}", self)
     }
 }
 
 #[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
 #[cfg_attr(any(test, feature = "fuzzing"), derive(Arbitrary))]
 pub struct EventWithVersion {
-    pub transaction_version: u64, // Should be `Version`
+    pub transaction_version: u64,
+    // Should be `Version`
```
msmouse, line 330: I think this comment was for the previous field? (maybe just remove?)

msmouse (approved):
LGTM

storage/aptosdb/src/event_store/mod.rs
```diff
@@ -471,19 +471,19 @@ impl EventStore {
...
-                    None
-                }
-            }) {
-                db_batch.delete::<EventByVersionSchema>(&(
-                    *event.key(),
-                    current_version,
-                    event.sequence_number(),
-                ))?;
-                db_batch.delete::<EventByKeySchema>(&(*event.key(), event.sequence_number()))?;
+            for event in (events?).into_iter().enumerate() {
```
msmouse, line 474: what's wrong with `for (idx, event) in ...` here?

aptos-move/aptos-vm/src/verifier/event_validation.rs
```diff
@@ -0,0 +1,149 @@
...
+                        })?;
+                    match param {
+                        StructInstantiation(index, _) | Struct(index) => {
+                            let struct_handle = &module.struct_handle_at(*index);
+                            let struct_name = module.identifier_at(struct_handle.name);
+                            if struct_handle.module != module.self_handle_idx() {
+                                metadata_validation_err(format!("{} passed to 0x1::event::emit function is not defined in the same module", struct_name).as_str())
+                            } else if !event_structs.contains(struct_name.as_str()) {
+                                metadata_validation_err(format!("#[event] attribute is missing for {} passed to 0x1::event::emit function", struct_name).as_str())
+                            } else {
```
davidiw, line 104: `Missing #[event] attribute on {}. The #[event] attribute is required for all structs passed into 0x1::event::emit`.

aptos-move/aptos-vm/src/verifier/event_validation.rs
```diff
@@ -0,0 +1,149 @@
...
+                                Ok(())
+                            }
+                        },
+                        _ => metadata_validation_err(
+                            "Reference to non-struct parameter for 0x1::event::emit function",
```
davidiw, line 109: `Passed in a non-struct parameter into 0x1::event::emit.`

aptos-move/aptos-vm/src/verifier/event_validation.rs
```diff
@@ -0,0 +1,149 @@
...
+    let metadata = session.load_module(module_id).map(|module| {
+        CompiledModule::deserialize(&module)
+            .map(|module| aptos_framework::get_metadata_from_compiled_module(&module))
+    });
+
+    if let Ok(Ok(Some(metadata))) = metadata {
+        extract_event_metadata(&metadata)
+    } else {
+        Ok(HashSet::new())
+    }
```
davidiw, line 133: shouldn't we fail if we can't load the module or can't deserialize it? granted I don't see how we get there, but ... it is a bit concerning we're discarding errors in line 129

aptos-move/aptos-vm/src/verifier/event_validation.rs
```diff
@@ -0,0 +1,149 @@
...
+/// Given a module id extract all event metadata
+pub(crate) fn extract_event_metadata(
+    metadata: &RuntimeModuleMetadataV1,
+) -> VMResult<HashSet<String>> {
+    let mut event_structs = HashSet::new();
+    for (struct_, attrs) in &metadata.struct_attributes {
+        for attr in attrs {
+            if attr.is_event() && !event_structs.insert(struct_.clone()) {
+                metadata_validation_err("Found duplicate event attribute")?;
+            }
```
davidiw, line 145: again we have this logic where we validate the logic on the metadata struct attributes?

aptos-move/framework/aptos-framework/sources/event.move
```diff
@@ -6,10 +6,33 @@ module aptos_framework::event {
...
+    use std::features;
 
     friend aptos_framework::account;
     friend aptos_framework::object;
 
+    /// Module event feature is not supported.
+    const EMODULE_EVENT_NOT_SUPPORTED: u64 = 1;
+
+    /// Emit an event with payload `msg` by using `handle_ref`'s key and counter.
+    public fun emit<T: store + drop>(msg: &T) {
```
davidiw, line 18: Per conversations `&T` -> `T` or pending alternative outcomes...

aptos-move/framework/aptos-framework/sources/event.move
```diff
@@ -6,10 +6,33 @@ module aptos_framework::event {
...
+        assert!(features::bulletproofs_enabled(), std::error::invalid_state(EMODULE_EVENT_NOT_SUPPORTED));
```
davidiw, line 19: huh which feature?

aptos-move/framework/src/extended_checks.rs
```diff
@@ -428,6 +435,87 @@ impl<'a> ExtendedChecker<'a> {
...
+                            .symbol_pool()
+                            .string(struct_.get_name())
+                            .to_string(),
+                    )
+                    .or_default()
+                    .push(KnownAttribute::event());
+            }
+        }
+        for fun in module.get_functions() {
+            if fun.is_inline() || fun.is_native() {
```
davidiw, line 462: why don't we check inline functions? is that because it is already expressed in the full form of an actual function?

aptos-move/framework/src/extended_checks.rs
```diff
@@ -428,6 +435,87 @@ impl<'a> ExtendedChecker<'a> {
...
+                        self.env
+                            .symbol_pool()
+                            .string(struct_.get_name())
+                            .to_string(),
+                    )
+                    .or_default()
+                    .push(KnownAttribute::event());
+            }
+        }
+        for fun in module.get_functions() {
```
davidiw, line 461: what happens if the function is a generic function?

execution/executor-types/src/parsed_transaction_output.rs
```diff
@@ -16,7 +16,13 @@ pub struct ParsedTransactionOutput {
 
 impl ParsedTransactionOutput {
     pub fn parse_reconfig_events(events: &[ContractEvent]) -> impl Iterator<Item = &ContractEvent> {
-        events.iter().filter(|e| *e.key() == *NEW_EPOCH_EVENT_KEY)
+        events.iter().filter(|e| {
+            if let ContractEvent::V1(v1) = e {
+                *v1.key() == *NEW_EPOCH_EVENT_KEY
+            } else {
+                false
+            }
```
davidiw, line 24: might be good to just add these acceors to the ContractEvent itself rather than having to always do this match statement nonsense...

third_party/move/tools/move-cli/src/sandbox/utils/mod.rs
davidiw: we are stripping out move events from the system, since there really isn't such a thing

davidiw (changes requested):
This is pretty incredible... Lots of small feedbacks and waiting on alignment on `&T`

One other thing to consider is whether or not you can introduce a generic 

`fn 0xbob::Module::emit<T>()` that calls `0x1::event::emit` or would that fail?

aptos-move/framework/src/extended_checks.rs
davidiw: Leave a comment please

davidiw:
BTW doesn't this need to be added to releases to be  tested by forge?

aptos-move/framework/src/extended_checks.rs
```diff
@@ -428,6 +435,87 @@ impl<'a> ExtendedChecker<'a> {
     }
 }
 
+// ----------------------------------------------------------------------------------
+// Events
+
+impl<'a> ExtendedChecker<'a> {
+    fn check_and_record_events(&mut self, module: &ModuleEnv) {
```
movekevin, line 442: Does this run for scripts at all? I wonder how we can disable emit event calls for scripts.

aptos-move/move-examples/event/sources/event.move
```diff
@@ -0,0 +1,54 @@
...
+
+    public entry fun emit(num: u64) {
+        let i = 0;
+        while (i < num) {
+            let event = MyEvent {
+                seq: i,
+                field: Field { field: false },
+                bytes: vector[]
+            };
+            event::emit(&event);
```
movekevin, line 27: Can you add a quick test/function for emitting events in an inline function as well?

aptos-move/move-examples/event/sources/event.move
```diff
@@ -0,0 +1,54 @@
...
+    use aptos_framework::event;
+    #[test_only]
+    use std::vector;
+
+    struct Field has store, drop {
+        field: bool,
+    }
+
+    #[event]
+    struct MyEvent has store, drop {
```
movekevin, line 13: What happens if you have weird data structures in here such as Table?

api/types/src/transaction.rs
```diff
@@ -529,10 +536,16 @@ pub struct Event {
...
-            ContractEvent::V0(v0) => Self {
-                guid: (*v0.key()).into(),
-                sequence_number: v0.sequence_number().into(),
-                typ: v0.type_tag().clone().into(),
+            ContractEvent::V1(v1) => Self {
+                guid: (*v1.key()).into(),
+                sequence_number: v1.sequence_number().into(),
+                typ: v1.type_tag().clone().into(),
+                data,
+            },
```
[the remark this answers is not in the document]
lightmark: I was doing v0/v1 and @msmouse suggested to v1/v2. I think it is consistent with naming as v1/v2 doesn't hurt serialization as enums. It is worthwhile to change that in rust code to me.

api/types/src/transaction.rs
```diff
@@ -557,11 +570,18 @@ pub struct VersionedEvent {
...
     fn from((event, data): (&EventWithVersion, serde_json::Value)) -> Self {
         match &event.event {
-            ContractEvent::V0(v0) => Self {
+            ContractEvent::V1(v1) => Self {
+                version: event.transaction_version.into(),
+                guid: (*v1.key()).into(),
+                sequence_number: v1.sequence_number().into(),
+                typ: v1.type_tag().clone().into(),
+                data,
+            },
```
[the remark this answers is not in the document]
lightmark: not really. Without that naming change, I still to add those code everywhere because `Deref` is not working anymore since we have a second discriminant.

aptos-move/framework/aptos-framework/sources/event.move
lightmark: This was a mistake. I changed later.

aptos-move/framework/src/extended_checks.rs
lightmark: copied from @wrwg 's PR. I think it is `0x1::event::emit` is not an inline.

aptos-move/move-examples/event/sources/event.move
lightmark: I think the handle will be serialized.

aptos-move/aptos-vm/src/verifier/event_validation.rs
```diff
@@ -0,0 +1,149 @@
...
+                            metadata_validation_error(
+                                "Missing parameter for 0x1::event::emit function",
+                            )
+                        })?;
+                    match param {
+                        StructInstantiation(index, _) | Struct(index) => {
+                            let struct_handle = &module.struct_handle_at(*index);
+                            let struct_name = module.identifier_at(struct_handle.name);
+                            if struct_handle.module != module.self_handle_idx() {
+                                metadata_validation_err(format!("{} passed to 0x1::event::emit function is not defined in the same module", struct_name).as_str())
```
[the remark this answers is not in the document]
lightmark: We are checking that event struct emitted from this module is defined in this module from bytecode. Why is it duplicate?
Compiler will do that in extended_checks but people can bypass that.

aptos-move/aptos-vm/src/verifier/event_validation.rs
lightmark: I think I just copied what resource_group does in this case without thinking through it too much. If we wanna change, I think we should change both places but it may has replay issues.

aptos-move/aptos-vm/src/verifier/event_validation.rs
lightmark: do we have that? I don't think so...

aptos-move/framework/src/extended_checks.rs
lightmark: it is also in `FunctionEnv`. Generic functions have non-empty `type_params` in `FunctionData`

aptos-move/framework/src/extended_checks.rs
lightmark: confirmed with him that you're right. Here are just declarations and the actually function called this inline function would have the actual check.

aptos-move/framework/src/extended_checks.rs
lightmark: maybe not. I think we are going to solve the issue.

aptos-move/aptos-vm/src/verifier/event_validation.rs
```diff
@@ -0,0 +1,149 @@
...
+    Err(metadata_validation_error(msg))
+}
+
+fn metadata_validation_error(msg: &str) -> VMError {
+    PartialVMError::new(StatusCode::CONSTRAINT_NOT_SATISFIED)
+        .with_message(format!("metadata and code bundle mismatch: {}", msg))
+        .finish(Location::Undefined)
+}
+
+/// Validate resource group metadata on modules one by one:
```
movekevin, line 33: Comment needs to be updated

aptos-move/aptos-vm/src/verifier/event_validation.rs
```diff
@@ -0,0 +1,149 @@
...
+/// * Extract the event metadata
+/// * Verify all changes are compatible upgrades (existing event attributes cannot be removed)
+pub(crate) fn validate_module_events(
```
movekevin, line 36: cc @brianrmurphy I think this new attribute might need to be added somewhere in the compiler so it doesn't generate warnings for unknown attributes

aptos-move/aptos-vm/src/verifier/event_validation.rs
```diff
@@ -0,0 +1,149 @@
...
+    Ok(())
+}
+
+/// Validate all the `0x1::event::emit` calls have the struct defined in the same module with event
+/// attribute.
+pub(crate) fn validate_emit_calls(
+    event_structs: &HashSet<String>,
+    module: &CompiledModule,
+) -> VMResult<()> {
+    for fun in module.function_defs() {
```
movekevin, line 70: How expensive is it if a module has a lot of emit functions? Gas is not currently charged for verifier work. This is just limited to module publishing but just in case this is expensive enough

aptos-move/e2e-move-tests/src/tests/module_event.rs
```diff
@@ -0,0 +1,116 @@
...
+#[test]
+fn verify_module_event_upgrades() {
+    let mut h = MoveHarness::new_with_features(vec![FeatureFlag::MODULE_EVENT], vec![]);
+    let account = h.new_account_at(AccountAddress::from_hex_literal("0xf00d").unwrap());
+
+    // Initial code
+    let source = r#"
+        module 0xf00d::M {
+            #[event]
+            struct Event1 { }
```
movekevin, line 75: Should we also verify in the verifier code that event structs have the right abilities?

aptos-move/e2e-move-tests/src/tests/module_event.rs
```diff
@@ -0,0 +1,116 @@
...
+}
+
+#[derive(Debug, Serialize, Deserialize, Eq, PartialEq)]
+struct MyEvent {
+    seq: u64,
+    field: Field,
+    bytes: Vec<u64>,
+}
+
+#[test]
```
movekevin, line 23: Can we also tests for more cases such as:
1. Publishing modules with event emit calls with non-event structs.
2. Events with nested types (phantom vs not)
3. etc.

movekevin (approved):

aptos-move/e2e-move-tests/src/tests/module_event.rs
lightmark: 1. The compilation will not go through.

aptos-move/framework/src/natives/event.rs
```diff
@@ -50,11 +90,20 @@ fn native_write_to_event_store(
         EVENT_WRITE_TO_EVENT_STORE_BASE
             + EVENT_WRITE_TO_EVENT_STORE_PER_ABSTRACT_VALUE_UNIT * context.abs_val_size(&msg),
     )?;
+    let ty_tag = context.type_to_type_tag(&ty)?;
+    let ty_layout = context.type_to_type_layout(&ty)?;
+    let blob = msg.simple_serialize(&ty_layout).ok_or_else(|| {
+        SafeNativeError::InvariantViolation(PartialVMError::new(
+            StatusCode::VALUE_DESERIALIZATION_ERROR,
```
vgao1996, line 97: Please change to another status code under the invariant violation category. `SafeNativeError::InvariantViolation` is only supposed to take invariant violations. Use `UNKNOWN_INVARIANT_VIOLATION_ERROR` if you are not sure what to use.

aptos-move/framework/src/natives/event.rs
```diff
@@ -79,21 +128,121 @@ fn native_emitted_events_by_handle(
...
+    let ty_tag = context.type_to_type_tag(&ty)?;
+    let ty_layout = context.type_to_type_layout(&ty)?;
+    let ctx = context.extensions_mut().get_mut::<NativeEventContext>();
+    let events = ctx
+        .emitted_v1_events(&key, &ty_tag)?
+        .into_iter()
+        .map(|blob| {
+            Value::simple_deserialize(blob, &ty_layout).ok_or_else(|| {
+                SafeNativeError::InvariantViolation(PartialVMError::new(
+                    StatusCode::VALUE_DESERIALIZATION_ERROR,
```
vgao1996, line 155: Same here. VALUE_DESERIALIZATION_ERROR is not an invariant violation.

vgao1996 (changes requested):

aptos-move/e2e-move-tests/src/tests/module_event.rs
lightmark: verifier does not have metadata info...

lightmark:
@davidiw @vgao1996 Comments addressed.

aptos-move/aptos-vm/src/aptos_vm.rs
```diff
@@ -994,6 +1000,7 @@ impl AptosVM {
                 .map_err(|err| Self::metadata_validation_error(&err.to_string()))?;
         }
         verifier::resource_groups::validate_resource_groups(session, modules)?;
+        verifier::event_validation::validate_module_events(session, modules)?;
```
vgao1996, line 1003: As we just discussed offline, this needs to be feature gated

lightmark: As just discussed offline, line 1000 is already feature gated so this line is good.

aptos-move/aptos-vm/src/verifier/event_validation.rs
```diff
@@ -0,0 +1,184 @@
...
+
+        let original_event_structs =
+            extract_event_metadata_from_module(session, &module.self_id())?;
+
+        for member in original_event_structs {
+            // Fail if we see a removal of an event attribute.
+            if !new_event_structs.remove(&member) {
+                metadata_validation_err("Invalid change in event attributes")?;
+            }
+        }
```
vgao1996, line 59: @wrwg it's a bit hacky that we have to implement the compatibility check logic here in a slightly different way. Really should offer a way to extend the logic of the original backward compatibility checker.

vgao1996 (approved):
LGTM for now

aptos-move/aptos-vm/src/verifier/event_validation.rs
lightmark: I fixed that

aptos-move/aptos-vm/src/verifier/event_validation.rs
lightmark: it should be included in the base fee

lightmark:
@davidiw

davidiw (commented):
You haven't implemented events v2 for the API when reading transaction outputs nor have you added supported for the grpc indexer code.

For the API, you could take a look at my resource group / tokenv2 stuff.

I don't know how high a bar we want for the grpc, but it could be as simple as making a v2 look like a v1 and leaving it for future work...

lightmark:
per `when reading transaction outputs`. I can add that quicky.
per `proto` it's ready because the struct can be reused, and I just use a dummy `eventID` + seq 0 for all v2 events. The only thing needed for indexer team is to distinguish them here: https://github.com/aptos-labs/aptos-indexer-processors/blob/496ccd90b89a16040bdeb7fc9c9bf0e18be9aa5f/rust/processor/src/processors/default_processor.rs#L278-L283

lightmark:
replay verify https://github.com/aptos-labs/aptos-core/actions/runs/5932514429

api/src/tests/events_test.rs
```diff
@@ -136,6 +138,48 @@ async fn test_get_events_by_invalid_account_event_handle_field_type() {
     context.check_golden_output(resp);
 }
 
+#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
```
davidiw, line 141: good man

davidiw (approved):

PR #9417, opened 2023-08-01
[Forge] PFN should have a container name of fullnode, not validator

bchocho:
### Test Plan

Confirm the PFN is no longer labeled a validator in grafana system dashboard and humio.


rustielin (approved):
nice catch!

JoshLind (approved):

PR #9421, opened 2023-08-02
[TS SDK] Support local endpoint in Provider

0xmaayan:
### Description
relive https://github.com/aptos-labs/aptos-core/pull/8850 - support `Network.LOCAL` enum for local development. It isn't ideal though, because an indexer client still exists, it just won't work correctly.

### DIFF
--- a/ecosystem/typescript/sdk/src/providers/provider.ts
+++ b/ecosystem/typescript/sdk/src/providers/provider.ts
@@ -14,6 +14,10 @@ type NetworkWithCustom = Network | "CUSTOM";
  * This class holds both AptosClient and IndexerClient classes's methods and properties so we
  * can instantiate the Provider class and use it to query full node and/or Indexer.
  *
+ * NOTE: Indexer client can be undefined/not set when we use Network.LOCAL (since Indexer
+ * does not support local environment) or when we use a CUSTOM network to support applications
+ * that only use custom fullnode and not Indexer
+ *
  * @example An example of how to use this class with a live network
  * ```
  * const provider = new Provider(Network.DEVNET)
@@ -28,7 +32,7 @@ type NetworkWithCustom = Network | "CUSTOM";
  * const account = await provider.getAccount("0x123");
  * ```
  *
- * @example An example of how to use this class with a lcustom network.
+ * @example An example of how to use this class with a custom network.
  * ```
  * const provider = new Provider({fullnodeUrl:"my-fullnode-url",indexerUrl:"my-indexer-url"})
  * const account = await provider.getAccount("0x123");
@@ -36,7 +40,7 @@ type NetworkWithCustom = Network | "CUSTOM";
  * ```
  *
  * @param network enum of type Network - MAINNET | TESTNET | DEVNET | LOCAL or custom endpoints of type CustomEndpoints
- * @param config AptosClient config arg - additional configuration options for the generated Axios client.
+ * @param config optional ClientConfig config arg - additional configuration we can pass with the request to the server.
  */
 export class Provider {
   aptosClient: AptosClient;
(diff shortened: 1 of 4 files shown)


ecosystem/typescript/sdk/src/utils/api-endpoints.ts
```diff
@@ -2,24 +2,28 @@ export const NetworkToIndexerAPI: Record<string, string> = {
...
   testnet: "https://fullnode.testnet.aptoslabs.com/v1",
   devnet: "https://fullnode.devnet.aptoslabs.com/v1",
+  local: "http://localhost:8080/v1",
 };
 
 export const NodeAPIToNetwork: Record<string, string> = {
   "https://fullnode.mainnet.aptoslabs.com/v1": "mainnet",
   "https://fullnode.testnet.aptoslabs.com/v1": "testnet",
   "https://fullnode.devnet.aptoslabs.com/v1": "devnet",
+  "http://localhost:8080/v1": "local",
```
xbtmatt, line 19: I can't remember, was this added then removed before because it broke the ci tests?

ecosystem/typescript/sdk/src/providers/provider.ts
```diff
@@ -14,14 +14,28 @@ type NetworkWithCustom = Network | "CUSTOM";
...
+ * ```
+ *
+ * @example An example of how to use this class with a lcustom network.
+ * ```
+ * const provider = new Provider({fullnodeUrl:"my-fullnode-url",indexerUrl:"my-indexer-url"})
+ * const account = await provider.getAccount("0x123");
+ * const accountTokens = await provider.getOwnedTokens("0x123");
+ * ```
+ *
+ * @param network enum of type Network - MAINNET | TESTNET | DEVENET | LOCAL or custom endpoints of type CustomEndpoints
```
xbtmatt, line 38: `DEVENET` => `DEVNET`

xbtmatt:
I can't remember, was this added then removed before because it broke the ci tests?

davidiw (approved):

0xmaayan:
> I can't remember, was this added then removed before because it broke the ci tests?

yes. ci didn't pass but the test wasn't land blocking so it got in and then broke next PRs lol

ecosystem/typescript/sdk/src/tests/e2e/provider.test.ts
```diff
@@ -3,26 +3,46 @@ import { Network, NetworkToIndexerAPI, NetworkToNodeAPI } from "../../utils";
...
     const provider = new Provider({ fullnodeUrl: "full-node-url", indexerUrl: "indexer-url" });
     expect(provider.aptosClient.nodeUrl).toBe("full-node-url/v1");
-    expect(provider.indexerClient.endpoint).toBe("indexer-url");
+    expect(provider.indexerClient?.endpoint).toBe("indexer-url");
+  });
+
+  it("does not set indexer client when indexer url is not provided with a custom network", async () => {
+    const provider = new Provider({ fullnodeUrl: "full-node-url" });
+    expect(provider.aptosClient.nodeUrl).toBe("full-node-url/v1");
+    expect(provider.indexerClient).toBe(undefined);
```
banool, line 28: I wonder is there some way we can make this more obvious? Seeing that it is undefined isn't super obvious, maybe some error or explanation somehow? Idk if TS has a great way to do this though, so not a blocker.

banool (approved):

ecosystem/typescript/sdk/src/tests/e2e/provider.test.ts
0xmaayan: interesting! I guess it is not really an error since we can use is without an indexer client. I'll add a note on this on the Provider class itself

PR #9422, opened 2023-08-02
[Python SDK] Add a token_transfer function to the python SDK

xbtmatt:
### Description

Adding a `token_transfer` function to the python client. It is really just an object transfer, but wasn't sure where to put an object transfer for now.

The transfer function and the extra documentation tags are for my next PR for the `your first non-fungible asset` tutorial that I'm updating ASAP.

----

e: Moved the transfer call to the rest client as a generic object_transfer function and updated the token_transfer function to call object_transfer. Added a simple example of it to the end of the `aptos-token.py` example:

```yaml
=== Transferring the Token from Alice to Bob ===
Alice: 0x8a6a0b29ae3ccbf00a2aba89612b19b265ad5162ccce9159b6af424774c45bae
Bob:   0x409dba9c7d4f4d6866071895fe79dbf7ba7db16486f47343ba4be0da849d76a9
Token: 0xd9f96e6cfd4eead5ba84b877b654de72fddf1ecba5abf8eba035be5721821df6

Owner: 0x8a6a0b29ae3ccbf00a2aba89612b19b265ad5162ccce9159b6af424774c45bae
    ...transferring...    
Owner: 0x409dba9c7d4f4d6866071895fe79dbf7ba7db16486f47343ba4be0da849d76a9
```

### DIFF
--- a/ecosystem/python/sdk/aptos_sdk/aptos_token_client.py
+++ b/ecosystem/python/sdk/aptos_sdk/aptos_token_client.py
@@ -9,8 +9,7 @@ from .account import Account
 from .account_address import AccountAddress
 from .async_client import RestClient
 from .bcs import Deserializer, Serializer
-from .transactions import (EntryFunction, TransactionArgument,
-                           TransactionPayload)
+from .transactions import EntryFunction, TransactionArgument, TransactionPayload
 from .type_tag import StructTag, TypeTag
 
 
--- a/ecosystem/python/sdk/aptos_sdk/aptos_token_client.py
+++ b/ecosystem/python/sdk/aptos_sdk/aptos_token_client.py
@@ -9,7 +9,8 @@ from .account import Account
 from .account_address import AccountAddress
 from .async_client import RestClient
 from .bcs import Deserializer, Serializer
-from .transactions import EntryFunction, TransactionArgument, TransactionPayload
+from .transactions import (EntryFunction, TransactionArgument,
+                           TransactionPayload)
 from .type_tag import StructTag, TypeTag
 
 
@@ -383,6 +384,7 @@ class AptosTokenClient:
 
         return TransactionPayload(payload)
 
+    # :!:>create_collection
     async def create_collection(
         self,
         creator: Account,
@@ -401,7 +403,7 @@ class AptosTokenClient:
         tokens_freezable_by_creator: bool,
         royalty_numerator: int,
         royalty_denominator: int,
-    ) -> str:
+    ) -> str:  # <:!:create_collection
         payload = AptosTokenClient.create_collection_payload(
             description,
             max_supply,
@@ -458,6 +460,7 @@ class AptosTokenClient:
 
         return TransactionPayload(payload)
 
+    # :!:>mint_token
     async def mint_token(
         self,
         creator: Account,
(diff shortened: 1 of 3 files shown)


davidiw (commented):
I'm a little hesitant to recommend this, perhaps we could just make `transfer_object` in the async_client::RestClient? Thoughts?

gregnazario:
> I'm a little hesitant to recommend this, perhaps we could just make `transfer_object` in the async_client::RestClient? Thoughts?

Yes, let's add transfer object generically.

If we want to make it more "friendly" we could always just call the transfer object from the token client

xbtmatt:
@gregnazario @davidiw moved it into the rest client as `object_transfer` and made `token_transfer` call it. Added it to the end of the `aptos-token.py` example

Also are there no unit tests for the python sdk? 😛

ecosystem/python/sdk/aptos_sdk/async_client.py
```diff
@@ -742,6 +742,27 @@ async def get_collection(
...
+            TransactionArgument(object, Serializer.struct),
+            TransactionArgument(to, Serializer.struct),
+        ]
+
+        payload = EntryFunction.natural(
+            "0x1::object",
+            "transfer_call",
+            [],
+            transaction_arguments,
+        )
```
gregnazario, line 759: We may want to make this the object input one, so we give people good fundamentals.

gregnazario (approved):

ecosystem/python/sdk/aptos_sdk/aptos_token_client.py
```diff
@@ -515,6 +517,12 @@ async def mint_soul_bound_token(
         )
         return await self.client.submit_bcs_transaction(signed_transaction)
 
+    # :!:>transfer_token
+    async def transfer_token(
+        self, owner: Account, token: AccountAddress, to: AccountAddress
+    ) -> str:
+        return await self.client.transfer_object(owner, token, to) # <:!:transfer_token
+
```
davidiw, line 525: As a minimalist, I'm inclined to oppose this. It masks from the user a real operation and instead gives a pseudo-operation. but ahh... sigh


also why are we showing the body of the function?

it feels inconsistent with the mint? if anything is really actionable, it would be that...

davidiw (approved):

ecosystem/python/sdk/aptos_sdk/aptos_token_client.py
xbtmatt: Hmm, I mean I'm okay with removing it. I just figured it might be nice to have an obvious transfer function in the AptosTokenClient itself. But I see your point, it is most likely unnecessary, especially since this is going in the tutorial.

I'll remove the doc annotation from this and annotate the transfer_object function declaration then just explain how to use it in the tutorial

davidiw: I don't really care all that much, just major issue is we expose the return statement here but not in the mint function.

We won't be changing the SDK all that much in this space, so if it is convenient who am I to argue...

xbtmatt: Yeah I see what you're saying, I only included it so that it's obvious it's just calling the `transfer_object` function under the hood. But I can just explain that

PR #9425, opened 2023-08-02
[ts-sdk example] Adding an example for rotation offer capability and signer offer capability with signed structs

xbtmatt:
### Description

It is sort of unclear how to sign Move structs with the typescript SDK, this PR is two examples that should make it more clear.

It is also an example of how to use the `account::offer_rotate_capability` and `account::offer_signer_capability` Move functions.

Note that for some reason the assertion failed for me a few times even though the offer was correct. Not sure what was going on there or how to reproduce it. It was on localnet, I thought it was a stale sequence number at first but wasn't able to produce the error again after casting the addresses to strings

### Test Plan

`cd ~/aptos-core/ecosystem/typescript/sdk/examples/typescript-esm && pnpm run offer_capabilities`

The scripts assert at the end that the account's resources reflect the new offers (Alice offers to Bob).

### DIFF
--- a/ecosystem/typescript/sdk/examples/typescript-esm/offer_capabilities.ts
+++ b/ecosystem/typescript/sdk/examples/typescript-esm/offer_capabilities.ts
@@ -1,20 +1,53 @@
 import { AptosAccount, FaucetClient, Network, Provider, HexString, TxnBuilderTypes, BCS, Types } from "aptos";
 import assert from "assert";
 
-const CORE_CODE_ADDRESS = new HexString("0x0000000000000000000000000000000000000000000000000000000000000001");
 const ED25519_ACCOUNT_SCHEME = 0;
 
-// Works for both SignerCapabilityOffer and RotationCapabilityOffer
-// Except the structName fields are different and the chainId isn't used in SignerCapabilityOffer
-type CapabilityOfferProofChallengeV2 = {
-  accountAddress: TxnBuilderTypes.AccountAddress;
-  moduleName: string;
-  structName: string;
-  sequenceNumber: number;
-  sourceAddress: TxnBuilderTypes.AccountAddress;
-  recipientAddress: TxnBuilderTypes.AccountAddress;
-  chainId?: number; // not used in SignerCapabilityOffer
-};
+class SignerCapabilityOfferProofChallengeV2 {
+  public readonly moduleAddress: TxnBuilderTypes.AccountAddress = TxnBuilderTypes.AccountAddress.CORE_CODE_ADDRESS;
+  public readonly moduleName: string = "account";
+  public readonly structName: string = "SignerCapabilityOfferProofChallengeV2";
+  public readonly functionName: string = "offer_signer_capability";
+
+  constructor(
+    public readonly sequenceNumber: number,
+    public readonly sourceAddress: TxnBuilderTypes.AccountAddress,
+    public readonly recipientAddress: TxnBuilderTypes.AccountAddress,
+  ) {}
+
+  serialize(serializer: BCS.Serializer): void {
+    this.moduleAddress.serialize(serializer);
+    serializer.serializeStr(this.moduleName);
+    serializer.serializeStr(this.structName);
+    serializer.serializeU64(this.sequenceNumber);
+    this.sourceAddress.serialize(serializer);
+    this.recipientAddress.serialize(serializer);
+  }
+}
+
+class RotationCapabilityOfferProofChallengeV2 {
+  public readonly moduleAddress: TxnBuilderTypes.AccountAddress = TxnBuilderTypes.AccountAddress.CORE_CODE_ADDRESS;
+  public readonly moduleName: string = "account";
+  public readonly structName: string = "RotationCapabilityOfferProofChallengeV2";
+  public readonly functionName: string = "offer_rotation_capability";
+
+  constructor(
+    public readonly chainId: number,
+    public readonly sequenceNumber: number,
+    public readonly sourceAddress: TxnBuilderTypes.AccountAddress,
+    public readonly recipientAddress: TxnBuilderTypes.AccountAddress,
+  ) {}
+
+  serialize(serializer: BCS.Serializer): void {
+    this.moduleAddress.serialize(serializer);
+    serializer.serializeStr(this.moduleName);
+    serializer.serializeStr(this.structName);
+    serializer.serializeU8(this.chainId);
+    serializer.serializeU64(this.sequenceNumber);
+    this.sourceAddress.serialize(serializer);
+    this.recipientAddress.serialize(serializer);
+  }
+}
 
 const createAndFundAliceAndBob = async (
   faucetClient: FaucetClient,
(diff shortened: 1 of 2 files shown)


ecosystem/typescript/sdk/examples/typescript-esm/offer_capabilities.ts
```diff
@@ -0,0 +1,137 @@
...
+  recipientAddress: TxnBuilderTypes.AccountAddress;
+};
+
+const createAndFundAliceAndBob = async (
+  faucetClient: FaucetClient,
+): Promise<{ alice: AptosAccount; bob: AptosAccount }> => {
+  console.log(`-------------------  Creating and funding a new Bob & Alice  -------------------`);
+  const alice = new AptosAccount();
+  const bob = new AptosAccount();
+  await faucetClient.fundAccount(alice.address(), 1 * Math.pow(10, APTOS_COIN_DECIMALS));
```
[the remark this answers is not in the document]
xbtmatt: Forgot to change it from the last one heh, will just use 100_000_000

ecosystem/typescript/sdk/examples/typescript-esm/offer_capabilities.ts
```diff
@@ -0,0 +1,133 @@
...
+const CORE_CODE_ADDRESS = new HexString("0x0000000000000000000000000000000000000000000000000000000000000001");
+const ED25519_ACCOUNT_SCHEME = 0;
+
+// Works for both SignerCapabilityOffer and RotationCapabilityOffer.
+// The structName changes and chainId isn't used in SignerCapabilityOffer
+type CapabilityOfferProofChallengeV2 = {
+  accountAddress: TxnBuilderTypes.AccountAddress;
+  moduleName: string;
+  structName: string;
+  chainId?: number; // not used in SignerCapabilityOffer
```
0xmaayan, line 13: nite: usually optional params are listed at the bottom - not big of a deal here but nice to follow good practices

0xmaayan (approved):

ecosystem/typescript/sdk/examples/typescript-esm/offer_capabilities.ts
xbtmatt: ah yea my bad! Forgot to re-arrange it because I needed the bytes to be in a specific order and was moving the params around without the function, will move it to the end

ecosystem/typescript/sdk/examples/typescript-esm/offer_capabilities.ts
```diff
@@ -0,0 +1,134 @@
...
+      alice,
+      "offer_rotation_capability",
+      {
+        accountAddress: TxnBuilderTypes.AccountAddress.fromHex(CORE_CODE_ADDRESS),
+        moduleName: "account",
+        structName: "RotationCapabilityOfferProofChallengeV2",
+        sequenceNumber: Number((await provider.getAccount(alice.address())).sequence_number),
+        sourceAddress: TxnBuilderTypes.AccountAddress.fromHex(alice.address()),
+        recipientAddress: TxnBuilderTypes.AccountAddress.fromHex(bob.address()),
+        chainId: await provider.aptosClient.getChainId(),
```
gregnazario, line 58: Yeah, actually all things that you do provider.aptosClient, you can actually just do provider, since the interface is on both.

gregnazario, line 58: Generally, I'm in favor of doing these prior to the call instead of inside them for readability / sequencing.  Vs right now, the inner calls happen, then the outer.

ecosystem/typescript/sdk/examples/typescript-esm/offer_capabilities.ts
```diff
@@ -0,0 +1,134 @@
...
+      },
+    );
+    console.log({ hash, version, success, payload });
```
gregnazario, line 61: I would explain what these are in the message, like `Submitted transaction: `

ecosystem/typescript/sdk/examples/typescript-esm/offer_capabilities.ts
gregnazario, line 58: Note some of this is duplicated and static, so you could get these and then use them multiple places below, make it a little more readable.
```
const aliceAddress = TxnBuilderTypes.AccountAddress.fromHex(alice.address())
const bobAddress = TxnBuilderTypes.AccountAddress.fromHex(bob.address())
```

ecosystem/typescript/sdk/examples/typescript-esm/offer_capabilities.ts
```diff
@@ -0,0 +1,134 @@
...
+  // Note that the proof bytes must be constructed in this specific order- the order of the struct data on-chain.
+  const proofBytes = new Uint8Array([
+    ...BCS.bcsToBytes(struct.accountAddress),
+    ...BCS.bcsSerializeStr(struct.moduleName),
+    ...BCS.bcsSerializeStr(struct.structName),
+    ...(struct.chainId ? BCS.bcsSerializeU8(struct.chainId) : []),
+    ...BCS.bcsSerializeUint64(struct.sequenceNumber),
+    ...BCS.bcsToBytes(struct.sourceAddress),
+    ...BCS.bcsToBytes(struct.recipientAddress),
+  ]);
```
gregnazario, line 112: This is super clean

ecosystem/typescript/sdk/examples/typescript-esm/offer_capabilities.ts
```diff
@@ -0,0 +1,134 @@
...
+
+  // This is the actual signature of the struct.
+  const signedMessage = signer.signBuffer(proofBytes).toUint8Array();
+
+  // Note the hard-coded account scheme. You may need to accomodate MultiEd25519.
+  const payload = new TxnBuilderTypes.TransactionPayloadEntryFunction(
+    TxnBuilderTypes.EntryFunction.natural(
```
gregnazario, line 119: Never to this day will I understand what natural means in this context.

gregnazario (commented):
Hey, looks great, just some comments about how to cut down on some of the duplicate logic / make it a little more readable

ecosystem/typescript/sdk/examples/typescript-esm/offer_capabilities.ts
xbtmatt: 🤣

ecosystem/typescript/sdk/examples/typescript-esm/offer_capabilities.ts
xbtmatt: I love the spread operator, one of the best things about js/ts

ecosystem/typescript/sdk/examples/typescript-esm/offer_capabilities.ts
```diff
@@ -0,0 +1,153 @@
...
+// Works for both SignerCapabilityOffer and RotationCapabilityOffer
+// Except the structName fields are different and the chainId isn't used in SignerCapabilityOffer
+type CapabilityOfferProofChallengeV2 = {
+  accountAddress: TxnBuilderTypes.AccountAddress;
+  moduleName: string;
+  structName: string;
+  sequenceNumber: number;
+  sourceAddress: TxnBuilderTypes.AccountAddress;
+  recipientAddress: TxnBuilderTypes.AccountAddress;
+  chainId?: number; // not used in SignerCapabilityOffer
```
davidiw, line 16: From a BCS approach, it would be better to Make RotationCapabilityOffer contain a SignerCapabilityOffer.

I would further suggest though that we actually inline the first three fields, that shouldn't be part of the input

ecosystem/typescript/sdk/examples/typescript-esm/offer_capabilities.ts
```diff
@@ -0,0 +1,153 @@
...
+  const bobAccountAddress = TxnBuilderTypes.AccountAddress.fromHex(bob.address());
+  const moduleAddress = TxnBuilderTypes.AccountAddress.fromHex(CORE_CODE_ADDRESS);
+
+  // Offer Alice's rotation capability to Bob
+  {
+    // Construct the RotationCapabilityOfferProofChallengeV2 struct
+    const rotationCapProof: CapabilityOfferProofChallengeV2 = {
+      accountAddress: moduleAddress,
+      moduleName: "account",
+      structName: "RotationCapabilityOfferProofChallengeV2",
```
davidiw, line 53: this can lead to accidents.

ecosystem/typescript/sdk/examples/typescript-esm/offer_capabilities.ts
```diff
@@ -0,0 +1,153 @@
...
+  struct: CapabilityOfferProofChallengeV2,
+): Promise<any> => {
+  // The proof bytes are just the individual BCS serialized
+  // data concatenated into a single byte array.
+  // Note that the proof bytes must be constructed in this specific order: the order of the struct data on-chain.
+  const proofBytes = new Uint8Array([
+    ...BCS.bcsToBytes(struct.accountAddress),
+    ...BCS.bcsSerializeStr(struct.moduleName),
+    ...BCS.bcsSerializeStr(struct.structName),
+    ...(struct.chainId ? BCS.bcsSerializeU8(struct.chainId) : []),
```
davidiw, line 127: this is tricky, because the ideal way to write this code is that this is a serialize function on the struct

davidiw (commented):
I know this is largely a copy of the code in the TS SDK, but I'd love for us to be a bit more systematic in defining types and how to serialize them. Also this feels like something that should be entirely within the SDK or making its way into there and we just use the examples as an intermediary ... maybe that's a way I can also experiment with graphql in Python :thinking:

ecosystem/typescript/sdk/examples/typescript-esm/offer_capabilities.ts
xbtmatt: Yeah I realized there's actually already a struct for the rotation offer in the sdk, and then after seeing that + working with Kent on it I realized it's pretty janky to do it this way rather than the struct being a class that implements serialize

I'll change it to that before submitting

xbtmatt:
> I know this is largely a copy of the code in the TS SDK, but I'd love for us to be a bit more systematic in defining types and how to serialize them. Also this feels like something that should be entirely within the SDK or making its way into there and we just use the examples as an intermediary ... maybe that's a way I can also experiment with graphql in Python 🤔

I agree btw, I talked to mayaan about this a few times, I think we should show how to do it in an example but handle it all in the SDK (in a way the wallet adapter can use, too)

I just changed the example to reflect this, I think it makes a lot more sense this way and we can easily put it into the SDK

davidiw (approved):
nice

PR #9427, opened 2023-08-02
gas-schedule numbers updates.md

cryptomolot:
taken numbers from today transaction.rs and storage_gas.md

### DIFF
--- a/developer-docs-site/docs/concepts/base-gas.md
+++ b/developer-docs-site/docs/concepts/base-gas.md
@@ -252,8 +252,8 @@ As of the time of this writing, `min_price_per_gas_unit` in [`transaction.rs`](h
 | Constant                  | Value  |
 |---------------------------|--------|
 | `min_price_per_gas_unit`  | 100    |
-| `max_price_per_gas_unit`  | 10,000 |
-| `gas_unit_scaling_factor` | 10,000 |
+| `max_price_per_gas_unit`  | 10,000,000,000 |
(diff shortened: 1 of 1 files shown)


saharct:
@msmouse @vgao1996 can you help update the storage gas section to take into account [AIP-38 - Deprecate storage gas curves](https://github.com/aptos-foundation/AIPs/blob/main/aips/aip-38.md)?

developer-docs-site/docs/concepts/base-gas.md
```diff
@@ -276,7 +276,7 @@ Hence, in terms of octas, initial mainnet gas costs can be estimated as follows
 | Operation       | Operation | Minimum octas |
 |-----------------|-----------|---------------|
 | Per-item read   | $r_i$     | 3000          |
-| Per-item create | $c_i$     | 50,000        |
+| Per-item create | $c_i$     | 3000          |
```
davidiw, line 279:
```suggestion
| Per-item create | $c_i$     | 50,000          |
```

davidiw (approved):
looks good but the per-item create wasn't modified, still at 50K https://github.com/aptos-labs/aptos-core/blob/main/aptos-move/aptos-gas-schedule/src/gas_schedule/transaction.rs#L145

davidiw:
Thanks for these changes, it would be great if you could make a single PR that encompasses all the changes to a single doc. I'll land these ones as is.

PR #9428, opened 2023-08-02
Add functions that allow using vector like a set

banool:
### Description
These new functions make it possible to easily use `vector` as a set, removing the need for something like a SimpleSet (which just wraps a vector).

`push_unique` is just contains + push_back in a single function, which is what people using vector with unique elements have to do today themselves.

`swap_remove_value` provides a more efficient, unstable method for removing an item from a vector, which is fine if you're using the vector as a traditional set with no ordering of values.

I'm not sure if I should change the code in third_party too, it seems out of date compared to this version of vector.

### Test Plan
From `aptos-move/framework/move-stdlib`
```
$ aptos move test | grep -E 'test_push_unique|test_find_remove'
[ PASS    ] 0x1::vector_tests::test_find_remove
[ PASS    ] 0x1::vector_tests::test_find_remove_by
[ PASS    ] 0x1::vector_tests::test_push_unique
```

### DIFF
--- a/aptos-move/framework/move-stdlib/sources/vector.move
+++ b/aptos-move/framework/move-stdlib/sources/vector.move
@@ -211,6 +211,18 @@ module std::vector {
         pragma intrinsic = true;
     }
 
+    /// Push a new element to the back of the vector if it is not already in the vector.
+    /// This is O(n) in the worst case and preserves ordering of elements in the vector.
+    /// Return `true` if `key` did not already exist in the vector and `false` otherwise.
+    public fun push_unique<Element: drop>(v: &mut vector<Element>, e: Element): bool {
+        if (contains(v, &e)) {
+            false
+        } else {
+            push_back(v, e);
+            true
+        }
+    }
+
     /// Remove the `i`th element of the vector `v`, shifting all subsequent elements.
     /// This is O(n) and preserves ordering of elements in the vector.
     /// Aborts if `i` is out of bounds.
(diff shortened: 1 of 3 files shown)


aptos-move/framework/move-stdlib/sources/vector.move
```diff
@@ -211,6 +211,21 @@ module std::vector {
...
+    /// Return `true` if `key` did not already exist in the vector and `false` otherwise.
+    public fun push_unique<Element: drop>(v: &mut vector<Element>, e: Element): bool {
+        if (contains(v, &e)) {
+            false
+        } else {
+            push_back(v, e);
+            true
+        }
+    }
+    spec push_unique {
```
wrwg, line 225: No need to use this. This is only for functions which have builtin treatment in the prover, this is just a regular Move function. Same below.

aptos-move/framework/move-stdlib/sources/vector.move
```diff
@@ -260,6 +275,24 @@ module std::vector {
         pragma intrinsic = true;
     }
 
+    /// Remove the first occurrence of a given value in the vector `v` and return it in
+    /// a vector, swapping the last element in the vector with the removed item.
+    /// This is O(n) + O(1) and does NOT preserve ordering of elements in the vector.
+    /// This returns an empty vector if the value isn't present in the vector.
+    /// Note that this cannot return an option as option uses vector and there'd be a
+    /// circular dependency between option and vector.
+    public fun swap_remove_value<Element>(v: &mut vector<Element>, val: &Element): vector<Element> {
```
wrwg, line 284: I wonder whether this should be generalized to take a predicate, and whether we really need the 'swap' in the name. I would perhaps just have
```
inline fun find_remove<E>(v: &mut vector<E>, p: |Element|bool) -> vector<E>
```

wrwg (approved):
Great! I would have thought we may also want intersection and union, but we can introduce as we see need for it.

wrwg:
> Great! I would have thought we may also want intersection and union, but we can introduce as we see need for it.

Also, another set of functions which might be important are relations (==, <=).

aptos-move/framework/move-stdlib/sources/vector.move
```diff
@@ -211,6 +211,21 @@ module std::vector {
         pragma intrinsic = true;
     }
 
+    /// Push a new element to the back of the vector if it is not already in the vector.
+    /// This is O(n) in the worst case and preserves ordering of elements in the vector.
+    /// Return `true` if `key` did not already exist in the vector and `false` otherwise.
+    public fun push_unique<Element: drop>(v: &mut vector<Element>, e: Element): bool {
```
davidiw, line 217: My concern with this is that if used improperly, you could end up in a case where you have an inconsistent array.

I'd much rather suggest someone use a well-typed variant of this that doesn't allow for non-unique pushes.

aptos-move/framework/move-stdlib/sources/vector.move
davidiw: Agreed on naming.

aptos-move/framework/move-stdlib/sources/vector.move
banool: What do you mean by inconsistent? For example duplicate items by virtue of using `push_unique` mixed with `push` / `insert` / etc?

If by a well typed variant you mean a separate module, then I have previously implemented and proposed this and it has been widely knocked back: https://aptos-org.slack.com/archives/C036X27DZNG/p1690572419217229.

banool:
I just made the following changes:
- Removed the `spec` blocks.
- Renamed `swap_remove_value` to `find_remove`

Later I will try to do the following, pending resolving [this issue](https://aptos-org.slack.com/archives/C036X27DZNG/p1691973990843039):
- Add a version of `find_remove` that takes a predicate call `find_remove_by`. This naming matches the naming used by functions like `dedup_by` in Rust: https://doc.rust-lang.org/std/vec/struct.Vec.html#method.dedup_by. As well as what I propose in https://www.notion.so/aptoslabs/Additional-vector-methods-845d5cbf9d5f41ba957f68a1db685095.

banool:
Okay, I got `find_remove_by` and the related tests working.

banool:
Still waiting for review.

PR #9430, opened 2023-08-02
Refer to create rather than create_from_account for token v2

banool:
### Description
The docs refer to create_from_account but this function is deprecated, so I make them refer to create instead.

Additionally, create_from_account is marked as deprecated but nothing says what to use instead. I add a comment explaining what to do.

As you can see my editor also obliterates trailing spaces, hopefully that's alright. We should probably make the linter do that anyway.

### Test Plan
👀

### DIFF
--- a/aptos-move/framework/aptos-token-objects/sources/token.move
+++ b/aptos-move/framework/aptos-token-objects/sources/token.move
@@ -141,6 +141,8 @@ module aptos_token_objects::token {
     }
 
     #[deprecated]
+    /// DEPRECATED: Use `create` instead for identical behavior.
+    ///
     /// Creates a new token object from an account GUID and returns the ConstructorRef for
     /// additional specialization.
     public fun create_from_account(
--- a/developer-docs-site/docs/standards/digital-asset.md
+++ b/developer-docs-site/docs/standards/digital-asset.md
@@ -7,9 +7,9 @@ import useBaseUrl from '@docusaurus/useBaseUrl';
 # Aptos Digital Asset Standard
 
 ## Overview of NFTs
-An [NFT](https://en.wikipedia.org/wiki/Non-fungible_token) is a non-fungible [token](https://github.com/aptos-labs/aptos-core/blob/main/aptos-move/framework/aptos-token-objects/sources/token.move) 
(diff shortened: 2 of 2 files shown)


movekevin (commented):
Lets hold off on this a bit as this is not enabled on testnet and mainnet yet

xbtmatt:
I brought this up with Kevin already but yeah we may need to think more about how the new `create_object` is going to be inserted into contracts bc 1. it makes objects even harder to track, especially in terms of when it's invoked by a contract that's built around using GUIDs, but also importantly 2. because it will break a few tests in some of the token examples that assume a GUID is being used to create an object.

davidiw:
> 2. because it will break a few tests in some of the token examples that assume a GUID is being used to create an object.

@xbtmatt can you enumerate those? I updated the Python example to use events, it wasn't too bad.

xbtmatt:
> > 2. because it will break a few tests in some of the token examples that assume a GUID is being used to create an object.
> 
> @xbtmatt can you enumerate those? I updated the Python example to use events, it wasn't too bad.

I thought there were more but maybe not :p

These are the only two I could find, in @gregnazario's test_utils module for the marketplace contract

https://github.com/aptos-labs/aptos-core/blob/59f629396cd466d6e358fd144959c3453b959233/aptos-move/move-examples/marketplace/sources/test_utils.move#L90

https://github.com/aptos-labs/aptos-core/blob/59f629396cd466d6e358fd144959c3453b959233/aptos-move/move-examples/marketplace/sources/test_utils.move#L110

davidiw (approved):
now is the time!

aptos-move/framework/aptos-token-objects/sources/token.move
```diff
@@ -141,6 +141,8 @@ module aptos_token_objects::token {
     }
 
     #[deprecated]
+    /// DEPRECATED: Use `create` instead for identical behavior.
+    ///
```
davidiw, line 145: gah, why you do this, I'm gonna remove it so I can land this...
```suggestion
```

davidiw:
Gah there are other conflicts...

davidiw:
@banool please update and we'll land...

movekevin (approved):

PR #9433, opened 2023-08-02
[TS SDK] Return txn data on success and txn failure reason on failure

0xmaayan:
### Description
The sdk provides a way to return the txn data when it succeed, added an example/test on how to use it.

For when a txn has failed, we throw an error with the txn hash and the failure reason message - Error class is a bit weird as it only outputs the "message" prop, unless we wrap it with try/catch and then we can get access to the other class properties, it feels like a big change for this use case and that should be good for now.

### DIFF
--- a/ecosystem/typescript/sdk/src/providers/aptos_client.ts
+++ b/ecosystem/typescript/sdk/src/providers/aptos_client.ts
@@ -759,7 +759,7 @@ export class AptosClient {
     }
     if (!(lastTxn as any)?.success) {
       throw new FailedTransactionError(
-        `Transaction ${txnHash} committed to the blockchain but execution failed`,
+        `Transaction ${txnHash} failed with an error: ${(lastTxn as any).vm_status}`,
         lastTxn,
       );
     }
--- a/ecosystem/typescript/sdk/CHANGELOG.md
+++ b/ecosystem/typescript/sdk/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to the Aptos Node SDK will be captured in this file. This ch
 ## Unreleased
 
 - Add support for a fee payer transaction
+- Return transaction message error when transaction has failed when `checkSuccess` is set to true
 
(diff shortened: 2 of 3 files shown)


ecosystem/typescript/sdk/src/providers/aptos_client.ts
```diff
@@ -672,7 +672,7 @@ export class AptosClient {
     }
     if (!(lastTxn as any)?.success) {
       throw new FailedTransactionError(
-        `Transaction ${txnHash} committed to the blockchain but execution failed`,
+        `Transaction ${txnHash} failed with an error: ${(lastTxn as any).vm_status}`,
```
xbtmatt, line 675: Yay :D Is there any way we can put the `vm_status` in a field in the `FailedTransactionError` class? I actually don't know if this makes sense, but it would make it easier to use instead of parsing it. Is it generally not a good idea to add fields to existing classes if the sdk exposes them with export, or would it not really matter

0xmaayan: we have the full txn object part of the `FailedTransactionError` class. but as mentioned in the PR comment, Error class is a bit weird as it only outputs the "message" prop, unless we wrap it with try/catch and then we can get access to the other class properties, it feels like a big change for this use case and that should be good for now.

We can always just return the full transaction object, but then we lost the "message" part, i.e you won't see `Transaction ${txnHash} failed with an error: ${(lastTxn as any).vm_status}` but only a txn object and then you would need to look for the `success` prop to see if it succeed or failed - basically same response for a success txn

xbtmatt: oh duh, I forgot the txn object is in the error! Yeah this is great, I think it makes most sense this way then

xbtmatt (approved):

gregnazario (approved):

PR #9442, opened 2023-08-02
Native functions for aggregator snapshots feature

vusirikala:
### Description

Add `snapshot`, `read_snapshot` native functions for aggregator snapshots feature.


aptos-move/framework/src/natives/aggregator_natives/aggregator_snapshot.rs
```diff
@@ -0,0 +1,127 @@
...
+
+fn native_snapshot_with_u64_limit(
+    context: &mut SafeNativeContext,
+    _ty_args: Vec<Type>,
+    mut args: VecDeque<Value>,
+) -> SafeNativeResult<SmallVec<[Value; 1]>> {
+    debug_assert_eq!(args.len(), 1);
+
+    context.charge(AGGREGATOR_V2_SNAPSHOT_WITH_U64_LIMIT_BASE)?;
+    let (aggregator_id, _limit) = aggregator_info(&safely_pop_arg!(args, StructRef))?;
```
igor-aptos, line 60: you should check that limit is <= u64::MAX, and if it is not , abort

aptos-move/aptos-aggregator/src/aggregator_extension.rs
```diff
@@ -64,6 +64,33 @@ impl AggregatorID {
...
+    // aggregator overflows.
+    // TODO: Currently this is a single u128 value since we use 0 as a trivial
+    // lower bound. If we want to support custom lower bounds, or have more
+    // complex postconditions, we should factor this out in its own struct.
+    limit: u128,
+    // Describes values seen by this aggregator. Note that if aggregator knows
+    // its value, then storing history doesn't make sense.
+    history: Option<History>,
+    // The AggregatorID of the aggregator from which the snapshot is taken.
+    base_aggregator: AggregatorID,
```
igor-aptos, line 91: originally you were thinking of adding base_aggregator to every Aggregator, and having AggregatorSnapshot only have base_aggregator. 
and then snapshot creates first a cloned aggregator, and then snapshot that points to it. 

just wondering what reasoning you had for adding all these in the AggregatorSnapshot, do you think it will be simpler?

aptos-move/aptos-aggregator/src/aggregator_extension.rs
```diff
@@ -445,6 +474,30 @@ impl AggregatorData {
...
+            });
+        snapshot_id
+    }
+
+    pub fn read_snapshot(&self, id: AggregatorSnapshotID) -> u128 {
+        let snapshot = self
+            .aggregator_snapshots
+            .get(&id)
+            .expect("AggregatorSnapshot doesn't exist");
+        snapshot.value
```
igor-aptos, line 498: don't you need here to actually do an equivalent of aggregator read, that jumps over to base_aggregator to continue ?

igor-aptos (changes requested):
looks good , a few small changes:

aptos-move/aptos-aggregator/src/aggregator_extension.rs
georgemitenkov: Linking with base aggregator is also easier for maintainability? Given that we plan to change quite a few things

aptos-move/aptos-aggregator/src/aggregator_extension.rs
```diff
@@ -64,6 +64,33 @@ impl AggregatorID {
     }
 }
 
+/// Uniquely identifies each aggregator snapshot instance during the block execution.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord)]
+pub struct AggregatorSnapshotID {
```
georgemitenkov, line 69: Should we have a single ID wrapper for both snapshots and aggregators? What do you think?

aptos-move/aptos-aggregator/src/aggregator_extension.rs
```diff
@@ -445,6 +474,30 @@ impl AggregatorData {
         }
     }
 
+    pub fn snapshot(&mut self, id: &AggregatorID) -> AggregatorSnapshotID {
+        let snapshot_id = AggregatorSnapshotID {
+            id: self.generate_id(),
+        };
+        let aggregator = self.aggregators.get(id).expect("Aggregator doesn't exist");
```
georgemitenkov, line 481: This shouldn't throw but maybe returning result is cleaner

aptos-move/aptos-aggregator/src/aggregator_extension.rs
```diff
@@ -445,6 +474,30 @@ impl AggregatorData {
...
+                base_aggregator: *id,
+            });
+        snapshot_id
+    }
+
+    pub fn read_snapshot(&self, id: AggregatorSnapshotID) -> u128 {
+        let snapshot = self
+            .aggregator_snapshots
+            .get(&id)
+            .expect("AggregatorSnapshot doesn't exist");
```
georgemitenkov, line 497: ditto

aptos-move/framework/src/natives/aggregator_natives/aggregator_snapshot.rs
```diff
@@ -0,0 +1,127 @@
...
+    let aggregator_context = context.extensions().get::<NativeAggregatorContext>();
+
+    match ty_args[0] {
+        Type::U64 => {
+            let snapshot_id = aggregator_snapshot_u64_info(&safely_pop_arg!(args, StructRef))?;
+            let aggregator_data = aggregator_context.aggregator_data.borrow_mut();
+            let value = aggregator_data.read_snapshot(AggregatorSnapshotID { id: snapshot_id });
+            assert!(
+                value <= u64::MAX as u128,
+                "Snapshot value can't exceed u64::MAX"
```
georgemitenkov, line 94: But it can? Should we just do a check comparing to the aggregator limit?

aptos-move/framework/src/natives/aggregator_natives/aggregator_snapshot.rs
```diff
@@ -0,0 +1,127 @@
...
+            assert!(
+                snapshot_id <= u64::MAX as u128,
+                "Snapshot ID can't exceed u64::MAX"
+            );
+            let value = aggregator_data.read_snapshot(AggregatorSnapshotID {
+                id: snapshot_id as u64,
+            });
+            Ok(smallvec![Value::u128(value)])
+        },
+        _ => unreachable!("Snapshot can only be u64 or u128"),
```
georgemitenkov, line 110: Let's return an error here just in case? snapshot can be constructed only from an aggregator, but seems cleaner than panicking

aptos-move/framework/src/natives/aggregator_natives/aggregator_snapshot.rs
```diff
@@ -0,0 +1,127 @@
...
+                "Snapshot value can't exceed u64::MAX"
+            );
+            Ok(smallvec![Value::u64(value as u64)])
+        },
+        Type::U128 => {
+            let snapshot_id = aggregator_snapshot_u128_info(&safely_pop_arg!(args, StructRef))?;
+            let aggregator_data = aggregator_context.aggregator_data.borrow_mut();
+            assert!(
+                snapshot_id <= u64::MAX as u128,
+                "Snapshot ID can't exceed u64::MAX"
```
georgemitenkov, line 103: Why not? if I have an aggregator, snapshot (and not snapshot_u64), and read, I can have something > u64::MAX?

aptos-move/framework/src/natives/aggregator_natives/aggregator_snapshot.rs
vusirikala: So, we can't crate a snapshot_u64 if the limit is above u64::MAX?

aptos-move/framework/src/natives/aggregator_natives/aggregator_snapshot.rs
vusirikala: How can the snapshot's value exceed u64::MAX when reading AggregatorSnapshot<u64>. If the value actually does exceed u64::MAX, it's an overflow and we need to abort right?

aptos-move/framework/src/natives/aggregator_natives/aggregator_snapshot.rs
vusirikala: It's the snapshot's `id` generating using a u64 counter. Why would it exceed u64::MAX?

aptos-move/aptos-aggregator/src/aggregator_extension.rs
vusirikala: > Linking with base aggregator is also easier for maintainability? 

@georgemitenkov Do you mind expanding on this more? What do you mean by linking with base aggregator?

aptos-move/framework/src/natives/aggregator_natives/aggregator_snapshot.rs
georgemitenkov: Indeed, if I have a limit of u128::max, but my aggregator stores 10, why cannot I create a snapshot_u64...
Unless we want to be restrictive here, as allowing the above means we only check overflow on reading / materializing the snapshot.

aptos-move/framework/src/natives/aggregator_natives/aggregator_snapshot.rs
georgemitenkov: I meant we can have the following:
1. create an aggregator with limit > u64::MAX
2. snapshot it as snapshot it (note that the types on Move side are u64, but the value is stored in extension as u128)
3. read_snapshot_u64, which fails if value is greater than  > u64::MAX.

I think Igor suggested a check in (2) to not allow this, it is a possible solution. Otherwise we have to check here?

aptos-move/framework/src/natives/aggregator_natives/aggregator_snapshot.rs
georgemitenkov: Oh, that's ID, all good :)

aptos-move/aptos-aggregator/src/aggregator_extension.rs
georgemitenkov: I meant the same thing as Igor, having a `base_aggregator`

vusirikala: > just wondering what reasoning you had for adding all these in the AggregatorSnapshot, do you think it will be simpler?

Earlier we had the notion of "derived aggregators". The goal at the time was to not introduce a new primitive and merge the notions of aggregator and derived aggregator. Over time, we separated out the notion of aggregators and aggregator snapshots quite a lot. Storing the aggregator snapshot information in rust code in an aggregator struct now seems more confusing to me.

aptos-move/aptos-aggregator/src/aggregator_extension.rs
vusirikala: Done.

aptos-move/aptos-aggregator/src/aggregator_extension.rs
vusirikala: Given that we have separate data structures for aggregators and aggregator snapshots, I felt it is a cleaner to separate out the identifiers for them as well. It felt a little awkward to have `BTreeMap<AggregatorID, Aggregator>` and `BTreeMap<AggregatorID, AggregatorSnapshot>`.

aptos-move/framework/src/natives/aggregator_natives/aggregator_snapshot.rs
vusirikala: Gotcha. In storage, the snapshot<u64> can only store values less than u64::MAX. If the aggregator's value exceeds u64::MAX when the snapshot was created, the read_snapshot has to abort as in the current code right?
If we instead want the abort to happen at the time of creating snapshot<u64>, then we will have to compute the aggregator's value when creating the snapshot and check if the value is below u64::MAX. I believe we want to avoid that?

aptos-move/framework/src/natives/aggregator_natives/aggregator_snapshot.rs
vusirikala: Isn't using "unreachable" more cleaner? It semantically says that this case just doesn't occur.

aptos-move/aptos-aggregator/src/aggregator_extension.rs
georgemitenkov: Indeed :) But when we collect aggregators and snapshots in a change set their deltas will have the same key type though, but we can do a conversion to some other type, no strong preference

aptos-move/framework/src/natives/aggregator_natives/aggregator_snapshot.rs
georgemitenkov: Or alternatively not allow snapshots u64 from aggregators with max value above u64::MAX

aptos-move/framework/src/natives/aggregator_natives/aggregator_snapshot.rs
georgemitenkov: I don't trust Move😂 I think the current guideline for native functions is to be safe and always use result even if something is unreachable, etc. unless it is 100% should not happen, see more here: https://www.notion.so/aptoslabs/Draft-Native-Function-Guidelines-55509921df77485e99274f48880a7ab6?pvs=4

aptos-move/framework/src/natives/aggregator_natives/aggregator_snapshot.rs
igor-aptos: My suggestion is that we don't want code to sometimes be able to cast to u64, sometimes not,  and then need to handle that too. 
And second reason is that since limit is fixed, casting success or abort is never wrong in speculation, and so casting never triggers re-execution

Basically I think this should work as if Aggregator was a templated on type, and there is no casting - u64 aggregator creates u64 snapshot, u128 aggregator creates u128 snapshot. 

I think having Aggregator be a template is more complicated,  but if it isn't,  maybe we should do that?

aptos-move/aptos-aggregator/src/aggregator_extension.rs
igor-aptos: Sounds good, thanks for the explanation!

aptos-move/aptos-aggregator/src/aggregator_extension.rs
igor-aptos: Since they are necessary to be unique against each other - i.e. AggregatorId cannot be the same as AggregatorSnapshotId, I would use the same class. If you don't like the name maybe EphemeralId for both is better?

But you can also think that it can be based on the module, and both are in 
 aggregator module

igor-aptos: If we use EphemeralId elsewhere, for s different class, EphemeralContextId is fine too

aptos-move/framework/src/natives/aggregator_natives/aggregator_snapshot.rs
igor-aptos: I was suggesting (and comment in Move code refers to it), that we can cast to u64 only if max_value (i.e.limit) can be cast to u64

aptos-move/framework/src/natives/aggregator_natives/aggregator_snapshot.rs
igor-aptos: Given that aggregator can never be replaced in backward compatible way having only u128 version of aggregator seems simpler


But what I meant to say,  above suggestion makes it that casting to u64 should never fail,  if it wasn't a coding error - i.e. if you need to cast to u64 snapshot,  you should assert limit is within u64 bounds before you create an aggregator, and you should be good to go.

PR #9445, opened 2023-08-02
[TS SDK] Support fee payer transaction

0xmaayan:
### Description
Adds support for fee payer transaction with
- a function to generate a fee payer raw transaction which returns a FeePayerRawTransaction
- a function to sign a fee payer raw transaction which returns the signer authenticator
- a function to submit the signed fee payer raw transaction

### Test Plan
tests are passing

### DIFF
--- a/ecosystem/typescript/sdk/src/providers/aptos_client.ts
+++ b/ecosystem/typescript/sdk/src/providers/aptos_client.ts
@@ -23,6 +23,7 @@ import {
   TransactionBuilderRemoteABI,
   RemoteABIBuilderConfig,
   TransactionBuilderMultiEd25519,
+  TransactionBuilder,
 } from "../transaction_builder";
 import {
   bcsSerializeBytes,
@@ -35,7 +36,14 @@ import {
   Uint64,
   AnyNumber,
 } from "../bcs";
-import { Ed25519PublicKey, MultiEd25519PublicKey, RawTransaction } from "../aptos_types";
+import {
+  AccountAddress,
+  Ed25519PublicKey,
+  FeePayerRawTransaction,
+  MultiAgentRawTransaction,
+  MultiEd25519PublicKey,
+  RawTransaction,
+} from "../aptos_types";
 import { get, post, ClientConfig, AptosApiError } from "../client";
 
 export interface OptionalTransactionArgs {
@@ -306,6 +314,91 @@ export class AptosClient {
 
     const builder = new TransactionBuilderRemoteABI(this, config);
     return builder.build(payload.function, payload.type_arguments, payload.arguments);
+  }
+
+  /**
+   * Generates a fee payer transaction that can be signed and submitted to chain
+   *
+   * @param sender the sender's account address
+   * @param payload the transaction payload
+   * @param fee_payer the fee payer account
+   * @param secondary_signer_accounts an optional array of the secondary signers accounts
+   * @returns a fee payer raw transaction that can be signed and submitted to chain
+   */
+  async generateFeePayerTransaction(
+    sender: MaybeHexString,
+    payload: Gen.EntryFunctionPayload,
+    fee_payer: AptosAccount,
+    secondary_signer_accounts: Array<AptosAccount> = [],
+    options?: Partial<Gen.SubmitTransactionRequest>,
+  ): Promise<TxnBuilderTypes.FeePayerRawTransaction> {
+    const rawTxn = await this.generateTransaction(sender, payload, options);
+
+    const signers: Array<AccountAddress> = secondary_signer_accounts.map((account) =>
+      AccountAddress.fromHex(account.address()),
+    );
+
+    const feePayerTxn = new TxnBuilderTypes.FeePayerRawTransaction(
+      rawTxn,
+      signers,
+      AccountAddress.fromHex(fee_payer.address()),
+    );
+    return feePayerTxn;
+  }
+
+  /**
+   * Submits fee payer transaction to chain
+   *
+   * @param feePayerTransaction the raw transaction to be submitted, of type FeePayerRawTransaction
+   * @param senderAuthenticator the sender account authenticator (can get from signMultiTransaction() method)
+   * @param feePayerAuthenticator the feepayer account authenticator (can get from signMultiTransaction() method)
+   * @param signersAuthenticators an optional array of the signer account authenticators
+   * @returns The pending transaction
+   */
+  async submitFeePayerTransaction(
+    feePayerTransaction: TxnBuilderTypes.FeePayerRawTransaction,
+    senderAuthenticator: TxnBuilderTypes.AccountAuthenticatorEd25519,
+    feePayerAuthenticator: TxnBuilderTypes.AccountAuthenticatorEd25519,
+    signersAuthenticators: Array<TxnBuilderTypes.AccountAuthenticatorEd25519> = [],
+  ): Promise<Gen.PendingTransaction> {
+    const txAuthenticatorFeePayer = new TxnBuilderTypes.TransactionAuthenticatorFeePayer(
+      senderAuthenticator,
+      feePayerTransaction.secondary_signer_addresses,
+      signersAuthenticators,
+      { address: feePayerTransaction.fee_payer_address, authenticator: feePayerAuthenticator },
+    );
+
+    const bcsTxn = bcsToBytes(
+      new TxnBuilderTypes.SignedTransaction(feePayerTransaction.raw_txn, txAuthenticatorFeePayer),
+    );
+    const transactionRes = await this.submitSignedBCSTransaction(bcsTxn);
+
+    return transactionRes;
+  }
+
+  /**
+   * Signs a multi transaction type (multi agent / fee payer) and returns the
+   * signer authenticator to be used to submit the transaction.
+   *
+   * @param signer the account to sign on the transaction
+   * @param rawTxn a MultiAgentRawTransaction or FeePayerRawTransaction
+   * @returns signer authenticator
+   */
+  // eslint-disable-next-line class-methods-use-this
+  async signMultiTransaction(
+    signer: AptosAccount,
+    rawTxn: MultiAgentRawTransaction | FeePayerRawTransaction,
+  ): Promise<TxnBuilderTypes.AccountAuthenticatorEd25519> {
+    const signerSignature = new TxnBuilderTypes.Ed25519Signature(
+      signer.signBuffer(TransactionBuilder.getSigningMessage(rawTxn)).toUint8Array(),
+    );
+
+    const signerAuthenticator = new TxnBuilderTypes.AccountAuthenticatorEd25519(
+      new TxnBuilderTypes.Ed25519PublicKey(signer.signingKey.publicKey),
+      signerSignature,
+    );
+
+    return Promise.resolve(signerAuthenticator);
   }
 
   /** Converts a transaction request produced by `generateTransaction` into a properly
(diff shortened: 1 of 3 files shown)


xbtmatt:
Woohoo fee payer! Looks good to me, just a small nit w wording

ecosystem/typescript/sdk/src/providers/aptos_client.ts
```diff
@@ -308,6 +316,91 @@ export class AptosClient {
     return builder.build(payload.function, payload.type_arguments, payload.arguments);
   }
 
+  /**
+   * Generates a fee payer transaction that can be signed and submitted to chain
+   *
+   * @param sender sender account address
```
xbtmatt, line 322:
```suggestion
   * @param sender the sender's account address
```

ecosystem/typescript/sdk/src/providers/aptos_client.ts
```diff
@@ -308,6 +316,91 @@ export class AptosClient {
...
+      signers,
+      AccountAddress.fromHex(fee_payer.address()),
+    );
+    return feePayerTxn;
+  }
+
+  /**
+   * Submits fee payer transaction to chain
+   *
+   * @param feePayerTransaction the fee payer transaction, of type FeePayerRawTransaction
```
xbtmatt, line 352:
```suggestion
   * @param feePayerTransaction the raw transaction to be submitted, of type FeePayerRawTransaction
```

only to make it a little more clear. Up to you though, not sure if it's clearer this way or the other way

xbtmatt (approved):

junkil-park (approved):
Great! I am wondering when this feature will be released?

ecosystem/typescript/sdk/src/providers/aptos_client.ts
```diff
@@ -308,6 +316,91 @@ export class AptosClient {
...
+      signer.signBuffer(TransactionBuilder.getSigningMessage(rawTxn)).toUint8Array(),
+    );
+
+    const signerAuthenticator = new TxnBuilderTypes.AccountAuthenticatorEd25519(
+      new TxnBuilderTypes.Ed25519PublicKey(signer.signingKey.publicKey),
+      signerSignature,
+    );
+
+    return Promise.resolve(signerAuthenticator);
+  }
```
gregnazario, line 402: We could just make this `signTransaction` for any transaction right?  It should be the same either way?

ecosystem/typescript/sdk/src/providers/aptos_client.ts
```diff
@@ -308,6 +316,91 @@ export class AptosClient {
...
+   * @param senderAuthenticator the sender account authenticator (can get from signMultiTransaction() method)
+   * @param feePayerAuthenticator the feepayer account authenticator (can get from signMultiTransaction() method)
+   * @param signersAuthenticators an optional array of the signer account authenticators
+   * @returns The pending transaction
+   */
+  async submitFeePayerTransaction(
+    feePayerTransaction: TxnBuilderTypes.FeePayerRawTransaction,
+    senderAuthenticator: TxnBuilderTypes.AccountAuthenticatorEd25519,
+    feePayerAuthenticator: TxnBuilderTypes.AccountAuthenticatorEd25519,
+    signersAuthenticators: Array<TxnBuilderTypes.AccountAuthenticatorEd25519> = [],
```
gregnazario, line 362: Let's drop `Authenticator` from the input names.

And we should name these:
1. sender
2. feePayer
3. additionalSigners

Additional because the sender and the fee payer are technically also signers

ecosystem/typescript/sdk/src/providers/aptos_client.ts
```diff
@@ -308,6 +316,91 @@ export class AptosClient {
...
+   * Submits fee payer transaction to chain
+   *
+   * @param feePayerTransaction the raw transaction to be submitted, of type FeePayerRawTransaction
+   * @param senderAuthenticator the sender account authenticator (can get from signMultiTransaction() method)
+   * @param feePayerAuthenticator the feepayer account authenticator (can get from signMultiTransaction() method)
+   * @param signersAuthenticators an optional array of the signer account authenticators
+   * @returns The pending transaction
+   */
+  async submitFeePayerTransaction(
+    feePayerTransaction: TxnBuilderTypes.FeePayerRawTransaction,
```
gregnazario, line 359: probably can just rename to just transaction or raw transaction, since the input is for a FeePayer

ecosystem/typescript/sdk/src/providers/aptos_client.ts
```diff
@@ -308,6 +316,91 @@ export class AptosClient {
...
+   * @param secondary_signer_accounts an optional array of the secondary signers accounts
+   * @returns a fee payer raw transaction that can be signed and submitted to chain
+   */
+  async generateFeePayerTransaction(
+    sender: MaybeHexString,
+    payload: Gen.EntryFunctionPayload,
+    fee_payer: AptosAccount,
+    secondary_signer_accounts: Array<AptosAccount> = [],
+    options?: Partial<Gen.SubmitTransactionRequest>,
+  ): Promise<TxnBuilderTypes.FeePayerRawTransaction> {
```
gregnazario, line 334: nit: We're mixing `snake_case` with `camelCase`, we should stick with one?

ecosystem/typescript/sdk/src/providers/aptos_client.ts
```diff
@@ -308,6 +316,91 @@ export class AptosClient {
...
+   * @param payload the transaction payload
+   * @param fee_payer the fee payer account
+   * @param secondary_signer_accounts an optional array of the secondary signers accounts
+   * @returns a fee payer raw transaction that can be signed and submitted to chain
+   */
+  async generateFeePayerTransaction(
+    sender: MaybeHexString,
+    payload: Gen.EntryFunctionPayload,
+    fee_payer: AptosAccount,
+    secondary_signer_accounts: Array<AptosAccount> = [],
```
gregnazario, line 332: These accounts don't require private keys right?  Just the address, maybe the public key?

gregnazario (commented):
My comments are mostly around consistency, just wondering if we need private keys for `AptosAccount`

gregnazario:
> Great! I am wondering when this feature will be released?

Soon!

ecosystem/typescript/sdk/src/providers/aptos_client.ts
0xmaayan: yep. amazed that lint didn't complain on it

ecosystem/typescript/sdk/src/providers/aptos_client.ts
0xmaayan: it might be confusing to understand **what** transaction to pass, rawTxn or feePayerTransaction, tho there is the input type, but naming it as `feePayerTransaction` can be more meaningful - but dont have strong opinion on that

ecosystem/typescript/sdk/src/providers/aptos_client.ts
0xmaayan: I find adding `Authenticator` makes it clearer that we are expecting an `Authenticator` and not just an `account`

ecosystem/typescript/sdk/src/providers/aptos_client.ts
0xmaayan: We already have a `signTransaction` that signs with `Ed25519Signature` and returns a`Uint8Array` type, for feepayer/multiagent transaction we need to return the signer authenticator so users would submit the transaction with it (unless we want to introduce another function in the flow that return the authenticator, which i dont think it is necessary) . because of different return type issues, we might be better with 2 separate functions

ecosystem/typescript/sdk/src/providers/aptos_client.ts
0xmaayan: yea... I think we should standardize the account input type through out the sdk. some functions ask for the account, some for the account address and some for a hex string. we should probably have one input type and a way to derive whatever the function needs within the sdk.

xbtmatt: In this case the e2e flow will often not include the signers so I think we should definitely use addresses, not accounts, I meant to start a discussion about this yesterday but forgot to hit send I guess

ecosystem/typescript/sdk/src/providers/aptos_client.ts
gregnazario: Okay let's keep it then for now

ecosystem/typescript/sdk/src/providers/aptos_client.ts
gregnazario: I see... then let's keep this how it is.  We can do better when we rework the whole thing in V2.

ecosystem/typescript/sdk/src/providers/aptos_client.ts
gregnazario: okay, let's still rename the last one to additionalSignerAuthenticators

ecosystem/typescript/sdk/src/providers/aptos_client.ts
gregnazario: Then can we take in an address then instead, or does it need more than the address?

0xmaayan: oh wait my bad, misreading, FeePayer expects an Array of `AccountAddress` and not `AptosAccount` 🤦‍♀️
 ill change the input to accept AptosAccount.

xbtmatt: I'm fairly sure the actual raw txn does not need anything other than the secondary addresses and the fee payer address. Obviously the order of the addresses matter, because this is the order of the `&signer`s in a multi-signer function call.

But yes, this transaction will need to be constructed in multiple contexts where the signers are signing independently of one another, so it only makes sense to use addresses here

ecosystem/typescript/sdk/src/providers/aptos_client.ts
```diff
@@ -308,6 +316,87 @@ export class AptosClient {
...
+   * @param sender the sender's account address
+   * @param payload the transaction payload
+   * @param fee_payer the fee payer account
+   * @param secondarySignerAccounts an optional array of the secondary signers accounts
+   * @returns a fee payer raw transaction that can be signed and submitted to chain
+   */
+  async generateFeePayerTransaction(
+    sender: MaybeHexString,
+    payload: Gen.EntryFunctionPayload,
+    feePayer: AptosAccount,
```
gregnazario, line 331: One last thing, both of these should be just the address.

Then each individual user should sign with their AptosAccount

ecosystem/typescript/sdk/src/providers/aptos_client.ts
```diff
@@ -308,6 +316,83 @@ export class AptosClient {
...
+   * @param sender the sender's account address
+   * @param payload the transaction payload
+   * @param fee_payer the fee payer account
+   * @param secondarySignerAccounts an optional array of the secondary signers accounts
+   * @returns a fee payer raw transaction that can be signed and submitted to chain
+   */
+  async generateFeePayerTransaction(
+    sender: MaybeHexString,
+    payload: Gen.EntryFunctionPayload,
+    feePayer: AccountAddress,
```
gregnazario, line 331: We should probably make both MaybeHexString, I swear AccountAddress vs MaybeHexString is confusing.

gregnazario (approved):

ecosystem/typescript/sdk/src/providers/aptos_client.ts
0xmaayan: haha ok I think we got everything as it should be, one last look?

PR #9448, opened 2023-08-03
[TS SDK] Initialize sdk_v2 folder structure and basic usage

0xmaayan:
### Description
This PR is the first step for the [sdk_v2 development process](https://www.notion.so/aptoslabs/TS-SDK-V2-b6ecea4ec1744e0d8f728bb9bedaabf5?pvs=4#66ebc37862d048fe886535817da75ea4). It includes only the basic structure, client files, [Aptos](https://github.com/aptos-labs/aptos-core/pull/9448/files#diff-cef902eb6acbe9baa210f333e8bd5be26a3e69a48536b7ded08ae76a85135db5R5) class in `aptos.ts`  and [AptosConfig](https://github.com/aptos-labs/aptos-core/pull/9448/files) in `aptos_config.ts`
```
src/
 api
  aptos.ts // main class to initialize
  aptosConfig.ts // global configuration type
 bcs // bcs related operations
 client // core client files
 crypto // holds all crypto related functions (ED25519, multi ED25519, etc) 
 internal // holds internal operations like api calls
 types // sdk types 
 utils // util functions
 index.ts // main file to export all modules
```


ecosystem/typescript/sdk_v2/src/internal/account.ts
```diff
@@ -0,0 +1,106 @@
...
+
+export async function getResources(
+  aptosConfig: AptosConfig,
+  accountAddress: MaybeHexString,
+  query?: { ledgerVersion?: AnyNumber },
+): Promise<Gen.MoveResource[]> {
+  const out = await paginateWithCursor<{}, Gen.MoveResource[]>({
+    url: aptosConfig.network,
+    endpoint: `accounts/${accountAddress}/resources`,
+    params: { ledger_version: query?.ledgerVersion, limit: 9999 },
```
gregnazario, line 85: Hmm, why is limit set to 9999?

ecosystem/typescript/sdk_v2/src/internal/account.ts
```diff
@@ -0,0 +1,106 @@
...
+
+export async function getModules(
+  aptosConfig: AptosConfig,
+  accountAddress: MaybeHexString,
+  query?: { ledgerVersion?: AnyNumber },
+): Promise<Gen.MoveModuleBytecode[]> {
+  const response = await paginateWithCursor<{}, Gen.MoveModuleBytecode[]>({
+    url: aptosConfig.network,
+    endpoint: `accounts/${HexString.ensure(accountAddress).hex()}/modules`,
+    params: { ledger_version: query?.ledgerVersion, limit: 1000 },
```
gregnazario, line 26: Similarly why is limit set to 1000?  These should be provided by the user with a default that makes sense right?

We may want to expose the pagination the whole way up now that we can rewrite it

gregnazario (approved):
I now realize this is mostly just copy pasta, so I'm going to stop commenting on it, but keep in mind anything that's been copied over, and had a bug that was fixed in it, will now diverge.

May just be better to write it from scratch with empty files, and just have all the license, and setup files etc.

ecosystem/typescript/sdk_v2/src/client/types.ts
```diff
@@ -0,0 +1,95 @@
...
+    this.url = response.url;
+    this.status = response.status;
+    this.statusText = response.statusText;
+    this.data = response.data;
+    this.request = request;
+  }
+}
+
+export interface PaginationArgs {
+  start?: bigint | number;
```
gregnazario, line 93: This is good for now, but let's be consistent this round I saw there was the AnyNumber type that represented this sometimes in the old SDK

So maybe we make that type now?

ecosystem/typescript/sdk_v2/src/version.ts
```diff
@@ -0,0 +1,2 @@
+// hardcoded for now, we would want to have it injected dynamically
+export const VERSION = "2.0.0";
```
gregnazario, line 2: I wonder are alpha versions supported in npm?

gregnazario (approved):

0xjinn (approved):
Is the plan to create a new sdk_v2 folder, then slowly migrating everything over to sdk_v2, and eventually deprecating/replacing the existing one?

0xmaayan:
> Is the plan to create a new sdk_v2 folder, then slowly migrating everything over to sdk_v2, and eventually deprecating/replacing the existing one?

the plan is to create a sdk_v2 that holds sdk v2 files. some would be ported over from v1 and new ones. eventually we would have 2 folders, sdk and sdk_v2 and just publish sdk_v2 under a new name. we keep the sdk folder b/c we need to keep maintaining it for some time

PR #9451, opened 2023-08-03
[NFT Metadata Crawler] DB migrations in code

just-in-chang:
### Description
Run DB migrations in code instead of as a command


larry-aptos (approved):

bowenyang007 (approved):
nice!

PR #9452, opened 2023-08-03
[dag] Integrate Order Rule with dag driver

ibalajiarun:
### Description

This PR integrates the Order rule with the DAG driver. Whenever the dag driver adds a certified node to the DAG, it calls the order rule to try and commit nodes that have enough voting power.

### Test Plan

Existing tests.

### DIFF
--- a/consensus/src/dag/dag_driver.rs
+++ b/consensus/src/dag/dag_driver.rs
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use super::{
+    order_rule::OrderRule,
     storage::DAGStorage,
     types::{CertifiedAck, DAGMessage},
     RpcHandler,
@@ -43,6 +44,7 @@ pub(crate) struct DagDriver {
     time_service: TimeService,
     rb_abort_handle: Option<AbortHandle>,
     storage: Arc<dyn DAGStorage>,
+    order_rule: OrderRule,
 }
 
 impl DagDriver {
@@ -55,6 +57,7 @@ impl DagDriver {
         current_round: Round,
         time_service: TimeService,
         storage: Arc<dyn DAGStorage>,
+        order_rule: OrderRule,
     ) -> Self {
         // TODO: rebroadcast nodes after recovery
         Self {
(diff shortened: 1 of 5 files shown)


ibalajiarun:
* main
  * **PR #9452** <a href="https://app.graphite.dev/github/pr/aptos-labs/aptos-core/9452" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/aptos-labs/aptos-core/9452?utm_source=stack-comment).

sasha8 (approved):
LGTM

consensus/src/dag/dag_driver.rs
```diff
@@ -149,11 +153,15 @@ impl RpcHandler for DagDriver {
...
             if dag_reader.exists(node.metadata()) {
-                return Ok(CertifiedAck::new(node.metadata().epoch()));
+                return Ok(CertifiedAck::new(epoch));
             }
         }
 
-        self.add_node(node)?;
+        let node_metadata = node.metadata().clone();
+        self.add_node(node).and_then(|_| {
+            self.order_rule.process_new_node(&node_metadata);
```
sasha8, line 162: It is not enough to call order_rule here. We should call it every time we enter a node to the dag. For example, certified node response in dag_fetcher. Otherwise, we might lose liveness.

sasha8 (commented):
There might be a liveness issue.

consensus/src/dag/dag_driver.rs
```diff
@@ -149,11 +153,15 @@ impl RpcHandler for DagDriver {
...
+            Ok(())
+        })?;
```
sasha8, line 164: It is not enough to call order_rule here. We should call it every time we enter a node to the dag. For example, certified node response in dag_fetcher. Otherwise, we might lose liveness.

consensus/src/dag/dag_driver.rs
sasha8, line 162: It is not enough to call order_rule here. We should call it every time we enter a node to the dag. For example, certified node response in dag_fetcher. Otherwise, we might lose liveness.

zekun000 (approved):
discussed offline, we'll make the order rule able to scan through all the nodes, so we only need to call it once after all dependencies added

sasha8 (approved):

PR #9454, opened 2023-08-03
[dag] rename reliable_broadcast.rs to rb_handler.rs

ibalajiarun:
### Description

Since reliable broadcast sender is moved to its own crate, we now only handle the receiver side of the broadcast in DAG. Hence, renaming the module appropriately

### Test Plan

Existing tests

PR #9453, opened 2023-08-03
[dag] Integrate Dag Fetcher with Dag Driver

ibalajiarun:
### Description

This PR integrates the dag fetching logic with the dag driver.

PR #9455, opened 2023-08-03
[dag] bootstrap logic

ibalajiarun:


ibalajiarun:
* main
        * **PR #9455** <a href="https://app.graphite.dev/github/pr/aptos-labs/aptos-core/9455" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/aptos-labs/aptos-core/9455?utm_source=stack-comment).

consensus/src/dag/bootstrap.rs
```diff
@@ -0,0 +1,117 @@
...
+    let anchor_election = Box::new(RoundRobinAnchorElection::new(validators));
+    let order_rule = OrderRule::new(
+        epoch_state.clone(),
+        latest_ledger_info,
+        dag.clone(),
+        anchor_election,
+        ordered_nodes_tx,
+    );
+
+    let (dag_fetcher, fetch_requester) = DagFetcher::new(
```
sasha8, line 73: I think dag_fetcher should also get order_rule, so it can call it every time a certificate is added to the dag (see my comments on the first PR).

Or actually, maybe it should be only part of Dag and called from add_node therein.

ibalajiarun: I like the idea actually, but given how Zekun implemented oneshot receiver, I felt like that should be handled by the handler task and not the fetcher, because order rule can be a heavy operation imo.

consensus/src/dag/bootstrap.rs
```diff
@@ -0,0 +1,119 @@
...
+        dag_rpc_rx,
+        rb_handler,
+        dag_driver,
+        fetch_handler,
+        node_fetch_waiter,
+        certified_node_fetch_waiter,
+    );
+
+    let (nh_abort_handle, nh_abort_registration) = AbortHandle::new_pair();
+    let (df_abort_handle, df_abort_registration) = AbortHandle::new_pair();
```
zekun000, line 108: are those used for shutdown?

zekun000 (approved):

sasha8 (approved):

consensus/src/dag/bootstrap.rs
ibalajiarun: yes, that was my idea.

zekun000: typically we don't need those since dropping the sender would result in the task stops if it's something like this
```
  while let Some(x) = rx.next().await {}
```

ibalajiarun: I am planning to introduce an interval timer for garbage collection. Let me keep it for now, and we can clean up later if it becomes unnecessary.

PR #9461, opened 2023-08-03
[Spec] Ensures of transaction_validation

UIZorrot:
### Description

We have finished the remained post-condition in transaction_validation.move.
As far as we know, high-level properties 1,2,3 is remained unsolved, which the document requires a manual audit

### Test Plan

aptos move prove -f transcation_validation


rahxephon89 (approved):

junkil-park (approved):
LGTM

PR #9462, opened 2023-08-03
[Spec] Ensures for voting

UIZorrot:
### Description

We have finished the remained post-condition in voting.move.
As far as we know, high-level properties is all complete

### Test Plan

aptos move prove -f voting


rahxephon89 (changes requested):
@UIZorrot I think `resolve_proposal_v2` will cause inconsistency issue. Could you use the flag  `--unconditional-abort-as-inconsistency` and `--check-inconsistency` to check whether there is an issue?

UIZorrot:
The inconsistency is coming from event.move, the latest updated fix the problem

rahxephon89 (changes requested):

aptos-move/framework/aptos-framework/sources/voting.spec.move
```diff
@@ -57,15 +55,11 @@ spec aptos_framework::voting {
         is_multi_step_proposal: bool,
     ): u64 {
         use aptos_framework::chain_status;
-        pragma verify_duration_estimate = 120; // TODO: set because of timeout (property proved)
 
         requires chain_status::is_operating();
         include CreateProposalAbortsIf<ProposalType>;
+        // property 1: Verify the proposal_id of the newly created proposal.
         ensures result == old(global<VotingForum<ProposalType>>(voting_forum_address)).next_proposal_id;
-        ensures global<VotingForum<ProposalType>>(voting_forum_address).next_proposal_id
```
junkil-park, line 65: Why are these removed?

rahxephon89: They are moved to the schema below

rahxephon89 (changes requested):

aptos-move/framework/aptos-framework/sources/voting.spec.move
```diff
@@ -176,25 +211,70 @@ spec aptos_framework::voting {
...
+
+        let multi_step_key = std::string::spec_utf8(IS_MULTI_STEP_PROPOSAL_KEY);
+        aborts_if simple_map::spec_contains_key(proposal.metadata, multi_step_key) &&
+            !from_bcs::deserializable<bool>(simple_map::spec_get(proposal.metadata, multi_step_key));
+        let is_multi_step = simple_map::spec_contains_key(proposal.metadata, multi_step_key) &&
+            from_bcs::deserialize(simple_map::spec_get(proposal.metadata, multi_step_key));
+        aborts_if !is_multi_step && len(next_execution_hash) != 0;
+
+        aborts_if len(next_execution_hash) == 0 && !exists<timestamp::CurrentTimeMicroseconds>(@aptos_framework);
+        aborts_if len(next_execution_hash) == 0 && is_multi_step && simple_map::spec_contains_key(proposal.metadata, multi_step_in_execution_key);
```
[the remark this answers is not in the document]
UIZorrot: oh yes, i think you are right

rahxephon89 (approved):

junkil-park (approved):

PR #2042, opened 2022-07-18
testing: identify flaky tests

geekflyer:

PR #9463, opened 2023-08-03
[devtools] (Re)Introducing target determinator cargo cli

ibalajiarun:
### Description

This PR (re)introduces the target-determinator based cargo command wrapper from the diem project, albeit in a simplified and very opinionated way. The implementation details diverge from the original implementation and is tailored to the everyday use-case at Aptos. It uses the `determinator` [crate](https://docs.rs/determinator/latest/determinator/index.html), which was originally used in diem repo, to identify potentially affected packages between local changes and remote. 

As v0, the features are very limited:
- Only identifies changed packages with respect to local origin/main ref. It won't work for other branch such as release branches yet.
- Does not yet allow excluding certain packages such as smoke-tests if they are identified as affected.
- It downloads cargo metadata of origin/main ref from a GCS bucket, if not present locally. This may affect offline development.

There are two binaries:
- aptos-cargo-cli is the default CLI that can be installed and used via `aptos-cargo-cli`.
- cargo-x is the cargo subcommand, which will allow invoking the CLI via `cargo x ...`

The command supports the following subcommands: `check`, `fmt`, `nextest`, `test`, `xclippy`

### Test Plan

To install: `cargo install --path devtools/aptos-cargo-cli --bin cargo-x`

Manual testing


devtools/aptos-cargo-cli/src/bin/cargo-x.rs
```diff
@@ -0,0 +1,36 @@
+// Copyright © Aptos Foundation
```
JoshLind, line 1: nit: technically we should be adding the full license, but I recognize the linter doesn't currently enforce this 😢 
```
```

JoshLind (approved):
Thanks @ibalajiarun! This looks reasonable to me. Any chance you could link the old PR/code version you're pulling this functionality from? 😄

devtools/aptos-cargo-cli/src/bin/cargo-x.rs
ibalajiarun: Good point @JoshLind, we should discuss this, because the implementation details are totally different. I only borrow the idea of calling cargo commands with the appropriate --package flags and implement it very differently. I use the public determinator and guppy crates in crates.io to find the affected packages to pass to the --package flag, so that attribution is implicit. Given this, I wonder if we need to change the license. wdyt?

The original code was removed here: https://github.com/aptos-labs/aptos-core/pull/2042

perryjrandall (approved):

perryjrandall:
thank u boss man

devtools/aptos-cargo-cli/src/bin/cargo-x.rs
JoshLind: Hmm.. I'm so-so. If there's no copy/pasta and things are sufficiently different, I don't think we need to worry too much 😄 But, if you feel like there's some overlap, it probably makes sense to use the shared license?

ibalajiarun: I added the shared license to couple of files that look like there is an overlap, just to be safe.

Issue #7906, opened 2023-04-24
[CLI] Ledger support


alnoki:
Per recent discussion with @gregnazario @davidiw @CapCap on Move Monday.

Currently Petra supports the Ledger hardware wallet integration, but the Aptos CLI does not.

The lack of support on the CLI imposes burdensome development costs on builders: if they want to publish, sign a transaction etc. from a hardware wallet, they will need to develop some kind of bespoke interface that feeds in a transaction request to Petra.

Adding ledger support to the CLI will improve the pace of development, without discouraging secure practices (as a result of burdensome inconveniences).

See also https://docs.solana.com/wallet-guide/hardware-wallets/ledger#use-ledger-nano-with-solana-cli

dylanschultzie:
+1. Adding Ledger support to CLI will allow us at Lavender.Five to rotate our operator key to our hardware wallet. Currently we still have a soft key to manage pulling commissions, then immediately move the commission to our hardware wallet.

ericjohncarlson:
+1 to this as well.  Hugely valuable to validator operations.

gregnazario:
Yes, we plan to add Ledger support to the CLI, I as well would love to use it instead of a hot wallet :)

alnoki:
@0xjinn

banool:
I just put discussing this item on the agenda for our team sync, stay posted!

alnoki:
> I just put discussing this item on the agenda for our team sync, stay posted!

@banool thanks for the update! One detail worth mentioning is the ability to sign publication transactions up to the 64k transaction limit, since I believe this may involve chunking into the ledger internal RAM

alnoki:
Bumping this issue per stale label

Ledger support is a key feature for secure development

0xjinn:
Hi @alnoki , the ledger integration is in a feature branch right now - https://github.com/aptos-labs/aptos-core/blob/feature_branch/cli_ledger_integration, and here is the doc - [https://github.com/aptos-labs/aptos-core/blob/feature_branch/cli_ledger_integratio[…]oper-docs-site/docs/tools/aptos-cli/use-cli/use-aptos-ledger.md](https://github.com/aptos-labs/aptos-core/blob/feature_branch/cli_ledger_integration/developer-docs-site/docs/tools/aptos-cli/use-cli/use-aptos-ledger.md)

Please feel free to test it out in devnet/testnet. We will be releasing this soon.

dylanschultzie:
[...]

PR #9465, opened 2023-08-03
[CLI][Ledger] update derivation path format doc

0xjinn:
### Description

Looks like the format has causes some confusion. See reference - https://github.com/aptos-labs/aptos-core/issues/7906#issuecomment-1664226876

This PR updates the format comment

### Test Plan

```
➜  aptos-core git:(feature_branch/cli_ledger_integration) ✗ aptos-debug init --ledger --profile ledger --derivation-path m/44\'/637\'/0\'/0\'/0\'

Configuring for profile ledger

Using command line argument for derivation path m/44'/637'/0'/0'/0'
Account 59836ba1dd0c845713bdab34346688d6f1dba290dbf677929f2fc20593ba0cfb has been already found onchain

---
Aptos CLI is now set up for account 59836ba1dd0c845713bdab34346688d6f1dba290dbf677929f2fc20593ba0cfb as profile ledger!  Run `aptos --help` for more information about commands
{
  "Result": "Success"
}

```

### DIFF
--- a/crates/aptos/src/common/types.rs
+++ b/crates/aptos/src/common/types.rs
@@ -742,8 +742,8 @@ pub trait ParsePrivateKey {
 pub struct HardwareWalletOptions {
     /// Derivation Path of your account in hardware wallet
     ///
-    /// e.g format - [m/44'/637'/0'/0'/0]
-    /// Make sure your wallet is unblocked and have Aptos opened
+    /// e.g format - m/44\'/637\'/0\'/0\'/0\'
+    /// Make sure your wallet is unlocked and have Aptos opened
     #[clap(long)]
     pub derivation_path: Option<String>,
 


crates/aptos/src/common/types.rs
```diff
@@ -742,8 +742,8 @@ pub trait ParsePrivateKey {
 pub struct HardwareWalletOptions {
     /// Derivation Path of your account in hardware wallet
     ///
-    /// e.g format - [m/44'/637'/0'/0'/0]
-    /// Make sure your wallet is unblocked and have Aptos opened
+    /// e.g format - m/44\'/637\'/0\'/0\'/0\'
+    /// Make sure your wallet is unlocked and have Aptos opened
```
gregnazario, line 746: Technically, it is the original one, but then you have to pass it in through the CLI, which gets messy.

gregnazario (approved):

crates/aptos/src/common/types.rs
0xjinn: Yeah. I think people will forget to escape the `'`.  Hopefully this make it clear when using the CLI

PR #9466, opened 2023-08-03
fix SerializeKey/DeserializeKey macros to work for generic structs

alinush:
### Description

Much like https://github.com/aptos-labs/aptos-core/pull/9117.

Currently cannot get this to work. `DeserializeKey` is trickier to get right: see the `serialize_key.rs` test currently failing below.

Also, a bit confused about whether the `serde` name of the `Value` struct, in the `deserialize_key` procedural macro, needs to be changed to account for the type params also.

```
alinush@macbook [~/repos/aptos-core/crates/aptos-crypto] (alin/fix-serializekey-macros) $ cargo test
   Compiling aptos-crypto v0.0.3 (/Users/alinush/repos/aptos-core/crates/aptos-crypto)
error: expected one of `#`, `>`, `const`, identifier, or lifetime, found `<`
  --> crates/aptos-crypto/src/unit_tests/serialize_key.rs:28:24
   |
28 |     struct TestGenerics<SubType> {
   |                        ^ expected one of `#`, `>`, `const`, identifier, or lifetime

error: proc-macro derive produced unparsable tokens
  --> crates/aptos-crypto/src/unit_tests/serialize_key.rs:27:28
   |
27 |     #[derive(SerializeKey, DeserializeKey)]
   |                            ^^^^^^^^^^^^^^

error[E0277]: the trait bound `for<'de> TestGenerics<SubType>: test_utils::_::_serde::Deserialize<'de>` is not satisfied
  --> crates/aptos-crypto/src/unit_tests/serialize_key.rs:32:64
   |
32 |     impl<SubType: ValidCryptoMaterial> ValidCryptoMaterial for TestGenerics<SubType> {
   |                                                                ^^^^^^^^^^^^^^^^^^^^^ the trait `for<'de> test_utils::_::_serde::Deserialize<'de>` is not implemented for `TestGenerics<SubType>`
   |
   = help: the following other types implement trait `test_utils::_::_serde::Deserialize<'de>`:
             <&'a Path as test_utils::_::_serde::Deserialize<'de>>
             <&'a [u8] as test_utils::_::_serde::Deserialize<'de>>
             <&'a serde_bytes::Bytes as test_utils::_::_serde::Deserialize<'de>>
             <&'a str as test_utils::_::_serde::Deserialize<'de>>
             <() as test_utils::_::_serde::Deserialize<'de>>
             <(T0, T1) as test_utils::_::_serde::Deserialize<'de>>
             <(T0, T1, T2) as test_utils::_::_serde::Deserialize<'de>>
             <(T0, T1, T2, T3) as test_utils::_::_serde::Deserialize<'de>>
           and 189 others
   = note: required for `TestGenerics<SubType>` to implement `DeserializeOwned`
note: required by a bound in `traits::ValidCryptoMaterial`
  --> crates/aptos-crypto/src/traits.rs:63:72
   |
61 | pub trait ValidCryptoMaterial:
   |           ------------------- required by a bound in this trait
62 |     // The for<'a> exactly matches the assumption "deserializable from any lifetime".
63 |     for<'a> TryFrom<&'a [u8], Error=CryptoMaterialError> + Serialize + DeserializeOwned
   |                                                                        ^^^^^^^^^^^^^^^^ required by this bound in `ValidCryptoMaterial`

For more information about this error, try `rustc --explain E0277`.
error: could not compile `aptos-crypto` (lib test) due to 3 previous errors
```

### DIFF
--- a/crates/aptos-crypto-derive/src/lib.rs
+++ b/crates/aptos-crypto-derive/src/lib.rs
@@ -151,11 +151,15 @@ pub fn deserialize_key(source: TokenStream) -> TokenStream {
     let ast: DeriveInput = syn::parse(source).expect("Incorrect macro input");
     let name = &ast.ident;
     let name_string = name.to_string();
-    let (impl_generics, ty_generics, where_clause) = &ast.generics.split_for_impl();
 
+    let impl_params = ast.generics.params.iter();
+    let (_, params, where_clause) = ast.generics.split_for_impl();
+
+    // Note: Ideally, we should generate a lifetime with a unique name so that it does not collide with
+    //       the lifetime parameters of the type itself.
     let gen = quote! {
-        impl<'de, #impl_generics> ::serde::Deserialize<'de> for #name #ty_generics
-        #where_clause
+        impl<'de, #(#impl_params),*> ::serde::Deserialize<'de> for #name #params
+            #where_clause
         {
             fn deserialize<D>(deserializer: D) -> std::result::Result<Self, D::Error>
             where
@@ -174,7 +178,7 @@ pub fn deserialize_key(source: TokenStream) -> TokenStream {
                     struct Value<'a>(&'a [u8]);
 
                     let value = Value::deserialize(deserializer)?;
-                    #name #ty_generics::try_from(value.0).map_err(|s| {
+                    #name::try_from(value.0).map_err(|s| {
                         <D::Error as ::serde::de::Error>::custom(format!("{} with {}", s, #name_string))
                     })
                 }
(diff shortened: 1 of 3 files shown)


crates/aptos-crypto-derive/src/lib.rs
```diff
@@ -151,8 +151,12 @@ pub fn deserialize_key(source: TokenStream) -> TokenStream {
     let ast: DeriveInput = syn::parse(source).expect("Incorrect macro input");
     let name = &ast.ident;
     let name_string = name.to_string();
+    let (impl_generics, ty_generics, where_clause) = &ast.generics.split_for_impl();
+
     let gen = quote! {
-        impl<'de> ::serde::Deserialize<'de> for #name {
+        impl<'de, #impl_generics> ::serde::Deserialize<'de> for #name #ty_generics
```
ibalajiarun, line 157: I think instead of doing this, you need to add `'de` into `impl_generics` above and just use `#impl_generics` here. Not sure how to do that exactly.

davidiw (commented):
Curious ... why? The code you'd generate would be a bit different than the code the existing macro is producing.

Your current example code is missing things like <#type> (<>) and #name::#type (::).
So you likely either need another macro or you can get really fancy and push a conditional -- if there are type arguments do this if there are none do that. Rust isn't really dynamic.
You could look into what you can do within and around quote! to do incremental builds of the function as well.

Might be good to google examples as well....

PR #9470, opened 2023-08-03
[aptos-release-builder] Print out on chain configs

runtian-zhou:
### Description

Added an command to print out on chain configs.

### Test Plan

Tested with following command: 
```
cargo run -- print-configs --endpoint https://fullnode.mainnet.aptoslabs.com/v1
```

And it printed out the values of configs properly

### DIFF
--- a/aptos-move/aptos-release-builder/src/main.rs
+++ b/aptos-move/aptos-release-builder/src/main.rs
@@ -57,6 +57,9 @@ pub enum Commands {
         /// Url endpoint for the desired network. e.g: https://fullnode.mainnet.aptoslabs.com/v1.
         #[clap(short, long)]
         endpoint: url::Url,
+        /// Whether to print out the full gas schedule.
+        #[clap(short, long)]
+        print_gas_schedule: bool,
     },
 }
 
@@ -170,7 +173,10 @@ async fn main() -> Result<()> {
                 .set_fast_resolve(DEFAULT_RESOLUTION_TIME)
                 .await
         },
-        Commands::PrintConfigs { endpoint } => {
+        Commands::PrintConfigs {
+            endpoint,
+            print_gas_schedule,
+        } => {
             use aptos_types::on_chain_config::*;
 
             let client = aptos_rest_client::Client::new(endpoint);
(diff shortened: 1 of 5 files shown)


alinush:
Can you update a README.md with instructions on this command?

runtian-zhou:
I updated the comments for the argument. Instruction could be found if you just use `cargo run -- --help`

movekevin (approved):
This is awesome! Do you have a sample of the output?

aptos-move/aptos-release-builder/src/main.rs
```diff
@@ -48,6 +52,12 @@ pub enum Commands {
         #[clap(long)]
         mint_to_validator: bool,
     },
+    /// Print out current values of on chain configs.
+    PrintConfigs {
+        /// Url endpoint for the desired network. e.g: https://fullnode.mainnet.aptoslabs.com/v1.
+        #[clap(short, long)]
```
perryjrandall, line 58: it would be nice if we could also accept a short name alias like 'testnet' 'mainnet' 'devnet' as well as a fully formed url

perryjrandall (approved):
beautiful <3

runtian-zhou:
Here's a sampel output https://gist.github.com/runtian-zhou/783984144676be824e1ca3bd469e5ea0

perryjrandall:
> Here's a sampel output https://gist.github.com/runtian-zhou/783984144676be824e1ca3bd469e5ea0

Can we summarize or minimize the gas schedule by default? Its usually only necessary if you want to specifically check gas

movekevin:
> > Here's a sampel output https://gist.github.com/runtian-zhou/783984144676be824e1ca3bd469e5ea0
> 
> Can we summarize or minimize the gas schedule by default? Its usually only necessary if you want to specifically check gas

Yeah or better separate it out into its own command to not overwhelm the output

perryjrandall:
> > > Here's a sampel output https://gist.github.com/runtian-zhou/783984144676be824e1ca3bd469e5ea0
> > 
> > 
> > Can we summarize or minimize the gas schedule by default? Its usually only necessary if you want to specifically check gas
> 
> Yeah or better separate it out into its own command to not overwhelm the output

Yeah a subcommand structure could be very nice here

```
aptos-release-builder print [gas|execution|features|consensus]
```

PR #9471, opened 2023-08-03
[NFT Metadata Crawler] Default to no ACK after parsing

just-in-chang:
### Description
Only send ACK to PubSub if release flag is set

### Test Plan
Tested locally; looks good

Fixes EI-284


rtso (approved):
Don't forget to the example config.yaml too in documentation

davidiw (approved):

PR #9474, opened 2023-08-03
Cleanup OnChainExecutionConfig and update to V4

bchocho:
### Description

All versions before registration are now prefixed with `Deprecated` and always return shuffle and gas limit as off. The next governance should update to V4.

### DIFF
--- a/aptos-move/aptos-release-builder/src/components/mod.rs
+++ b/aptos-move/aptos-release-builder/src/components/mod.rs
@@ -10,8 +10,7 @@ use aptos_temppath::TempPath;
 use aptos_types::{
     account_config::CORE_CODE_ADDRESS,
     on_chain_config::{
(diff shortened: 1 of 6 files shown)


igor-aptos (approved):
ok this looks cleaner to me

sitalkedia (commented):
TBH, this looks more confusing to figure out the behavior based on the on-chain config value. For example, the transaction shuffling is set to 1, which is enabled but it will not be very clear that V3 is deprecated and we are overriding the shuffling to no shuffling in V3. 

The previous approach made it much clearer, some one taking a look at the on-chain config value immediately can figure out that it has been deprecated and we are doing no shuffling.

bchocho:
@sitalkedia would it make sense to do these changes (rename of V1-V3 to deprecated) while keeping the DeprecatedSenderAwareV1 as well?

PR #9484, opened 2023-08-04
[Refactor Transaction-generator-lib] [1/n] Add txn pattern generator trait and 2 impls

manudhundi:
Add txn pattern generator trait and 2 implementations 'random txn pattern'
and 'connected txn pattern generator'.

### DIFF
--- a/crates/transaction-generator-lib/src/lib.rs
+++ b/crates/transaction-generator-lib/src/lib.rs
@@ -13,6 +13,11 @@ use aptos_sdk::{
 };
 use args::TransactionTypeArg;
 use async_trait::async_trait;
+use rand::{rngs::StdRng, seq::SliceRandom, Rng};
+#[cfg(test)]
+use rand_core::SeedableRng;
+#[cfg(test)]
+use std::collections::HashSet;
 use std::{
     collections::HashMap,
     sync::{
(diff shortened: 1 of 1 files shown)


crates/transaction-generator-lib/src/lib.rs
```diff
@@ -90,6 +95,231 @@ pub trait TransactionGenerator: Sync + Send {
...
+        1. The class implementing the trait should ensure that the pool is large enough.
+        2. The trait is independent of the contents of the pool. In case the contents of the pool
+           matter, then the class implementing the trait should handle it.
+        3. The trait returns: array of [senders, [multi_sig_signers], [receivers]]
+    */
+    fn generate_tx_pattern(
+        &mut self,
+        rng: &mut StdRng,
+        pool_start: usize,
+        pool_size: usize,
```
igor-aptos, line 109: these seems like arguments for the constructor of the below implementations, not something all implementations care about?

crates/transaction-generator-lib/src/lib.rs
```diff
@@ -90,6 +95,231 @@ pub trait TransactionGenerator: Sync + Send {
...
+        num_txns: usize,
+        num_multi_sig_signers_per_txn: usize,
+        num_receivers_per_txn: usize,
+    ) -> Vec<(usize, Vec<usize>, Vec<usize>)>;
```
igor-aptos, line 113: It might be too early to tell (i.e. the rest of the changes will influence what is needed), but seems like in some cases we will want TxnPatternGenerator to be managing account pool, in which case these need to be types - LocalAccount / AccountAddress, and not usize

igor-aptos (changes requested):
for the TxnPatternGenerator trait and signatures, think of what would be needed for an implementation(s) that represents account_pool.

for AccountCreation , every call to generate_tx_pattern should create new LocalAccounts from rng, and give those IDs out, putting them into the pool, and on the other side every caller that needs a burner account, claims it from that pool. I don't think current signature would work for that well

crates/transaction-generator-lib/src/lib.rs
manudhundi: I was thinking we just generate the pattern and return them as an index to sample from a pool.
Thanks for pointing the shortcomings. 
So I suppose we keep the trait more generic and allow the implementations of the trait to either 'pick from a pool' or 'generate randomly'.
Does this trait signature seem reasonable at this point ?
    fn generate_tx_pattern(
        &mut self,
        rng: &mut StdRng,
        num_txns: usize,
        num_multi_sig_signers_per_txn: usize,
        num_receivers_per_txn: usize,
    ) -> Vec<(LocalAccount, Vec<LocalAccount>, Vec<LocalAccount>)>;
    
    One small downside of this could be that, we will be passing around 'LocalAccount' for cases (P2P generator) where 'AccountAddress' is sufficient.

manudhundi: The api got formatted out missing some content. Below is what I meant actually.

fn generate_tx_pattern(
    &mut self,
    rng: &mut StdRng,
    num_txns: usize,
    num_multi_sig_signers_per_txn: usize,
    num_receivers_per_txn: usize,
) -> Vec<(LocalAccount, Vec < LocalAccount > , Vec < LocalAccount > )>;

sitalkedia: @manudhundi - Is there any generator today that actually uses the `multi-sig-signer` if not, I would simplify the API by removing it. Of course, we can add it back later when we add support for txn type that uses it.

PR #9485, opened 2023-08-04
Update deprecated dependencies

gregnazario:
### Description
This removes the move cli `info` command, that prints out the dependency tree.  We don't really support the move CLI.

Additionally, updates some more dependencies so there's no longer a warning about deprecation in a future version.

### Test Plan
```
cargo build -p aptos
```


davidiw:
Why are we removing the dependency graph generator? I couldn't follow the logic in the commit about being blocked by a future Rust update.

gregnazario:
> Why are we removing the dependency graph generator? I couldn't follow the logic in the commit about being blocked by a future Rust update.

As far as I could tell it was not used anywhere.

We need to have a replacement for ptree because the author hasn't responded to the request for update prior.

We could fork the dependency and create our own branch for it as well.

davidiw (approved):
ptree is not actively maintained... approving

wrwg (changes requested):
Unless there is a pressing reason, I suggest to lets have a discussion first about aptos-cli and move-cli relation. There are multiple options -- ranging from removing the move CLI, or moving some functionality from aptos CLI into move CLI (invisible to the user), so that aptos-move becomes an independent component (in third_party).

wrwg (approved):
See the point with `ptree`, but lets continue to discuss.

PR #9488, opened 2023-08-04
A more efficient implementation of the multi-round block partitioner

zjma:
### Description

Implemented `PartitionerV2` which is ~2.5x faster than the default partitioner.
Also included:
- Partitioner selection via envvars.
- A sharded execution bug fix.
- Giving the last executor shard extra threads to handle merged last round.
- Minor refactoring in aptos-block-partitioner.


execution/executor-benchmark/src/block_partitioning.rs
```diff
@@ -14,20 +14,22 @@ use aptos_types::{
...
 impl BlockPartitioningStage {
-    pub fn new(num_shards: usize) -> Self {
-        let maybe_partitioner = if num_shards <= 1 {
+    pub fn new(num_executor_shards: usize) -> Self {
+        let maybe_partitioner = if num_executor_shards <= 1 {
             None
         } else {
-            let partitioner = ShardedBlockPartitioner::new(num_shards);
-            Some(partitioner)
+            //TODO: build partitioner from configuration.
```
zjma, line 27: cc @sitalkedia, i plan to do it in a separate PR.

execution/block-partitioner/src/v2/mod.rs
```diff
@@ -0,0 +1,719 @@
...
+}
+
+mod conflicting_txn_tracker;
+mod counters;
+
+/// Basically `ShardedBlockPartitioner` but:
+/// - Not pre-partitioned by txn sender.
+/// - implemented more efficiently.
+pub struct PartitionerV2 {
+    thread_pool: ThreadPool,
```
sitalkedia, line 98: checkout my latest PR, we can use the `BlockPartitionerConfiguration` class for both V1 and V2.

execution/block-partitioner/src/v2/mod.rs
```diff
@@ -0,0 +1,719 @@
...
+            drop(new_idxs);
+        });
+        let duration = timer.stop_and_record();
+        info!("drop={duration}");
+        ret
+    }
+}
+
+#[test]
+fn test_partitioner_v2_correctness() {
```
sitalkedia, line 656: Let's move all tests into a test module. Also, can you move all V1 tests to a common module and ensure the relevant tests work with both V1 and V2?

execution/block-partitioner/src/v2/mod.rs
```diff
@@ -0,0 +1,719 @@
...
+}
+
+impl BlockPartitioner for PartitionerV2 {
+    fn partition(
+        &self,
+        txns: Vec<AnalyzedTransaction>,
+        num_executor_shards: usize,
+    ) -> Vec<SubBlocksForShard<AnalyzedTransaction>> {
+        let timer = MISC_TIMERS_SECONDS
+            .with_label_values(&["preprocess"])
```
sitalkedia, line 522: let's rename the counter from `preprocess` to `build-index` or something like that?

execution/block-partitioner/src/v2/mod.rs
```diff
@@ -0,0 +1,719 @@
...
+            DashMap::with_shard_amount(self.dashmap_num_shards);
+        let trackers: DashMap<StorageKeyIdx, RwLock<ConflictingTxnTracker>> =
+            DashMap::with_shard_amount(self.dashmap_num_shards);
+        for txn in txns.iter() {
+            senders.push(RwLock::new(None));
+            wsets.push(RwLock::new(HashSet::with_capacity(txn.write_hints().len())));
+            rsets.push(RwLock::new(HashSet::with_capacity(txn.read_hints().len())));
+        }
+        let timer_1 = MISC_TIMERS_SECONDS
+            .with_label_values(&["preprocess__main"])
```
sitalkedia, line 542: may be build-index-main?

execution/block-partitioner/src/v2/mod.rs
```diff
@@ -0,0 +1,719 @@
...
+                        .or_insert_with(|| num_senders.fetch_add(1, Ordering::SeqCst));
+                    *senders[txn_idx].write().unwrap() = Some(sender_idx);
+                    let num_writes = txn.write_hints().len();
+                    for (i, storage_location) in txn
+                        .write_hints()
+                        .iter()
+                        .chain(txn.read_hints().iter())
+                        .enumerate()
+                    {
+                        let key = storage_location.state_key().clone();
```
sitalkedia, line 561: Why do you need a clone here - this can be expensive? Can we avoid it?

execution/block-partitioner/src/v2/mod.rs
```diff
@@ -0,0 +1,719 @@
...
+                        let key_idx = *key_idx_table
+                            .entry(key)
+                            .or_insert_with(|| num_keys.fetch_add(1, Ordering::SeqCst));
+                        let is_write = i < num_writes;
+                        if is_write {
```
sitalkedia, line 566: This is a bit confusing.. why not iterate through read_set and write_set separately?

execution/block-partitioner/src/v2/mod.rs
```diff
@@ -0,0 +1,719 @@
...
+                                ))
+                            })
+                            .write()
+                            .unwrap()
+                            .add_candidate(txn_idx, is_write);
+                    }
+                });
+        });
+        let duration_1 = timer_1.stop_and_record();
+        info!("preprocess__main={duration_1}");
```
sitalkedia, line 588: Let's remove them if they are temporarily added or convert them to `trace` if we plan to keep them. I would prefer removing them as we already have counters for them?

execution/block-partitioner/src/v2/mod.rs
```diff
@@ -0,0 +1,719 @@
...
+        txns: Vec<AnalyzedTransaction>,
+        num_executor_shards: usize,
+    ) -> Vec<SubBlocksForShard<AnalyzedTransaction>> {
+        let timer = MISC_TIMERS_SECONDS
+            .with_label_values(&["preprocess"])
+            .start_timer();
+        let num_txns = txns.len();
+        let num_senders = AtomicUsize::new(0);
+        let num_keys = AtomicUsize::new(0);
+        let mut senders: Vec<RwLock<Option<SenderIdx>>> = Vec::with_capacity(num_txns);
```
sitalkedia, line 527: Can you move the building index part to a separate function so that it makes this function a bit shorter and easier to follow?

execution/block-partitioner/src/v2/mod.rs
```diff
@@ -0,0 +1,719 @@
...
+            &start_index_matrix,
+            &new_idxs,
+            &trackers,
+        );
+        let duration = timer.stop_and_record();
+        info!("add_edges={duration}");
+        let timer = MISC_TIMERS_SECONDS
+            .with_label_values(&["drop"])
+            .start_timer();
+        self.thread_pool.spawn(move || {
```
sitalkedia, line 637: Let's also move the drops to a separate `async_drop` function?

execution/block-partitioner/src/v2/mod.rs
```diff
@@ -0,0 +1,719 @@
...
+    }
+}
+
+#[test]
+fn test_partitioner_v2_determinism() {
+    let partitioner = Arc::new(PartitionerV2::new(4, 4, 10, 64));
+    assert_deterministic_result(partitioner);
+}
+
+/// Evenly divide 0..n-1. Example: uniform_partition(11,3) == [[0,1,2,3],[4,5,6,7],[8,9,10]]
```
sitalkedia, line 676: The example is actually incorrect, as per your code, we would have chunk sizes of 3,4,4. Am I missing anything?

execution/block-partitioner/src/v2/mod.rs
```diff
@@ -0,0 +1,719 @@
...
+        info!("add_edges__return_obj={duration}");
+
+        self.thread_pool.spawn(move || {
+            drop(sub_block_matrix);
+            drop(txns);
+        });
+        ret
+    }
+
+    fn multi_rounds(
```
sitalkedia, line 402: nit: better name for this function..

execution/block-partitioner/src/v2/mod.rs
```diff
@@ -0,0 +1,719 @@
...
+        start_txns: &[OriginalTxnIdx],
+        mut remaining_txns: Vec<Vec<OriginalTxnIdx>>,
+    ) -> (
+        Vec<Vec<Vec<OriginalTxnIdx>>>,
+        Vec<Vec<OriginalTxnIdx>>,
+        Vec<RwLock<TxnIndex>>,
+    ) {
+        let finalized_indexs: Vec<RwLock<TxnIndex>> =
+            (0..num_txns).map(|_tid| RwLock::new(0)).collect();
+        let mut txn_idx_matrix: Vec<Vec<Vec<OriginalTxnIdx>>> = Vec::new();
```
sitalkedia, line 419: Rename this to accepted_txns?

execution/block-partitioner/src/v2/mod.rs
```diff
@@ -0,0 +1,719 @@
...
+        &self,
+        round_id: RoundId,
+        rsets: &[RwLock<HashSet<StorageKeyIdx>>],
+        wsets: &[RwLock<HashSet<StorageKeyIdx>>],
+        senders: &[RwLock<Option<SenderIdx>>],
+        trackers: &DashMap<StorageKeyIdx, RwLock<ConflictingTxnTracker>>,
+        start_txns: &[OriginalTxnIdx],
+        remaining_txns: Vec<Vec<OriginalTxnIdx>>,
+    ) -> (Vec<Vec<OriginalTxnIdx>>, Vec<Vec<OriginalTxnIdx>>) {
+        let timer = MISC_TIMERS_SECONDS
```
sitalkedia, line 134: Let's change this to a timer with round as a dimension.

execution/block-partitioner/src/v2/mod.rs
```diff
@@ -0,0 +1,719 @@
+// Copyright © Aptos Foundation
+
+#[cfg(test)]
```
sitalkedia, line 3: Still need to go through the logic in the module deeper but a high level, this module seems overloaded with doing too many things, can you think about refactoring/offloading some functionality to separate modules? For example, we have a `conflict_detector`and `dependency_edge_creator` modules in V1, that offloads some of the functionality. 

Also individual functions are pretty overloaded, let's think about breaking them into smaller pieces and add more comments.

execution/block-partitioner/src/v2/conflicting_txn_tracker.rs
```diff
@@ -0,0 +1,165 @@
...
+    /// candidates=[T4(W), T7(W), T8(R), T9(W), T11(W)], start=3, end=5 => true
+    /// candidates=[T4(W), T7(W), T8(R), T9(W), T11(W)], start=3, end=6 => true
+    /// candidates=[T4(W), T7(W), T8(R), T9(W), T11(W)], start=4, end=4 => false
+    /// candidates=[T4(W), T7(W), T8(R), T9(W), T11(W)], start=4, end=6 => true
+    /// candidates=[T4(W), T7(W), T8(R), T9(W), T11(W)], start=5, end=6 => false
+    /// candidates=[T4(W), T7(W), T8(R), T9(W), T11(W)], start=7, end=9 => true
+    /// candidates=[T4(W), T7(W), T8(R), T9(W), T11(W)], start=8, end=9 => false
+    /// candidates=[T4(W), T7(W), T8(R), T9(W), T11(W)], start=12, end=4 => false
+    /// candidates=[T4(W), T7(W), T8(R), T9(W), T11(W)], start=12, end=5 => true
+    pub fn has_write_in_range(
```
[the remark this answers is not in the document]
zjma: it's actually used by every actual round to determine discarding or not.

execution/block-partitioner/src/v2/conflicting_txn_tracker.rs
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    /// Check if there is a txn writing to the current storage location and its txn_id in the given range.
+    /// The txn list is considered a ring, and the range can be wrapped.
+    /// Below are some examples.
+    ///
+    /// candidates=[T4(W), T7(W), T8(R), T9(W), T11(W)], start=3, end=4 => false
+    /// candidates=[T4(W), T7(W), T8(R), T9(W), T11(W)], start=3, end=5 => true
+    /// candidates=[T4(W), T7(W), T8(R), T9(W), T11(W)], start=3, end=6 => true
+    /// candidates=[T4(W), T7(W), T8(R), T9(W), T11(W)], start=4, end=4 => false
```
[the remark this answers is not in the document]
zjma: Sorry for the ambiguity, it should represent an empty interval [4,4) which always contains no writers...

execution/block-partitioner/src/v2/conflicting_txn_tracker.rs
```diff
@@ -0,0 +1,165 @@
...
+    /// candidates=[T4(W), T7(W), T8(R), T9(W), T11(W)], start=4, end=6 => true
+    /// candidates=[T4(W), T7(W), T8(R), T9(W), T11(W)], start=5, end=6 => false
+    /// candidates=[T4(W), T7(W), T8(R), T9(W), T11(W)], start=7, end=9 => true
+    /// candidates=[T4(W), T7(W), T8(R), T9(W), T11(W)], start=8, end=9 => false
```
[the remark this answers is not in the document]
zjma: 9 writes, but the interval is [8,9).
I will make sure to call out the interval details in the comment...

aptos-move/aptos-vm/src/sharded_block_executor/tests.rs
```diff
@@ -49,5 +64,57 @@ fn test_sharded_block_executor_with_random_transfers_sequential() {
...
     let sharded_block_executor = ShardedBlockExecutor::new(client);
-    test_utils::sharded_block_executor_with_random_transfers(sharded_block_executor, 1)
+    let mut partitioner = Box::new(ShardedBlockPartitioner::new(num_shards));
+    partitioner.max_rounds = Some(8);
+    partitioner.xshard_dep_avoid_threshold = Some(0.9);
+    test_utils::sharded_block_executor_with_random_transfers(partitioner, sharded_block_executor, 1)
+}
+
+#[test]
+fn test_partitioner_v2_sharded_block_executor_no_conflict() {
```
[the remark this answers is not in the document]
zjma: why is this duplication?

execution/block-partitioner/src/v2/mod.rs
```diff
@@ -0,0 +1,719 @@
...
+/// Represent a sender in a partitioning session.
+type SenderIdx = usize;
+
+/// Represent a txn after its position is finalized.
+/// Different from `aptos_types::block_executor::partitioner::ShardedTxnIndex`,
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
+pub struct ShardedTxnIndex2 {
+    pub round_id: RoundId,
+    pub shard_id: ShardId,
+    pub ori_txn_idx: OriginalTxnIdx,
```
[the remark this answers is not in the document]
zjma: It has to be original index. that's why I need a `ShardedTxnIndex2`.

execution/block-partitioner/src/v2/mod.rs
```diff
@@ -0,0 +1,719 @@
...
+/// Represent a specific storage location in a partitioning session.
+type StorageKeyIdx = usize;
+
+/// Represent a sender in a partitioning session.
+type SenderIdx = usize;
+
+/// Represent a txn after its position is finalized.
+/// Different from `aptos_types::block_executor::partitioner::ShardedTxnIndex`,
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
+pub struct ShardedTxnIndex2 {
```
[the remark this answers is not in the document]
zjma: https://github.com/aptos-labs/aptos-core/pull/9488#discussion_r1289133683

execution/block-partitioner/src/main.rs
```diff
@@ -1,61 +1,68 @@
...
+use aptos_logger::info;
 use aptos_types::transaction::analyzed_transaction::AnalyzedTransaction;
 use clap::Parser;
 use rand::rngs::OsRng;
 use rayon::iter::{IntoParallelIterator, ParallelIterator};
 use std::{sync::Mutex, time::Instant};
 
+#[cfg(unix)]
+#[global_allocator]
+static ALLOC: jemallocator::Jemalloc = jemallocator::Jemalloc;
```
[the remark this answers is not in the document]
zjma: merging the discussion.
https://github.com/aptos-labs/aptos-core/pull/9488#discussion_r1288943639

execution/block-partitioner/src/v2/mod.rs
zjma: required by being a key of hashmap `key_idx_table`.

execution/block-partitioner/src/v2/mod.rs
zjma: i double checked, the comment is consistent with the test cases.
I'm letting the bigger chunks go first.

execution/executor-benchmark/src/block_partitioning.rs
zjma: in this PR now...

execution/block-partitioner/src/lib.rs
```diff
@@ -3,34 +3,244 @@
...
+                            .entry((src_tid.round_id, src_tid.shard_id, key))
+                            .or_insert_with(|| 0);
+                        *value += 1;
+                    }
+                }
+                for (dst_tid, locs) in td.cross_shard_dependencies.dependent_edges().iter() {
+                    for loc in locs.iter() {
+                        let key = loc.clone().into_state_key();
+                        let key_str = CryptoHash::hash(&key).to_hex();
+                        println!("MATRIX_REPORT - round={}, shard={}, old_tid={}, new_tid={}, send key={} to round={}, shard={}, new_tid={}", round_id, shard_id, old_tid, tid, key_str, dst_tid.round_id, dst_tid.shard_id, dst_tid.txn_index);
```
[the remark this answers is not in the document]
zjma: Leaning to keeping them. The function is now test-only so the prints are usually captured.

execution/block-partitioner/src/lib.rs
```diff
@@ -3,34 +3,244 @@
...
+                let tid = sub_block.start_index + local_tid;
+                for loc in td.txn.write_hints().iter() {
+                    let key = loc.clone().into_state_key();
+                    let key_str = CryptoHash::hash(&key).to_hex();
+                    println!(
+                        "MATRIX_REPORT - round={}, shard={}, old_tid={}, new_tid={}, write_hint={}",
+                        round_id, shard_id, old_tid, tid, key_str
+                    );
+                }
+                for (src_tid, locs) in td.cross_shard_dependencies.required_edges().iter() {
```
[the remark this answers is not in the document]
zjma: Do not yet have a good way to break down - I tried pulling them out and the whole file gets even longer because several objects with long types are passed down across multiple layers...

execution/block-partitioner/src/v2/mod.rs
```diff
@@ -0,0 +1,719 @@
...
+        atomic::{AtomicUsize, Ordering},
+        Mutex, RwLock,
+    },
+};
+
+/// The position of a txn in the block *before partitioning*.
+type OriginalTxnIdx = usize;
+
+/// Represent a specific storage location in a partitioning session.
+type StorageKeyIdx = usize;
```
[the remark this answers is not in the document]
zjma: why? we are just counting here...

execution/block-partitioner/src/main.rs
```diff
@@ -1,61 +1,68 @@
...
 
     #[clap(long, default_value_t = 100000)]
     pub block_size: usize,
 
-    #[clap(long, default_value_t = 10)]
+    #[clap(long, default_value_t = 9)]
     pub num_blocks: usize,
 
-    #[clap(long, default_value_t = 12)]
+    #[clap(long, default_value_t = 60)]
```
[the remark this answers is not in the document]
zjma: 60 is recently more frequently used...

execution/block-partitioner/src/v2/conflicting_txn_tracker.rs
```diff
@@ -0,0 +1,165 @@
...
+/// For txns that claimed to access the same storage location,
+/// it caches some metadata about the location and also keeps track of their status (pending or position finalized) throughout the partitioning process.
+#[derive(Debug, Clone, Eq, PartialEq, Serialize, Deserialize)]
+pub struct ConflictingTxnTracker {
+    pub storage_location: StorageLocation,
+    pub anchor_shard_id: ShardId,
+    pending_reads: BTreeSet<OriginalTxnIdx>,
+    pending_writes: BTreeSet<OriginalTxnIdx>,
+    pub writer_set: HashSet<OriginalTxnIdx>,
+    pub finalized_all: BTreeSet<ShardedTxnIndex2>,
```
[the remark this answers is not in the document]
zjma: the current design is to require a RwLock outside, then the experience is like a DB transaction: moving a txn out of pending set and into finalized set should together be an atomic operation. And all these BTreeSets are just like DB indexes ...

not sure what it's like if separated. Maybe a sub-PR?

execution/block-partitioner/src/v2/mod.rs
zjma: preprocess contains allocation and initialization and parallel index building. The last is more what you referred to as "build-index" and is currently called "main".

execution/block-partitioner/src/v2/conflicting_txn_tracker.rs
```diff
@@ -0,0 +1,159 @@
...
+    /// Txns that write the current storage location.
+    pub writer_set: HashSet<OriginalTxnIdx>,
+    /// Txns that have been accepted.
+    pub finalized_all: BTreeSet<ShardedTxnIndex2>,
+    /// Txns that (1) write the current storage location and (2) have been accepted.
+    pub finalized_writes: BTreeSet<ShardedTxnIndex2>,
+}
+
+impl ConflictingTxnTracker {
+    pub fn new(storage_location: StorageLocation, anchor_shard_id: ShardId) -> Self {
```
sitalkedia, line 35: nit: rather than passing the `anchor_shard_id`, you can just calculate here. That would make the constructor cleaner.

execution/block-partitioner/src/v2/conflicting_txn_tracker.rs
```diff
@@ -0,0 +1,159 @@
...
+    /// The storage location on which conflicting txns are being tracked by this tracker.
+    pub storage_location: StorageLocation,
+    /// A randomly chosen owner shard of the storage location, for conflict resolution purpose.
+    pub anchor_shard_id: ShardId,
+    /// Txns that (1) read the current storage location and (2) have not been accepted.
+    pending_reads: BTreeSet<OriginalTxnIdx>,
+    /// Txns that (1) write the current storage location and (2) have not been accepted.
+    pending_writes: BTreeSet<OriginalTxnIdx>,
+    /// Txns that write the current storage location.
+    pub writer_set: HashSet<OriginalTxnIdx>,
```
sitalkedia, line 27: What's the difference between `pending_writes` and `writer_set` ?

execution/block-partitioner/src/v2/mod.rs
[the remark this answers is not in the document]
sitalkedia: Let's rename this to `ShardedTxnIndexV2` in that case.

execution/block-partitioner/src/v2/mod.rs
```diff
@@ -0,0 +1,750 @@
...
+
+impl PartitionerV2 {
+    pub fn new(
+        num_threads: usize,
+        num_rounds_limit: usize,
+        avoid_pct: u64,
+        dashmap_num_shards: usize,
+        merge_discarded: bool,
+    ) -> Self {
+        info!("Creating a PartitionerV2 instance with num_threads={num_threads}, num_rounds_limit={num_rounds_limit}, avoid_pct={avoid_pct}, dashmap_num_shards={dashmap_num_shards}, merge_discarded={merge_discarded}");
```
sitalkedia, line 116: Better to derive #Debug for this struct to print useful info like this.

execution/block-partitioner/src/v2/mod.rs
```diff
@@ -0,0 +1,750 @@
...
+        num_executor_shards: usize,
+    ) -> PartitionedTransactions {
+        let timer = MISC_TIMERS_SECONDS
+            .with_label_values(&["preprocess"])
+            .start_timer();
+        let num_txns = txns.len();
+        let num_senders = AtomicUsize::new(0);
+        let num_keys = AtomicUsize::new(0);
+        let mut senders: Vec<RwLock<Option<SenderIdx>>> = Vec::with_capacity(num_txns);
+        let mut wsets: Vec<RwLock<HashSet<StorageKeyIdx>>> = Vec::with_capacity(num_txns);
```
sitalkedia, line 556: nit: would prefer naming `write_sets` and `read_sets`

execution/block-partitioner/src/v2/mod.rs
sitalkedia: you have extra _ in the counter name it seems.

execution/block-partitioner/src/v2/mod.rs
```diff
@@ -0,0 +1,750 @@
...
+                            .unwrap();
+                        if is_write {
+                            tracker.add_write_candidate(txn_idx);
+                        } else {
+                            tracker.add_read_candidate(txn_idx);
+                        }
+                    }
+                });
+        });
+        let _duration_1 = timer_1.stop_and_record();
```
sitalkedia, line 620: better naming for timer_1 and `_duration_1`

execution/block-partitioner/src/v2/mod.rs
```diff
@@ -0,0 +1,750 @@
...
+        let _duration = timer.stop_and_record();
+
+        let timer = MISC_TIMERS_SECONDS
+            .with_label_values(&["pre_partition_uniform"])
+            .start_timer();
+        let remaining_txn_idxs = uniform_partition(num_txns, num_executor_shards);
```
sitalkedia, line 626: This can be done as a follow up, but we need to create an abstraction for the partitioner here.. so that we can plug in other types of partitioner as a next evolution.

execution/block-partitioner/src/test_utils.rs
```diff
@@ -79,3 +88,202 @@ pub fn create_signed_p2p_transaction(
...
+}
+
+#[cfg(test)]
+pub fn assert_deterministic_result(partitioner: Arc<dyn BlockPartitioner>) {
+    let mut rng = thread_rng();
+    let block_gen = P2PBlockGenerator::new(1000);
+    for _ in 0..10 {
+        let txns = block_gen.rand_block(&mut rng, 100);
+        let result_0 = partitioner.partition(txns.clone(), 10);
+        for _ in 0..2 {
```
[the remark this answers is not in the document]
zjma: This way it's ~1s on my laptop, >1s would be too expensive for a single test case i feel.

execution/block-partitioner/src/v2/conflicting_txn_tracker.rs
zjma: `pending_writes` tracks all writes that haven't been accepted, and is frequently updated during discarding rounds.

`writer_set` is a constant lookup table needed in `add_edges` to check if a txn writes to the current key or not. 

Example:
```
// before any discarding

writer_set == [0, 7, 9]
pending_writes == [0, 7, 9]
accepted == []

// during discarding round 0, mark_txn_accepted(7, round_id=0, shard_id=1)

writer_set == [0, 7, 9]
pending_writes == [0, 9]
accepted == [T7@(0, 1)]

// during discarding round 2, mark_txn_accepted(9, round_id=2, shard_id=3)

writer_set == [0, 7, 9]
pending_writes == [0]
accepted == [T7@(0,1), T9@(2,3)]

// during discarding round 4, mark_txn_accepted(9, round_id=4, shard_id=0)

writer_set == [0, 7, 9]
pending_writes == []
accepted == [T7@(0,1), T9@(2,3), T0@(4,0)]
```

execution/block-partitioner/src/v2/conflicting_txn_tracker.rs
zjma: then the constructor needs a `num_shards`... Not much difference?

execution/block-partitioner/src/v2/conflicting_txn_tracker.rs
zjma: This `writer_set` is not needed any more!

aptos-move/aptos-vm/src/sharded_block_executor/tests.rs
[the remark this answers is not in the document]
zjma: we already have and reuse `test_utils::test_sharded_block_executor_no_conflict(partitioner, sharded_block_executor)`...

execution/block-partitioner/src/v2/mod.rs
zjma: i'd like to keep struct and tests in the same file. Now all the files are much smaller, that should not hurt readability any more?

aptos-move/aptos-vm/src/sharded_block_executor/tests.rs
```diff
@@ -21,11 +24,13 @@ fn setup_sharded_block_executor<S: StateView + Sync + Send + 'static>(
 
 #[test]
 fn test_sharded_block_executor_no_conflict() {
-    for last_round_partition in [true, false].iter() {
-        let sharded_block_executor = setup_sharded_block_executor(8, Some(4));
+    let num_shards = 8;
+    for last_round_partition in [true, false] {
+        let mut partitioner = Box::new(ShardedBlockPartitioner::new(num_shards, 2, 0.9, last_round_partition));
```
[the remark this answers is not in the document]
zjma: But V2 has new parameters `num_threads` and `dashmap_num_shards`...

execution/block-partitioner/src/v2/init.rs
```diff
@@ -0,0 +1,59 @@
...
+    v2::{
+        conflicting_txn_tracker::ConflictingTxnTracker, counters::MISC_TIMERS_SECONDS,
+        state::PartitionState, types::PreParedTxnIdx, PartitionerV2,
+    },
+};
+use rayon::{iter::ParallelIterator, prelude::IntoParallelIterator};
+use std::sync::RwLock;
+
+impl PartitionerV2 {
+    pub(crate) fn init(state: &mut PartitionState) {
```
sitalkedia, line 14: This function really belong in state.rs and also should be called from the constructor of the state?

execution/block-partitioner/src/v2/conflicting_txn_tracker.rs
sitalkedia: IMO, its just cleaner to pass `num_shards` in the constructor.. but up to you.

execution/block-partitioner/src/v2/partition_to_matrix.rs
```diff
@@ -0,0 +1,189 @@
...
+use rayon::{
+    iter::ParallelIterator,
+    prelude::{IntoParallelIterator, IntoParallelRefIterator},
+};
+use std::{mem, sync::RwLock};
+
+impl PartitionerV2 {
+    /// Populate `state.finalized_txn_matrix` with txns flattened into a matrix (num_rounds by num_shards),
+    /// in a way that avoid in-round cross-shard conflicts.
+    pub(crate) fn partition_to_matrix(state: &mut PartitionState) {
```
sitalkedia, line 19: nit: this function can be renamed to `remove_cross_shard_dependencies` ?

execution/block-partitioner/src/v2/partition_to_matrix.rs
```diff
@@ -0,0 +1,189 @@
...
+    ) -> (Vec<Vec<PreParedTxnIdx>>, Vec<Vec<PreParedTxnIdx>>) {
+        let _timer = MISC_TIMERS_SECONDS
+            .with_label_values(&[format!("round_{round_id}").as_str()])
+            .start_timer();
+
+        let num_shards = remaining_txns.len();
+        let mut discarded: Vec<RwLock<Vec<PreParedTxnIdx>>> = Vec::with_capacity(num_shards);
+        let mut tentatively_accepted: Vec<RwLock<Vec<PreParedTxnIdx>>> =
+            Vec::with_capacity(num_shards);
+        let mut finally_accepted: Vec<RwLock<Vec<PreParedTxnIdx>>> = Vec::with_capacity(num_shards);
```
sitalkedia, line 85: Add some comment explaining the difference between `tentatively_accepted` and `finally_accepted`

execution/block-partitioner/src/v2/partition_to_matrix.rs
```diff
@@ -0,0 +1,189 @@
...
+                .enumerate()
+                .collect::<Vec<_>>()
+                .into_par_iter()
+                .for_each(|(shard_id, txn_idxs)| {
+                    txn_idxs.into_par_iter().for_each(|txn_idx| {
+                        let in_round_conflict_detected = state
+                            .all_hints(txn_idx)
+                            .into_iter()
+                            .any(|key_idx| state.key_owned_by_another_shard(shard_id, key_idx));
+                        //TODO: early stop.
```
sitalkedia, line 109: Can you implement this TODO?

execution/block-partitioner/src/v2/state.rs
```diff
@@ -0,0 +1,349 @@
+// Copyright © Aptos Foundation
+
+use crate::{
```
sitalkedia, line 3: Few high level thoughts 

1. Can you take a closer look into places where we are cloning the Transaction or the `StateKey` and see if we can avoid them by dealing with references instead? I have commented in few places but there might be more.

2. Wherever possible, use `par_iter()` with `min_length` to avoid rayon spawning very small task for processing. This can be adverse effect on the performance.

execution/block-partitioner/src/pre_partition/uniform.rs
```diff
@@ -0,0 +1,57 @@
...
+
+/// Evenly divide txns. Example: processing txns 0..11 results in [[0,1,2,3],[4,5,6,7],[8,9,10]].
+pub struct UniformPartitioner {}
+
+impl PrePartitioner for UniformPartitioner {
+    fn pre_partition(
+        &self,
+        transactions: &[AnalyzedTransaction],
+        num_shards: usize,
+    ) -> Vec<Vec<PreParedTxnIdx>> {
```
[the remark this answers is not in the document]
zjma: i actually meant `PrePar[tition]edTxnIdx` ...

execution/block-partitioner/src/v2/config.rs
```diff
@@ -0,0 +1,30 @@
+// Copyright © Aptos Foundation
+
+use crate::v2::PartitionerV2;
+
+#[derive(Clone, Copy, Debug)]
+pub struct PartitionerV2Config {
```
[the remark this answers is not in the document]
zjma: It seems `num_shards` is not needed - it can partition for any `num_shards` (given as part of `partition()` invocation) after `new`ed.

types/src/block_executor/partitioner.rs
```diff
@@ -36,9 +50,41 @@ impl ShardedTxnIndex {
 /// Denotes a set of cross shard edges, which contains the set (required or dependent) transaction
 /// indices and the relevant storage locations that are conflicting.
 pub struct CrossShardEdges {
-    edges: HashMap<ShardedTxnIndex, Vec<StorageLocation>>,
+    pub edges: HashMap<ShardedTxnIndex, Vec<StorageLocation>>,
 }
 
+impl PartialEq for CrossShardEdges {
+    fn eq(&self, other: &Self) -> bool {
```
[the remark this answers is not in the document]
zjma: that's needed by determinism tests where we run partitioner 2+ times and compare with `==`.

execution/block-partitioner/src/pre_partition/mod.rs
```diff
@@ -0,0 +1,25 @@
+// Copyright © Aptos Foundation
+
+use crate::v2::types::PreParedTxnIdx;
+use aptos_types::{
+    block_executor::partitioner::ShardId, transaction::analyzed_transaction::AnalyzedTransaction,
+};
+
+pub trait PrePartitioner: Send {
+    fn pre_partition(
```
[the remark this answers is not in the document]
zjma: ran out of ideas... added comment instead.

execution/block-partitioner/src/v2/types.rs
```diff
@@ -0,0 +1,88 @@
...
+/// Represents positions of a txn after it is assigned to a sub-block.
+///
+/// Different from `aptos_types::block_executor::partitioner::ShardedTxnIndex`,
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
+pub struct ShardedTxnIndexV2 {
+    pub sub_block_idx: SubBlockIdx,
+    pub ori_txn_idx: PreParedTxnIdx,
+}
+
+impl Ord for ShardedTxnIndexV2 {
```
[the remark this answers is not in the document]
zjma: Needed by BTreeSets, which is needed by ConflictTxnTrackers

execution/block-partitioner/src/v2/types.rs
```diff
@@ -0,0 +1,88 @@
...
+        SubBlockIdx { round_id, shard_id }
+    }
+
+    pub fn global() -> SubBlockIdx {
+        SubBlockIdx::new(GLOBAL_ROUND_ID, GLOBAL_SHARD_ID)
+    }
+}
+
+/// The position of a txn in the block after `pre_partition` and before `flatten_to_rounds`.
+pub type PreParedTxnIdx = usize;
```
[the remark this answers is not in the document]
zjma: In partition logic we have other `TxnIdx`s...
`PreParedTxnIdx` with some comments is already the best bet IMO...

execution/block-partitioner/src/v2/mod.rs
```diff
@@ -0,0 +1,176 @@
...
+    dashmap_num_shards: usize,
+    merge_discarded: bool,
+}
+
+impl PartitionerV2 {
+    pub fn new(
+        num_threads: usize,
+        num_rounds_limit: usize,
+        avoid_pct: u64,
+        dashmap_num_shards: usize,
```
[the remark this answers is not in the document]
zjma: I'm seeing 20% or even more regression if not configured optimally.

execution/block-partitioner/src/v2/state.rs
```diff
@@ -0,0 +1,349 @@
...
+    pub(crate) avoid_pct: u64,
+    pub(crate) merge_discarded: bool,
+    pub(crate) thread_pool: Arc<ThreadPool>,
+
+    // Holding all the txns.
+    pub(crate) txns: Vec<RwLock<Option<AnalyzedTransaction>>>,
+
+    // Pre-partitioning results.
+    pub(crate) pre_partitioned: Vec<Vec<PreParedTxnIdx>>,
+    pub(crate) start_txn_idxs_by_shard: Vec<PreParedTxnIdx>,
```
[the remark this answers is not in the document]
zjma: There is value. Otherwise
```
    pub(crate) start_index_matrix: Vec<Vec<PreParedTxnIdx>>,
    pub(crate) new_txn_idxs: Vec<RwLock<TxnIndex>>,
```
becomes
```
    pub(crate) start_index_matrix: Vec<Vec<TxnIdx>>,
    pub(crate) new_txn_idxs: Vec<RwLock<TxnIndex>>,
```
which is terrible...

execution/block-partitioner/src/v2/state.rs
```diff
@@ -0,0 +1,349 @@
...
+            wsets.push(RwLock::new(HashSet::with_capacity(txn.write_hints().len())));
+            rsets.push(RwLock::new(HashSet::with_capacity(txn.read_hints().len())));
+        }
+        let takable_txns = thread_pool.install(|| {
+            txns.into_par_iter()
+                .map(|txn| RwLock::new(Some(txn)))
+                .collect()
+        });
+        let start_txn_idxs_by_shard = start_txn_idxs(&pre_partitioned);
+        Self {
```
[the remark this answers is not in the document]
zjma: `new()` is allocation. `init` is to fill in the initial value. Better keep them separated (and we can time them separatedly).

execution/block-partitioner/src/v2/init.rs
zjma: See [here](https://github.com/aptos-labs/aptos-core/pull/9488#discussion_r1294772987).

execution/block-partitioner/src/v2/state.rs
zjma: 1. fixed. Clone is only needed as when initially numbering them.
2. Will find another time to bench it and decide the right threshold. Maybe another PR?

execution/block-partitioner/src/test_utils.rs
```diff
@@ -79,3 +110,213 @@ pub fn create_signed_p2p_transaction(
...
+/// Assert partitioner correctness for `ShardedBlockPartitioner` and `V2Partitioner`:
+/// - Transaction set remains the same after partitioning.
+/// - The relative order of the txns from the same sender
+/// - For a cross-shard dependency, the consumer txn always comes after the provider txn in the sharded block.
+/// - Required edge set matches dependency edge set.
+/// - Before the last round, there is no in-round cross-shard dependency.
+///
+/// Also print a summary of the partitioning result.
+#[cfg(test)]
+pub fn verify_partitioner_output(
```
sitalkedia, line 153: nit: this function it pretty long, can you take a look into breaking it down so that its easier to follow?

execution/block-partitioner/src/v2/conflicting_txn_tracker.rs
```diff
@@ -0,0 +1,151 @@
...
+/// For txns that claimed to access the same storage location,
+/// it caches some metadata about the location and also keeps track of their status (pending or position finalized) throughout the partitioning process.
+#[derive(Debug, Clone, Eq, PartialEq, Serialize, Deserialize)]
+pub struct ConflictingTxnTracker {
+    /// The storage location on which conflicting txns are being tracked by this tracker.
+    pub storage_location: StorageLocation,
+    /// A randomly chosen owner shard of the storage location, for conflict resolution purpose.
+    pub anchor_shard_id: ShardId,
+    /// Txns that (1) read the current storage location and (2) have not been accepted.
+    pending_reads: BTreeSet<PreParedTxnIdx>,
```
sitalkedia, line 23: can we not get rid of `PreParedTxnIdx` type and use `TxnIdx` instead? The name is very confusing and doesn't serve any purpose IMO.

execution/block-partitioner/src/v2/state.rs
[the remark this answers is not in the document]
sitalkedia: You can still time them separately and keep them as separate functions.. just call `init` from constructor, there is really no need to call `init` from outside.

execution/block-partitioner/src/v2/mod.rs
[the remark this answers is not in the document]
sitalkedia: What's the optimal configuration?

execution/block-partitioner/src/v2/state.rs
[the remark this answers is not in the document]
sitalkedia: As discussed offline, let's rename this to `PrePartitionedTxnIdx`

sitalkedia (approved):
Looks pretty good % Minor nits.. thank you for patiently addressing the review comments.

aptos-move/aptos-vm/src/sharded_block_executor/tests.rs
```diff
@@ -7,6 +7,9 @@ use crate::{
     },
     ShardedBlockExecutor,
 };
+use aptos_block_partitioner::{
+    sharded_block_partitioner::config::PartitionerV1Config, v2::config::PartitionerV2Config,
```
manudhundi, line 11: Is there a good reason for the hierarchies of PartitionerV1Config and PartitionerV2Config to be different ?

execution/block-partitioner/src/lib.rs
```diff
@@ -2,61 +2,60 @@
...
+    hash::{Hash, Hasher},
+};
+use v2::config::PartitionerV2Config;
+mod pre_partition;
+
+pub trait BlockPartitioner: Send {
+    fn partition(
+        &self,
+        transactions: Vec<AnalyzedTransaction>,
+        num_shards: usize,
```
manudhundi, line 27: Other configs like 'num_threads', 'num_rounds_limit', etc are not a part of this trait but rather part of the obj (PartitionerV2) implementing this trait. Why is 'num_shards' a parameter here ? That is, what is special about 'num_shards' vs say 'num_rounds_limit' ?

execution/block-partitioner/src/main.rs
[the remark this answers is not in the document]
manudhundi: `jemallocator` change should be probably merged as a separate PR or atleast as 'un-squashed' commit ?

execution/block-partitioner/src/sharded_block_partitioner/config.rs
```diff
@@ -0,0 +1,58 @@
+// Copyright © Aptos Foundation
+
+use crate::{sharded_block_partitioner::ShardedBlockPartitioner, BlockPartitioner};
+use aptos_types::block_executor::partitioner::RoundId;
+
+#[derive(Clone, Copy, Debug)]
+pub struct PartitionerV1Config {
```
manudhundi, line 7: Follow up on the comment regarding the v1, v2 config hierarchies.
We have a file called `../../config.rs`, but it has only  `PartitionerV1Config`

execution/block-partitioner/src/pre_partition/mod.rs
```diff
@@ -0,0 +1,25 @@
...
+        &self,
+        transactions: &[AnalyzedTransaction],
+        num_shards: usize,
+    ) -> Vec<Vec<PrePartitionedTxnIdx>>;
+}
+
+pub mod uniform_partitioner;
+
+pub fn start_txn_idxs(pre_partitioned: &Vec<Vec<PrePartitionedTxnIdx>>) -> Vec<PrePartitionedTxnIdx> {
```
manudhundi: Looks correct, but probably good idea to have a unit test for this function.

execution/block-partitioner/src/pre_partition/uniform_partitioner.rs
```diff
@@ -0,0 +1,55 @@
...
+
+impl PrePartitioner for UniformPartitioner {
+    fn pre_partition(
+        &self,
+        transactions: &[AnalyzedTransaction],
+        num_shards: usize,
+    ) -> Vec<Vec<PrePartitionedTxnIdx>> {
+        let num_items = transactions.len();
+        let num_chunks = num_shards;
+        let num_big_chunks = num_items % num_chunks;
```
manudhundi, line 21: Calling it 'big_chunk' and 'small_chunk' will make reading the code harder. 
You can write this logic with 'chunk_size' instead of 'small_chunk_size'; 'num_chunks_with_overflow' or 'num_chunks_with_extra' instead of 'num_big_chunks'

execution/block-partitioner/src/pre_partition/uniform_partitioner.rs
```diff
@@ -0,0 +1,55 @@
...
+    let txns = block_gen.rand_block(&mut rng, 18);
+    let partitioner = UniformPartitioner {};
+    let actual = partitioner.pre_partition(txns.as_slice(), 5);
+    assert_eq!(
+        vec![4, 4, 4, 3, 3],
+        actual.iter().map(|v| v.len()).collect::<Vec<usize>>()
+    );
+    assert_eq!((0..18).collect::<Vec<usize>>(), actual.concat());
+
+    let actual = partitioner.pre_partition(txns.as_slice(), 3);
```
manudhundi, line 49: A better test case would also test the 'extra' txns per shard logic ? Currently 18/3 divides the txns evenly across all shards

execution/block-partitioner/src/v2/partition_to_matrix.rs
```diff
@@ -0,0 +1,224 @@
...
+
+impl PartitionerV2 {
+    /// Populate `state.finalized_txn_matrix` with txns flattened into a matrix (num_rounds by num_shards),
+    /// in a way that avoid in-round cross-shard conflicts.
+    pub(crate) fn remove_cross_shard_dependencies(state: &mut PartitionState) {
+        let _timer = MISC_TIMERS_SECONDS
+            .with_label_values(&["remove_cross_shard_dependencies"])
+            .start_timer();
+
+        let mut remaining_txns = mem::take(&mut state.pre_partitioned);
```
manudhundi, line 33: Why are we copying all the txns and then reseting the state ? Why not directly work on the mutable ref 'state.pre_partitioned' ?

aptos-move/aptos-vm/src/sharded_block_executor/tests.rs
zjma: v1 is actually an alias of `sharded_block_partitioner`, so the hierarchy is the same?

execution/block-partitioner/src/lib.rs
zjma: IMO `num_shards ` actually doesn't have to be a parameter here. I think we made the trait too early...

execution/block-partitioner/src/sharded_block_partitioner/config.rs
zjma: Same reason: `sharded_block_partitioner`=`v1`...

execution/block-partitioner/src/pre_partition/mod.rs
zjma: fixed

execution/block-partitioner/src/pre_partition/uniform_partitioner.rs
zjma: That test case actually already have both 18/5 and 18/3.

execution/block-partitioner/src/v2/partition_to_matrix.rs
zjma: I think there's no copy, only object ownership transfer. So already equivalent to working on the mutable ref?

execution/block-partitioner/src/pre_partition/uniform_partitioner.rs
zjma: fixed

aptos-move/aptos-vm/src/sharded_block_executor/tests.rs
zjma: I added some clarification comments on top of `PartitionerV1Config`...

execution/block-partitioner/src/pre_partition/uniform.rs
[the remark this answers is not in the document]
manudhundi: PrePartitionedTxnIdx sounds good to me.
If that seems long, then how about 'PrePartndIdx' or 'PrePrtitndIdx' ?

execution/block-partitioner/src/lib.rs
manudhundi: Can you fix the trait then, if it is simple ? or else maybe create a low priority tech debt ?
Upto you, I am not particular on either

aptos-move/aptos-vm/src/sharded_block_executor/tests.rs
manudhundi: Then, perhaps call it `sharded_block_partitioner_v2` ?

execution/block-partitioner/src/pre_partition/uniform_partitioner.rs
manudhundi: cool, thanks

execution/block-partitioner/src/pre_partition/uniform_partitioner.rs
manudhundi: Oh my bad, I missed 18/5

aptos-move/aptos-vm/src/sharded_block_executor/tests.rs
zjma: it's tricky... the v2 is also only a multithreaded program rather than really being "sharded"...

zjma: I'd rather rename the old one to v1, but that unnecessarily makes the current diff bigger...

execution/block-partitioner/src/lib.rs
zjma: > create a low priority tech debt ?
do you mean create a TODO comment?

execution/block-partitioner/src/v2/partition_to_matrix.rs
manudhundi: ok got it

aptos-move/aptos-vm/src/sharded_block_executor/tests.rs
manudhundi: ok

execution/block-partitioner/src/lib.rs
manudhundi: sure TODO comment is good. You can resolve this comment after that

zjma: fixed

aptos-move/aptos-vm/src/sharded_block_executor/tests.rs
zjma: added a TODO as well.

execution/block-partitioner/src/pre_partition/uniform.rs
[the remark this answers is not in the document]
zjma: Has been renamed to `PrePartitionedTxnIdx`, not too long!

manudhundi (approved):

PR #9492, opened 2023-08-04
[TS SDK] bump version to 1.17.0

0xmaayan:
### Description
- Add support for a fee payer transaction
- Return transaction message error when transaction has failed when `checkSuccess` is set to true


junkil-park (approved):

junkil-park (commented):
Thank you!

xbtmatt (approved):

PR #9493, opened 2023-08-04
[indexer-grpc] k6 loadtest

rustielin:
### Description

An initial pass at a simple gRPC load test for indexer gRPC. It spins up 10 vus (virtual users), which are configurable, and initiates the RawTransactions stream. 

The loadtest does not complete other than by the configured timeout, since the transaction stream is supposed to be continuous. In future iterations, we should add a k6 check to catch connection drops. And also measure the txn/s throughput. 

### Test Plan

Run it locally. We can run this in CI after it's stabilized. 

```
# Run indexer gRPC and single-node testnet locally via docker-compose
# From testsuite/ directory
$ IMAGE_TAG=nightly_1063755d6de638450a5f3d9f6f6a33c4c34cfbd6 ./indexer_grpc_local.py --verbose start

# Transpile
$ cd testsuite/loadtest-k6
$ yarn install
$ yarn start

# Run load test from within its dir
$ k6 run dist/indexer-grpc-test.js

     data_received................: 244 MB 4.1 MB/s
     data_sent....................: 50 kB  825 B/s
     grpc_streams.................: 10     0.166665/s
     grpc_streams_msgs_received...: 710    11.833216/s
     grpc_streams_msgs_sent.......: 10     0.166665/s
     vus..........................: 10     min=10      max=10
     vus_max......................: 10     min=10      max=10


running (1m00.0s), 00/10 VUs, 0 complete and 10 interrupted iterations
contacts ✓ [======================================] 10 VUs  30s
```


larry-aptos:
If we want to test finite streaming, we can specify `transaction_count` in the request.

rustielin:
merging as reference for now

PR #9495, opened 2023-08-04
[testsuite][pangu][pte] Creating a Pangu Command to Create a Transaction Emitter

olsenbudanur:
### Description
This PR creates a simple command to create a transaction emitter pod in an existing testnet. 
### Test Plan
The code is functional at creating a transaction emitter in aptos-forge-exp-0, and the k8s abstraction it uses is already well tested.


testsuite/pangu_lib/testnet_commands/transaction_emitter.py
```diff
@@ -0,0 +1,39 @@
...
+
+    pod_spec: client.V1PodSpec = client.V1PodSpec(
+        restart_policy="Never",
+        containers=[container],
+    )
+
+    pod: client.V1Pod = client.V1Pod(
+        api_version="v1",
+        kind="Pod",
+        metadata=client.V1ObjectMeta(name=f"{testnet_name}-tx-emitter"),
```
rustielin, line 35: you might want to give it a random suffix or something, so you can run multiple transaction emitters in parallel, or even just to run them in sequence if you don't delete the pods afterwards

olsenbudanur: Makes sense, on it

testsuite/pangu_lib/testnet_commands/transaction_emitter.py
```diff
@@ -0,0 +1,39 @@
...
+
+def transaction_emitter_main(testnet_name: str, system_context: SystemContext):
+    pod: client.V1Pod = create_pod(testnet_name)
+    system_context.kubernetes.create_resource(pod, testnet_name)
+
+
+def create_pod(testnet_name: str) -> client.V1Pod:
+    container: client.V1Container = client.V1Container(
+        name=f"{testnet_name}-tx-emitter",
+        image="aptoslabs/tools:devnet_performance",
```
rustielin, line 15: this image should be configurable as well ;) maybe if it's not specified, you can inherit from the top level pangu config

olsenbudanur: It'd be UTIL file imo, not the pangu config. Can't access the pangu config here.

testsuite/pangu_lib/testnet_commands/transaction_emitter.py
```diff
@@ -0,0 +1,76 @@
...
+    # Get logs
+    command = ["kubectl", "logs", "-f", pod_name, "-n", testnet_name]
+    while True:
+        system_context.shell.run(command, stream_output=True)
+        user_input = input(
+            '-------------------------------------------------------\n-Press Enter to get the newest logs,\n-Ctrl + c to end looking at the logs...\n-Type "delete" to delete the transaction emitter pod\n-------------------------------------------------------\n'
+        )
+        if user_input == "delete":
+            system_context.kubernetes.delete_resource(pod, testnet_name)
+            break
```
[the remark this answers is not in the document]
olsenbudanur: On it!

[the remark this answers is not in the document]
olsenbudanur: Done, lmk if you like it

testsuite/pangu_lib/testnet_commands/transaction_emitter.py
```diff
@@ -0,0 +1,95 @@
...
+    container: client.V1Container = client.V1Container(
+        name=pods_name,
+        image=util.DEFAULT_TRANSACTION_EMITTER_IMAGE,
+        env=[
+            client.V1EnvVar(name="RUST_BACKTRACE", value="1"),
+            client.V1EnvVar(name="REUSE_ACC", value="1"),
+        ],
+        command=command_array,
+        resources=client.V1ResourceRequirements(
+            requests={"cpu": "1", "memory": "1Gi"},  # Check if too much/not enough
```
rustielin, line 78: change the request as well to be higher

rustielin (approved):
very nice! thanks for getting this out so quickly

some nits above, but generally LGTM

testsuite/pangu_lib/testnet_commands/transaction_emitter.py
```diff
@@ -0,0 +1,95 @@
...
+        random.choices(string.ascii_lowercase + string.digits, k=8)
+    )
+    pod_name = f"{testnet_name}-tx-emitter-{random_postfix}"
+
+    #
+    # Create Pod
+    log.info("Starting transaction emitter...")
+    pod: client.V1Pod = create_transaction_emitter_pod(pod_name, command_array)
+    system_context.kubernetes.create_resource(pod, testnet_name)
+    log.info("Transaction emitter started...")
```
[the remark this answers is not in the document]
olsenbudanur: Ok, 2 ways to go about dry_run. To have the dry run functionality, the user would need to provide a dry_run bool and workspace str. These 2 ways exist because of how we are piping all options thru to the transaction emitter. 

- The CLI can take this as an arg, in which case the user would need to explicitly provide this info every time (can't be optional) even if they dont want to do a dry run. 
- We can let the user have the illusion of providing an option for the dry-run/workspace, but on the code side we'd traverse the args that are parsed to be sent to the transaction emitter to see if they've sent a --dry_run and --workspace option.

What's your preference? I prefer the first option.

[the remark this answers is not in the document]
olsenbudanur: Nvm, fixed this.

rustielin (approved):

ibalajiarun (approved):
LGTM. Thanks!

PR #9497, opened 2023-08-04
Revert "[Data Streaming Service] Increase data prefetcher for PFNs."

JoshLind:
### Description
- This reverts commit 0f07a3083a271412544c35f30b808ca91e8ff7af.
- This PR was created using: `git revert 0f07a3083a271412544c35f30b808ca91e8ff7af `

The network stack is yet to handle the increased aggression and some test PFNs (with weak links) are seeing health check failures due to bandwidth saturation.

### Test Plan
Existing test infrastructure.

PR #9500, opened 2023-08-04
[forge][land blocking] recalibrate latency limits

igor-aptos:
There were a few sporadic failures on landblocking PRs. increasing limits, and printing them better.

### DIFF
--- a/testsuite/forge-cli/src/main.rs
+++ b/testsuite/forge-cli/src/main.rs
@@ -1532,10 +1532,10 @@ fn realistic_env_max_load_test(
                 .add_latency_threshold(3.4, LatencyType::P50)
                 .add_latency_threshold(4.5, LatencyType::P90)
                 .add_latency_breakdown_threshold(LatencyBreakdownThreshold::new_strict(vec![
-                    (LatencyBreakdownSlice::QsBatchToPos, 0.3),
-                    (LatencyBreakdownSlice::QsPosToProposal, 0.25),
+                    (LatencyBreakdownSlice::QsBatchToPos, 0.35),
+                    (LatencyBreakdownSlice::QsPosToProposal, 0.5),
                     (LatencyBreakdownSlice::ConsensusProposalToOrdered, 0.8),
-                    (LatencyBreakdownSlice::ConsensusOrderedToCommit, 0.6),
+                    (LatencyBreakdownSlice::ConsensusOrderedToCommit, 0.65),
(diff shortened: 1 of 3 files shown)


JoshLind (approved):

testsuite/forge-cli/src/main.rs
```diff
@@ -1532,10 +1532,10 @@ fn realistic_env_max_load_test(
                 .add_latency_threshold(3.4, LatencyType::P50)
                 .add_latency_threshold(4.5, LatencyType::P90)
                 .add_latency_breakdown_threshold(LatencyBreakdownThreshold::new_strict(vec![
-                    (LatencyBreakdownSlice::QsBatchToPos, 0.3),
-                    (LatencyBreakdownSlice::QsPosToProposal, 0.25),
+                    (LatencyBreakdownSlice::QsBatchToPos, 0.35),
```
bchocho, line 1535: I've seen in a couple cases where the value gets pretty high, but drops back down. (Not sure why, might need some digging into.) In these case, I think we need to use a max_breach_pct to avoid failures (10% seems like it would work.)

Examples (have to expand the comment):
* https://github.com/aptos-labs/aptos-core/pull/9482#issuecomment-1665798925
* https://github.com/aptos-labs/aptos-core/pull/9497#issuecomment-1666061643

igor-aptos: yeah, I want to have stats that I print here, and then I'll probably have a limit for both max and p75/p90.
but don't want to make more noise without having better data.

bchocho: Sounds reasonable!

bchocho (approved):

testsuite/testcases/src/lib.rs
```diff
@@ -184,14 +184,31 @@ impl NetworkTest for dyn NetworkLoadTest {
...
+                "Latency breakdown for phase {}: {:?}",
+                phase,
+                phase_stats
+                    .latency_breakdown
+                    .keys()
+                    .into_iter()
+                    .map(|slice| {
+                        let slice_samples = phase_stats.latency_breakdown.get_samples(&slice);
+                        format!(
+                            "{:?}: max: {}, avg: {}",
```
igor-aptos, line 204:
```suggestion
                            "{:?}: max: {:.3}, avg: {:.3}",
```

PR #9501, opened 2023-08-04
[gas] fix abstract gas cost for storage

vgao1996:
This fixes a bug in the original abstract gas implementation, allowing the io gas to be expressed as an abstract amount rather than a concrete one. This should not change the behavior of the production binary at all.

### DIFF
--- a/aptos-move/aptos-vm-types/src/storage.rs
+++ b/aptos-move/aptos-vm-types/src/storage.rs
@@ -96,10 +96,6 @@ pub struct StoragePricingV2 {
 }
 
 impl StoragePricingV2 {
-    pub fn zeros() -> Self {
-        Self::new_without_storage_curves(LATEST_GAS_FEATURE_VERSION, &AptosGasParameters::zeros())
-    }
-
     pub fn new_with_storage_curves(
         feature_version: u64,
         storage_gas_schedule: &StorageGasSchedule,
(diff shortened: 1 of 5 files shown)


runtian-zhou (approved):
How can we make sure the behavior won't change?

vgao1996:
@runtian-zhou Through replay: https://github.com/aptos-labs/aptos-core/actions/runs/5766629688

aptos-move/aptos-vm-types/src/storage.rs
```diff
@@ -253,10 +253,10 @@ impl StoragePricing {
                     gas_params,
                 )),
             },
-            10.. => V2(StoragePricingV2::new_without_storage_curves(
```
msmouse, line 256: remove `new_without_storage_curves`?

msmouse: https://github.com/aptos-labs/aptos-core/blob/137acee4c6dddb1c86398dce25b041d78a3028d3/aptos-move/aptos-vm-types/src/storage.rs#L429 change this to V3 probably?

vgao1996: Yeah, trying this right now

vgao1996: Done.

aptos-move/e2e-testsuite/Cargo.toml
```diff
@@ -19,7 +19,7 @@ aptos-crypto = { workspace = true }
 aptos-framework = { workspace = true }
 aptos-gas-algebra = { workspace = true }
 aptos-gas-meter = { workspace = true }
-aptos-gas-schedule = { workspace = true }
+aptos-gas-schedule = { workspace = true, features = ["testing"] }
```
msmouse, line 22: why is this?

msmouse (approved):

aptos-move/e2e-testsuite/Cargo.toml
vgao1996: This is another bug caused by the original PR. If you run `cargo test` for the e2e-testsuite crate alone, you'll notice half of the test failing due to error GAS_UNIT_PRICE_BELOW_MIN_BOUND. CI failed to catch this because it performs feature unification for all crates.

movekevin (changes requested):
This currently leads to a difference of 1 in gas (rounding error?) when running replay

vgao1996:
New replay task: https://github.com/aptos-labs/aptos-core/actions/runs/5793651188

vgao1996:
@movekevin can you re-approve the PR since it's now passing replay? Thanks!

movekevin (approved):

PR #9503, opened 2023-08-05
[GHA] Skip non-required tests in PRs and reduce scheduled jobs.

JoshLind:
### Description

This PR offers several changes (each in their commit):
1. To avoid running unnecessary tests on each PR, we skip non-required (i.e., non-land blocking) tests. If PR authors still want to run these tests, they can add the PR label `CICD:non-required-tests`.
2. We remove the scheduled runs of non-critical forge tests. Currently, only the `stable` forge tests are considered critical and high-signal. Other tests are less stable, flaky and really only need to be run on an adhoc basis (e.g., to test performance improvements, etc.). Going forward, we may want to change this (e.g., introduce separate categories of forge jobs, split up the stable jobs, etc.). But, until then, it makes sense to disable the nightly runs and just use ad-hoc scheduling for now.
3. We reduce the schedule frequency of several non-critical fullnode sync jobs (from daily, to every 3 days).
4. Add new PR labels to the non-required jobs file (to ensure jobs run when labels are added).

### Test Plan
Existing test infrastructure.

### DIFF
--- a/.github/workflows/check-sdk-examples.yaml
+++ b/.github/workflows/check-sdk-examples.yaml
@@ -1,6 +1,7 @@
 name: "Check SDK examples"
 on:
   pull_request:
+    types: [labeled, opened, synchronize, reopened, auto_merge_enabled]
   push:
     branches:
       - devnet
--- a/.github/workflows/check-sdk-examples.yaml
+++ b/.github/workflows/check-sdk-examples.yaml
@@ -15,6 +15,7 @@ jobs:
   # whereas we run the test-sdk-confirm-client-generated-publish against a node
   # built from the same commit and run as part of that CI job.
   run-examples:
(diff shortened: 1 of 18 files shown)


JoshLind:
@igor-aptos, @bchocho -- are you folks okay with removing the automatic scheduling of the `unstable` and `PFN` jobs? In my opinion:
1. If we want to keep running the `unstable` tests, we should fix all the broken ones and see if we're actively monitoring their signals? I personally never look at them.
2. @bchocho, I don't require the `PFN` jobs to run everyday. Ad-hoc seems more than enough for me. But, I wanted to check if you're okay with it, too?

igor-aptos (approved):
Sounds good to me!

bchocho (approved):
Yeah makes sense to run these adhoc until they become stable

PR #9506, opened 2023-08-05
[framework] Add Aptos-token mint function that returns object

gregnazario:
### Description
Function probably needs a better name, but the long needed non-entry version of mint.

### Test Plan
See move tests


movekevin (approved):
Thanks for adding this! Can you add a similar function for create_collection as well?

gregnazario:
Updated the rest of the tests including in examples, and added for soulbound & collections

xbtmatt (approved):
Awesome!

Issue #9251, opened 2023-07-23
[Bug] aptos-move unit tests are run without extended_checks


brmataptos:
# 🐛 Bug

Unit tests initiated in [aptos-core/aptos-move/framework/tests/move_unit_test.rs](https://github.com/aptos-labs/aptos-core/blob/cf1accf5611b1afa62c62514e371acf44357eed4/aptos-move/framework/tests/move_unit_test.rs#L15) are not running compiler extended_checks.  This means that:
- these extended_checks are not exercised on much code in testing.
- any added functionality needed to process aptos-specific move code needs a different code path.

Tests run here are as follows:
```
aptos-move/framework/tests/move_unit_test.rs
15: fn run_tests_for_pkg(path_to_pkg: impl Into<String>) {
53:     run_tests_for_pkg("aptos-framework");
58:     run_tests_for_pkg("aptos-stdlib");
63:     run_tests_for_pkg("move-stdlib");
68:     run_tests_for_pkg("aptos-token");
73:     run_tests_for_pkg("aptos-token-objects");
```

The implementation of `run_tests_for_pkg` leverages `third_party` 's  method here:
[`move_cli::base::test::run_move_unit_tests`](https://github.com/aptos-labs/aptos-core/blob/cf1accf5611b1afa62c62514e371acf44357eed4/third_party/move/tools/move-cli/src/base/test.rs#L150) which doesn't have access to `extended_checks`, which lives outside of `third_party`

The `aptos-framework` tests need something done to call `extended_checks::run_extended_checks()` on the built model.  Note that [code implementing `aptos move TestPackage`](`https://github.com/aptos-labs/aptos-core/blob/cf1accf5611b1afa62c62514e371acf44357eed4/crates/aptos/src/move_tool/mod.rs#L461`) uses the same test runner but first "[builds] `the Move model for extended checks`".

Perhaps these tests should do something similar to better test the `extended_checks` code.

Ideally, Aptos Move-specific code would be better integrated with Third-Party Move code through less hacky dialect plugins, unless the code were merged and specialized for Aptos's Move dialect.

## A bit more info with code pointers.

You can see here:
https://github.com/aptos-labs/aptos-core/blob/cf1accf5611b1afa62c62514e371acf44357eed4/aptos-move/framework/tests/move_unit_test.rs#L9
and here:
[...]

PR #9507, opened 2023-08-05
[events v2] Sample of how to check for event type attributes in the extended checker

wrwg:
Since this kind of check (where we look at each call) hasn't been done before, this PR demonstrates how the plumbing could look like.

### DIFF
--- a/aptos-move/framework/src/extended_checks.rs
+++ b/aptos-move/framework/src/extended_checks.rs
@@ -12,11 +12,16 @@ use move_core_types::{
 use move_model::{
     ast::{Attribute, AttributeValue, Value},
     model::{
-        FunctionEnv, GlobalEnv, Loc, ModuleEnv, NamedConstantEnv, Parameter, QualifiedId,
+        FunId, FunctionEnv, GlobalEnv, Loc, ModuleEnv, NamedConstantEnv, Parameter, QualifiedId,
         StructEnv, StructId,
     },
     symbol::Symbol,
     ty::{PrimitiveType, ReferenceKind, Type},
+};
+use move_stackless_bytecode::{
+    function_target::{FunctionData, FunctionTarget},
+    stackless_bytecode::{AttrId, Bytecode, Operation},
+    stackless_bytecode_generator::StacklessBytecodeGenerator,
 };
 use std::{collections::BTreeMap, rc::Rc, str::FromStr};
 use thiserror::Error;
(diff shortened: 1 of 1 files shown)


davidiw (commented):
Wow! Nice.

@lightmark , maybe your dreams can come true....

you can use this to get your options 2 and 3 without doing anything too fun upstream.

aptos-move/framework/src/extended_checks.rs
```diff
@@ -67,6 +72,7 @@ impl<'a> ExtendedChecker<'a> {
                 self.check_and_record_resource_group_members(module);
                 self.check_and_record_view_functions(module);
                 self.check_entry_functions(module);
+                self.check_events(module);
```
lightmark, line 75: should we do both check and record?

lightmark:
> > Thanks @wrwg . My understanding is that the check will be called when compiling the code but not will not be verified by verifier. So if someone bypass compilation it is still possible to emit `T` w/o `#[event]` attribute, right? Does this mean we have to either have similar logic in verifier or runtime native functions?
> 
> This check can be applied during module publishing: https://github.com/aptos-labs/aptos-core/blob/main/aptos-move/aptos-vm/src/aptos_vm.rs#L994

So the final static check is at module publishing phase, basically repeat the logic of this PR?

brmataptos:
Note this, however: https://github.com/aptos-labs/aptos-core/issues/9251, though it's not impossible that I'm mistaken.

wrwg:
> > > Thanks @wrwg . My understanding is that the check will be called when compiling the code but not will not be verified by verifier. So if someone bypass compilation it is still possible to emit `T` w/o `#[event]` attribute, right? Does this mean we have to either have similar logic in verifier or runtime native functions?
> > 
> > 
> > This check can be applied during module publishing: https://github.com/aptos-labs/aptos-core/blob/main/aptos-move/aptos-vm/src/aptos_vm.rs#L994
> 
> So the final static check is at module publishing phase, basically repeat the logic of this PR?

Yes, that is what we also discussed previously. However, I'm not convinced that we need to do any runtime check here. Keep in mind that runtime code verification is all about safety. So what specifically is unsafe about calling emit with a struct which does not have the event attribute? Just wondering.

lightmark:
> So what specifically is unsafe about calling emit with a struct which does not have the event attribute? Just wondering.
I think it is not about safety but following our event emission protocol. So I finally didn't do any native function runtime check but only check that in module publishing request as what resource group did. I believe it reaches a balance between complexity, performance hit and protocol robustness.

davidiw:
Closing out, as we now have the code being used in a functioning PR. https://github.com/aptos-labs/aptos-core/pulls?q=is%3Apr+is%3Aopen+events

PR #9510, opened 2023-08-05
[Execution][Sharding] Add support for global executor in sharding

sitalkedia:
### Description

- Adds support for global execution in sharding. On a high level a `partition_last_round` flag has been added to the partitioner, if it is false, then the last round is added as a global round.  

- Cleaned up the partitioning API to returned `PartitionedTransactions`. This holds both partitioned txns and global txns. 

- Added support for the global executor in the local execution client. 

### Test Plan
UTs and ran the benchmark.  I am not making the global executor as default as I see around 5% TPS regression with this enabled by default.

### DIFF
--- a/aptos-move/aptos-vm/src/sharded_block_executor/local_executor_shard.rs
+++ b/aptos-move/aptos-vm/src/sharded_block_executor/local_executor_shard.rs
@@ -21,6 +21,7 @@ use aptos_types::{
 use crossbeam_channel::{unbounded, Receiver, Sender};
 use move_core_types::vm_status::VMStatus;
 use std::{sync::Arc, thread};
+use crate::sharded_block_executor::counters::WAIT_FOR_SHARDED_OUTPUT_SECONDS;
 
 /// Executor service that runs on local machine and waits for commands from the coordinator and executes
 /// them in parallel.
@@ -146,6 +147,7 @@ impl<S: StateView + Sync + Send + 'static> LocalExecutorClient<S> {
     }
 
     fn get_output_from_shards(&self) -> Result<Vec<Vec<Vec<TransactionOutput>>>, VMStatus> {
+        let _timer = WAIT_FOR_SHARDED_OUTPUT_SECONDS.start_timer();
         trace!("LocalExecutorClient Waiting for results");
         let mut results = vec![];
         for (i, rx) in self.result_rxs.iter().enumerate() {
--- a/aptos-move/aptos-vm/src/sharded_block_executor/local_executor_shard.rs
+++ b/aptos-move/aptos-vm/src/sharded_block_executor/local_executor_shard.rs
@@ -3,6 +3,7 @@
 
 use crate::sharded_block_executor::{
     coordinator_client::CoordinatorClient,
+    counters::WAIT_FOR_SHARDED_OUTPUT_SECONDS,
     cross_shard_client::CrossShardClient,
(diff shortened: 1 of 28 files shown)


aptos-move/aptos-vm/src/sharded_block_executor/local_executor_shard.rs
```diff
@@ -145,18 +184,20 @@ impl<S: StateView + Sync + Send + 'static> ExecutorClient<S> for LocalExecutorCl
...
-            results.push(
-                rx.recv()
-                    .unwrap_or_else(|_| panic!("Did not receive output from shard {}", i))?,
-            );
-        }
-        Ok(results)
+        // This means that we are executing the global transactions concurrently with the individual shards but the
+        // global transactions will be blocked for cross shard transaction results. This hopefully will help with
+        // finishing the global transactions faster but we need to evaluate if this causes thread contention. If it
+        // does, then we can simply move this call to the end of the function.
```
zjma, line 191: so it's effectively a special shard...

Could it be too early to make a such a formal change to introduce global round/shard, before we are sure it's needed...?

execution/executor-benchmark/src/block_partitioning.rs
```diff
@@ -19,11 +18,17 @@ pub(crate) struct BlockPartitioningStage {
 }
 
 impl BlockPartitioningStage {
-    pub fn new(num_shards: usize) -> Self {
+    pub fn new(num_shards: usize, partition_last_round: bool) -> Self {
```
zjma, line 21: i think `partition_last_round` is a param very specific to V1/V2. It makes more sense to either pass in a pointer of `<dyn BlockPartitioner>`, or, inside the new() function, `build_from_envvar()`.

execution/executor-benchmark/src/main.rs
```diff
@@ -96,6 +96,8 @@ pub struct PipelineOpt {
     num_executor_shards: usize,
     #[clap(long)]
     async_partitioning: bool,
+    #[clap(long)]
+    use_global_executor: bool,
```
zjma, line 100: (the same idea as above) This is very specific to V1/V2 and should be inside some "MultiRoundPartitionerConfig" rather than a direct member of `PipelineConfig`.

zjma, line 100: (same as above)

aptos-move/aptos-vm/src/sharded_block_executor/local_executor_shard.rs
sitalkedia: @zjma - this will be needed sooner than later. For non-P txns for example, it might be tricky or very inefficient to shard them. So we can put them in a global shard.

sitalkedia: Also, this is optional, so we can run various tests to see the performance before making this as a default.

execution/executor-benchmark/src/main.rs
sitalkedia: Good point. How about I take this as a follow up to clean up/reorganize some configs related to sharding in executor benchmark?

execution/executor-benchmark/src/block_partitioning.rs
sitalkedia: How about this as a follow up as well?

zjma (approved):

msmouse (approved):

Issue #7351, opened 2023-03-23
[Bug] Python SDK is out of date on the PyPI


alnoki:
The Python SDK is at version `0.6.0` ([ref 1](https://github.com/aptos-labs/aptos-core/blob/main/ecosystem/python/sdk/setup.py#L22), [ref 2](https://github.com/aptos-labs/aptos-core/blob/154e450a77c593e93bbd1ba224bb2fd532d4ffb8/ecosystem/python/sdk/pyproject.toml#L3)), but the PyPI reflects [0.5.1](https://pypi.org/project/aptos-sdk/).

@davidiw @banool is the PyPI release process automated?

dannysheridan:
@alnoki if you want to use a publishing pipeline to keep PyPI up to date, check out our open source project: www.buildwithfern.com

banool:
It is not currently automated, we can look at making a release this week.

alnoki:
Bump

banool:
Just released a new version (0.6.1) and included publishing instructions and the like: https://github.com/aptos-labs/aptos-core/pull/8149.

dannysheridan:
@banool are you interested in automating the generation and publishing of the Python SDK? here's [a python SDK](https://github.com/vellum-ai/vellum-client-python) where the team uses Fern.

banool:
Hi, I've previously spoken with someone from Fern at length and we decided that Fern was not a good fit for our needs, given we're moving away from the OpenAPI stack.

davidiw:
tonight!

PR #9512, opened 2023-08-06
Feat/pypi ci

zhenfeng-zhu:
### Description

This is for issue https://github.com/aptos-labs/aptos-core/issues/7351

### DIFF
--- a/.github/workflows/python-sdk-publish.yaml
+++ b/.github/workflows/python-sdk-publish.yaml
@@ -2,7 +2,6 @@ name: "Run Python SDK Publish"
 
 on:
   workflow_dispatch:
-  push: 
 
 jobs:
   release:
@@ -18,12 +17,6 @@ jobs:
 
       - name: Build project for distribution
         run: poetry build
-        working-directory: ./ecosystem/python/sdk
-
-      - name: Check Version
-        id: check-version
-        run: |
-          [[ "$(poetry version --short)" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]] && [[ $(grep $(poetry version --short) CHANGELOG.md) =~ $(poetry version --short)  ]] || echo prerelease=true >> $GITHUB_OUTPUT
         working-directory: ./ecosystem/python/sdk
 
       - name: Publish to PyPI
--- a/.github/workflows/python-sdk-publish.yaml
+++ b/.github/workflows/python-sdk-publish.yaml
@@ -1,0 +1,33 @@
+name: "Run Python SDK Publish"
+
+on:
+  workflow_dispatch:
+  push: 
+
+jobs:
+  release:
+    name: Release
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      - uses: ./.github/actions/python-setup
+        with:
+          pyproject_directory: ./ecosystem/python/sdk
+
+      - name: Build project for distribution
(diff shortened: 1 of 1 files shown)


gedigi (approved):

.github/workflows/python_sdk_publish.yaml
```diff
@@ -0,0 +1,37 @@
+name: Release
+
+on:
```
banool, line 3: I don't think our current Python SDK testing process is strict enough for this to run on every push right now. Could we make this work on workflow_dispatch instead for now?

.github/workflows/python_sdk_publish.yaml
```diff
@@ -0,0 +1,37 @@
...
+        run: echo "$HOME/.local/bin" >> $GITHUB_PATH
+
+      - name: Build project for distribution
+        run: cd ./ecosystem/python/sdk && poetry build
+
+      - name: Check Version
+        id: check-version
+        run: |
+          [[ "$(poetry version --short)" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]] || echo prerelease=true >> $GITHUB_OUTPUT
+
```
banool, line 33: Ideally we would also check that there is an entry in the CHANGELOG for this new version.

banool (changes requested):
Thanks for the PR! Requesting changes for the workflow trigger.

.github/workflows/python_sdk_publish.yaml
```diff
@@ -0,0 +1,37 @@
...
+        uses: actions/checkout@v3
+
+      - name: Set up Python 3.10
+        uses: actions/setup-python@v4
+        with:
+          python-version: "3.10"
+
+      - name: Install Poetry
+        run: |
+          curl -sSL https://install.python-poetry.org | python - -y
```
banool, line 21: Let's use `actions/python-setup/action.yaml` instead, it sets up the correct Python version as configured in the repo, installs Poetry, etc.

zhenfeng-zhu:
Thanks for all your help, I have resolved some comment. Please review it!

.github/workflows/python-sdk-publish.yaml
```diff
@@ -0,0 +1,33 @@
+name: "Run Python SDK Publish"
+
+on:
+  workflow_dispatch:
+  push:
```
rustielin, line 5: we don't need the `push` trigger, can we just do it for `workflow_dispatch`?

.github/workflows/python-sdk-publish.yaml
```diff
@@ -0,0 +1,33 @@
...
+          pyproject_directory: ./ecosystem/python/sdk
+
+      - name: Build project for distribution
+        run: poetry build
+        working-directory: ./ecosystem/python/sdk
+
+      - name: Check Version
+        id: check-version
+        run: |
+          [[ "$(poetry version --short)" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]] && [[ $(grep $(poetry version --short) CHANGELOG.md) =~ $(poetry version --short)  ]] || echo prerelease=true >> $GITHUB_OUTPUT
```
rustielin, line 26: where is this github output being used?

rustielin (changes requested):

banool (approved):
Looks good, thanks for changing to workflow_dispatch!!

sionescu (approved):

Issue #8710, opened 2023-06-16
[Feature Request] Support for generating signed messages in SDK and CLI


movekevin:
It'd be great to have support in SDK and CLI for generating signed messages (signing over blobs of data that represent Move structs with type info). This is a very common use case, for example framework's account key rotation, rotation cap offering, signer cap offering, and ecosystem uses of authentication via signed messages.

Example: https://github.com/aptos-labs/aptos-core/blob/main/aptos-move/framework/aptos-framework/sources/account.move#L385

xbtmatt:
cc @movekevin 

Been thinking about this a bit lately, is it important to have the ability to sign a random struct? Or can we offer the ability to sign for specific structs?

The TS SDK can actually sign whatever, but there are security implications with the wallet adapter signing any arbitrary message, so unless there's a big need for signing structs defined in arbitrary, non-framework modules, maybe it makes the most sense to just offer the ability to sign structs for the specific functions that need them?

PR #9516, opened 2023-08-07
[TS SDK example] Add a working example of converting a MultiEd25519 account to a MultiSig account

xbtmatt:
As a part of an effort to address https://github.com/aptos-labs/aptos-core/issues/8710, this is the second example of signed structs with the Typescript SDK

### Description

**This example demonstrates how to convert a MultiEd25519 account to a MultiSig account and revoke its auth key using the `0x1::multisig_account` module.**

1. Initialize N accounts and fund them.
2. Initialize a MultiEd25519 account with the created accounts as owners and a signature threshold K as NUM_SIGNATURES_REQUIRED
      - See: https://aptos.dev/concepts/accounts/#multi-signer-authentication for more information on MultiEd25519 accounts.
3. Create a proof struct for at minimum K of the N accounts to sign.
4. Gather the signatures from the accounts.
5. Assemble a MultiEd25519 signed proof struct with the gathered signatures.
6. Call the `0x1::multisig_account::create_with_existing_account_and_revoke_auth_key` function with the assembled proof struct and other logistical information
      - Because the function requires a signed proof by the MultiEd25519 account, it does not require or check the signer, meaning anyone can submit the transaction with the proof struct.
      - We submit it as a randomly generated account here to convey this.
7. The transaction will be executed and the following occurs on chain:
      a. The MultiEd25519 account is converted into a MultiSig account.
      b. The resulting account can from then on be used as a MultiSig account, potentially with new owners and/or a new minimum signature threshold.
      c. The original MultiEd25519 account has its authentication key rotated, handing over control to the `0x1::multisig_account` contract.

### Test Plan

Run the script and observe the output. The authentication key, sequence number, rotation capability offer, and signer capability offer are checked at the end.

```shell
cd ~/aptos-core/ecosystem/typescript/sdk/examples/typescript-esm
pnpm install
pnpm run multi_ed25519_to_multisig
```

### DIFF
--- a/ecosystem/typescript/sdk/examples/typescript-esm/multi_ed25519_to_multisig.ts
+++ b/ecosystem/typescript/sdk/examples/typescript-esm/multi_ed25519_to_multisig.ts
@@ -14,16 +14,32 @@ import assert from "assert";
 const ED25519_ACCOUNT_SCHEME = 0;
 const MULTI_ED25519_ACCOUNT_SCHEME = 1;
 
-type MultiSigAccountCreationWithAuthKeyRevocationMessage = {
-  moduleAddress: TxnBuilderTypes.AccountAddress;
-  moduleName: string;
-  structName: string;
-  chainId: number;
-  multiSigAddress: TxnBuilderTypes.AccountAddress;
-  sequenceNumber: number;
-  owners: Array<TxnBuilderTypes.AccountAddress>;
-  numSignaturesRequired: number;
-};
+class MultiSigAccountCreationWithAuthKeyRevocationMessage {
+  public readonly moduleAddress: TxnBuilderTypes.AccountAddress = TxnBuilderTypes.AccountAddress.CORE_CODE_ADDRESS;
+  public readonly moduleName: string = "multisig_account";
+  public readonly structName: string = "MultisigAccountCreationWithAuthKeyRevocationMessage";
+  public readonly functionName: string = "create_with_existing_account_and_revoke_auth_key";
+
+  constructor(
+    public readonly chainId: number,
+    public readonly multiSigAddress: TxnBuilderTypes.AccountAddress,
+    public readonly sequenceNumber: number,
+    public readonly owners: Array<TxnBuilderTypes.AccountAddress>,
+    public readonly numSignaturesRequired: number,
+  ) {}
+
+  serialize(serializer: BCS.Serializer): void {
+    this.moduleAddress.serialize(serializer);
+    serializer.serializeStr(this.moduleName);
+    serializer.serializeStr(this.structName);
+    serializer.serializeU8(this.chainId);
+    this.multiSigAddress.serialize(serializer);
+    serializer.serializeU64(this.sequenceNumber);
+    serializer.serializeU32AsUleb128(this.owners.length);
+    this.owners.forEach((owner) => owner.serialize(serializer));
+    serializer.serializeU64(this.numSignaturesRequired);
+  }
+}
 
 //////////////////////////////////////////////////////////////////////////////////////////
 //////////////////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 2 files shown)


ecosystem/typescript/sdk/examples/typescript-esm/multi_ed25519_to_multisig.ts
```diff
@@ -0,0 +1,437 @@
...
+  const NUM_SIGNATURES_REQUIRED = 3;
+
+  // Step 1.
+  // Initialize N accounts and fund them. See: https://aptos.dev/concepts/accounts/#multi-signer-authentication
+  // Works with any # of addresses, you just need to change NUM_SIGNATURES_REQUIRED and the signingAddresses
+  const account1 = new AptosAccount();
+  const account2 = new AptosAccount();
+  const account3 = new AptosAccount();
+
+  await faucetClient.fundAccount(account1.address(), 100_000_000);
```
movekevin, line 66: We really need a oneApt() library function in the SDK that returns 10^8 :)

ecosystem/typescript/sdk/examples/typescript-esm/multi_ed25519_to_multisig.ts
```diff
@@ -0,0 +1,437 @@
...
+//////////////////////////////////////////////////////////////////////////////////////////
+//////////////////////////////////////////////////////////////////////////////////////////
+
+// Funds and thus creates the derived MultiEd25519 account and prints out the derived address, authentication key, and public key.
+const initializeMultiEd25519 = async (
+  faucetClient: FaucetClient,
+  accounts: Array<AptosAccount>,
+  numSignaturesRequired: number,
+): Promise<Array<string>> => {
+  const multiSigPublicKey = new TxnBuilderTypes.MultiEd25519PublicKey(
```
movekevin, line 227: Let's add some comments here that creates an account with multiEd25519 auth key requires the public keys, which are different from account addresses. This has always been a very confusing detail for both devs and users as there's no easy way to get the public keys - some wallets don't really show them either.

ecosystem/typescript/sdk/examples/typescript-esm/multi_ed25519_to_multisig.ts
```diff
@@ -0,0 +1,437 @@
...
+      expireTimestamp: BigInt(Math.floor(Date.now() / 1000) + 60),
+    });
+
+    // Step 7.
+    // Wait for the transaction to complete, then observe and assert that the authentication key for the original MultiEd25519 account
+    // has been rotated and all capabilities revoked.
+    const txnInfo = await provider.waitForTransactionWithResult(txn);
+    printRelevantTxInfo(txnInfo as Types.UserTransaction);
+    assertChangesAndPrint(provider, multiSigAddress, sequenceNumber, false, metadataKeys, metadataValues);
+  } else {
```
movekevin, line 161: Is this option 2 really useful? Once the challenge message has been signed by enough signatures to form the multied25519 key, it's better submitted by a single account? The only use case I ever see here is to not have to pay gas, but gas is so low, it doesn't really matter. Probably costs cents to do the migration

movekevin (approved):
This is amazing work!!!!

ecosystem/typescript/sdk/examples/typescript-esm/multi_ed25519_to_multisig.ts
xbtmatt: No, it's not really, haha. It was more just to show how to sign a transaction that needs a proof with a multied account.

Lowkey I didn't realize until the very end that the auth key revocation function in multisig_account.move doesn't need a `&signer`, but I was so far into it I just finished the example as if it needed it

ecosystem/typescript/sdk/examples/typescript-esm/multi_ed25519_to_multisig.ts
```diff
@@ -0,0 +1,370 @@
...
+  const structSignatures = [structSig1, structSig2, structSig3].map((sig) => sig.toUint8Array());
+
+  // This is the bitmap indicating which original owners are present as signers. It takes a diff of the original owners and the signing owners and creates the bitmap based on that.
+  const bitmap = createBitmapFromDiff(accountAddresses, signingAddresses);
+
+  // Step 5.
+  // Assemble a MultiEd25519 signed proof struct with the gathered signatures.
+  // This represents the multisig signed struct by all owners. This is used as proof of authentication by the overall MultiEd25519 account, since there's no signer
+  // checked in the entry function we're using.
+  const multiSigStruct = createMultiSigStructFromSignedStructs(structSignatures, bitmap);
```
davidiw, line 131: these really should be a single call

ecosystem/typescript/sdk/examples/typescript-esm/package.json
```diff
@@ -7,8 +7,13 @@
   "scripts": {
     "build": "rm -rf dist/* && tsc -p .",
     "test": "pnpm build && node ./dist/index.js",
+<<<<<<< HEAD
     "offer_capabilities": "ts-node --esm offer_capabilities.ts",
     "rotate_key": "ts-node --esm rotate_key.ts"
+=======
+    "rotate_key": "ts-node --esm rotate_key.ts",
+    "multi_ed25519_to_multisig": "ts-node --esm multi_ed25519_to_multisig.ts"
+>>>>>>> 1aafc62c48 (An example of converting a MultiEd25519 account to a MultiSig account)
```
davidiw, line 16: yucky

davidiw (commented):
overall looks fine, but curious if we could make this a library :P
and a little more ... user-friendly via a single call to construct the signature

xbtmatt:
> overall looks fine, but curious if we could make this a library :P
and a little more ... user-friendly via a single call to construct the signature

yea, I mainly split things up to try to convey what the actual flow would be in a dapp where you're waiting on users to sign the different messages, but maybe it ended up just being more confusing...

ecosystem/typescript/sdk/examples/typescript-esm/package.json
xbtmatt: that's embarrassing...

davidiw (approved):

PR #9518, opened 2023-08-07
[Prover] Revive the Prover lab

junkil-park:
### Description
The Prover lab is a benchmarking infrastructure for Prover, which was not working because it hasn't been used and updated for a while.

This PR revives the Prover lab to investigate the timeout issues that the Prover currently has.

To make it work again, this PR updates the Prover lab to adapt the new APIs of `clap` and to adapt the recent change in Prover (e.g., proper initialization with `create_init_num_operation_state`).

Tested the Prover lab with the simple data as well as with the diem framework.

### Test Plan
In a test directory:
`./run.sh`
`./plot.sh`


rahxephon89 (approved):
LGTM, thanks!

aalok-t (approved):

wrwg (approved):

wrwg:
Would it make sense to check in together with this change the new_boogie experiment we reviewed today, including the generated diagrams?

junkil-park:
> Would it make sense to check in together with this change the new_boogie experiment we reviewed today, including the generated diagrams?

Yes, I am checking them in.

PR #9520, opened 2023-08-07
[API] Return 410 when txn has been pruned, not 500

banool:
### Description
A missing txn is not an internal server error, 410 is a more appropriate error code.

The long term fix is to make storage return typed errors, but this is a solid fix for now.

### Test Plan
Run a local testnet with a modified config where the pruning window is much shorter:
```
cargo run -p aptos -- node run-local-testnet --force-restart --assume-yes --with-faucet --test-config-override ~/myfullnode.yaml
```

Config: https://gist.github.com/banool/7bf5f89cbd9ab999b16eb641413902a8.

Let it run for a while.

See that now instead of it being a 500, it is a 410 when a txn is missing:
```
curl -v localhost:8080/v1/transactions/by_version/1
```

Before:
```
< HTTP/1.1 500 Internal Server Error
{"message":"Failed to get transaction by version 1: Transaction at version 1 is pruned, min available version is 403.","error_code":"internal_error","vm_error_code":null}
```

After:
```
< HTTP/1.1 410 Gone
{"message":"Ledger version(1) has been pruned","error_code":"version_pruned","vm_error_code":null}
```

Unit testing:
```
cd api
cargo test
```

### DIFF
--- a/api/src/transactions.rs
+++ b/api/src/transactions.rs
@@ -12,8 +12,8 @@ use crate::{
     page::Page,
     response::{
         api_disabled, api_forbidden, transaction_not_found_by_hash,
-        transaction_not_found_by_version, BadRequestError, BasicError, BasicErrorWith404,
-        BasicResponse, BasicResponseStatus, BasicResult, BasicResultWith404,
+        transaction_not_found_by_version, version_pruned, BadRequestError, BasicError,
+        BasicErrorWith404, BasicResponse, BasicResponseStatus, BasicResult, BasicResultWith404,
         InsufficientStorageError, InternalError,
     },
     ApiTags,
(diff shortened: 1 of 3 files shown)


sitalkedia (approved):
LGTM

sitalkedia:
I think there is still a race condition when the transaction being queried is right on the pruning window, the current check will pass but the DB might return an error. Do we need to be worried about that case or current fix is good enough?

banool:
I think that case should be rare enough that we shouldn't worry about it for now.

banool:
Issue to track making this use 410 for when storage eventually has nicely typed responses: https://github.com/aptos-labs/aptos-core/issues/9523.

api/src/transactions.rs
```diff
@@ -770,6 +770,9 @@ impl TransactionsApi {
         if version > ledger_info.version() {
             return Ok(None);
         }
+        if version < ledger_info.oldest_version() {
+            return Ok(None);
+        }
```
gregnazario, line 775: This should work in most cases.  Remembering that it is possible for the version being used to be pruned prior to processing it (there is an offset though, so it should be fine).

gregnazario (approved):

banool:
Update, I made it return 410 if the version has been pruned. I have confirmed that if you query for a version that the node hasn't synced to yet, it still returns 404.

```
< HTTP/1.1 410 Gone
{"message":"Ledger version(1) has been pruned","error_code":"version_pruned","vm_error_code":null}
```

api/src/transactions.rs
```diff
@@ -1325,3 +1331,9 @@ fn override_gas_parameters(
     // TODO: Check that signature is null, this would just be helpful for downstream use
     SignedTransaction::new_with_authenticator(raw_txn, signed_txn.authenticator())
 }
+
+enum GetByVersionResponse {
+    VersionTooNew,
+    VersionTooOld,
+    Found(TransactionData),
+}
```
gregnazario, line 1339: Note this should be cleaned up in the error handling updates for storage

PR #9522, opened 2023-08-07
add semgrep for github workflows

gedigi:
### Description
This PR creates a workflow for semgrep (semgrep.dev) that checks for insecure usage of `pull_request_target` in GitHub workflows.

### Test Plan
The workflow can be triggered/tested manually and automatically run at every code change.

As a test I removed a "permission-check" from one of the workflows and the workflow caught it: https://github.com/aptos-labs/aptos-core/actions/runs/5801417544/job/15725640542


rustielin (commented):
The approach LGTM. Can you add a GHA workflow run link to your test plan that runs the actual workflow (e.g. a canary run that adds a new `pull_request_target`)?

rustielin (approved):

PR #9524, opened 2023-08-07
[prover] lock version dependency for z3 and boogie (#8718)

aalok-t:
### Description

The upgrade of Z3 and Boogie is mostly backward compatible, but not always. This fix allows Prover to warn users for the higher version of Z3 and Boogie because those may be incompatible (refer: https://github.com/aptos-labs/aptos-core/issues/8718)

### Test Plan

Tested on Mac Arm64 with no current boogie or Z3 installation
Code is visibly unchanged for other cases

### DIFF
--- a/third_party/move/move-prover/boogie-backend/src/options.rs
+++ b/third_party/move/move-prover/boogie-backend/src/options.rs
@@ -19,9 +19,17 @@ const DEFAULT_BOOGIE_FLAGS: &[&str] = &[
     "-proverOpt:O:model_validate=true",
 ];
 
-const MIN_BOOGIE_VERSION: &str = "2.15.8";
-const MIN_Z3_VERSION: &str = "4.11.0";
-const MIN_CVC5_VERSION: &str = "0.0.3";
+/// Versions for boogie, z3, and cvc5. The upgrade of boogie and z3 is mostly backward compatible,
+/// but not always. Setting the max version allows Prover to warn users for the higher version of
+/// boogie and z3 because those may be incompatible.
+const MIN_BOOGIE_VERSION: Option<&str> = Some("2.15.8.0");
+const MAX_BOOGIE_VERSION: Option<&str> = Some("2.15.8.0");
+
+const MIN_Z3_VERSION: Option<&str> = Some("4.11.2");
+const MAX_Z3_VERSION: Option<&str> = Some("4.11.2");
+
+const MIN_CVC5_VERSION: Option<&str> = Some("0.0.3");
+const MAX_CVC5_VERSION: Option<&str> = None;
 
 #[derive(Debug, Clone, Copy, Serialize, Deserialize)]
 pub enum VectorTheory {
@@ -306,17 +314,27 @@ impl BoogieOptions {
                 version_arg,
                 r"version ([0-9.]*)",
             )?;
-            Self::check_version_is_greater("boogie", &version, MIN_BOOGIE_VERSION)?;
+            Self::check_version_is_compatible(
+                "boogie",
+                &version,
+                MIN_BOOGIE_VERSION,
+                MAX_BOOGIE_VERSION,
+            )?;
         }
         if !self.z3_exe.is_empty() && !self.use_cvc5 {
             let version =
                 Self::get_version("z3", &self.z3_exe, &["--version"], r"version ([0-9.]*)")?;
-            Self::check_version_is_greater("z3", &version, MIN_Z3_VERSION)?;
+            Self::check_version_is_compatible("z3", &version, MIN_Z3_VERSION, MAX_Z3_VERSION)?;
         }
         if !self.cvc5_exe.is_empty() && self.use_cvc5 {
             let version =
                 Self::get_version("cvc5", &self.cvc5_exe, &["--version"], r"version ([0-9.]*)")?;
-            Self::check_version_is_greater("cvc5", &version, MIN_CVC5_VERSION)?;
+            Self::check_version_is_compatible(
+                "cvc5",
+                &version,
+                MIN_CVC5_VERSION,
+                MAX_CVC5_VERSION,
+            )?;
         }
         Ok(())
     }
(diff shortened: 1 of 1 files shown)


third_party/move/move-prover/boogie-backend/src/options.rs
```diff
@@ -20,8 +20,13 @@ const DEFAULT_BOOGIE_FLAGS: &[&str] = &[
 ];
 
 const MIN_BOOGIE_VERSION: &str = "2.15.8";
+const MAX_BOOGIE_VERSION: &str = "2.15.8";
+
 const MIN_Z3_VERSION: &str = "4.11.0";
+const MAX_Z3_VERSION: &str = "4.11.0";
+
 const MIN_CVC5_VERSION: &str = "0.0.3";
+const MAX_CVC5_VERSION: &str = "0.0.3";
```
rahxephon89, line 29: This one caused an linter error: https://github.com/aptos-labs/aptos-core/actions/runs/5789663535/job/15691166679 You could remove it since it is not used.

third_party/move/move-prover/boogie-backend/src/options.rs
```diff
@@ -365,4 +372,32 @@ impl BoogieOptions {
         }
         Ok(())
     }
+    fn check_version_is_leq(tool: &str, given: &str, expected: &str) -> anyhow::Result<()> {
```
wrwg, line 375: Can we avoid this code duplication? This seems to do 95% the same as `check_version_is_greater`. Also, as you are as with this, rename `_greater` into `_geq` for consistency with the new function.

third_party/move/move-prover/boogie-backend/src/options.rs
aalok-t: Removed it. Thank you.

third_party/move/move-prover/boogie-backend/src/options.rs
aalok-t: I have created a separate function for string compatibility and renamed functions for consistency. The functions `check_version_is_geq` and `check_version_is_leq` are still pretty similar. I am not sure if I should further simplify.

rahxephon89: @aalok-t Maybe you could try to use macro rules to avoid duplication further?

rahxephon89: @aalok-t Or does it make sense to call `check_version_is_geq` two times in sequence with different arguments: 
Self::check_version_is_geq("z3", &version, MIN_Z3_VERSION)?;
Self::check_version_is_geq("z3", MAX_Z3_VERSION, &version,)?;

aalok-t: One cannot directly do that as the error messages are slightly different. I have effectively done that in my update and will share it soon.

third_party/move/move-prover/boogie-backend/src/options.rs
```diff
@@ -340,31 +363,73 @@ impl BoogieOptions {
...
+            let bn = b.parse::<usize>().unwrap();
+            if an < bn {
+                result = false;
+                break;
+            }
+        }
+        result
+    }
+
+    fn check_version_op(tool: &str, given: &str, expected: &str, op: &str) -> anyhow::Result<()> {
```
brmataptos, line 395: It is generally a bad idea in most languages to use a string (or string slice, here) to select different behavior from a function.  In Rust, it's easy enough to use an enum:
```
enum Op { Max, Min }
```
and then pass Op::Max or Op::Min instead of a string "max" or "min".

Though: to minimize code, since these really only affect the ordering used to compare parameters and the string output with the error case you can separate comparison parameters from output parameters and use a method with a signature like:
```
  fn check_version_le(
        expect_lesser: &str,
        expect_greater: &str,
        relative_term: &str,
        expected: &str,
        given: &str,
        tool: &str,
    ) -> anyhow::Result<()> {
        ...
        return Err(anyhow!("expected at {} version {} but found {} for `{}`",
            relative_term, expected, given, tool);
        ...
}
```
and call it with
```
check_version_le(min, given, "least", min, given, tool)
```
and
```
check_version_le(given, max, "most", max, given, tool)
```

third_party/move/move-prover/boogie-backend/src/options.rs
```diff
@@ -306,17 +314,32 @@ impl BoogieOptions {
                 version_arg,
                 r"version ([0-9.]*)",
             )?;
-            Self::check_version_is_greater("boogie", &version, MIN_BOOGIE_VERSION)?;
+            Self::check_version_comparable(
```
brmataptos, line 317: Suggest changing this method name to "check_version_is_compatible"

third_party/move/move-prover/boogie-backend/src/options.rs
```diff
@@ -19,9 +19,22 @@ const DEFAULT_BOOGIE_FLAGS: &[&str] = &[
...
+const MIN_BOOGIE_VERSION: Option<&str> = Some("2.15.8");
+const MAX_BOOGIE_VERSION: Option<&str> = Some("2.15.8");
+
+const MIN_Z3_VERSION: Option<&str> = Some("4.11.0");
+const MAX_Z3_VERSION: Option<&str> = Some("4.11.0");
+
+const MIN_CVC5_VERSION: Option<&str> = Some("0.0.3");
+const MAX_CVC5_VERSION: Option<&str> = None;
+
+enum Op {
```
rahxephon89, line 34: As also mentioned by @brmataptos, to minimize the code change, you actually don't need this enum to distinguish two cases.

third_party/move/move-prover/boogie-backend/src/options.rs
```diff
@@ -19,9 +19,17 @@ const DEFAULT_BOOGIE_FLAGS: &[&str] = &[
...
+/// but not always. Setting the max version allows Prover to warn users for the higher version of
+/// boogie and z3 because those may be incompatible.
+const MIN_BOOGIE_VERSION: Option<&str> = Some("2.15.8");
+const MAX_BOOGIE_VERSION: Option<&str> = Some("2.15.8");
+
+const MIN_Z3_VERSION: Option<&str> = Some("4.11.0");
+const MAX_Z3_VERSION: Option<&str> = Some("4.11.0");
+
+const MIN_CVC5_VERSION: Option<&str> = Some("0.0.3");
+const MAX_CVC5_VERSION: Option<&str> = None;
```
aalok-t, line 32: Using Option type to allow cases where the min version/max version may not be specified.

third_party/move/move-prover/boogie-backend/src/options.rs
aalok-t: Thanks Brian. I have updated the function accordingly.

third_party/move/move-prover/boogie-backend/src/options.rs
aalok-t: I have removed enum.

rahxephon89 (approved):
LGTM, thanks!

third_party/move/move-prover/boogie-backend/src/options.rs
```diff
@@ -340,29 +358,55 @@ impl BoogieOptions {
...
+            Self::check_version_le(expected, given, "least", expected, given, tool)?;
+        }
+        if let Some(expected) = expected_max {
+            Self::check_version_le(given, expected, "most", expected, given, tool)?;
+        }
+        Ok(())
+    }
+
+    // This function checks if expected_lesser is actually less than or equal to expected_greater
+    fn check_version_le(
```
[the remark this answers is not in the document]
brmataptos: Perhaps add that explanation as a comment?

wrwg (approved):

brmataptos:
Note that you should "git rebase main" rather than "git merge main" so that you get a cleaner PR in the end.

I'm happy with this but would prefer an extra comment.

brmataptos (approved):

rahxephon89:
@aalok-t You need to set the version to "4.11.2". This is the info from the failed unit test: https://github.com/aptos-labs/aptos-core/actions/runs/5838031208/job/15834487894?pr=9524

PR #9526, opened 2023-08-07
[framework] Add Aptos-token mint function that returns object

movekevin:
### Description
Cherrypicking https://github.com/aptos-labs/aptos-core/commit/78fb3739692a45235bd1b59ea7ed1a71fab00959 into v1.6


gregnazario (approved):

igor-aptos:
thanks, looks good

PR #9528, opened 2023-08-07
[Spec] Fix spec for daily prover test

rahxephon89:
### Description

Need to turn off verification of `create_stake_pool` to avoid timeout of inconsistency check in the daily prover test job.


aalok-t (approved):

junkil-park (approved):
Thanks!

PR #9534, opened 2023-08-07
brian/typo

bchocho:
- Fix typo in sweep
- Update PFN realistic env 100 TPS test, move to forge stable. Remove other PFN TPS tests from forge PFN

### DIFF
--- a/.github/workflows/forge-pfn.yaml
+++ b/.github/workflows/forge-pfn.yaml
@@ -95,49 +95,10 @@ jobs:
 
   ### Public fullnode tests
 
-  # Measures PFN latencies with a constant TPS
-  run-forge-pfn-const-tps:
-    if: ${{ github.event_name != 'pull_request' }}
-    needs: determine-test-metadata
-    uses: aptos-labs/aptos-core/.github/workflows/workflow-run-forge.yaml@main
-    secrets: inherit
-    with:
-      IMAGE_TAG: ${{ needs.determine-test-metadata.outputs.IMAGE_TAG }}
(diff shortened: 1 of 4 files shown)


testsuite/testcases/src/load_vs_perf_benchmark.rs
```diff
@@ -218,7 +218,7 @@ fn to_table(results: &[SingleRunStats]) -> Vec<String> {
         "batch->pos",
         "pos->prop",
         "prop->order",
-        "prop->commit",
```
bchocho, line 221: This is unrelated to the main PR, but it looks like a typo. cc: @igor-aptos

.github/workflows/forge-pfn.yaml
```diff
@@ -95,49 +95,10 @@ jobs:
 
   ### Public fullnode tests
 
-  # Measures PFN latencies with a constant TPS
```
JoshLind, line 98: Anyone reason we're removing these? One of the things I liked about them was that it was easy for us see/debug changes in performance/latency across tests (i.e., because we only change 1 variable at a time). Also, given that we're no longer regularly running the PFN jobs, it also doesn't hurt to keep these here, right?

JoshLind (approved):
Thanks @bchocho. LGTM! Only ask is if we can keep the const PFN tests in the `forge-pfn.yaml` file? 😄

.github/workflows/forge-pfn.yaml
bchocho: I though they were done with, but I'll revive them!

igor-aptos (approved):

PR #9536, opened 2023-08-07
[NFT Metadata Crawler] Parse chain id from Pubsub and prep config for deployment

just-in-chang:
### Description
Allows worker to parser chain id from Pubsub. Also set google account credentials to optional in config yaml for deployment because it is automatically set in cloud run instsances.

### DIFF
--- a/ecosystem/nft-metadata-crawler-parser/src/worker.rs
+++ b/ecosystem/nft-metadata-crawler-parser/src/worker.rs
@@ -40,7 +40,7 @@ use tracing::{error, info};
 #[derive(Clone, Debug, Deserialize, Serialize)]
 #[serde(deny_unknown_fields)]
 pub struct ParserConfig {
-    pub google_application_credentials: String,
+    pub google_application_credentials: Option<String>,
     pub bucket: String,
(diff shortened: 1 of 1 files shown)


ecosystem/nft-metadata-crawler-parser/src/worker.rs
```diff
@@ -214,6 +218,7 @@ impl Worker {
         token_uri: String,
         last_transaction_version: i32,
         last_transaction_timestamp: chrono::NaiveDateTime,
+        chain_id: i32,
```
bowenyang007, line 221: What's calling this?

just-in-chang: Nothing yet, will add db checks in following PR

bowenyang007 (approved):

ecosystem/nft-metadata-crawler-parser/src/worker.rs
bowenyang007: Can you add follow up PRs. will make this easier to understand

rtso (approved):

PR #9539, opened 2023-08-08
[NFT Metadata Crawler] Use separate DB schema

just-in-chang:
### Description
Move tables for the nft metadata crawler to its own postgres schema


bowenyang007:
@rtso btw this is a great example of how we can do a different schema in diesel.

bowenyang007 (approved):

rtso (approved):

rtso:
@bowenyang007 nice, i think this should be doable to split up for the current processor setup. each processor would need to define which schema its tables belong to so it knows which migration to apply on runtime.

PR #9542, opened 2023-08-08
Framework upgrade to include aptos_token changes

movekevin:


davidiw:
huh?

movekevin:
> huh?

Needed this to only release the framework to testnet (1.6.1) as all other featurea have already been enabled as part of 1.6.0

PR #9543, opened 2023-08-08
Sanitize pruner window configs

msmouse:
### Description

So that a user won't be able to accidentally set it to a super large number and fail requests.

### Test Plan
eyes.
I feel a unit test doesn't add much value.

### DIFF
--- a/config/src/config/storage_config.rs
+++ b/config/src/config/storage_config.rs
@@ -6,6 +6,7 @@ use crate::{
     config::{config_sanitizer::ConfigSanitizer, node_config_loader::NodeType, Error, NodeConfig},
     utils,
 };
+use aptos_logger::warn;
 use aptos_types::chain_id::ChainId;
 use serde::{Deserialize, Serialize};
 use std::{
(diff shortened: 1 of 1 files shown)


grao1991 (approved):

JoshLind (approved):
> I feel a unit test doesn't add much value.

LGTM, but disagree on the lack of tests 😄

config/src/config/storage_config.rs
```diff
@@ -291,11 +292,53 @@ impl StorageConfig {
...
+            .storage_pruner_config
+            .epoch_snapshot_pruner_config
+            .prune_window;
+        let user_pruning_window_offset = config
+            .storage_pruner_config
+            .ledger_pruner_config
+            .user_pruning_window_offset;
+
+        if ledger_prune_window < 50_000_000 {
+            warn!("Ledger prune window is too small, harming network data availability.");
```
banool, line 320: Nit: I'd probably just say `ledger_prune_window` rather than `Ledger prune window`.

banool (approved):

config/src/config/storage_config.rs
msmouse: there's no config called `ledger_prune_window`, I'll say "Ledger prune_window"

PR #9545, opened 2023-08-08
[GHA] Continue migrating away from non-required tests.

JoshLind:
### Description
This PR offers two changes (each in their own commit):
1. Avoid running non-required jobs on PRs unless the correct label is applied. This includes even if PR authors re-push their changes to a branch on the repository.
2. Remove/skip no longer needed tests. These tests are already covered in other flows (e.g., `general-lints`). Also, don't run the SDK tests on docs only changes.

### Test Plan
Existing test infrastructure.


perryjrandall (approved):

sionescu (approved):

.github/workflows/ts-sdk-e2e-tests.yaml
```diff
@@ -49,17 +62,20 @@ jobs:
           AWS_DOCKER_ARTIFACT_REPO: ${{ secrets.AWS_DOCKER_ARTIFACT_REPO }}
           GIT_CREDENTIALS: ${{ secrets.GIT_CREDENTIALS }}
       - uses: ./.github/actions/run-ts-sdk-e2e-tests
+        if: needs.file_change_determinator.outputs.only_docs_changed != 'true'
         with:
           BRANCH: main
           GCP_DOCKER_ARTIFACT_REPO: ${{ secrets.GCP_DOCKER_ARTIFACT_REPO }}
+      - run: echo "Skipping the tests on the main branch! Unrelated changes detected."
+        if: needs.file_change_determinator.outputs.only_docs_changed == 'true'
```
ibalajiarun, line 70: This is more of a note. I just realized that this will still trigger provisioning a VM (if one's not available), so we will still pay (non-zero cost) but only a fraction of the cost. We will still save big.

JoshLind: Yeah, great point! Ideally we could skip the entire job in the future (once we've reworked the entire flow), to avoid this step 😄

PR #9546, opened 2023-08-08
[CLI] Disallow "rotating" to same private key

banool:
### Description
It isn't harmful that we allow this, it's just there is no point in allowing it.

### Test Plan
```
$ ~/a/core/target/debug/aptos account rotate-key --private-key 0x98b22207d27a4b843e833f292a7f7ea8bb29829a0db018b2db6444636061aea9 --new-private-key 0x98b22207d27a4b843e833f292a7f7ea8bb29829a0db018b2db6444636061aea9 --url https://fullnode.testnet.aptoslabs.com
{
  "Error": "Invalid arguments: New private key cannot be the same as the current private key"
}
```

### DIFF
--- a/crates/aptos/src/account/key_rotation.rs
+++ b/crates/aptos/src/account/key_rotation.rs
@@ -104,6 +104,12 @@ impl CliCommand<RotateSummary> for RotateKey {
 
         let (current_private_key, sender_address) = self.txn_options.get_key_and_address()?;
 
+        if new_private_key == current_private_key {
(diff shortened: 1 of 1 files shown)


0xjinn (approved):
Make sense

xbtmatt (approved):
Tested it out, works as expected, lgtm!

I disabled auto-merge in case you or greg wanted to look at it before merging, since I wasn't tagged in reviewers

gregnazario (approved):

PR #9549, opened 2023-08-08
[TS SDK] Support network and API endpoints on `AptosConfig`

0xmaayan:
### Description
This PR adds support for network and API endpoints on `AptosConfig`. Once `Aptos` is initialized, a network and API endpoints are being set.

- Ability to initialize `Aptos` with no network and then it assigns a default one (testnet)
- Ability to initializes `Aptos` with a local/devnet/test/mainnet network, and the sdk assigns the network endpoint for each service (fullnode, faucet, indexer).
- Ability to initialize `Aptos` with a custom network - it expects the user to also provide the API endpoints.

This would be much simpler once API v2/gateway is ready.

### DIFF
--- a/ecosystem/typescript/sdk_v2/src/utils/api-endpoints.ts
+++ b/ecosystem/typescript/sdk_v2/src/utils/api-endpoints.ts
@@ -1,0 +1,27 @@
+export const NetworkToIndexerAPI: Record<string, string> = {
+  mainnet: "https://indexer.mainnet.aptoslabs.com/v1/graphql",
+  testnet: "https://indexer-testnet.staging.gcp.aptosdev.com/v1/graphql",
+  devnet: "https://indexer-devnet.staging.gcp.aptosdev.com/v1/graphql",
+};
+
+export const NetworkToNodeAPI: Record<string, string> = {
+  mainnet: "https://fullnode.mainnet.aptoslabs.com/v1",
+  testnet: "https://fullnode.testnet.aptoslabs.com/v1",
+  devnet: "https://fullnode.devnet.aptoslabs.com/v1",
+  local: "http://localhost:8080/v1",
+};
+
+export const NetworkToFaucetAPI: Record<string, string> = {
+  mainnet: "https://faucet.mainnet.aptoslabs.com",
+  testnet: "https://faucet.testnet.aptoslabs.com",
+  devnet: "https://faucet.devnet.aptoslabs.com",
+  local: "http://localhost:8081",
+};
+
+export enum Network {
+  MAINNET = "mainnet",
+  TESTNET = "testnet",
+  DEVNET = "devnet",
+  LOCAL = "local",
+  CUSTOM = "custom",
+}
--- a/ecosystem/typescript/sdk_v2/src/utils/const.ts
+++ b/ecosystem/typescript/sdk_v2/src/utils/const.ts
@@ -1,3 +1,3 @@
 import { Network } from "./api-endpoints";
 
-export const DEFAULT_NETWORK = Network.TESTNET;
+export const DEFAULT_NETWORK = Network.DEVNET;
--- a/ecosystem/typescript/sdk_v2/src/utils/const.ts
+++ b/ecosystem/typescript/sdk_v2/src/utils/const.ts
@@ -1,0 +1,3 @@
+import { Network } from "./api-endpoints";
+
+export const DEFAULT_NETWORK = Network.TESTNET;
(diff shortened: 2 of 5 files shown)


ecosystem/typescript/sdk_v2/tests/unit/aptos_config.test.ts
```diff
@@ -0,0 +1,59 @@
+import { Aptos, AptosConfig } from "../../src";
+import { Network } from "../../src/utils/api-endpoints";
+
+describe("aptos config", () => {
+  test("it should set TESTNET network if network is not provided", async () => {
```
xbtmatt, line 5: This would be new right, in v1 before you couldn't not specify a network?

I'm not sure this even matters or needs to be addressed now, but if we're going to provide a default network we could consider devnet, depending on how much we care about random data being permanently on testnet

ecosystem/typescript/sdk_v2/src/utils/const.ts
```diff
@@ -0,0 +1,3 @@
+import { Network } from "./api-endpoints";
+
+export const DEFAULT_NETWORK = Network.TESTNET;
```
gregnazario, line 3: I'd default this to either `DEVNET` or `MAINNET`, but probably let's do `DEVNET` until we launch it or make another decision.

ecosystem/typescript/sdk_v2/tests/unit/aptos_config.test.ts
xbtmatt: Also slightly related: none of this is actually package into the `aptos` package yet, right? So we don't have to worry about breaking anything for the time being until releasing it?

ecosystem/typescript/sdk_v2/src/utils/api-endpoints.ts
```diff
@@ -0,0 +1,27 @@
...
+  local: "http://localhost:8081",
+};
+
+export enum Network {
+  MAINNET = "mainnet",
+  TESTNET = "testnet",
+  DEVNET = "devnet",
+  LOCAL = "local",
+  CUSTOM = "custom",
+}
```
gregnazario, line 27: Does this include a `toString()` and a `fromString()` kind of stuff?

I found myself having to do `toString() to compare sometimes to inputs from the wallet adapter wallets.

ecosystem/typescript/sdk_v2/tests/unit/aptos_config.test.ts
0xmaayan: > This would be new right, in v1 before you couldn't not specify a network?
> 
> I'm not sure this even matters or needs to be addressed now, but if we're going to provide a default network we could consider devnet, depending on how much we care about random data being permanently on testnet
> 
> But idk, some people use testnet way more than I do so maybe they'd have an opinion about it

Yep, we dont enforce specifying network, and simply set it to a default one. As for what is the default, we can go either way, I remember in the past people pitched for testnet so new devs would not get confused on why their data suddenly disappeared lol but devnet also makes sense, don't have strong opinion here.

0xmaayan: another new thing is that we initialize with a Network enum which is a "name" and not with a network url as in v1

ecosystem/typescript/sdk_v2/src/utils/api-endpoints.ts
0xmaayan: not completely sure what you mean.. you can do `Network.DEVNET === "devnet"` that would output `true`

ecosystem/typescript/sdk_v2/src/utils/const.ts
0xmaayan: done

ecosystem/typescript/sdk_v2/src/utils/api-endpoints.ts
gregnazario: Oh, now I remember, it's because one side was `"Devnet"` and the other was "Network.DEVNET", so had to lowercase one of them.

gregnazario (approved):

ecosystem/typescript/sdk_v2/src/utils/api-endpoints.ts
0xmaayan: ohh yes, wallet standard and co.

xbtmatt (approved):

ecosystem/typescript/sdk_v2/tests/unit/aptos_config.test.ts
xbtmatt: Ah okay, makes sense to just keep it the same then! I didn't realize it already had a default value.

And yes, I love it!it works with the indexer too now :D

PR #9550, opened 2023-08-08
[release-builder] Increase lockup before executing proposals

perryjrandall:
Increase the lockup before executing proposals

Currently the release flow started consistently failing because we need to increase the lockup

This ensures that the lockup is sufficient before executing transactions

Test Plan: not sure how to test other than by running against testnet???

Please advise


davidiw (commented):
what if we don't add the args? Also I don't see the args being passed in from another component -- is that done in another repo?

perryjrandall:
> what if we don't add the args? Also I don't see the args being passed in from another component -- is that done in another repo?

In this flow those are not optional args

rahxephon89 (approved):

runtian-zhou (approved):

PR #9552, opened 2023-08-08
[RFC] Concurrent event creation - ConcurrentEventHandle using Aggregator V2

igor-aptos:
This is how would parallelizing events seq_num computaiton look like. Depending on the timeline of AIP-44, this might be abandoned, as AIP-44 makes this unnecessary.

Creating ConcurrentEventHandle to enable parallel event creation using Aggregators V2.

Need to think where to put EventSeqNum and ChangeSetEvent to make it cleanest/remove unnecessary dependencies.


davidiw (commented):
If we weren't strapped for resources, I might actually be curious to see both approaches and see which the community gravitates towards... Might actually be interesting to inspect the API data to see how many people query by events.

PR #9554, opened 2023-08-08
[NFT Metadata Crawler] Check received chain id against existing chain id in parser db

just-in-chang:
### Description
Adds chain id check to make sure the parser is pulling and writing data for the write chain

### Test Plan
Tested locally, looks good. Code is also mostly copied from indexers

Fixes EI-282

### DIFF
--- a/ecosystem/nft-metadata-crawler-parser/migrations/2023-06-27-001532_create_tables/down.sql
+++ b/ecosystem/nft-metadata-crawler-parser/migrations/2023-06-27-001532_create_tables/down.sql
@@ -1,2 +1,3 @@
 DROP TABLE IF EXISTS nft_metadata_crawler.parsed_token_uris;
+DROP TABLE IF EXISTS ledger_infos;
 DROP SCHEMA IF EXISTS nft_metadata_crawler CASCADE;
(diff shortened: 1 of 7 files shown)


bowenyang007 (approved):

ecosystem/nft-metadata-crawler-parser/migrations/2023-06-27-001532_create_tables/down.sql
```diff
@@ -1,2 +1,3 @@
 DROP TABLE IF EXISTS nft_metadata_crawler.parsed_token_uris;
+DROP TABLE IF EXISTS ledger_infos;
```
bowenyang007, line 2: wait this needs to say `nft_metadata_crawler.ledger_infos`

You should always run DATABASE=... diesel migration redo

bowenyang007: ah in this case it'll still pass...it's just a typo i guess.

ecosystem/nft-metadata-crawler-parser/migrations/2023-06-27-001532_create_tables/up.sql
```diff
@@ -13,6 +13,8 @@ CREATE TABLE IF NOT EXISTS nft_metadata_crawler.parsed_token_uris (
   inserted_at TIMESTAMP NOT NULL DEFAULT NOW()
 );
 
+CREATE TABLE nft_metadata_crawler.ledger_infos (chain_id BIGINT UNIQUE PRIMARY KEY NOT NULL);
```
bowenyang007, line 16: Create TABLE if not exists. So that you don't get into trouble for duplicating the migration.

rtso (approved):

PR #9556, opened 2023-08-08
Aptos-profiling crate

yunuseozer:
Adding the aptos-profiler crate with cpu and memory profiling features.

### DIFF
--- a/crates/aptos-profiler/src/cpu_profiler.rs
+++ b/crates/aptos-profiler/src/cpu_profiler.rs
@@ -60,6 +60,6 @@ impl Profiler for CpuProfiler {
     /// Expose the results as SVG
     fn expose_svg_results(&self) -> Result<String> {
         let content = convert_svg_to_string(self.svg_result_path.as_path());
-        content;
+        content
     }
 }
--- a/crates/aptos-profiler/src/cpu_profiler.rs
+++ b/crates/aptos-profiler/src/cpu_profiler.rs
@@ -1,0 +1,65 @@
+// Copyright © Aptos Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::{
+    utils::{convert_svg_to_string, create_file_with_parents},
+    CpuProfilerConfig, Profiler,
+};
+use anyhow::Result;
+use std::{path::PathBuf, thread, time};
+
+pub struct CpuProfiler {
+    frequency: i32,
+    svg_result_path: PathBuf,
+}
+
+impl CpuProfiler {
+    pub(crate) fn new(config: &CpuProfilerConfig) -> Self {
+        Self {
+            frequency: config.frequency,
+            svg_result_path: config.svg_result_path.clone(),
+        }
+    }
+}
+
+impl Profiler for CpuProfiler {
+    /// Perform CPU profiling for the given duration
+    fn profile_for(&self, duration_secs: u64) -> Result<()> {
+        let guard = pprof::ProfilerGuard::new(self.frequency).unwrap();
+        thread::sleep(time::Duration::from_secs(duration_secs));
+
+        if let Ok(report) = guard.report().build() {
+            let file = create_file_with_parents(self.svg_result_path.as_path())?;
+            let _result = report.flamegraph(file);
+        };
+
+        Ok(())
+    }
+
+    /// Start profiling until it is stopped
+    fn start_profiling(&self) -> Result<()> {
+        let _guard = pprof::ProfilerGuard::new(self.frequency).unwrap();
+        let duration = u64::MAX;
+        thread::sleep(time::Duration::from_secs(duration));
+
+        Ok(())
+    }
+
+    /// End profiling
+    fn end_profiling(&self) -> Result<()> {
+        //TODO: pprof-rs crate may not have a direct way of stopping the profiling from another function.
+        //Potential approach: return guard object to original scope and pass it here to stop and report results
+        todo!();
+    }
+
+    /// Expose the results as TXT
+    fn expose_text_results(&self) -> Result<String> {
+        unimplemented!();
+    }
+
+    /// Expose the results as SVG
+    fn expose_svg_results(&self) -> Result<String> {
+        let content = convert_svg_to_string(self.svg_result_path.as_path());
+        return content;
+    }
+}
--- a/crates/aptos-profiler/src/cpu_profiler.rs
+++ b/crates/aptos-profiler/src/cpu_profiler.rs
@@ -60,6 +60,6 @@ impl Profiler for CpuProfiler {
     /// Expose the results as SVG
     fn expose_svg_results(&self) -> Result<String> {
         let content = convert_svg_to_string(self.svg_result_path.as_path());
-        return content;
+        content;
     }
 }
--- a/crates/aptos-profiler/src/memory_profiler.rs
+++ b/crates/aptos-profiler/src/memory_profiler.rs
@@ -1,0 +1,127 @@
+// Copyright © Aptos Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::{utils::convert_svg_to_string, MemProfilerConfig, Profiler};
+use anyhow::{anyhow, Result};
+use std::{path::PathBuf, process::Command, thread, time::Duration};
+
+pub struct MemProfiler {
+    txt_result_path: PathBuf,
+    svg_result_path: PathBuf,
+}
+
+impl MemProfiler {
+    pub(crate) fn new(config: &MemProfilerConfig) -> Self {
+        Self {
+            txt_result_path: config.txt_result_path.clone(),
+            svg_result_path: config.svg_result_path.clone(),
+        }
+    }
+}
+
+impl Profiler for MemProfiler {
+    fn profile_for(&self, duration_secs: u64) -> Result<()> {
+        let mut prof_active: bool = true;
+
+        let result = unsafe {
+            jemalloc_sys::mallctl(
+                b"prof.active\0".as_ptr() as *const _,
+                std::ptr::null_mut(),
+                std::ptr::null_mut(),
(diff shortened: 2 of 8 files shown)


crates/aptos-profiler/src/cpu_profiler.rs
```diff
@@ -0,0 +1,55 @@
+// Copyright © Aptos Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+use anyhow::Result;
+use crate::{CpuProfilerConfig, Profiler};
+use crate::utils::{convert_svg_to_string, create_file_with_parents};
+use std::{thread, time, path::PathBuf};
+
+pub struct CpuProfiler {
+    duration: u64,
```
grao1991, line 10: nit: either use a Duration struct, or rename the variable with unit (e.g. duration_secs)

crates/aptos-profiler/src/cpu_profiler.rs
```diff
@@ -0,0 +1,55 @@
...
+        Self {
+            duration: config.duration,
+            frequency: config.frequency,
+            cpu_profiling_data_file: config.cpu_profiling_result.clone(),
+        }
+    }
+}
+
+impl Profiler for CpuProfiler {
+    fn start_profiling(&self) -> Result<()> {
```
grao1991, line 26: avoid unwrap in this function

crates/aptos-profiler/src/cpu_profiler.rs
```diff
@@ -0,0 +1,55 @@
...
+        Ok(())
+    }
+    // End profiling
+    fn end_profiling(&self) -> Result<()> {
+        
+    }
+        
+    // Expose the results as TXT
+    fn expose_text_results(&self) -> Result<String> {
+        Ok()
```
grao1991, line 45: Will this be supported?

use either todo!("...") or unimplemented!("...")

crates/aptos-profiler/src/cpu_profiler.rs
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    // Expose the results as SVG
+    fn expose_svg_results(&self) -> Result<String> {
+        let content = convert_svg_to_string(self.cpu_profiling_data_file.as_path());
+        return Ok(content.unwrap());
```
grao1991, line 51: avoid unwrap if possible

crates/aptos-profiler/src/cpu_profiler.rs
```diff
@@ -0,0 +1,55 @@
...
+        if let Ok(report) = guard.report().build() {
+            let file = create_file_with_parents(self.cpu_profiling_data_file.as_path())?;
+
+            report.flamegraph(file).unwrap();
+        };
+        
+        Ok(())
+    }
+    // End profiling
+    fn end_profiling(&self) -> Result<()> {
```
grao1991, line 39: do we really need an end_profiling function?

crates/aptos-profiler/src/memory_profiler.rs
```diff
@@ -0,0 +1,108 @@
+// Copyright © Aptos Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+use anyhow::{anyhow, Result};
+use crate::{MemProfilerConfig, Profiler, utils::convert_svg_to_string};
+use std::{thread, fs, path::PathBuf, time::Duration, process::Command};
+
+pub struct MemProfiler {
+    duration: u64,
```
grao1991, line 9: ditto

crates/aptos-profiler/src/memory_profiler.rs
```diff
@@ -0,0 +1,108 @@
...
+                &mut prof_active as *mut _ as *mut _,
+                std::mem::size_of::<bool>(),
+            )
+        };
+    
+        if result != 0 {
+            return Err(anyhow!("Failed to deactivate jemalloc profiling"));
+        }
+
+        Command::new("python3")
```
grao1991, line 60: add a TODO to switch away from python3.

crates/aptos-profiler/src/memory_profiler.rs
```diff
@@ -0,0 +1,108 @@
...
+            .arg(&self.memory_profiling_result_svg.to_string_lossy().as_ref())
+            .output()
+            .expect("Failed to execute command");
+        Ok(())
+    }
+
+    // Expose the results in TXT format
+    fn expose_text_results(&self) -> Result<String> {
+        let content = fs::read_to_string(self.memory_profiling_result_txt.as_path())
+        .expect("Failed to read input");
```
grao1991, line 99: return error instead of crashing here.

crates/aptos-profiler/src/memory_profiler.rs
```diff
@@ -0,0 +1,108 @@
...
+        return Ok(content);
+    }
+    // Expose the results in SVG format
+    fn expose_svg_results(&self) -> Result<String> {
+        let content = convert_svg_to_string(self.memory_profiling_result_svg.as_path())
+        .expect("Failed to read input");
```
grao1991, line 105: ditto

crates/aptos-profiler/src/memory_profiler.rs
```diff
@@ -0,0 +1,108 @@
...
+        Ok(())
+    }
+
+    // Expose the results in TXT format
+    fn expose_text_results(&self) -> Result<String> {
+        let content = fs::read_to_string(self.memory_profiling_result_txt.as_path())
+        .expect("Failed to read input");
+        return Ok(content);
+    }
+    // Expose the results in SVG format
```
areshand, line 102: 1. have a blank line between functions. 
2. use /// for function comments eg: /// Expose the results in SVG format

crates/aptos-profiler/src/memory_profiler.rs
```diff
@@ -0,0 +1,108 @@
+// Copyright © Aptos Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+use anyhow::{anyhow, Result};
+use crate::{MemProfilerConfig, Profiler, utils::convert_svg_to_string};
+use std::{thread, fs, path::PathBuf, time::Duration, process::Command};
+
+pub struct MemProfiler {
+    duration: u64,
+    memory_profiling_result_txt: PathBuf,
```
areshand, line 10: rename to txt_result_path?

crates/aptos-profiler/src/memory_profiler.rs
```diff
@@ -0,0 +1,108 @@
...
+    memory_profiling_result_svg: PathBuf,
```
areshand, line 11: svg_result_path

areshand:
you also need to add the crate to aptos-core/cargo.toml eg: ```
members = [
    "api",
    "api/openapi-spec-generator",
    "api/test-context",
    "api/types",
    "aptos-move/aptos-aggregator",
    "aptos-move/aptos-debugger",
```

crates/aptos-profiler/src/cpu_profiler.rs
yunuseozer: it seems that we will for executor benchmark integration

grao1991: How would that work? Are you going to stop the profiler by the duration_secs or manually?

yunuseozer: currently, it stops after duration_secs by default. the end profiling would stop it manually in concept. I currently do it for memory profiling, but it seems trickier for cpu because of the crate I'm using.

yunuseozer: and we'll likely need it for executor benchmark integration because when we are running it for the entire duration, it's hard to give an exact duration, so we will start with duration = max int and stop when needed. this is unimplemented though.

grao1991: In that case, I'd suggest to separate the interfaces, and do not store the duration as a member variable of the struct.

for profiling a predefined duration, take it as a parameter. e.g. fn profile_for(duration_secs: u64) -> Result<...>
for manually starting/stopping, have a fn start_profiling() -> Result<...> and fn stop_profiling() -> Result<...>

yunuseozer: The thing is, for CPU profiling, we have to provide a duration because it automatically stops when ProfilerGuard object goes out of scope. In that case, I think it would work if I give have max int as the duration for manually starting/stopping. Although it's not here yet, I'm currently working on adding a function to load configs from a yaml file, instead of just having default configs like here, so that will give the option for providing input, but I can create separate functions for running for a duration and running/stopping manually as you said as well.

grao1991: `I think it would work if I have max int as the duration for manually starting/stopping.`
OK

`so that will give the option for providing input`
Sounds good, however yaml file might be an overkill. http parameter for node and commandline flag for benchmark seem to be more clean and elegant.

`but I can create separate functions for running for a duration and running/stopping manually as you said as well.`
Yes, this is what I want, the interface need to be clean and well document.

yunuseozer: I've pushed an implementation. However, for CPU profiling without a duration, it starts the profiling, but stopping and generating results are unimplemented with the end_profiling function. Still exploring options on that as it wasn't a clear need before.

areshand:
> you also need to add the crate to `aptos-core/cargo.toml` 

Also add to [workplace.dependences] section
> 
> eg: 
> 
> ```
> 
> members = [
> 
>     "api",
> 
>     "api/openapi-spec-generator",
> 
>     "api/test-context",
> 
>     "api/types",
> 
>     "aptos-move/aptos-aggregator",
> 
>     "aptos-move/aptos-debugger",
> 
> ```

areshand (approved):
Lgtm. We can close some remaining issues after this lands

grao1991 (approved):

PR #9559, opened 2023-08-08
[Draft] Store `speculative_start_value` for each aggregator

vusirikala:
### Description

In this PR, we store a speculative start value inside each aggregator. Initially, the speculative start value is initialized with the last committed value of the aggregator. When the performs a read operation on the aggregator, we perform an expensive read operation to compute the most recent value of the aggregator.


aptos-move/aptos-aggregator/src/aggregator_extension.rs
```diff
@@ -119,279 +121,390 @@ pub struct AggregatorSnapshot {
...
+                Ok(value)
             },
-            AggregatorState::NegativeDelta => {
-                self.value = subtraction(value_from_storage, self.value)?;
-            },
-            AggregatorState::Data => {
-                unreachable!("Materialization only happens in Delta state")
+            AggregatorState::Delta { speculative_start_value, speculative_source, delta, history } => {
+                // If we performed a "cheap read" or "expensive read" operation before, use it.
+                if speculative_source != SpeculativeValueSource::Default {
```
igor-aptos, line 436: this is different than I had in mind, but I am liking having Default source, and lazily deciding to fetch on demand.

aptos-move/aptos-aggregator/src/aggregator_extension.rs
```diff
@@ -119,279 +121,390 @@ pub struct AggregatorSnapshot {
...
+                    return match delta {
+                        DeltaValue::Positive(value) => Ok(addition(speculative_start_value, value, self.max_value)?),
+                        DeltaValue::Negative(value) => Ok(subtraction(speculative_start_value, value)?),
```
igor-aptos, line 439: these shouldn't ever fail. not sure if we want to have them return exceptions?

aptos-move/aptos-aggregator/src/aggregator_extension.rs
```diff
@@ -119,279 +121,390 @@ pub struct AggregatorSnapshot {
...
+                    };
+                }
+                // Otherwise, we have to go to storage and read the value.
+                let value_from_storage = resolver.resolve_last_committed_aggregator_value(&self.id).map_err(|e| {
+                    extension_error(format!("Could not find the value of the aggregator: {}", e))
+                })?;
+        
+                // Validate history and apply the delta.
```
igor-aptos, line 447: if we are only now doing cheap read, history should be empty, maybe we should check that instead?

aptos-move/aptos-aggregator/src/aggregator_extension.rs
```diff
@@ -119,279 +121,390 @@ pub struct AggregatorSnapshot {
...
+                self.validate_history(value_from_storage)?;
+                self.state = AggregatorState::Delta {
+                    speculative_start_value: value_from_storage,
+                    speculative_source: SpeculativeValueSource::LastCommittedValue,
```
igor-aptos, line 451: this seems a bit more convoluted (and potentially more expensive due to copies) than just changing self.sate.speculative_start_value and self.state.speculative_source ?

aptos-move/aptos-aggregator/src/aggregator_extension.rs
```diff
@@ -428,14 +541,20 @@ impl AggregatorData {
...
+            state: AggregatorState::Delta {
+                speculative_start_value: 0,
+                speculative_source: SpeculativeValueSource::Default,
+                delta: DeltaValue::Positive(0),
+                history: DeltaHistory::new(),
+            },
             max_value,
-            history: Some(History::new()),
         });
+        aggregator.read_last_committed_aggregator_value(resolver);
```
igor-aptos, line 557: given how you've implemented this, maybe we should do this call in try_add/try_sub (i.e. snapshot() call doesn't need it, and read() call can just directly do the expensive read)

aptos-move/aptos-aggregator/src/aggregator_extension.rs
```diff
@@ -495,7 +612,25 @@ impl AggregatorData {
             .aggregator_snapshots
             .get(&id)
             .expect("AggregatorSnapshot doesn't exist");
-        snapshot.value
+        match snapshot.state {
```
igor-aptos, line 615: I think we want read_snapshot() to be equivalent to read(), as in - doing the expensive read.

igor-aptos (changes requested):
amount of issues in DeltaHistory suggests that you should really have a better tests for that component

aptos-move/aptos-aggregator/src/aggregator_extension.rs
```diff
@@ -2,24 +2,47 @@
...
+/// `AggregatorState` is obtained. 
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum SpeculativeValueSource {
+    // The speculative_start_value is assigned to 0.
+    Default,
+    // The speculative_start_value is obtained by reading
+    // the last committed value of the aggregator from MVHashmap.
+    LastCommittedValue,
+    // The speculative_start_value is obtained by aggregating all
+    // the previous deltas of the aggregator from MVHashmap.
```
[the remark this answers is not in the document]
vusirikala: Done.

aptos-move/aptos-aggregator/src/aggregator_extension.rs
```diff
@@ -119,279 +121,390 @@ pub struct AggregatorSnapshot {
...
-                    // of aggregator, we can figure out if the output of try_add/try_sub changes.
-                    // When self.value + value exceeds u128::MAX, we know that no matter what the starting value of the
-                    // aggregator is, the transaction always results in an underflow. So, we are 100% sure the output of the
-                    // transaction is an underflow.
-                    if self.value < u128::MAX - value {
-                        self.record_underflow(self.value + value);
+                    DeltaValue::Negative(current_delta) => {
+                        // If speculative_start_value - current_delta - input is less than 0, it's an underflow.
+                        // Otherwise, we update the delta with delta - input.
+                        if speculative_start_value > current_delta + input {
```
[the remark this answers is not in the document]

aptos-move/aptos-aggregator/src/aggregator_extension.rs

aptos-move/aptos-aggregator/src/aggregator_extension.rs
vusirikala: I would like to use `self.state.speculative_start_value` as well. But unfortunately, our syntax won't allow it as self.state is an enum.

PR #9564, opened 2023-08-09
[TS SDK v2] Add AccountAddress

banool:
### Description
This PR implements AccountAddress for the v2 TS SDK. This implementation conforms to AIP-40.

### Test Plan
```
cd ecosystem/typescript/sdk_v2
npx jest -- tests/unit/account_address.test.ts
```

### DIFF
--- a/ecosystem/typescript/sdk_v2/src/core/account_address.ts
+++ b/ecosystem/typescript/sdk_v2/src/core/account_address.ts
@@ -1,0 +1,340 @@
+// Copyright © Aptos Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+import { bytesToHex, hexToBytes } from "@noble/hashes/utils";
+import { HexInput } from "../types";
+import { ParsingError, ParsingResult } from "./common";
+
+/**
+ * This enum is used to explain why an address was invalid.
+ */
+export enum AddressInvalidReason {
+  INCORRECT_NUMBER_OF_BYTES = "incorrect_number_of_bytes",
+  INVALID_HEX_CHARS = "invalid_hex_chars",
+  TOO_SHORT = "too_short",
+  TOO_LONG = "too_long",
+  LEADING_ZERO_X_REQUIRED = "leading_zero_x_required",
+  LONG_FORM_REQUIRED_UNLESS_SPECIAL = "long_form_required_unless_special",
+}
+
+/**
+ * NOTE: Only use this class for account addresses. For other hex data, e.g. transaction
+ * hashes, use the Hex class.
+ *
+ * AccountAddress is used for working with account addresses. Account addresses, when
+ * represented as a string, generally look like these examples:
+ * - 0x1
+ * - 0xaa86fe99004361f747f91342ca13c426ca0cccb0c1217677180c9493bad6ef0c
+ *
+ * Proper formatting and parsing of account addresses is defined by AIP-40.
+ * To learn more about the standard, read the AIP here:
+ * https://github.com/aptos-foundation/AIPs/blob/main/aips/aip-40.md.
+ *
+ * The comments in this class make frequent reference to the LONG and SHORT formats,
+ * as well as "special" addresses. To learn what these refer to see AIP-40.
+ */
+export class AccountAddress {
+  /*
+   * This is the internal representation of an account address.
+   */
+  readonly data: Uint8Array;
+
+  /*
+   * The number of bytes that make up an account address.
+   */
+  static readonly LENGTH: number = 32;
+
+  /*
+   * The length of an address string in LONG form without a leading 0x.
+   */
+  static readonly LONG_STRING_LENGTH: number = 64;
+
+  static ONE: AccountAddress = AccountAddress.fromString({ input: "0x1" });
+
+  static TWO: AccountAddress = AccountAddress.fromString({ input: "0x2" });
+
+  static THREE: AccountAddress = AccountAddress.fromString({ input: "0x3" });
+
+  static FOUR: AccountAddress = AccountAddress.fromString({ input: "0x4" });
+
+  /**
+   * Creates an instance of AccountAddress from a Uint8Array.
+   *
+   * @param args.data A Uint8Array representing an account address.
+   */
+  constructor(args: { data: Uint8Array }) {
+    if (args.data.length !== AccountAddress.LENGTH) {
+      throw new ParsingError(
+        "AccountAddress data should be exactly 32 bytes long",
+        AddressInvalidReason.INCORRECT_NUMBER_OF_BYTES,
+      );
+    }
+    this.data = args.data;
+  }
+
+  /**
+   * Returns whether an address is special, where special is defined as 0x0 to 0xf
+   * inclusive. In other words, the last byte of the address must be < 0b10000 (16)
+   * and every other byte must be zero.
+   *
+   * For more information on how special addresses are defined see AIP-40:
+   * https://github.com/aptos-foundation/AIPs/blob/main/aips/aip-40.md.
+   *
+   * @returns true if the address is special, false if not.
+   */
+  isSpecial(): boolean {
+    return (
+      this.data.slice(0, this.data.length - 1).every((byte) => byte === 0) && this.data[this.data.length - 1] < 0b10000
+    );
+  }
+
+  // ===
+  // Methods for representing an instance of AccountAddress as other types.
+  // ===
+
+  /**
+   * Return the AccountAddress as a string as per AIP-40.
+   * https://github.com/aptos-foundation/AIPs/blob/main/aips/aip-40.md.
+   *
+   * In short, it means that special addresses are represented in SHORT form, meaning
+   * 0x0 through to 0xf inclusive, and every other address is represented in LONG form,
+   * meaning 0x + 64 hex characters.
+   *
+   * @returns AccountAddress as a string conforming to AIP-40.
+   */
+  toString(): string {
+    return `0x${this.toStringWithoutPrefix()}`;
+  }
+
+  /**
+   * NOTE: Prefer to use `toString` where possible.
+   *
+   * Return the AccountAddress as a string as per AIP-40 but without the leading 0x.
+   *
+   * Learn more by reading the docstring of `toString`.
+   *
+   * @returns AccountAddress as a string conforming to AIP-40 but without the leading 0x.
+   */
+  toStringWithoutPrefix(): string {
+    let hex = bytesToHex(this.data);
+    if (this.isSpecial()) {
+      hex = hex[hex.length - 1];
+    }
+    return hex;
+  }
+
+  /**
+   * NOTE: Prefer to use `toString` where possible.
+   *
+   * Whereas toString will format special addresses (as defined by isSpecial) using the
+   * SHORT form (no leading 0s), this format the address in the LONG format
+   * unconditionally.
+   *
+   * This means it will be 0x + 64 hex characters.
+   *
+   * @returns AccountAddress as a string in LONG form.
+   */
+  toStringLong(): string {
+    return `0x${this.toStringLongWithoutPrefix()}`;
+  }
+
+  /*
+   * NOTE: Prefer to use `toString` where possible.
+   *
+   * Whereas toString will format special addresses (as defined by isSpecial) using the
+   * SHORT form (no leading 0s), this function will include leading zeroes. The string
+   * will not have a leading zero.
+   *
+   * This means it will be 64 hex characters without a leading 0x.
+   *
+   * @returns AccountAddress as a string in LONG form without a leading 0x.
+   */
+  toStringLongWithoutPrefix(): string {
+    return bytesToHex(this.data);
+  }
+
+  /**
+   * Get the inner hex data. The inner data is already a Uint8Array so no conversion
+   * is taking place here, it just returns the inner data.
+   *
+   * @returns Hex data as Uint8Array
+   */
+  toUint8Array(): Uint8Array {
+    return this.data;
+  }
+
+  // ===
+  // Methods for creating an instance of AccountAddress from other types.
+  // ===
+
+  /**
+   * NOTE: This function has strict parsing behavior. For relaxed behavior, please use
+   * the `fromStringRelaxed` function.
+   *
+   * Creates an instance of AccountAddress from a hex string.
+   *
+   * This function allows only the strictest formats defined by AIP-40. In short this
+   * means only the following formats are accepted:
+   *
+   * - LONG
+   * - SHORT for special addresses
+   *
+   * Where:
+   * - LONG is defined as 0x + 64 hex characters.
+   * - SHORT for special addresses is 0x0 to 0xf inclusive.
+   *
+   * This means the following are not accepted:
+   * - SHORT for non-special addresses.
+   * - Any address without a leading 0x.
+   *
+   * Learn more about the different address formats by reading AIP-40:
+   * https://github.com/aptos-foundation/AIPs/blob/main/aips/aip-40.md.
+   *
+   * @param args.input A hex string representing an account address.
+   *
+   * @returns An instance of AccountAddress.
+   */
+  static fromString(args: { input: string }): AccountAddress {
+    // Assert the string starts with 0x.
+    if (!args.input.startsWith("0x")) {
+      throw new ParsingError("Hex string must start with a leading 0x.", AddressInvalidReason.LEADING_ZERO_X_REQUIRED);
+    }
+
+    const address = AccountAddress.fromStringRelaxed(args);
+
+    // Assert that only special addresses can use short form.
+    if (args.input.slice(2).length !== this.LONG_STRING_LENGTH && !address.isSpecial()) {
+      throw new ParsingError(
+        "Hex string is not a special address, it must be represented as 0x + 64 chars.",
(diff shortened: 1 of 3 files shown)


ecosystem/typescript/sdk_v2/src/types/account_address.ts
```diff
@@ -0,0 +1,241 @@
...
+  ADDRESS_TOO_LONG = "address_too_long",
+  LEADING_ZERO_X_REQUIRED = "leading_zero_x_required",
+  LONG_FORM_REQUIRED = "long_form_required",
+  LONG_FORM_REQUIRED_UNLESS_SPECIAL = "long_form_required_unless_special",
+}
+
+/**
+ * This error is used to explain why an address was invalid.
+ */
+export class AddressInvalidError extends Error {
```
0xmaayan, line 53: I was thinking on how the SDK should handle errors. One option is to throw an error as we do (slightly) in v1. I found it a bit annoying as the Error class is a bit weird in the way that by default it handles and logs only the `message` part. For example, with the `AddressInvalidError`, the message will be logged but the `reason` part not. I was thinking on maybe stop throwing errors and instead start returning Object with a `status`, `message`, etc. something like
```
type AptosResponse = {
 status: "success" | "error"
 message: string
}
```

also wonder what you all think @xbtmatt @gregnazario @0xjinn

ecosystem/typescript/sdk_v2/src/types/account_address.ts
```diff
@@ -0,0 +1,241 @@
...
+// wallet and dapp folks.
+export const defaultValidityOptions: ValidityOptions = {
+  requireLongFormUnlessSpecial: false,
+  requireLongForm: false,
+  requireLeadingZeroX: false,
+};
+
+export enum AddressInvalidReason {
+  INCORRECT_NUMBER_OF_BYTES = "incorrect_number of bytes",
+  ADDRESS_TOO_SHORT = "address_too_short",
```
0xmaayan, line 43: maybe?
```suggestion
  ADDRESS_TOO_SHORT = "address too short should be between 1 to 64",
```

ecosystem/typescript/sdk_v2/src/types/account_address.ts
```diff
@@ -0,0 +1,241 @@
...
+import { Serializer, Deserializer, Bytes } from "../bcs";
+
+// TODO: Get feedback on these error messages, ideally they can be used directly in
+// dapps and wallets. The enum is meant to make it possible for daps to user their
+// own error messages if they want, but the defaults should still be good, and they're
+// not great right now.
+
+// These options influence fromStr and the isValid* functions. You may use them to make
+// the parsing of those functions strictier. Not setting these options is the best way
+// to comply with AIP 40 but while migrating it may be helpful to be able to enforce
```
0xmaayan, line 14: in what cases we would want to `be able to enforce stricter behavior`?

ecosystem/typescript/sdk_v2/src/types/account_address.ts
```diff
@@ -0,0 +1,241 @@
...
+  // The number of bytes that make up an account address.
+  static readonly LENGTH: number = 32;
+
+  // The length of an address string in long form without a leading 0x.
+  static readonly LONG_STRING_LENGTH: number = 64;
+
+  // This is the internal representation of an account address.
+  readonly address: Bytes;
+
+  static ADDRESS_ONE: AccountAddress = AccountAddress.fromStr("0x1");
```
0xmaayan, line 75: what are these for?

0xmaayan (commented):
Looks good! left some comments

ecosystem/typescript/sdk_v2/src/types/account_address.ts
gregnazario: Only problem with returning a response that's not an error, is that now you can't easily use `try catch` and instead have to change your whole flow to use type matching on errors.  I'm not sure how that works with the average developer

ecosystem/typescript/sdk_v2/src/types/account_address.ts
```diff
@@ -0,0 +1,241 @@
...
+
+  // Only addresses in the long form (64 characters + optionally a leading 0x) will be
+  // considered valid, even if the address is special.
+  requireLongForm?: boolean;
+
+  // If set, the address string will be considered invalid if there is no leading 0x.
+  requireLeadingZeroX?: boolean;
+};
+
+// TODO: As part of reviewing this we should discuss these options. I'm partially
```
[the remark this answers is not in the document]
banool: That's a fair point. So here's what I want to update the standard to:

Functions that parse / validate strings as AccountAddress MUST accept the following:

a. LONG
b. SHORT for special addresses

The following correspond to representations actively in circulation. Implementations MAY accept these formats for the sake of backwards compatibility with the existing ecosystem.

c. LONG with or without leading 0x
d. SHORT with or without leading 0x for all addresses (not just special)

These are allowed for the sake of migration, since some tools return account addresses in these latter formats and cannot be updated without breaking backwards compatibility. Where possible, developers should avoid accepting these formats (e.g. when building APIs, wallets, tools, etc) but it is allowed in order to maintain compatibility.

ecosystem/typescript/sdk_v2/src/types/account_address.ts
```diff
@@ -0,0 +1,241 @@
...
+// actually be enabled by default. Maayan and Greg to provide their thoughts as well as
+// wallet and dapp folks.
+export const defaultValidityOptions: ValidityOptions = {
+  requireLongFormUnlessSpecial: false,
+  requireLongForm: false,
+  requireLeadingZeroX: false,
+};
+
+export enum AddressInvalidReason {
+  INCORRECT_NUMBER_OF_BYTES = "incorrect_number of bytes",
```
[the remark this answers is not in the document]
banool: Works for me!

ecosystem/typescript/sdk_v2/src/types/account_address.ts
```diff
@@ -0,0 +1,241 @@
...
+
+    return address;
+  }
+
+  /**
+   * Checks if the string is a valid AccountAddress.
+   *
+   * @returns True if the string is valid, false if not.
+   */
+  static isValid(str: string, validityOptions: ValidityOptions = defaultValidityOptions): boolean {
```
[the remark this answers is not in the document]
banool: You think we should just remove `isValid`? I suppose that's a good idea, then it's up to the user which of the values they use in the response.

ecosystem/typescript/sdk_v2/src/types/account_address.ts
```diff
@@ -0,0 +1,241 @@
...
+      return {
+        valid: false,
+        invalidReason: addressInvalidError.message,
+        invalidReasonCode: addressInvalidError.reason,
+      };
+    }
+  }
+
+  /**
+   * Returns whether an address is special, where special is defined as 0x0 to 0xf
```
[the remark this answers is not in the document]
banool: I could yeah, though what do you think about linking to the AIP instead, where it explains it in detail?

ecosystem/typescript/sdk_v2/src/types/account_address.ts
banool: I agree JS errors are sorta nasty for exactly the reason you describe Maayan, but I also agree with Greg that returning essentially a Result is not really standard for JS.

ecosystem/typescript/sdk_v2/src/types/account_address.ts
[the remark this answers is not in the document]
banool: So these aren't really meant to be the messages, these are just the enum values. See comment below.

ecosystem/typescript/sdk_v2/src/types/account_address.ts
banool: So this isn't really meant to be the explanation, that comes from within the function. The value for this enum could just as easily be a number rather than a string, which maybe I'll do.

ecosystem/typescript/sdk_v2/src/types/account_address.ts
banool: We already had ADDRESS_ONE (under a different name) in the previous AccountAddress class. We tend to deploy all framework code on 0x1, 0x2, 0x3, or 0x4. So it's handy to have constants for those. We do the same thing in Rust and Python.

ecosystem/typescript/sdk_v2/src/types/account_address.ts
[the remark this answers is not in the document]
banool: We spoke offline, I'll make a change to the AIP.

ecosystem/typescript/sdk_v2/src/types/account_address.ts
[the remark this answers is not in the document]
0xmaayan: yes we can. maybe something like "special address is reserved/used by aptos framework" and then a link to it

ecosystem/typescript/sdk_v2/src/types/account_address.ts
```diff
@@ -0,0 +1,317 @@
+// Copyright © Aptos Foundation
```
0xmaayan, line 1: following this comment on HexData class, lets move this class to either `utils` or `code` folder https://github.com/aptos-labs/aptos-core/pull/9595/files#r1290573135

ecosystem/typescript/sdk_v2/src/bcs/consts.ts
```diff
@@ -0,0 +1,12 @@
...
+
+import { Uint128, Uint16, Uint32, Uint64, Uint8, Uint256 } from "./types";
+
+// Upper bound values for uint8, uint16, uint64 and uint128
+export const MAX_U8_NUMBER: Uint8 = 2 ** 8 - 1;
+export const MAX_U16_NUMBER: Uint16 = 2 ** 16 - 1;
+export const MAX_U32_NUMBER: Uint32 = 2 ** 32 - 1;
+export const MAX_U64_BIG_INT: Uint64 = BigInt(2 ** 64) - BigInt(1);
+export const MAX_U128_BIG_INT: Uint128 = BigInt(2 ** 128) - BigInt(1);
+export const MAX_U256_BIG_INT: Uint256 = BigInt(2 ** 256) - BigInt(1);
```
gregnazario, line 12: Curious... does BigInt(2**64) work if, it's above the value for a standard number?

I just realized this (we use this in the other SDK too right?)

ecosystem/typescript/sdk_v2/src/bcs/deserializer.ts
```diff
@@ -0,0 +1,188 @@
...
+   * and then reads "l" bytes of content. Decodes the byte array into a string.
+   *
+   * BCS layout for "string": string_length | string_content. string_length is the bytes length of
+   * the string that is uleb128 encoded. string_length is a u32 integer.
+   *
+   * @example
+   * ```ts
+   * const deserializer = new Deserializer(new Uint8Array([24, 0xc3, 0xa7, 0xc3, 0xa5, 0xe2, 0x88, 0x9e,
+   * 0xe2, 0x89, 0xa0, 0xc2, 0xa2, 0xc3, 0xb5, 0xc3, 0x9f, 0xe2, 0x88, 0x82, 0xc6, 0x92, 0xe2, 0x88, 0xab]));
+   * assert(deserializer.deserializeStr() === "çå∞≠¢õß∂ƒ∫");
```
gregnazario, line 41: Lol, these bytes, maybe a more human readable example?

ecosystem/typescript/sdk_v2/src/bcs/serializer.ts
```diff
@@ -0,0 +1,241 @@
...
+
+  private offset: number;
+
+  constructor() {
+    this.buffer = new ArrayBuffer(64);
+    this.offset = 0;
+  }
+
+  private ensureBufferWillHandleSize(bytes: number) {
+    while (this.buffer.byteLength < this.offset + bytes) {
```
gregnazario, line 27: Nit: same comparison as way above, but inverted condition :)

ecosystem/typescript/sdk_v2/src/bcs/serializer.ts
```diff
@@ -0,0 +1,241 @@
...
+
+  /**
+   * Serializes a boolean value.
+   *
+   * BCS layout for "boolean": One byte. "0x01" for True and "0x00" for False.
+   */
+  serializeBool(value: boolean): void {
+    if (typeof value !== "boolean") {
+      throw new Error("Value needs to be a boolean");
+    }
```
gregnazario, line 99: Interesting that this is validated in boolean, but not in the others?

Oh, I guess there are checks in `checkNumberRange`

ecosystem/typescript/sdk_v2/src/bcs/serializer.ts
```diff
@@ -0,0 +1,241 @@
...
+   * ```ts
+   * const serializer = new Serializer();
+   * serializer.serializeU32(305419896);
+   * assert(serializer.getBytes() === new Uint8Array([0x78, 0x56, 0x34, 0x12]));
+   * ```
+   */
+  @checkNumberRange(0, MAX_U32_NUMBER)
+  serializeU32(value: Uint32): void {
+    this.serializeWithFunction(DataView.prototype.setUint32, 4, value);
+  }
```
gregnazario, line 144: These already handle little endianness automatically?

ecosystem/typescript/sdk_v2/src/core/account_address.ts
```diff
@@ -0,0 +1,342 @@
...
+   */
+  static readonly LONG_STRING_LENGTH: number = 64;
+
+  static ADDRESS_ONE: AccountAddress = AccountAddress.fromString({ str: "0x1" });
+
+  static ADDRESS_TWO: AccountAddress = AccountAddress.fromString({ str: "0x2" });
+
+  static ADDRESS_THREE: AccountAddress = AccountAddress.fromString({ str: "0x3" });
+
+  static ADDRESS_FOUR: AccountAddress = AccountAddress.fromString({ str: "0x4" });
```
gregnazario, line 59: I forget can statics be used in Typescript like `AccountAddress.ONE`, that may be easier then.

ecosystem/typescript/sdk_v2/src/core/account_address.ts
```diff
@@ -0,0 +1,342 @@
...
+
+  /**
+   * Creates an instance of AccountAddress from a Uint8Array.
+   *
+   * @param args.data A Uint8Array representing an account address.
+   */
+  constructor(args: { data: Uint8Array }) {
+    if (args.data.length !== AccountAddress.LENGTH) {
+      throw new ParsingError("Expected address of length 32", AddressInvalidReason.INCORRECT_NUMBER_OF_BYTES);
```
gregnazario, line 68: 32 bytes

ecosystem/typescript/sdk_v2/tests/unit/account_address.test.ts
```diff
@@ -0,0 +1,216 @@
...
+      ADDRESS_OTHER.longWith0x,
+    );
+    expect(AccountAddress.fromStringRelaxed({ str: ADDRESS_OTHER.longWithout0x }).toString()).toBe(
+      ADDRESS_OTHER.longWith0x,
+    );
+  });
+});
+
+// These tests show that fromString only parses addresses with a leading 0x and only
+// SHORT if it is a specil address.
```
gregnazario, line 109: *special

gregnazario (commented):
I'm not sure you wanted to include BCS, but I think if we do, we should have some preset tests to ensure serialization / deserialization works properly (I don't know if I missed it)

banool:
It'll probably be cleaner if I remove the serde functions from AccountAddress and remove the BCS stuff from this PR. Then we can add the BCS stuff later.

ecosystem/typescript/sdk_v2/src/types/account_address.ts
xbtmatt: Yes, I use the `0x1` a lot, so much that I have it memorized lol

```typescript
TxnBuilderTypes.AccountAddress.CORE_CODE_ADDRESS
```

But yeah, ADDRESS_ONE probably makes more sense, esp cuz you can extend it to 3 and 4 for the token stuff

ecosystem/typescript/sdk_v2/src/bcs/deserializer.ts
xbtmatt: Lol yea this was copied from the example in the original sdk idk why it was this string...

whenever we do do this we should use something like this:

```typescript
const deserializer = new Deserializer(new Uint8Array([6, 0x61, 0x62, 0x63, 0x31, 0x32, 0x33]));
assert(deserializer.deserializeStr() == "abc123");
```

ecosystem/typescript/sdk_v2/src/core/account_address.ts
banool: Good call let me do that.

0xmaayan (commented):
LGTM, left come comments

banool:
I just added tests for fromHexInput and fromHexInputRelaxed. I also added a toUint8Array method (like we have with Hex( and added tests for that.

0xmaayan (approved):

gregnazario (approved):

PR #9576, opened 2023-08-09
[CLI] Handle empty network address in CLI

0xjinn:
### Description

When the validator network address is empty, the whole validator won't show up when running `aptos node show-validator-set`. 

This PR adds the updates the conditon to following

```
Empty Address -> return vec![]
Not parsable -> Return error instead of actual result
Valid address -> Ok()
```

### Test Plan

1. Force the bcs parsing to fail and result in the following result
```
➜  aptos-core git:(jin/cli_handle_empty_network_addresses) ✗ aptos-debug node show-validator-set --profile validator1
{
  "Error": "Error (de)serializing 'Validator Set': unexpected end of input"
}
```

2. Empty network address
```
{
  "Result": {
    "scheme": "BLS12381",
    "active_validators": [
      {
        "account_address": "df5cdcfb4286b841736f5e36f9c68b22c4565ac62b0319bfbed4782652d7d2e7",
        "consensus_voting_power": 10,
        "config": {
          "consensus_public_key": "0xb0fdf93c841281e9633d842772bad1382a2907a1bdf4c8cc3d04307358ceffb372befdfd886111a0a0073fa88479cb89",
          "validator_network_addresses": [
            "/ip4/0.0.0.0/tcp/6180/noise-ik/0x44ffbba0c41d7cf82c07bd94914ba4c90f6b7ad558a5b5901d00bfd3c76c3a17/handshake/0"
          ],
          "fullnode_network_addresses": [
            "/ip4/0.0.0.0/tcp/6181/noise-ik/0x850102ff8a92270b6b95012c6c69a7fb090f044adb5033f29525306786dc0930/handshake/0"
          ],
          "validator_index": 0
        }
      },
      {
        "account_address": "3a3a058973df823ceaf238c90c199d75ceba7aff5acfb1e94e7edc9b2ed70eba",
        "consensus_voting_power": 1,
        "config": {
          "consensus_public_key": "0xacd224997d91ac55b00beaf54ac8c28b1941ff58f85565a7cdc7016ee6d2c5fe50937dedfb1a20786559fac819ccbf66",
          "validator_network_addresses": [],
          "fullnode_network_addresses": [],
          "validator_index": 1
        }
      }
    ],
    "pending_inactive": [],
    "pending_active": [],
    "total_voting_power": 11,
    "total_joining_power": 0
  }
}
```


crates/aptos/src/node/mod.rs
```diff
@@ -1009,8 +1009,14 @@ impl TryFrom<&ValidatorConfig> for ValidatorConfigSummary {
         Ok(ValidatorConfigSummary {
             consensus_public_key,
             // TODO: We should handle if some of these are not parsable
-            validator_network_addresses: config.validator_network_addresses()?,
-            fullnode_network_addresses: config.fullnode_network_addresses()?,
+            validator_network_addresses: match config.validator_network_addresses() {
+                Ok(addresses) => addresses,
+                Err(_) => vec![],
```
gregnazario, line 1014: Maybe use `eprintln!("")` to print out the error so that we know that it's not empty?

Either that or we probably need to fix the parsing of the network addresses to skip broken ones

0xjinn: @gregnazario  Use `eprintln!("")` to print out the error on top of the whole result?

I think that works, another idea I have is to divide it into following:

Empty Address -> return vec![]
Not parsable -> Return error instead of actual result
Valid address -> Ok()

gregnazario: That also works as well

0xjinn:
Updated the PR to throw an error when validator is not parsable. See description for corresponding ressult

davidiw (approved):

banool (approved):

PR #9584, opened 2023-08-09
[TS SDK] create HexData module

0xmaayan:
### Description
a class that handles validation, standardization, etc for non account address hex string data - public key, private key, txn hex, etc

Both `Account` and `AccountAddress` depends on this module.
https://www.notion.so/aptoslabs/be09c985c1cb4575a53cece9d35f05aa?v=de52612cb01441848573341ba0bf3e34

### DIFF
--- a/ecosystem/typescript/sdk_v2/src/utils/hexData.ts
+++ b/ecosystem/typescript/sdk_v2/src/utils/hexData.ts
@@ -1,0 +1,67 @@
+import { bytesToHex, hexToBytes } from "@noble/hashes/utils";
+import { Hex } from "../types";
+
+export class HexData {
+  private hexData: Hex;
+
+  /**
+   * Creates new HexData instance from Hex type.
+   * @param hex Uint8Array
+   */
+  constructor(hex: Hex) {
+    this.hexData = hex;
+  }
+
+  /**
+   * Getter method to get the inner hexData
+   * @returns inner hexData
+   */
+  public get hex(): Hex {
+    return this.hexData;
+  }
+
+  /**
+   * Static method to convert given hex string to HexData
+   * @returns HexData
+   */
+  static fromString(hex: string): HexData {
+    if (hex.startsWith("0x")) {
+      return new HexData(HexData.toBytes(hex));
+    }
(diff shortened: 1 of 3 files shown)


ecosystem/typescript/sdk_v2/src/utils/hexData.ts
```diff
@@ -0,0 +1,67 @@
...
+import { Hex } from "../types";
+
+export class HexData {
+  private hexData: Hex;
+
+  /**
+   * Creates new HexData instance from Hex type.
+   * @param hex Uint8Array
+   */
+  constructor(hex: Hex) {
```
xbtmatt, line 11:
```suggestion
  constructor(hex: Uint8Array) {
```

I assume you meant this, right? Given the `@param hex Uint8Array` notation and the usage of `fromString`

ecosystem/typescript/sdk_v2/src/utils/hexData.ts
```diff
@@ -0,0 +1,67 @@
+import { bytesToHex, hexToBytes } from "@noble/hashes/utils";
+import { Hex } from "../types";
+
+export class HexData {
```
banool, line 4: My initial understanding was that this was going to just be a class full of static methods, not something you can instantiate. If we're going with this approach, it should probably be in types, not utils. Also it's weird that this is something you can instantiate but then it's just all static methods mostly.

ecosystem/typescript/sdk_v2/src/utils/hexData.ts
```diff
@@ -0,0 +1,67 @@
...
+  private hexData: Hex;
```
banool, line 5: If the class has an internal representation of hex data, let's not use Hex, this is what made HexString so confusing. Can we instead represent hex data using Uint8Array instead, similar to how I do it in AccountAddress?

banool:
I spoke to Maayan offline and suggested some changes. We'll move forward with this PR: https://github.com/aptos-labs/aptos-core/pull/9595.

0xmaayan:
closing in favor of https://github.com/aptos-labs/aptos-core/pull/9595

PR #9457, opened 2023-08-03
[dag] e2e integration test

ibalajiarun:
### Description

This PR implements an e2e integration test for the DAG modules. The test currently only asserts the good case behavior and doesn't cover all possible invocations. It uses the NetworkPlayground to simulate the network.

PR #9456, opened 2023-08-03
[dag] network sender implementation

ibalajiarun:
### Description

This PR implements the DAG Network Sender trait.

PR #9585, opened 2023-08-09
[dag] Integrate dag fetcher with RB node handler

ibalajiarun:
### Description

This PR integrates DAG fetcher to the RB node handler. Whenever the node is missing parents, the fetcher is requested to fetch those missing parents and their ancestors. To make sure the GCed DAG is not refetched, it changes the lowest_round in dag_store to provide the information.

### Test Plan

Existing tests.

### DIFF
--- a/consensus/src/dag/rb_handler.rs
+++ b/consensus/src/dag/rb_handler.rs
@@ -1,7 +1,7 @@
 // Copyright © Aptos Foundation
 // SPDX-License-Identifier: Apache-2.0
 
-use super::{storage::DAGStorage, NodeId};
+use super::{dag_fetcher::TFetchRequester, storage::DAGStorage, NodeId};
 use crate::dag::{
     dag_network::RpcHandler,
     dag_store::Dag,
@@ -21,8 +21,8 @@ pub enum NodeBroadcastHandleError {
     InvalidParent,
     #[error("missing parents")]
     MissingParents,
-    #[error("parents do not meet quorum voting power")]
-    NotEnoughParents,
+    #[error("stale round number")]
+    StaleRound(Round),
 }
 
 pub(crate) struct NodeBroadcastHandler {
@@ -31,6 +31,7 @@ pub(crate) struct NodeBroadcastHandler {
     signer: ValidatorSigner,
     epoch_state: Arc<EpochState>,
     storage: Arc<dyn DAGStorage>,
+    fetch_requester: Arc<dyn TFetchRequester>,
 }
 
 impl NodeBroadcastHandler {
@@ -39,6 +40,7 @@ impl NodeBroadcastHandler {
         signer: ValidatorSigner,
         epoch_state: Arc<EpochState>,
         storage: Arc<dyn DAGStorage>,
+        fetch_requester: Arc<dyn TFetchRequester>,
     ) -> Self {
         let epoch = epoch_state.epoch;
         let votes_by_round_peer = read_votes_from_storage(&storage, epoch);
@@ -49,6 +51,7 @@ impl NodeBroadcastHandler {
             signer,
             epoch_state,
             storage,
+            fetch_requester,
         }
     }
 
@@ -67,21 +70,15 @@ impl NodeBroadcastHandler {
         self.storage.delete_votes(to_delete)
     }
 
-    fn validate(&self, node: &Node) -> anyhow::Result<()> {
+    fn validate(&self, node: Node) -> anyhow::Result<Node> {
         let current_round = node.metadata().round();
 
-        // round 0 is a special case and does not require any parents
-        if current_round == 0 {
-            return Ok(());
-        }
-
-        let prev_round = current_round - 1;
-
         let dag_reader = self.dag.read();
-        // check if the parent round is missing in the DAG
+        let lowest_round = dag_reader.lowest_round();
+
         ensure!(
-            prev_round >= dag_reader.lowest_round(),
-            NodeBroadcastHandleError::MissingParents
+            current_round >= lowest_round,
+            NodeBroadcastHandleError::StaleRound(current_round)
         );
 
         // check which parents are missing in the DAG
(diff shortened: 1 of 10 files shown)


ibalajiarun:
* main
  * **PR #9312** <a href="https://app.graphite.dev/github/pr/aptos-labs/aptos-core/9312" target="_blank">Graphite</a> 
          * **PR #9455** <a href="https://app.graphite.dev/github/pr/aptos-labs/aptos-core/9455" target="_blank">Graphite</a> 
            * **PR #9456** <a href="https://app.graphite.dev/github/pr/aptos-labs/aptos-core/9456" target="_blank">Graphite</a> 
              * **PR #9457** <a href="https://app.graphite.dev/github/pr/aptos-labs/aptos-core/9457" target="_blank">Graphite</a> 
                * **PR #9585** <a href="https://app.graphite.dev/github/pr/aptos-labs/aptos-core/9585" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/aptos-labs/aptos-core/9585?utm_source=stack-comment).

consensus/src/dag/rb_handler.rs
```diff
@@ -67,22 +72,24 @@ impl NodeBroadcastHandler {
...
     }
 
-    fn validate(&self, node: &Node) -> anyhow::Result<()> {
+    fn validate(&self, node: Node) -> anyhow::Result<Node> {
         let current_round = node.metadata().round();
 
         // round 0 is a special case and does not require any parents
         if current_round == 0 {
-            return Ok(());
+            bail!(NodeBroadcastHandleError::InvalidRound);
```
zekun000, line 80: nit: we should start to carry more debug information into the errors, like exact round, missing parents id, why it's invalid etc

consensus/src/dag/rb_handler.rs
```diff
@@ -67,22 +72,24 @@ impl NodeBroadcastHandler {
...
         let prev_round = current_round - 1;
 
         let dag_reader = self.dag.read();
         // check if the parent round is missing in the DAG
-        ensure!(
-            prev_round >= dag_reader.lowest_round(),
-            NodeBroadcastHandleError::MissingParents
-        );
+        if prev_round >= dag_reader.lowest_round() {
+            if let Err(err) = self.fetch_requester.request_for_node(node) {
```
zekun000, line 88: I think we should at least unify this condition with the one below, it also doesn't look correct to me, maybe what you meant to check is prev_round > `highest_round`?

previously the check is to filter out node that below lowest round, the error was misleading

zekun000 (changes requested):

consensus/src/dag/rb_handler.rs
ibalajiarun: Fix this now. I will do a separate pass for improving the errors overall and start adding metric counters.

consensus/src/dag/rb_handler.rs
ibalajiarun: Good catch. I did not account for check inversion - I was doing ensure! before. But, I removed this condition and just use the one with missing parents check below. Previously I was not checking for signatures if none of the parents existed locally, but I should have.

consensus/src/dag/rb_handler.rs
```diff
@@ -67,22 +72,24 @@ impl NodeBroadcastHandler {
...
+                error!("request to fetch failed: {}", err);
+            }
+            bail!(NodeBroadcastHandleError::MissingParents);
+        }
```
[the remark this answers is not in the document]
ibalajiarun: I merged this check with the below one. This one will miss the case when a node has all invalid parents.

consensus/src/dag/rb_handler.rs
```diff
@@ -67,22 +72,13 @@ impl NodeBroadcastHandler {
...
-            return Ok(());
-        }
-
-        let prev_round = current_round - 1;
+        // ensure round is greater than 0. Don't accept a zero-round.
+        ensure!(current_round > 0, NodeBroadcastHandleError::InvalidRound(current_round));
 
         let dag_reader = self.dag.read();
-        // check if the parent round is missing in the DAG
-        ensure!(
```
zekun000, line 82: we still need to filter out the lower round right? otherwise we'd fetch all those nodes from gc'ed rounds

consensus/src/dag/rb_handler.rs
```diff
@@ -92,18 +88,21 @@ impl NodeBroadcastHandler {
             .cloned()
             .collect();
         if !missing_parents.is_empty() {
-            // For each missing parent, verify their signatures and voting power
+            // For each missing parent, verify their signatures and voting power. Otherwise, a malicious node can send 
+            // bad nodes with fake parents and cause this peer to issue unnecessary fetch requests.
             ensure!(
                 missing_parents
                     .iter()
                     .all(|parent| { parent.verify(&self.epoch_state.verifier).is_ok() }),
```
sasha8, line 96: Should this be in node.verify()?

consensus/src/dag/rb_handler.rs
sasha8: That is a good point. I can send you a node from a very old round and force you to fetch if it is below the GC level. We should ignore nodes from rounds below the GC level.

consensus/src/dag/rb_handler.rs
ibalajiarun: This is to only verify the missing parents. Ideally, we won't miss any parents in the normal case and don't have to verify at all.

consensus/src/dag/rb_handler.rs
ibalajiarun: It's becoming hard to write unit tests without a verifier trait. I will follow-up to introduce that trait and then we can have more unit tests for this case. Added TODO for now.

sasha8 (approved):

zekun000 (approved):

PR #9586, opened 2023-08-09
Profiling exec benchmark

yunuseozer:
Started implementation of manual cpu profiling start/stop.

### DIFF
--- a/crates/aptos-profiler/src/jeprof.py
+++ b/crates/aptos-profiler/src/jeprof.py
@@ -2,7 +2,6 @@ import subprocess
 import sys
 
 def execute_command(command):
-    print("this is error")
     try:
         output = subprocess.check_output(command, shell=True, stderr=subprocess.STDOUT)
         return output.decode('utf-8').strip()
--- a/crates/aptos-profiler/src/jeprof.py
+++ b/crates/aptos-profiler/src/jeprof.py
@@ -2,6 +2,7 @@ import subprocess
 import sys
 
 def execute_command(command):
+    print("this is error")
     try:
         output = subprocess.check_output(command, shell=True, stderr=subprocess.STDOUT)
         return output.decode('utf-8').strip()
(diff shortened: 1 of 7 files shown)


crates/aptos-profiler/src/jeprof.py
```diff
@@ -2,6 +2,7 @@
 import sys
 
 def execute_command(command):
+    print("this is error")
```
areshand, line 5: remove this line

execution/executor-benchmark/src/main.rs
```diff
@@ -343,11 +360,32 @@ fn main() {
...
 
+    let config = ProfilerConfig::new_with_defaults();
+    let handler = ProfilerHandler::new(config);
+    let mut cpu_profiler = handler.get_cpu_profiler();
+    let mut memory_profiler = handler.get_mem_profiler();
+    let memory_profiling = opt.profiler_opt.memory_profiling;
+    let cpu_profiling = opt.profiler_opt.cpu_profiling;
+
+    if cpu_profiling {
+        let _cpu_prof = cpu_profiler.start_profiling();
```
areshand, line 371: move `let mut cpu_profiler = handler.get_cpu_profiler();` here? similarly for `memory_profiler`

areshand (commented):
almost there, only a few comments

areshand (approved):

grao1991 (approved):
It seems the `Profiler` trait is not a good abstraction..

execution/executor-benchmark/src/main.rs
yunuseozer: had to move it back outside the if (my understanding of your comment was putting it inside) because end_profiling also uses the cpu_profiler object and caused lint errors

areshand: ok, make sense

PR #9480, opened 2023-08-03
Update StateValueMetadata and reenable metadata tracking on devnet

msmouse:
### Description
1. stop tracking the payer
2. implement deposit tracking (the slot fee) 
3. enable tracking on devnet

Implementing the refund next

### Test Plan
1. `test_metadata_tracking()` updated
2. replay-verify https://github.com/aptos-labs/aptos-core/actions/runs/5799529398

PR #9588, opened 2023-08-09
Storage Deletion Refund

msmouse:
### Description

Based on #6652, #6745, #7600, #8219, #9480, #9582 and module events (#9414), implementing AIP-32.

Commits are individually reviewable.

### Test Plan

move e2e tests added

replay-verify: https://github.com/aptos-labs/aptos-core/actions/runs/5922181699

### DIFF
--- a/aptos-move/aptos-gas-meter/src/traits.rs
+++ b/aptos-move/aptos-gas-meter/src/traits.rs
@@ -55,6 +55,15 @@ pub trait GasAlgebra {
         gas_unit_price: FeePerGasUnit,
     ) -> PartialVMResult<()>;
 
+    /// Refunds storage fee.
+    ///
+    /// The amount refunded can be a quantity or an abstract expression containing
+    /// gas parameters.
+    fn refund_storage_fee(
+        &mut self,
+        abstract_amount: impl GasExpression<VMGasParameters, Unit = Octa>,
+    ) -> PartialVMResult<()>;
+
     /// Returns the amount of gas used under the execution category.
     fn execution_gas_used(&self) -> InternalGas;
 
@@ -91,6 +100,12 @@ pub trait AptosGasMeter: MoveGasMeter {
         gas_unit_price: FeePerGasUnit,
     ) -> PartialVMResult<()>;
 
+    /// Refunds storage fee.
+    ///
+    /// Amount is in APT/Octa. Refund is not deducted from the gas meter balance hence no need to
+    /// convert the amount to gas units.
+    fn refund_storage_fee(&mut self, amount: Fee) -> PartialVMResult<()>;
+
     /// Charges an intrinsic cost for executing the transaction.
     ///
     /// The cost stays constant for transactions below a certain size, but will grow proportionally
@@ -105,6 +120,9 @@ pub trait AptosGasMeter: MoveGasMeter {
 
     /// Calculates the storage fee for a state slot allocation.
     fn storage_fee_for_state_slot(&self, op: &WriteOp) -> Fee;
+
+    /// Calculates the storage fee refund for a state slot deallocation.
+    fn storage_fee_refund_for_state_slot(&self, op: &WriteOp) -> Fee;
 
     /// Calculates the storage fee for state bytes.
     fn storage_fee_for_state_bytes(&self, key: &StateKey, op: &WriteOp) -> Fee;
@@ -131,6 +149,7 @@ pub trait AptosGasMeter: MoveGasMeter {
         change_set: &mut VMChangeSet,
         txn_size: NumBytes,
         gas_unit_price: FeePerGasUnit,
+        is_storage_deletion_refund_enabled: bool,
     ) -> VMResult<()> {
         // The new storage fee are only active since version 7.
         if self.feature_version() < 7 {
@@ -145,15 +164,20 @@ pub trait AptosGasMeter: MoveGasMeter {
         }
 
         // Calculate the storage fees.
-        let write_fee = change_set
-            .write_set_iter_mut()
-            .fold(Fee::new(0), |acc, (key, op)| {
-                let slot_fee = self.storage_fee_for_state_slot(op);
-                let bytes_fee = self.storage_fee_for_state_bytes(key, op);
-                Self::maybe_record_storage_deposit(op, slot_fee);
+        let mut write_fee = Fee::new(0);
+        for (key, op) in change_set.write_set_iter_mut() {
+            let slot_fee = self.storage_fee_for_state_slot(op);
+            let refund = self.storage_fee_refund_for_state_slot(op);
+            let bytes_fee = self.storage_fee_for_state_bytes(key, op);
 
-                acc + slot_fee + bytes_fee
-            });
+            Self::maybe_record_storage_deposit(op, slot_fee);
+            if is_storage_deletion_refund_enabled {
+                self.refund_storage_fee(refund)
+                    .map_err(|err| err.finish(Location::Undefined))?;
+            }
+
+            write_fee += slot_fee + bytes_fee
+        }
         let event_fee = change_set.events().iter().fold(Fee::new(0), |acc, event| {
             acc + self.storage_fee_per_event(event)
         });
--- a/aptos-move/aptos-gas-meter/src/traits.rs
+++ b/aptos-move/aptos-gas-meter/src/traits.rs
@@ -75,6 +75,9 @@ pub trait GasAlgebra {
 
     /// Returns the amount of storage fee used.
     fn storage_fee_used(&self) -> Fee;
+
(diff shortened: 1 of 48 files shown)


aptos-move/framework/aptos-framework/sources/coin.move
```diff
@@ -365,6 +365,19 @@ module aptos_framework::coin {
         merge(&mut coin_store.coin, coin);
     }
 
+    /// Deposit the coin balance into the recipient's account without checking if the account is frozen.
+    /// This is for internal use only and doesn't emit an DepositEvent.
+    public(friend) fun force_deposit<CoinType>(account_addr: address, coin: Coin<CoinType>) acquires CoinStore {
```
movekevin, line 370: We technically don't need this as CoinStore<APT> would never be frozen.

msmouse: I need a method that doesn't emit the DepositEvent anyway. If you like I can rename it to `deposit_internal`

movekevin: Are we still planning to do a v2 event for this?

aptos-move/framework/aptos-framework/sources/transaction_validation.move
```diff
@@ -270,15 +270,27 @@ module aptos_framework::transaction_validation {
...
             transaction_fee::collect_fee(gas_payer, transaction_fee_amount);
         } else {
             // Otherwise, just burn the fee.
             // TODO: this branch should be removed completely when transaction fee collection
             // is tested and is fully proven to work well.
-            transaction_fee::burn_fee(gas_payer, transaction_fee_amount);
+            amount_to_burn = transaction_fee_amount;
+        };
+
+        if (amount_to_burn > storage_fee_refunded) {
```
movekevin, line 288: Why not just mint first. Then charge gas?

msmouse: One less function call and the logical is clearer (to me). Lmk you feel strongly about it.

aptos-move/e2e-move-tests/src/tests/storage_refund.rs
```diff
@@ -0,0 +1,128 @@
...
+
+use crate::{assert_success, tests::common::test_dir_path, MoveHarness};
+use aptos_language_e2e_tests::account::Account;
+use aptos_types::on_chain_config::FeatureFlag;
+use aptos_vm::testing::{testing_only::inject_error_once, InjectedError};
+use move_core_types::account_address::AccountAddress;
+use serde::Serialize;
+
+#[test]
+fn test_refunds() {
```
movekevin, line 12: Can you test this on dropping smart tables? cc @lightmark

aptos-move/framework/aptos-framework/sources/transaction_validation.move
movekevin: Not strongly so this is fine :)

aptos-move/framework/aptos-framework/sources/coin.move
msmouse: Yes it's in this implementation already.

movekevin: Missed it. LGTM

aptos-move/e2e-move-tests/src/tests/storage_refund.rs
msmouse: What specifically are we worried about here? It's no different than a normal resource, right?

movekevin: It's potentially a lot more if the table is very large (e.g. 1000 buckets)?

movekevin (approved):

grao1991 (approved):

aptos-move/e2e-move-tests/src/tests/storage_refund.rs
bowenyang007: And the table owner gets the refund? Isn't this kinda broken? e.g. I create a smart table where other ppl write stuff into and pay gas, and I drop the table and get a huge refund.

msmouse: @bowenyang007 Yes, that was the debate (and you were one of the believers of refunding to the deleter). People think it makes more sense than the other way around, partly because as a user you trust the contract developer in the first place.

bowenyang007: Well tables are weird b/c it's not just one person paying gas. It's potentially A LOT of people. How do other chains do it?

msmouse: Just mentioned you in the comparison doc.

movekevin: There are no tables on other chains. Closest to mapping on EVM but you have to delete one slot at a time. It's good to encourage destroying an entire table though such as for airdrop whitelisting as it's just dead data afterward. If the developers leave a way to delete it, they can get a lot more refunds.

aptos-move/framework/aptos-framework/sources/transaction_fee.spec.move
```diff
@@ -169,6 +169,29 @@ spec aptos_framework::transaction_fee {
         } else {
             option::spec_is_none(post_maybe_supply)
         };
+        ensures coin::supply<AptosCoin> == old(coin::supply<AptosCoin>) - amount;
+    }
+
+    spec mint_and_refund(account: address, refund: u64) {
```
davidiw, line 175: wow.....

aptos-move/framework/aptos-framework/sources/transaction_validation.move
```diff
@@ -270,15 +270,27 @@ module aptos_framework::transaction_validation {
...
             // If transaction fees are redistributed to validators, collect them here for
             // later redistribution.
             transaction_fee::collect_fee(gas_payer, transaction_fee_amount);
         } else {
             // Otherwise, just burn the fee.
             // TODO: this branch should be removed completely when transaction fee collection
             // is tested and is fully proven to work well.
-            transaction_fee::burn_fee(gas_payer, transaction_fee_amount);
+            amount_to_burn = transaction_fee_amount;
+        };
```
davidiw, line 286: why not do the ...
```
let amount_to_burn = if (...) {
  ...
  0
} else {
  ..
  transaction_fee_amount
};

**Review by davidiw — APPROVED** [review]

File: aptos-move/framework/aptos-framework/sources/transaction_validation.move (hunk 2 above)
Comment (zekun000) [reply aptos-move/framework/aptos-framework/sources/transaction_validation.move:288]: this is ugly to me with the fee distribution branching above

File: aptos-move/aptos-gas-meter/src/traits.rs [hunk 6]
```diff
@@ -145,15 +167,20 @@ pub trait AptosGasMeter: MoveGasMeter {
...
-            .write_set_iter_mut()
-            .fold(Fee::new(0), |acc, (key, op)| {
-                let slot_fee = self.storage_fee_for_state_slot(op);
-                let bytes_fee = self.storage_fee_for_state_bytes(key, op);
-                Self::maybe_record_storage_deposit(op, slot_fee);
-
-                acc + slot_fee + bytes_fee
-            });
+        let mut write_fee = Fee::new(0);
+        for (key, op) in change_set.write_set_iter_mut() {
```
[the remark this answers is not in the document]
msmouse: `Self::maybe_record_storage_deposit(op, slot_fee);` updates the write op in place according to the current pricing.

aptos-move/framework/aptos-framework/sources/coin.move
```diff
@@ -365,6 +365,19 @@ module aptos_framework::coin {
...
+        assert!(
+            is_account_registered<CoinType>(account_addr),
```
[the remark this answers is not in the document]
msmouse: Refund goes to the gas payer. (The epilogue only has access to her.)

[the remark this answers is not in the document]
davidiw: Actually this is an interesting point... I do think with fee payer, one could create an account without a coinstore :thinking:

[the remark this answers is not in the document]
zekun000: @msmouse told me the refund goes to the gas payer not the txn sender, so we ensure it always has a coin store in prologue

zekun000 (commented):
I think we need to deal with mulsig txn failure cleanup, especially clear_refund

msmouse:
@zekun000  See the second last commit, looks safer now?

PR #9589, opened 2023-08-09
[refactoring] Remove unused genesis check from StateView

georgemitenkov:
### Description

`is_genesis()` was defined in `StateView` but never used apart from a single test with `MockVM`.
But the test set it to false always... Removing it to have better view interfaces.


msmouse (approved):
Love it. Awesome. As long as tests pass.

sitalkedia (approved):

PR #9592, opened 2023-08-10
[docker] Do not overwrite release images unless we want to overwrite them

perryjrandall:
Previously this overwrote a tag, this should be fine except for release tags

This ensures that we dont overwrite a release tag unless we want to by specifying the overwrite option

Test Plan: running action in GH, run script locally

### DIFF
--- a/.github/workflows/copy-images-to-dockerhub.yaml
+++ b/.github/workflows/copy-images-to-dockerhub.yaml
@@ -10,6 +10,9 @@ on:
         required: false
         type: string
         description: the git sha to use for the image tag. If not provided, the git sha of the triggering branch will be used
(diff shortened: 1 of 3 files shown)


.github/workflows/copy-images-to-dockerhub.yaml
```diff
@@ -21,6 +24,9 @@ on:
         required: false
         type: string
         description: the git sha to use for the image tag. If not provided, the git sha of the triggering branch will be used
+      OVERWRITE:
+        type: boolean
+        description: Whether or not we should overwrite images
```
sherry-x, line 29: default should be false?

sherry-x (approved):

PR #9593, opened 2023-08-10
[TS SDK] filter amount > 0 on getTokenOwnersData and include all query fields

0xmaayan:
### Description
- Add a filter amount > 0 on getTokenOwnersData
- Include missing query fields

### DIFF
--- a/ecosystem/typescript/sdk/src/providers/indexer.ts
+++ b/ecosystem/typescript/sdk/src/providers/indexer.ts
@@ -366,6 +366,7 @@ export class IndexerClient {
 
     const whereCondition: any = {
       token_data_id: { _eq: tokenAddress },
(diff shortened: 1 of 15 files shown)


ecosystem/typescript/sdk/src/providers/indexer.ts
```diff
@@ -364,6 +364,7 @@ export class IndexerClient {
 
     const whereCondition: any = {
       token_data_id: { _eq: tokenAddress },
+      amount: { _gt: "0" },
```
banool, line 367: Do we have a test in the e2e tests that confirms that this works as intended?

banool (approved):

kent-white (approved):

PR #9594, opened 2023-08-10
Add ECDSA signatures over NIST-P256 to `aptos-crypto`

mstraka100:
### Description

This PR adds support for ECDSA signatures over NIST-P256, inside of the `aptos-crypto` crate. This is necessary for adding transaction authenticators that interact with popular implementations of the WebAuthn standard. The API for this signature scheme is identical to that of the already-implemented EdDSA over Ed25519 signature scheme. Signatures are guaranteed to be output in canonical form, and signatures not in this form are rejected by verification, to prevent malleability attacks. 

### Test Plan

Unit tests are included in `crates/aptos-crypto/unit_tests/p256_test.rs`. They can be run with `cargo test` in `crates/aptos-crypto`.


crates/aptos-crypto/src/p256/p256_keys.rs
```diff
@@ -0,0 +1,270 @@
...
+    /// The length of the P256PrivateKey
+    pub const LENGTH: usize = P256_PRIVATE_KEY_LENGTH;
+
+    /// Serialize a P256PrivateKey.
+    pub fn to_bytes(&self) -> [u8; P256_PRIVATE_KEY_LENGTH] {
+        self.0.to_bytes().into()
+    }
+
+    /// Deserialize an P256PrivateKey without any validation checks apart from expected key size.
+    fn from_bytes_unchecked(
```
hariria, line 55: I'm kind of curious why the impl for `from_bytes_unchecked` is public for public keys and not for private keys?

types/src/transaction/authenticator.rs
```diff
@@ -46,6 +46,11 @@ pub enum TransactionAuthenticator {
         public_key: Ed25519PublicKey,
         signature: Ed25519Signature,
     },
+    /// Single signature
+    /*P256 {
+        public_key: P256PublicKey,
+        signature: P256PublicKey,
+    },*/
```
hariria, line 53: Remove

crates/aptos-crypto/src/p256/p256_keys.rs
```diff
@@ -0,0 +1,270 @@
+// Copyright © Aptos Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+//! This file implements traits for Ed25519 private keys and public keys.
```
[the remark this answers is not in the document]
mstraka100: Fixed

crates/aptos-crypto/Cargo.toml
```diff
@@ -46,6 +46,10 @@ static_assertions = { workspace = true }
 thiserror = { workspace = true }
 tiny-keccak = { workspace = true }
 x25519-dalek = { workspace = true }
+p256 = { workspace = true }
+# TODO: Remove this?
```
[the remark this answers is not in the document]
mstraka100: Removed

Cargo.toml
```diff
@@ -535,6 +535,10 @@ num-traits = "0.2.15"
 once_cell = "1.10.0"
 ouroboros = "0.15.6"
 owo-colors = "3.5.0"
+# TODO: Do we actually use serde?
```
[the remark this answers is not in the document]
mstraka100: Removed the serde feature

crates/aptos-crypto/src/p256/p256_keys.rs
mstraka100: It's pub(crate) for public keys so it can be used in unit tests. Making it fully public would be a bad idea

crates/aptos-crypto/src/p256/mod.rs
alinush, line 1: Please rename this directory to `ecdsa_p256`.

crates/aptos-crypto/src/p256/p256_keys.rs
```diff
@@ -0,0 +1,270 @@
...
+        match p256::ecdsa::SigningKey::from_slice(bytes) {
+            Ok(p256_secret_key) => Ok(P256PrivateKey(p256_secret_key)),
+            Err(_) => Err(CryptoMaterialError::DeserializationError),
+        }
+    }
+
+    /// Private function aimed at minimizing code duplication between sign
+    /// methods of the SigningKey implementation. This should remain private.
+    fn sign_arbitrary_message(&self, message: &[u8]) -> P256Signature {
+        let secret_key: &p256::ecdsa::SigningKey = &self.0;
```
alinush, line 67: Surprising that you need to specify the types here.

crates/aptos-crypto/src/p256/p256_keys.rs
```diff
@@ -0,0 +1,270 @@
...
+        let canonical_sig = P256Signature::make_canonical(&sig);
+
+        canonical_sig
+    }
+}
+
+impl P256PublicKey {
+    /// Serialize a P256PublicKey.
+    pub fn to_bytes(&self) -> [u8; P256_PUBLIC_KEY_LENGTH] {
+        // The RustCrypto P256 `to_sec1_bytes` call here should never return an array of the wrong length and cause a panic
```
alinush, line 78: Define "should"? Under what circumstances does it panic here?

crates/aptos-crypto/src/p256/p256_keys.rs
```diff
@@ -0,0 +1,270 @@
...
+        P256PrivateKey::sign_arbitrary_message(self, message)
+    }
+}
+
+impl Uniform for P256PrivateKey {
+    fn generate<R>(rng: &mut R) -> Self
+    where
+       R: ::rand::RngCore + ::rand::CryptoRng + ::rand_core::CryptoRng + ::rand_core::RngCore,
+    {
+        let mut bytes: [u8; P256_PRIVATE_KEY_LENGTH] = Default::default();
```
alinush, line 127: You do not need the `Default` call.

crates/aptos-crypto/src/p256/p256_keys.rs
```diff
@@ -0,0 +1,270 @@
...
+
+        canonical_sig
+    }
+}
+
+impl P256PublicKey {
+    /// Serialize a P256PublicKey.
+    pub fn to_bytes(&self) -> [u8; P256_PUBLIC_KEY_LENGTH] {
+        // The RustCrypto P256 `to_sec1_bytes` call here should never return an array of the wrong length and cause a panic
+        (*self.0.to_sec1_bytes()).try_into().unwrap()
```
alinush, line 79: What is the difference between `to_sec1_bytes` and `to_bytes` for a ` p256::ecdsa::VerifyingKey`?

crates/aptos-crypto/src/p256/p256_keys.rs
```diff
@@ -0,0 +1,270 @@
...
+
+impl Length for P256PublicKey {
+    fn length(&self) -> usize {
+        P256_PUBLIC_KEY_LENGTH
+    }
+}
+
+impl ValidCryptoMaterial for P256PublicKey {
+    fn to_bytes(&self) -> Vec<u8> {
+        self.0.to_sec1_bytes().to_vec()
```
alinush, line 245: Should probably just be `self.to_bytes()` which in turn calls `to_sec1_bytes()`, for easier, single-place updates to the code.

crates/aptos-crypto/src/p256/p256_sigs.rs
```diff
@@ -0,0 +1,191 @@
...
+        };
+        let s = self.0.s();
+        let r = self.0.r();
+        // NonZeroScalar::try_from throws an error on being passed the curve order, so we use the
+        // order minus one, then add one later to compute (n-1) - s + 1 = n-s
+        let order_minus_one = NonZeroScalar::try_from(&ORDER_MINUS_ONE[..]).unwrap();
+        let one = NonZeroScalar::from_uint(<NistP256 as Curve>::Uint::ONE).unwrap();
+        // Dereferencing a NonZeroScalar makes it a Scalar, which implements subtraction
+        let new_s = *order_minus_one - *s + *one;
+        let new_s_nonzero = NonZeroScalar::new(new_s).unwrap();
```
alinush, line 104: This is `n-s`, which is guaranteed to be non-zero because the signing returns a non-zero `s`, correct?

alinush, line 104: Also, curious, there isn't a `negate()` function that you can call here, instead of doing this arithmetic on your own?

crates/aptos-crypto/src/p256/p256_sigs.rs
```diff
@@ -0,0 +1,191 @@
...
+        let new_sig = p256::ecdsa::Signature::from_scalars(&r, &new_s_nonzero).unwrap();
```
alinush, line 105: Lastly, in case I forget later, is this malleability code tested?

crates/aptos-crypto/src/p256/p256_sigs.rs
```diff
@@ -0,0 +1,191 @@
...
+    /// Check if S < ORDER_HALF to capture invalid signatures.
+    fn check_s_lt_order_half(s: &[u8]) -> bool {
+        for i in 0..32 {
+            match s[i].cmp(&ORDER_HALF[i]) {
+                Ordering::Less => return true,
+                Ordering::Greater => return false,
+                _ => {},
+            }
+        }
+        // As this stage S == ORDER_HALF which implies a non canonical S.
```
alinush, line 85: Typo: **At** this stage

crates/aptos-crypto/src/p256/p256_sigs.rs
```diff
@@ -0,0 +1,191 @@
...
+
+impl private::Sealed for P256Signature {}
+
+impl P256Signature {
+    /// The length of the P256Signature
+    pub const LENGTH: usize = P256_SIGNATURE_LENGTH;
+
+    /// Serialize an P256Signature.
+    pub fn to_bytes(&self) -> [u8; P256_SIGNATURE_LENGTH] {
+        // The RustCrypto P256 `to_bytes` call here should never return a byte array of the wrong length
```
alinush, line 35: Should or never does? Under what conditions does it return a byte array of the wrong length?

crates/aptos-crypto/src/unit_tests/p256_test.rs
```diff
@@ -0,0 +1,204 @@
...
+        );
+        prop_assert!(check);
+    }
+
+    #[test]
+    fn test_keys_encode(keypair in uniform_keypair_strategy::<P256PrivateKey, P256PublicKey>()) {
+        {
+            let encoded = keypair.private_key.to_encoded_string().unwrap();
+            // Hex encoding of a 64-bytes key is 128 (2 x 64) characters.
+            prop_assert_eq!(2 + 2 * P256_PRIVATE_KEY_LENGTH, encoded.len());
```
alinush, line 41: Where are the extra 2 bytes from?

crates/aptos-crypto/src/unit_tests/p256_test.rs
```diff
@@ -0,0 +1,204 @@
...
+        message in random_serializable_struct(),
+        keypairs in proptest::array::uniform10(uniform_keypair_strategy::<P256PrivateKey, P256PublicKey>())
+    ) {
+        let mut signatures: Vec<(P256PublicKey, P256Signature)> = keypairs.iter().map(|keypair| {
+            (keypair.public_key.clone(), keypair.private_key.sign(&message).unwrap())
+        }).collect();
+        prop_assert!(P256Signature::batch_verify(&message, signatures.clone()).is_ok());
+        // We swap message and signature for the last element,
+        // resulting in an incorrect signature
+        let (key, _sig) = signatures.pop().unwrap();
```
alinush, line 65: `key` should be `pk`

crates/aptos-crypto/src/unit_tests/p256_test.rs
```diff
@@ -0,0 +1,204 @@
...
+            prop_assert_eq!(Some(keypair.public_key), decoded.ok());
+        }
+    }
+
+    #[test]
+    fn test_batch_verify(
+        message in random_serializable_struct(),
+        keypairs in proptest::array::uniform10(uniform_keypair_strategy::<P256PrivateKey, P256PublicKey>())
+    ) {
+        let mut signatures: Vec<(P256PublicKey, P256Signature)> = keypairs.iter().map(|keypair| {
```
alinush, line 59: `signatures` should be `pks_and_sigs`

crates/aptos-crypto/src/unit_tests/p256_test.rs
```diff
@@ -0,0 +1,204 @@
...
+        prop_assert_ne!(serialized_old, serialized);
+
+        // Check that malleable signatures will pass verification and deserialization in the RustCrypto
+        // p256 crate.
+        // Construct the corresponding RustCrypto p256 public key.
+        let rustcrypto_public_key = p256::ecdsa::VerifyingKey::from_sec1_bytes(
+            &keypair.public_key.to_bytes()
+        ).unwrap();
+
+        // Construct the corresponding RustCrypto p256 Signature. This signature is malleable.
```
alinush, line 164: Technically, should say "This signature has been **mauled**". "Malleability" is a property of the verification algorithm, not the signature itself.

crates/aptos-crypto/src/unit_tests/p256_test.rs
```diff
@@ -0,0 +1,204 @@
...
+        // RustCrypto p256 will deserialize the malleable
+        // signature. It does not detect it.
+        prop_assert!(rustcrypto_sig.is_ok());
+
+        let msg_bytes = bcs::to_bytes(&message);
+        prop_assert!(msg_bytes.is_ok());
+
+        let rustcrypto_sig = rustcrypto_sig.unwrap();
+        // RustCrypto p256 verify will NOT accept the mauled signature
+        prop_assert!(rustcrypto_public_key.verify(msg_bytes.as_ref().unwrap(), &rustcrypto_sig).is_err());
```
alinush, line 176: But if the malleability check is done in `p256` why are we redoing it in our crate?

crates/aptos-crypto/src/p256/p256_sigs.rs
```diff
@@ -0,0 +1,191 @@
...
+        public_key: &P256PublicKey,
+    ) -> Result<()> {
+        Self::verify_arbitrary_msg(self, &signing_message(message)?, public_key)
+    }
+
+    /// Checks that `self` is valid for an arbitrary &[u8] `message` using `public_key`.
+    /// Outside of this crate, this particular function should only be used for native signature
+    /// verification in Move.
+    ///
+    /// This function will check both the signature and `public_key` for small subgroup attacks.
```
alinush, line 132: `secp256r1` is [a prime-order group](https://safecurves.cr.yp.to/www.secg.org/sec2-v2.pdf). There are no small subgroup attacks.

Can you change this to say "This function will check if the signature has been mauled."?

crates/aptos-crypto/src/p256/p256_sigs.rs
```diff
@@ -0,0 +1,191 @@
...
+    /// Check for correct size and third-party based signature malleability issues.
+    /// This method is required to ensure that given a valid signature for some message under some
+    /// key, an attacker cannot produce another valid signature for the same message and key.
+    ///
+    /// We use the technique described in
+    /// [BIP146](https://github.com/bitcoin/bips/blob/master/bip-0146.mediawiki) to prevent
+    /// malleability of ECDSA signatures. Signatures comprise elements {R, S}, and S can be
+    /// enforced to be of canonical form by ensuring it is less than the order of the P256 curve
+    /// divided by 2. If this is not done, a value S > n/2 can be replaced by S' = n - S to form another distinct valid
+    /// signature, where n is the curve order. This check is not performed by the RustCrypto P256 library
```
alinush, line 64: Here you contradict the comment you made in the tests, where you say it does check...

alinush (changes requested):
Did a pass! Looks alright, except for the confusion on whether the `p256` library does [not] check for malleability & some of the _"should not fail"_ type of comments.

Cargo.toml
```diff
@@ -648,7 +652,7 @@ walkdir = "2.3.3"
 warp = { version = "0.3.5", features = ["tls"] }
 warp-reverse-proxy = "1.0.0"
 which = "4.2.5"
-x25519-dalek = "1.2.0"
+x25519-dalek = { git = "https://github.com/aptos-labs/x25519-dalek" }
```
gedigi, line 655: why are we forking dalek? how do we plan to maintain it current with upstream?

crates/aptos-crypto/src/p256/p256_keys.rs
```diff
@@ -0,0 +1,270 @@
...
+        let serialized: &[u8] = &(self.to_bytes());
+        P256PrivateKey::try_from(serialized).unwrap()
+    }
+}
+
+/// A P256 public key
+#[derive(DeserializeKey, Clone, SerializeKey)]
+pub struct P256PublicKey(pub(crate) p256::ecdsa::VerifyingKey);
+
+impl private::Sealed for P256PublicKey {}
```
gedigi, line 43: should this be in `traits.rs` where other implementations are also specified?

Cargo.toml
hariria: For context, the upstream branch has a very old `zeroize` dependency. Many people have had to fork in order to make it work with newer dependencies (see this issue i created in 2022 before starting Aptos for example -> https://github.com/dalek-cryptography/curve25519-dalek/issues/388). I think we can pretty easily sync changes with the upstream but will let @mstraka100 or @alinush chime in here

gedigi: have you considered submitting a PR upstream to update zeroize?

crates/aptos-crypto/src/unit_tests/p256_test.rs
mstraka100: Fixed, the non-canonical signature was being computed incorrectly

crates/aptos-crypto/src/p256/p256_sigs.rs
mstraka100: Changed to "Checks for and rejects non-canonical signatures (r,s) where s > (n/2), where n is the group order"

crates/aptos-crypto/src/p256/p256_sigs.rs
mstraka100: Fixed, was computing the mauled non-canonical signature incorrectly

Cargo.toml
mstraka100: There actually is an update to x25519-dalek (2.0.0-rc.3) that updates zeroize, but it's still a release candidate and has not been finalized yet

gedigi: 2.0.0 seems stable now: https://docs.rs/x25519-dalek/2.0.0/x25519_dalek/

crates/aptos-crypto/src/p256/p256_keys.rs
mstraka100: Hmm yep, it's not needed. Removed. Must have been interacting with some other compiler issue I fixed already.

crates/aptos-crypto/src/p256/p256_keys.rs
mstraka100: `let mut bytes: [0u8; P256_PRIVATE_KEY_LENGTH];` gives a compiler error

crates/aptos-crypto/src/p256/p256_keys.rs
mstraka100: I don't follow - `self.to_bytes()` is what is being implemented in this function

crates/aptos-crypto/src/p256/p256_sigs.rs
mstraka100: Fixed

crates/aptos-crypto/src/unit_tests/p256_test.rs
mstraka100: Fixed

crates/aptos-crypto/src/unit_tests/p256_test.rs
mstraka100: Fixed

crates/aptos-crypto/src/p256/mod.rs
mstraka100: Changed

Cargo.toml
mstraka100: I believe there are other parts of our codebase that use our dalek fork anyways, so I'd propose updating it in a separate PR if we do update it

crates/aptos-crypto/src/unit_tests/p256_test.rs
mstraka100: It's for the `0x` prepended to the hex encoding. Changed the comment to explain this

crates/aptos-crypto/src/p256/p256_keys.rs
mstraka100: `P256PublicKey`? No, `traits.rs` only contains traits that are agnostic to the signature scheme used, while these structs are specific to ECDSA over P256

mstraka100: Oh you meant `Sealed`. I think that makes sense to put it in `traits.rs` with the rest

mstraka100: Moved Sealed to `traits.rs`

crates/aptos-crypto/src/p256/p256_sigs.rs
mstraka100: Yeah you're right. Changed the code

mstraka100: Yes

crates/aptos-crypto/src/p256/p256_keys.rs
mstraka100: It should never panic. The length of the encoding ultimately gets its length from this function which is constant (modulo difference between compressed/uncompressed points): https://github.com/RustCrypto/formats/blob/master/sec1/src/point.rs#L520

It uses the sec1 standard which has constant size bytes lengths of encodings for elliptic curves, per section 2.3.3: https://www.secg.org/sec1-v2.pdf

Possibly cosmic radiation could change the returned length and cause a panic

crates/aptos-crypto/src/p256/p256_sigs.rs
mstraka100: See related reply in https://github.com/aptos-labs/aptos-core/pull/9594#discussion_r1341601992

crates/aptos-crypto/src/p256/p256_keys.rs
alinush: Yes, but here you are implementing `ValidCryptoMaterial::to_bytes`.

You already have a `P256PublicKey::to_bytes`, which I'm trying to suggest should be called here:

```
impl P256PublicKey {
    /// Serialize a P256PublicKey.
    pub fn to_bytes(&self) -> [u8; P256_PUBLIC_KEY_LENGTH] {
        // The RustCrypto P256 `to_sec1_bytes` call here should never return an array of the wrong length and cause a panic
        (*self.0.to_sec1_bytes()).try_into().unwrap()
    }
```

You can probably call it via `P256PublicKey::to_bytes(&self)`.

Otherwise, you have two places in which you call `to_sec1_bytes`, and when you change one place you might forget the other.

crates/aptos-crypto/src/ecdsa_p256/mod.rs
alinush, line 1: For symmetry with David's `sep256k1_ecdsa` crate, should we rename this to `secp256r1_ecdsa` or to `p256_ecdsa`?

crates/aptos-crypto/src/p256/p256_keys.rs
alinush: Ugh. Because `:` should be a `=`...

crates/aptos-crypto/src/p256/p256_keys.rs
mstraka100: You are correct - changed the line to `self.to_bytes().to_vec()`

crates/aptos-crypto/src/ecdsa_p256/mod.rs
mstraka100: Don't see why not. Changed the name

Cargo.toml
```diff
@@ -649,7 +653,7 @@ walkdir = "2.3.3"
...
```
zjma, line 656: should we pin to a commit? just to be safer...

Cargo.toml
```diff
@@ -546,6 +546,10 @@ once_cell = "1.10.0"
 ordered-float = "3.9.1"
 ouroboros = "0.15.6"
 owo-colors = "3.5.0"
+# TODO: Do we actually use serde?
+p256 = { version = "0.13.2" } #, features = ["serde"] }
```
zjma, line 550: According to https://docs.rs/p256/latest/p256 ...
```
The elliptic curve arithmetic contained in this crate has never been independently audited!
```

zjma: taking the risk?

mstraka100: We also tried the [openssl crate](https://crates.io/crates/openssl) for p256 but it was nearly impossible to figure out where their rng was coming from in their underlying C code. Generally the `RustCrypto` libraries seem to be well-regarded in the blockchain cryptography space. We should still discuss whether we want to have an audit done ourselves on their code

crates/aptos-crypto/src/p256_ecdsa/p256_keys.rs
```diff
@@ -0,0 +1,266 @@
...
+/// A P256 public key
+#[derive(DeserializeKey, Clone, SerializeKey)]
+pub struct P256PublicKey(pub(crate) p256::ecdsa::VerifyingKey);
+
+impl P256PrivateKey {
+    /// The length of the P256PrivateKey
+    pub const LENGTH: usize = P256_PRIVATE_KEY_LENGTH;
+
+    /// Serialize a P256PrivateKey.
+    pub fn to_bytes(&self) -> [u8; P256_PRIVATE_KEY_LENGTH] {
```
zjma, line 46: Is this some standardized serialization format (which I believe is not included in [NIST SP 800-186](https://csrc.nist.gov/publications/detail/sp/800-186/final) ), or simply "whatever p256 uses"?

zjma: Same comments for `P256PublicKey` and `P256Signature`.

crates/aptos-crypto/src/p256_ecdsa/p256_sigs.rs
```diff
@@ -0,0 +1,188 @@
...
+        let encoded_signature = self.to_bytes();
+        state.write(&encoded_signature);
+    }
+}
+
+impl TryFrom<&[u8]> for P256Signature {
+    type Error = CryptoMaterialError;
+
+    fn try_from(bytes: &[u8]) -> std::result::Result<P256Signature, CryptoMaterialError> {
+        P256Signature::check_s_malleability(bytes)?;
```
zjma, line 164: @alinush in David's PR you mentioned this has to happen in verification, not deserialization?

zjma: Oh, just saw that's in verification too. It's still unclear to me high-s should be a deserialization failure, or verification failure, or it simply doesn't matter :)

crates/aptos-crypto/src/p256_ecdsa/p256_keys.rs
```diff
@@ -0,0 +1,266 @@
...
+    ) -> std::result::Result<P256PrivateKey, CryptoMaterialError> {
+        match p256::ecdsa::SigningKey::from_slice(bytes) {
+            Ok(p256_secret_key) => Ok(P256PrivateKey(p256_secret_key)),
+            Err(_) => Err(CryptoMaterialError::DeserializationError),
+        }
+    }
+
+    /// Private function aimed at minimizing code duplication between sign
+    /// methods of the SigningKey implementation. This should remain private.
+    fn sign_arbitrary_message(&self, message: &[u8]) -> P256Signature {
```
zjma, line 62: Add some notes for the hash function applied to the message, if available?

crates/aptos-crypto/src/p256_ecdsa/p256_sigs.rs
alinush: Oh, indeed, this is just a paranoid check & could be avoided.

Cargo.toml
mstraka100: Done

Cargo.toml
mstraka100: Discussed this further with @gedigi and decided an audit on `RustCrypto` isn't necessary for now

crates/aptos-crypto/src/p256_ecdsa/p256_keys.rs
mstraka100: Added. The hash function used is SHA2-256

crates/aptos-crypto/src/p256_ecdsa/p256_keys.rs
mstraka100: Added comments. They all use SEC1

crates/aptos-crypto/src/p256_ecdsa/p256_keys.rs
```diff
@@ -0,0 +1,270 @@
...
+        P256PrivateKey::sign_arbitrary_message(self, message)
+    }
+}
+
+impl Uniform for P256PrivateKey {
+    fn generate<R>(rng: &mut R) -> Self
+    where
+       R: ::rand::RngCore + ::rand::CryptoRng + ::rand_core::CryptoRng + ::rand_core::RngCore,
+    {
+        let mut bytes = [0u8; P256_PRIVATE_KEY_LENGTH];
```
zjma, line 127: technically non-uniform?

zjma: i mean L128 and L129

zjma (approved):

crates/aptos-crypto/src/p256_ecdsa/p256_keys.rs
mstraka100: You're right, this is insecure. I've changed it to sample a byte array twice the size of the private key length, then reduce by the field order, per @alinush's suggestion

crates/aptos-crypto/src/p256_ecdsa/p256_keys.rs
```diff
@@ -0,0 +1,275 @@
...
+    // Uses a hack to get around the incompatability of the `aptos-crypto` RngCore trait and the
+    // `RustCrypto` RngCore trait
+    fn generate<R>(rng: &mut R) -> Self
+    where
+        R: ::rand::RngCore + ::rand::CryptoRng + ::rand_core::CryptoRng + ::rand_core::RngCore,
+    {
+        let mut bytes = [0u8; P256_PRIVATE_KEY_LENGTH * 2];
+        rng.fill_bytes(&mut bytes);
+        let bignum = BigUint::from_bytes_be(&bytes[..]);
+        let order = BigUint::from_bytes_be(&ORDER);
```
alinush, line 132: I hope the LE/BE encoding of `ORDER` here is right. If we get these flipped, the key space from which we sample could be much smaller and disaster will ensue.

Are there any tests we can do here?

crates/aptos-crypto/src/p256_ecdsa/mod.rs
```diff
@@ -0,0 +1,64 @@
...
+    0xBC, 0xE6, 0xFA, 0xAD, 0xA7, 0x17, 0x9E, 0x84, 0xF3, 0xB9, 0xCA, 0xC2, 0xFC, 0x63, 0x25, 0x51,
+];
+
+/// The value (q-1)/2, where q is the order of P256 as defined in [NIST SP 800-186](https://csrc.nist.gov/publications/detail/sp/800-186/final).
+/// Computed with the following SageMath code:
+///
+/// # Curve order
+/// qq = 0xFFFFFFFF00000000FFFFFFFFFFFFFFFFBCE6FAADA7179E84F3B9CAC2FC632551
+/// q_half = (qq-1)/2
+pub const ORDER_HALF: [u8; 32] = [
```
alinush, line 51: Clarify if this is in LE or BE encoding...

crates/aptos-crypto/src/p256_ecdsa/mod.rs
```diff
@@ -0,0 +1,64 @@
...
+
+/// The length in bytes of the P256PrivateKey
+pub const P256_PRIVATE_KEY_LENGTH: usize = 32;
+/// The length in bytes of the P256PublicKey
+pub const P256_PUBLIC_KEY_LENGTH: usize = 65;
+/// The length in bytes of the P256Signature
+pub const P256_SIGNATURE_LENGTH: usize = 64;
+
+/// The order of P256 as defined in [NIST SP 800-186](https://csrc.nist.gov/publications/detail/sp/800-186/final)
+const ORDER: [u8; 32] = [
```
alinush, line 40: Clarify if this is in LE or BE encoding...

crates/aptos-crypto/src/p256_ecdsa/p256_keys.rs
mstraka100: Yeah I'm trying to figure this out. The encoding for `RustCrypto` is surprisingly opaque

crates/aptos-crypto/src/p256_ecdsa/p256_keys.rs
```diff
@@ -0,0 +1,275 @@
...
+        let remainder = bignum.mod_floor(&order);
+        P256PrivateKey(p256::ecdsa::SigningKey::from_slice(&remainder.to_bytes_le()).unwrap())
```
alinush, line 134: Will this panic if the `bytes` do not encode a valid scalar in that field?

alinush: Also, why not call `P256PrivateKey::from_bytes_unchecked` here?

crates/aptos-crypto/src/p256_ecdsa/p256_keys.rs
```diff
@@ -0,0 +1,275 @@
...
+impl PartialEq<Self> for P256PrivateKey {
+    fn eq(&self, other: &Self) -> bool {
+        self.to_bytes() == other.to_bytes()
+    }
+}
+
+impl Eq for P256PrivateKey {}
+
+// We could have a distinct kind of validation for the PrivateKey: e.g., checking the derived
+// PublicKey is valid?
```
alinush, line 147: Not sure what you mean... why would the public key not be valid if computed using the `RustCrypto` library?

crates/aptos-crypto/src/p256_ecdsa/p256_keys.rs
mstraka100: Ok converting the array `[0,0,0,1]` to a `BigUint` with `from_bytes_be` then printing the result gives `[1]`, while converting the same array and printing the resulting `BigUint` with `from_bytes_le` gives `[16777216]`. So `_be` is correct here since the `ORDER` constant array is also has the more important integers on the left of the less important ones

crates/aptos-crypto/src/p256_ecdsa/mod.rs
mstraka100: Big-endian. Comment added

crates/aptos-crypto/src/p256_ecdsa/mod.rs
mstraka100: Big-endian. Comment added

crates/aptos-crypto/src/p256_ecdsa/p256_keys.rs
mstraka100: You mean if `remainder` doesn't encode a valid scalar in the field? How would that happen when remainder is the result of `bytes` being taken mod the scalar field order?

mstraka100: Changed to `from_bytes_unchecked`

mstraka100: Check my logic on deciding `to_bytes_be` is correct here - I defined a constant `ORDER_TOO_MUCH` that is identical to `ORDER`, except the least significant (rightmost) byte is `0xFF`. I then convert it to a `BigUint` using `from_bytes_be` (this is correct from the tests I did on `ORDER`). Then I convert the BigUint into bytes again using `to_bytes_(le|be)`, and call `P256PrivateKey::from_bytes_unchecked` on the result. `_be` panics most of the p256 tests, which means it is the right option, as it correctly results in a value which is greater than the field order

crates/aptos-crypto/src/p256_ecdsa/p256_keys.rs
mstraka100: I agree. Removed the comment

crates/aptos-crypto/src/p256_ecdsa/p256_keys.rs
mstraka100: Added a unit test that makes this clearer. See https://github.com/aptos-labs/aptos-core/pull/9594/commits/a1506e2dfccaab44c7e1930c1e11b336d7f3e767

crates/aptos-crypto/src/p256_ecdsa/p256_keys.rs
alinush: Indeed, [`ORDER` is in BE](https://github.com/RustCrypto/elliptic-curves/blob/60664a93bc17c866974f145c04905d2bab87e555/p256/src/lib.rs#L108).

crates/aptos-crypto/src/p256_ecdsa/p256_keys.rs
```diff
@@ -0,0 +1,273 @@
...
+    fn generate<R>(rng: &mut R) -> Self
+    where
+        R: ::rand::RngCore + ::rand::CryptoRng + ::rand_core::CryptoRng + ::rand_core::RngCore,
+    {
+        let mut bytes = [0u8; P256_PRIVATE_KEY_LENGTH * 2];
+        rng.fill_bytes(&mut bytes);
+        let bignum = BigUint::from_bytes_be(&bytes[..]);
+        let order = BigUint::from_bytes_be(&ORDER);
+        let remainder = bignum.mod_floor(&order);
+        P256PrivateKey::from_bytes_unchecked(&remainder.to_bytes_be()).unwrap()
```
alinush, line 134: Also, documenting for posterity, I traced down the deserialization code to [this line](https://github.com/RustCrypto/traits/blob/elliptic-curve-v0.13.6/elliptic-curve/src/field.rs#L37), which does indeed use BE.

PR #9595, opened 2023-08-10
[TS SDK v2] Add Hex and HexInput types

banool:
### Description
This PR defines the `Hex` class. This class provides a way for working with non-account-address hex data, e.g. txn hashes. It provides methods for creating instances of the Hex class as well as static methods to make it easy to quickly work with hex data. This is explained further in the file level comment.

### Test Plan
```
cd ecosystem/typescript/sdk_v2
npx jest -- tests/unit/hex.test.ts
```

### DIFF
--- a/ecosystem/typescript/sdk_v2/src/core/common.ts
+++ b/ecosystem/typescript/sdk_v2/src/core/common.ts
@@ -1,0 +1,40 @@
+// Copyright © Aptos Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+/**
+ * This error is used to explain why parsing failed.
+ */
+export class ParsingError<T> extends Error {
+  /**
+   * This provides a programmatic way to access why parsing failed. Downstream devs
+   * might want to use this to build their own error messages if the default error
+   * messages are not suitable for their use case. This should be an enum.
+   */
+  public invalidReason: T;
+
+  constructor(message: string, invalidReason: T) {
+    super(message);
+    this.invalidReason = invalidReason;
+  }
+}
+
+/**
+ * Whereas ParsingError is thrown when parsing fails, e.g. in a fromString function,
+ * this type is returned from "defensive" functions like isValid.
+ */
+export type ParsingResult<T> = {
+  /**
+   * True if valid, false otherwise.
+   */
+  valid: boolean;
+
+  /*
+   * If valid is false, this will be a code explaining why parsing failed.
+   */
+  invalidReason?: T;
+
+  /*
+   * If valid is false, this will be a string explaining why parsing failed.
+   */
+  invalidReasonMessage?: string;
+};
--- a/ecosystem/typescript/sdk_v2/src/core/hex.ts
+++ b/ecosystem/typescript/sdk_v2/src/core/hex.ts
@@ -1,0 +1,165 @@
+// Copyright © Aptos Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+import { bytesToHex, hexToBytes } from "@noble/hashes/utils";
+import { HexInput } from "../types";
+import { ParsingError, ParsingResult } from "./common";
+
+/**
+ * This enum is used to explain why parsing might have failed.
+ */
+export enum HexInvalidReason {
+  TOO_SHORT = "too_short",
+  INVALID_LENGTH = "invalid_length",
+  INVALID_HEX_CHARS = "invalid_hex_chars",
+}
+
+/**
+ * NOTE: Do not use this class when working with account addresses, use AccountAddress.
+ *
+ * NOTE: When accepting hex data as input to a function, prefer to accept HexInput and
+ * then use the static helper methods of this class to convert it into the desired
+ * format. This enables the greatest flexibility for the developer.
+ *
+ * Hex is a helper class for working with hex data. Hex data, when represented as a
+ * string, generally looks like this, for example: 0xaabbcc, 45cd32, etc.
+ *
+ * You might use this class like this:
+ *
+ * ```ts
+ * getTransactionByHash(txnHash: HexInput): Promise<Transaction> {
+ *   const txnHashString = Hex.fromHexInput({ hexInput: txnHash }).toString();
+ *   return await getTransactionByHashInner(txnHashString);
+ * }
+ * ```
+ *
+ * This call to `Hex.fromHexInput().toString()` converts the HexInput to a hex string
+ * with a leading 0x prefix, regardless of what the input format was.
+ *
+ * These are some other ways to chain the functions together:
+ * - `Hex.fromString({ hexInput: "0x1f" }).toUint8Array()`
+ * - `new Hex({ data: [1, 3] }).toStringWithoutPrefix()`
+ */
+export class Hex {
+  private data: Uint8Array;
+
+  /**
+   * Create a new Hex instance from a Uint8Array.
+   *
+   * @param hex Uint8Array
+   */
+  constructor(args: { data: Uint8Array }) {
+    this.data = args.data;
+  }
+
+  // ===
+  // Methods for representing an instance of Hex as other types.
+  // ===
+
+  /**
+   * Get the inner hex data. The inner data is already a Uint8Array so no conversion
+   * is taking place here, it just returns the inner data.
+   *
+   * @returns Hex data as Uint8Array
+   */
+  toUint8Array(): Uint8Array {
+    return this.data;
+  }
+
+  /**
+   * Get the hex data as a string without the 0x prefix.
+   *
+   * @returns Hex string without 0x prefix
+   */
+  toStringWithoutPrefix(): string {
+    return bytesToHex(this.data);
+  }
+
+  /**
+   * Get the hex data as a string with the 0x prefix.
+   *
+   * @returns Hex string with 0x prefix
+   */
+  toString(): string {
+    return `0x${this.toStringWithoutPrefix()}`;
+  }
+
+  // ===
+  // Methods for creating an instance of Hex from other types.
+  // ===
+
+  /**
+   * Static method to convert a hex string to Hex
+   *
+   * @param str A hex string, with or without the 0x prefix
+   *
+   * @returns Hex
+   */
+  static fromString(args: { str: string }): Hex {
+    let input = args.str;
+
+    if (input.startsWith("0x")) {
+      input = input.slice(2);
+    }
+
+    if (input.length === 0) {
+      throw new ParsingError(
+        "Hex string is too short, must be at least 1 char long, excluding the optional leading 0x.",
+        HexInvalidReason.TOO_SHORT,
+      );
+    }
+
+    if (input.length % 2 !== 0) {
+      throw new ParsingError("Hex string must be an even number of hex characters.", HexInvalidReason.INVALID_LENGTH);
+    }
+
+    try {
+      return new Hex({ data: hexToBytes(input) });
+    } catch (e) {
+      const error = e as Error;
+      throw new ParsingError(
+        `Hex string contains invalid hex characters: ${error.message}`,
+        HexInvalidReason.INVALID_HEX_CHARS,
+      );
+    }
+  }
+
+  /**
+   * Static method to convert an instance of HexInput to Hex
+   *
+   * @param str A HexInput (string or Uint8Array)
+   *
+   * @returns Hex
+   */
+  static fromHexInput(args: { hexInput: HexInput }): Hex {
+    if (args.hexInput instanceof Uint8Array) return new Hex({ data: args.hexInput });
+    return Hex.fromString({ str: args.hexInput });
+  }
+
+  // ===
+  // Methods for checking validity.
+  // ===
+
+  /**
+   * Check if the string is valid hex.
+   *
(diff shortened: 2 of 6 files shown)


ecosystem/typescript/sdk_v2/src/types/hex.ts
```diff
@@ -0,0 +1,91 @@
+import { bytesToHex, hexToBytes } from "@noble/hashes/utils";
+import { HexInput } from ".";
```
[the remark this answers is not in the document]
hardsetting: > but I think in general it would be better to put the types in a different files, maybe types.ts

oh I see that `hex.ts` is already under a `types` folder.
At least to me, a type in Typescript is either an `interface` or a `type`, so maybe we could move `Hex` to a `utils` folder since it's a utility class?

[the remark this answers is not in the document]
banool: @0xmaayan what do you think? I don't really mind but to me utils is just assorted utility functions, not a place for core "types" (classes in this case) that we use throughout the whole SDK.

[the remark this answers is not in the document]
hardsetting: We can also just move `HexInput` under `hex.ts` :D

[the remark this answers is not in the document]
0xmaayan: yes the idea is to have all types or interfaces on a types.ts file
```
types // holds local types
```

Let's move this class into utils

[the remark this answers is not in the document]
0xmaayan: yes the idea is to have all types and interfaces under `types.ts`
```
types // holds local types
```

let's have this class under `utils` folder

ecosystem/typescript/sdk_v2/src/types/hex.ts
```diff
@@ -0,0 +1,91 @@
...
+    const hex = Hex.fromHexInput(args);
+    return hex.toUint8Array();
+  }
+
+  static hexInputToString(args: { hexInput: HexInput }): string {
+    const hex = Hex.fromHexInput(args);
+    return hex.toString();
+  }
+
+  static hexInputToStringWithoutPrefix(args: { hexInput: HexInput }): string {
```
xbtmatt, line 74: is there a namespace conflict even if it's static? 

Would be nice if it's consistent, bc ultimately you'll call it as `Hex.toStringWithoutPrefix(...)`

ecosystem/typescript/sdk_v2/tests/unit/hex.test.ts
```diff
@@ -0,0 +1,71 @@
...
+  expect(hex).toEqual(mockHex.withPrefix);
+});
+
+test("converts hex bytes to string without 0x prefix", () => {
+  const hex = Hex.hexInputToStringWithoutPrefix({ hexInput: mockHex.withPrefix });
+  expect(hex).toEqual(mockHex.withoutPrefix);
+});
+
+test("throws when parsing invalid hex char", () => {
+  expect(() => Hex.fromString({ str: "0xxyz" })).toThrow();
```
0xmaayan, line 61: why `0xxyz` is invalid? the function removes the `0x` part and left with `xyz` - that should be valid according  to the code?

xbtmatt: I think the throw would be from `noble` bc xyz aren't valid hex chars

ecosystem/typescript/sdk_v2/src/types/hex.ts
xbtmatt: Overall this looks nice to work with, except for the single arg object 😅 For stuff like this, like simple conversion functions, it definitely seems annoying and unintuitive because you need to name two args even though it's a single input

ecosystem/typescript/sdk_v2/tests/unit/hex.test.ts
```diff
@@ -0,0 +1,71 @@
...
+});
+
+test("throws when parsing hex of length zero", () => {
+  expect(() => Hex.fromString({ str: "0x" })).toThrow();
```
0xmaayan, line 65: can add the message it throws. i.e here it throws `Hex string is too short, must be at least 1 char long, excluding the optional leading 0x.`

ecosystem/typescript/sdk_v2/src/types/hex.ts
```diff
@@ -0,0 +1,91 @@
...
+    return hex.toStringWithoutPrefix();
+  }
+
+  static isValid(args: { str: string }): { valid: boolean; invalidReason: string | null } {
+    try {
+      Hex.fromString(args);
+      return { valid: true, invalidReason: null };
+    } catch (e) {
+      const error = e as Error;
+      return {
```
0xmaayan, line 85: is there a reason we hold `invalidReasonCode` in AptosAccount but not here? I am not saying we shouldn't have it but wonder why there and not here.

Also, these Objects seem pretty similar here and in AptosAccount, should we create a type of it and define the return type?
```
type Error = {
valid: boolean, invalidReason?: string, invalidReasonCode?: string }
}

static isValid(args: { str: string }): Error
```

ecosystem/typescript/sdk_v2/src/types/hex.ts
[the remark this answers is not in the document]
banool: Spoke offline, we'll do `core`.

ecosystem/typescript/sdk_v2/src/types/hex.ts
```diff
@@ -0,0 +1,150 @@
...
+  constructor(data: Uint8Array) {
+    this.data = data;
+  }
+
+  /**
+   * Static method to convert a hex string to Hex
+   * @param str A hex string, with or without the 0x prefix
+   * @returns Hex
+   */
+  static fromString(str: string): Hex {
```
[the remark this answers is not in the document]
banool: I figure this isn't a helper method, it's a parsing method. This way works better in my brain.

ecosystem/typescript/sdk_v2/src/types/hex.ts
banool: Yeah... But we agreed we'll take this for a spin, so let's see how it goes!

ecosystem/typescript/sdk_v2/src/types/hex.ts
banool: Well there are only two places where parsing can fail, the check for length and the call to the noble function. So it seemed overkill to have a code. I agree it's more uniform though, let me add a code.

ecosystem/typescript/sdk_v2/src/core/hex.ts
```diff
@@ -0,0 +1,120 @@
...
+      );
+    }
+  }
+
+  static fromHexInput(args: { hexInput: HexInput }): Hex {
+    if (args.hexInput instanceof Uint8Array) return new Hex({ data: args.hexInput });
+    return Hex.fromString({ str: args.hexInput });
+  }
+
+  toUint8Array(): Uint8Array {
```
0xmaayan, line 80: toUint8Array feels like we convert it to Uint8Array, if we just return the class property can use a getter
```
public get hex(): Uint8Array {
 return this.data
}
```

ecosystem/typescript/sdk_v2/src/core/hex.ts
```diff
@@ -0,0 +1,120 @@
...
+ * getTransactionByHash(txnHash: HexInput): Promise<Transaction> {
+ *   const txnHashString = Hex.hexInputToString({ hexInput: txnHash });
+ *   return await getTransactionByHashInner(txnHashString);
+ * }
+ * ```
+ *
+ * This call to `Hex.hexInputToString` converts the HexInput to a hex string with a
+ * leading 0x prefix, regardless of what the input format was.
+ */
+export class Hex {
```
0xmaayan, line 39: have we lost all method comments?

ecosystem/typescript/sdk_v2/src/types/hex.ts
banool: I figure prefixing with `hexInput` makes it a bit clearer that it's taking a hex input and that it's a static function.

ecosystem/typescript/sdk_v2/src/core/hex.ts
```diff
@@ -0,0 +1,120 @@
...
+  private data: Uint8Array;
+
+  constructor(args: { data: Uint8Array }) {
+    this.data = args.data;
+  }
+
+  static fromString(args: { str: string }): Hex {
```
0xmaayan, line 46: nit: can reorganize this file where all non-static methods at the top and follow by static methods

ecosystem/typescript/sdk_v2/src/core/hex.ts
banool: I feel like this makes it inconsistent with the other functions though, it makes it weird and unnatural imo when some are getters and some are normal functions. At first I had `asUint8Array`, perhaps we could do that?

ecosystem/typescript/sdk_v2/src/core/hex.ts
banool: Oh yeah wth, idk where they went. I'll add them back.

ecosystem/typescript/sdk_v2/src/core/hex.ts
0xmaayan: sounds good!

0xmaayan (approved):

ecosystem/typescript/sdk_v2/src/core/hex.ts
```diff
@@ -0,0 +1,192 @@
...
+    if (input.length === 0) {
+      throw new ParsingError(
+        "Hex string is too short, must be at least 1 char long, excluding the optional leading 0x.",
+        HexInvalidReason.TOO_SHORT,
+      );
+    }
+
+    if (input.length % 2 !== 0) {
+      throw new ParsingError("Hex string must be an even number of hex characters.", HexInvalidReason.INVALID_LENGTH);
+    }
```
gregnazario, line 110: We... may want to be careful here cause 0x1, it won't be accepted right?  May need to just make it an even number of chars.  Oh nvm, that's what HexInput is for

(always the problem with removing the leading 0s)

ecosystem/typescript/sdk_v2/src/core/hex.ts
```diff
@@ -0,0 +1,192 @@
...
+  /**
+   * Static method to convert HexInput to a string without the 0x prefix.
+   *
+   * @returns hex as string without 0x prefix
+   */
+  static hexInputToStringWithoutPrefix(args: { hexInput: HexInput }): string {
+    const hex = Hex.fromHexInput(args);
+    return hex.toStringWithoutPrefix();
+  }
+
```
gregnazario, line 169: Wondering, why have these as static, if they can just be exactly the same as creating, then output?

ecosystem/typescript/sdk_v2/src/core/hex.ts
```diff
@@ -0,0 +1,192 @@
...
+/**
+ * NOTE: Do not use this class when working with account addresses, use AccountAddress.
+ *
+ * NOTE: When accepting hex data as input to a function, prefer to accept HexInput and
+ * then use the static helper methods of this class to convert it into the desired
+ * format. This enables the greatest flexibility for the developer.
+ *
+ * Hex is a helper class for working with hex data. Hex data, when represented as a
+ * string, generally looks like this, for example: 0xaabbcc, 45cd32, etc.
+ *
```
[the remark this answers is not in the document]
banool: I think people should see the important callouts first.

ecosystem/typescript/sdk_v2/src/core/common.ts
```diff
@@ -0,0 +1,32 @@
...
+ */
+export type ParsingResult<T> = {
+  // True if valid, false otherwise.
+  valid: boolean;
+
+  // If valid is false, this will be a code explaining why parsing failed.
+  invalidReason: T | null;
+
+  // If valid is false, this will be a string explaining why parsing failed.
+  invalidReasonMessage: string | null;
```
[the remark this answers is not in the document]
banool: Great question, I legit have no idea which is better lol. I can do some googling.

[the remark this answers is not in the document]
0xmaayan: yes, lets do `invalidReasonMesssage?:string` . that would be converted into `invalidReasonMesssage:string | undefined`

ecosystem/typescript/sdk_v2/src/core/hex.ts
banool: Yeah so 0x1 is not valid, Hex enforces that the string is a valid hex string as-is, without any padding.

ecosystem/typescript/sdk_v2/src/core/hex.ts
banool: Discussing offline.

gregnazario (approved):

xbtmatt (approved):

PR #9600, opened 2023-08-10
Update release.yaml

thepomeranian:
added steps


aptos-move/aptos-release-builder/data/release.yaml
```diff
@@ -1,6 +1,6 @@
 ---
 remote_endpoint: ~
-name: "v1.6"
+name: "step_1_v1.6"
```
movekevin, line 3: Why not name this step_1_upgrade_framework to be consistent with previous versions?

runtian-zhou (approved):

movekevin (approved):

PR #9603, opened 2023-08-10
[gas] convert output to InternalGas cost

wlawt:
### Description

Currently the output isn't that helpful, as the numbers represent the gas usage per instruction. This PR converts these numbers to the InternalGas cost, which allows for future roadmap items like updating the gas schedule much easier. 

Before:
```
where the gas parameter values are (microsecond per instruction):

0.03320208133434005 ADD
0.02167781151796508 AND
0.03401204199678623 BIT_AND
```

After:
```
internal gas cost (66666 InternalGas per 1µ):

ADD = 2213.4499542351136
AND = 1445.17298265666
BIT_AND = 2267.446791757751
```

### DIFF
--- a/aptos-move/aptos-gas-calibration/src/solve.rs
+++ b/aptos-move/aptos-gas-calibration/src/solve.rs
@@ -9,8 +9,11 @@ use crate::{
     },
     math_interface::generic_map,
 };
+use aptos_gas_schedule::{InitialGasSchedule, TransactionGasParameters};
 use nalgebra::DMatrix;
 use std::collections::BTreeMap;
+
+const MILLISECONDS_TO_MICROSECONDS: u64 = 1000;
 
 /// wrapper function to build a coefficient matrix
 ///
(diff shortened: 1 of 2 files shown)


aptos-move/aptos-gas-calibration/src/solve.rs
```diff
@@ -168,3 +166,33 @@ fn report_undetermined_gas_params(
...
+    x_hat: &mut DMatrix<f64>,
+    max_execution_time: u64,
+    gas_params: Vec<String>,
+) {
+    let one_microsec_per_internal_gas =
+        (MAX_EXECUTION_GAS / max_execution_time) / MILLISECONDS_TO_MICROSECONDS;
+
+    println!(
+        "\ninternal gas cost (1µ = {} InternalGas):\n",
+        one_microsec_per_internal_gas
```
vgao1996, line 187: It should be called the opposite -- internal gas per microsecond

aptos-move/aptos-gas-calibration/src/solve.rs
```diff
@@ -12,6 +12,9 @@ use crate::{
 use nalgebra::DMatrix;
 use std::collections::BTreeMap;
 
+const MAX_EXECUTION_GAS: u64 = 20000000000;
```
vgao1996, line 15: Consider reading the value from `TransactionGasParameters::initial()` instead of hardcoding it?

vgao1996 (commented):
Can you also attach a sample output in the summary? This will make it easier for reviewers to see what you are trying to achieve with changes like this.

vgao1996 (approved):

runtian-zhou (approved):

PR #9605, opened 2023-08-10
[deps] Update clap dependency to be consistent across crates

gregnazario:
### Description
This mainly just cleans up a little compiler v2 interfacing so all of the codebase is on the same version of clap.  It should just make every input a 0 or more argument input, and if there are any that you want fixed, feel free to call them out.

### Test Plan
build seems to work, let's see how CI handles it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -237,7 +237,7 @@ dependencies = [
  "base64 0.13.0",
  "bcs 0.1.4",
  "chrono",
- "clap 4.3.5",
+ "clap 4.3.21",
  "clap_complete",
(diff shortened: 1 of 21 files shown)


wrwg (approved):
Really should use `{ workspace = true }` or such to get rid of this problem, but lets do it next time.

gregnazario:
> Really should use `{ workspace = true }` or such to get rid of this problem, but lets do it next time.

Yeah, we have to probably set that up for the `third-party` crates, unsure what that involves.

vgao1996 (approved):

PR #9608, opened 2023-08-10
Crate disentangle

brianolson:
### Description

`network/` crate becomes `network/framework/` so that no frameworks are nested within it, other crates under `network/...` are unaffected.

### Test Plan

Unit tests and lint should catch all the compile dependencies. No code changes.


JoshLind:
I have mixed feelings about this. I agree it's interesting that `network/` is a crate, but I'm not sure this PR is the right approach. Generally, things under `crates` should be self-contained and reusable. But, moving the networking code into it, feels odd.

Regarding the peer monitoring service, I'd rather we bring this up into its own directory (i.e., at the same level as state sync, mempool and consensus). Note sure where else to put the networking code though (e.g., `network-discovery`, `network-builder`, etc.)

brianolson:
The net change is now `network/` -> `network/framework/`, and `network/{crate}` remains for everything else. All crate names in the root Cargo.toml remain the same.

davidiw:
nice, nit though, framework -> core, thoughts?

brianolson:
sorry, "netcore" is taken.
"framework" actually feels right to me. Kinda similar to some HTTP frameworks I've worked with they parse the protocol, do event handling, and pass structs for Headers and Body data, and handle reassembly of the outbound message.

JoshLind (approved):
LGTM. Unblocking 😄

bchocho (approved):
LGTM

PR #9612, opened 2023-08-10
[txn emitter] allow reuse of accounts using seed, also add a mode to exclusively create accounts

bchocho:
### Description

Account creation is a pretty expensive part of the emitter, so allow reuse. Note, the seed allows re-creation of all private keys of the test accounts, so must be kept safe (e.g., in gcloud secrets).

Also create an exclusive `create-accounts` mode for separating out account creation and thus making it easier to exactly specify the gas to use in each step.

### Test Plan

Run `% cargo run -p aptos-transaction-emitter --release -- create-accounts -t https://fullnode.testnet.aptoslabs.com --coin-source-key ${APT_KEY} --chain-id testnet --count 1 --max-gas-per-txn 1009` and check output via explorer.

### DIFF
--- a/crates/transaction-emitter-lib/src/args.rs
+++ b/crates/transaction-emitter-lib/src/args.rs
@@ -70,10 +70,6 @@ pub struct ClusterArgs {
     #[clap(long, conflicts_with = "targets")]
     pub targets_file: Option<String>,
 
-    /// If set, try to use public peers instead of localhost.
-    #[clap(long)]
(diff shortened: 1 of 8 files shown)


bchocho:
Ping @igor-aptos

igor-aptos (approved):
Thanks, looks good! sorry for delay

sitalkedia (approved):

PR #9613, opened 2023-08-10
remove redundant test

0xmaayan:
### Description
This test is not really a SDK test but more an API test. if we really want to test the sdk, we should probably mock the response, and then check if on second request cookies are still there.
Having this test here can block CI when devnet is down


banool (approved):
Could you open an issue to test this on the API side instead?

0xmaayan:
> Could you open an issue to test this on the API side instead?

wasn't sure where to put it b/c it is not a bug, but we have it here https://github.com/orgs/aptos-labs/projects/15/views/1?filterQuery=maayan&pane=issue&itemId=35625080

JoshLind (approved):

gregnazario (approved):

PR #9616, opened 2023-08-11
[Execution Benchmark] Reduce cost

danielxiangzl:
### Description
Remove parallel execution test for every PR and reduce the number of runs for sequential execution test.

### DIFF
--- a/.github/workflows/workflow-run-execution-performance.yaml
+++ b/.github/workflows/workflow-run-execution-performance.yaml
@@ -64,30 +64,6 @@ jobs:
       - run: echo "Skipping sequential execution performance! Unrelated changes detected."
         if: needs.file_change_determinator.outputs.only_docs_changed == 'true'
 
-  # Run parallel execution performance tests
-  parallel-execution-performance:
-    needs: file_change_determinator
-    timeout-minutes: 60
-    runs-on: ${{ inputs.RUNNER_NAME }}
-    steps:
-      - uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8 # pin@v3
-        with:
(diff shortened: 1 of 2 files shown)


sitalkedia:
Why remove the parallel execution and keep the sequential one and not the other way round? I would assume running the parallel execution actually gives us more signal as it covers any block-stm regression as well as slowness in other part of the system covered by the sequential execution?

danielxiangzl:
> Why remove the parallel execution and keep the sequential one and not the other way round? I would assume running the parallel execution actually gives us more signal as it covers any block-stm regression as well as slowness in other part of the system covered by the sequential execution?

Two reasons: (1) sequential is cheaper than parallel execution which requires testing various block sizes and conflict rates to get enough signals, (2) we mostly want to detect VM regression in this test, as we trust folks making changes to BlockSTM should be serious enough about the performance and should run the benchmark locally for verification (which is what we do before having this benchmark).

sitalkedia:
> we mostly want to detect VM regression in this test,

We already have `single node performance` test for this.. I am not sure having a sequential in-memory benchmark is going to help it any further. If we are confident on not introducing any regression to block-stm, I would say this test is not adding much value. May be we move both the sequential execution and parallel execution to be running periodically on main instead of running on every PR?

danielxiangzl:
> > we mostly want to detect VM regression in this test,
> 
> We already have `single node performance` test for this.. I am not sure having a sequential in-memory benchmark is going to help it any further. If we are confident on not introducing any regression to block-stm, I would say this test is not adding much value. May be we move both the sequential execution and parallel execution to be running periodically on main instead of running on every PR?

Sounds good to me!

PR #9620, opened 2023-08-11
update serde and fix `generate-format` tests

ibalajiarun:
### Description

`cargo test -p generate-format` in #9463 fail because it updates serde due to a dependency requirement and it seems breaks the generate-format tests. This PR fixes those test while updating serde.

### Test Plan

Existing tests


third_party/move/move-core/types/src/language_storage.rs
```diff
@@ -108,7 +108,7 @@ pub struct StructTag {
     pub module: Identifier,
     pub name: Identifier,
     // alias for compatibility with old json serialized data.
-    #[serde(rename = "type_args", alias = "type_params")]
+    #[serde(rename = "type_args")]
```
gregnazario, line 111: It's interesting how this causes issues for this test, though I'm not sure what the implications are

wrwg: As I said on slack, its unclear to me why this is needed and what are the implications. Why doesn't this work any longer? Bug in serde?

Cargo.lock
```diff
@@ -12083,7 +12083,18 @@ version = "0.3.5"
...
  "once_cell",
- "serde 1.0.149",
+ "serde 1.0.183",
+ "thiserror",
+]
+
+[[package]]
+name = "serde-reflection"
+version = "0.3.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
```
davidiw, line 12093: we're pulling in two versions of serde-reflection now, could we fix that -- would be curious if that might be the culprit here...

davidiw (changes requested):

Cargo.lock
ibalajiarun: The crates.io version also suffers from the same problem, and the test fails.

wrwg: Do you mean you assured both versions are the same? Not sure what you mean with 'the crates.io version', all those versions are from crates.io.

PR #9623, opened 2023-08-11
[cleanup] Remove foreign contracts

georgemitenkov:
### Description

Seems like `foreign_contracts` are not used?


alinush (approved):
I think we removed mirai a while back. Confirm with Wolfgang just to be sure!

davidiw (approved):

davidiw:
There's nothing really here about mirai, but yes that was removed quite a while ago.

wrwg (approved):

PR #9624, opened 2023-08-11
typo spec-lang.md

cryptomolot:
typo

### DIFF
--- a/developer-docs-site/docs/move/prover/spec-lang.md
+++ b/developer-docs-site/docs/move/prover/spec-lang.md
@@ -237,7 +237,7 @@ function in such a situation, one can use the notation `::len(v)`.
 In MSL, expressions have partial semantics. This is in contrast to Move program expressions, which
 have total semantics, since they either deliver a value or abort.
 
-An expression `e[X]` that depends on some some variables `X` may have a known interpretation for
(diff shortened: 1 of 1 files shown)


developer-docs-site/docs/move/prover/spec-lang.md
```diff
@@ -237,7 +237,7 @@ function in such a situation, one can use the notation `::len(v)`.
 In MSL, expressions have partial semantics. This is in contrast to Move program expressions, which
 have total semantics, since they either deliver a value or abort.
 
-An expression `e[X]` that depends on some some variables `X` may have a known interpretation for
+An expression `e[X]` that depends on some variables `X` may have a known interpretation for
```
gregnazario, line 240: This I believe is generated from source code, so that will need to be updated first, then these would be generated from it.

cryptomolot: Thanks for an explanation! Anything needed from my side?

davidiw: This might be copied content, but it is in the path that is static and not generated. Good to go.

davidiw (approved):

PR #9626, opened 2023-08-11
[token_v2][framework] Introduce new index field to token.move in preparation for aggregators v2

igor-aptos:

### DIFF
--- a/aptos-move/framework/aptos-token-objects/sources/token.move
+++ b/aptos-move/framework/aptos-token-objects/sources/token.move
@@ -10,6 +10,7 @@ module aptos_token_objects::token {
     use std::string::{Self, String};
     use std::signer;
     use std::vector;
+    use aptos_framework::aggregator_v2::{Self, AggregatorSnapshot};
     use aptos_framework::event;
     use aptos_framework::object::{Self, ConstructorRef, Object};
     use aptos_token_objects::collection::{Self, Collection};
@@ -37,8 +38,11 @@ module aptos_token_objects::token {
     struct Token has key {
         /// The collection from which this token resides.
         collection: Object<Collection>,
+        /// Deprecated in favor of index inside TokenAppendix1.
+        /// Will be populated until concurrent_token_v2_enabled feature flag is enabled.
+        ///
         /// Unique identifier within the collection, optional, 0 means unassigned
-        index: u64,
+        deprecated_index: u64,
         /// A brief description of the token.
         description: String,
         /// The name of the token, which should be unique within the collection; the length of name
@@ -49,6 +53,13 @@ module aptos_token_objects::token {
         uri: String,
         /// Emitted upon any mutation of the token.
         mutation_events: event::EventHandle<MutationEvent>,
+    }
+
+    #[resource_group_member(group = aptos_framework::object::ObjectGroup)]
+    /// Represents first addition to the common fields for all tokens
+    struct TokenAppendix1 has key {
+        /// Unique identifier within the collection, optional, 0 means unassigned
+        index: AggregatorSnapshot<u64>,
     }
 
     /// This enables burning an NFT, if possible, it will also delete the object. Note, the data
@@ -91,15 +102,21 @@ module aptos_token_objects::token {
         let collection = object::address_to_object<Collection>(collection_addr);
         let id = collection::increment_supply(&collection, signer::address_of(&object_signer));
 
+        let index = option::get_with_default(&mut id, 0);
         let token = Token {
             collection,
-            index: option::get_with_default(&mut id, 0),
+            deprecated_index: index,
             description,
             name,
             uri,
             mutation_events: object::new_event_handle(&object_signer),
         };
         move_to(&object_signer, token);
+
+        let token_appendix_1 = TokenAppendix1 {
+            index: aggregator_v2::create_snapshot(index),
+        };
+        move_to(&object_signer, token_appendix_1);
 
         if (option::is_some(&royalty)) {
             royalty::init(constructor_ref, option::extract(&mut royalty))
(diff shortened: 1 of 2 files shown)


aptos-move/framework/aptos-token-objects/sources/token.move
```diff
@@ -37,8 +38,11 @@ module aptos_token_objects::token {
     struct Token has key {
         /// The collection from which this token resides.
         collection: Object<Collection>,
+        /// Deprecated in favor of index inside TokenAppendix1.
+        /// Will be populated until concurrent_token_v2_enabled feature flag is enabled.
+        ///
         /// Unique identifier within the collection, optional, 0 means unassigned
-        index: u64,
+        deprecated_index: u64,
```
movekevin, line 45: This would violate the backward compat policy for module ugprades :)

igor-aptos: haha, I wanted to ask that.

Would it, as it is just name, and BCS is ignoring it. are names themselves enforced not to change by module upgrade?

otherwise I can just add a comment to mark it that way

zekun000: does it? I thought field name is irrelevant

xbtmatt: It doesn't matter for function args, it does for struct fields! Just tested it

zekun000: if it does, it sounds like unnecessary constraint that we may be able to relax? let's discuss in #move-eng or somewhere

aptos-move/framework/aptos-token-objects/sources/token.move
```diff
@@ -254,6 +271,23 @@ module aptos_token_objects::token {
...
+            aggregator_v2::copy_snapshot(&borrow_global<TokenAppendix1>(token_address).index)
+        } else {
+            aggregator_v2::create_snapshot(borrow(token).deprecated_index)
+        }
+    }
+
+    #[view]
+    /// Avoid this method in the same transaction as the token is minted
+    /// as that would prohibit transactions to be executed in parallel.
+    public fun index_view<T: key>(token: Object<T>): u64 acquires Token, TokenAppendix1 {
```
davidiw, line 287: let's just call this `index` to go along with the other names

aptos-move/framework/aptos-token-objects/sources/token.move
```diff
@@ -254,6 +271,23 @@ module aptos_token_objects::token {
         }
     }
 
+    /// This method allows minting to happen in parallel, making it efficient.
+    public fun index_snapshot<T: key>(token: &Object<T>): AggregatorSnapshot<u64> acquires Token, TokenAppendix1 {
+        let token_address = object::object_address(token);
+        if (exists<TokenAppendix1>(token_address)) {
+            aggregator_v2::copy_snapshot(&borrow_global<TokenAppendix1>(token_address).index)
+        } else {
+            aggregator_v2::create_snapshot(borrow(token).deprecated_index)
```
davidiw, line 280: is this efficient? asking if below, we should just perform the check again.

aptos-move/framework/aptos-token-objects/sources/token.move
```diff
@@ -51,6 +55,13 @@ module aptos_token_objects::token {
         mutation_events: event::EventHandle<MutationEvent>,
     }
 
+    #[resource_group_member(group = aptos_framework::object::ObjectGroup)]
+    /// Represents first addition to the common fields for all tokens
+    struct TokenAppendix1 has key {
+        /// Unique identifier within the collection, optional, 0 means unassigned
+        index: AggregatorSnapshot<u64>,
+    }
```
davidiw, line 63: I'd prefer we just call this struct `Index` or `TokenIndex`. `TokenAppendix1` has no useful information about it and we're not going to be able to add more fields anyway.

aptos-move/framework/aptos-token-objects/sources/token.move
davidiw: While it won't break anything for BCS compatibility, it will break it for JSON: https://github.com/aptos-labs/aptos-core/blob/main/ecosystem/python/sdk/aptos_sdk/aptos_token_client.py#L118

might be too risky a move...

davidiw (changes requested):
leaving requested changes to track status... feel free to remove if all parties are good

aptos-move/framework/aptos-token-objects/sources/token.move
vusirikala: The original motivation to introduce aggregator snapshots is to replace `read(agg)` with `snapshot(agg)`, and materialize the snapshot at a later point in time. How are we achieving that with this `create_snapshot` method? Here, we are creating a snapshot from an index (u64) instead of an aggregator. How is a snapshot helping us here?

aptos-move/framework/aptos-token-objects/sources/token.move
```diff
@@ -229,7 +287,19 @@ module aptos_token_objects::token {
...
+            aggregator_v2::create_snapshot(borrow(token).name)
+        }
+    }
+
     #[view]
+    /// Avoid this method in the same transaction as the token is minted
+    /// as that would prohibit transactions to be executed in parallel.
     public fun name<T: key>(token: Object<T>): String acquires Token {
         borrow(&token).name
     }
```
davidiw, line 305: since name is a deprecated field, we should probably check to see first if the new field exists and return that and then return name otherwise.

igor-aptos: Omg :/ nice catch. 

By the way, this is the PR that we might potentially deploy as a point release.

Full PR that we want to go out with 1.8 is this one, if you have time to take a look - https://github.com/aptos-labs/aptos-core/pull/9971/files (it has the same bug, of course :/ )

igor-aptos:
not going with a partial rollout in 1.7, closing.

PR #9633, opened 2023-08-11
[gas][docs] example guides

wlawt:
### Description
Improve docs based on user feedback

### DIFF
--- a/aptos-move/aptos-gas-calibration/README.md
+++ b/aptos-move/aptos-gas-calibration/README.md
@@ -39,14 +39,14 @@ cargo run -p aptos -- move init --name MY-PROJECT
 Calibration Functions need to be marked with `entry` and have a prefix of `calibrate_`. For example, the following functions would work:
 
 ```Move
-//// acceptable formats
+//// VALID
 public entry fun calibrate() {}
 
 public entry fun calibrate_another_txn() {}
 
 public entry fun calibrate123() {}
 
-//// inacceptable formats
+//// INVALID
 public fun calibrate() {}
 
 public fun test_my_txn() {}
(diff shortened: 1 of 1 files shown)


aptos-move/aptos-gas-calibration/README.md
```diff
@@ -56,23 +56,73 @@ public entry fun calibrate_addition(_x: u64, _y: u64) {}
...
+        aptos_hash::blake2b_256(msg);
+        aptos_hash::blake2b_256(msg);
+        aptos_hash::blake2b_256(msg);
+        aptos_hash::blake2b_256(msg);
+        aptos_hash::blake2b_256(msg);
+        aptos_hash::blake2b_256(msg);
+        aptos_hash::blake2b_256(msg);        
+        i += 1;
+    }
+}
```
vgao1996, line 111: This example is confusing. The parameter `n` is supposed to control the length of the vector used but it's not.

aptos-move/aptos-gas-calibration/README.md
```diff
@@ -56,23 +56,73 @@ public entry fun calibrate_addition(_x: u64, _y: u64) {}
...
+public entry fun calibrate_my_test_should_error() {}
+
+public entry fun calibrate_should_error() {}
+
+//// INVALID
+public entry fun should_error_calibrate() {}
+
+public entry fun calibrate_test_error() {}
+
+public entry fun calibrate_test_should_error_() {}
```
vgao1996, line 70: Might be clearer if you explicitly state that these are valid calibration functions but they are not considered ones that should error?

aptos-move/aptos-gas-calibration/README.md
```diff
@@ -90,4 +140,18 @@ For every Calibration Function, the Abstract Gas Usage and running time are dete
 
 If the matrix that represents this system is not invertible, then we report the undetermined gas parameters, or the linearly dependent combinations of gas parameters. The exact math can be found under `/src/math.rs`. 
 
-Otherwise, the User can expect to see all the values, expressed as Gas Usage per Microsecond, along with the running times and any outliers. 
\ No newline at end of file
+Otherwise, the User can expect to see all the values, expressed as Gas Usage per Microsecond, along with the running times and any outliers.
```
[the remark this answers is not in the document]
wlawt: Good point, removed

aptos-move/aptos-gas-calibration/README.md
wlawt: Maybe a better name is to call `n` the `num_iterations` instead for the while-loop

vgao1996: Oh my bad I misread

vgao1996 (approved):

aptos-move/aptos-gas-calibration/README.md
```diff
@@ -90,4 +141,18 @@ For every Calibration Function, the Abstract Gas Usage and running time are dete
...
+If you happen to see something like:
+
+```
+linearly dependent variables are:
+
+- gas parameter: HASH_BLAKE2B_256_BASE
+- gas parameter: HASH_BLAKE2B_256_PER_BYTE
+```
+
+It is possible that you may not have enough Calibration Functions. In this example, there should be at least two Calibration Functions.
```
runtian-zhou, line 157: Can you list out the possibilities of why getting "linearly dependent variable"? I think sth like
"Linearly dependent variable indicates either you don't have enough samples on the specific parameter you are trying to calibrate or you have more gas parameter than you would need"? Would be better to elaborate why this could happen.

wlawt: Good point, I noted some of the possibilities. There aren't many, but I explained a couple that could happen

runtian-zhou (approved):

PR #9634, opened 2023-08-11
trivial: fix gas meter mutability

msmouse:
### Description
Doesn't need to be `mut` in a few places.

### Test Plan

existing tests

### DIFF
--- a/aptos-move/aptos-vm/src/aptos_vm.rs
+++ b/aptos-move/aptos-vm/src/aptos_vm.rs
@@ -234,7 +234,7 @@ impl AptosVM {
     pub fn failed_transaction_cleanup(
         &self,
         error_code: VMStatus,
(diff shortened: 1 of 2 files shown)


aptos-move/aptos-vm/src/aptos_vm.rs
```diff
@@ -326,7 +327,6 @@ impl AptosVM {
                 ) {
                     return discard_error_vm_status(e);
                 }
-                let fee_statement = AptosVM::fee_statement_from_gas_meter(txn_data, gas_meter);
```
msmouse, line 329: For a moment I thought this would contain different numbers than the epilogue saw since the epilogue.

danielxiangzl (approved):

bitcoinfinancier (approved):

zjma (approved):

PR #9640, opened 2023-08-13
[TS SDK v2] Reject invalid SHORT form for special addresses in fromString

banool:
### Description
In the v2 TS SDK the `fromString` function enforces strict parsing. As written though, it wasn't exactly spec compliant in that it allowed padding zeroes in addresses. This PR fixes that.

### Test Plan
```
npx jest -- tests/unit/account_address.test.ts
```

### DIFF
--- a/ecosystem/typescript/sdk_v2/src/core/account_address.ts
+++ b/ecosystem/typescript/sdk_v2/src/core/account_address.ts
@@ -208,14 +208,14 @@ export class AccountAddress {
     if (args.input.length != AccountAddress.LONG_STRING_LENGTH + 2) {
       if (!address.isSpecial()) {
         throw new ParsingError(
-          "Hex string is not a special address, it must be represented as 0x + 64 chars.",
+          "The given hex string is not a special address, it must be represented as 0x + 64 chars.",
           AddressInvalidReason.LONG_FORM_REQUIRED_UNLESS_SPECIAL,
         );
       } else {
         // 0x + one hex char is the only valid SHORT form for special addresses.
         if (args.input.length != 3) {
           throw new ParsingError(
-            "Hex string is a special address not in LONG form, it must be 0x0 to 0xf, no padding zeroes.",
+            "The given hex string is a special address not in LONG form, it must be 0x0 to 0xf with no padding zeroes.",
             AddressInvalidReason.INVALID_PADDING_ZEROES,
           );
         }
--- a/ecosystem/typescript/sdk_v2/src/core/account_address.ts
+++ b/ecosystem/typescript/sdk_v2/src/core/account_address.ts
@@ -182,7 +182,7 @@ export class AccountAddress {
    *
    * Where:
    * - LONG is defined as 0x + 64 hex characters.
-   * - SHORT for special addresses is 0x0 to 0xf inclusive, no padding zeroes.
+   * - SHORT for special addresses is 0x0 to 0xf inclusive without padding zeroes.
    *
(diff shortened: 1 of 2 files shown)


ecosystem/typescript/sdk_v2/src/core/account_address.ts
```diff
@@ -202,10 +203,29 @@ export class AccountAddress {
 
     const address = AccountAddress.fromStringRelaxed(args);
 
+    // Check if the address is in LONG form. If it is not, this is only allowed for
+    // special addresses, in which case we check it is in proper SHORT form.
+    if (args.input.length != AccountAddress.LONG_STRING_LENGTH + 2) {
+      if (!address.isSpecial()) {
+        throw new ParsingError(
+          "Hex string is not a special address, it must be represented as 0x + 64 chars.",
```
davidiw, line 211: This reads a little weird... maybe "The given hex string" ?

ecosystem/typescript/sdk_v2/src/core/account_address.ts
```diff
@@ -202,10 +203,29 @@ export class AccountAddress {
...
+          AddressInvalidReason.LONG_FORM_REQUIRED_UNLESS_SPECIAL,
+        );
+      } else {
+        // 0x + one hex char is the only valid SHORT form for special addresses.
+        if (args.input.length != 3) {
+          throw new ParsingError(
+            "Hex string is a special address not in LONG form, it must be 0x0 to 0xf, no padding zeroes.",
```
davidiw, line 218: nit no comma instead just add "with"

ecosystem/typescript/sdk_v2/src/core/account_address.ts
```diff
@@ -181,7 +182,7 @@ export class AccountAddress {
    *
    * Where:
    * - LONG is defined as 0x + 64 hex characters.
-   * - SHORT for special addresses is 0x0 to 0xf inclusive.
+   * - SHORT for special addresses is 0x0 to 0xf inclusive, no padding zeroes.
```
davidiw, line 185: nit no comma instead just add "with"

davidiw (approved):
seems fine

xbtmatt:
In general the `async`s are not necessary right? 

Also, have you considered using `fromStringStrict` with `fromString` as the default behavior instead?

I think `strict` is more intuitively clear than `relaxed`, and in general it might be nice to have the default behavior "assume" the intent, since that's generally what people expect/want (at least I would)

Also not fully aware of if these terms are defined in the AIP already, so feel to ignore if so

xbtmatt (approved):

banool:
I'm not 100% sure I follow your point about relaxed / strict, could you clarify what you're suggesting we change here? 

If you're suggesting we rename fromString to fromStringStrict, I'd say I'd rather have an obvious default and the stricter one should be that default.

xbtmatt:
> I'm not 100% sure I follow your point about relaxed / strict, could you clarify what you're suggesting we change here?
> 
> If you're suggesting we rename fromString to fromStringStrict, I'd say I'd rather have an obvious default and the stricter one should be that default.

Yeah, that's what I was suggesting! The only reason I say that is because I find that most people end up using whatever's easiest in the end, which in some cases is just the shorter function name

But now that I think about it, I think it would make more sense to default to strict, then. Anyway, was just curious your thoughts on it!

0xmaayan (approved):

PR #9642, opened 2023-08-14
 [Python SDK] Make from_str strict, add from_str_relaxed

banool:
### Description
In this PR the `from_str` function on `AccountAddress` has been updated to conform to the strict parsing described by [AIP-40](https://github.com/aptos-foundation/AIPs/blob/main/aips/aip-40.md). For the relaxed parsing behavior of this function prior to this change, folks can use `AccountAddress.from_str_relaxed`.

Whereas with Rust we will almost certainly keep `from_str` the relaxed variant for now, with Python we're still at 0.x.x, so we're at greater liberty to make changes like this. Open to contrary thoughts though.

### Test Plan
```
poetry run python -m unittest aptos_sdk/account_address.py
```

### DIFF
--- a/ecosystem/python/sdk/CHANGELOG.md
+++ b/ecosystem/python/sdk/CHANGELOG.md
@@ -3,16 +3,17 @@
 All notable changes to the Aptos Python SDK will be captured in this file. This changelog is written by hand for now.
 
 ## 0.7.0
-- Port remaining sync examples to async (hello-blockchain, multisig, your-coin)
-- Updated token client to use events to acquire minted tokens
-- Update many dependencies and set Python 3.8.1 as the minimum requirement
-- Add support for an experimental chunked uploader
-- Add experimental support for the Aptos CLI enabling local end-to-end testing, package building, and package integration tests
+- **[Breaking Change]**: The `from_str` function on `AccountAddress` has been updated to conform to the strict parsing described by [AIP-40](https://github.com/aptos-foundation/AIPs/blob/main/aips/aip-40.md). For the relaxed parsing behavior of this function prior to this change, use `AccountAddress.from_str_relaxed`.
 - **[Breaking Change]**: Rewrote the large package publisher to support large modules too
 - **[Breaking Change]**: Delete sync client
 - **[Breaking Change]**: Removed the `hex` function from `AccountAddress`. Instead of `addr.hex()` use `str(addr)`.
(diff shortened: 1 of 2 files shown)


ecosystem/python/sdk/aptos_sdk/account_address.py
```diff
@@ -72,11 +72,113 @@ def is_special(self):
...
+
+        Parameters:
+        - address (str): A hex string representing an account address.
+
+        Returns:
+        - AccountAddress: An instance of AccountAddress.
+        """
+        # Assert the string starts with 0x.
+        if not address.startswith("0x"):
+            raise RuntimeError("Hex string must start with a leading 0x.")
```
0xmaayan, line 105: do we have errors in the sdk?

0xmaayan (commented):
LGTM. defer to @davidiw to do first stamp

gregnazario (approved):

ecosystem/python/sdk/aptos_sdk/account_address.py
banool: Don't think so, but let me define one.

ecosystem/python/sdk/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 All notable changes to the Aptos Python SDK will be captured in this file. This changelog is written by hand for now.
 
+## Unreleased
```
davidiw, line 5: 0.7.0 is unreleased still :)

davidiw (approved):
Thanks for doing this.

I do have one concern though... this file used to be super tiny and now it is 600 lines long to just parse a pretty trivial string. Do we need this verbosity? Should we instead make the test cases some data corpus of comparisons?

In other words, this code isn't easy to replicate, which goes against the goal for this SDK :grimacing:

banool:
To clarify, are you suggesting we move the tests out into a separate location?

davidiw:
> To clarify, are you suggesting we move the tests out into a separate location?

Well I don't have a great suggestion but I noticed a few things:
* A lot of this is about comparing that X is roughly equal to x
* Another set is that certain things fail in a specific fashion

This feels more like a data corpus test written in code... so it might be better to actually define some form of IDL, say a CSV, that says, `left value`, `left value op`, `right value`, `right value op`, expected outcome....

That way you can easily produce the test in like 5-10 lines of code and make it less risky as we port it to new languages....

banool:
Yeah totally agree, it'd be great to find some tool that supports a language agnostic way to define inputs + expected outputs and then has libraries in major languages for hooking into those definitions. I can open an issue + add this to our roadmap.

PR #9644, opened 2023-08-14
[refactoring] Relax trait bounds, reduce dependence on `StateView`

georgemitenkov:
### Description

API only needs move resolver, and not state view. Decoupling
`MoveResolverExt` into `AptosMoveResolver + StateView` so
that any future changes were not visible on API side.

1. Moved API checks for resource groups where it should be.

2. Making trait bounds for config not require state view. Right now
state view is used all over the place although it seems this can
be avoided.

3. Reducing the usage of `MoveResolverExt`: e.g. session only
needs to talk to resolver.

4. State value metadata has its own resolver. This makes storage
accesses almost uniform (still need to fix `read_write_set`) which
should be implemented in storage adapter.

5. Some minor typo fixes and dead code removal.

@msmouse @davidiw what do you think?


.github/linters/semgrep/pull-request-target-code-checkout.yaml
```diff
@@ -47,6 +47,15 @@ rules:
             ...
             $JOBNAME:
               ...
+      - pattern-not-inside: |
+          needs: [..., permission-check, ...]
```
vusirikala, line 51: What are these changes? These changes seem to be in the linters directory.

georgemitenkov: I think it is a CI fix, will rebase later

aptos-move/aptos-vm/src/data_cache.rs
```diff
@@ -235,3 +233,30 @@ impl<S: StateView> AsMoveResolver<S> for S {
...
     }
 }
+
+impl<'a, S: StateView> StateValueMetadataResolver for StorageAdapter<'a, S> {
+    fn get_state_value_metadata(
+        &self,
+        state_key: &StateKey,
+    ) -> anyhow::Result<Option<Option<StateValueMetadata>>> {
+        let maybe_state_value = self.state_store.get_state_value(state_key)?;
+        Ok(maybe_state_value.map(|state_value| state_value.into_metadata()))
```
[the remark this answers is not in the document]
gelash: So the outer Option is because of the maybe_state_value being empty, and inside is if the into_metadata doesn't succeed? can't we map_err to storage error on the outer one here to convert it to the result? (seems to be the way the caller uses it)?

gelash (approved):

aptos-move/aptos-vm/src/move_vm_ext/resolver.rs
```diff
@@ -30,18 +49,9 @@ pub trait MoveResolverExt:
...
-                .iter()
-                .find(|attr| attr.is_resource_group())?;
-            Some(())
-        })
-        .is_some()
-    }
 }
+
+// TODO: Remove dependency on StateView.
+pub trait MoveResolverExt: AptosMoveResolver + StateView {}
```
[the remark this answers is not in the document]
georgemitenkov: We need it for `StateView` in `RespawnedSession` (so executing user transaction essentially) and `read_write_set()`.

zekun000 (approved):
the respawned session is a bit ugly but probably for another pr

aptos-move/aptos-vm/src/aptos_vm.rs
```diff
@@ -119,16 +118,21 @@ macro_rules! unwrap_or_discard {
 }
 
 impl AptosVM {
-    pub fn new(state: &impl StateView) -> Self {
-        Self(AptosVMImpl::new(state))
+    pub fn new(config_storage: &impl ConfigStorage) -> Self {
```
vusirikala, line 121: What is `ConfigStorage`? How is it different from `StateView`? Why is `ConfigStorage` enough here instead of `StateView`?

api/src/context.rs
```diff
@@ -43,9 +43,10 @@ use aptos_types::{
     },
     transaction::{SignedTransaction, TransactionWithProof, Version},
 };
+use aptos_utils::aptos_try;
 use aptos_vm::{
     data_cache::{AsMoveResolver, StorageAdapter},
-    move_vm_ext::MoveResolverExt,
+    move_vm_ext::AptosMoveResolver,
```
vusirikala, line 49: Is `AptosMoveResolver` just a renaming for `MoveResolverExt`? Or are they different in functionality?

aptos-move/aptos-vm/src/data_cache.rs
```diff
@@ -48,7 +52,7 @@ pub struct StorageAdapter<'a, S> {
         RefCell<BTreeMap<AccountAddress, BTreeMap<StructTag, BTreeMap<StructTag, Vec<u8>>>>>,
 }
 
-impl<'a, S: StateView> StorageAdapter<'a, S> {
+impl<'a, S> StorageAdapter<'a, S> {
```
vusirikala, line 55: Why is `StateView` trait bound removed here? Isn't StorageAdapter defined only for StateView?

aptos-move/aptos-vm/src/data_cache.rs
```diff
@@ -235,3 +233,30 @@ impl<S: StateView> AsMoveResolver<S> for S {
...
+//      in epilogue.
+impl<'a, S: StateView> TStateView for StorageAdapter<'a, S> {
+    type Key = StateKey;
+
+    fn get_state_value(&self, state_key: &Self::Key) -> anyhow::Result<Option<StateValue>> {
+        self.state_store.get_state_value(state_key)
+    }
+
+    fn get_usage(&self) -> anyhow::Result<StateStorageUsage> {
+        self.state_store.get_usage()
```
vusirikala, line 260: What is usage here?

aptos-move/aptos-vm/src/move_vm_ext/session.rs
```diff
@@ -428,14 +428,17 @@ impl<'r> WriteOpConverter<'r> {
...
 
-        let existing_value_opt = self.remote.get_state_value(state_key).map_err(|_| {
-            VMStatus::error(
-                StatusCode::STORAGE_ERROR,
-                err_msg("Storage read failed when converting change set."),
-            )
-        })?;
-
-        let write_op = match (existing_value_opt, move_storage_op) {
+        let maybe_existing_metadata =
```
vusirikala, line 431: Instead of using existing value, we are now using the metadata of existing value. What's the motivation behind this change?

aptos-move/aptos-vm/src/aptos_vm.rs
georgemitenkov: `ConfigStorage` is an abstraction which allows you to get resources which are configs, e.g. gas schedule, etc. In case of VM, we don't care about state view when we initialise, all we care is the configs.

aptos-move/aptos-vm/src/data_cache.rs
georgemitenkov: There are 2 impls: all functionality relies on state view, but one should be able to create an adapter on top of some other type and implement move resolver / state view on top.

aptos-move/aptos-vm/src/data_cache.rs
georgemitenkov: I think this is needed for Move functions to query how many bytes are used in storage, etc. for pricing.

aptos-move/aptos-vm/src/move_vm_ext/session.rs
georgemitenkov: We never needed the value there - only metadata to calculate storage refunds. Motivation behind having a resolver is to avoid using a fixed `get_state_value` method and use resolver abstraction so we can use any storage backend - for indexer, execution, etc.

Maybe we should expose this in resource/move resolvers, but this suffices for now I think

aptos-move/aptos-vm/src/data_cache.rs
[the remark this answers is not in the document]
georgemitenkov: Not really: the outer thing can be None if the state value does not exist: in this case for storage refunds this is acceptable. If inner does not exist, then there is no metadata associated, so we kind of need both...

I guess using named enums instead of option would be nicer in general😄

aptos-move/aptos-vm/src/block_executor/vm_wrapper.rs
```diff
@@ -30,7 +31,10 @@ impl<'a, S: 'a + StateView + Sync> ExecutorTask for AptosExecutorTask<'a, S> {
     type Txn = PreprocessedTransaction;
 
     fn init(argument: &'a S) -> Self {
-        let vm = AptosVM::new(argument);
+        // AptosVM has to be initialized using configs from storage.
+        // Using adapter allows us to fetch those.
+        let config_storage = StorageAdapter::new(argument);
```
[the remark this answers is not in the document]
georgemitenkov: Oh, so for this change doing this allows us to not have a state view bound here actually, so it's easier to implement things in the future (I started doing that and ended up refactoring). Maybe using `from_state_view` which hides it is better?

aptos-move/aptos-vm/src/aptos_vm.rs
```diff
@@ -1231,7 +1235,7 @@ impl AptosVM {
         }
     }
 
-    fn read_writeset<'a>(
+    fn read_write_set<'a>(
```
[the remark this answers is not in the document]
georgemitenkov: renamed to `read_change_set` instead, should be cleaner (and this takes change set as argument now)

aptos-move/aptos-vm/src/block_executor/vm_wrapper.rs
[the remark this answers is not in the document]
georgemitenkov: Added a comment about that

aptos-move/aptos-vm/src/data_cache.rs
```diff
@@ -235,3 +233,30 @@ impl<S: StateView> AsMoveResolver<S> for S {
...
+        Ok(maybe_state_value.map(|state_value| state_value.into_metadata()))
+    }
+}
+
+// We need to implement StateView for adapter because:
+//   1. When processing write set payload, storage is accessed
+//      directly.
+//   3. When stacking Storage adapters on top of each other, e.g.
+//      in epilogue.
+impl<'a, S: StateView> TStateView for StorageAdapter<'a, S> {
```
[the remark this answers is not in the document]
georgemitenkov: Nice! But this crate does not allow trait implementations, only function call resolution? And here we need StorageAdapter to satisfy trait bounds. Not sure how it helps?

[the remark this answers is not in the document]
gelash: lol, for that there also seems to be a crate: https://docs.rs/ambassador/latest/ambassador/
but I'm not sure using two crates to save 5 lines is worth it )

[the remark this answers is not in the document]
georgemitenkov: Yep, I think with all the impls delegate can be pretty useful actually :) not sure about here, can revisit this, and anyway with new traits we will not need StateView here

PR #9647, opened 2023-08-14
[Executor-benchmark] Make connected grps of txns fully conflicting grps

manudhundi:
Fully conflicting grps are more meaningful pattern for testing out the sharded
executor because they cannot be executed parallely (that is what we intend to benchmark).

Cmd:
cargo run -p aptos-executor-benchmark -- --block-size 100 --connected-tx-grps 5 --shuffle-connected-txns run-executor --main-signer-accounts 1000 --data-dir /tmp/some-db --checkpoint-dir /tmp/some-checkpoint --blocks 2

For a good random workload we would want 'num_accounts_per_grp' to be much greater than 'num_txns_per_grp'.
The command enforces 'num_signer_accounts' > '2 * num_txns_per_grp'.


manudhundi:
#9164

manudhundi:
#9371

execution/executor-benchmark/src/transaction_generator.rs
```diff
@@ -451,91 +453,90 @@ impl TransactionGenerator {
...
+                            (index2, index1)
+                        }
+                    })
+                    .collect();
+                conflicting_indices
             })
             .collect();
-        transfer_indices.shuffle(rng);
+        if shuffle_indices {
+            println!("Shuffling conflicting txns*******************************");
```
sitalkedia, line 514: Remove?

sitalkedia (approved):
LGTM

execution/executor-benchmark/src/transaction_generator.rs
manudhundi: Yes, will do

zjma (approved):

execution/executor-benchmark/src/main.rs
```diff
@@ -128,6 +128,9 @@ struct Opt {
     #[clap(long, default_value_t = 0)]
     connected_tx_grps: usize,
 
+    #[clap(long)]
+    shuffle_connected_txns: bool,
```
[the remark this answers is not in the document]
manudhundi: Not shuffling is an interesting case where sharding does really well. Ideally, we want the sharding to do equally good in the case when it is shuffled. So, 'not shuffling' is a good benchmark to have.

Default value is false. There is some rust annotation issue if we specify default as positive. Can get around it by having a negated flag, but thought it is fine this way.

PR #9649, opened 2023-08-14
[NFT Metadata Crawler] Stop parsing completely if token_uri has already been parsed

just-in-chang:
### Description
Title, add more context for logs, add more panics (failed to write to gcs, failed to commit postgres tx)

### Test Plan
Works locally


bowenyang007:
Wouldn't all these panics break the entire server? Why is that desirable?

bowenyang007 (approved):

rtso (approved):

PR #9650, opened 2023-08-14
Update workflows

gedigi:
### Description
Update GitHub workflows:
- adds permission-check
- adds cron for semgrep
- updates semgrep rule to skip false positive

### Test Plan
- main: https://github.com/aptos-labs/aptos-core/actions/runs/5859783927
- this branch: https://github.com/aptos-labs/aptos-core/actions/runs/5859971040/job/15887078986?pr=9650


.github/workflows/semgrep.yaml
```diff
@@ -4,6 +4,8 @@ on:
   workflow_dispatch:
   pull_request:
     types: [labeled, opened, synchronize, reopened, auto_merge_enabled]
+  schedule:
+    - cron: '37 13 * * *'
```
ibalajiarun, line 8:
```suggestion
    - cron: '0 * * * *'
```

ibalajiarun: Probably run this every hour.

ibalajiarun (approved):

PR #9651, opened 2023-08-14
add the header to data service.

larry-aptos:
### Description

* Accepts the headers to data service.
* To avoid cardinality issue, use classification + name as the fields. 
### Test Plan

image


bowenyang007 (approved):
looks fine but hard to tell without e2e lol

PR #9654, opened 2023-08-14
add account and hex data modules

0xmaayan:

### DIFF
--- a/ecosystem/typescript/sdk_v2/src/core/account.ts
+++ b/ecosystem/typescript/sdk_v2/src/core/account.ts
@@ -1,0 +1,106 @@
+import nacl from "tweetnacl";
+import * as bip39 from "@scure/bip39";
+import { bytesToHex } from "@noble/hashes/utils";
+import { derivePath, Memoize } from "../utils";
+
+export class Account {
+  /**
+   * A private key and public key, associated with the given account
+   */
+
+  /**
+   * Address associated with the given account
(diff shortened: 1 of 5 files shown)


ecosystem/typescript/sdk_v2/src/core/account.ts
```diff
@@ -0,0 +1,106 @@
...
+  @Memoize()
+  static authKey(publicKey: Uint8Array): Hex {
+    const pubKey = new Ed25519PublicKey(publicKey);
+    const authKey = AuthenticationKey.fromEd25519PublicKey(pubKey);
+    return authKey.derivedAddress();
+  }
+
+  /**
+   * Creates new account with bip44 path and mnemonics,
+   * @param path. (e.g. m/44'/637'/0'/0'/0')
```
xbtmatt, line 64: This is the default that the wallets use right? Or does each wallet use a different derivation path?

I remember when I used to import my accounts by seed phrase I wouldn't understand why the accounts created were different and it was cause of the derivation path, would be nice if we could make it clear

0xjinn: Can do!

Btw this is a draft, I am gonna re-do the whole thing 😄

xbtmatt: Gotcha :D

PR #9664, opened 2023-08-15
config sanitize in test mode

brianolson:
### Description

Fix config sanitization in `aptos-node --test` mode (it wasn't happening at all before)

### Test Plan

Works locally when running `aptos-node --test ...`


config/src/config/node_config_loader.rs
```diff
@@ -126,6 +126,18 @@ fn optimize_and_sanitize_node_config(
...
+pub fn sanitize_node_config(node_config: &mut NodeConfig) -> Result<(), Error> {
+    let node_type = NodeType::extract_from_config(node_config);
+    let chain_id = match get_chain_id(node_config) {
+        Ok(chain_id) => chain_id,
+        Err(error) => {
+            println!("Failed to get the chain ID from the genesis blob! Skipping config sanitization. Error: {:?}", error);
+            return Ok(());
+        },
+    };
+    NodeConfig::sanitize(node_config, node_type, chain_id)
```
gregnazario, line 138: This sanitizes the testing configs?

brianolson: I think I found out that the sanitizer wasn't running because there was something in my overlay that it would have caught but instead I got a crash later (probably `None.unwrap()` for some `Option` in the config, but the sanitizer would have replaced None with a default value)

JoshLind: Yeah, I originally never supported this path because I wasn't sure how useful it was. But, if you've found it useful, then it might be worth doing.

@bchocho: @brianolson is only calling `sanitize()` and not `optimize()` here, so it won't change any of the configs. Sanitize was meant to leave it unchanged 🤞

config/src/config/node_config_loader.rs
```diff
@@ -126,6 +126,18 @@ fn optimize_and_sanitize_node_config(
     NodeConfig::sanitize(node_config, node_type, chain_id)
 }
 
+pub fn sanitize_node_config(node_config: &mut NodeConfig) -> Result<(), Error> {
+    let node_type = NodeType::extract_from_config(node_config);
```
JoshLind, line 130: nit: it might be worth adding a helper function to do the node type and chain ID extraction so that we don't have the duplication with the function above.

JoshLind (approved):
LGTM. If test mode still works, I'm happy 😄

aptos-node/src/lib.rs
```diff
@@ -288,6 +288,8 @@ where
             genesis_waypoint.to_string().as_bytes(),
         )?;
 
+        aptos_config::config::sanitize_node_config(&mut validators[0].config)?;
```
JoshLind, line 291: Any reason we don't sanitize all node configs?

brianolson: `--test` mode is only one validator. If that changes this wouldn't work right, but we have `forge` local swarms for things beyond this test mode.

config/src/config/node_config_loader.rs
brianolson: The sanitizer is effectively where some defaults live, not in the value loaded by serde Default trait implementation just at/after the constructor, but later if something hasn't been set then fix it up at sanitizer time.

bchocho (approved):

PR #9666, opened 2023-08-15
balanced replay-verify

areshand:
### Description
Parallel partitions of each runner decreased to 16 from 32 so that we can scale on cheaper machines with 200G mem
Runner number increase to 16 from 8
Load is balanced across runners. 

### Test Plan
Aws shows the finishing time is closed to 2 hrs. the outlier is manually adjusted
https://github.com/aptos-labs/aptos-core-private/actions/runs/5869091896

test replay https://github.com/aptos-labs/aptos-core/actions/runs/5872812275

### DIFF
--- a/.github/workflows/workflow-run-replay-verify.yaml
+++ b/.github/workflows/workflow-run-replay-verify.yaml
@@ -82,7 +82,7 @@ jobs:
     runs-on: ${{ inputs.RUNS_ON }}
     strategy:
         matrix:
-          number: [0, 1, 2, 3, 4, 5, 6, 7] # runner number
+          number: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15] # runner number
     steps:
       - name: Echo Runner Number
         run: echo "Runner is ${{ matrix.number }}"
@@ -106,7 +106,7 @@ jobs:
 
       - name: Run replay-verify in parallel
         shell: bash
-        run: testsuite/replay_verify.py ${{ matrix.number }} 8 # first argument is the runner number, second argument is the total number of runners
+        run: testsuite/replay_verify.py ${{ matrix.number }} 16 # first argument is the runner number, second argument is the total number of runners
         env:
           BUCKET: ${{ inputs.BUCKET }}
           SUB_DIR: ${{ inputs.SUB_DIR }}
--- a/testsuite/replay_verify.py
+++ b/testsuite/replay_verify.py
@@ -16,52 +16,66 @@ from verify_core.common import clear_artifacts, query_backup_latest_version
 # This script runs the replay-verify from the root of aptos-core
 # It assumes the aptos-db-tool binary is already built with the release profile
 
-testnet_runner_mapping = {
-    0:[250000000, 255584106],
-    1:[255584107, 271874718],
-    2:[271874719, 305009463],
-    3:[305009464, 324904819],
-    4:[324904820, 347234877],
-    5:[347234878, 366973577],
-    6:[366973578, 399489396],
-    7:[399489397, 430909965],
-    8:[430909966, 449999999],
-    9:[450000000, 462114510],
-    10:[462114511, 477825432],
-    11:[477825433, 485000000],
-    12:[485000001, 516281795],
-    13:[516281796, 551052675],
-    14:[551052676, 582481398],
-    15:[582481399, sys.maxsize]
-}
 
-mainnet_runner_mapping = {
-    0:[0, 14949498],
-    1:[14949499, 30518131],
-    2:[30518132, 49314011],
-    3:[49314012, 69611025],
-    4:[69611026, 90057535],
-    5:[90057536, 109821002],
-    6:[109821003, 125881567],
-    7:[125881568, 134463753],
-    8:[134463754, 153497556],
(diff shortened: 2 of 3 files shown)


grao1991 (approved):

testsuite/replay_verify.py
```diff
@@ -16,6 +16,43 @@
 # This script runs the replay-verify from the root of aptos-core
 # It assumes the aptos-db-tool binary is already built with the release profile
 
+testnet_runner_mapping = {
```
sherry-x, line 19: can we add a comment on what is this mapping?

sherry-x: also why their number is so random? Is this depends on the transaction we need to skip?

areshand: updated. The number is generated using a script initially linked in the comment as well

msmouse: Should add steps to 1. how we "seal" the latest range when it reaches a similar run time compared with previous ranges (and that the oncall will receive a low pri alert) 2. how we retire the oldest range when according to our policy the backward compatibility is no longer required.

testsuite/replay_verify.py
```diff
@@ -16,15 +16,58 @@
...
-    latest_version: int,
-    txns_to_skip: Tuple[int],
-    backup_config_template_path: str,
+        n: int,
+        N: int,
+        history_start: int,
+        per_partition: int,
+        latest_version: int,
+        txns_to_skip: Tuple[int],
+        backup_config_template_path: str,
```
msmouse, line 70: accidental?

testsuite/replay_verify.py
```diff
@@ -106,10 +149,9 @@ def main():
         runner_cnt = 1
 
     assert (
-        runner_no >= 0 and runner_no < runner_cnt
+            runner_no >= 0 and runner_no < runner_cnt
```
msmouse, line 152: accidental?

testsuite/replay_verify.py
```diff
@@ -129,12 +171,17 @@ def main():
...
+        else mainnet_runner_mapping
+    )
+    runner_start = runner_mapping[runner_no][0]
+    runner_end = runner_mapping[runner_no][1]
+    if runner_no == runner_cnt - 1:
+        runner_end = LATEST_VERSION
     print("runner start %d end %d" % (runner_start, runner_end))
     # run replay-verify in parallel
-    N = 32
+    N = 16
```
msmouse, line 184: should this be derived from the strategy?

testsuite/replay_verify.py
msmouse: (the steps can be easier if we make this an vector instead of a map, and automatically give runners names? (removes the  `strategy.matrix.number` array as well?) -- Looks like in the current way we are coding it, adding a range will require changes in three places.

msmouse (commented):
Beautiful

msmouse (commented):
Beautiful. We are below 2 hours with this. ❤️

testsuite/replay_verify.py
areshand: how do we make sure different runners get different ranges without passing in a seq#? Each runner doesn't know about other runner and cannot communicate.

msmouse: The vector items is keyed by the index anyway? I'm a jerk and I didn't read, but in my illusion when you spawn the runners just read the length of the vector, and assign 0..len to the runners

areshand: I see. Github spawns the runners.  But I guess you mean I can put the ranges in Yaml to coordinator the runners.

testsuite/replay_verify.py
```diff
@@ -16,52 +16,66 @@
...
-    7:[399489397, 430909965],
-    8:[430909966, 449999999],
-    9:[450000000, 462114510],
-    10:[462114511, 477825432],
-    11:[477825433, 485000000],
-    12:[485000001, 516281795],
-    13:[516281796, 551052675],
-    14:[551052676, 582481398],
-    15:[582481399, sys.maxsize]
+    0: [250000000, 255584106],
```
msmouse, line 33: make it
```
testnet_runner_mapping = [
    [255584107, 271874718],
    [271874719, 305009463],
    ...
```
?

areshand: Ok, updated, I thought you wanted to remove strategy.matrix.number array

msmouse: both, I want ALL. 🚀

.github/workflows/workflow-run-replay-verify.yaml
```diff
@@ -82,7 +82,7 @@ jobs:
     runs-on: ${{ inputs.RUNS_ON }}
     strategy:
         matrix:
-          number: [0, 1, 2, 3, 4, 5, 6, 7] # runner number
+          number: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15] # runner number
```
msmouse, line 85: so this can't be removed / derived?

areshand: Cannot, it is used by GitHub to spawn runners

msmouse (approved):

PR #9677, opened 2023-08-16
[NFT Metadata Crawler] Increase HTTP request for large files

just-in-chang:
### Description
Previously the timeout was too short so many of the semi-large could not finish downloading, this increases that timeout.


wlawt (approved):
lgtm would just change the number to a const variable

rtso (approved):

PR #9548, opened 2023-08-08
Swich to gcp for reading backup data

areshand:
### Description
install google sdk
use the gcp bucket instead

### Test Plan
https://github.com/aptos-labs/aptos-core/actions/runs/5815172944/job/15766231496 Test replay verify successfully started

PR #9679, opened 2023-08-17
Revert "Swich to gcp for reading backup data"

areshand:
Reverts aptos-labs/aptos-core#9548


sherry-x (approved):
Looks fine to me, but any reason for why we're doing this?

areshand:
> Looks fine to me, but any reason for why we're doing this?

our pods cannot reliable read gcs. here is the context https://aptos-org.slack.com/archives/C03EP0XM99D/p1692117776821539

PR #9687, opened 2023-08-17
[DAG] Adding randomness field in dag node

danielxiangzl:
### Description
Reserve the field for randomness shares in the dag node, to make future randomness upgrades compatible. 
 
### Test Plan

Existing tests.


consensus/src/dag/types.rs
```diff
@@ -46,6 +46,7 @@ struct NodeWithoutDigest<'a> {
     timestamp: u64,
     payload: &'a Payload,
     parents: &'a Vec<NodeCertificate>,
+    shares: &'a Vec<u8>, // serialized bytes of randomness shares
```
ibalajiarun, line 49: This should be an `Option<>`?

ibalajiarun: Also, annotate with `#[serde(with = "serde_bytes")]` (lessons from last week 😬 )

zekun000 (approved):

consensus/src/dag/types.rs
```diff
@@ -38,6 +38,18 @@ impl TDAGMessage for CertifiedAck {
     }
 }
 
+#[derive(Clone, Serialize, Deserialize, CryptoHasher, Debug, PartialEq)]
+pub enum Extensions {
+    Empty,
```
ibalajiarun, line 43: nit: None is probably more appropriate.

```suggestion
    None,
```

ibalajiarun (approved):
So, what's the idea with the enums? Will there be versioned structs as we add more fields later? Like V1(...), V2(...), etc?

zekun000:
@ibalajiarun Yes, it's basically a versioning for more fields.

PR #9693, opened 2023-08-20
trivial: add required feature to dependency to fix tests

msmouse:
`e2e-move-tests` when running alone doesn't pass because `pub const GAS_UNIT_PRICE: u64 = 0;` is not in effect

### Test Plan

`RUST_MIN_STACK=4297152 RUST_BACKTRACE=1 cargo test -p e2e-move-tests -- --nocapture`


grao1991 (approved):
Add spaces before and after `=` ?

zekun000 (approved):

wrwg (approved):

Issue #9696, opened 2023-08-21
[compiler v2] Support inline functions


wrwg:
Compiler v2 does not yet support inline functions. While the type checker does support inline functions including lambdas, the inlines are not expanded, leading to the errors in the stackless bytecode generator. 

We likely should support lambdas and higher functions in the stackless bytecode as that is where we eventually want to go, also in the Move VM. Thus expansion of inline functions should happen likely as phase in the bytecode processing pipeline, and not at bytecode generation times.

Issue #9330, opened 2023-07-27
[compiler v2] Constant folding and constant references


wrwg:
Apparently, the v1 compiler supports constant folding and this effects type checking and allowed values for constants, for example:

```
const C: u64 = 1 + 2;
const BYTES: vector<u8> = b"hello";
const VALUES: vector<u64> = vector[1, 2, 3];
```

It is also possible to take references for constants, e.g. `&BYTES`. This is not supported right now by the v2 compiler. There are a number of test cases around this in `move-compiler/transactional-tests/tests/constants/.

wrwg:
It also turns out that the v1 compiler supports something like `1 << var` and the result will still be an integer of any desired type, similar as a direct constant. This lets compiler v2 currently fail to compile `fixed_point32` in the Move stdlib.

wrwg:
> It also turns out that the v1 compiler supports something like `1 << var` and the result will still be an integer of any desired type, similar as a direct constant. This lets compiler v2 currently fail to compile `fixed_point32` in the Move stdlib.

This particular issue has been fixed, so its only constant folding itself.

vineethk:
Some clarification here: what is remaining to be done is implement constant folding as an optimization pass.

wrwg:
> Some clarification here: what is remaining to be done is implement constant folding as an optimization pass.

Yes, but there is one functional aspect as well, namely that one can write `const FOO: u64 = 1 + 2`, which is currently not possible. However, I don't think there are many use cases for this.

brmataptos:
I take this as the problem of (1) writing move-model AST which represents non-literal constant expressions such as the above, (2) [optional] a pass to rewrite those non-literal expressions to literals before code generation, and (3) supporting constants like `&BYTES`.

vineethk:
We should revert this change: https://github.com/aptos-labs/aptos-core/pull/11101/files (i.e., change this function back to a constant: https://github.com/aptos-labs/aptos-core/blob/d1f883907cd11d3fee5847fcbbc7d19545018fb5/aptos-move/framework/aptos-stdlib/tests/fixedpoint64_tests.move#L5) after this issue is resolved.

PR #9697, opened 2023-08-21
[compiler v2] Integration into package system and Aptos CLI

wrwg:
### Description

This integrates the v2 compiler into the `move-package` code and makes it available to the `aptos` CLI. One can now compile with the v2 compiler as in

```
aptos move compile --compiler-version=v2
```

However, compilation of already the Move stdlib fails because of #9330 (effecting bit operations in `fixed_point32.move`) and #9696 (effecting inline functions with lambdas in `vector.move`). Fixing those issues does not fit in this PR which is by itself not trivial.

Implementation wise, since the package system is build on top of internals of the v1 compiler, we provide a new v2 driver function parameter to the general entrypoint `compile_with_driver`. We also add a new flag for the compiler version which is plumbed through many layers of configurations from Aptos CLI down to Move compiler (something which cleaning up is out of scope for this PR).

This PR also disables the use of the bytecode produced by v2 for production, by assigning a new experimental version number to it which is only allowed in `testing` feature mode.

### Test Plan

Tested manually.


third_party/move/move-binary-format/src/file_format_common.rs
```diff
@@ -405,9 +405,14 @@ pub const VERSION_5: u32 = 5;
...
 pub const VERSION_6: u32 = 6;
 
-// Mark which version is the latest version
+/// Mark which version is the latest version
 pub const VERSION_MAX: u32 = VERSION_6;
 
+/// A unique version value which is used for experimental code which is not allowed in
+/// production. The bytecode deserializer accepts modules with this version only when the
+/// cargo feature `testing` is enabled.
+pub const VERSION_EXPERIMENTAL: u32 = u32::MAX;
```
rahxephon89, line 414: We will keep this until compiler v2 is in production?

third_party/move/tools/move-package/src/lib.rs
```diff
@@ -148,6 +154,22 @@ pub struct BuildConfig {
...
+    /// Compiler version to use
+    #[clap(long = "compiler-version", global = true)]
+    pub compiler_version: Option<CompilerVersion>,
+}
+
+#[derive(ValueEnum, Clone, Copy, Debug, Serialize, Deserialize, Eq, PartialEq, PartialOrd)]
+pub enum CompilerVersion {
+    V1,
+    V2,
+}
```
gregnazario, line 167: We may want to consider more fine grained compiler versions in the future.

third_party/move/tools/move-cli/src/base/test.rs
```diff
@@ -13,13 +13,14 @@ use move_compiler::{
     PASS_CFGIR,
 };
 use move_coverage::coverage_map::{output_map_to_file, CoverageMap};
-use move_package::{compilation::build_plan::BuildPlan, BuildConfig};
+// if windows
```
gregnazario, line 16: This looks like an awkward place to have this, is it supposed to be conditional inputs?

third_party/move/testing-infra/transactional-test-runner/src/framework.rs
```diff
@@ -680,18 +680,15 @@ fn compile_source_unit_v2(
...
     let error_str = String::from_utf8_lossy(&error_writer.into_inner()).to_string();
-    let (mut modules, mut scripts) =
+    let (_, mut units) =
         result.map_err(|_| anyhow::anyhow!("compilation errors:\n {}", error_str))?;
-    let unit = if modules.is_empty() && scripts.len() == 1 {
-        (None, scripts.pop())
-    } else if scripts.is_empty() && modules.len() == 1 {
-        (modules.pop(), None)
+    let unit = if units.len() != 1 {
+        anyhow::bail!("expected either one script or one module")
```
gregnazario, line 686: Any reason not to give the module and scripts size anymore?

third_party/move/move-compiler-v2/transactional-tests/tests/control_flow/sorter.exp
```diff
@@ -286,7 +286,7 @@ B0:
 
 >>> V2 Compiler {
 == BEGIN Bytecode ==
-// Move bytecode v6
+// Move bytecode v4294967295
```
gregnazario, line 289: This seems wrong

third_party/move/move-compiler-v2/transactional-tests/tests/control_flow/sorter.exp
```diff
@@ -567,7 +567,7 @@ B5:
 task 1 'run'. lines 67-73:
 
 == BEGIN Bytecode ==
-// Move bytecode v6
+// Move bytecode v4294967295
```
gregnazario, line 570: Similarly

third_party/move/move-compiler-v2/tests/file-format-generator/assign.exp
```diff
@@ -106,7 +106,7 @@ fun assign::assign_struct($t0: &mut assign::S) {
 
 
 ============ disassembled file-format ==================
-// Move bytecode v6
```
gregnazario, line 109: Similarly on the move bytecode version needs to be checked

third_party/move/move-compiler/src/compiled_unit.rs
```diff
@@ -44,8 +44,6 @@ pub struct SpecInfo {
 #[derive(Debug, Clone)]
 pub struct FunctionInfo {
     pub spec_info: BTreeMap<SpecId, SpecInfo>,
-    pub parameters: Vec<(Var, VarInfo)>,
-    pub attributes: Attributes,
 }
```
gregnazario, line 47: Where is this stuff stored now?

third_party/move/move-analyzer/src/symbols.rs
```diff
@@ -645,40 +648,45 @@ impl Symbolicator {
...
+                    Err(diags) => {
+                        let failure = false;
+                        diagnostics = Some((diags, failure));
+                        eprintln!("bytecode compilation failed");
+                        return Ok((files, vec![]));
+                    },
+                };
+                // warning diagnostics (if any) since compilation succeeded
+                if !diags.is_empty() {
+                    // assign only if non-empty, otherwise return None to reset previous diagnostics
```
gregnazario, line 681: Please move this to a separate function for readability purposes.

third_party/move/move-binary-format/src/file_format_common.rs
gregnazario, line 414: Is there supposed to be a different version of the bytecode for the V2 compiler, or will it be completely compatible with the old version?

third_party/move/move-compiler-v2/Cargo.toml
```diff
@@ -12,10 +12,15 @@ edition = "2021"
 [dependencies]
 anyhow = "1.0.62"
 move-binary-format = { path = "../move-binary-format" }
+move-bytecode-source-map = { path = "../move-ir-compiler/move-bytecode-source-map" }
+move-command-line-common = { path = "../move-command-line-common" }
+#move-package = { path = "../tools/move-package" }
```
gregnazario, line 17: let's remove this line

third_party/move/move-compiler-v2/src/file_format_generator/module_generator.rs
```diff
@@ -83,12 +90,26 @@ impl ModuleGenerator {
...
+                let name = IR_AST::ModuleName(IR_SYMBOL::Symbol::from(
+                    ctx.symbol_to_str(module_env.get_name().name()),
+                ));
+                Some(IR_AST::ModuleIdent::new(
+                    name,
+                    module_env.get_name().addr().expect_numerical(),
+                ))
+            };
+            SourceMap::new(ctx.env.to_ir_loc(&module_env.get_loc()), module_name_opt)
+        };
```
gregnazario, line 112: Any particular reason it's in the surrounding block?  Makes it a little less readable

gregnazario (approved):
Seems mostly fine, I mostly have readability comments.

The UNUSED one, maybe we should make just a really high known number, since I suspect we'll never get that high, and it'll be more obvious, like 60,000

third_party/move/move-binary-format/src/file_format_common.rs
wrwg: In the first iteration, we aim at v6 (the current version), so fully compatible.

third_party/move/move-compiler-v2/Cargo.toml
wrwg: Will do next iteration

third_party/move/move-analyzer/src/symbols.rs
wrwg: That's actually old code, it was just indented. I agree it looks very messy. Hopefully we retire this path together with the old compiler. I factor it out next time I get along.

third_party/move/tools/move-cli/src/base/test.rs
wrwg: This was moved by autoformatter after reformating imports. Move it back to where it was, before the cfg!

PR #9698, opened 2023-08-21
[compiler v2] Move the stackless bytecode crate out of the prover

wrwg:
### Description

This moves the `move-stackless-bytecode` crate out of its current location in the `move-prover` into the `move-model` tree. This is a pure relocation. Subsequent PRs will split off prover specific code and move it back to the prover tree. 

### Test Plan

Stuff still works


movekevin (approved):

rahxephon89 (approved):
LGTM, thanks!

PR #9704, opened 2023-08-21
bump CLI version to 2.1.0

0xjinn:

### DIFF
--- a/crates/aptos/CHANGELOG.md
+++ b/crates/aptos/CHANGELOG.md
@@ -2,8 +2,12 @@
 
 All notable changes to the Aptos CLI will be captured in this file. This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html) and the format set out by [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
-## Unreleased
+## [2.1.0] - 2023/08/21
+### Updated
 - Updated CLI source compilation to use rust toolchain version 1.71.1 (from 1.71.0).
+### Added
+- Added basic ledger support for CLI
+  - Example: `aptos init --ledger` to create a new account from ledger
 
 ## [2.0.3] - 2023/08/04
 ### Fixed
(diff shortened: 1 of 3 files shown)


crates/aptos/CHANGELOG.md
```diff
@@ -2,8 +2,12 @@
 
 All notable changes to the Aptos CLI will be captured in this file. This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html) and the format set out by [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
-## Unreleased
```
gregnazario, line 5: I would leave the `Unreleased` so people don't forget about it.

crates/aptos/CHANGELOG.md
```diff
@@ -2,8 +2,12 @@
...
+## [2.1.0] - 2023/08/21
+### Updated
 - Updated CLI source compilation to use rust toolchain version 1.71.1 (from 1.71.0).
+### Added
+- Added basic ledger support for CLI
+  - Example: `aptos init --ledger` to create a new account from ledger
```
gregnazario, line 10: Maybe we should say you can do `aptos init --ledger` to create a profile from ledger, and from then on you can use it the same as any other profile?

0xjinn: Good call!

crates/aptos/CHANGELOG.md
```diff
@@ -3,7 +3,13 @@
 All notable changes to the Aptos CLI will be captured in this file. This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html) and the format set out by [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
+
+## [2.1.0] - 2023/08/21
+### Updated
 - Updated CLI source compilation to use rust toolchain version 1.71.1 (from 1.71.0).
+### Added
+- Added basic ledger support for CLI
+  - Example: `aptos init --ledger` to create a new profile from ledger. After this, you can use it the same way as other normal profile.
```
gregnazario, line 12:
```suggestion
  - Example: `aptos init --ledger` to create a new profile from ledger. After this, you can use it the same way as other profiles.
```

gregnazario (approved):

xbtmatt (approved):

PR #9713, opened 2023-08-22
[compiler v2] Stackless Bytecode Refactoring

wrwg:
### Description

This is the 2nd step in the refactoring started in #9698.  This splits off prover specific parts from the crate `move-model/bytecode` into `move-prover/bytecode-pipeline`. Sharable dataflow analysis and transformation processors, like livevar and reaching definitions, stay. The tests have been split as well, and a common testing driver has been moved in its own test utility crate.

Github does not nicely show diffs like this, but this is functionally a no-op which only Moves code around.

### Test Plan

Existing tests

The diff in the baseline files of livevar and reaching def analysis comes from that we now run this independent of prover specific processors like reference elimination, and is expected.


aalok-t (approved):

third_party/move/move-model/bytecode/tests/reaching_def/basic_test.exp
```diff
@@ -57,8 +57,8 @@ fun ReachingDefTest::basic($t0|a: u64, $t1|b: u64): u64 {
 
 
 [variant baseline]
-fun ReachingDefTest::create_resource($t0|sender: signer) {
-     var $t1: signer
+fun ReachingDefTest::create_resource($t0|sender: &signer) {
+     var $t1: &signer
```
davidiw, line 61: how do changes like this make it in here?

these are generated, right?

curious if this will also change the resulting byte code?

davidiw (approved):
seems fine or tests would be blowing up :grin:

third_party/move/move-model/bytecode/tests/reaching_def/basic_test.exp
wrwg: As mentioned in the PR description, for some of those tests, previously we run some bytecode transformations specific to the prover before the tested feature's transformation (here, reaching_defs is not longer preceded by reference elimination). This is not longer possible after splitting the crates.

PR #9719, opened 2023-08-22
[Stake] Added `update_commission_percentage`

junkil-park:
### Description
This PR allows updating the commission percentage for a vesting contract, which is the feature that the [AIP-48](https://github.com/aptos-foundation/AIPs/blob/main/aips/aip-48.md) proposes.
- Added the function `vesting::update_commission_percentage` and test cases
- Deprecated `staking_contract::update_commision` because the function name has a typo.
- Instead, added `staking_contract::update_commission_percentage`

### Test Plan
aptos move test

### DIFF
--- a/aptos-move/framework/aptos-framework/sources/vesting.move
+++ b/aptos-move/framework/aptos-framework/sources/vesting.move
@@ -760,6 +760,21 @@ module aptos_framework::vesting {
         update_operator(admin, contract_address, new_operator, commission_percentage);
     }
 
+    public entry fun update_commission_percentage(
+        admin: &signer,
+        contract_address: address,
+        new_commission_percentage: u64,
+    ) acquires VestingContract {
+        let operator = operator(contract_address);
+        let vesting_contract = borrow_global_mut<VestingContract>(contract_address);
+        verify_admin(admin, vesting_contract);
+        let contract_signer = &get_vesting_account_signer_internal(vesting_contract);
+        staking_contract::update_commision(contract_signer, operator, new_commission_percentage);
+        vesting_contract.staking.commission_percentage = new_commission_percentage;
+        // This function does not emit an event. Instead, `staking_contract::update_commission_percentage`
+        // emits the event for this commission percentage update.
+    }
+
     public entry fun update_voter(
         admin: &signer,
         contract_address: address,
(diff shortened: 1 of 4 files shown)


aptos-move/framework/aptos-framework/sources/vesting.move
```diff
@@ -1460,6 +1475,84 @@ module aptos_framework::vesting {
...
+        let expected_commission = 0;
+
+        // Stake pool earns some rewards.
+        stake::end_epoch();
+        let (_, accumulated_rewards, _) = staking_contract::staking_contract_amounts(contract_address, operator_address);
+
+        // Update commission percentage to 20%. This also immediately requests commission.
+        update_commission_percentage(admin, contract_address, 20);
+        // Commission is calculated using the previous commission percentage which is 10%.
+        expected_commission = expected_commission + (accumulated_rewards / 10);
```
movekevin, line 1534: Nit: why not just define expected_commission here since it’s 0 before here anyway

aptos-move/framework/aptos-framework/sources/vesting.move
```diff
@@ -1460,6 +1475,84 @@ module aptos_framework::vesting {
         assert!(operator_commission_percentage(contract_address) == 10, 0);
     }
 
+    #[test(aptos_framework = @0x1, admin = @0x123, operator = @0x345)]
+    public entry fun test_update_commission_percentage(
```
movekevin, line 1479: This test seems redundant with the other test added below. Does this test any code path that the other test doesn't?

movekevin (approved):

movekevin:
Can you add a link to the correspondIng AIP here? (Aip 48)

junkil-park:
> Can you add a link to the correspondIng AIP here? (Aip 48)

@movekevin , I added the link to the description of this PR.

aptos-move/framework/aptos-framework/sources/vesting.move
junkil-park: It is a more basic test case, but it seems redundant. It's merged into the other one now.

aptos-move/framework/aptos-framework/sources/vesting.move
junkil-park: Updated accordingly.

aptos-move/framework/aptos-framework/sources/staking_contract.move
```diff
@@ -372,9 +372,13 @@ module aptos_framework::staking_contract {
         emit_event(&mut store.reset_lockup_events, ResetLockupEvent { operator, pool_address });
     }
 
-    /// Convenience function to allow a staker to update the commission percentage paid to the operator.
-    /// TODO: fix the typo in function name. commision -> commission
+    #[deprecated]
```
sherry-x, line 375: shall we leave a todo here to cleanup later? what's the process to cleanup deprecated entry functions? looks like lots of places would need to be updated if we remove this code.

junkil-park: I see. 

@movekevin , @brianrmurphy, what is the implication of the attribute `#[deprecated]`, and what is the proper clean up process?

brianrmurphy:
Better to use brmataptos, but:

#[deprecated] does nothing now.  Eventually we'll have a warning from the
compiler if you use the thing marked as deprecated.

aptos-move/framework/aptos-framework/sources/staking_contract.move
gregnazario: You cannot remove deprecated functions due to backwards compatibility rules.

You first can put `#[deprecated]` and then later make an update to make it abort rather than do any changes.

aptos-move/framework/aptos-framework/sources/staking_contract.move
```diff
@@ -372,9 +372,13 @@ module aptos_framework::staking_contract {
...
     public entry fun update_commision(staker: &signer, operator: address, new_commission_percentage: u64) acquires Store, StakingGroupUpdateCommissionEvent {
+        update_commission_percentage(staker, operator, new_commission_percentage);
+    }
+
+    /// Allow staker to update the commission percentage paid to the operator.
+    public entry fun update_commission_percentage(staker: &signer, operator: address, new_commission_percentage: u64) acquires Store, StakingGroupUpdateCommissionEvent {
```
gregnazario, line 381: I'm not sure why the previous one needs to be deprecated if it just calls the other one?  Let's stick with the original one so anyone who's automated around it doesn't have to change, will never be able to get rid of it.

But I see the main thing you're doing here is fixing the spelling.

gregnazario (approved):

aptos-move/framework/aptos-framework/sources/staking_contract.move
movekevin: Yeah unfortunately we cannot delete public functions as that'd break backward compat

aptos-move/framework/aptos-framework/sources/staking_contract.move
junkil-park: @gregnazario, I see. We should keep the original one `update_commision` if the compatibility breaks without it.

How about the new one (i.e., `staking_contract::update_commission_percentage`)? Should we not add a new one?

aptos-move/framework/aptos-framework/sources/staking_contract.move
junkil-park: I updated the PR to keep the original one, not adding a new one.

PR #9725, opened 2023-08-23
[Experimental] Add ThreadManager.

grao1991:
### Description
This is just a start point. We have too many threads in the system. I'm trying to come up a way to manager all threads at a common place. I'm just putting two simple strategies as an example, please just treat this as an example and playground, and implement better things on top of it.

### DIFF
--- a/crates/aptos-runtimes/src/lib.rs
+++ b/crates/aptos-runtimes/src/lib.rs
@@ -1,8 +1,7 @@
 // Copyright © Aptos Foundation
 // SPDX-License-Identifier: Apache-2.0
 
-#![forbid(unsafe_code)]
-
+use rayon::{ThreadPool, ThreadPoolBuilder};
 use std::sync::atomic::{AtomicUsize, Ordering};
 use tokio::runtime::{Builder, Runtime};
 
@@ -14,6 +13,17 @@ const MAX_THREAD_NAME_LENGTH: usize = 12;
 /// Returns a tokio runtime with named threads.
 /// This is useful for tracking threads when debugging.
 pub fn spawn_named_runtime(thread_name: String, num_worker_threads: Option<usize>) -> Runtime {
+    spawn_named_runtime_with_start_hook(thread_name, num_worker_threads, || {})
+}
+
+pub fn spawn_named_runtime_with_start_hook<F>(
+    thread_name: String,
+    num_worker_threads: Option<usize>,
+    on_thread_start: F,
+) -> Runtime
+where
+    F: Fn() + Send + Sync + 'static,
+{
     // Verify the given name has an appropriate length
     if thread_name.len() > MAX_THREAD_NAME_LENGTH {
         panic!(
@@ -31,6 +41,7 @@ pub fn spawn_named_runtime(thread_name: String, num_worker_threads: Option<usize
             let id = atomic_id.fetch_add(1, Ordering::SeqCst);
             format!("{}-{}", thread_name_clone, id)
         })
+        .on_thread_start(on_thread_start)
         .disable_lifo_slot()
         .enable_all();
     if let Some(num_worker_threads) = num_worker_threads {
@@ -41,6 +52,49 @@ pub fn spawn_named_runtime(thread_name: String, num_worker_threads: Option<usize
     builder.build().unwrap_or_else(|error| {
         panic!(
             "Failed to spawn named runtime! Name: {:?}, Error: {:?}",
+            thread_name, error
+        )
+    })
+}
+
+/// Returns a rayon threadpool with threads.
+/// This is useful for tracking threads when debugging.
+pub fn spawn_rayon_thread_pool(
+    thread_name: String,
+    num_worker_threads: Option<usize>,
+) -> ThreadPool {
+    spawn_rayon_thread_pool_with_start_hook(thread_name, num_worker_threads, || {})
+}
+
+pub fn spawn_rayon_thread_pool_with_start_hook<F>(
+    thread_name: String,
+    num_worker_threads: Option<usize>,
+    on_thread_start: F,
+) -> ThreadPool
+where
+    F: Fn() + Send + Sync + 'static,
+{
+    // Verify the given name has an appropriate length
+    if thread_name.len() > MAX_THREAD_NAME_LENGTH {
+        panic!(
+            "The given runtime thread name is too long! Max length: {}, given name: {}",
+            MAX_THREAD_NAME_LENGTH, thread_name
+        );
+    }
+
+    let thread_name_clone = thread_name.clone();
+    let mut builder = ThreadPoolBuilder::new()
+        .thread_name(move |index| format!("{}-{index}", thread_name_clone))
+        .start_handler(move |_| on_thread_start());
+
+    if let Some(num_worker_threads) = num_worker_threads {
+        builder = builder.num_threads(num_worker_threads);
+    }
+
+    // Spawn and return the threadpool
+    builder.build().unwrap_or_else(|error| {
+        panic!(
+            "Failed to spawn named rayon thread pool! Name: {:?}, Error: {:?}",
             thread_name, error
         )
     })
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1356,6 +1356,19 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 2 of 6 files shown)


crates/aptos-runtimes/src/thread_manager.rs
```diff
@@ -0,0 +1,218 @@
...
+        &self.io_threads
+    }
+
+    fn new(strategy: ThreadConfigStrategy) -> Self {
+        match strategy {
+            ThreadConfigStrategy::DefaultStrategy => {
+                let exe_threads = Self::create_threads(ThreadsConfigs::default(), "exe");
+                let non_exe_threads = Self::create_threads(ThreadsConfigs::default(), "non_exe");
+                let io_threads =
+                    Self::create_io_threads(ThreadsConfig::new_without_start_hook(64), "io");
```
sitalkedia, line 115: 64 is very high, is this what we are using today?

crates/aptos-runtimes/src/thread_manager.rs
```diff
@@ -0,0 +1,218 @@
...
+                let mut non_exe_cpu_set = Self::new_cpu_set();
+                for core_id in core_ids.iter().take(num_exe_cpu) {
+                    unsafe { CPU_SET(core_id.id, &mut exe_cpu_set) };
+                }
+                for core_id in core_ids.iter().skip(num_exe_cpu) {
+                    unsafe { CPU_SET(core_id.id, &mut non_exe_cpu_set) };
+                }
+
+                let exe_threads = Self::create_threads(
+                    ThreadsConfigs {
```
sitalkedia, line 138: Why do we need tokio for executor threads?

crates/aptos-runtimes/src/thread_manager.rs
```diff
@@ -0,0 +1,218 @@
...
+
+                let non_exe_threads = Self::create_threads(
+                    ThreadsConfigs {
+                        tokio_config: ThreadsConfig::new(
+                            core_ids.len() - num_exe_cpu,
+                            Self::pin_cpu_set(exe_cpu_set),
+                        ),
+                        rayon_config: ThreadsConfig::new(
+                            core_ids.len() - num_exe_cpu,
+                            Self::pin_cpu_set(exe_cpu_set),
```
sitalkedia, line 159: `non_exe_cpu_set` ?

sitalkedia (commented):
Wondering if you observed any performance improvement using CPU pinning approach.

crates/aptos-runtimes/src/lib.rs
```diff
@@ -45,3 +58,46 @@ pub fn spawn_named_runtime(thread_name: String, num_worker_threads: Option<usize
...
+    if thread_name.len() > MAX_THREAD_NAME_LENGTH {
+        panic!(
+            "The given runtime thread name is too long! Max length: {}, given name: {}",
+            MAX_THREAD_NAME_LENGTH, thread_name
+        );
+    }
+
+    let thread_name_clone = thread_name.clone();
+    let mut builder = ThreadPoolBuilder::new()
+        .thread_name(move |index| format!("{}-{index}", thread_name_clone))
```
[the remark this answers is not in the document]
areshand: Yunus's play with some tool in his intern project that can trim the suffix and visualize the threads together. One caveat the tool is licenced so we cannot merge it into our public main branch

crates/aptos-runtimes/src/thread_manager.rs
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    pub fn get_io_pool(&self) -> &ThreadPool {
+        &self.io_threads
+    }
+
+    fn new(strategy: ThreadConfigStrategy) -> Self {
+        match strategy {
+            ThreadConfigStrategy::DefaultStrategy => {
+                let exe_threads = Self::create_threads(ThreadsConfigs::default(), "exe");
```
areshand, line 112: have a enum for "exe", "non-exe" and "io"?

crates/aptos-runtimes/src/thread_manager.rs
```diff
@@ -0,0 +1,218 @@
...
+use tokio::runtime::{Handle, Runtime};
+
+pub static THREAD_MANAGER: Lazy<ThreadManager> =
+    Lazy::new(|| ThreadManager::new(ThreadManager::get_thread_config_strategy()));
+
+static THREAD_CONFIG_STRATEGY: OnceCell<ThreadConfigStrategy> = OnceCell::new();
+
+struct ThreadsConfig {
+    num_threads: usize,
+    on_thread_start: Box<dyn Fn() + Send + Sync + 'static>,
```
areshand, line 18: thread is able to take different functions, should we include the function in the thread_config? for example, the same IO pool was used for JMT and adding chunk data.

areshand:
just curious how would you solve the too many threads problem once you manage all threads with the `thread manager` here?

crates/aptos-runtimes/src/lib.rs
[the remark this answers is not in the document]
msmouse: @grao1991  has dealt with that https://github.com/aptos-labs/aptos-core/pull/9721

msmouse (commented):
Man I'm not too comfortable landing code that's not tested or called at all. Call it somewhere or maybe move to /experimental

msmouse (commented):
Man I'm not too comfortable landing code that's not tested or called at all. Call it somewhere or maybe create separate crate (and move to /experimental)

crates/aptos-runtimes/src/thread_manager.rs
grao1991: We are using multiple threadpools with 32 threads today. 

64 is not really high for real io if we have multiple disks, or using network disks.

crates/aptos-runtimes/src/thread_manager.rs
grao1991: I'm doing some experiments with async code. We can remove it if we end up not using it.

crates/aptos-runtimes/src/thread_manager.rs
```diff
@@ -0,0 +1,218 @@
...
+    rayon_config: ThreadsConfig,
+}
+
+struct Threads {
+    runtime: Runtime,
+    cpu_pool: ThreadPool,
+}
+
+pub struct ThreadManager {
+    exe_threads: Threads,
```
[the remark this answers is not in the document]
grao1991: Just a temporary way to separate pools for threads that potentially do IO v.s. threads that only do computation. 

I will need to do some experiments to find out a good strategy, it might end up in a totally different way. This PR is just to setup a playground so people can start playing with it.

crates/aptos-runtimes/src/thread_manager.rs
grao1991: This is the function you want to core when the thread is starting. Usually pinning to cpu cores, or setting thread priorities, etc. If these are different they probably should use different thread pool.

grao1991:
> Wondering if you observed any performance improvement using CPU pinning approach.

This depends on what we run.. I can find some 

> just curious how would you solve the too many threads problem once you manage all threads with the `thread manager` here?

Once we manager all threads in the same place we can start looking if we should combine some of the thread pools, adjusting the thread configs, etc. Now threads are created everywhere, which make it hard to track and manage.

crates/aptos-runtimes/src/thread_manager.rs
grao1991: Now they are just used as a string as the thread name.

crates/aptos-runtimes/src/thread_manager.rs
grao1991: Good catch, copied the wrong line when doing refactoring.

sitalkedia:
> Man I'm not too comfortable landing code that's not tested or called at all. Call it somewhere or maybe create separate crate (and move to /experimental)

+1, may be we can create a separate crate for this experimental code?

grao1991:
> > Man I'm not too comfortable landing code that's not tested or called at all. Call it somewhere or maybe create separate crate (and move to /experimental)
> 
> +1, may be we can create a separate crate for this experimental code?

Moved to experimental/

sitalkedia (approved):
LGTM

msmouse (approved):

PR #9729, opened 2023-08-23
[TS-SDK] Fix the coin client transfer/checkBalance functions to allow custom CoinTypes

xbtmatt:
### Description

Updating `CoinClient` `transfer()` and `checkBalance()` to check for a custom `CoinType` first before converting to a `HexString` address.

As of right now, the function assumes it's an address and tries to parse it as such, resulting in a failed conversion to `HexString`:

```typescript
throw new Error('Invalid byte sequence');
```

Added a simple `.includes("::")` check to skip the hex string conversion so it doesn't error if the user sends in a custom cointype.

```typescript
const isTypeTag = (extraArgs?.coinType ?? "").toString().includes("::");
if (extraArgs?.coinType && !isTypeTag && AccountAddress.isValid(extraArgs.coinType)) {
```

Also made the `coinType` arg consistent between the two functions by changing the `checkBalance` type to `string | MaybeHexString` instead of just `string`.

Also added two unit tests that use a custom cointype in the form of `APTOS_COIN` aka `0x1::aptos_coin::AptosCoin` to test it (even that didn't work before).

### Test Plan
```shell
yarn run jest coin_client.test.ts
```

### DIFF
--- a/ecosystem/typescript/sdk/src/plugins/coin_client.ts
+++ b/ecosystem/typescript/sdk/src/plugins/coin_client.ts
@@ -130,7 +130,7 @@ export class CoinClient {
       // The coin type to use, defaults to 0x1::aptos_coin::AptosCoin.
       // If you want to check the balance of a fungible asset, set this param to be the
       // fungible asset address
-      coinType?: string | MaybeHexString;
+      coinType?: MaybeHexString;
     },
   ): Promise<bigint> {
     const isTypeTag = (extraArgs?.coinType ?? "").toString().includes("::");
--- a/ecosystem/typescript/sdk/src/plugins/coin_client.ts
+++ b/ecosystem/typescript/sdk/src/plugins/coin_client.ts
@@ -67,12 +67,13 @@ export class CoinClient {
       createReceiverIfMissing?: boolean;
     },
   ): Promise<string> {
-    // Since we can receive either a fully qualified type tag like "0x1::coin_type::CoinType" or a fungible object address "0x1234...6789"
-    // we first check to see if the raw string value includes "::" to make sure it's not supposed to be a fungible asset object address.
+    // Since we can receive either a fully qualified type tag like "0x1::coin_type::CoinType"
+    // or a fungible object address "0x1234...6789" we first check to see if the raw string value includes "::"
+    // This is to make sure it's not supposed to be a fungible asset object address.
     const isTypeTag = (extraArgs?.coinType ?? "").toString().includes("::");
 
-    // If the coin type exists, definitely isn't a type tag, and is a valid account address, then we enter this if block
-    // under the assumption that it's a fungible asset object address.
+    // If the coin type exists, definitely isn't a type tag, and is a valid account address,
+    // then we enter this if block under the assumption that it's a fungible asset object address.
     if (extraArgs?.coinType && !isTypeTag && AccountAddress.isValid(extraArgs.coinType)) {
       /* eslint-disable no-console */
       console.warn("to transfer a fungible asset, use `FungibleAssetClient()` class for better support");
@@ -138,12 +139,13 @@ export class CoinClient {
       coinType?: string | MaybeHexString;
     },
   ): Promise<bigint> {
-    // Since we can receive either a fully qualified type tag like "0x1::coin_type::CoinType" or a fungible object address "0x1234...6789"
-    // we first check to see if the raw string value includes "::" to make sure it's not supposed to be a fungible asset object address.
+    // Since we can receive either a fully qualified type tag like "0x1::coin_type::CoinType"
+    // or a fungible object address "0x1234...6789" we first check to see if the raw string value includes "::"
+    // This is to make sure it's not supposed to be a fungible asset object address.
     const isTypeTag = (extraArgs?.coinType ?? "").toString().includes("::");
 
-    // If the coin type exists, definitely isn't a type tag, and is a valid account address, then we enter this if block
-    // under the assumption that it's a fungible asset object address.
+    // If the coin type exists, definitely isn't a type tag, and is a valid account address,
+    // then we enter this if block under the assumption that it's a fungible asset object address.
     if (extraArgs?.coinType && !isTypeTag && AccountAddress.isValid(extraArgs.coinType)) {
       /* eslint-disable no-console */
       console.warn("to check balance of a fungible asset, use `FungibleAssetClient()` class for better support");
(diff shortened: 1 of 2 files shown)


ecosystem/typescript/sdk/src/plugins/coin_client.ts
```diff
@@ -129,10 +130,11 @@ export class CoinClient {
       // The coin type to use, defaults to 0x1::aptos_coin::AptosCoin.
       // If you want to check the balance of a fungible asset, set this param to be the
       // fungible asset address
-      coinType?: string;
+      coinType?: string | MaybeHexString;
```
0xmaayan, line 133:
```suggestion
      coinType?: MaybeHexString;
```

since `type MaybeHexString = HexString | string | HexEncodedBytes;`

xbtmatt: oh right, I'll change the other one then

xbtmatt: well actually, I think the string is indicating it can be a regular `TypeTag` string as well. Does it make sense to make it a `MaybeHexString` if it's not intended to be used as a `HexString`?

xbtmatt:
> haha we can also argue them `to transfer a fungible asset, use `FungibleAssetClient()` class for better support`

does the `FungibleAssetClient` support the original `CoinType` generics type of coins as well? But yea :p will at least save people from getting stuck on this

ecosystem/typescript/sdk/src/tests/e2e/coin_client.test.ts
```diff
@@ -19,10 +20,10 @@ test(
     await faucetClient.fundAccount(alice.address(), 100_000_000);
     await faucetClient.fundAccount(bob.address(), 0);
 
-    const txnHash1 = await coinClient.transfer(alice, bob, 42);
+    const txnHash1 = await coinClient.transfer(alice, bob, 42, { coinType: APTOS_COIN });
```
movekevin, line 23: Can we update the your coin example to use the transfer function from coin_client?

ecosystem/typescript/sdk/src/tests/e2e/coin_client.test.ts
```diff
@@ -6,6 +6,7 @@ import { getFaucetClient, longTestTimeout, NODE_URL } from "../unit/test_helper.
 import { AptosAccount } from "../../account/aptos_account";
 import { COIN_TRANSFER, CoinClient, TRANSFER_COINS } from "../../plugins/coin_client";
 import { EntryFunctionPayload, Transaction_UserTransaction } from "../../generated";
+import { APTOS_COIN } from "../../utils";
 
 test(
   "transfer and checkBalance works",
```
movekevin, line 12: This test is gigantic and is chewing off more than it should. We should break this into several different tests:
1. Transferring and checking APT balances. This includes specifying vs not specifying coinType
2. Transferring and checking non-APT balances
3. Transferring and checking fungible assets balances
4. Error cases such as insufficient balances, invalid coin types, non-existent coin types/fungible assets

movekevin: This should also have at least a simple test for fungible asset transfer since it has that branch

davidiw (approved):

davidiw:
Agree with Kevin, but probably should file an issue.

ecosystem/typescript/sdk/src/tests/e2e/coin_client.test.ts
xbtmatt: Ok, I was going to put it in here, but the test case requires publishing and a few other logistical things, so I'll just file an issue for it so that we can better track all the test coverage we're missing

ecosystem/typescript/sdk/src/tests/e2e/coin_client.test.ts
movekevin: You can file an issue for this as well @xbtmatt

movekevin (approved):

gregnazario (commented):
Can we test if it works on a custom coin type?  I'm reading this code, and I'm not sure how it's supposed to work?

ecosystem/typescript/sdk/src/plugins/coin_client.ts
```diff
@@ -129,10 +130,11 @@ export class CoinClient {
       // The coin type to use, defaults to 0x1::aptos_coin::AptosCoin.
       // If you want to check the balance of a fungible asset, set this param to be the
       // fungible asset address
-      coinType?: string;
+      coinType?: MaybeHexString;
     },
   ): Promise<bigint> {
-    if (extraArgs?.coinType && AccountAddress.isValid(extraArgs.coinType)) {
+    const isTypeTag = (extraArgs?.coinType ?? "").toString().includes("::");
+    if (extraArgs?.coinType && !isTypeTag && AccountAddress.isValid(extraArgs.coinType)) {
```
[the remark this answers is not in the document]
xbtmatt: It's possible I made a wrong assumption that both functions checkBalance and coinTransfer can handle both v1 and v2 coins (CoinType and addresses) but that is why this exists. It's confusing because the string option can be for a CoinType aka "0x1::coin_type::CoinType" or a loosely typed hex string for the address of a fungible asset object address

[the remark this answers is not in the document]
gregnazario: Ah, I see V2 type coins, gotta test that out then

ecosystem/typescript/sdk/src/plugins/coin_client.ts
```diff
@@ -129,10 +130,11 @@ export class CoinClient {
       // The coin type to use, defaults to 0x1::aptos_coin::AptosCoin.
       // If you want to check the balance of a fungible asset, set this param to be the
       // fungible asset address
-      coinType?: string;
+      coinType?: MaybeHexString;
```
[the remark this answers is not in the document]
movekevin: This function also handles fungible assets so this can be an address in that case

ecosystem/typescript/sdk/src/plugins/coin_client.ts
xbtmatt: I'm going to change it back (to the new change, aka `string | MaybeHexString`) to make it clearer that you can use either a TypeTag string or a (Maybe)HexString address

gregnazario (approved):

PR #9739, opened 2023-08-23
Add graph streaming partitioners

xosmig:
A prototype of transaction partitioning in the order-then-partition paradigm in the batched streaming model based on well-known graph partitioning algorithms.

Three partitioners are implemented:
* Random (baseline) 
* Fennel -- state-of-the-art single-pass streaming partitioning protocol. It is supposed to be lightning-fast, with reasonable partitioning quality.
* Metis -- state-of-the-art multilevel partitioning software. The implementation in this PR only provides a wrapper that requires [Metis](http://glaros.dtc.umn.edu/gkhome/metis/metis/overview) to be installed (see the documentation for the [metis crate](https://crates.io/crates/metis)). It partitions the whole graph at once and, hence, doesn't really fit the batched streaming model, but is useful as a baseline.

On a test with 100k p2p transactions on 1k accounts split over 100 batches and 60 partitions, the fractions of cut edge weights are as follows:
* Random: 0.98
* Fennel: 0.13
* Metis: 0.05

Here, an edge is a dependency between two transactions and its weight is a function of how far the transactions are from each other in the serialization order. An edge is "cut" if the two transactions are in different partitions.

Can be reproduced with the following command (remove `--features metis-partitioner` if Metis is not installed):
```
cargo test -p aptos-streaming-partitioner --features metis-partitioner -- --nocapture --test-threads 1
```

This PR also provides several abstractions to facilitate the implementation of graph algorithms and computation in the batched streaming model. 

Next, I'm considering trying to improve the partitioning quality of Fennel by adding a round of community detection.


zjma:
Looks promising. @xosmig can these partitioners partition deterministically?

zjma:
also any built-in load balancing?

xosmig:
Both implementations are deterministic. However, in the order-then-partition paradigm, deterministic partitioning is not necessary as the end result of the execution only depends on the serialization order, which is fixed before the partitioning. 

Eventually, I would like to try to actually use this fact by implementing dynamic load balancing, with (non-deterministic) feedback from the executioner.

xosmig:
> also any built-in load balancing?

You can specify the "balance constraint", which is how far the weight of a partition is allowed to deviate from the total weight divided by the number of partitions. By default, I set it to 10%, i.e., if the total weight of all transactions is `W` and there are `K` partitions, the weight of transactions in each partition must be at most `1.1 W / K`.

In the case of Metis, it will try to minimize the weight of cut edges while respecting the given balancing constraint, but it will not try to minimize the imbalance as long as it is within the allowed range.

In the case of Fennel, it has the objective function that combines the weight of cut edges and the load balancing, so it will "prefer" more balanced partitioning.

xosmig:
I added the computation of the imbalance to the test output. As expected, Metis has exactly the allowed 10% while Fennel has only 6%.

PR #9740, opened 2023-08-23
[Execution][Benchmark] Add support for a pre-generating all transactions in the block-stm benchmark

sitalkedia:
### Description

Add a change to support pre-generating all transactions - this helps to isolate the noise from CPU usage in generating the transactions and also measures CPU utilization of execution only phase.

### Test Plan

```
cargo run --release -p aptos-transaction-benchmarks execute --generate-then-execute --num-accounts 100000 --block-size 50000 --concurrency-level-per-shard 10
```

### DIFF
--- a/aptos-move/aptos-transaction-benchmarks/src/main.rs
+++ b/aptos-move/aptos-transaction-benchmarks/src/main.rs
@@ -186,8 +186,7 @@ fn execute(opt: ExecuteOpt) {
         opt.num_executor_shards,
         opt.concurrency_level_per_shard,
         opt.remote_executor_addresses,
-        false,
-        //opt.no_conflict_txns,
+        opt.no_conflict_txns,
         opt.maybe_block_gas_limit,
         opt.generate_then_execute,
     );
--- a/aptos-move/aptos-transaction-benchmarks/src/main.rs
+++ b/aptos-move/aptos-transaction-benchmarks/src/main.rs
@@ -204,6 +204,7 @@ fn main() {
             .unwrap()
             .as_millis() as i64,
     );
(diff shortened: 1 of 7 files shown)


aptos-move/aptos-transaction-benchmarks/src/transaction_bench_state.rs
```diff
@@ -0,0 +1,326 @@
+// Copyright © Aptos Foundation
+
+use crate::{transactions, transactions::RAYON_EXEC_POOL};
```
sitalkedia, line 3: This is just moving the `TransactionBenchState` to a new module, no logic change.

danielxiangzl (approved):
LGTM!

grao1991 (approved):

aptos-move/aptos-transaction-benchmarks/src/main.rs
```diff
@@ -181,8 +186,10 @@ fn execute(opt: ExecuteOpt) {
         opt.num_executor_shards,
         opt.concurrency_level_per_shard,
         opt.remote_executor_addresses,
-        opt.no_conflict_txns,
+        false,
+        //opt.no_conflict_txns,
```
manudhundi, line 190: remove this commented line ?

manudhundi (approved):
Looks fine. Just left a minor comment

aptos-move/aptos-transaction-benchmarks/src/main.rs
sitalkedia: Good eye, I actually didn't mean to comment this out. Let me fix it.

PR #9741, opened 2023-08-23
[State Sync] Add fast sync support to genesis.

JoshLind:
### Description
This PR updates state sync to support fast syncing to genesis. When a node starts up using fast sync and no epoch changes have occurred in the network, the node will fast sync to genesis before continuing. If an epoch change has occurred, fast sync will still choose the latest epoch ending ledger version (as it already does).

This is required to unblock genesis-less fast syncing (i.e., fast syncing a node without writing the genesis transaction to the DB at all).

### Test Plan
New and existing tests!

### DIFF
--- a/state-sync/state-sync-v2/state-sync-driver/src/bootstrapper.rs
+++ b/state-sync/state-sync-v2/state-sync-driver/src/bootstrapper.rs
@@ -463,8 +463,19 @@ impl<
         let highest_known_ledger_info = self.get_highest_known_ledger_info()?;
         let highest_known_ledger_version = highest_known_ledger_info.ledger_info().version();
 
-        // If we've already synced to the highest known version, there's nothing to do
-        if highest_synced_version >= highest_known_ledger_version {
+        // Check if we need to sync more data
+        if self.get_bootstrapping_mode().is_fast_sync()
+            && highest_synced_version == GENESIS_TRANSACTION_VERSION
+            && highest_known_ledger_version == GENESIS_TRANSACTION_VERSION
+        {
+            // The node is fast syncing and an epoch change isn't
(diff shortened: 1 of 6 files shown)


state-sync/state-sync-v2/state-sync-driver/src/bootstrapper.rs
```diff
@@ -947,7 +968,12 @@ impl<
         // Initialize the state value synchronizer (if not already done)
         if !self.state_value_syncer.initialized_state_snapshot_receiver {
             // Fetch all verified epoch change proofs
-            let epoch_change_proofs = self.verified_epoch_states.all_epoch_ending_ledger_infos();
+            let version_to_sync = ledger_info_to_sync.ledger_info().version();
+            let epoch_change_proofs = if version_to_sync == GENESIS_TRANSACTION_VERSION {
+                vec![ledger_info_to_sync.clone()] // Sync to genesis
+            } else {
+                self.verified_epoch_states.all_epoch_ending_ledger_infos() // Sync beyond genesis
```
areshand, line 975: curious why introducing a branch here. what is contained in the all_epoch_ending_ledger_infos when the version_to_sync is genesis?

areshand (approved):

msmouse (approved):

state-sync/state-sync-v2/state-sync-driver/src/bootstrapper.rs
JoshLind: Aah, it's because `verified_epoch_states.all_epoch_ending_ledger_infos` only contains epoch ending ledger infos beyond genesis. Thus, when the version to sync is genesis, `verified_epoch_states.all_epoch_ending_ledger_infos` is empty.

PR #9742, opened 2023-08-23
[gas-calibration][simple] ignore terms with 0-coefficients

vgao1996:
This revises `fn aggregate_terms` so that it ignores all terms with 0 as coefficients, preventing them from causing errors in the gas calibration tool downstream.

### DIFF
--- a/aptos-move/aptos-abstract-gas-usage/src/algebra_helpers.rs
+++ b/aptos-move/aptos-abstract-gas-usage/src/algebra_helpers.rs
@@ -1,98 +1,98 @@
 // Copyright © Aptos Foundation
 // SPDX-License-Identifier: Apache-2.0
 
-use anyhow::{anyhow, Result};
+use anyhow::{bail, Result};
 use aptos_gas_algebra::DynamicExpression;
 use std::collections::{
     btree_map::Entry::{Occupied, Vacant},
     BTreeMap,
 };
 
-/// expand out the AST and collect only additive terms
+/// Expands a gas expression into a list of terms that add up to the original expression.
+/// A term is an expression that does not contain the add operation.
 ///
-/// ### Arguments
+/// Note that the result is NOT guaranteed to be simplified or in canonical form.
 ///
-/// * `expr` - Root node of the AST
-///
-/// ### Example
-///
-/// * term -> num
-/// * term -> var
-/// * term -> term * term
-/// * case 1: num -> Vec(num)
-/// * case 2: var -> Vec(var)
-/// * case 3: e1 * e2 -> t1 * t2
-/// * case 4: e1 * e2 -> t1 + t2
+/// Here are some examples:
+/// - `1 => [1]`
+/// - `x => [x]`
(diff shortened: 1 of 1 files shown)


wlawt (approved):

aptos-move/aptos-abstract-gas-usage/src/algebra_helpers.rs
```diff
@@ -79,46 +79,48 @@ pub fn expand_terms(expr: DynamicExpression) -> Vec<DynamicExpression> {
 ///
 /// (A + A) * (5 + 5) => 5A + 5A + 5A + 5A => 20A
 pub fn aggregate_terms(terms: Vec<DynamicExpression>) -> Result<BTreeMap<String, u64>> {
+    use DynamicExpression::*;
```
brmataptos, line 82: Usually it's better to enumerate what you use, so changes don't break your code.

aptos-move/aptos-abstract-gas-usage/src/algebra_helpers.rs
```diff
@@ -79,46 +79,48 @@ pub fn expand_terms(expr: DynamicExpression) -> Vec<DynamicExpression> {
...
+                // Seeing a concrete quantity means that the user is not encoding the gas
+                // cost in an abstract way using GasExpression.
+                //
+                // This makes calibration impossible and we should error.
                 if value != 0 {
-                    return Err(anyhow!(
-                        "Concrete quantity provided. Should be GasExpression."
-                    ));
+                    bail!(
+                        "Expected simple term got non-zero concrete value {}.",
```
brmataptos, line 94: Do we all know what "simple term" means?  I assume it's (constant * var) ?

brmataptos (approved):

aptos-move/aptos-abstract-gas-usage/src/algebra_helpers.rs
wlawt: Agreed, we should probably define "simple term" for clarity

aptos-move/aptos-abstract-gas-usage/src/algebra_helpers.rs
davidiw: agreed, I don't see how this is better.

davidiw (approved):

aptos-move/aptos-abstract-gas-usage/src/algebra_helpers.rs
vgao1996: I now manually import all the variants. Thanks for pointing this out!

aptos-move/aptos-abstract-gas-usage/src/algebra_helpers.rs
vgao1996: Revised the comments and now they clearly explain what a simple term means.

vgao1996:
@davidiw @brmataptos I've made the changes you requested and cleaned up the code a bit, getting rid of some unnecessary allocations.

PR #9743, opened 2023-08-24
sdk v2 types

0xmaayan:
### Description
My attempt to get rid of openAPI response types and maintain and declare local types. Biggest chunk is the `Transaction` type and all its sub-types. Probably some more types will be added as we keep developing sdk_v2.

The beauty is
1. all types are under `types/index.ts` file and easier to understand what is going on.
2. when building the package (`pnpm run build`) all types are under `dist/types/index.d.ts` and much more clear.


ecosystem/typescript/sdk_v2/src/types/index.ts
```diff
@@ -84,6 +94,451 @@ export type MoveModuleBytecode = {
...
 };
 
+/////// TRANSACTION TYPES ///////
+
+export type Transaction =
+  | PendingTransaction
+  | UserTransaction
+  | GenesisTransaction
+  | BlockMetadataTransaction
+  | StateCheckpointTransaction;
```
xbtmatt, line 104:
```suggestion
export type TransactionResponse =
  | PendingTransactionResponse
  | UserTransactionResponse
  | GenesisTransactionResponse
  | BlockMetadataTransactionResponse
  | StateCheckpointTransactionResponse;
```

commenting as a reminder of our earlier conversation, not necessary but may make things a little clearer

xbtmatt (approved):
Sooo much better than the generated stuff, lgtm

davidiw (approved):
feels like these should be potentially bound to serializer / deserializer / init functions? is this a first step of just shedding the ugly open api code?

0xmaayan:
> feels like these should be potentially bound to serializer / deserializer / init functions? is this a first step of just shedding the ugly open api code?

yep. it is to get rid of the generated types and have everything defined locally.

PR #9744, opened 2023-08-24
Delegate all the methods from sub classes

0xmaayan:
### Description

EDIT: following PR comments we will roll with this idea.

This is an idea for how the end user would use the new sdk.

Currently, one initializes `Aptos`. i.e `const aptos = new Aptos()`, and then has access to all the sub classes, i.e `account` and for all future sub classes.
```
const aptos = new Aptos();
aptos.account.getInfo("0x1")
```

This PR proposes to delegate all sub classes methods to the Aptos class, so user would not need to use the namespace but simply call any sub class function
```
const aptos = new Aptos();
aptos.getInfo("0x1")
```

image

This way, sdk users can find the query they are looking for easily without trying to figure out what namespace a function lives in, and sdk maintainers can keep functions in different files and under different namespaces which helps with maintenance and development process.

note: it would not work for static methods, which is expected.

### DIFF
--- a/ecosystem/typescript/sdk_v2/src/api/account.ts
+++ b/ecosystem/typescript/sdk_v2/src/api/account.ts
@@ -33,7 +33,7 @@ export class Account {
    * }
    * ```
    */
-  async getInfo(args: { accountAddress: HexInput }): Promise<AccountData> {
+  async getAccountInfo(args: { accountAddress: HexInput }): Promise<AccountData> {
     const data = await getInfo({ aptosConfig: this.config, ...args });
     return data;
   }
@@ -48,7 +48,7 @@ export class Account {
    * @returns Account modules
    */
 
-  async getModules(args: {
+  async getAccountModules(args: {
     accountAddress: HexInput;
     options?: PaginationArgs & LedgerVersion;
   }): Promise<MoveModuleBytecode[]> {
@@ -72,7 +72,7 @@ export class Account {
    * }
    * ```
    */
-  async getModule(args: {
+  async getAccountModule(args: {
     accountAddress: HexInput;
     moduleName: string;
     options?: LedgerVersion;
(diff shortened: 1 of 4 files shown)


0xjinn:
I think this is a nice feature, but feeling a bit overwhelm when API are under `aptos` namespace, and it might be confusing for new develoeprs.

e.g `aptos.getInfo()` - Users might not know if this is getting account info, config info, or other info. Unless we keep keep the function name consistent with the class name included, -> getAccountInfo().

hariria:
Looks great!

0xmaayan:
> I think this is a nice feature, but feeling a bit overwhelm when API are under `aptos` namespace, and it might be confusing for new develoeprs.
> 
> e.g `aptos.getInfo()` - Users might not know if this is getting account info, config info, or other info. Unless we keep keep the function name consistent with the class name included, -> getAccountInfo().

yes, as mentioned in the PR description, function names can be changed to a more meaningful names

xbtmatt:
Thought about it a bit more, I suppose it's not really a big deal as long as there's a clear list of subclasses somewhere. It is definitely nice to have the mixins sometimes

xbtmatt (approved):
Thought about it a bit more, I suppose it's not really a big deal since the subclasses are right there in the definition and if you really want to narrow it down to just subclasses you can look in the class definition.

repeating `aptos.account` everywhere makes it difficult to read, so it is definitely nice to have the mixins

0xmaayan:
> Thought about it a bit more, I suppose it's not really a big deal since the subclasses are right there in the definition and if you really want to narrow it down to just subclasses you can look in the class definition.
> 
> repeating `aptos.account` everywhere makes it difficult to read, so it is definitely nice to have the mixins

sounds good! I'll keep this PR around if anyone has more thoughts. meanwhile, labeled it as "do not merge" and can get back to it when I am back.

davidiw (commented):
oh the beauty of dynamic languages....

davidiw (approved):

davidiw:
note, you don't demonstrate this works in any examples, might be good to update the examples as such....

PR #9745, opened 2023-08-24
[Framework] AIP-45: Allow soft burning objects

movekevin:
### Description
Context: https://github.com/aptos-foundation/AIPs/blob/main/aips/aip-45.md

### Test Plan
Unit tests. Manual tests in devnet/testnet/mainnet later

### DIFF
--- a/aptos-move/framework/aptos-framework/sources/object.move
+++ b/aptos-move/framework/aptos-framework/sources/object.move
@@ -45,6 +45,8 @@ module aptos_framework::object {
     const EMAXIMUM_NESTING: u64 = 6;
     /// The resource is not stored at the specified address.
     const ERESOURCE_DOES_NOT_EXIST: u64 = 7;
+    /// Cannot reclaim objects that weren't burnt.
+    const EOBJECT_NOT_BURNT: u64 = 8;
 
     /// Explicitly separate the GUID space between Object and Account to prevent accidental overlap.
     const INIT_GUID_CREATION_NUM: u64 = 0x4000000000000;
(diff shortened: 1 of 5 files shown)


davidiw (approved):
you're such a perfectionist.....

bowenyang007:
What are we doing for secondary stores?

bowenyang007 (approved):

aptos-move/framework/aptos-framework/sources/object.move
```diff
@@ -84,6 +86,9 @@ module aptos_framework::object {
     /// derivation to produce an object address.
     const OBJECT_FROM_SEED_ADDRESS_SCHEME: u8 = 0xFE;
 
+    /// Address where unwanted objects can be forcefully transferred to.
+    const BURN_ADDRESS: address = @0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff;
```
lightmark, line 90: I wonder how we choose between 0x0 and 0xffffffffff.... 0 is not that error prone to me.

lightmark (commented):
do we want a short security review for this?

aptos-move/framework/aptos-framework/sources/primary_fungible_store.move
```diff
@@ -126,6 +126,8 @@ module aptos_framework::primary_fungible_store {
     /// Withdraw `amount` of fungible asset from the given account's primary store.
     public fun withdraw<T: key>(owner: &signer, metadata: Object<T>, amount: u64): FungibleAsset {
         let store = primary_store(signer::address_of(owner), metadata);
+        // Check if the store object has been burnt or not. If so, unburn it first.
+        may_be_unburn(owner, store);
```
lightmark, line 130: Why is this necessary? it seems fine to error out in this case.

movekevin: To make it seem more seamless to transition between burn and unburn

aptos-move/framework/aptos-framework/sources/object.move
xbtmatt: Also, does it matter if this is a reserved framework address or not?

Issue #9717, opened 2023-08-22
[Bug] UNKNOWN_INVARIANT_VIOLATION_ERROR when using inline function to loop vector  with break


fgfm999:
# 🐛 Bug

Got `UNKNOWN_INVARIANT_VIOLATION_ERROR ` with codes below

`guess_flips` will failed at `let _v2 =  flips;`  with `UNKNOWN_INVARIANT_VIOLATION_ERROR`.

The other three functions are all passed, the logic is the same, just tried to locate this issue as accurately as possible.

## To reproduce

```rust
module 0xcafe::vectors {
    use std::vector;
    public entry fun guess_flips(flips: vector<u8>) {
        loops_with_break(&flips);
        let _v = copy flips; // this is ok
        
        // this will fail, a UNKNOWN_INVARIANT_VIOLATION_ERROR (code 2000)
        let _v2 =  flips;
    }
    
    // no function call
    public entry fun guess_flips_directly(flips: vector<u8>) {
        let i = 0;
        while (i < vector::length(&flips)) {
            if (*vector::borrow(&flips, i) != 0) {
                break
            };
            i = i + 1;
        };
        let _v = copy flips; // this is ok
        let _v2 =  flips; // this is ok
    }

    // call function, no inline, with `break`
    public entry fun guess_with_break_without_inline(flips: vector<u8>) {
        loops_with_break_on_inline(&flips);
        let _v = copy flips; // this is ok
        let _v2 =  flips; // this is ok
    }

    // call `inline` function, without `break`
    public entry fun guess_without_break_with_inline(flips: vector<u8>) {
        loops_without_break(&flips);
        let _v = flips; // this is ok
        let _v2 = copy flips; // this is ok
    }

    
    inline fun loops_with_break(flips: &vector<u8>) {
        let i = 0;
        while (i < vector::length(flips)) {
            if (*vector::borrow(flips, i) != 0) {
                break 
            };
            i = i + 1;
        };
    }

    fun loops_with_break_on_inline(flips: &vector<u8>) {
        let i = 0;
        while (i < vector::length(flips)) {
            if (*vector::borrow(flips, i) != 0) {
                break
            };
            i = i + 1;
        };
    }
    
    inline fun loops_without_break(flips: &vector<u8>) {
        let i = 0;
        while (i < vector::length(flips)) {
            assert!(*vector::borrow(flips, i) == 0, 3);
            i = i + 1;
        };
    }

    #[test]
[...]

Issue #6615, opened 2023-02-15
[Bug] inline function


lightmark:
🐛 Bug
When implement inline fun all<K, V> to SmartTable and run test_all.
The following error shows up:

2023-02-15T18:39:37.467108Z ERROR {"message":"Error: VMError { major_status: UNKNOWN_INVARIANT_VIOLATION_ERROR, sub_status: None, message: Some("moving container with dangling references"), exec_state: Some(ExecutionState { stack_trace: [(Some(ModuleId { address: 0000000000000000000000000000000000000000000000000000000000000001, name: Identifier("smart_table") }), FunctionDefinitionIndex(7), 12), (Some(ModuleId { address: 0000000000000000000000000000000000000000000000000000000000000001, name: Identifier("smart_table") }), FunctionDefinitionIndex(19), 69)] }), location: Module(ModuleId { address: 0000000000000000000000000000000000000000000000000000000000000001, name: Identifier("table_with_length") }), indices: [], offsets: [(FunctionDefinitionIndex(10), 13)] }\nCORE DUMP: >>>>>>>>>>>>\nCall stack:\n frame #0: 0x0000000000000000000000000000000000000000000000000000000000000001::smart_table::test_all [pc = 69]\n frame #1: 0x0000000000000000000000000000000000000000000000000000000000000001::smart_table::destroy [pc = 12]\nframe #2: 0x0000000000000000000000000000000000000000000000000000000000000001::table_with_length::remove [pc = 13]:\n0> CopyLoc(0)\n1> MutBorrowFieldGeneric(FieldInstantiationIndex(0))\n2> MoveLoc(1)\n3> CallGeneric(7)\n4> StLoc(2)\n5> CopyLoc(0)\n6> ImmBorrowFieldGeneric(FieldInstantiationIndex(1))\n7> ReadRef\n8> LdU64(1)\n9> Sub\n10> MoveLoc(0)\n11> MutBorrowFieldGeneric(FieldInstantiationIndex(1))\n12> WriteRef\n13 MoveLoc(2)\nLocals (120f35e08):\n[0] Invalid\n[1] Invalid\n[2] (container 120f37878: [(container 120f36738: [U64(13618872872554492724), U64(47), U64(47)])])\nOperand Stack:\n\n<<<<<<<<<<<<\n"}

To reproduce
git checkout smart_table_debug
cd aptos-core/aptos-move/move-examples/data_structures
aptos move test
[...]

PR #9752, opened 2023-08-24
[smart_table] a lot of inline functions added

lightmark:
### Description
see the implementations. all is missing here because it triggers the error mentioned in
https://github.com/aptos-labs/aptos-core/issues/9717
https://github.com/aptos-labs/aptos-core/issues/6615
again.
cc @wrwg @brmataptos

### Test Plan
ut

### DIFF
--- a/aptos-move/framework/aptos-stdlib/sources/data_structures/smart_vector.move
+++ b/aptos-move/framework/aptos-stdlib/sources/data_structures/smart_vector.move
@@ -14,8 +14,10 @@ module aptos_std::smart_vector {
     const EVECTOR_EMPTY: u64 = 3;
     /// bucket_size cannot be 0
     const EZERO_BUCKET_SIZE: u64 = 4;
+    /// The length of the smart vectors are not equal.
+    const ESMART_VECTORS_LENGTH_MISMATCH: u64 = 0x20005;
 
-    /// A Scalable vector implementation based on tables, elements are grouped into buckets with `bucket_size`.
+    /// A Scalable vector implementation based on tables, Ts are grouped into buckets with `bucket_size`.
     /// The option wrapping BigVector saves space in the metadata associated with BigVector when smart_vector is
     /// so small that inline_vec vector can hold all the data.
     struct SmartVector<T> has store {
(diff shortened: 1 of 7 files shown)


aptos-move/framework/aptos-stdlib/sources/data_structures/smart_vector.move
```diff
@@ -336,11 +336,124 @@ module aptos_std::smart_vector {
...
+        v: SmartVector<T>,
+        p: |&T|bool
+    ): SmartVector<T> {
+        let result = aptos_std::smart_vector::new<T>();
+        aptos_std::smart_vector::for_each(v, |elem| {
+            if (p(&elem)) aptos_std::smart_vector::push_back(&mut result, elem);
+        });
+        result
+    }
+
```
movekevin, line 456: What about zip, zip_ref, etc. too :)

aptos-move/framework/aptos-stdlib/sources/data_structures/smart_vector.move
```diff
@@ -336,11 +336,124 @@ module aptos_std::smart_vector {
...
+        v: SmartVector<T1>,
+        f: |T1|T2
+    ): SmartVector<T2> {
+        let result = aptos_std::smart_vector::new<T2>();
+        aptos_std::smart_vector::for_each(v, |elem| push_back(&mut result, f(elem)));
+        result
+    }
+
+    /// Filter the vector using the boolean function, removing all Ts for which `p(e)` is not true.
+    public inline fun filter<T: store + drop>(
```
movekevin, line 446: What about find that accepts a lambda (instead of value)? vector needs one such function too :P

movekevin (approved):

davidiw (approved):

aptos-move/framework/aptos-stdlib/sources/data_structures/smart_vector.move
lightmark: vector already has that. I 'll add one.

lightmark: find doesn't work cuz it will trigger the bug  https://github.com/aptos-labs/aptos-core/issues/9717
https://github.com/aptos-labs/aptos-core/issues/6615 again.

PR #9754, opened 2023-08-24
[verifier] fix counting bug in signature checker v2

vgao1996:
This fixes a panicking bug in the signature checker caused by `max_num_of_ty_params_or_args` not counting the type parameters of a script. This was originally implemented as part of https://github.com/aptos-labs/aptos-core/pull/9344 but it has been split out since that PR needs some re-work.


third_party/move/move-bytecode-verifier/src/signature_v2.rs
```diff
@@ -992,8 +993,11 @@ pub fn verify_module(module: &CompiledModule) -> VMResult<()> {
     res.map_err(|e| e.finish(Location::Module(module.self_id())))
 }
 
-pub fn verify_script(script: &CompiledScript) -> VMResult<()> {
-    let max_num = max_num_of_ty_params_or_args(BinaryIndexedView::Script(script));
+pub fn verify_script(config: &VerifierConfig, script: &CompiledScript) -> VMResult<()> {
+    let mut max_num = max_num_of_ty_params_or_args(BinaryIndexedView::Script(script));
+    if config.sig_checker_v2_fix_script_ty_param_count {
+        max_num = max_num.max(script.type_parameters.len());
+    }
```
davidiw, line 1000:
```suggestion
    let max_num = if config.sig_checker_v2_fix_script_ty_param_count {
        max_num = max_num.max(script.type_parameters.len())
    } else {
        max_num_of_ty_params_or_args(BinaryIndexedView::Script(script))
    }
```

vgao1996: I kinda get what you're trying to change, but new logic you proposed doesn't seem quite right. Do you actually mean
```
let max_num = if config.sig_checker_v2_fix_script_ty_param_count {
        usize::max(max_num_of_ty_params_or_args(BinaryIndexedView::Script(script)), script.type_parameters.len())
    } else {
        max_num_of_ty_params_or_args(BinaryIndexedView::Script(script))

```
?

wrwg (approved):

runtian-zhou (approved):

PR #9581, opened 2023-08-09
[move] extra randomness APIs + raffle and lottery examples

alinush:
### Description

1. Updates the randomness API to include more functions.
2. Gives two example of randapps: a raffle and a lottery.
    - The lottery example contains both a secure variant and an insecure one (susceptible to undergasing)

PR #9759, opened 2023-08-24
Publish `veiled_coin` and `drand` example modules under resource accounts

mstraka100:
### Description

This PR modifies the VeiledCoin and Drand Move example contracts to each be deployed under a resource account. Previously in both examples `init_module` would be deployed under a user account which kept signing access to the resource account. This has been changed so that the contract must be deployed under a resource account, whose authentication key is rotated in `init_module` so that the deploying user account loses signing access to it.

### Test Plan
The VeiledCoin and Drand unit tests have integrated the above change and can be run with `RUST_MIN_STACK=104857600 cargo test -- veiled_coin` and `RUST_MIN_STACK=104857600 cargo test -- drand` respectively in `aptos-core/aptos-move/move-examples`.


aptos-move/move-examples/tests/move_unit_tests.rs
```diff
@@ -117,7 +118,12 @@ fn test_hello_blockchain() {
 
 #[test]
 fn test_drand_lottery() {
-    test_common("drand");
+let addr = AccountAddress::from_hex_literal("0x7654").unwrap();
```
alinush, line 121: indent

aptos-move/move-examples/drand/Move.toml
```diff
@@ -10,4 +10,5 @@ AptosFramework = { local = "../../framework/aptos-framework" }
 std = "0x1"
 aptos_std = "0x1"
 aptos_framework = "0x1"
+source_addr = "_"
```
alinush, line 13: Would it be better to rename this to `developer_address`. I find "developer" more informative than "source" and in also in line with [the lottery PR](https://github.com/aptos-labs/aptos-core/pull/9581/files#diff-6f8716d830670b98d3861da5eec4442bfd3e2f6db3b35e340bb63c146ac05729R14))

aptos-move/move-examples/veiled_coin/Move.toml
```diff
@@ -10,3 +10,4 @@ std = "0x1"
 aptos_std = "0x1"
 aptos_framework = "0x1"
 veiled_coin = "_"
+source_addr = "_"
```
alinush, line 13: Same suggestion of using `developer_address`.

aptos-move/move-examples/veiled_coin/sources/veiled_coin.move
```diff
@@ -100,6 +100,14 @@
 /// the resource account can be used to transfer out the normal from its coin store. Transfering out a coin like this
 /// requires a `signer` for the resource account, which the `veiled_coin` module can obtain via a `SignerCapability`.
 ///
+/// ## Deploying this contract
```
alinush, line 103: Sweet!! :D

aptos-move/move-examples/veiled_coin/sources/veiled_coin.move
```diff
@@ -248,15 +258,16 @@ module veiled_coin::veiled_coin {
...
+        let signer_cap = resource_account::retrieve_resource_account_cap(
+            resource_account, @source_addr
+        );
 
-        move_to(deployer,
-            VeiledCoinMinter {
-                signer_cap
-            }
-        )
+        coin::register<CoinType>(resource_account);
```
alinush, line 267: Why must this be done here? There was code in `veiled_coin` that handled this before-hand.

Plus, the veiled coin module handles *any* type of coins, not just `CoinType`.

aptos-move/move-examples/veiled_coin/sources/veiled_coin.move
```diff
@@ -237,7 +247,7 @@ module veiled_coin::veiled_coin {
 
     /// Initializes a so-called "resource" account which will maintain a `coin::CoinStore<T>` resource for all `Coin<T>`'s
     /// that have been converted into a `VeiledCoin<T>`.
-    fun init_module(deployer: &signer) {
+    fun init_module<CoinType>(resource_account: &signer) {
```
alinush, line 250: I don't think `init_module` can be generic. Plus, it's not in line with the functionality we want: we want the `veiled_coin` module to be deployed **once** for **all** coin types.

aptos-move/move-examples/veiled_coin/sources/veiled_coin.move
```diff
@@ -763,9 +774,14 @@ module veiled_coin::veiled_coin {
...
 
     #[test_only]
-    /// So we can call this from `veiled_coin_tests.move`.
-    public fun init_module_for_testing(deployer: &signer) {
-        init_module(deployer)
+    /// So we can call this from `veiled_coin_tests.move`. `source` must be the signer of @source_addr
+    public fun init_module_for_testing<CoinType>(source: &signer, resource_acct: &signer) {
+        let seed = vector::empty();
+        let auth_key = vector::empty();
+        // When deploying this contract we must have both a source account which deploys the contract, and a resource account under which the contract is published. These two lines mock this flow for testing
```
alinush, line 781: Sweet! 😁 Again, consider replacing `source` with `developer`.

aptos-move/move-examples/veiled_coin/sources/veiled_coin.move
alinush: I believe the code that handles this should be in `veiled_mint_from_coin`. If it needs to be in other places, I would add `coin::register` there, not in `init_module`.

aptos-move/move-examples/veiled_coin/sources/veiled_coin.move
mstraka100: Done

aptos-move/move-examples/veiled_coin/Move.toml
mstraka100: Replaced with `developer_addr`

aptos-move/move-examples/drand/Move.toml
mstraka100: Replaced with `developer_addr`

aptos-move/move-examples/tests/move_unit_tests.rs
mstraka100: Done

aptos-move/move-examples/veiled_coin/sources/veiled_coin.move
mstraka100: Correct, this was a mistake. Removed the register line from `init_module`

aptos-move/move-examples/veiled_coin/sources/veiled_coin.move
mstraka100: Correct. Changed it back to be unparameterized

alinush (approved):

zjma (approved):

aptos-move/move-examples/drand/sources/lottery.move
```diff
@@ -129,7 +143,7 @@ module drand::lottery {
     /// Allows anyone to close the lottery (if enough time has elapsed) and to decide the winner, by uploading
     /// the correct _drand-signed bytes_ associated with the committed draw time in `Lottery::draw_at`.
     /// These bytes will then be verified and used to extract randomness.
-    public entry fun close_lottery(drand_signed_bytes: vector<u8>): Option<address> acquires Lottery {
+    public entry fun close_lottery(drand_signed_bytes: vector<u8>) acquires Lottery {
```
alinush, line 146: Yep... Ideally, if you can, consider having the `public entry close_lottery(...)` function call a `public close_lottery_internal(...):  Option<address>` function, where everything is implemented, and ignore its return value.

This would allow other contracts to call `drand::lottery::close_lottery_internal` and check who won. Right now, there is no easy way of doing this.

davidiw:
Considering that we have multi-user accounts, I'm curious what we're hoping to teach user's with this change?

alinush:
> Considering that we have multi-user accounts, I'm curious what we're hoping to teach user's with this change?

I'm not sure why multi-user (multisig v2?) accounts are relevant here. Are you suggesting a contract like `veiled_coin` should be deployed in a multisig v2 account?

What we're hoping people learn here is how to deploy the contract at a resource account that nobody controls. This seems important for contracts like `veiled_coin`, where an attacker could steal everybody's coins if he upgrades the contract.

However, if there's a better flow you have in mind, let's discuss!

davidiw:
> > Considering that we have multi-user accounts, I'm curious what we're hoping to teach user's with this change?
> 
> I'm not sure why multi-user (multisig v2?) accounts are relevant here. Are you suggesting a contract like `veiled_coin` should be deployed in a multisig v2 account?
> 
> What we're hoping people learn here is how to deploy the contract at a resource account that nobody controls. This seems important for contracts like `veiled_coin`, where an attacker could steal everybody's coins if he upgrades the contract.
> 
> However, if there's a better flow you have in mind, let's discuss!

Ahh, if that's the goal, then I would recommend setting the upgrade policy to immutable, but that's where I think it is important to decouple the goal of the code from the deployment considerations for the code.

* Immutable contract -- makes it clear who deployed it and its lifetime
* Resource account -- decouples ownership but does not prevent malicious upgrades in and of itself
* Multi-user account -- so long a quorum of owners agree to perform an operation it is viable

if you think you write bug free software or distrust of admins is greater than correctness, immutable contract.

otherwise leveraging multi-user with a high expectation for upgrades should suffice.

Most defi apps leverage a multi-ed or multi-user framework so that they can manage upgrades.

I think there are interesting aspects to tradeoff between the two here.

In veiled, what if there is a bug and coins get stuck somewhere? Could an upgrade fix it?

In drand, if there's a bug, we can just publish a new contract and have existing contracts point to that... but what if there's an immutable contract? Well I guess it would need an immutable drand :grin: ... In that case, I would say a drand router package that can be upgraded should point to immutable instances of versioned drand contracts.

mstraka100:
@davidiw Is it your opinion that it is generally a bad practice to deploy contracts to resource accounts, and that it is better to deploy a contract to a user account with an immutable upgrade policy? What do you mean that deploying to a resource account "does not prevent malicious upgrades in and of itself"? Particularly if the authentication key for the resource account is rotated to 0x0?

alinush:
As far as I can tell, we need better documentation on the many flows that can be used to deploy a smart contract and what is the recommended one / tradeoffs. Both @mstraka100 and I are confused about this. 

@davidiw, your write-up below is a good starting point, but we'd need more details & time to understand. If this documentation already exists, please point us to it. We couldn't find it (e.g., searching "immutable" on aptos.dev only yielded [this discussion](https://aptos.dev/guides/nfts/mint-nft-cli/#3-add-admin-account-and-functions)). Maybe we can help write it too. 

For now, we'll go ahead and merge this with the `aptos move create-resource-account-and-publish-package` flow, which AFAWK, rotates the authkey to zero and ensures nobody can control the resource account where the contract is deployed. Advanced users can customize this, if they want.

davidiw (changes requested):
let's discuss how to eliminate resource accounts and only demonstrate them in examples where they add value.

davidiw:
@movekevin , I think for examples, we really ought not be to clever with deployment strategies and just deploy on the deployers address. Then we can have advanced deployment methods as a single set of examples that can be combined. What are your thoughts?

PR #9767, opened 2023-08-24
[Mempool] reduce mempool log noise

bchocho:
### Description

Reduce the noise added by #9309. Make expected "errors" trace, and sample regardless.

### DIFF
--- a/mempool/src/shared_mempool/tasks.rs
+++ b/mempool/src/shared_mempool/tasks.rs
@@ -69,8 +69,17 @@ pub(crate) async fn execute_broadcast<NetworkClient, TransactionValidator>(
                 )
                 .peer(&peer)
                 .error(&error)),
+                BroadcastError::NoTransactions(_) | BroadcastError::PeerNotPrioritized(_, _) => {
+                    sample!(
+                        SampleRate::Duration(Duration::from_secs(60)),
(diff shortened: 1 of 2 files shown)


mempool/src/shared_mempool/tasks.rs
```diff
@@ -69,8 +69,17 @@ pub(crate) async fn execute_broadcast<NetworkClient, TransactionValidator>(
                 )
                 .peer(&peer)
                 .error(&error)),
+                BroadcastError::NoTransactions(_) | BroadcastError::PeerNotPrioritized(_, _) => {
+                    sample!(
+                        SampleRate::Duration(Duration::from_secs(60)),
```
JoshLind, line 74: nit: you could consider making this a const, e.g.:

```
const BROADCAST_LOG_FREQ_SECS: u64 = 60;
```

JoshLind (approved):
LGTM!

zekun000 (approved):

mempool/src/shared_mempool/tasks.rs
ibalajiarun: Do we need to sample trace? Just sampling overhead for something that we don't use quite often.

ibalajiarun (approved):

PR #9768, opened 2023-08-24
[Executor-benchmark] Fix a bug in measuring 'vm_start_time' when '--generate-then-execute' flag is used

manudhundi:
[Executor-benchmark] Fix a bug in measuring 'vm_start_time' when '--generate-then-execute' flag is used

### DIFF
--- a/execution/executor-benchmark/src/lib.rs
+++ b/execution/executor-benchmark/src/lib.rs
@@ -241,7 +241,7 @@ pub fn run_benchmark<V>(
         .collect::<HashMap<_, _>>();
     let start_commit_total = APTOS_EXECUTOR_COMMIT_BLOCKS_SECONDS.get_sample_sum();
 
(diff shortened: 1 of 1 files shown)


sitalkedia (approved):

execution/executor-benchmark/src/lib.rs
```diff
@@ -259,6 +259,7 @@ pub fn run_benchmark<V>(
         );
     }
     if pipeline_config.delay_execution_start {
+        start_vm_time = APTOS_EXECUTOR_VM_EXECUTE_BLOCK_SECONDS.get_sample_sum();
```
zjma, line 262: but it looks like `APTOS_EXECUTOR_VM_EXECUTE_BLOCK_SECONDS` the counter is not supposed to change between L244 and L262?

msmouse (approved):

PR #9747, opened 2023-08-24
[TS SDK] Add support for fungible asset queries

0xmaayan:
### Description
Adds fungible asset queries support
- getAccountCoinsData
- getAccountCoinsDataCount
### Test Plan
tests are passing

PR #9773, opened 2023-08-25
[TS SDK] Add support to filter on coin data query

0xmaayan:
### Description
This PR adds support for the user to filter on the coin data query as some products might want to exclude secondary stores data until better ecosystem support.
https://github.com/aptos-labs/aptos-core/pull/9747#discussion_r1304956904

### DIFF
--- a/ecosystem/typescript/sdk/src/providers/indexer.ts
+++ b/ecosystem/typescript/sdk/src/providers/indexer.ts
@@ -52,6 +52,7 @@ import {
   Account_Transactions_Order_By,
   Current_Collections_V2_Order_By,
   Current_Collection_Ownership_V2_View_Order_By,
+  Current_Fungible_Asset_Balances_Bool_Exp,
   Current_Fungible_Asset_Balances_Order_By,
   Current_Token_Datas_V2_Order_By,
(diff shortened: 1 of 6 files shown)


ecosystem/typescript/sdk/src/providers/indexer.ts
```diff
@@ -830,19 +831,17 @@ export class IndexerClient {
     extraArgs?: {
       options?: IndexerPaginationArgs;
       orderBy?: IndexerSortBy<Current_Fungible_Asset_Balances_Order_By>[];
+      filter?: [Current_Fungible_Asset_Balances_Bool_Exp];
```
movekevin, line 834: I think we need to default to filtering out secondary stores. Otherwise the fungible_asset_client also needs to be rewritten to allow transferring/checking balances from secondary stores. Currently no operations regarding secondary stores are supported. This means developers can run into a very weird experience here:
1. They call check getAccountCoinsData, which returns secondary stores
2. They sum up the balances and display to the users which now show balances > what they have in primary stores
3. They try to send > what they have in primary store => Errors
4. Everyone is now confused. Users are confused that they have less money than what wallets show them. Developers are confused why things error out

PR #9785, opened 2023-08-25
[object] refactor burn/unburn

lightmark:
### Description
shorten our code.

### Test Plan
ut

### DIFF
--- a/aptos-move/framework/aptos-framework/sources/object.move
+++ b/aptos-move/framework/aptos-framework/sources/object.move
@@ -467,21 +467,22 @@ module aptos_framework::object {
     ) acquires ObjectCore {
         let owner_address = signer::address_of(owner);
         verify_ungated_and_descendant(owner_address, object);
+        transfer_raw_inner(object, to);
+    }
 
+    inline fun transfer_raw_inner(object: address, to: address) acquires ObjectCore {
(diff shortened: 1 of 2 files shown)


aptos-move/framework/aptos-framework/sources/object.move
```diff
@@ -467,21 +467,22 @@ module aptos_framework::object {
...
+        if (object_core.owner != to) {
+            event::emit_event(
+                &mut object_core.transfer_events,
+                TransferEvent {
+                    object,
+                    from: object_core.owner,
+                    to,
+                },
+            );
+            object_core.owner = to;
```
davidiw, line 484: please, please no..... now everyone that is not has to read to the bottom lol, I like the rest of the PR but why this....

davidiw (approved):

aptos-move/framework/aptos-framework/sources/object.move
movekevin: Yeah less indentation previously is better

movekevin (approved):

aptos-move/framework/aptos-framework/sources/object.move
lightmark: because inline function does not support `return`...

PR #9787, opened 2023-08-25
add ser/de to CompiledModule focused fuzzers

gedigi:
### Description
- add ser/de to CompiledModule focused fuzzers
- add repro script

### Test Plan
- `test-fuzzers.sh` passes as expected
- `repro.sh` works as expected


davidiw (commented):
why does it see it as deadcode?

wrwg (approved):

gedigi:
> > why does it see it as deadcode?
> 
> I think this can be prevented by making this `cfg(test)`

it actually doesn't. I tried to gate it with `cfg(test)` but it still shows as dead code. maybe it's because the fuzz tests themselves have no main and are not directly linked in the workspace?

davidiw (approved):
I see, this is a deficiency in the way the linter picks up fuzzing targets.... (maybe)

PR #9788, opened 2023-08-26
Merge indexer docs

banool:
This PR rebasees the indexer-docs feature branch into main. Each commit here came from a separate PR. Each commit is independent, let's rebase this rather than merge.


rtso (approved):

developer-docs-site/sidebars.js
```diff
@@ -359,6 +358,51 @@ const sidebars = {
         "tutorials/build-e2e-dapp/handle-tasks",
       ],
     },
+    {
+      type: "category",
+      label: "Learn about the Aptos Indexer",
```
davidiw, line 363: for now, I'd probably still put this under advanced guides or integrate with aptos

agreed it probably belongs under infrastructure and nodes becomes more general...

davidiw (commented):
before we land this, can we please setup redirects for recently moved content?

banool:
Good call, that was on the todolist and subsequently forgotten 🏃‍♀️

developer-docs-site/sidebars.js
banool: Imo that's buried a little too deep. Anyway it doesn't really matter if we're gonna move it again in a few weeks.

banool:
(I've gotta add another redirect and move the content again, this isn't ready to go)

banool:
Okay we're good to go!

bowenyang007 (approved):

PR #9789, opened 2023-08-26
[Execution][Storage] Split the ledger update to a separate phase

sitalkedia:
### Description

Today the ledger update being sequential and in the critical path of execution consumes around ~15% of the execution. This change introduces a new phase for ledger update. Please note that the change only affects the executor benchmark, and current production path should not be affected. We will have a separate change to add a new pipeline stage for this. 

### Test Plan

Ran the benchmark and observed ~10% TPS gain in sharded execution.

### DIFF
--- a/execution/executor/src/metrics.rs
+++ b/execution/executor/src/metrics.rs
@@ -79,6 +79,29 @@ pub static APTOS_EXECUTOR_EXECUTE_BLOCK_SECONDS: Lazy<Histogram> = Lazy::new(|| 
     .unwrap()
 });
 
+pub static APTOS_EXECUTOR_LEDGER_UPDATE_SECONDS: Lazy<Histogram> = Lazy::new(|| {
+    register_histogram!(
+        // metric name
+        "aptos_executor_ledger_update_seconds",
+        // metric description
+        "The total time spent in ledger update in the block executor.",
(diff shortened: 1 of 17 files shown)


execution/executor/src/metrics.rs
```diff
@@ -79,6 +79,29 @@ pub static APTOS_EXECUTOR_EXECUTE_BLOCK_SECONDS: Lazy<Histogram> = Lazy::new(||
...
+        // metric name
+        "aptos_executor_ledger_update_seconds",
+        // metric description
+        "The total time spent in ledger update in the block executor.",
+        exponential_buckets(/*start=*/ 1e-3, /*factor=*/ 2.0, /*count=*/ 20).unwrap(),
+    )
+    .unwrap()
+});
+
+pub static APTOS_EXECUTOR_LEDGER_UPDATE_OTHER_SECONDS: Lazy<HistogramVec> = Lazy::new(|| {
```
msmouse, line 93: what's wrong with just using `APTOS_EXECUTOR_OTHER_TIMERS_SECONDS`?

msmouse (approved):
LGTM.

execution/executor/src/metrics.rs
sitalkedia: Nothing wrong, I thought its just better to have a separate counter for breaking down the ledger update phase but change it back if strong opinion about it.

grao1991 (approved):

PR #9792, opened 2023-08-27
[compiler v2][type checker] Make arithmetic operators generic with constraints

wrwg:
### Description

Until now, we had represented arithmetic operators via overloads. For instance, we had `fun _+_(u8,u8):u8, fun _+_(u16,u16):u16` and so on for all integer types. The problem with this approach is that it does not work for constants, as described in #9330. Move allows to write `1+2` and this expression behaves from the typing perspective like `3`, which has a type which is automatically adapted to the usage context.

In this PR, arithmetic ops are represented not longer as overloads but as _generic functions with constraints_. For this, the already existing `SomeNumber(..)` type constraint is used. Basically:

```
fun _+_<A>(x: A, y: A): A where A is SomeNumber({u8,u16,u32,..,u256});
fun _-_<A>(...)...
```

This has the following effects:

- Expressions over constants inherit their generic result type (that is, `1 + 2` does not fix the operator to a specific integer type. Instead the context may do this.)
- Error messages get better. Before we just printed non-matching overloads, now we see more precise error messages.

Technically, this needed some refinements of the type constraint system, which brings us even closer to a type checker supporting traits/type classes. Also, some error reporting has been fixed which was before only working in very specific contexts.

### Test Plan

Updates baselines files, adds a new test for constant expression type inference


third_party/move/move-model/src/builder/builtins.rs
```diff
@@ -89,49 +97,241 @@ pub(crate) fn declare_builtins(trans: &mut ModelBuilder<'_>) {
...
-        declare_bin(Implies, Operation::Implies, bool_t, bool_t, bool_t, Spec);
-        declare_bin(Iff, Operation::Iff, bool_t, bool_t, bool_t, Spec);
-        declare_bin(And, Operation::And, bool_t, bool_t, bool_t, SpecAndImpl);
-        declare_bin(Or, Operation::Or, bool_t, bool_t, bool_t, SpecAndImpl);
+        let declare_arithm_ops = |trans: &mut ModelBuilder,
+                                  type_params: &[TypeParameter],
+                                  type_constraints: &BTreeMap<usize, Constraint>,
+                                  ty: Type,
+                                  visibility: EntryVisibility| {
+            declare_bin_gen(
```
brmataptos, line 126: Could you use a loop to call declare_bin_gen for all of these more compactly?  (You need a loop over pairs.)

third_party/move/move-model/src/builder/builtins.rs
```diff
@@ -195,6 +398,7 @@ pub(crate) fn declare_builtins(trans: &mut ModelBuilder<'_>) {
                 loc: loc.clone(),
                 oper: Operation::MaxU8,
                 type_params: vec![],
+                type_param_constraints: BTreeMap::default(),
```
brmataptos, line 401: Actually, you could group similar operations into loops over a literal tuples down here, too, and make the code much more compact and simpler to change in the future.

third_party/move/tools/move-cli/src/base/test.rs
```diff
@@ -13,14 +13,16 @@ use move_compiler::{
...
+use move_package::{
+    compilation::{build_plan::BuildPlan, compiled_package::unimplemented_v2_driver},
+    BuildConfig, CompilerConfig,
+};
 use move_unit_test::UnitTestingConfig;
 use move_vm_test_utils::gas_schedule::CostTable;
 // if unix
 #[cfg(target_family = "unix")]
 use std::os::unix::prelude::ExitStatusExt;
+// if windows
```
brmataptos, line 25: These would actually be clearer with blank lines instead of the // if unix and // if windows comments.  The #[cfg ...] are pretty self-documenting.

third_party/move/move-compiler-v2/tests/checking/typing/binary_add_invalid.exp
```diff
@@ -1,170 +1,79 @@
 
 Diagnostics:
-error: no matching declaration of `+`
+error: invalid call of `+`: expected `integer` but found `bool` for argument 1
```
brmataptos, line 3: This is much better than before, but is there some way to recognize built-ins and not refer tot hem as calls in the error message?  New users might be a little confused.

brmataptos (commented):
Looks pretty nice.

third_party/move/move-model/src/builder/builtins.rs
wrwg: Done.

third_party/move/move-model/src/builder/builtins.rs
wrwg: Lets not overdo this. Having this declarations one-by-one is also useful. Moreover, you are asking here to refactor existing code which I have not really touched as part of this PR.

third_party/move/tools/move-cli/src/base/test.rs
wrwg: That is not me. I'm just correcting a formatting mistake and bring this comment back to where it was before the last PR. My last PR accidentally messed this up, and Greg noticed and asked to fix.

third_party/move/move-compiler-v2/tests/checking/typing/binary_add_invalid.exp
wrwg: Not sure what is wrong with `+`. What else would you like to display here?

This code also did not change, the error message is only now looking like this in this context. We can fix this later if we see need.

third_party/move/move-model/src/builder/builtins.rs
brmataptos: Just a thought.

third_party/move/move-compiler-v2/tests/checking/typing/binary_add_invalid.exp
brmataptos: The problem is that Move's "+" is is an operator, not a function, so is not "called".  Perhaps "invalid use of `+`'.

But I know that it looks like a call in your IR so you may need a list of "operator" functions which would get a different error message.  So if that's difficult, probably just change the message to "invalid use of ", which kind of works for both cases.

wrwg: I do not see the problem at all. In my book an 'calling an operator' is a totally valid notion. I do not understand your concern, and disagree with that this should be renamed.

third_party/move/tools/move-cli/src/base/test.rs
brmataptos: Ok.  Though these are pretty useless comments.

third_party/move/move-compiler-v2/tests/checking/typing/bind_wrong_type.exp
```diff
@@ -42,7 +42,7 @@ error: expected `u64` but found `()`
 19 │         let (): u64 = ();
    │                       ^^
 
-error: expected `u64` but found `()`
+error: expected `()` but found `u64`
```
brmataptos, line 45: This error message seems backwards.

third_party/move/move-compiler-v2/tests/checking/typing/constant_inference.exp
```diff
@@ -0,0 +1,7 @@
+
+Diagnostics:
+error: invalid call of `+`: expected `u8` but found `u16|u32|u64|u128|u256` for argument 2
```
brmataptos, line 3: This is wrong.  The invalid type is in argument 1.

third_party/move/move-compiler-v2/tests/checking/typing/lambda.exp
```diff
@@ -6,20 +6,13 @@ error: `M::reduce` is a function and not a macro
...
-   = outruled candidate `+(u32, u32): u32` (expected `u32` but found `&integer` for argument 2)
-   = outruled candidate `+(u64, u64): u64` (expected `u64` but found `&integer` for argument 2)
-   = outruled candidate `+(u128, u128): u128` (expected `u128` but found `&integer` for argument 2)
-   = outruled candidate `+(u256, u256): u256` (expected `u256` but found `&integer` for argument 2)
 
-error: expected `()` but found `u64`
+error: expected `integer` but found `()`
    ┌─ tests/checking/typing/lambda.move:73:43
    │
 73 │         foreach(&v, |e| { sum = sum + *e; *e }) // expected to have wrong result type of lambda
```
brmataptos, line 18: Error is reversed.

third_party/move/move-model/src/ast.rs
```diff
@@ -1209,7 +1209,7 @@ impl Operation {
         )
     }
 
-    /// Whether the operation alllows to take reference parameters instead of values. This applies
+    /// Whether the operation allows to take reference parameters instead of values. This applies
```
brmataptos, line 1212: Given that you're changing this, let's make it actually grammatical.  As written, the verb "to take" is lacking a subject in the subclause.  Two options:
- Whether the operation allows reference parameters instead of values.
- Whether the operations can take reference parameters instead of values.

brmataptos (changes requested):

third_party/move/move-model/src/ty.rs
```diff
@@ -1403,22 +1474,12 @@ impl TypeUnificationError {
...
-                            .map(|p| Type::new_prim(*p).display(display_context).to_string())
-                            .join("|")
-                    };
+                Constraint::SomeNumber(_) => {
+                    let options_str = constr.display(display_context);
                     let type_str = ty.display(display_context).to_string();
                     let (expected, actual) = match order {
-                        WideningOrder::Join | WideningOrder::LeftToRight => (type_str, options_str),
-                        WideningOrder::RightToLeft => (options_str, type_str),
+                        WideningOrder::Join | WideningOrder::LeftToRight => (options_str, type_str),
```
rahxephon89, line 1481: Related to @brmataptos's comment on the reversed error message , why is the order reversed here?

third_party/move/move-compiler-v2/tests/checking/typing/binary_add_invalid.exp
brmataptos: I call an operator by dialing 0 on my phone. 

I might call an operator **function**, and if we were in a language that allows users to override operators using functions, then a programmer would understand that underlying each operator is a function, and we could say that we "call an operator" as a sort of **metonymy**.

Given that Move is not currently such a language, I would might "use" or "apply" an operator.  I don't "call an operator" in such a language.  Under the hood, yes, perhaps a function is called by the implementation, and we (as language implementors) are aware of this and can be sloppy in our language usage.  But the programmer should not have to be aware of the implementation so we should avoid distracting him/her by drawing attention to it.

This is a minor issue, and if it's nontrivial to fix, then I would understand.

wrwg: Its perhaps not that hard to fix, but it would be ugly to fix because you would suddenly special case things in error messages in an unnatural way.

I find your request unreasonable, frankly. The message is `invalid call of +` and what the heck is wrong with this? I just don't get it. It sounds totally fine to me, whether operator or not (the message does not mention operator). I would prefer real technical feedback on this PR instead of you imposing taste issues here and blocking this PR.

third_party/move/move-model/src/ast.rs
wrwg: I don't mind about such English issues, there are grammatical errors all over the place. I do not have time to fix this.

wrwg:
I attempted to fix some of the orientation errors (where it appears 'expected' is wrong wr.t. 'found'), and in course of this removed some heuristics which appeared to be broken. Those heuristics have been added in the last review round  on pressure of reviewers. Please note: _error messages are not perfect and this is not the objective at this point_. Specifically, in a contra-variance scenario (when assigning values), the expected vs found might be now a bit confusing again, but at least I hope it is correct.

third_party/move/move-compiler-v2/tests/checking/typing/binary_add_invalid.exp
brmataptos: This issue was not an urgent issue.  The type ordering issues were the blocking point.

But don't deceive yourself that this error message is clear to everyone.

brmataptos:
I filed a bug to follow up on the clarity of some of these error messages.

third_party/move/move-compiler-v2/tests/checking/typing/binary_add_invalid.exp
brmataptos: We should discuss with a few more potential developers but not waste time on this now.  Filed an issue here: https://github.com/aptos-labs/aptos-core/issues/9863

brmataptos (approved):
I'm ok with this for now, although I filed 2 followup issues.

PR #9793, opened 2023-08-27
Update typos in lib.rs

so-schen:

### DIFF
--- a/aptos-move/block-executor/src/lib.rs
+++ b/aptos-move/block-executor/src/lib.rs
@@ -48,7 +48,7 @@ dependencies. In particular, an incarnation of transaction tx_j stops and waits
 on a condition variable whenever it reads a value marked as an ESTIMATE that was
 written by a lower transaction tx_k. When the execution of tx_k finishes, it
 signals the condition variable and the execution of tx_j continues. This way,
-tx_k does not read a value that is likely to cause an abort in the future due to a
+tx_j does not read a value that is likely to cause an abort in the future due to a
 validation failure, which would happen if the next incarnation of tx_k would
 indeed write to the same location (the ESTIMATE markers that are not overwritten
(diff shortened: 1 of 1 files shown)


aptos-move/block-executor/src/lib.rs
```diff
@@ -33,7 +33,7 @@ describes the updates made by the incarnation as (memory location, value) pairs.
 The write-set of the incarnation is applied to shared memory (the multi-version
 data-structure) at the end of execution. After an incarnation executes it needs
 to pass validation. The validation re-reads the read-set and compares the
-observed versions. Intuitively, a successful validation implies that writes
+observed versions. Intuitively, a successful validation implies that reads
```
danielxiangzl, line 36: Both `reads` and `writes` are up-to-date. Here saying `writes` may be more consistent with rest of the comments.

aptos-move/block-executor/src/lib.rs
```diff
@@ -48,7 +48,7 @@ dependencies. In particular, an incarnation of transaction tx_j stops and waits
 on a condition variable whenever it reads a value marked as an ESTIMATE that was
 written by a lower transaction tx_k. When the execution of tx_k finishes, it
 signals the condition variable and the execution of tx_j continues. This way,
-tx_k does not read a value that is likely to cause an abort in the future due to a
+tx_j does not read a value that is likely to cause an abort in the future due to a
```
danielxiangzl, line 51: Good catch!

aptos-move/block-executor/src/lib.rs
davidiw, line 36:
```suggestion
observed versions. Intuitively, a successful validation implies that writes
```

davidiw (approved):

PR #9795, opened 2023-08-27
Run transfer with hotspot in executor-benchmark

zjma:
### Description

When running executor-benchmark run-transfer mode, you can now optionally specify `--hotspot-probability x` with `x` in `[0.5, 1)` to treat a fraction of `1-x` of the accounts as the "hotspot" and use it `x` of the time.

For example, with `--hotspot-probability 0.8 ... --main-signer-accounts 100`, when we need to sample an account, with probability 0.8 we sample from account 0..20 (and the remaining probability 0.2 goes to account 20..100).

NOTE:
- If `connected-tx-grps >= 1`, `--hotspot-probability` is ignored.
- if `connected-tx-grps==0` and a valid `hotspot-probability` is provided, `transactions-per-sender` is ignored.

Usage:
```
cargo run --release -p aptos-executor-benchmark -- --block-size 10000 --hotspot-probability 0.8 --split-ledger-db --use-sharded-state-merkle-db --concurrency-level 1 --async-partitioning --num-executor-shards 48 run-executor --data-dir /tmp/some-db --checkpoint-dir /tmp/some-ck --blocks 100 --main-signer-accounts 100000
```


execution/executor-benchmark/src/main.rs
```diff
@@ -189,6 +189,9 @@ struct Opt {
     #[clap(long)]
     shuffle_connected_txns: bool,
 
+    #[clap(long, default_value_t = 0.0)]
+    hotspot_probability: f32,
```
msmouse, line 193: Can be `Option<f32>`?

execution/executor-benchmark/src/main.rs
```diff
@@ -345,13 +348,20 @@ where
                 Some(mix_per_phase[0].clone())
             };
 
+            let maybe_hotspot_probability =
+                if opt.hotspot_probability >= 0.5 && opt.hotspot_probability < 1.0 {
```
msmouse, line 352: I'd rather panic when it's out of range, instead of swallowing it.

msmouse: okay I see you assert down the stack.. but anyways, easier to read if asserted here.

execution/executor-benchmark/src/transaction_generator.rs
```diff
@@ -576,6 +626,12 @@ impl TransactionGenerator {
                 connected_tx_grps,
                 shuffle_connected_txns,
             );
+        } else if maybe_hotspot_probability.is_some() {
```
msmouse, line 629: In the code somewhere assert both are not provided at the same time. (maybe as an argument group https://github.com/aptos-labs/aptos-core/blob/3e78e8a78b788f85bb84906a568e0d24242dda83/execution/executor-benchmark/src/main.rs#L168)

execution/executor-benchmark/src/transaction_generator.rs
```diff
@@ -453,6 +455,53 @@ impl TransactionGenerator {
...
+                        rand_with_hotspot(&mut rng, num_accounts, num_hotspot_accounts);
+                    let receiver = self.main_signer_accounts.as_ref().unwrap().accounts
+                        [receiver_idx]
+                        .address();
+                    let sender =
+                        &mut self.main_signer_accounts.as_mut().unwrap().accounts[sender_idx];
+                    let amount = 1;
+                    let txn = sender.sign_with_transaction_builder(
+                        self.transaction_factory.transfer(receiver, amount),
+                    );
```
msmouse, line 492: Can't reuse `get_random_transfer_batch()`?  Maybe pass in a range to indicate which accounts to use.

execution/executor-benchmark/src/lib.rs
```diff
@@ -103,6 +103,7 @@ pub fn run_benchmark<V>(
     mut transactions_per_sender: usize,
     connected_tx_grps: usize,
     shuffle_connected_txns: bool,
+    maybe_hotspot_probability: Option<f32>,
```
manudhundi, line 106: I think 'hotspot_probability' name should be good enough. Option<f32> will indicate the maybe

execution/executor-benchmark/src/transaction_generator.rs
```diff
@@ -453,6 +455,53 @@ impl TransactionGenerator {
...
+    /// Whenever we need to sample an account, with probability 0.8 we sample from {A0, A1} uniformly at random;
+    /// with probability 0.2 we sample from {A2, ..., A9} uniformly at random.
+    pub fn gen_random_transfers_with_hotspot(
+        &mut self,
+        block_size: usize,
+        num_blocks: usize,
+        hotspot_probability: f32,
+    ) {
+        assert!((0.5..1.0).contains(&hotspot_probability));
+        println!("Started generating random transfers with hotspot.");
```
sitalkedia, line 472: Remove this or change it to use logger instead.

sitalkedia (commented):
Curious what's the performance of sharded and unsharded execution with hotspot?

execution/executor-benchmark/src/transaction_generator.rs
zjma: `get_random_transfer_batch` takes a `txns_per_sender` which doesn't fit in the new block generation being proposed...

which i think is suggesting a new function.

execution/executor-benchmark/src/transaction_generator.rs
```diff
@@ -453,6 +455,53 @@ impl TransactionGenerator {
...
+                        &mut self.main_signer_accounts.as_mut().unwrap().accounts[sender_idx];
+                    let amount = 1;
+                    let txn = sender.sign_with_transaction_builder(
+                        self.transaction_factory.transfer(receiver, amount),
+                    );
+                    Transaction::UserTransaction(txn)
+                })
+                .chain(once(Transaction::StateCheckpoint(HashValue::random())))
+                .collect();
+            self.version += transactions.len() as Version;
```
manudhundi, line 497: If possible, can you better quantify (log/debug log) the generated txns pattern in the laymen terms that you can use when you present the results. It maybe better to say with 'C' connected groups or with 'H' account hotspots (accounts with over 'T' txns threshold) we got 'Y tps' vs saying with 'hotspot_probability 0.xx' we got Y tps.

execution/executor-benchmark/src/transaction_generator.rs
zjma: fixed with `conflicts_with_all` macro, should also be fine?

execution/executor-benchmark/src/main.rs
zjma: fixed

execution/executor-benchmark/src/main.rs
zjma: fixed

execution/executor-benchmark/src/lib.rs
zjma: fixed

execution/executor-benchmark/src/transaction_generator.rs
zjma: fixed

execution/executor-benchmark/src/transaction_generator.rs
zjma: you just mean to print out the block generator details?

also, I assume you want it to be printed at the end of the execution benchmark?

manudhundi: Yes the details of block generator output (txn pattern generated). Only if it is possible to quantify the txn pattern in some other way which is more intuitive than hotspot ratio.

Doesn't matter where it is printed because we can always look it up.

zjma: still can't come up with anything more representative than a single `hotspot_probability` 😂  
Anyway, it's printing all block generator parameters now.

msmouse (approved):

grao1991 (approved):

PR #9796, opened 2023-08-28
Remove explicit copy statement for vector

fgfm999:
### Description
Seems the `copy` keyword is not needed for a copyable vector, Tried with aptos-move `2.0.3`. 
Also talked about this with Matt on Discord.

By the way, the original statement makes sense for why explicit copy is needed. Very grateful if reviewers could talk about the reason for this change.

> Copies of large vectors can be expensive, so the compiler requires explicit copy's to make it easier to see where they are happening. 

### Test Plan
The following test passed.

```move
module 0xcafe::capy {
    #[test_only]
    use std::vector;

    struct Book has copy, drop {}

    #[test]
    fun implicty_copy() {
        let v0 = vector[Book {}, Book {}, Book {}];

        let v1 = v0; // implicit copy. The keyword `copy` is not required.

        assert!(vector::length(&v0) == 3, 0);
        assert!(vector::length(&v1) == 3, 1);

        let _ = vector::pop_back(&mut v1);

        assert!(vector::length(&v0) == 3, 2); // the original vector is not modified
        assert!(vector::length(&v1) == 2, 3);

    }
}
```

### DIFF
--- a/developer-docs-site/docs/move/book/variables.md
+++ b/developer-docs-site/docs/move/book/variables.md
@@ -714,14 +714,11 @@ before it is assigned a value.
 As mentioned above, the Move compiler will infer a `copy` or `move` if one is not indicated. The
 algorithm for doing so is quite simple:
 
-- Any scalar value with the `copy` [ability](./abilities.md) is given a `copy`.
+- Any value with the `copy` [ability](./abilities.md) is given a `copy`.
 - Any reference (both mutable `&mut` and immutable `&`) is given a `copy`.
   - Except under special circumstances where it is made a `move` for predictable borrow checker
     errors.
 - Any other value is given a `move`.
-  - This means that even though other values might be have the `copy` [ability](./abilities.md), it
-    must be done _explicitly_ by the programmer.
-  - This is to prevent accidental copies of large data structures.
 
 For example:
 
--- a/developer-docs-site/docs/move/book/variables.md
+++ b/developer-docs-site/docs/move/book/variables.md
@@ -723,13 +723,21 @@ algorithm for doing so is quite simple:
 For example:
 
 ```move
+struct Foo {
+    f: u64
+}
+
+struct Coin has copy {
+    value: u64
+}
+
 let s = b"hello";
 let foo = Foo { f: 0 };
 let coin = Coin { value: 0 };
 
-let s2 = s; // move
+let s2 = s; // copy
 let foo2 = foo; // move
-let coin2 = coin; // move
+let coin2 = coin; // copy
 
 let x = 0;
 let b = false;
--- a/developer-docs-site/docs/move/book/variables.md
+++ b/developer-docs-site/docs/move/book/variables.md
@@ -719,6 +719,9 @@ algorithm for doing so is quite simple:
   - Except under special circumstances where it is made a `move` for predictable borrow checker
     errors.
 - Any other value is given a `move`.
+- If the compiler can prove that the source value with copy ability is not used after the 
+  assignment, then a move may be used instead of a copy for performance, but this will be invisible
+  to the programmer (except in possible decreased time or gas cost).
 
 For example:
 
--- a/developer-docs-site/docs/move/book/vector.md
+++ b/developer-docs-site/docs/move/book/vector.md
@@ -157,7 +157,8 @@ fun destroy_droppable_vector<T: drop>(vec: vector<T>) {
 }
 ```
 
-Similarly, vectors cannot be copied unless the element type has `copy`. 
+Similarly, vectors cannot be copied unless the element type has `copy`. In other words, a
+`vector<T>` has `copy` if and only if `T` has `copy`.
 
 For more details see the sections on [type abilities](./abilities.md) and [generics](./generics.md).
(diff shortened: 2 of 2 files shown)


brmataptos (commented):
This should really be a compiler bug, not a documentation bug, and we should fix it.  Requiring an explicit copy in the source is a good idea.

brmataptos:
I filed a bug to track this issue in the compiler: https://github.com/aptos-labs/aptos-core/issues/9916

brmataptos:
Thanks for finding this.  I assume it was not a deliberate change, but a bug.

fgfm999:
Hi, @brmataptos, `Explicit copy ` for vectors really makes sense 🙂.

What is the **default behavior** when assigning a vector to another variable, 
or there's not,  we **must** add a `move` or `copy` keyword to tell the compiler what we intend to do explicitly due to docs？

```Move
let x = vector::singleton<u64>(10);
let y = copy x; // compiler error without the copy!
```

And what about a **copyable** struct that contains a vector field, like `SimpleMap`? Explicited copy needed?

developer-docs-site/docs/move/book/vector.md
```diff
@@ -157,22 +157,11 @@ fun destroy_droppable_vector<T: drop>(vec: vector<T>) {
...
-implicitly copied:
-
-```move
-let x = vector::singleton<u64>(10);
-let y = copy x; // compiler error without the copy!
-```
-
-Copies of large vectors can be expensive, so the compiler requires explicit `copy`'s to make it
-easier to see where they are happening.
+Similarly, vectors cannot be copied unless the element type has `copy`.
```
xbtmatt, line 160:
```suggestion
Similarly, vectors cannot be copied unless the element type has `copy`. In other words, a
`vector<T>` has `copy` if and only if `T` has `copy`.
```

I think if we do change this, we should at least leave this part in, it's a little clearer!

brmataptos:
> Hi, @brmataptos, `Explicit copy ` for vectors really makes sense 🙂.
> 
> And what about a **copyable** struct that contains a vector field, like `SimpleMap`? Explicited copy needed?

Ok, I got info from someone who's been around longer:  An explicit copy was originally needed , but it was a big pain when a lot of code was written, so it was removed.  If all fields have copy, then struct should have copy, and assignment becomes copy if the source is not provably dead after this point.

Although perhaps there are cases where subsequent use of the struct/vector can be proven to not happen, and (less expensive) move can occur instead.  I guess if you want to be sure the compiler does the right thing, you can be explicit about move/copy. 

Ideally, we would state that in this doc: 

An assignment statement by default _moves_ a value from one variable/location to another.  If the source variable is not provably dead after the assignment, then a _move_ would be illegal, so a compile error will indicate this unless the type of the value has `copy` and can be transformed into a _copy_ operation.  In most cases, the compiler will get it right, but for large types (e.g., struct or vector) you can force a _move_ by explicitly including that.

brmataptos:
And this code also needs updating:

https://aptos.dev/move/book/variables/#inference which is located in (developer-docs-site/docs/move/book/variables.md)
.

developer-docs-site/docs/move/book/vector.md
fgfm999: Get it back. 
By the way, are you the Matt I talked with in Discord 😀?

fgfm999:
@brmataptos 
Thank you for your detailed explanation.
I had updated `variables.md`, please take a look.

developer-docs-site/docs/move/book/variables.md
```diff
@@ -714,25 +714,30 @@ before it is assigned a value.
 As mentioned above, the Move compiler will infer a `copy` or `move` if one is not indicated. The
 algorithm for doing so is quite simple:
 
-- Any scalar value with the `copy` [ability](./abilities.md) is given a `copy`.
+- Any value with the `copy` [ability](./abilities.md) is given a `copy`.
 - Any reference (both mutable `&mut` and immutable `&`) is given a `copy`.
   - Except under special circumstances where it is made a `move` for predictable borrow checker
     errors.
 - Any other value is given a `move`.
```
brmataptos, line 721: If the compiler can prove that the source value with `copy` ability is not used after the assignment, then a move may be used instead of a copy for performance, but this will be invisible to the programmer (except in possible decreased time or gas cost).

fgfm999: done

fgfm999: BTW, maybe the following code demonstrates the compiler will always prefer `move` if possible.

```move
    struct Car has store, copy {}
    struct User has key {
        car: Car
    }
    #[test]
    fun test_move_and_copy() {
        let s1 = create_signer_for_test(@0xcafe);
        let s2 = create_signer_for_test(@0xabcd);

        let c = Car {};
        let u1 = User { car: copy c }; // copy
        let u2 = User { car: move c }; // move, since car can not be dropped, and this test passed.
        move_to(&s1, u1);
        move_to(&s2, u2);
    }
```

davidiw (approved):

PR #9801, opened 2023-08-28
[Configs] Add seed peers for testnet and mainnet.

JoshLind:
### Description
This PR adds dedicated seed peer support for testnet and mainnet. This is useful for nodes that wish to sync to the network with only the genesis blob.

### Test Plan
Existing test infrastructure.

### DIFF
--- a/config/src/config/config_optimizer.rs
+++ b/config/src/config/config_optimizer.rs
@@ -4,16 +4,54 @@
 use crate::{
     config::{
         node_config_loader::NodeType, utils::get_config_name, Error, InspectionServiceConfig,
-        LoggerConfig, MempoolConfig, NodeConfig, PeerMonitoringServiceConfig, StateSyncConfig,
+        LoggerConfig, MempoolConfig, NodeConfig, Peer, PeerMonitoringServiceConfig, PeerRole,
+        PeerSet, StateSyncConfig,
     },
     network_id::NetworkId,
 };
-use aptos_types::chain_id::ChainId;
+use aptos_crypto::{x25519, ValidCryptoMaterialStringExt};
+use aptos_types::{chain_id::ChainId, network_address::NetworkAddress, PeerId};
+use maplit::hashset;
 use serde_yaml::Value;
+use std::{collections::HashMap, str::FromStr};
 
 // Useful optimizer constants
 const OPTIMIZER_STRING: &str = "Optimizer";
+const PUBLIC_NETWORK_OPTIMIZER_NAME: &str = "PublicNetworkConfigOptimizer";
 const VALIDATOR_NETWORK_OPTIMIZER_NAME: &str = "ValidatorNetworkConfigOptimizer";
+
+// Mainnet seed peers. Each seed peer entry is a tuple
+// of (account address, public key, network address).
+const MAINNET_SEED_PEERS: [(&str, &str, &str); 1] = [(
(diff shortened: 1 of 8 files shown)


config/src/config/config_optimizer.rs
```diff
@@ -4,17 +4,47 @@
...
+const PUBLIC_NETWORK_OPTIMIZER_NAME: &str = "PublicNetworkConfigOptimizer";
 const VALIDATOR_NETWORK_OPTIMIZER_NAME: &str = "ValidatorNetworkConfigOptimizer";
 
+// Mainnet seed peer addresses and keys
+const MAINNET_SEED_PEER_ADDRESSES: [&str; 0] = [];
+const MAINNET_SEED_PEER_NETWORK_ADDRESSES: [&str; 0] = [];
+const MAINNET_SEED_PEER_PUBLIC_KEYS: [&str; 0] = [];
+
+// Testnet seed peer addresses and keys
+const TESTNET_SEED_PEER_ADDRESSES: [&str; 4] = [
```
brianolson, line 29: I wish there was either a loud comment "these three are one thing and must be the same length and correspond by index", or (really this) I wish it was one list of something with three elements. maybe [[&str; 3]; 4] or vec![] of (str, str, str) tuples?

config/src/config/config_optimizer.rs
```diff
@@ -4,17 +4,47 @@
...
 
 // Useful optimizer constants
 const OPTIMIZER_STRING: &str = "Optimizer";
+const PUBLIC_NETWORK_OPTIMIZER_NAME: &str = "PublicNetworkConfigOptimizer";
 const VALIDATOR_NETWORK_OPTIMIZER_NAME: &str = "ValidatorNetworkConfigOptimizer";
 
+// Mainnet seed peer addresses and keys
+const MAINNET_SEED_PEER_ADDRESSES: [&str; 0] = [];
+const MAINNET_SEED_PEER_NETWORK_ADDRESSES: [&str; 0] = [];
+const MAINNET_SEED_PEER_PUBLIC_KEYS: [&str; 0] = [];
```
sionescu, line 26: It would be better to use a more specific type that 

 1) Ties together these 3 properties
 2) Parses the strings into something more appropriate, like a bignum

config/src/config/config_optimizer.rs
JoshLind: SGTM! I've gone for an array of tuples of (str, str, str) 😄

config/src/config/config_optimizer.rs
sionescu: The address is not a string, it's a large integer, so it should be parsed as such. The string is only its printed representation.

sherry-x (approved):

config/src/config/config_optimizer.rs
JoshLind: Aah, gotcha 😄
1. The peer address here is a hex formatted string. When we build the seed from it, we use `PeerId::from_hex(...)`, (which appears to be the standardized way of handling addresses across our code). Likewise, it's the format expected by the serde implementation for the object (e.g., if you were to define these in the node config manually).
2. I also think it's much more readable than storing a large number in the code. For example, I can directly look it up in our [explorer](https://explorer.aptoslabs.com/account/31e55012a7d439dcd16fee0509cd5855c1fbdc62057ba7fac3f7c88f5453dd8e?network=testnet) 😄  

So, I'm inclined to leave this as it is. I think it's superior to a large number 🤔

sionescu (approved):

PR #9808, opened 2023-08-28
disable module_event feature after framework update is done

msmouse:
### Description
The module event feature is supposed to be enabled only briefly during framework upgrade to allow the new framework to publish, until the feature is approved by governance and enabled again. This tests the workflow and tooling work.

### Test Plan

framework_upgrade forge test


davidiw (approved):
so fancy

grao1991 (approved):

PR #9814, opened 2023-08-29
[TS SDK v2] Add BCS serialization/deserialization

xbtmatt:
### Description

The following is a WIP. I might remove the ability to make procedural calls, but I want to finish the full loop of transaction serialization first before seeing what makes the most sense.

I've added the ability to serialize values procedurally, because it makes it much clearer to the developer that they should generally be constructing fully Serializable values and getting the bytes, rather than serializing individual fields/arguments.

I renamed `bcsToBytes` to `serialize` and put it in the `Serializable` class.  You can now chain calls to this function like so:

```typescript
    const serializedBytes = serializer
      .serializeBytes(new Uint8Array([0x41, 0x70, 0x74, 0x6f, 0x73]))
      .serializeBool(true)
      .serializeBool(false)
      .serializeU8(254)
      .serializeU8(255)
      .serializeU16(65535)
      .serializeU16(4660)
      .serializeU32(4294967295)
      .serializeU32(305419896)
      .serializeU64(BigInt("18446744073709551615"))
      .serializeU64(BigInt("1311768467750121216"))
      .serializeU128(BigInt("340282366920938463463374607431768211455"))
      .serializeU128(BigInt("1311768467750121216"))
      .getBytes();
```

You can also now call `serialize` easily on `Serializable` values:

```typescript
    class MoveStructB implements Serializable {
      constructor(
        public moveStructA: MoveStructA,
        public name: string,
        public description: string,
        public numbers: Array<number>,
      ) {}

      serialize(serializer: Serializer): void {
        serializer
          .serialize(this.moveStructA)
          .serializeStr(this.name)
          .serializeStr(this.description)
          .serializeBytes(new Uint8Array(this.numbers));
      }
    }

    // where MoveStructA is another Serializable class
    const moveStructA = new MoveStructA("abc", "123", false, [1, 2, 3, 4]);
    const moveStructB = new MoveStructB(moveStructA, "def", "456", [5, 6, 7, 8]);

    const serializedBytes = new Serializer().serialize(moveStructB).getBytes();
```

You can still do everything the old way, it's just possible to chain serialization calls now.

The plan is to discourage developers from serializing individual transaction arguments and encourage them to create Serializable classes that are more intuitive and readable.

If we do it like this, we can offer the developer two choices whenever they need to serialize a value:

1. They can use a Serializable value, and whatever function they're using serializes it for them
2. They can use a Uint8Array, and whatever function they're using assumes it is BCS serialized correctly.

WRT the transaction builder pattern, specifically for say, an entry function, the former would mean they declare and later instantiate the class to use in a payload like this:

```typescript
class MyFunctionPayload implements Serializable {
    constructor(
        public amount: number,
        public enabled: boolean,
        public description: string,
    ) {}
    
    // for standalone usage, maybe this could go in Serializer, though
    toBCSBytes(): Uint8Array {
        const serializer = new Serializer();
        this.serialize(serializer);
        return serializer.getBytes();
    }

    // for composition
    serialize(serializer: Serializer): void {
        serializer
            .serializeU64(this.amount)
            .serializeBool(this.enabled)
            .serializeStr(this.description);
    }
}

const rawTxn = new EntryFunction(
    "0x1::my_module::my_function",
    [], // type args
    new MyFunctionPayload(1, false, "hello world").toBCSBytes(),
    // or
    // new Serializer().serialize(new MyfunctionPayload(1, false, "hello world")).getBytes(),
);
```

Or in the latter case, the developer can serialize their own arguments directly in the transaction entry function call like before:

```typescript
const rawTxn = new EntryFunction(
    "0x1::some_module::some_function",
    [], // type args
    new Serializer()
        .serializeU64(numValue)
        .serializeBool(boolValue)
        .serializeStr(stringValue)
        .getBytes()
);
```

This leaves open the possibility for ABI generation with Serializable types, eventually.

We can also continue to support the ability to serialize args with the remote ABI builder, too.

TODO:

- Add unit tests for `serializeVector<T>`
- Add `deserializeVector<T>`
- Integrate `TypeTag` values, either as `TypeTag` or `MoveType` or something similar
- Add `TypeTag` serialization with `serialize` to unit tests
- Test examples of a new transaction builder that utilizes procedural serialization and composition
- Do the above with `TypeTag`
- Consider whether or not the helper functions need to be as comprehensive as the tx remote abi builder
- Add an example for serializing `StructTags/Structs` either with the old process or some new way

### Test Plan
I've added unit tests for the above in the serializer unit tests. I also included all the old unit tests.

### DIFF
--- a/ecosystem/typescript/sdk_v2/src/aptos_types/ed25519.ts
+++ b/ecosystem/typescript/sdk_v2/src/aptos_types/ed25519.ts
@@ -1,0 +1,49 @@
+// Copyright © Aptos Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+import { Deserializer, Serializer } from "../bcs";
+
+export class Ed25519PublicKey {
+  static readonly LENGTH: number = 32;
+
+  readonly value: Uint8Array;
+
+  constructor(value: Uint8Array) {
+    if (value.length !== Ed25519PublicKey.LENGTH) {
+      throw new Error(`Ed25519PublicKey length should be ${Ed25519PublicKey.LENGTH}`);
+    }
+    this.value = value;
+  }
+
+  toBytes(): Uint8Array {
+    return this.value;
+  }
+
+  serialize(serializer: Serializer): void {
+    serializer.serializeByteVector(this.value);
+  }
+
+  static deserialize(deserializer: Deserializer): Ed25519PublicKey {
+    const value = deserializer.deserializeBytes();
+    return new Ed25519PublicKey(value);
+  }
+}
+
+export class Ed25519Signature {
+  static readonly LENGTH = 64;
+
+  constructor(public readonly value: Uint8Array) {
+    if (value.length !== Ed25519Signature.LENGTH) {
+      throw new Error(`Ed25519Signature length should be ${Ed25519Signature.LENGTH}`);
+    }
+  }
+
+  serialize(serializer: Serializer): void {
+    serializer.serializeByteVector(this.value);
+  }
+
+  static deserialize(deserializer: Deserializer): Ed25519Signature {
+    const value = deserializer.deserializeBytes();
+    return new Ed25519Signature(value);
+  }
+}
--- a/ecosystem/typescript/sdk_v2/src/bcs/serializer.ts
+++ b/ecosystem/typescript/sdk_v2/src/bcs/serializer.ts
@@ -12,6 +12,10 @@ import {
 } from "./consts";
 import { AnyNumber, Uint16, Uint32, Uint8 } from "./types";
 
+export interface Serializable {
+  serialize(serializer: Serializer): void;
+}
+
 export class Serializer {
   private buffer: ArrayBuffer;
 
@@ -30,7 +34,7 @@ export class Serializer {
     }
   }
 
-  protected serialize(values: Uint8Array) {
+  protected appendToBuffer(values: Uint8Array) {
     this.ensureBufferWillHandleSize(values.length);
     new Uint8Array(this.buffer, this.offset).set(values);
     this.offset += values.length;
@@ -49,7 +53,7 @@ export class Serializer {
 
   /**
    * Serializes a string. UTF8 string is supported.
-   * 
+   *
    * The number of bytes in the string content is serialized first, as a uleb128-encoded u32 integer.
    * Then the string content is serialized as UTF8 encoded bytes.
    *
@@ -63,9 +67,10 @@ export class Serializer {
    * assert(serializer.getBytes() === new Uint8Array([8, 49, 50, 51, 52, 97, 98, 99, 100]));
    * ```
    */
-  serializeStr(value: string): void {
+  serializeStr(value: string): this {
     const textEncoder = new TextEncoder();
     this.serializeBytes(textEncoder.encode(value));
+    return this;
   }
 
   /**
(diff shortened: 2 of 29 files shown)


ecosystem/typescript/sdk_v2/src/bcs/types.ts
```diff
@@ -0,0 +1,10 @@
+// Copyright © Aptos Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+export type Uint8 = number;
+export type Uint16 = number;
+export type Uint32 = number;
+export type Uint64 = bigint;
+export type Uint128 = bigint;
+export type Uint256 = bigint;
+export type AnyNumber = bigint | number;
```
0xjinn, line 10: Are we skipping the alias from sdk v1 `export type Bytes = Uint8Array;`? In this case, we will be using `Uint8Array` instead of `Bytes` on sdk v2?

xbtmatt: Yeah, we talked about it a bit back that we wanted to get rid of the reference to `Bytes`. It was in a random slack thread, lol. It is easy to add back in, but we don't really use it anywhere else, so I removed it for now. It's fairly easy to add back in

xbtmatt: Also with some of this stuff, I am playing around with different types of serializtaion/deserialization. The core BCS serialization/deserialization should generally stay the same, though

0xjinn: Cool! I am gonna start rebasing some of my stuff on top of this branch 😄

xbtmatt: Ok cool :) Lmk if something blocks you or is confusing. I kept most of the same file structure even

ecosystem/typescript/sdk_v2/src/bcs/serializer.ts
```diff
@@ -0,0 +1,308 @@
...
+    return this;
+  }
+
+  /**
+   * Serializes an array of bytes.
+   *
+   * BCS layout for "bytes": bytes_length | bytes
+   * where bytes_length is a u32 integer encoded as a uleb128 integer, equal to the length of the bytes array.
+   */
+  serializeByteVector(value: Uint8Array): this {
```
0xjinn, line 82: I noticed that the name in v1 is `serializeBytes`, shall we keep it the same? I think Bytes and Uint8Array feel much the same.

xbtmatt: yeah I might go back to it, the only reason I changed it is because in the old SDK `serializeBytes` is confusing, because there's also `serializeFixedBytes` and `bcsToBytes`.

They all do 3 different things.

Where as `serializeByteVector` is very clear that you're serializing a representation of `vector<u8>`

ecosystem/typescript/sdk_v2/src/aptos_types/ed25519.ts
```diff
@@ -0,0 +1,49 @@
+// Copyright © Aptos Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+import { Deserializer, Serializer } from "../bcs";
+
+export class Ed25519PublicKey {
```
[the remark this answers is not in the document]
xbtmatt: Yeah, I had to add them all for the tx builder. Maybe we should add the base classes and merge them so we don't have different versions floating around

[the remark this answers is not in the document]
0xjinn: Ohh wait, is the BCS stuff includes txn builder?

PR #9820, opened 2023-08-29
Fix number of accounts in transaction benchmark

sitalkedia:
### Description

The number of accounts was smaller than the block size which causes issue with no conflict txn case (default)


sitalkedia:
Closing in favor of https://github.com/aptos-labs/aptos-core/pull/9821

PR #9827, opened 2023-08-30
[features] fix flags

davidiw:
* Some flags were not enabled by default even though they are marked for release
* One flag was off by 3
* Flags were in an inconsistent order making it hard to insure they were enabled..

let's see if this blows up...


movekevin (approved):

aptos-move/e2e-testsuite/src/tests/data_store.rs
```diff
@@ -102,7 +102,6 @@ fn borrow_after_move() {
     let (module, txn) = add_module_txn(&sender, 10);
     executor.execute_and_apply(txn);
 
-    println!("HERE!");
```
lightmark, line 105: hhh

aptos-move/e2e-move-tests/src/tests/module_event.rs
```diff
@@ -45,22 +45,24 @@ fn test_module_event_enabled() {
...
-        );
-        assert_eq!(
-            bcs::from_bytes::<MyEvent>(module_event.event_data()).unwrap(),
-            MyEvent {
-                seq: i as u64,
-                field: Field { field: false },
-                bytes: vec![],
-            }
-        );
+    assert_eq!(events.len(), 12);
```
lightmark, line 48: why now it is 12?

davidiw: because it never accounted for fees, I guess

aptos-move/e2e-testsuite/src/tests/data_store.rs
davidiw: :yawning_face:

msmouse:
What's this, shall we consolidate?
https://github.com/aptos-labs/aptos-core/blob/99170f0528f9b0cd24064a1b47b60ffee2a37931/types/src/on_chain_config/aptos_features.rs#L51

aptos-move/e2e-move-tests/src/tests/module_event.rs
lightmark: @msmouse 's fees? This is so annoying from devss perspective.

lightmark (approved):

PR #9835, opened 2023-08-30
[move] remove aptos transactional test

davidiw:
when I was updating the feature set I ran into this monstrosity

we have had zero asks for support on this... I am highly confident we built it with great expectations and then did little to support it... the better approach is to make it easier to test from SDKs as I recently did in Python... which is pretty easy to do in TS.


gregnazario (approved):
Can you remove it from the docs as well?

movekevin (approved):

wrwg (approved):

brmataptos:
How do I sign up to be notified about changes like this?

I find the transactional-test framework useful to debug some compiler/VM validation bugs.  Why do I need to write a custom .rs driver in e2e-move-tests just to compile and load some move code illustrating a bug?  While we still have transactional-tests in third_party, some bug reports involve aptos libs.

This change is quite a pain, as I already had reproduced a bug on a branch using this and now I have to go and rebase back up the tree or something to reproduce it.  And then write some Rust code driver just to add a regression test for it.

zi0Black:
> How do I sign up to be notified about changes like this?
> 
> I find the transactional-test framework useful to debug some compiler/VM validation bugs. Why do I need to write a custom .rs driver in e2e-move-tests just to compile and load some move code illustrating a bug? While we still have transactional-tests in third_party, some bug reports involve aptos libs.
> 
> This change is quite a pain, as I already had reproduced a bug on a branch using this and now I have to go and rebase back up the tree or something to reproduce it. And then write some Rust code driver just to add a regression test for it.

We had consultants who used this in the process of writing POCs and had the same question as @brmataptos.

wrwg:
> > How do I sign up to be notified about changes like this?
> > I find the transactional-test framework useful to debug some compiler/VM validation bugs. Why do I need to write a custom .rs driver in e2e-move-tests just to compile and load some move code illustrating a bug? While we still have transactional-tests in third_party, some bug reports involve aptos libs.
> > This change is quite a pain, as I already had reproduced a bug on a branch using this and now I have to go and rebase back up the tree or something to reproduce it. And then write some Rust code driver just to add a regression test for it.
> 
> We had consultants who used this in the process of writing POCs and had the same question as @brmataptos.

Yes, this was a misunderstanding about internal use of this framework and for end users. Its not for end users. We have a work item to bring it back for internal use. (Feel freee to revert this PR if you need it soon.)

PR #9838, opened 2023-08-30
[indexer-docs] Use absolute paths for links

banool:
### Description
There is some weird behavior in docusaurus where relative links don't work when the current URL has a trialing slash. See more information here: https://aptos-org.slack.com/archives/C04PF1X2UKY/p1693348785934909. I fix this by using absolute links instead.

### Test Plan
CI.


larry-aptos (approved):

banool:
Relevant issue: https://github.com/facebook/docusaurus/issues/3372.

PR #9842, opened 2023-08-30
Generate transactions in multiple threads.

grao1991:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1228,6 +1228,7 @@ dependencies = [
  "bcs 0.1.4",
  "chrono",
  "clap 4.3.21",
+ "derivative",
  "indicatif 0.15.0",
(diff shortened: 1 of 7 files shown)


sitalkedia (approved):
LGTM. Please take care of making the number of workers configurable before landing and also may be make the number of worker 1 by default?

grao1991:
> LGTM. Please take care of making the number of workers configurable before landing and also may be make the number of worker 1 by default?

Made it configurable. 
We need at least 3 workers to make it faster than the rest of the pipeline. I'd like to set the default to 4 since I like power of 2. 
Given the channel between the generator and the rest of pipeline is bounded, I don't worry about setting it to a larger value.

lightmark (approved):

PR #9844, opened 2023-08-30
[consensus-only] execution and fakeaptosdb bugfix and realistic env test

ibalajiarun:
### Description

This PR fixes/improves the consensus only benchmark:
- It fixes a bug that prevented running the benchmark due to execution results divergence during state-sync because we now add the StateCheckpoint transaction output with zero gas post-execution in the normal flow that conflicted with state-sync execution flow where in consensus-only mode, the StateCheckpoint transaction output set a non-zero gas.
- Mocks the `get_block_timestamp()` call to return the highest known block timestamp instead of passing through to the underlying aptosdb that made the call to fail and txn-emitter to consequentially fail.
- It also a realistic env forge test with tuned parameters to achieve 50-60k tps with up to 100 replicas. 

### Test Plan

Now able to get 50-60k tps consistently.

https://github.com/aptos-labs/aptos-core/pull/9844#issuecomment-1704424974

### DIFF
--- a/.github/workflows/docker-build-test.yaml
+++ b/.github/workflows/docker-build-test.yaml
@@ -349,11 +349,11 @@ jobs:
     secrets: inherit
     with:
       GIT_SHA: ${{ needs.determine-docker-build-metadata.outputs.gitSha }}
-      FORGE_TEST_SUITE: consensus_only_perf_benchmark
+      FORGE_TEST_SUITE: consensus_only_realistic_env_max_tps
       IMAGE_TAG: consensus_only_perf_test_${{ needs.determine-docker-build-metadata.outputs.gitSha }}
       FORGE_RUNNER_DURATION_SECS: 300
-      COMMENT_HEADER: forge-consensus-only-perf-test
-      FORGE_NAMESPACE: forge-consensus-only-perf-test-${{ needs.determine-docker-build-metadata.outputs.targetCacheId }}
+      COMMENT_HEADER: consensus-only-realistic-env-max-tps
+      FORGE_NAMESPACE: forge-consensus-only-realistic-env-max-tps-${{ needs.determine-docker-build-metadata.outputs.targetCacheId }}
 
   # Run forge multiregion test. This test uses the multiregion forge cluster that deploys pods in three GCP regions.
   forge-multiregion-test:
--- a/execution/executor-types/src/executed_chunk.rs
+++ b/execution/executor-types/src/executed_chunk.rs
@@ -133,7 +133,9 @@ impl ExecutedChunk {
                     .ledger_info()
                     .transaction_accumulator_hash()
                     == result_accumulator.root_hash(),
-                "Root hash in target ledger info does not match local computation."
+                "Root hash in target ledger info does not match local computation. {:?} != {:?}",
+                verified_target_li,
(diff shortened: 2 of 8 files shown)


sitalkedia:
Hmm.. the CPU utilization is pretty high - we are using around 10 cpu core for consensus only - https://aptoslabs.grafana.net/d/system/system?from=1693780311000&to=1693781027000&var-Datasource=fHo-R604z&var-BigQuery=axNEitxVz&var-metrics_source=All&var-chain_name=forge-0&var-cluster=All&var-namespace=forge-consensus-only-realistic-env-max-tps-pr-9844&var-kubernetes_pod_name=All&orgId=1. Do you have insight what is using so much CPU?

ibalajiarun:
> Hmm.. the CPU utilization is pretty high - we are using around 10 cpu core for consensus only - https://aptoslabs.grafana.net/d/system/system?from=1693780311000&to=1693781027000&var-Datasource=fHo-R604z&var-BigQuery=axNEitxVz&var-metrics_source=All&var-chain_name=forge-0&var-cluster=All&var-namespace=forge-consensus-only-realistic-env-max-tps-pr-9844&var-kubernetes_pod_name=All&orgId=1. Do you have insight what is using so much CPU?

I haven't debugged @sitalkedia . May be the perf tool can help us figure?

sitalkedia:
> > Hmm.. the CPU utilization is pretty high - we are using around 10 cpu core for consensus only - https://aptoslabs.grafana.net/d/system/system?from=1693780311000&to=1693781027000&var-Datasource=fHo-R604z&var-BigQuery=axNEitxVz&var-metrics_source=All&var-chain_name=forge-0&var-cluster=All&var-namespace=forge-consensus-only-realistic-env-max-tps-pr-9844&var-kubernetes_pod_name=All&orgId=1. Do you have insight what is using so much CPU?
> 
> I haven't debugged @sitalkedia . May be the perf tool can help us figure?

Is the execution/ledger update part in this test a no-op and the CPU is purely coming from mempool/consensus?

ibalajiarun:
> > > Hmm.. the CPU utilization is pretty high - we are using around 10 cpu core for consensus only - https://aptoslabs.grafana.net/d/system/system?from=1693780311000&to=1693781027000&var-Datasource=fHo-R604z&var-BigQuery=axNEitxVz&var-metrics_source=All&var-chain_name=forge-0&var-cluster=All&var-namespace=forge-consensus-only-realistic-env-max-tps-pr-9844&var-kubernetes_pod_name=All&orgId=1. Do you have insight what is using so much CPU?
> > 
> > 
> > I haven't debugged @sitalkedia . May be the perf tool can help us figure?
> 
> Is the execution/ledger update part in this test a no-op and the CPU is purely coming from mempool/consensus?

The execution is a no-op. From grafana, the txn shuffle takes 200ms single-threaded. There is db update part (txn accumulator update) that takes 200ms and  is single threaded as well. So, I don't see execution and db update taking more than one CPU. I would say its mostly quorum store, mempool, consensus and state sync (if any).

sitalkedia (approved):
We need to followup to understand the root cause of such high CPU usage though..

danielxiangzl (approved):

PR #9852, opened 2023-08-31
[Performance] Refactor ThreadManager, and use it in some places.

grao1991:
### Description
Default to the "default" strategy.
Adjusted the threadpool, under the default strategy, commit is slightly slower, execution is slightly faster.

This PR doesn't cover all the threadpool usages. There are some other things that may need to move to a different thread pool, but I'm not super confident that they won't affect production, need more investigations.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -963,6 +963,7 @@ dependencies = [
  "aptos-crypto",
  "aptos-db-indexer",
  "aptos-executor-types",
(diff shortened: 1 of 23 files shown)


msmouse (approved):
let's land after the cut -_-

zekun000 (approved):

sitalkedia:
@grao1991 - Can we quantify the performance improvement we get from this change?

PR #9854, opened 2023-08-31
add different suites to single_node_performance.py

igor-aptos:

### DIFF
--- a/execution/executor-benchmark/src/main.rs
+++ b/execution/executor-benchmark/src/main.rs
@@ -192,8 +192,11 @@ struct Opt {
     #[clap(long, conflicts_with_all = &["connected_tx_grps", "transactions_per_sender"])]
     hotspot_probability: Option<f32>,
 
-    #[clap(long)]
-    concurrency_level: Option<usize>,
+    #[clap(
+        long,
+        help = "Number of threads to use for execution. Generally replaces --concurrency-level flag (directly for default case, and as a total across all shards for sharded case)"
+    )]
+    execution_threads: Option<usize>,
 
     #[clap(flatten)]
     pruner_opt: PrunerOpt,
@@ -225,18 +228,14 @@ struct Opt {
 }
 
 impl Opt {
-    fn concurrency_level(&self) -> usize {
-        match self.concurrency_level {
+    fn execution_threads(&self) -> usize {
+        match self.execution_threads {
             None => {
-                let level = (num_cpus::get() as f64 / self.pipeline_opt.num_executor_shards as f64)
-                    .ceil() as usize;
-                println!(
-                    "\nVM concurrency level defaults to {} for number of shards {} \n",
-                    level, self.pipeline_opt.num_executor_shards
-                );
-                level
+                let cores = num_cpus::get();
+                println!("\nExecution threads defaults to number of cores: {}", cores,);
+                cores
             },
-            Some(level) => level,
+            Some(threads) => threads,
         }
     }
 }
@@ -414,9 +413,19 @@ fn main() {
     aptos_node_resource_metrics::register_node_metrics_collector();
     let _mp = MetricsPusher::start_for_local_run("executor-benchmark");
 
-    AptosVM::set_concurrency_level_once(opt.concurrency_level());
-    AptosVM::set_num_shards_once(opt.pipeline_opt.num_executor_shards);
-    NativeExecutor::set_concurrency_level_once(opt.concurrency_level());
+    let execution_threads = opt.execution_threads();
+    let execution_shards = opt.pipeline_opt.num_executor_shards;
+    assert!(
+        execution_threads % execution_shards == 0,
+        "Execution threads ({}) must be divisible by the number of execution shards ({}).",
+        execution_threads,
+        execution_shards
+    );
+    let execution_threads_per_shard = execution_threads / execution_shards;
+
+    AptosVM::set_num_shards_once(execution_shards);
+    AptosVM::set_concurrency_level_once(execution_threads_per_shard);
+    NativeExecutor::set_concurrency_level_once(execution_threads_per_shard);
 
     let config = ProfilerConfig::new_with_defaults();
     let handler = ProfilerHandler::new(config);
--- a/testsuite/single_node_performance.py
+++ b/testsuite/single_node_performance.py
@@ -10,40 +10,83 @@ import json
 from typing import Callable, Optional, Tuple, Mapping, Sequence, Any
 from tabulate import tabulate
 from subprocess import Popen, PIPE, CalledProcessError
-from dataclasses import dataclass
+from dataclasses import dataclass, field
+from enum import Flag, auto
+
+
+class Flow(Flag):
+    # Tests that are run on PRs
+    LAND_BLOCKING = auto()
+    # Tests that are run continuously on main (in addition to LAND_BLOCKING ones)
+    CONTINUOUS = auto()
+    # Tests that are run manually when using a smaller representative mode.
+    # (i.e. for measuring speed of the machine)
+    REPRESENTATIVE = auto()
+
+
+@dataclass
+class RunGroupKey:
+    transaction_type: str
+    module_working_set_size: int = field(default=1)
+    executor_type: str = field(default="VM")
+
+    transaction_type_override: Optional[str] = field(default=None)
+    transaction_weights_override: Optional[str] = field(default=None)
+    sharding_traffic_flags: Optional[str] = field(default=None)
+
+
+@dataclass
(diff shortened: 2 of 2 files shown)


testsuite/single_node_performance.py
```diff
@@ -10,40 +10,73 @@
...
 from subprocess import Popen, PIPE, CalledProcessError
-from dataclasses import dataclass
+from dataclasses import dataclass, field
+from enum import Flag, auto
 
+class Flow(Flag):
+    NONE = 0
+    LAND_BLOCKING = auto()
+    CONTINUOUS = auto()
+    REPRESENTATIVE = auto()
```
sitalkedia, line 20: Can you add a comment here explaining what this represents?

testsuite/single_node_performance.py
```diff
@@ -10,40 +10,73 @@
...
-    # ("token-v1nft-mint-and-transfer-parallel", False, 20): (5300.0, True),
-    # ("token-v1ft-mint-and-store", False): 1000.0,
-    # ("token-v1nft-mint-and-store-sequential", False): 1000.0,
-    # ("token-v1nft-mint-and-store-parallel", False): 1000.0,
-    # ("no-op2-signers", False, 1): (18000.0, True),
-    # ("no-op5-signers", False, 1): (18000.0, True),
-    ("token-v2-ambassador-mint", False, 1): (1500.0, True),
-    ("token-v2-ambassador-mint", False, 20): (5500.0, True),
-}
+TESTS = [
```
sitalkedia, line 42: Don't we need one for sharded execution?

testsuite/single_node_performance.py
```diff
@@ -10,40 +10,73 @@
...
+    # RunGroupConfig(expected_tps=1000, key=RunGroupKey("token-v1nft-mint-and-store-sequential"), included_in=Flow.NONE),
+    # RunGroupConfig(expected_tps=1000, key=RunGroupKey("token-v1nft-mint-and-transfer-parallel"), included_in=Flow.NONE),
+
+    RunGroupConfig(expected_tps=18000, key=RunGroupKey("no-op2-signers"), included_in=Flow.CONTINUOUS),
+    RunGroupConfig(expected_tps=18000, key=RunGroupKey("no-op5-signers"), included_in=Flow.CONTINUOUS),
+   
+    RunGroupConfig(expected_tps=1500, key=RunGroupKey("token-v2-ambassador-mint"), included_in=Flow.LAND_BLOCKING | Flow.REPRESENTATIVE),
+    RunGroupConfig(expected_tps=5500, key=RunGroupKey("token-v2-ambassador-mint", module_working_set_size=20), included_in=Flow.LAND_BLOCKING | Flow.REPRESENTATIVE),
+
+    RunGroupConfig(expected_tps=50000, key=RunGroupKey("coin-transfer", executor_type="sharded", sharding_traffic_flags="--connected-tx-grps 5000"), included_in=Flow.REPRESENTATIVE),
```
igor-aptos, line 78: @sitalkedia  this is sharded execution. 

not run on PR or continuous, only when using FLOW=REPRESENTATIVE

testsuite/single_node_performance.py
```diff
@@ -10,40 +10,78 @@
...
+    # RunGroupConfig(expected_tps=1000, key=RunGroupKey("token-v1nft-mint-and-transfer-parallel"), included_in=Flow.NONE),
+
+    RunGroupConfig(expected_tps=18000, key=RunGroupKey("no-op2-signers"), included_in=Flow.CONTINUOUS),
+    RunGroupConfig(expected_tps=18000, key=RunGroupKey("no-op5-signers"), included_in=Flow.CONTINUOUS),
+   
+    RunGroupConfig(expected_tps=1500, key=RunGroupKey("token-v2-ambassador-mint"), included_in=Flow.LAND_BLOCKING | Flow.REPRESENTATIVE),
+    RunGroupConfig(expected_tps=5500, key=RunGroupKey("token-v2-ambassador-mint", module_working_set_size=20), included_in=Flow.LAND_BLOCKING | Flow.REPRESENTATIVE),
+
+    RunGroupConfig(expected_tps=50000, key=RunGroupKey("coin_transfer_connected_components", executor_type="sharded", sharding_traffic_flags="--connected-tx-grps 5000", transaction_type_override=""), included_in=Flow.REPRESENTATIVE),
+    RunGroupConfig(expected_tps=50000, key=RunGroupKey("coin_transfer_hotspot", executor_type="sharded", sharding_traffic_flags="--hotspot-probability 0.8", transaction_type_override=""), included_in=Flow.REPRESENTATIVE),
```
sitalkedia, line 83: Hmm we would expect around 25k TPS with this setting..let's verify this with your setup and we can update this number

sitalkedia, line 83: This is around 43k with my setup, so may be set this to a little lower like 45k?

sitalkedia (approved):
LGTM - thanks for taking care of it.

igor-aptos:
for numbers, let's re-calibrate after PR that improves hotspot configs, but I want to land this first - to land changes to the flags, to not conflict with them later.
I also need to think how to have different expected_tps for different number of execution threads, as that obviously affects things a lot.

msmouse (approved):

execution/executor-benchmark/src/main.rs
```diff
@@ -193,7 +193,7 @@ struct Opt {
     hotspot_probability: Option<f32>,
 
     #[clap(long)]
-    concurrency_level: Option<usize>,
+    execution_threads: Option<usize>,
```
zjma, line 196: This rename can break many informal docs🤔

igor-aptos: explanation of why I am changing this: 

for parallelism, I think that "--concurrency-level 1 --num-executor-shards 48" is a bit confusing overloading of the flags

we should have one flag of number of cores given (allowed to be used) by the execution stage. and then additional flags if sharding needs it.

currently --concurrency-level is that flag, so I would assume 8 means 8 cores are given
and then having --num-executor-shards be 8, if you want to run 8 BlockSTMs.

I can leave concurrency-level, but that is a bit ambiguous, so changing to execution_threads (in the table we print "exe 8" for example already), because that is what it means


I prefer breaking a few old commands for clarity. I can add some comment though

testsuite/single_node_performance.py
```diff
@@ -239,61 +279,62 @@ def print_table(
...
-        cur_block_size = int(min([expected_tps, BLOCK_SIZE]))
+        cur_block_size = int(min([test.expected_tps, MAX_BLOCK_SIZE]))
 
-        executor_type = "native" if use_native_executor else "VM"
+        sharding_traffic_flags = test.key.sharding_traffic_flags or ""
 
-        use_native_executor_str = "--use-native-executor" if use_native_executor else ""
-        common_command_suffix = f"{use_native_executor_str} --generate-then-execute --transactions-per-sender 1 --block-size {cur_block_size} --split-ledger-db --use-sharded-state-merkle-db --skip-index-and-usage run-executor --transaction-type {transaction_type} --transaction-weights {transaction_weights} --module-working-set-size {module_working_set_size} --main-signer-accounts {MAIN_SIGNER_ACCOUNTS} --additional-dst-pool-accounts {ADDITIONAL_DST_POOL_ACCOUNTS} --data-dir {tmpdirname}/db  --checkpoint-dir {tmpdirname}/cp"
+        if test.key.executor_type == "VM":
+            executor_type_str = "--transactions-per-sender 1"
```
zjma, line 307: @msmouse wanted me to add [this](https://github.com/aptos-labs/aptos-core/blob/0056a7b570d3e31cb0218059092cc50434b11a29/execution/executor-benchmark/src/main.rs#L192C1-L194C1) to mark `--transactions-per-sender` and `--hotspot-probability` conflicting with each other...

zjma:
```
error: the argument '--transactions-per-sender <TRANSACTIONS_PER_SENDER>' cannot be used with '--hotspot-probability <HOTSPOT_PROBABILITY>'
```

zjma: I don't know, maybe just drop the constraint and swallow it?

igor-aptos: thanks for raising!

for this PR that is ok - as test.key.executor_type == "sharded" is where we don't add --transactions-per-sender.

but will follow-up:
- is executor type a right place to switch between sharded and not sharded, i.e. can we use sharding with native executor?
- we might want to use --hotspot-probability with non-sharded workload - to have identical workload for testing, which this doesn't allow

zjma (approved):

manudhundi (approved):

PR #9858, opened 2023-08-31
Make it possible to configure which image registries to log in to

banool:
### Description
Up until now we have assumed that everyone will want to upload their images to all image registries, but that is no longer the case with some of our projects, e.g. indexer where we only upload to GCP GAR. This PR makes it possible to only login into certain repos.

### Test Plan
I manually changed `aptos-labs/aptos-core/.github/actions/docker-setup@main` to `./.github/actions/docker-setup` everywhere and changed `pull_request_target` to `pull_request` in docker-built-test.yaml. I then changed rust-images to use the workflow locally too: `./.github/workflows/workflow-run-docker-rust-build.yaml`. This means the pull_request run of rust-images used the new docker-setup workflow. See the successful run here: https://github.com/aptos-labs/aptos-core/actions/runs/6050060856/job/16418569649?pr=9858.

```
AWS args were supplied and are vaild, we will log into ECR
```
```
AWS ECR detected with us-west-2 region
Logging into https://***...
Login Succeeded!
```

This demonstrates that AWS login still works if configured. It should be pretty obvious that AWS login won't occur if those args aren't passed but if something is wrong, I'll fix it later once I use I use this in aptos-indexer-processors.

This was the commit where the test passed, in case I need to go back to it to test further: e2862aea1affbfd97662d9115a45b118956f1dbc.


.github/actions/docker-setup/action.yaml
```diff
@@ -1,31 +1,35 @@
...
+  * Logs in to docker image registries
+    * Only if the GCP args are provided will it log into GAR.
+    * Only if the AWS args are provided will it log into ECR.
   * Setup for buildx and other dependencies (crane)
   * Sets git credentials for private builds
 inputs:
   # GCP auth
   GCP_WORKLOAD_IDENTITY_PROVIDER:
-    required: true
+    required: false
```
rustielin, line 12: maybe can we default GAR on since it's our primary registry? if we ever want to push to ECR, we always push to GAR as well.

sionescu: I actually like this: we could have our private repository on GCP, and sync that to ECR somehow.

banool: Done, I made GAR required and ECR optional. Having something sync from GCP to ECR in the background sounds good, maybe easier to maintain than uploading to both.

banool:
Looks good, the 3 failures seem unrelated to this change.

larry-aptos (approved):

rustielin (approved):

PR #9859, opened 2023-08-31
[indexer-grpc] Format addresses, besides inside Move data, as per AIP-40

banool:
### Description
This is an implementation of the Transaction Stream Service section of the propssal here: https://www.notion.so/aptoslabs/Proposal-Indexer-AccountAddress-Handling-451c6e15e3b04f7697dc2b5731787cb2.

Discussion here: https://aptos-org.slack.com/archives/C03N83P7QUC/p1693475077945499.

### Test Plan
These tests were already broken so this does nothing right now, but I took a crack at fixing them. If we like this PR I can fix them all the way.
```
cd ecosystem/indexer-grpc/indexer-grpc-fullnode
cargo test
```


bowenyang007:
This will still mean that api is not padded correctly right? Also what if the address is in a struct?

bowenyang007:
Is it possible to change the way we do address.to_string? Seems like this would be the more generalized solution although idk if things like table handle is an address. I think so?

banool:
It was a tricky thing to migrate safely but I did it, to_string is now conforms to AIP-40. I can update this PR to use that instead.

PR #9864, opened 2023-08-31
Full single node performance (all tests) run on continuous

igor-aptos:
PR on top of igor/benchmark_suites (#9854)

### DIFF
--- a/.github/workflows/execution-performance.yaml
+++ b/.github/workflows/execution-performance.yaml
@@ -13,6 +13,8 @@ jobs:
         github.event_name == 'workflow_dispatch' ||
         github.event_name == 'schedule' ||
         contains(github.event.pull_request.labels.*.name, 'CICD:run-e2e-tests') ||
+        contains(github.event.pull_request.labels.*.name, 'CICD:run-execution-performance-test') ||
+        contains(github.event.pull_request.labels.*.name, 'CICD:run-execution-performance-full-test') ||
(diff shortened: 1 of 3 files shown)


sionescu (approved):

.github/workflows/execution-performance.yaml
```diff
@@ -21,4 +23,5 @@ jobs:
     with:
       GIT_SHA: ${{ github.event.pull_request.head.sha || github.sha }}
       RUNNER_NAME: executor-benchmark-runner
-      RUN_ONLY_SINGLE_NODE_PERF: ${{ github.event_name != 'schedule' }} # Run all tests on the scheduled cadence
+       # Run all tests only on the scheduled cadence, or explicitly requested
+      IS_FULL_RUN: ${{ github.event_name != 'schedule' || contains(github.event.pull_request.labels.*.name, 'CICD:run-execution-performance-full-test') }}
```
JoshLind, line 27: Silly question: should this be inverted? i.e., previously we ran the single node tests iff it wasn't a scheduled run (e.g., a PR run). But, now, we're saying, run all the tests iff it's a scheduled run, or it's been requested, right? 😄

JoshLind (approved):
LGTM other than my silly question 😄

.github/workflows/execution-performance.yaml
igor-aptos: argh, you are right :)

PR #9865, opened 2023-08-31
Fix assorted grammar errors

alnoki:


saharct (approved):

banool (approved):

developer-docs-site/docs/guides/transaction-management.md
```diff
@@ -104,4 +104,4 @@ In this model, each worker has access to the `SignerCap` of the shared account,
 
 Another model, if viable, is to decouple the `signer` altogether away from permissions and to make an application specific capability. Then this capability can be given to each worker that let’s them operate on the shared infrastructure.
 
-Note that parallelization on the shared infrastructure can be limited if any transaction would have any read, write or write, write conflicts. This won’t prevent multiple transactions from executing within a block, but can impact maximum blockchain performance.
+Note that parallelization on the shared infrastructure can be limited if any transaction would have any read or write conflicts. This won’t prevent multiple transactions from executing within a block, but can impact maximum blockchain performance.
```
xbtmatt, line 107: I actually think this slightly changed the meaning of what was said.

I think it was supposed to read like read/write and write/write.

You can't have a conflict solely based around multiple reads so I think that's why it was phrased the way it was. May need to go back and change this

PR #9869, opened 2023-08-31
[TS SDK v2] Add base BCS serialization and deserialization classes and unit tests

xbtmatt:
### Description

Adding the base BCS serialization and deserialization classes and their respective unit tests.

This is nearly exactly the same as before, with a few changes:

1. `Bytes` has been removed, we only have `Uint8Array` now
5. I've changed some of the documentation, tried to make it clearer/more intuitive

### Test Plan
`pnpm jest`

all old tests cases pass still


ecosystem/typescript/sdk_v2/src/bcs/deserializer.ts
```diff
@@ -0,0 +1,192 @@
...
+interface Deserializable<T> {
+  deserialize(deserializer: Deserializer): T;
+}
+
+export class Deserializer {
+  private buffer: ArrayBuffer;
+
+  private offset: number;
+
+  constructor(data: Uint8Array) {
```
0xjinn, line 18: Curious - do we want to use `Uint8Array` here or `HexInput`? I am guessing we will stick with `Uint8Array` for any data/buffer related?

ecosystem/typescript/sdk_v2/src/bcs/consts.ts
```diff
@@ -0,0 +1,12 @@
...
+
+import { Uint128, Uint16, Uint32, Uint64, Uint8, Uint256 } from "./types";
+
+// Upper bound values for uint8, uint16, uint64 and uint128
+export const MAX_U8_NUMBER: Uint8 = 2 ** 8 - 1;
+export const MAX_U16_NUMBER: Uint16 = 2 ** 16 - 1;
+export const MAX_U32_NUMBER: Uint32 = 2 ** 32 - 1;
+export const MAX_U64_BIG_INT: Uint64 = BigInt(2 ** 64) - BigInt(1);
+export const MAX_U128_BIG_INT: Uint128 = BigInt(2 ** 128) - BigInt(1);
+export const MAX_U256_BIG_INT: Uint256 = BigInt(2 ** 256) - BigInt(1);
```
gregnazario, line 12: How do these work, if the numbers inside aren't bigints?

ecosystem/typescript/sdk_v2/src/bcs/deserializer.ts
```diff
@@ -0,0 +1,192 @@
...
+/* eslint-disable no-bitwise */
+import { MAX_U32_NUMBER } from "./consts";
+import { Uint128, Uint16, Uint256, Uint32, Uint64, Uint8 } from "./types";
+
+// The class must implement a static deserialize method.
+interface Deserializable<T> {
+  deserialize(deserializer: Deserializer): T;
+}
+
+export class Deserializer {
```
gregnazario, line 13: We'll make a pass at the end to add docs

ecosystem/typescript/sdk_v2/src/bcs/deserializer.ts
```diff
@@ -0,0 +1,192 @@
...
+  private buffer: ArrayBuffer;
+
+  private offset: number;
+
+  constructor(data: Uint8Array) {
+    // copies data to prevent outside mutation of buffer.
+    this.buffer = new ArrayBuffer(data.length);
+    new Uint8Array(this.buffer).set(data, 0);
```
gregnazario, line 21: you don't have to set this to something?

ecosystem/typescript/sdk_v2/src/bcs/serializer.ts
```diff
@@ -0,0 +1,246 @@
...
+      value >>>= 7;
+    }
+    valueArray.push(value);
+    this.appendToBuffer(new Uint8Array(valueArray));
+  }
+
+  /**
+   * Returns the buffered bytes
+   */
+  getBytes(): Uint8Array {
```
0xjinn, line 222: Should we rename this to `toUint8Array`?

0xjinn (approved):
Just some nit comment and queston.

Otherwise, looks good!

ecosystem/typescript/sdk_v2/src/bcs/deserializer.ts
```diff
@@ -0,0 +1,192 @@
...
+   *
+   * BCS layout for "uint16": Two bytes. Binary format in little-endian representation.
+   * @example
+   * ```ts
+   * const deserializer = new Deserializer(new Uint8Array([0x34, 0x12]));
+   * assert(deserializer.deserializeU16() === 4660);
+   * ```
+   */
+  deserializeU16(): Uint16 {
+    return new DataView(this.read(2)).getUint16(0, true);
```
gregnazario, line 106: Is DataView platform dependent?  How is little endian vs big endian?

ecosystem/typescript/sdk_v2/src/bcs/serializer.ts
```diff
@@ -0,0 +1,246 @@
...
+  serialize(serializer: Serializer): void;
+}
+
+export class Serializer {
+  private buffer: ArrayBuffer;
+
+  private offset: number;
+
+  constructor() {
+    this.buffer = new ArrayBuffer(64);
```
gregnazario, line 25: Maybe we should make an optional preallocated size?

ecosystem/typescript/sdk_v2/src/bcs/serializer.ts
```diff
@@ -0,0 +1,246 @@
...
+  }
+
+  protected appendToBuffer(values: Uint8Array) {
+    this.ensureBufferWillHandleSize(values.length);
+    new Uint8Array(this.buffer, this.offset).set(values);
+    this.offset += values.length;
+  }
+
+  private serializeWithFunction(
+    fn: (byteOffset: number, value: number, littleEndian?: boolean) => void,
```
gregnazario, line 44: I think BCS is only little endian?

gregnazario (approved):

ecosystem/typescript/sdk_v2/src/bcs/deserializer.ts
xbtmatt: yeah I'll stick to `Uint8Array` for now, just to keep this all the same. I think we can address whether or not it should change in next meeting. And then address the lack of object args, too

ecosystem/typescript/sdk_v2/src/bcs/serializer.ts
xbtmatt: yeah good idea! Could be much clearer

ecosystem/typescript/sdk_v2/src/bcs/consts.ts
xbtmatt: Apparently (as you figured out and according to what I read about it) it only works because it's a floating point value internally and it happens to not lose precision because it's a power of two.

Which is why `BigInt(2 ** 256) - BigInt(1) != BigInt(2**256 - 1)`

I'll change it to `BigInt(2) ** BigInt(256)`

ecosystem/typescript/sdk_v2/src/bcs/deserializer.ts
xbtmatt: no, apparently the `new Uint8Array(this.buffer)` serves as a sort of mutable reference to the data in the `ArrayBuffer` in the heap. So `.set` directly mutates the value in `this.buffer`, changing its data to `data` starting at the byte at index `0`

ecosystem/typescript/sdk_v2/src/bcs/deserializer.ts
xbtmatt: so normally `DataView` is big-endian, actually, but the second argument in `getUint16` (in our case, `true`) is the `littleEndian` flag.

ecosystem/typescript/sdk_v2/src/bcs/serializer.ts
xbtmatt: so the reason it has this is so the function signature matches the `Dataview.setUintX` functions later:

like you noticed:

```typescript
    fn: (byteOffset: number, value: number, littleEndian?: boolean) => void,
    // ...
    fn.apply(dv, [0, value, true]);
```
but if you see this:

```typescript
  serializeU16(value: Uint16) {
    this.serializeWithFunction(DataView.prototype.setUint16, 2, value);
  }
```

the `DataView.prototype.setUint16` is going to have a function signature with the `littleEndian` flag as the 3rd arg

PR #9871, opened 2023-08-31
Minor improvement to partitioning phase in executor benchmark

sitalkedia:
### Description

- Remove `--async-partioning` flag as we default to it now and this also reduces unnecessary code duplication.
- Add stats for partitioning at the end of the log
- Reduce number of threads from 8 to 4 - this will help reduce some context switch overhead.

### DIFF
--- a/execution/block-partitioner/src/v2/config.rs
+++ b/execution/block-partitioner/src/v2/config.rs
@@ -51,7 +51,7 @@ impl PartitionerV2Config {
 impl Default for PartitionerV2Config {
     fn default() -> Self {
         Self {
-            num_threads: 8,
+            num_threads: 4,
             max_partitioning_rounds: 4,
             cross_shard_dep_avoid_threshold: 0.9,
             dashmap_num_shards: 64,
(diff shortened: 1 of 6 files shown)


grao1991 (approved):

execution/block-partitioner/src/v2/config.rs
```diff
@@ -51,7 +51,7 @@ impl PartitionerV2Config {
 impl Default for PartitionerV2Config {
     fn default() -> Self {
         Self {
-            num_threads: 8,
+            num_threads: 4,
```
zjma, line 54: I actually never tried running V2 with default config. I always use 32 threads 🤣

sitalkedia: The executor benchmark is using the default config right?

zjma: maybe. Anyway I always overwrite it... All my numbers are from runs with 32 threads.

sitalkedia: I tried it out and the partitioning performance is fine with 4 threads, this is going to reduce some thread contention so less thread is better.

zjma: ok i will also re-run with 4 threads and see if there's any TPS diff.

zjma (approved):

PR #9875, opened 2023-08-31
[pangu] expose cpu and memory to config

rustielin:
### Description

So user can specify CPU and memory of validators+VFNs+PFNs.
* for validators+VFNs, it's done via the pangu blueprint (top level config)
* for PFNs, it's done via CLI for now

### Test Plan

existing unit tests, updated fixtures. also run pangu manually and confirm updated resource request


testsuite/pangu_lib/testnet_commands/commands.py
```diff
@@ -26,8 +27,8 @@
 @click.option("--layout-path", help="Pass the path to the layout file (yaml).")
 @click.option(
     "--framework-path",
-    default=f"{util.TEMPLATE_DIRECTORY}/framework.mrb",
-    help="Pass in the path to the compiled move framework (head.mrb, or framework.mrb) file. Defaults to the default framework in the pangu_lib.",
+    required=True,
```
zekun000, line 30: it'd be good to link to an example one, or instructions how to get one. I spent some time figuring out what this is

zekun000 (approved):

testsuite/pangu_lib/testnet_commands/commands.py
rustielin: added to the help message :) 

```
$ pangu testnet create --help           
Usage: pangu.py testnet create [OPTIONS]

  Create a testnet.

Options:
...
  --framework-path TEXT           Pass in the path to the compiled move framework (head.mrb, or framework.mrb) file. To compile it, run: $
                                  cargo run --locked --package aptos-framework -- release  [required]
```

sherry-x (approved):

PR #9880, opened 2023-09-01
[GHA] Use versioned actions/checkout instead of pinning a commit

sionescu:
### Description

Trying to get rid of annoying warnings that "The `save-state` command is deprecated and will be disabled soon."

### DIFF
--- a/.github/actions/docker-setup/action.yaml
+++ b/.github/actions/docker-setup/action.yaml
@@ -74,7 +74,7 @@ runs:
 
     - id: auth
       name: "Authenticate to Google Cloud"
-      uses: "google-github-actions/auth@dac4e13deb3640f22e3ffe758fd3f95e6e89f712" # pin@v0
+      uses: "google-github-actions/auth@v1"
       with:
         create_credentials_file: false
         token_format: "access_token"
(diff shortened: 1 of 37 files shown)


banool:
Coming full circle hahah. So we pinned these in the first place to mitigate the risk of upstream security vulnerabilities coming from this actions. Could we just upgrade the pinned version?

sionescu:
The upstreams here are Github, Google, Docker and NPM so I think it's safe to say that they have good security practices and we don't need to worry about vulnerabilities.
For marocchino/sticky-pull-request-comment, I can update the pin.

rustielin (approved):
I wonder what @gedigi thinks about it, so leaving one review. I'm ok with unpinning the commit hashes for "trusted" GH orgs like you said above ^ 

Do you know which actions are actually upgraded in this process btw @sionescu ?

gedigi (approved):

gedigi:
I think that using hash values vs version numbers doesn't make us necessarily more secure, but possibly less secure. While it's true that there's a supply chain risk with bad code being pushed to a newer version of an action, it is more likely that a vulnerability is discovered in released code, and with a hash-based pin we wouldn't be automatically patched.

So for me +1 on this change.

sionescu:
Then I'll go ahead and merge this. Each "unpin" is in its own commit so if something breaks we can revert only the offending change.

PR #9881, opened 2023-09-01
[TS SDK v2] Adding the ability to serialize and deserialize classes in the base Serializer/Deserializer classes

xbtmatt:
### Description
This PR is a small QOL add to the base Serializer and Deserializer classes that adds the ability to serialize and deserialize class instances that implement `serialize` and `deserialize` respectively.

Please note the majority of the code here is unit tests/documentation- this code already existed in the sdk in the helper.ts file before in the form of `bcsToBytes` and the general usage of `const myClassName: ClassName = ClassName.deserialize(...)`, but it wasn't clear how to use these functions exactly because they weren't documented or explained.

There is definitely some test overlap here (can probably put it all into a single test case, maybe 2) but I wanted to be thorough.

```typescript
/// serializer.ts
// ...
serialize<T extends Serializable>(value: T) {
  // NOTE: The `serialize` method called by `value` is defined in `value`'s
  // Serializable interface, not the one defined in this class.
  value.serialize(this);
}

// deserializer.ts
// ...
deserialize<T>(cls: Deserializable<T>): T {
  // NOTE: The `deserialize` method called by `cls` is defined in the `cls`'s
  // Deserializable interface, not the one defined in this class.
  return cls.deserialize(this);
}
```

This is to simplify the process of serializing and deserializing non-primitive, Serializable values.

You can still serialize/deserialize these values the old way, as well.

### Test Plan
Two examples are in the unit tests for the respective files.

```shell
# this will run both serializer and deserializer unit tests
pnpm jest serializer.test.ts --verbose
```

### Notes
Note that you cannot call `deserialize` on a subclass for an abstract factory class that implements the deserialization with enum variant index. This is explained and demonstrated in the unit test.

This is also a general demonstration of the basis for the Typescript implementation of various types of BCS representations with enums.

### DIFF
--- a/ecosystem/typescript/sdk_v2/src/bcs/deserializer.ts
+++ b/ecosystem/typescript/sdk_v2/src/bcs/deserializer.ts
@@ -191,7 +191,7 @@ export class Deserializer {
 
   /**
    * This function deserializes a Deserializable value. The bytes must be loaded into the Serializer already.
-   * Note that it does NOT take in the value, it takes in the class type of the value that implements Serializable.
+   * Note that it does not take in the value, it takes in the class type of the value that implements Serializable.
    *
    * The process of using this function is as follows:
    * 1. Serialize the value of class type T using its `serialize` function.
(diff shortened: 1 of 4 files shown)


ecosystem/typescript/sdk_v2/src/bcs/deserializer.ts
```diff
@@ -5,8 +5,7 @@
 import { MAX_U32_NUMBER } from "./consts";
 import { Uint128, Uint16, Uint256, Uint32, Uint64, Uint8 } from "./types";
 
-// The class must implement a static deserialize method.
-interface Deserializable<T> {
+export interface Deserializable<T> {
```
gregnazario, line 8: Export everything!

gregnazario (approved):

0xjinn (approved):
LGTM. 

Lets get this in so I can use the new form of `bcsToBytes` 😎

ecosystem/typescript/sdk_v2/tests/unit/deserializer.test.ts
```diff
@@ -130,4 +129,187 @@ describe("BCS Deserializer", () => {
       deserializer.deserializeStr();
     }).toThrow("Reached to the end of buffer");
   });
+
+  it("deserializes a single deserializable class", () => {
```
0xmaayan, line 133: this flow feels more of an example rather than testing the serializer/deserializer classes? wdyt?

0xjinn:
Note: Merging this in right now. Will come back to revisit unit test if needed.

PR #9882, opened 2023-09-01
[Sharded-Executor]: Add support for 'total_supply' in sharded executor.

manudhundi:
We achieve this by aggregating the total_supply deltas from individual shards
and then updating the txns on all shards with the correct total supply.


sitalkedia:
Can you run some benchmark to see what is the performance implication of this change?

manudhundi:
> Can you run some benchmark to see what is the performance implication of this change?

I ran the perf tests with 50k block size, 100k accounts, 5000 connected groups. Perf is around 56.5K tps.

sitalkedia:
> > Can you run some benchmark to see what is the performance implication of this change?
> 
> I ran the perf tests with 50k block size, 100k accounts, 5000 connected groups. Perf is around 56.5K tps.

So no performance regression with this change?

manudhundi:
> > > Can you run some benchmark to see what is the performance implication of this change?
> > 
> > 
> > I ran the perf tests with 50k block size, 100k accounts, 5000 connected groups. Perf is around 56.5K tps.
> 
> So no performance regression with this change?

There is some perf regression. We were seeing 60K VM tps in this case before. That is about 6% regression (60K to 56.5K)..

aptos-move/aptos-vm/src/sharded_block_executor/local_executor_shard.rs
```diff
@@ -189,14 +189,24 @@ impl<S: StateView + Sync + Send + 'static> ExecutorClient<S> for LocalExecutorCl
...
             maybe_block_gas_limit,
         )?;
 
-        let sharded_output = self.get_output_from_shards()?;
+        let mut sharded_output = self.get_output_from_shards()?;
+
+        // We could call sharded_aggregator_service::aggregate_and_update_total_supply() from here,
+        // calling on the global shard might help us get the resources on the global shard (say
+        // like thread pool)
+        self.global_executor.aggregate_results(
```
sitalkedia, line 204: Hmm you are not using the thread pool in global shard..

aptos-move/aptos-vm/src/sharded_block_executor/sharded_aggregator_service.rs
```diff
@@ -0,0 +1,155 @@
...
+        }
+    }
+}
+
+impl ops::Add for DeltaU128 {
+    type Output = Self;
+
+    fn add(self, rhs: Self) -> Self::Output {
+        if self.is_positive == rhs.is_positive {
+            Self {
```
sitalkedia, line 72: This function's readability can be improved by reducing if nesting levels..how about returning `Self` from here and remove the else condition below?

aptos-move/aptos-vm/src/sharded_block_executor/sharded_aggregator_service.rs
```diff
@@ -0,0 +1,155 @@
...
+        .map(move |value| bcs::from_bytes(value.as_slice()));
+    value.transpose().map_err(anyhow::Error::msg).unwrap()
+}
+
+/// This class ensures that deltas can use all 128 bits without having to let go of the sign bit for
+/// cases where the delta is negative. That is, we don't have to use conversions to i128.
+/// However, it does not handle overflow and underflow. That is, it will indicate to the caller of
+/// the faulty logic with their usage of deltas.
+#[derive(Clone, Copy)]
+struct DeltaU128 {
```
sitalkedia, line 29: can you add some UTs for this?

aptos-move/aptos-vm/src/sharded_block_executor/sharded_aggregator_service.rs
```diff
@@ -0,0 +1,155 @@
...
+
+    // No need to parallelize this as the runtime is O(num_shards)
+    // TODO: Get this from the individual shards while getting 'sharded_output'
+    sharded_output
+        .iter()
+        .enumerate()
+        .for_each(|(shard_id, shard_output)| {
+            for (round, txn_outputs) in shard_output.iter().enumerate() {
+                for last_txn in txn_outputs.iter().rev() {
+                    if let Some(last_txn_total_supply) =
```
sitalkedia, line 116: This is potentially dangerous and unnecessary.. all txns should have total supply except last txn. If so, can you handle the last txn as a special case but in all other case `unwrap` for the total supply value instead of using the iterator?

aptos-move/aptos-vm/src/sharded_block_executor/sharded_aggregator_service.rs
```diff
@@ -0,0 +1,155 @@
...
+pub fn aggregate_and_update_total_supply<S: StateView>(
+    sharded_output: &mut Vec<Vec<Vec<TransactionOutput>>>,
+    global_output: &mut [TransactionOutput],
+    state_view: &S,
+) {
+    let num_shards = sharded_output.len();
+    let num_rounds = sharded_output[0].len();
+    let total_supply_base_val: u128 = get_state_value(&TOTAL_SUPPLY_STATE_KEY, state_view).unwrap();
+
+    let mut aggr_total_supply_delta = vec![DeltaU128::default(); num_shards * num_rounds + 1];
```
sitalkedia, line 106: why do you need a +1 here?

aptos-move/aptos-vm/src/sharded_block_executor/sharded_aggregator_service.rs
```diff
@@ -0,0 +1,155 @@
...
+        });
+
+    for idx in 1..aggr_total_supply_delta.len() - 1 {
+        aggr_total_supply_delta[idx + 1] =
+            aggr_total_supply_delta[idx + 1] + aggr_total_supply_delta[idx];
+    }
+
+    // Runtime is O(num_txns), hence parallelized at the shard level and at the txns level.
+    sharded_output
+        .par_iter_mut()
```
sitalkedia, line 134: Hmm.. this will use global rayon thread pool which is not ideal.. can you pass the global executor thread pool here and use that instead?

aptos-move/aptos-vm/src/sharded_block_executor/sharded_aggregator_service.rs
```diff
@@ -0,0 +1,155 @@
...
+                        txn_output.write_set().get_value(&TOTAL_SUPPLY_STATE_KEY)
+                    {
+                        txn_output.update_total_supply(delta_for_round.add_delta(total_supply));
+                    }
+                });
+            }
+        });
+
+    let delta_for_global_shard = aggr_total_supply_delta[num_shards * num_rounds];
+    global_output.par_iter_mut().for_each(|txn_output| {
```
sitalkedia, line 150: Use `with_min_len(25)` here to avoid rayon spawning too small task.. also avoid using the global rayon thread pool.

types/src/transaction/mod.rs
```diff
@@ -1057,6 +1057,10 @@ impl TransactionOutput {
         &self.write_set
     }
 
+    pub fn update_total_supply(&mut self, value: u128) {
```
sitalkedia, line 1060: Add some comment here explaining why its needed.

types/src/write_set.rs
```diff
@@ -280,6 +302,26 @@ impl WriteSetV0 {
     pub fn get(&self, key: &StateKey) -> Option<&WriteOp> {
         self.0.get(key)
     }
+
+    pub fn get_value(&self, key: &StateKey) -> Option<u128> {
```
sitalkedia, line 306: Hmm.. get_value function name is generic but it is returning `u128`, which is specific to total supply.. how about we change this to get_total_supply instead?

manudhundi: ok done

types/src/transaction/mod.rs
manudhundi: done

aptos-move/aptos-vm/src/sharded_block_executor/sharded_aggregator_service.rs
manudhundi: ok done

aptos-move/aptos-vm/src/sharded_block_executor/sharded_aggregator_service.rs
manudhundi: done and added comments

aptos-move/aptos-vm/src/sharded_block_executor/sharded_aggregator_service.rs
manudhundi: done

aptos-move/aptos-vm/src/sharded_block_executor/local_executor_shard.rs
manudhundi: we are using the thread pool in global shard as per another comment below

aptos-move/aptos-vm/src/sharded_block_executor/global_executor.rs
```diff
@@ -59,4 +60,17 @@ impl<S: StateView + Sync + Send + 'static> GlobalExecutor<S> {
             maybe_block_gas_limit,
         )
     }
+
+    pub fn aggregate_results(
```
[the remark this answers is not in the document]
manudhundi: we are using the thread pool in global shard as per another comment below. So we need to call via global executor

aptos-move/aptos-vm/src/sharded_block_executor/sharded_aggregator_service.rs
manudhundi: done

aptos-move/aptos-vm/src/sharded_block_executor/sharded_aggregator_service.rs
manudhundi: done

aptos-move/aptos-vm/src/sharded_block_executor/sharded_aggregator_service.rs
manudhundi: First value is 0 which is delta for shard 0;
+1 is the delta for global shard

aptos-move/aptos-vm/src/sharded_block_executor/cross_shard_state_view.rs
```diff
@@ -179,3 +180,55 @@ mod tests {
         wait_thread.join().unwrap();
     }
 }
+
+pub const TOTAL_SUPPLY_AGGR_BASE_VAL: u128 = u128::MAX >> 1;
+#[derive(Clone)]
+pub struct CrossShardStateViewAggrOverride<'a, S> {
```
sitalkedia, line 186: It will be better if this implementation is generic i.e, it just stores a StateView base class instead of `CrossShardStateView`, and rename this to something like `AggregatorOverriddenStateView`

aptos-move/aptos-vm/src/sharded_block_executor/cross_shard_state_view.rs
```diff
@@ -179,3 +180,55 @@ mod tests {
...
+
+impl<'a, S: StateView + Sync + Send> CrossShardStateViewAggrOverride<'a, S> {
+    pub fn new(cross_shard_keys: HashSet<StateKey>, base_view: &'a S) -> Self {
+        Self {
+            cross_shard_state_view: CrossShardStateView::new(cross_shard_keys, base_view),
+        }
+    }
+
+    pub fn set_value(&self, state_key: &StateKey, state_value: Option<StateValue>) {
+        self.cross_shard_state_view
```
sitalkedia, line 198: This really belongs in `CrossShardStateView` instead. See my comment above for making this a generic StateView that doesn't care about the `CrossShardStateView`

aptos-move/aptos-vm/src/sharded_block_executor/global_executor.rs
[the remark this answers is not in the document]
sitalkedia: Still API-wise, it doesn't belong here. You may expose an API to get the global executor thread pool here instead?

aptos-move/aptos-vm/src/sharded_block_executor/sharded_aggregator_service.rs
sitalkedia: Let's add a comment here for the same :)

aptos-move/aptos-vm/src/sharded_block_executor/sharded_aggregator_service.rs
```diff
@@ -0,0 +1,308 @@
...
+fn is_total_supply_absent(
+    total_supply: Option<u128>,
+    shard_id: usize,
+    num_shards: usize,
+    txn_idx: usize,
+    txns_len: usize,
+) -> bool {
+    // only first and last txn of the block can have 'none' for 'total_supply'
+    if total_supply.is_none()
+        && ((shard_id == 0 && txn_idx == 0) || (shard_id == num_shards && txn_idx == txns_len))
```
sitalkedia, line 175: Hmm this check doesn't look right to me.. its possible that the shard_id 0 have no txn?

aptos-move/aptos-vm/src/sharded_block_executor/sharded_aggregator_service.rs
```diff
@@ -0,0 +1,308 @@
...
+    {
+        return true;
+    }
+    false
+}
+
+fn is_total_supply_absent_global_shard(
```
sitalkedia, line 182: I am wondering instead of doing this, it will be cleaner if we just remove the first and last `TransactionOutput` from the `sharded_output`/ `global_output` and process things in parallel without having to worry about the total supply being present?

aptos-move/aptos-vm/src/sharded_block_executor/sharded_aggregator_service.rs
```diff
@@ -0,0 +1,308 @@
...
+                    }
+                }
+            });
+        }
+    });
+
+    let delta_for_global_shard = aggr_total_supply_delta[num_shards * num_rounds] + base_val_delta;
+    let delta_for_global_shard_ref = &delta_for_global_shard;
+    let global_output_len = global_output.len();
+    if global_output.len() < 25 {
```
sitalkedia, line 282: You don't need this.. just use `iter_mut`  with min length..

aptos-move/aptos-vm/src/sharded_block_executor/sharded_aggregator_service.rs
sitalkedia: You can probably do this with skip_first/last on the iterator pretty easily..

aptos-move/aptos-vm/src/sharded_block_executor/cross_shard_state_view.rs
manudhundi: done

aptos-move/aptos-vm/src/sharded_block_executor/cross_shard_state_view.rs
manudhundi: done

aptos-move/aptos-vm/src/sharded_block_executor/sharded_aggregator_service.rs
manudhundi: Ok done

aptos-move/aptos-vm/src/sharded_block_executor/sharded_aggregator_service.rs
manudhundi: As discussed offline we don't have to assert in sharded execution module that all txns except first and last need to have total_supply

aptos-move/aptos-vm/src/sharded_block_executor/sharded_aggregator_service.rs
manudhundi: As discussed offline we don't have to assert in sharded execution module that all txns except first and last need to have total_supply

aptos-move/aptos-vm/src/sharded_block_executor/aggr_overridden_state_view.rs
```diff
@@ -0,0 +1,44 @@
...
+};
+
+pub const TOTAL_SUPPLY_AGGR_BASE_VAL: u128 = u128::MAX >> 1;
+#[derive(Clone)]
+pub struct AggregatorOverriddenStateView<'a, S> {
+    base_view: &'a S,
+}
+
+impl<'a, S: StateView + Sync + Send> AggregatorOverriddenStateView<'a, S> {
+    pub fn create_cross_shard_state_view_aggr_override(base_view: &'a S) -> Self {
```
sitalkedia, line 20: nit: `create_cross_shard_state_view_aggr_override` -> `new`

sitalkedia: Also, its probably cleaner to pass the `total_supply` in the constructor and store it in the struct.

aptos-move/aptos-vm/src/sharded_block_executor/global_executor.rs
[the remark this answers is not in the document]
sitalkedia: Can you address this?

aptos-move/aptos-vm/src/sharded_block_executor/sharded_aggregator_service.rs
```diff
@@ -0,0 +1,245 @@
...
+    let mut aggr_total_supply_delta = vec![DeltaU128::default(); num_shards * num_rounds + 1];
+
+    // No need to parallelize this as the runtime is O(num_shards)
+    // TODO: Get this from the individual shards while getting 'sharded_output'
+    sharded_output
+        .iter()
+        .enumerate()
+        .for_each(|(shard_id, shard_output)| {
+            for (round, txn_outputs) in shard_output.iter().enumerate() {
+                for txn in txn_outputs.iter().rev() {
```
sitalkedia, line 185: Add some comment here explaining why we need an iterator here..

sitalkedia (approved):
LGTM % minor nits..

aptos-move/aptos-vm/src/sharded_block_executor/aggr_overridden_state_view.rs
manudhundi: done

aptos-move/aptos-vm/src/sharded_block_executor/sharded_aggregator_service.rs
manudhundi: done

aptos-move/aptos-vm/src/sharded_block_executor/sharded_aggregator_service.rs
manudhundi: done

manudhundi: done

aptos-move/aptos-vm/src/sharded_block_executor/global_executor.rs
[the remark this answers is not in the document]
manudhundi: Sorry had missed this comment in the last review, addressed it now

aptos-move/aptos-vm/src/sharded_block_executor/sharded_aggregator_service.rs
```diff
@@ -0,0 +1,251 @@
...
+    // Runtime is O(num_txns), hence parallelized at the shard level and at the txns level.
+    executor_thread_pool.scope(|_| {
+        sharded_output
+            .par_iter_mut()
+            .enumerate()
+            .for_each(|(shard_id, shard_output)| {
+                for (round, txn_outputs) in shard_output.iter_mut().enumerate() {
+                    let delta_for_round =
+                        aggr_total_supply_delta_ref[round * num_shards + shard_id] + base_val_delta;
+                    txn_outputs.par_iter_mut().for_each(|txn_output| {
```
msmouse, line 227: no `with_min_len()` here?

aptos-move/aptos-vm/src/sharded_block_executor/sharded_aggregator_service.rs
```diff
@@ -0,0 +1,251 @@
...
+    // The first element is 0, which is the delta for shard 0 in round 0. +1 element will contain
+    // the delta for the global shard
+    let mut aggr_total_supply_delta = vec![DeltaU128::default(); num_shards * num_rounds + 1];
+
+    // No need to parallelize this as the runtime is O(num_shards)
+    // TODO: Get this from the individual shards while getting 'sharded_output'
+    sharded_output
+        .iter()
+        .enumerate()
+        .for_each(|(shard_id, shard_output)| {
```
msmouse, line 185: why not iterate all the rounds in the result txn version order and gather all the diffs in one go?

(better if we create a helper function `rounds_iter_in_version_order(&mut Vec<Vec<Vec<TransactionOutput>>>) -> Iterator<Item = &Vec<>>` )

aptos-move/aptos-vm/src/sharded_block_executor/sharded_aggregator_service.rs
```diff
@@ -0,0 +1,251 @@
...
+                            DeltaU128::get_delta(last_txn_total_supply, TOTAL_SUPPLY_AGGR_BASE_VAL);
+                        break;
+                    }
+                }
+            }
+        });
+
+    for idx in 1..aggr_total_supply_delta.len() - 1 {
+        aggr_total_supply_delta[idx + 1] =
+            aggr_total_supply_delta[idx + 1] + aggr_total_supply_delta[idx];
```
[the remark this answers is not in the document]
manudhundi: Using += has some rust lang issues.
`209 |         aggr_total_supply_delta[idx + 1] += (aggr_total_supply_delta[idx]);
    |         -------------------------------------^^^^^^^^^^^^^^^^^^^^^^^------
    |         |                                    |
    |         |                                    immutable borrow occurs here
    |         mutable borrow occurs here
    |         mutable borrow later used here
    |`
    
    https://www.reddit.com/r/rust/comments/offqy6/cannot_borrow_self_as_mutable_because_it_is_also/

I will have to use another local variable to get around this, and that will require 2 lines of code instead of 1

aptos-move/aptos-vm/src/sharded_block_executor/sharded_aggregator_service.rs
manudhundi: good catch, will use `with_min_len()` here

types/src/write_set.rs
```diff
@@ -280,6 +302,26 @@ impl WriteSetV0 {
...
+            .map(bcs::from_bytes);
+        value.transpose().map_err(anyhow::Error::msg).unwrap()
+    }
+
+    // This is a temporary method to update the total supply in the write set.
+    // TODO: get rid of this func() and use WriteSetMut instead; for that we need to change
+    //       VM execution such that to 'TransactionOutput' is materialized after updating
+    //       total_supply.
+    pub fn update_total_supply(&mut self, value: u128) {
+        self.0.write_set.insert(
```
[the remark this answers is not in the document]
manudhundi: We do this in `sharded_aggregator_service` where the function is called.
`if let Some(txn_total_supply) = txn_output.write_set().get_total_supply() {
                            txn_output
                                .update_total_supply(delta_for_round.add_delta(txn_total_supply));
                        }`

aptos-move/aptos-vm/src/sharded_block_executor/sharded_aggregator_service.rs
manudhundi: We are iterating such that we minimize going across shards. So once we go to a shard we iterate over all the rounds over there. 

If we iterate for each round first, and then every shard in each round, we could do the read and update of total supply in one pass over the Vec<Vec<Vec<TransactionOutput>>>, but it will require switching across shards for every round, which is costly since we eventually want shards across different nodes

types/src/write_set.rs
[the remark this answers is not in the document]
msmouse: I mean you can still assert inside here to defend against accidental changes outside. 
	`assert!(self.0.write_set.insert(...))` is all

aptos-move/aptos-vm/src/sharded_block_executor/sharded_aggregator_service.rs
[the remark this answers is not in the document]
msmouse: :trollface:  I'd prefer two lines but doesn't matter much.

aptos-move/aptos-vm/src/sharded_block_executor/sharded_aggregator_service.rs
msmouse: I don't quite buy this argument but my brain capacity (barely) works with two loops as well :trollface:

msmouse (approved):

msmouse (approved):

types/src/write_set.rs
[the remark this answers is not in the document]
manudhundi: oh ok, done

aptos-move/aptos-vm/src/sharded_block_executor/sharded_aggregator_service.rs
```diff
@@ -0,0 +1,251 @@
...
+        .map(move |value| bcs::from_bytes(value.as_slice()));
+    value.transpose().map_err(anyhow::Error::msg).unwrap()
+}
+
+/// This class ensures that deltas can use all 128 bits without having to let go of the sign bit for
+/// cases where the delta is negative. That is, we don't have to use conversions to i128.
+/// However, it does not handle overflow and underflow. That is, it will indicate to the caller of
+/// the faulty logic with their usage of deltas.
+#[derive(Clone, Copy)]
+struct DeltaU128 {
```
igor-aptos, line 30: We have almost identical thing in aggregators.

https://github.com/aptos-labs/aptos-core/pull/9368/files#diff-87ad56dd16ac7e11a72f7798a0f39b4428b59c7090b93eafe2b4ee98e13d02e7


We should consolidate. I suggest naming it SignedU128, with signature:

pub enum SignedU128 {
    Positive(u128),
    Negative(u128),
}

and appropriate methods returning BoundedMathResult<SignedU128> to indicate overflow and underflow, which you can call expect() on if you want to throw.

cc @vusirikala

aptos-move/aptos-vm/src/sharded_block_executor/sharded_aggregator_service.rs
```diff
@@ -0,0 +1,251 @@
...
+                    if let Some(last_txn_total_supply) = txn.write_set().get_total_supply() {
+                        aggr_total_supply_delta[round * num_shards + shard_id + 1] =
+                            DeltaU128::get_delta(last_txn_total_supply, TOTAL_SUPPLY_AGGR_BASE_VAL);
+                        break;
+                    }
+                }
+            }
+        });
+
+    for idx in 1..aggr_total_supply_delta.len() - 1 {
```
igor-aptos, line 201: is the order of versions also exactly like this? i.e. all transactions in shard 0 go before transactions in shard 1, which go before transactions in shard 2, etc?

aptos-move/aptos-vm/src/sharded_block_executor/sharded_aggregator_service.rs
manudhundi: As discussed, changed it to iterate round-wise for better readability

aptos-move/aptos-vm/src/sharded_block_executor/sharded_aggregator_service.rs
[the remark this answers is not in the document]
manudhundi: As per the change from addressing another comment, we now don't need '+=' operator

aptos-move/aptos-vm/src/sharded_block_executor/sharded_aggregator_service.rs
manudhundi: Then order is [round1_shard1], [round1_shard2], ... [round1_shardN], [round2_shard1], ...... [roundR_shardN]

aptos-move/aptos-vm/src/sharded_block_executor/sharded_aggregator_service.rs
manudhundi: Igor, yeah we should consolidate. Can I merge this change now though, and then consolidate it after the other PR is merged ? (I see that the other PR is still being worked upon)

PR #9883, opened 2023-09-01
Remove sharded block partitioner v1 and make v2 default

sitalkedia:
### Description

As above


execution/block-partitioner/src/tests.rs
```diff
@@ -0,0 +1,356 @@
+// Copyright © Aptos Foundation
```
zjma, line 1: i'm curious can V2 pass these test cases without any change to V2/test cases? V2 doesn't reorder by sender at the beginning and I was expecting some V1-specific tests will fail...

sitalkedia: I removed all sender specific tests.. all other tests are generic and they are passing on V2

execution/executor-benchmark/src/main.rs
```diff
@@ -132,22 +127,13 @@ impl PipelineOpt {
...
-                partition_last_round: self.use_global_executor,
-            }),
-            _ => panic!("Unknown partitioner version: {}", self.partitioner_version),
-        }
+        PartitionerConfig::V2(PartitionerV2Config {
+            num_threads: self.partitioner_v2_num_threads,
+            max_partitioning_rounds: self.max_partitioning_rounds,
+            cross_shard_dep_avoid_threshold: self.partitioner_cross_shard_dep_avoid_threshold,
+            dashmap_num_shards: self.partitioner_v2_dashmap_num_shards,
+            partition_last_round: self.use_global_executor,
```
zjma, line 135:
```suggestion
            partition_last_round: !self.use_global_executor,
```

manudhundi (commented):
If we are getting rid of sharded block partitioner and making v2 default, do we still call it v2 ?
That is, we don't have any v1 or v3, so we could perhaps have name for it ?

sitalkedia:
> If we are getting rid of sharded block partitioner and making v2 default, do we still call it v2 ? That is, we don't have any v1 or v3, so we could perhaps have name for it ?

Yes, I almost renamed everything but then realized that the rename is going to create a nightmare for @zjma to merge his changes. We can rename it post merging his pending changes :)

zjma:
> > If we are getting rid of sharded block partitioner and making v2 default, do we still call it v2 ? That is, we don't have any v1 or v3, so we could perhaps have name for it ?
> 
> Yes, I almost renamed everything but then realized that the rename is going to create a nightmare for @zjma to merge his changes. We can rename it post merging his pending changes :)

Yup, renaming can be taken care of in my PR.

zjma (approved):

grao1991 (approved):

PR #9884, opened 2023-09-01
Fix Python example tests

banool:
### Description
Odds are this is broken because of the new from_str handling. Not sure how this slipped through though.

The error was coming from the fact that with strict parsing, `b0b` is not a valid address string. I fix it by changing the example address we use to `0xf`.

### Test Plan
```
$ make test
poetry run python -m unittest discover -s aptos_sdk/ -p '*.py' -t ..
.....................................
----------------------------------------------------------------------
Ran 37 tests in 0.395s

OK
```

### DIFF
--- a/ecosystem/python/sdk/aptos_sdk/account_sequence_number.py
+++ b/ecosystem/python/sdk/aptos_sdk/account_sequence_number.py
@@ -178,7 +178,7 @@ class Test(unittest.IsolatedAsyncioTestCase):
 
         rest_client = RestClient("https://fullnode.devnet.aptoslabs.com/v1")
         account_sequence_number = AccountSequenceNumber(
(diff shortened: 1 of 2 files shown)


ecosystem/python/sdk/aptos_sdk/account_sequence_number.py
```diff
@@ -178,7 +178,7 @@ async def test_common_path(self):
 
         rest_client = RestClient("https://fullnode.devnet.aptoslabs.com/v1")
         account_sequence_number = AccountSequenceNumber(
-            rest_client, AccountAddress.from_str("b0b")
```
0xjinn, line 181: Interesting, i wonder how was this passing from the beginning lol

0xjinn (approved):

gregnazario (approved):

Issue #9878, opened 2023-08-31
[Bug] Local transaction stream docker compose fails due to redis


alnoki:
@bowenyang007 

# Background

System: 2021 Apple M1 Max 32 GB Memory

Docs page: https://aptos.dev/indexer/txn-stream/local-development

# Error

Following the instructions on the docs page:

```zsh
# From inside aptos-core/testsuite
poetry install
DOCKER_DEFAULT_PLATFORM=linux/amd64
export DOCKER_DEFAULT_PLATFORM
poetry run python indexer_grpc_local.py start
```

yields

```
No dependencies to install or update
[Thu, 31 Aug 2023 16:50:22] INFO [indexer_grpc_local.py.run_docker_compose:68] Running docker-compose up on docker/compose/validator-testnet/docker-compose.yaml
 Container validator-testnet-validator-1  Running
 Container validator-testnet-faucet-1  Running
[Thu, 31 Aug 2023 16:50:22] INFO [indexer_grpc_local.py.start:216] TESTNET STARTED: latest version @ 940
[Thu, 31 Aug 2023 16:50:22] INFO [indexer_grpc_local.py.run_docker_compose:68] Running docker-compose up on docker/compose/indexer-grpc/docker-compose.yaml
 Container indexer-grpc-redis-1  Creating
Error response from daemon: image with reference arm64v8/redis:6.2 was found but does not match the specified platform: wanted linux/amd64, actual: linux/arm64/v8
[Thu, 31 Aug 2023 16:50:22] INFO [indexer_grpc_local.py.wait_for_indexer_grpc_progress:150] Waiting for indexer grpc to start for 60s
[Thu, 31 Aug 2023 16:50:22] INFO [indexer_grpc_local.py.wait_for_indexer_grpc_progress:163] Exception: HTTPConnectionPool(host='localhost', port=18084): Max retries exceeded with url: /readiness (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x102076210>: Failed to establish a new connection: [Errno 61] Connection refused'))
```

Per above, it appears that there is a redis image issue

Hence, switching default platform to `linux/amd64/v8`:

```zsh
# From inside aptos-core/testsuite
poetry install
DOCKER_DEFAULT_PLATFORM=linux/amd64/v8
export DOCKER_DEFAULT_PLATFORM
poetry run python indexer_grpc_local.py start
```

```
No dependencies to install or update
[...]

PR #9885, opened 2023-09-01
Handle platform on user's behalf in indexer_grpc_local.py

banool:
### Description
This PR makes a variety of improvements to indexer_grpc_local.py:
- The script now does a series of checks to ensure that the environment is correct, such as whether docker is present, running, has docker compose, etc. and whether grpcurl is installed.
- The script already tried to use the correct Redis image based on the machine, but this was actually incorrect for x86 Macs. It now handles this correctly.
- It also sets DOCKER_DEFAULT_PLATFORM based on the OS + CPU architecture so the user shouldn't have to worry about it.
- It is now possible to use this script on Linux where you can only use Docker as root by setting the `--run-docker-as-root` flag.
- We use `docker compose` rather than `docker-compose`.
- I remove the `timeout` dependency and handle it in Python instead. As part of this I simplify the run functions. I'm pretty sure they're correct but let's see how it looks in CI.
- Some of the argument parsing has been simplified.

I've also updated the docs:
- Updated the prereqs section.
- Rearranged some of the debugging stuff.
- Added more guidance for making sure the script works on ARM MacOS.

### Test Plan
I tested this manually on both ARM MacOS and x86 Linux.

Unit tests:
```
$ poetry run python -m unittest indexer_grpc_local_test.py
.
----------------------------------------------------------------------
Ran 1 test in 0.000s

OK
```


banool:
Addresses https://github.com/aptos-labs/aptos-core/issues/9878.

banool:
Enabling auto merge to get some of the tests to run.

rustielin (approved):
LGTM! Grpc integration and e2e tests look good 

Looks a lot cleaner now -- thanks also for updating the docs and utils

testsuite/indexer_grpc_local.py
```diff
@@ -172,106 +182,175 @@ def wait_for_indexer_grpc_progress(shell: Shell, client: HttpClient) -> None:
...
-            res.exit_code == 124
-        ):  # timeout exits with 124 if it reaches the end of the timeout
+        try:
+            res = context.shell.run(
+                [
+                    GRPCURL_PATH,
+                    "-max-msg-sz",
+                    "10000000",
+                    "-d",
+                    '{ "starting_version": 0 }',
```
larry-aptos, line 192: optional: currently data service supports finite stream, i.e., `transactions_count`. The TimeoutExpired exception may indicate other issues if the finite stream is expected to end already.

larry-aptos (approved):

testsuite/indexer_grpc_local.py
banool: Okay good to know. I'll leave that as an optional improvement for now.

PR #9888, opened 2023-09-01
Update scheduler.rs inconsistent comments

so-schen:
inconsistent comments vs. state machine
and the code in `resume()` also did not increment the incarnation number


gelash (approved):
Nice catch, I think some earlier version used to preempt the waiting txn and then re-execute with incremented incarnation number, but now it's waiting and the same incarnation. The diagrams seem up-to-date.
@danielxiangzl @ziaptos to confirm

so-schen:
Any update?

danielxiangzl (approved):
Good catch! Thanks

PR #9889, opened 2023-09-01
[SDKv2] Migrate crypto classes over to sdk v2

0xjinn:
### Description

Everything stay the same, except the use of following

1. Bytes -> UInt8Array
2. Hex for key value
3. HexInput for hex input param

### Test Plan

pnpm jest

```
Test Suites: 9 passed, 9 total
Tests:       127 passed, 127 total
Snapshots:   0 total
Time:        2.352 s
```

### DIFF
--- a/ecosystem/typescript/sdk_v2/src/crypto/authentication_key.ts
+++ b/ecosystem/typescript/sdk_v2/src/crypto/authentication_key.ts
@@ -15,7 +15,7 @@ import { Ed25519PublicKey } from "./ed25519";
  * Account addresses can be derived from AuthenticationKey
  */
 export class AuthenticationKey {
-  // Length of AuthenticationKey in bytes(UInt8Array)
+  // Length of AuthenticationKey in bytes(Uint8Array)
   static readonly LENGTH: number = 32;
 
   // Scheme identifier for MultiEd25519 signatures used to derive authentication keys for MultiEd25519 public keys
@@ -28,18 +28,23 @@ export class AuthenticationKey {
   // authentication key) of a resource account.
   static readonly DERIVE_RESOURCE_ACCOUNT_SCHEME: number = 255;
 
-  private readonly _data: Hex;
+  public readonly data: Hex;
 
-  constructor(hexInput: HexInput) {
-    const hex = Hex.fromHexInput({ hexInput });
+  constructor(args: { data: HexInput }) {
+    const { data } = args;
+    const hex = Hex.fromHexInput({ hexInput: data });
     if (hex.toUint8Array().length !== AuthenticationKey.LENGTH) {
       throw new Error(`Authentication Key length should be ${AuthenticationKey.LENGTH}`);
     }
-    this._data = hex;
+    this.data = hex;
   }
 
-  get data(): Hex {
-    return this._data;
+  toString(): string {
+    return this.data.toString();
+  }
+
+  toUint8Array(): Uint8Array {
+    return this.data.toUint8Array();
   }
 
   /**
@@ -57,7 +62,7 @@ export class AuthenticationKey {
     const hash = sha3Hash.create();
     hash.update(bytes);
 
-    return new AuthenticationKey(hash.digest());
+    return new AuthenticationKey({ data: hash.digest() });
   }
 
   static fromEd25519PublicKey(publicKey: Ed25519PublicKey): AuthenticationKey {
@@ -70,7 +75,7 @@ export class AuthenticationKey {
     const hash = sha3Hash.create();
     hash.update(bytes);
 
-    return new AuthenticationKey(hash.digest());
+    return new AuthenticationKey({ data: hash.digest() });
   }
 
   /**
--- a/ecosystem/typescript/sdk_v2/src/crypto/authentication_key.ts
+++ b/ecosystem/typescript/sdk_v2/src/crypto/authentication_key.ts
@@ -1,0 +1,73 @@
+// Copyright © Aptos Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+import { sha3_256 as sha3Hash } from "@noble/hashes/sha3";
+import { AccountAddress, Hex } from "../core";
+import { HexInput } from "../types";
+import { MultiEd25519PublicKey } from "./multi_ed25519";
+import { Ed25519PublicKey } from "./ed25519";
+
+/**
+ * Each account stores an authentication key. Authentication key enables account owners to rotate
+ * their private key(s) associated with the account without changing the address that hosts their account.
+ * @see {@link * https://aptos.dev/concepts/accounts | Account Basics}
+ *
+ * Account addresses can be derived from AuthenticationKey
+ */
+export class AuthenticationKey {
+  static readonly LENGTH: number = 32;
+
+  static readonly MULTI_ED25519_SCHEME: number = 1;
+
+  static readonly ED25519_SCHEME: number = 0;
+
+  static readonly DERIVE_RESOURCE_ACCOUNT_SCHEME: number = 255;
+
+  readonly data: Hex;
+
+  constructor(hexInput: HexInput) {
+    if (hexInput.length !== AuthenticationKey.LENGTH) {
+      throw new Error("Expected a byte array of length 32");
+    }
+    this.data = Hex.fromHexInput({ hexInput });
+  }
+
+  /**
+   * Converts a K-of-N MultiEd25519PublicKey to AuthenticationKey with:
+   * `auth_key = sha3-256(p_1 | … | p_n | K | 0x01)`. `K` represents the K-of-N required for
+   * authenticating the transaction. `0x01` is the 1-byte scheme for multisig.
+   */
+  static fromMultiEd25519PublicKey(publicKey: MultiEd25519PublicKey): AuthenticationKey {
+    const pubKeyBytes = publicKey.toUint8Array();
+
+    const bytes = new Uint8Array(pubKeyBytes.length + 1);
+    bytes.set(pubKeyBytes);
+    bytes.set([AuthenticationKey.MULTI_ED25519_SCHEME], pubKeyBytes.length);
+
+    const hash = sha3Hash.create();
+    hash.update(bytes);
+
+    return new AuthenticationKey(hash.digest());
+  }
+
+  static fromEd25519PublicKey(publicKey: Ed25519PublicKey): AuthenticationKey {
+    const pubKeyBytes = publicKey.value.toUint8Array();
+
+    const bytes = new Uint8Array(pubKeyBytes.length + 1);
+    bytes.set(pubKeyBytes);
+    bytes.set([AuthenticationKey.ED25519_SCHEME], pubKeyBytes.length);
+
+    const hash = sha3Hash.create();
+    hash.update(bytes);
+
+    return new AuthenticationKey(hash.digest());
+  }
+
+  /**
+   * Derives an account address from AuthenticationKey. Since current AccountAddress is 32 bytes,
+   * AuthenticationKey bytes are directly translated to AccountAddress.
+   */
+  derivedAddress(): AccountAddress {
+    return new AccountAddress({ data: this.data.toUint8Array() });
+  }
+}
--- a/ecosystem/typescript/sdk_v2/src/crypto/authentication_key.ts
+++ b/ecosystem/typescript/sdk_v2/src/crypto/authentication_key.ts
@@ -26,10 +26,11 @@ export class AuthenticationKey {
   readonly data: Hex;
 
   constructor(hexInput: HexInput) {
-    if (hexInput.length !== AuthenticationKey.LENGTH) {
-      throw new Error("Expected a byte array of length 32");
+    const hex = Hex.fromHexInput({ hexInput });
+    if (hex.toUint8Array().length !== AuthenticationKey.LENGTH) {
+      throw new Error("Expected a hexinput of length 32");
     }
-    this.data = Hex.fromHexInput({ hexInput });
+    this.data = hex;
   }
 
   /**
--- a/ecosystem/typescript/sdk_v2/src/crypto/authentication_key.ts
+++ b/ecosystem/typescript/sdk_v2/src/crypto/authentication_key.ts
@@ -15,22 +15,31 @@ import { Ed25519PublicKey } from "./ed25519";
  * Account addresses can be derived from AuthenticationKey
  */
 export class AuthenticationKey {
+  // Length of AuthenticationKey in bytes(UInt8Array)
   static readonly LENGTH: number = 32;
 
+  // Scheme identifier for MultiEd25519 signatures used to derive authentication keys for MultiEd25519 public keys
   static readonly MULTI_ED25519_SCHEME: number = 1;
 
+  // Scheme identifier for Ed25519 signatures used to derive authentication key for MultiEd25519 public key
   static readonly ED25519_SCHEME: number = 0;
 
+  // Scheme identifier used when hashing an account's address together with a seed to derive the address (not the
+  // authentication key) of a resource account.
   static readonly DERIVE_RESOURCE_ACCOUNT_SCHEME: number = 255;
 
-  readonly data: Hex;
+  private readonly _data: Hex;
 
   constructor(hexInput: HexInput) {
     const hex = Hex.fromHexInput({ hexInput });
     if (hex.toUint8Array().length !== AuthenticationKey.LENGTH) {
-      throw new Error("Expected a hexinput of length 32");
+      throw new Error(`Authentication Key length should be ${AuthenticationKey.LENGTH}`);
     }
-    this.data = hex;
+    this._data = hex;
+  }
+
+  get data(): Hex {
+    return this._data;
   }
 
   /**
--- a/ecosystem/typescript/sdk_v2/src/crypto/ed25519.ts
+++ b/ecosystem/typescript/sdk_v2/src/crypto/ed25519.ts
@@ -1,0 +1,51 @@
+// Copyright © Aptos Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+import { Deserializer, Serializer } from "../bcs";
+import { Hex } from "../core";
+import { HexInput } from "../types";
+
+export class Ed25519PublicKey {
+  static readonly LENGTH: number = 32;
+
+  readonly value: Hex;
+
+  constructor(value: HexInput) {
+    if (value.length !== Ed25519PublicKey.LENGTH) {
+      throw new Error(`Ed25519PublicKey length should be ${Ed25519PublicKey.LENGTH}`);
+    }
+    this.value = Hex.fromHexInput({ hexInput: value });
+  }
+
+  toUint8Array(): Uint8Array {
+    return this.value.toUint8Array();
+  }
+
+  serialize(serializer: Serializer): void {
+    serializer.serializeBytes(this.value.toUint8Array());
+  }
+
+  static deserialize(deserializer: Deserializer): Ed25519PublicKey {
+    const value = deserializer.deserializeBytes();
+    return new Ed25519PublicKey(value);
+  }
+}
+
+export class Ed25519Signature {
+  static readonly LENGTH = 64;
+
+  constructor(public readonly value: Uint8Array) {
+    if (value.length !== Ed25519Signature.LENGTH) {
+      throw new Error(`Ed25519Signature length should be ${Ed25519Signature.LENGTH}`);
+    }
+  }
+
+  serialize(serializer: Serializer): void {
+    serializer.serializeBytes(this.value);
+  }
+
+  static deserialize(deserializer: Deserializer): Ed25519Signature {
+    const value = deserializer.deserializeBytes();
+    return new Ed25519Signature(value);
+  }
+}
--- a/ecosystem/typescript/sdk_v2/src/crypto/ed25519.ts
+++ b/ecosystem/typescript/sdk_v2/src/crypto/ed25519.ts
@@ -35,10 +35,13 @@ export class Ed25519PublicKey {
 export class Ed25519Signature {
   static readonly LENGTH = 64;
 
-  constructor(public readonly value: Uint8Array) {
+  public readonly value: Uint8Array;
+
+  constructor(value: Uint8Array) {
     if (value.length !== Ed25519Signature.LENGTH) {
       throw new Error(`Ed25519Signature length should be ${Ed25519Signature.LENGTH}`);
     }
+    this.value = value;
   }
 
   serialize(serializer: Serializer): void {
--- a/ecosystem/typescript/sdk_v2/src/crypto/ed25519.ts
+++ b/ecosystem/typescript/sdk_v2/src/crypto/ed25519.ts
@@ -6,24 +6,30 @@ import { Hex } from "../core";
 import { HexInput } from "../types";
 
 export class Ed25519PublicKey {
+  // Correct length of the public key in bytes (Uint8Array)
   static readonly LENGTH: number = 32;
 
-  readonly value: Hex;
+  // The public key in hex format
+  readonly data: Hex;
 
   constructor(hexInput: HexInput) {
     const hex = Hex.fromHexInput({ hexInput });
     if (hex.toUint8Array().length !== Ed25519PublicKey.LENGTH) {
       throw new Error(`Ed25519PublicKey length should be ${Ed25519PublicKey.LENGTH}`);
     }
-    this.value = hex;
+    this.data = hex;
   }
 
   toUint8Array(): Uint8Array {
-    return this.value.toUint8Array();
+    return this.data.toUint8Array();
+  }
+
+  toString(): string {
+    return this.data.toString();
   }
 
   serialize(serializer: Serializer): void {
-    serializer.serializeBytes(this.value.toUint8Array());
+    serializer.serializeBytes(this.data.toUint8Array());
   }
 
   static deserialize(deserializer: Deserializer): Ed25519PublicKey {
@@ -35,17 +41,17 @@ export class Ed25519PublicKey {
 export class Ed25519Signature {
   static readonly LENGTH = 64;
 
-  public readonly value: Uint8Array;
+  public readonly data: Uint8Array;
 
   constructor(value: Uint8Array) {
     if (value.length !== Ed25519Signature.LENGTH) {
       throw new Error(`Ed25519Signature length should be ${Ed25519Signature.LENGTH}`);
     }
-    this.value = value;
+    this.data = value;
   }
 
   serialize(serializer: Serializer): void {
-    serializer.serializeBytes(this.value);
+    serializer.serializeBytes(this.data);
   }
 
   static deserialize(deserializer: Deserializer): Ed25519Signature {
--- a/ecosystem/typescript/sdk_v2/src/crypto/ed25519.ts
+++ b/ecosystem/typescript/sdk_v2/src/crypto/ed25519.ts
@@ -10,11 +10,12 @@ export class Ed25519PublicKey {
 
   readonly value: Hex;
 
-  constructor(value: HexInput) {
-    if (value.length !== Ed25519PublicKey.LENGTH) {
+  constructor(hexInput: HexInput) {
+    const hex = Hex.fromHexInput({ hexInput });
+    if (hex.toUint8Array().length !== Ed25519PublicKey.LENGTH) {
       throw new Error(`Ed25519PublicKey length should be ${Ed25519PublicKey.LENGTH}`);
     }
-    this.value = Hex.fromHexInput({ hexInput: value });
+    this.value = hex;
   }
 
   toUint8Array(): Uint8Array {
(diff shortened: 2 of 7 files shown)


ecosystem/typescript/sdk_v2/src/crypto/ed25519.ts
```diff
@@ -0,0 +1,51 @@
...
+export class Ed25519PublicKey {
+  static readonly LENGTH: number = 32;
+
+  readonly value: Hex;
+
+  constructor(value: HexInput) {
+    if (value.length !== Ed25519PublicKey.LENGTH) {
+      throw new Error(`Ed25519PublicKey length should be ${Ed25519PublicKey.LENGTH}`);
+    }
+    this.value = Hex.fromHexInput({ hexInput: value });
```
0xjinn, line 17: Not sure if its just me. I found the obj for a single arg is not very user friendly

xbtmatt: Yeah not a fan. I'd be much happier with no object for single arg, object args for 2+ args, personally

ecosystem/typescript/sdk_v2/src/bcs/types.ts
```diff
@@ -1,6 +1,8 @@
 // Copyright © Aptos Foundation
 // SPDX-License-Identifier: Apache-2.0
 
+export type Seq<T> = T[];
```
xbtmatt, line 4: So I removed this so we don't have any aliased types. Same as `Bytes` <=> `Uint8Array`

Not sure if we want to go either way with it. Personally I like the way `Seq<T>` looks, so I see why jijun originally did it, but it might just confuse people 🤷

xbtmatt:
This is mostly the same right? Looks like nothing really changed, just checking

0xjinn:
> This is mostly the same right? Looks like nothing really changed, just checking

Yeah nothing really changed, except the key representation as `Hex`

ecosystem/typescript/sdk_v2/src/bcs/types.ts
0xjinn: I had the same thoughts. The Seq<> does looks a bit cleaner, but not sure if this is necessary. We can discuss this offline

xbtmatt (approved):

0xjinn:
Note: Removed `Seq<T> = T[]` alias. I don't think we need this in sdkv2

ecosystem/typescript/sdk_v2/src/crypto/authentication_key.ts
```diff
@@ -0,0 +1,73 @@
...
+  static readonly MULTI_ED25519_SCHEME: number = 1;
+
+  static readonly ED25519_SCHEME: number = 0;
+
+  static readonly DERIVE_RESOURCE_ACCOUNT_SCHEME: number = 255;
+
+  readonly data: Hex;
+
+  constructor(hexInput: HexInput) {
+    if (hexInput.length !== AuthenticationKey.LENGTH) {
```
0xmaayan, line 29: HexInput is of type `string | Uint8Array`, does the length check make sense for both types?

0xmaayan (changes requested):
can we add tests for each class?

ecosystem/typescript/sdk_v2/src/crypto/authentication_key.ts
0xjinn: Ahh good catch! I will change this to `Unint8Array` before checking for length

0xmaayan (commented):
any reason we dont add tests for ed and multi_ed classes?

0xjinn:
> any reason we dont add tests for ed and multi_ed classes?

@0xmaayan  Actually I have it in my local, but the multi_ed unit test will require `bcsToBytes`. Waiting for #9881 to land 😄

0xmaayan:
> > any reason we dont add tests for ed and multi_ed classes?
> 
> @0xmaayan Actually I have it in my local, but the multi_ed unit test will require `bcsToBytes`. Waiting for #9881 to land 😄

gotcha! sounds good thanks!

ecosystem/typescript/sdk_v2/src/crypto/authentication_key.ts
```diff
@@ -0,0 +1,74 @@
...
+  static readonly ED25519_SCHEME: number = 0;
+
+  static readonly DERIVE_RESOURCE_ACCOUNT_SCHEME: number = 255;
+
+  readonly data: Hex;
+
+  constructor(hexInput: HexInput) {
+    const hex = Hex.fromHexInput({ hexInput });
+    if (hex.toUint8Array().length !== AuthenticationKey.LENGTH) {
+      throw new Error("Expected a hexinput of length 32");
```
0xmaayan, line 31: nit
```suggestion
      throw new Error(`Authentication Key length should be ${AuthenticationKey.LENGTH)}`);
```

ecosystem/typescript/sdk_v2/src/crypto/authentication_key.ts
```diff
@@ -0,0 +1,74 @@
...
+ * Each account stores an authentication key. Authentication key enables account owners to rotate
+ * their private key(s) associated with the account without changing the address that hosts their account.
+ * @see {@link * https://aptos.dev/concepts/accounts | Account Basics}
+ *
+ * Account addresses can be derived from AuthenticationKey
+ */
+export class AuthenticationKey {
+  static readonly LENGTH: number = 32;
+
+  static readonly MULTI_ED25519_SCHEME: number = 1;
```
0xmaayan, line 20: could we add comments explaining each static fields?

ecosystem/typescript/sdk_v2/src/crypto/authentication_key.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  static readonly ED25519_SCHEME: number = 0;
+
+  static readonly DERIVE_RESOURCE_ACCOUNT_SCHEME: number = 255;
+
+  readonly data: Hex;
```
0xmaayan, line 26:
```suggestion
  private readonly data: Hex;
```

ecosystem/typescript/sdk_v2/tests/unit/authentication_key.test.ts
```diff
@@ -0,0 +1,43 @@
...
+import { AuthenticationKey } from "../../src/crypto/authentication_key";
+import { Ed25519PublicKey } from "../../src/crypto/ed25519";
+import { MultiEd25519PublicKey } from "../../src/crypto/multi_ed25519";
+
+// Auth key and Public key pair for testing
+const auth_key_hexInput = "0x6324287105756b0338e0f84025bd0ac80e58154eb94257b0d4f06ec6497e656e";
+const public_key = "0x719ab6a6d406931ca80efa922e3377390a8d2803e42ecdbf394e979f9a5e57bc";
+
+describe("AuthenticationKey", () => {
+  it("should create an instance with save the hexinput correctly", () => {
```
0xmaayan, line 13:
```suggestion
  it("should create an instance with provided hexinput correctly", () => {
```

ecosystem/typescript/sdk_v2/tests/unit/authentication_key.test.ts
```diff
@@ -0,0 +1,43 @@
...
+    const authKey = new AuthenticationKey(auth_key_hexInput);
+    expect(authKey).toBeInstanceOf(AuthenticationKey);
+    expect(authKey.data.toString()).toEqual(auth_key_hexInput);
```
0xmaayan, line 16: can we add
`expect(authKey.data).toBeInstanceOf(Hex);`

ecosystem/typescript/sdk_v2/tests/unit/authentication_key.test.ts
```diff
@@ -0,0 +1,43 @@
...
+    const authKey = AuthenticationKey.fromEd25519PublicKey(publicKey);
+    expect(authKey).toBeInstanceOf(AuthenticationKey);
+    expect(authKey.data.toString()).toEqual(auth_key_hexInput);
+  });
+
+  it("should create AuthenticationKey from MultiEd25519PublicKey", () => {
+    const publicKey = new MultiEd25519PublicKey([new Ed25519PublicKey(public_key)], 1);
+    const authKey = AuthenticationKey.fromMultiEd25519PublicKey(publicKey);
+    expect(authKey).toBeInstanceOf(AuthenticationKey);
+    expect(authKey.data.toString()).toEqual("0xd0cb1ed17413857dd59f4ee948b6678c339ad6d5cc97246f65825439f53fd944");
```
0xmaayan, line 35: can we store the expected auth key in a variable and compare to it?

ecosystem/typescript/sdk_v2/src/crypto/ed25519.ts
```diff
@@ -0,0 +1,52 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+import { Deserializer, Serializer } from "../bcs";
+import { Hex } from "../core";
+import { HexInput } from "../types";
+
+export class Ed25519PublicKey {
+  static readonly LENGTH: number = 32;
+
+  readonly value: Hex;
```
0xmaayan, line 11:
```suggestion
  private readonly value: Hex;
```

ecosystem/typescript/sdk_v2/src/crypto/ed25519.ts
```diff
@@ -0,0 +1,52 @@
...
+  static deserialize(deserializer: Deserializer): Ed25519PublicKey {
+    const value = deserializer.deserializeBytes();
+    return new Ed25519PublicKey(value);
+  }
+}
+
+export class Ed25519Signature {
+  static readonly LENGTH = 64;
+
+  constructor(public readonly value: Uint8Array) {
```
0xmaayan, line 38: why we dont accept HexInput here and then assign it to a class variable, same as all the other classes? Also the syntax is different, can we declare the value outside of the constructor ?

ecosystem/typescript/sdk_v2/src/crypto/multi_ed25519.ts
```diff
@@ -0,0 +1,159 @@
...
+   * transaction to be executed, at least K out of the N authorized signers have signed the transaction
+   * and passed the check conducted by the chain.
+   *
+   * @see {@link
+   * https://aptos.dev/guides/creating-a-signed-transaction#multisignature-transactions | Creating a Signed Transaction}
+   *
+   * @param public_keys A list of public keys
+   * @param threshold At least "threshold" signatures must be valid
+   */
+  constructor(public readonly public_keys: Ed25519PublicKey[], public readonly threshold: Uint8) {
```
0xmaayan, line 26: can we get the class variables outside the constructor? and keep the same syntax

ecosystem/typescript/sdk_v2/src/crypto/multi_ed25519.ts
```diff
@@ -0,0 +1,159 @@
...
+
+/* eslint-disable no-bitwise */
+import { Deserializer, Serializer } from "../bcs";
+import { Uint8 } from "../types";
+import { Ed25519PublicKey, Ed25519Signature } from "./ed25519";
+
+/**
+ * MultiEd25519 currently supports at most 32 signatures.
+ */
+const MAX_SIGNATURES_SUPPORTED = 32;
```
0xmaayan, line 12: thinking on maybe have it as a static variable in both classes instead of a global const... it doesn't make sny different but easier to understand and use for developers, i.e
```
MultiEd25519PublicKey.MAX_SIGNATURES_SUPPORTED

MultiEd25519Signature.MAX_SIGNATURES_SUPPORTED
```

ecosystem/typescript/sdk_v2/tests/unit/multi_ed25519.test.ts
```diff
@@ -0,0 +1,105 @@
...
+
+import { Deserializer } from "../../src/bcs/deserializer";
+import { Serializer } from "../../src/bcs/serializer";
+import { Hex } from "../../src/core/hex";
+import { Ed25519PublicKey, Ed25519Signature } from "../../src/crypto/ed25519";
+import { MultiEd25519PublicKey, MultiEd25519Signature } from "../../src/crypto/multi_ed25519";
+
+describe("MultiEd25519", () => {
+  it("public key serializes to bytes correctly", async () => {
+    const publicKey1 = "b9c6ee1630ef3e711144a648db06bbb2284f7274cfbee53ffcee503cc1a49200";
```
0xmaayan, line 12: we use consts in a lot of test files, wonder if we should have a consts/helper.ts file in the tests folder that keeps privateKeys, publicKeys, authKeys, etc and then use the same values in tests

ecosystem/typescript/sdk_v2/src/crypto/multi_ed25519.ts
```diff
@@ -0,0 +1,159 @@
...
+   * and passed the check conducted by the chain.
+   *
+   * @see {@link
+   * https://aptos.dev/guides/creating-a-signed-transaction#multisignature-transactions | Creating a Signed Transaction}
+   *
+   * @param public_keys A list of public keys
+   * @param threshold At least "threshold" signatures must be valid
+   */
+  constructor(public readonly public_keys: Ed25519PublicKey[], public readonly threshold: Uint8) {
+    if (threshold > MAX_SIGNATURES_SUPPORTED) {
```
0xmaayan, line 27: should we check for threshold < 0?

ecosystem/typescript/sdk_v2/src/crypto/multi_ed25519.ts
0xmaayan, line 26: wonder if we just use `threshold : number`

0xmaayan (changes requested):

ecosystem/typescript/sdk_v2/tests/unit/multi_ed25519.test.ts
0xjinn: Good suggestion!

ecosystem/typescript/sdk_v2/src/crypto/multi_ed25519.ts
0xjinn: Since the max threshold is 32, I think Uint8 should be fine here?

0xmaayan: `Uint8` is just a number type https://github.com/aptos-labs/aptos-core/blob/main/ecosystem/typescript/sdk_v2/src/types/index.ts#L11 - we declare it as Uint8 for BCS purposes. Since all BCS related code can technically live in its own library and outside the SDK, we should avoid using BCS related types throughout the SDK (might be wiser to move BCS types under the `bcs` folder and not place them in the global sdk types file - will add it as an item to the roadmap).

0xjinn: I see! Will update it to number.

ecosystem/typescript/sdk_v2/src/crypto/authentication_key.ts
```diff
@@ -0,0 +1,83 @@
...
+  static readonly MULTI_ED25519_SCHEME: number = 1;
+
+  // Scheme identifier for Ed25519 signatures used to derive authentication key for MultiEd25519 public key
+  static readonly ED25519_SCHEME: number = 0;
+
+  // Scheme identifier used when hashing an account's address together with a seed to derive the address (not the
+  // authentication key) of a resource account.
+  static readonly DERIVE_RESOURCE_ACCOUNT_SCHEME: number = 255;
+
+  private readonly _data: Hex;
```
0xmaayan, line 31: since we represent the data as `Hex`, should we also support `toUint8Array()` and `toString()` methods for easy retrieval? 
```
 toString(): string {
    return `0x${this.data.toString()}`;
  }

toUint8Array(): Uint8Array {
    return this.data.toUint8Array();
  }
```

ecosystem/typescript/sdk_v2/src/crypto/ed25519.ts
```diff
@@ -0,0 +1,55 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+import { Deserializer, Serializer } from "../bcs";
+import { Hex } from "../core";
+import { HexInput } from "../types";
+
+export class Ed25519PublicKey {
+  static readonly LENGTH: number = 32;
+
+  readonly value: Hex;
```
0xmaayan, line 11: `data` for consistency. 
```suggestion
  private readonly data: Hex;
```

ecosystem/typescript/sdk_v2/src/crypto/ed25519.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  constructor(hexInput: HexInput) {
+    const hex = Hex.fromHexInput({ hexInput });
+    if (hex.toUint8Array().length !== Ed25519PublicKey.LENGTH) {
+      throw new Error(`Ed25519PublicKey length should be ${Ed25519PublicKey.LENGTH}`);
+    }
+    this.value = hex;
+  }
+
+  toUint8Array(): Uint8Array {
```
0xmaayan, line 21: should we also support `toString()` method?

ecosystem/typescript/sdk_v2/src/crypto/ed25519.ts
0xmaayan, line 11: I wonder if we should represent the data as Uint8Array and not Hex. cc @banool for thoughts

ecosystem/typescript/sdk_v2/src/crypto/ed25519.ts
```diff
@@ -0,0 +1,55 @@
+// Copyright © Aptos Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+import { Deserializer, Serializer } from "../bcs";
+import { Hex } from "../core";
+import { HexInput } from "../types";
+
+export class Ed25519PublicKey {
+  static readonly LENGTH: number = 32;
```
0xmaayan, line 9: can we add comments explaining each field?

ecosystem/typescript/sdk_v2/src/crypto/multi_ed25519.ts
```diff
@@ -0,0 +1,173 @@
...
+
+  public readonly threshold: number;
+
+  /**
+   * Public key for a K-of-N multisig transaction. A K-of-N multisig transaction means that for such a
+   * transaction to be executed, at least K out of the N authorized signers have signed the transaction
+   * and passed the check conducted by the chain.
+   *
+   * @see {@link
+   * https://aptos.dev/guides/creating-a-signed-transaction#multisignature-transactions | Creating a Signed Transaction}
```
0xmaayan, line 22: `https://aptos.dev/guides/creating-a-signed-transaction#multisignature-transactions` gives me 404

ecosystem/typescript/sdk_v2/src/crypto/multi_ed25519.ts
```diff
@@ -0,0 +1,173 @@
...
+   *
+   * @param public_keys A list of public keys
+   * @param threshold At least "threshold" signatures must be valid
+   */
+  constructor(public_keys: Ed25519PublicKey[], threshold: number) {
```
0xmaayan, line 27:
```suggestion
  constructor(args:{public_keys: Ed25519PublicKey[], threshold: number}) {
```

ecosystem/typescript/sdk_v2/src/crypto/multi_ed25519.ts
```diff
@@ -0,0 +1,173 @@
+// Copyright © Aptos Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+/* eslint-disable no-bitwise */
+import { Deserializer, Serializer } from "../bcs";
+import { Uint8 } from "../types";
+import { Ed25519PublicKey, Ed25519Signature } from "./ed25519";
+
+export class MultiEd25519PublicKey {
+  static readonly MAX_SIGNATURES_SUPPORTED = 32;
```
0xmaayan, line 10: can we add comments explaining each field?

ecosystem/typescript/sdk_v2/src/crypto/multi_ed25519.ts
```diff
@@ -0,0 +1,173 @@
...
+   * @see {@link
+   * https://aptos.dev/guides/creating-a-signed-transaction#multisignature-transactions | Creating a Signed Transaction}
+   *
+   * @param public_keys A list of public keys
+   * @param threshold At least "threshold" signatures must be valid
+   */
+  constructor(public_keys: Ed25519PublicKey[], threshold: number) {
+    if (threshold < 0 || threshold > MultiEd25519PublicKey.MAX_SIGNATURES_SUPPORTED) {
+      throw new Error(
+        `"threshold" cannot be greater than 0 and larger than ${MultiEd25519PublicKey.MAX_SIGNATURES_SUPPORTED}`,
```
0xmaayan, line 30: maybe separate it to 2 different checks?
```suggestion
        `"threshold" cannot be lower than 0 and larger than ${MultiEd25519PublicKey.MAX_SIGNATURES_SUPPORTED}`,
```

ecosystem/typescript/sdk_v2/src/crypto/multi_ed25519.ts
```diff
@@ -0,0 +1,173 @@
...
+
+/* eslint-disable no-bitwise */
+import { Deserializer, Serializer } from "../bcs";
+import { Uint8 } from "../types";
+import { Ed25519PublicKey, Ed25519Signature } from "./ed25519";
+
+export class MultiEd25519PublicKey {
+  static readonly MAX_SIGNATURES_SUPPORTED = 32;
+
+  public readonly public_keys: Ed25519PublicKey[];
```
0xmaayan, line 12: js convention 
```suggestion
  public readonly publicKeys: Ed25519PublicKey[];
```

0xmaayan (changes requested):

ecosystem/typescript/sdk_v2/src/crypto/authentication_key.ts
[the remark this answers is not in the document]
0xjinn: Gotcha. I will update this to `public` as well, since it need to be publicly accessible.

0xjinn: Good call!

ecosystem/typescript/sdk_v2/src/crypto/ed25519.ts
```diff
@@ -0,0 +1,61 @@
...
+import { HexInput } from "../types";
+
+export class Ed25519PublicKey {
+  // Correct length of the public key in bytes (Uint8Array)
+  static readonly LENGTH: number = 32;
+
+  // The public key in hex format
+  readonly data: Hex;
+
+  constructor(hexInput: HexInput) {
```
0xmaayan, line 15: let's make sure we have the same syntax in all classes
```suggestion
  constructor(args:{data: HexInput}) {
```

ecosystem/typescript/sdk_v2/src/crypto/multi_ed25519.ts
```diff
@@ -0,0 +1,183 @@
...
+    for (let i = 0; i < bytes.length - 1; i += Ed25519PublicKey.LENGTH) {
+      const begin = i;
+      keys.push(new Ed25519PublicKey(bytes.subarray(begin, begin + Ed25519PublicKey.LENGTH)));
+    }
+    return new MultiEd25519PublicKey({ publieKeys: keys, threshold });
+  }
+}
+
+export class MultiEd25519Signature {
+  static MAX_SIGNATURES_SUPPORTED = 32;
```
0xmaayan, line 81: let's make sure we have comments explaining each class variables in all classes

ecosystem/typescript/sdk_v2/src/crypto/multi_ed25519.ts
```diff
@@ -0,0 +1,183 @@
...
+   * Signature for a K-of-N multisig transaction.
+   *
+   * @see {@link
+   * https://aptos.dev/guides/creating-a-signed-transaction#multisignature-transactions | Creating a Signed Transaction}
+   *
+   * @param signatures A list of ed25519 signatures
+   * @param bitmap 4 bytes, at most 32 signatures are supported. If Nth bit value is `1`, the Nth
+   * signature should be provided in `signatures`. Bits are read from left to right
+   */
+  constructor(signatures: Ed25519Signature[], bitmap: Uint8Array) {
```
0xmaayan, line 98:
```suggestion
  constructor(args:{signatures: Ed25519Signature[], bitmap: Uint8Array}) {
```

ecosystem/typescript/sdk_v2/tests/unit/authentication_key.test.ts
```diff
@@ -0,0 +1,46 @@
+// Copyright © Aptos Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+import { Hex } from "../../src/core/hex";
+import { AuthenticationKey } from "../../src/crypto/authentication_key";
+import { Ed25519PublicKey } from "../../src/crypto/ed25519";
+import { MultiEd25519PublicKey } from "../../src/crypto/multi_ed25519";
+
+// Auth key and Public key pair for testing
+const auth_key_hexInput = "0x6324287105756b0338e0f84025bd0ac80e58154eb94257b0d4f06ec6497e656e";
```
0xmaayan, line 10: let's get it to the `helper` file?

ecosystem/typescript/sdk_v2/tests/unit/helper.ts
```diff
@@ -0,0 +1,23 @@
+// Copyright © Aptos Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+export const multiEd25519PkTestObject = {
+  public_keys: [
```
0xmaayan, line 5:
```suggestion
  publicKeys: [
```

ecosystem/typescript/sdk_v2/src/crypto/ed25519.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  constructor(value: Uint8Array) {
+    if (value.length !== Ed25519Signature.LENGTH) {
+      throw new Error(`Ed25519Signature length should be ${Ed25519Signature.LENGTH}`);
+    }
+    this.data = value;
+  }
+
+  serialize(serializer: Serializer): void {
+    serializer.serializeBytes(this.data);
```
0xmaayan, line 54: `serializeBytes` expects `Uint8Array` type, `this.data` is of `Hex` type

0xmaayan (changes requested):
I am not sure we want to represent data as `Hex`. 
1) `Hex` is a class not a type. 
2) `Hex` should be used when working with `HexInput` type to convert it into the desired format by using the `Hex` class.
3) For all those classes, it looks like representing the inner `data` variable as `Uint8Array` makes more sense.

0xjinn:
@0xmaayan Ah ok I see. After I read more closely on what `Hex` class really is https://github.com/aptos-labs/aptos-core/blob/main/ecosystem/typescript/sdk_v2/src/core/hex.ts#L20-L26 Looks like its just a "helper" class to convert from/to Uint8Array/String. 

Ok I think this make sense now. I will migrate everything from Hex to Uint8Array

xbtmatt:
> @0xmaayan Ah ok I see. After I read more closely on what `Hex` class really is https://github.com/aptos-labs/aptos-core/blob/main/ecosystem/typescript/sdk_v2/src/core/hex.ts#L20-L26 Looks like its just a "helper" class to convert from/to Uint8Array/String.
> 
> Ok I think this make sense now. I will migrate everything from Hex to Uint8Array

@0xmaayan So just to clarify, we'll be using the HexInput to convert to/use the Hex class to represent Uint8Array data easily.

Whenever we need to actually pass around/use/store this data (a byte buffer) should we default to using the Hex class or the raw Uint8Array data?

Just clarifying general usage so we can keep it consistent internally so we don't end up with similar problems as the original sdk

ecosystem/typescript/sdk_v2/src/crypto/multi_ed25519.ts
```diff
@@ -0,0 +1,183 @@
...
+   * and passed the check conducted by the chain.
+   *
+   * @see {@link
+   * https://aptos.dev/integration/creating-a-signed-transaction/ | Creating a Signed Transaction}
+   *
+   * @param public_keys A list of public keys
+   * @param threshold At least "threshold" signatures must be valid
+   */
+  constructor(args: { publieKeys: Ed25519PublicKey[], threshold: number }) {
+    const { publieKeys, threshold } = args;
```
xbtmatt, line 31:
```suggestion
    const { publicKeys, threshold } = args;
```

ecosystem/typescript/sdk_v2/src/crypto/multi_ed25519.ts
```diff
@@ -0,0 +1,183 @@
...
+      );
+    }
+
+    if (threshold < 0) {
+      throw new Error(
+        '"threshold" cannot be less than 0',
+      );
+    }
+
+    this.publicKeys = publieKeys;
```
xbtmatt, line 44:
```suggestion
    this.publicKeys = publicKeys;
```

banool (commented):
There seems to be a lot of duplication with https://github.com/aptos-labs/aptos-core/pull/10005. If this is a stack, could you stack them properly?

I left some comments in that PR before I noticed the duplication.

banool:
The intent with the Hex class is that we actually use it to represent hex data, not Uint8Array. Otherwise we would've just made a file with a bunch of conversion functions rather than a class you can instantiate. Think of it as a Uint8Array with extra checks.

0xmaayan:
synced offline, we will go with Hex class for the inner data representation and HexInput for the input type. @0xjinn

PR #9895, opened 2023-09-01
[rosetta] Add docker build dependencies

gregnazario:
### Description
Fixes broken Rosetta Docker build
```
#11 32.79   Package libudev was not found in the pkg-config search path.
#11 32.79   Perhaps you should add the directory containing `libudev.pc'
#11 32.79   to the PKG_CONFIG_PATH environment variable
#11 32.79   No package 'libudev' found
```

### DIFF
--- a/docker/rosetta/rosetta.Dockerfile
+++ b/docker/rosetta/rosetta.Dockerfile
@@ -4,7 +4,9 @@ FROM ubuntu:20.04@sha256:a06ae92523384c2cd182dcfe7f8b2bf09075062e937d5653d7d0db0
 FROM rust:1.66.1-buster@sha256:e518dbab65069f4869f0159a460a989161ed277913c03427ed8b84542b771f7e AS rust-base
 
 WORKDIR /aptos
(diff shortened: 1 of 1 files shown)


rustielin (commented):
You might want to add the path to the dockerfile to the build trigger here, https://github.com/aptos-labs/aptos-core/blob/main/.github/workflows/docker-build-rosetta.yaml#L9, so you can build the image on this PR for testing

docker/rosetta/rosetta.Dockerfile
```diff
@@ -12,6 +12,21 @@ FROM rust-base as builder
...
+        cmake \
+        curl \
+        clang \
+        git \
+        pkg-config \
+        libssl-dev \
+        libpq-dev \
+        binutils \
+        lld \
+        libudev-dev
```
0xjinn, line 28: do we want to add it to line 7 instead?

0xjinn (approved):

docker/rosetta/rosetta.Dockerfile
gregnazario: Done

sionescu (approved):

rustielin (approved):

PR #9901, opened 2023-09-02
read hint

msmouse:
### Description
Real(er) read hints for p2p txns. 

### Test Plan

Printing out all fall back reads from the ptx executer and now seeing None data reads (module reads are still falling back to the base view)

### DIFF
--- a/types/src/transaction/analyzed_transaction.rs
+++ b/types/src/transaction/analyzed_transaction.rs
@@ -3,8 +3,10 @@
 
 use crate::{
     access_path::AccessPath,
-    account_config::{AccountResource, CoinStoreResource},
+    account_config::{AccountResource, CoinInfoResource, CoinStoreResource},
     block_metadata::BlockMetadata,
+    chain_id::ChainId,
+    on_chain_config::{CurrentTimeMicroseconds, Features, OnChainConfig, TransactionFeeBurnCap},
     state_store::{state_key::StateKey, table::TableHandle},
     transaction::{SignedTransaction, Transaction, TransactionPayload},
 };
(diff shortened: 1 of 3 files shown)


grao1991 (approved):

types/src/transaction/analyzed_transaction.rs
```diff
@@ -131,11 +133,18 @@ impl AnalyzedTransaction {
             // receiver account resource.
             write_hints.push(Self::account_resource_location(receiver_address));
         }
+
+        let read_hints = vec![
```
sitalkedia, line 137: Should we also add the framework code location here?

types/src/transaction/analyzed_transaction.rs
```diff
@@ -131,11 +133,18 @@ impl AnalyzedTransaction {
...
+            Self::current_ts_location(),
+            Self::features_location(),
+            Self::aptos_coin_info_location(),
+            Self::chain_id_location(),
+            Self::transaction_fee_burn_cap_location(),
```
sitalkedia, line 142: Also, how about account creation txn?

types/src/transaction/analyzed_transaction.rs
msmouse: That's a lot of modules, I'm thinking maybe we need to come up with some clever tricks down the road. For example, assuming commonly needed modules for all shards.

types/src/transaction/analyzed_transaction.rs
msmouse: Will need to print some logs to figure out and confirm. I'd test the performance impact with p2p first.

sitalkedia (approved):

types/src/transaction/analyzed_transaction.rs
msmouse: fails your test already :trollface: 
test_no_conflict_across_shards_in_non_last_rounds

sitalkedia: Let's rebase once https://github.com/aptos-labs/aptos-core/pull/9883 lands. I moved all these tests from v1 to v2. We will see if they fail in V2. cc - @zjma

PR #9908, opened 2023-09-04
Add print-bytecode support for Move source files in transactional tests.  Also fix v2 compiler handling of known_attributes in those tests.

brmataptos:
Add print-bytecode support for Move source files in transactional tests.  Also fix v2 compiler handling of known_attributes in those tests.

### Description
Allow the "print-bytecode" task in transactional-test-runner
to be applied to Move source code, as well as .mvir.

### Test Plan
Renamed the existing print_bytecode.move to print_bytecode.mvir since
MoveIR is what's in the file, and added a .move file to test that
Move source is detected, compiled, and bytecode printed.


brmataptos:
Btw, the fix to known_attributes doesn't do anything in compiler V2 at the moment, since it still is set to the default if not assigned, but if V2 is used on some tests in aptos-move then it might want to have the updated set of attributes which will get propagated there.

wrwg (commented):
Have you seen that a few weeks ago, `--print-bytecode` was introduced as an option to the `# publish` command? It is used e.g. here: 

https://github.com/aptos-labs/aptos-core/blob/main/third_party/move/move-compiler-v2/transactional-tests/tests/control_flow/sorter.move

I thought this time back that an option is better than an independent task because often you want to print this as a side effect of a real test which needs publish.

What does this PR do relative to this existing support?

brmataptos:
You can print byte code after compile even when publish or run fail.
Useful for debugging a failing test

brmataptos:
Btw, didn't want to change your complex control flow for that flag in case
you cared about it.

third_party/move/move-compiler-v2/transactional-tests/tests/control_flow/print_bytecode.exp
```diff
@@ -0,0 +1,31 @@
+comparison between v1 and v2 failed:
+= processed 2 tasks
+= 
+= task 0 'print-bytecode'. lines 1-4:
+- // Move bytecode v6
```
wrwg, line 5: Need to regenerate after rebase?

wrwg (approved):
Nice!

third_party/move/move-compiler-v2/transactional-tests/tests/control_flow/print_bytecode.exp
brmataptos: Not quite, V1 and V2 bytecode differs so this test doesn't work well as a V1V2Comparison.  I've added a new subdir /no-v1-comparison/ to allow this test to be run standalone.

rahxephon89 (approved):

third_party/move/move-compiler-v2/transactional-tests/tests/tests.rs
```diff
@@ -9,6 +9,18 @@ use std::path::Path;
 
 datatest_stable::harness!(run, TEST_DIR, r".*\.move$");
 
+/// Root of tests which require to skip v1-v2 comparison
+const SKIP_V1_COMPARISON_PATH: &str = "/no-v1-comparison/";
+
 fn run(path: &Path) -> Result<(), Box<dyn std::error::Error>> {
-    vm_test_harness::run_test_with_config(TestRunConfig::ComparisonV1V2, path)
+    let test_config: TestRunConfig = if let Some(p) = path.to_str() {
```
vineethk, line 16: I am not sure what our policy is on using features not available by default, but using let-chains would reduce some code duplication here:

`if let Some(p) = path.to_str() && p.contains(...) {...} else {...}`

However, to get this to work, we would need to add `#![feature(let_chains)]` above.

third_party/move/move-compiler-v2/transactional-tests/tests/tests.rs
```diff
@@ -9,6 +9,18 @@ use std::path::Path;
...
+        if p.contains(SKIP_V1_COMPARISON_PATH) {
```
vineethk, line 17: This check slightly differs from the comment above about "root of tests": we are not performing *that* check. Which might be okay, but we should say the appropriate thing in the comment above for `SKIP_V1_COMPARISON_PATH`.

vineethk (changes requested):
Minor changes requested.

third_party/move/move-compiler-v2/src/options.rs
```diff
@@ -30,6 +31,8 @@ pub struct Options {
     /// Do not complain about unknown attributes in Move code.
     #[clap(long, default_value = "false")]
     pub skip_attribute_checks: bool,
+    #[clap(skip)]
```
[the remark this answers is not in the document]
brmataptos: Ok.

third_party/move/testing-infra/transactional-test-runner/src/tasks.rs
```diff
@@ -215,6 +215,8 @@ pub struct PrintBytecodeCommand {
     /// The kind of input: either a script, or a module.
     #[clap(long = "input", value_enum, ignore_case = true, default_value_t = PrintBytecodeInputChoice::Script)]
     pub input: PrintBytecodeInputChoice,
+    #[clap(long = "syntax")]
```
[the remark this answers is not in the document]
brmataptos: Ok.

third_party/move/move-compiler-v2/transactional-tests/tests/tests.rs
brmataptos: I prefer not to do that, since it seems to be unstable.  How about. .unwrap_or_default() instead of the let match.

third_party/move/move-compiler-v2/transactional-tests/tests/tests.rs
brmataptos: A bit nit-picky.  But ok.

third_party/move/move-compiler-v2/transactional-tests/tests/tests.rs
vineethk: This was just a nit, and you are right that this is marked as "unstable feature". I don't have an opinion regarding `unwrap_or_default`.

vineethk (approved):
lgtm.

PR #9911, opened 2023-09-04
Upgrade version of Python used in SDK to fix tomllib dep

banool:
### Description
The new package publishing stuff depends on tomllib, which was only added in Python 3.11. This is one way to fix it, though I think we should not rely on 3.11 features because most folks are probably still on Python 3.9 ish.

### Test Plan
CI.


banool:
This way of fixing it is preferable: https://github.com/aptos-labs/aptos-core/pull/9941/.

PR #9919, opened 2023-09-05
Make Display impl of AccountAddress conform to AIP-40

banool:
### Description
This PR makes the Display impl of AccountAddress use to_standard_string. The motivation behind this change is to make it that when people use AccountAddress, they get standard compliant behavior by default.

I've taken particular care to try and make sure everything that uses AccountAddress is unchanged as a result of this PR, except in these specific cases where I think the change won't cause any issues. I have a previous attempt at this where I also try to change the Debug and Serialize impls, but it got out of hand quickly: https://github.com/aptos-labs/aptos-core/pull/9183. This PR just changes Display. Beyond that, it doesn't try to update the rest of the code to adopt the new Display implementation, but rather switches it to use the old underlying implementation. For example:

Before:
```rust
format!("0x{}::{}", addr, module_name);
```

After:
```rust
format!("0x{}::{}", addr.to_hex(), module_name);
```

This way nothing is changed. We can then go change these if we like later on piece by piece.

There are some cases where I have opted to keep the new Display impl because I know it is safe:
- Logging in the faucet

Places I’ve confirmed are unchanged:

- The CLI, e.g. `aptos init`, `aptos node show-validator-set`, `aptos config show-profiles --profile default`
- The validator configs, tested with `aptos node run-local-testnet`. The addresses are all still the same.
- Any storage of addresses by the node.
- How addresses are returned by the v1 node API.
- How addresses are returned by any existing indexer API.
- The Rust REST API client.
- Use of addresses in aptos-move and third_party/move.

### Test Plan
Manually running the CLI E2E tests:
```
DOCKER_DEFAULT_PLATFORM=linux/amd64 poetry run python main.py -d --base-network testnet --test-cli-path ~/a/core/target/debug/aptos
```

Confirming that the new CLI works with an older config.
```
cd /tmp/testing
aptos init
~/aptos-core/target/debug/aptos show-profiles
```

Confirming that a node from before this change happily starts with a set of configs generated by the new CLI (once I remove config sections that the new CLI generates that the node in the existing CLI doesn't know about, but not fixing any address stuff).
```
cd /tmp/node
~/aptos-core/target/debug/aptos node run-local-testnet
aptos node run-local-testnet
```

Beyond that I've gone through all the tests in third_party/move/, aptos-move/, api/, and more to make sure nothing is changed.

### DIFF
--- a/api/src/tests/resource_groups.rs
+++ b/api/src/tests/resource_groups.rs
@@ -49,8 +49,8 @@ async fn test_gen_resource_group() {
     context.publish_package(&mut admin1, txn).await;
 
     // Read default data
-    let primary = format!("0x{}::{}::{}", admin0.address(), "primary", "Primary");
-    let secondary = format!("0x{}::{}::{}", admin1.address(), "secondary", "Secondary");
+    let primary = format!("{}::{}::{}", admin0.address(), "primary", "Primary");
+    let secondary = format!("{}::{}::{}", admin1.address(), "secondary", "Secondary");
 
     let response = context.gen_resource(&admin0.address(), &primary).await;
     assert_eq!(response.unwrap()["data"]["value"], "3");
@@ -65,7 +65,7 @@ async fn test_gen_resource_group() {
     context
         .api_execute_entry_function(
             &mut user,
-            &format!("0x{}::secondary::init", admin1.address()),
+            &format!("{}::secondary::init", admin1.address()),
             json!([]),
             json!([55]),
         )
@@ -80,7 +80,7 @@ async fn test_gen_resource_group() {
     context
         .api_execute_entry_function(
             &mut user,
-            &format!("0x{}::primary::init", admin0.address()),
+            &format!("{}::primary::init", admin0.address()),
             json!([]),
             json!(["35"]),
         )
--- a/crates/aptos-rest-client/src/lib.rs
+++ b/crates/aptos-rest-client/src/lib.rs
@@ -879,7 +879,7 @@ impl Client {
         start: Option<u64>,
         limit: Option<u64>,
     ) -> AptosResult<Response<Vec<Transaction>>> {
-        let url = self.build_path(&format!("accounts/{}/transactions", address))?;
+        let url = self.build_path(&format!("accounts/{}/transactions", address.to_hex()))?;
 
         let mut request = self.inner.get(url);
         if let Some(start) = start {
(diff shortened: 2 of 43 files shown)


api/src/tests/resource_groups.rs
```diff
@@ -49,8 +49,8 @@ async fn test_gen_resource_group() {
     context.publish_package(&mut admin1, txn).await;
 
     // Read default data
-    let primary = format!("0x{}::{}::{}", admin0.address(), "primary", "Primary");
```
0xjinn, line 52: why are we removing "0x" here? Should we do a to_hex() here as well?

third_party/move/tools/move-cli/src/sandbox/utils/mod.rs
```diff
@@ -517,6 +517,7 @@ pub(crate) fn explain_execution_error(
             // TODO: map to source code location
             let location_explanation = match location {
                 AbortLocation::Module(id) => {
+                    // This isn't it.
```
gregnazario, line 520: probably remove this comment

crates/aptos-rest-client/src/lib.rs
```diff
@@ -1292,7 +1320,7 @@ impl Client {
         &self,
         address: AccountAddress,
     ) -> AptosResult<Response<AccountResource>> {
-        let url = self.build_path(&format!("accounts/{}", address))?;
+        let url = self.build_path(&format!("accounts/{}", address.to_hex()))?;
```
gregnazario, line 1323: These should probably not be `.to_hex()` if it should be accepting whatever any address input on the API.

gregnazario (commented):
lgtm, please cleanup the two spots I mentioned, I don't think we need to change the rest client at all since it should accept either.

crates/aptos-rest-client/src/lib.rs
banool: Perhaps, but there are lots of things like this in the PR. I'd rather make such changes in later PRs.

api/src/tests/resource_groups.rs
banool: In this case we're just using it for comparison in the tests and there are no goldens related to it or anything so it's fine. I could do `to_hex` but it wouldn't really be any different.

0xjinn: Ahh I didn't realize these were tests. sgtm

0xjinn (approved):

gregnazario (approved):

PR #9924, opened 2023-09-05
ExecutionPipeline

msmouse:
### Description
BufferManager schedules execution early into ExecutionPipeline, so that block execution and ledger apply can be carried out in a pipelined way.
Execution phase waits for the execution result.

### Test Plan

From forge runs we can see time has shifted out of the ExecutionPhase (now ExecutionWaitPhase). 
image

image

End to end improvement being confirmed.


zekun000 (commented):
lgtm

consensus/src/experimental/buffer_manager.rs
```diff
@@ -220,6 +227,15 @@ impl BufferManager {
             ordered_proof.commit_info(),
             self.buffer.len() + 1,
         );
+
+        let request = self.create_new_request(ExecutionRequest {
+            ordered_blocks: ordered_blocks.clone(),
+        });
+        self.execution_schedule_phase_tx
```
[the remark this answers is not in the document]
msmouse: Yes, for retry -- we need to re-issue the schedule to retry.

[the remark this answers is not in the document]
msmouse: Actually we can let `BufferManager`  hold the state computer and call `computer.schedule_compute(parent_id, ordered_blocks)` in place (how I coded it previously), but I think it's cleaner like this.

consensus/src/state_computer.rs
```diff
@@ -473,3 +481,149 @@ async fn test_commit_sync_race() {
...
+
+const PIPELINE_DEPTH: usize = 8;
+
+struct ExecutionPipeline {
+    block_tx: mpsc::Sender<ExecuteBlockCommand>,
+}
+
+impl ExecutionPipeline {
+    pub fn spawn(executor: Arc<dyn BlockExecutorTrait>, runtime: &tokio::runtime::Handle) -> Self {
+        let (block_tx, block_rx) = mpsc::channel(PIPELINE_DEPTH);
```
sitalkedia, line 493: What happens if we have more blocks in flight than the pipeline depth?

consensus/src/experimental/buffer_manager.rs
```diff
@@ -237,19 +253,16 @@ impl BufferManager {
...
-                let sender = self.execution_phase_tx.clone();
-                Self::spawn_retry_request(sender, request, Duration::from_millis(100));
-            } else {
-                self.execution_phase_tx
-                    .send(request)
-                    .await
-                    .expect("Failed to send execution request")
-            }
+            let sender = self.execution_schedule_phase_tx.clone();
+            Self::spawn_retry_request(sender, request, Duration::from_millis(100));
```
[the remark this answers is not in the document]
msmouse: Previously queued requests will all fail and be retried. Added protection in the latest commit.

consensus/src/state_computer.rs
```diff
@@ -473,3 +481,149 @@ async fn test_commit_sync_race() {
     assert!(executor.sync_to(generate_li(2, 8)).await.is_ok());
     assert_eq!(*recorded_commit.time.lock(), LogicalTime::new(2, 8));
 }
+
+const PIPELINE_DEPTH: usize = 8;
+
+struct ExecutionPipeline {
```
sitalkedia, line 487: May be better to move it to its own module?

consensus/src/experimental/execution_schedule_phase.rs
```diff
@@ -0,0 +1,89 @@
...
+                .await;
+            futs.push(fut)
+        }
+
+        // In the future being returned, wait for the compute results in order.
+        let fut = Box::pin(async move {
+            let mut results = vec![];
+            for (block, fut) in itertools::zip_eq(ordered_blocks, futs) {
+                debug!("try to receive compute result for block {}", block.id());
+                results.push(ExecutedBlock::new(block.ditch_result(), fut.await?));
```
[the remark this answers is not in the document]
msmouse: Done, see the latest commit.

consensus/src/state_computer.rs
```diff
@@ -473,3 +481,149 @@ async fn test_commit_sync_race() {
...
+            state_checkpoint_output,
+            result_tx,
+        }) = block_rx.recv().await
+        {
+            debug!("ledger_apply stage received block {}.", block_id);
+            let res = async {
+                let executor = executor.clone();
+                monitor!(
+                    "ledger_apply",
+                    tokio::task::spawn_blocking(move || {
```
sitalkedia, line 596: Why do we need to use spawn_blocking here and above?

sitalkedia:
Looks good overall. Can you update the PR with a meaningful summary?

consensus/src/state_computer.rs
msmouse: line 513 will "block" (switch out) -- I think it's better than throwing in a lot of things unexpectedly into the pipeline.

consensus/src/state_computer.rs
msmouse: That's the way it's supposed to execute run CPU and IO intensive tasks in async environment, to avoid blocking the async event loop. (Search for `spawn_blocking` to see how we do this today for all executor calls (hopefully))

consensus/src/state_computer.rs
msmouse: done

consensus/src/state_computer.rs
```diff
@@ -124,48 +125,54 @@ impl StateComputer for ExecutionProxy {
...
-        if let Err(e) = self
-            .txn_notifier
-            .notify_failed_txn(shuffled_txns, &compute_result)
-            .await
-        {
-            error!(
-                error = ?e, "Failed to notify mempool of rejected txns",
+        Box::pin(async move {
+            debug!(
+                block_id = %block_id,
```
zekun000, line 158: why %? that prints out the whole 32 bytes

zekun000 (approved):

consensus/src/state_computer.rs
msmouse: -_- looked nicer to me. But you are the king.

danielxiangzl (approved):
LGTM

PR #9929, opened 2023-09-05
Sharding v3: graph partitioning and interactive BlockSTM

zjma:
Multiple BlockSTM instances can now collaborate to execute a block reordered and partitioned by a serious graph partitioner like `FennelGraphPartitioner`.

Different from `PartitionerV2`-baesd sharding, it only has 1 round, and the dependency edges can cross. E.g. it's possible that:
- [T1, T3] are assigned to shard 0
- [T2, T4] are assigned to shard 1
- T4 depends on T1
- T3 depends on T2

Example command to try:
```
# create db
cargo run -p aptos-executor-benchmark -- --block-size 1000 --split-ledger-db --use-sharded-state-merkle-db create-db --data-dir /tmp/db2 --num-accounts 10000

# run transfer
cargo run -p aptos-executor-benchmark -- --use-sharding-v3 --block-size 1000  --num-executor-shards 2 --split-ledger-db --use-sharded-state-merkle-db run-executor --main-signer-accounts 10000 --data-dir /tmp/db2 --checkpoint-dir /tmp/ck2 --blocks 9
```

`FennelGraphPartitioner` is from #9739 .

Some major BlockSTM refactoring in this PR:
1. Operating based on the txns' global indices (so the update of the txns on other shard can easily participate in local execution/scheduling).
2. MVHashMap now starts with remote dependencies marked as `ESTIMATE`.
3. Dedicated commit thread(s), otherwise we get deadlock because of (2).


aptos-move/block-executor/src/executor.rs
```diff
@@ -122,11 +123,13 @@ where
     ) -> SchedulerTask {
         let _timer = TASK_EXECUTE_SECONDS.start_timer();
         let (idx_to_execute, incarnation) = version;
-        let txn = &signature_verified_block[idx_to_execute as usize];
+        let global_txn_idx = sharding_provider.global_idx_from_local(idx_to_execute);
+        sharding_provider.wait_for_remote_deps(idx_to_execute);
```
zjma, line 127: @xosmig I asked for key-level dependencies but during implementation I realized I can let local txn wait for their dep txns and there's no need to touch the MVHashMap...

execution/streaming-partitioner/src/main.rs
```diff
@@ -0,0 +1,192 @@
...
+                transactions,
+                TransactionGraphPartitioner::new(random_partitioner, params),
+                args,
+            )
+        },
+    };
+}
+
+pub fn edge_weight_function(idx1: SerializationIdx, idx2: SerializationIdx) -> EdgeWeight {
+    ((1. / (1. + idx1 as f64 - idx2 as f64)) * 100000.) as EdgeWeight
```
zjma, line 191: @xosmig how was 100000 chosen? is the ratio between this value and the default node weight matter?

xosmig: It was pretty arbitrary. This whole edge weight function is not very scientific and was just the first thing I came up with. 10^5 just seemed like enough precision for the tests I was running. However, on some inputs, it caused overflows.

I imagine, eventually, we will use our own partitioner implementation and will be able to use floats instead of i32, so there won't be such problems.

xosmig:
> a serious graph partitioner like FennelGraphPartitioner

Fennel is not particularly serious, it's, perhaps, the simplest graph partitioner there is. In the non-streaming model, it's probably better to test with Metis.

execution/streaming-partitioner/src/transaction_graph_partitioner.rs
```diff
@@ -0,0 +1,611 @@
...
+        let mut new_edges_weight = 0 as EdgeWeight;
+
+        // Add edges to the dependency graph based on the dependencies.
+        let basic_network_latency_penalty: EdgeWeight = 10;
+        let cross_round_discount: EdgeWeight = 10;
+        let per_key_penalty: EdgeWeight = 1;
+        let idx_round = idx / 48;
+        for (&dep, keys) in deps.iter() {
+            let dep_round = dep / 48;
+            let edge_weight = per_key_penalty * (keys.len() as EdgeWeight) + min(0, basic_network_latency_penalty - cross_round_discount * ((idx_round - dep_round) as EdgeWeight));
```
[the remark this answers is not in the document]
xosmig: Did you mean `max` instead of `min`? Also, with `basic_network_latency_penalty == cross_round_discount`, will it be just binary? I mean, the second part will be either 0 or 10, right?

In any case, only experiments will show which edge weight function to use, it's very hard to estimate costs theoretically.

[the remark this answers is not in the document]
zjma: Did mean `max`... thanks for catching!

In my very rough estimation, it currently takes up to 2ms for a txn to move from ready state to committed state in blockstm, which can be higher than the intrinsic network latency. Reordering may matter more.

Anyway, totally agree, will continue tuning and let the benchmark number drive.

And thanks for your great job!

PR #9933, opened 2023-09-05
[metadata] Metadata fixes

wrwg:
Measures metadata size


runtian-zhou (approved):

aptos-move/aptos-vm/src/aptos_vm_impl.rs
```diff
@@ -58,6 +58,7 @@ pub struct AptosVMImpl {
     storage_gas_params: Result<StorageGasParameters, String>,
     version: Option<Version>,
     features: Features,
+    timed_features: TimedFeatures,
```
zekun000, line 61: unrelated change?

wrwg: This is needed in the other repo and will help also in the future, so lets keep it.

gedigi (approved):

PR #9941, opened 2023-09-06
[Python SDK] Switch from tomllib to tomli

banool:
### Description
The newly added `package_publisher.py` uses tomllib. This is only available in Python 3.11. Rather than forcing SDK users to use Python 3.11 (which is not really the "standard" right now, for example in Ubuntu 22.04 it is 3.10), this PR switches us to tomli, a self described backport.

### Test Plan
```
make test
```
```
poetry run python -m examples.large_package_publisher
```
See that it worked here: https://explorer.aptoslabs.com/account/0x1f8f7237ba43e133ad6eb76649cf9eaec58f1938a9b283a91f44baed3a2c5596/transactions?network=testnet.


0xjinn (approved):

davidiw (approved):
when I wrote the Python sdk I never had compromise in mind... :sob:

PR #9944, opened 2023-09-06
[CLI][Coverage] Reduce execution time of coverage tool

### LINKED ISSUE #7295: [Bug][CLI] Coverage tool is prohibitively slow
@gregnazario @movekevin @wrwg @rahxephon89 #6893 

Running the [Econia](https://github.com/econia-labs/econia) unit test suite takes only about 17 seconds without coverage, but 
 over 7 minutes with coverage:

```zsh
# From inside Econia repo core dir
cd src/move/econia/
# Running without coverage
time aptos move test
...
Test result: OK. Total tests: 347; passed: 347; failed: 0
{
  "Result": "Success"
}
aptos move test  32.48s user 0.39s system 192% cpu 17.033 total
# Running with coverage
time aptos move test --coverage
...
+-------------------------+
| % Move Coverage: 100.00  |
+-------------------------+
Please use `aptos move coverage -h` for more detailed source or bytecode test coverage of this package
{
  "Result": "Success"
}
aptos move test --coverage  84.08s user 379.89s system 108% cpu 7:09.38 total
```

The coverage run in question also included up to two-minute delays between successive tests.

Run on a 2021 MackBook Pro M1 Max with 32 GB memory.

rahxephon89:
### Description

This PR reduces execution time of coverage tool by 1) only printing necessary data items to the trace file; 2) using buffer writer. Close #7295

### Test Plan

`time aptos move test --coverage`

After this change, the execution time of 1) aptos-framework is reduced from 3.5mins to 30s; 2) aptos-stdlib is reduced from 3.5mins to 30s and 3) move-stdlib is reduced from 2mins to 13s.

### DIFF
--- a/third_party/move/move-vm/runtime/src/tracing.rs
+++ b/third_party/move/move-vm/runtime/src/tracing.rs
@@ -18,14 +18,15 @@ use ::{
         env,
         fs::{File, OpenOptions},
         io::Write,
-        process,
         sync::Mutex,
-        thread,
     },
 };
 
 #[cfg(any(debug_assertions, feature = "debugging"))]
 const MOVE_VM_TRACING_ENV_VAR_NAME: &str = "MOVE_VM_TRACE";
+
+#[cfg(any(debug_assertions, feature = "debugging"))]
+const MOVE_VM_TRACING_FLUSH_ENV_VAR_NAME: &str = "MOVE_VM_TRACE_FLUSH";
 
 #[cfg(any(debug_assertions, feature = "debugging"))]
 const MOVE_VM_STEPPING_ENV_VAR_NAME: &str = "MOVE_VM_STEP";
@@ -36,23 +37,30 @@ static FILE_PATH: Lazy<String> = Lazy::new(|| {
 });
 
 #[cfg(any(debug_assertions, feature = "debugging"))]
-static TRACING_ENABLED: Lazy<bool> = Lazy::new(|| env::var(MOVE_VM_TRACING_ENV_VAR_NAME).is_ok());
+pub static TRACING_ENABLED: Lazy<bool> =
+    Lazy::new(|| env::var(MOVE_VM_TRACING_ENV_VAR_NAME).is_ok());
 
 #[cfg(any(debug_assertions, feature = "debugging"))]
 static DEBUGGING_ENABLED: Lazy<bool> =
     Lazy::new(|| env::var(MOVE_VM_STEPPING_ENV_VAR_NAME).is_ok());
 
 #[cfg(any(debug_assertions, feature = "debugging"))]
-static LOGGING_FILE: Lazy<Mutex<File>> = Lazy::new(|| {
-    Mutex::new(
-        OpenOptions::new()
-            .write(true)
-            .create(true)
-            .append(true)
-            .open(&*FILE_PATH)
-            .unwrap(),
-    )
+pub static LOGGING_FILE_WRITER: Lazy<Mutex<std::io::BufWriter<File>>> = Lazy::new(|| {
+    let file = OpenOptions::new()
+        .write(true)
+        .create(true)
+        .append(true)
+        .open(&*FILE_PATH)
+        .unwrap();
+    Mutex::new(std::io::BufWriter::with_capacity(
+        4096 * 1024, /* 4096KB */
+        file,
+    ))
 });
+
+#[cfg(any(debug_assertions, feature = "debugging"))]
+pub static SINGLE_STEP_FLUSHING: Lazy<bool> =
+    Lazy::new(|| env::var(MOVE_VM_TRACING_FLUSH_ENV_VAR_NAME).is_ok());
 
 #[cfg(any(debug_assertions, feature = "debugging"))]
 static DEBUG_CONTEXT: Lazy<Mutex<DebugContext>> = Lazy::new(|| Mutex::new(DebugContext::new()));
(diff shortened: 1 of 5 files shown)


third_party/move/move-vm/runtime/src/tracing.rs
```diff
@@ -69,16 +67,7 @@ pub(crate) fn trace(
 ) {
     if *TRACING_ENABLED {
         let f = &mut *LOGGING_FILE.lock().unwrap();
-        writeln!(
```
rahxephon89, line 72: `processId-threadId` is used as the key in the coverage map. However, it is not used when generating the coverage info and printing this time-consuming. So it is removed from the trace file. bytecode instruction is also removed due to the same reason

third_party/move/tools/move-coverage/src/coverage_map.rs
```diff
@@ -68,7 +68,7 @@ impl CoverageMap {
         for line in BufReader::new(file).lines() {
             let line = line.unwrap();
             let mut splits = line.split(',');
-            let exec_id = splits.next().unwrap();
+            let exec_id = "exec_id";
```
rahxephon89, line 71: Use a fake key so that the data structure of the coverage map does not need to be changed

third_party/move/move-vm/runtime/src/tracing.rs
```diff
@@ -69,16 +67,10 @@ pub(crate) fn trace(
...
-            "{}-{:?},{},{},{:?}",
-            process::id(),
-            thread::current().id(),
-            function_desc.pretty_string(),
-            pc,
-            instr,
-        )
-        .unwrap();
+        let mut buf_writer = std::io::BufWriter::new(f);
+        buf_writer
```
brmataptos, line 71: https://doc.rust-lang.org/std/io/struct.BufWriter.html# says:

> "It is critical to call [flush](https://doc.rust-lang.org/std/io/struct.BufWriter.html#method.flush) before BufWriter<W> is dropped. Though dropping will attempt to [flush](https://doc.rust-lang.org/std/io/struct.BufWriter.html#method.flush) the contents of the buffer, any errors that happen in the process of dropping will be ignored. Calling flush ensures that the buffer is empty and thus dropping will not even attempt file operations."

This whole approach seems a bit questionable.  Putting the BufWriter inside LOGGING_FILE would make more sense.  I'd also try using with_capacity() instead of new(), so you can specify some larger capacity, at least 16K (16384).  The default seems to be 8K.  Much larger may still be good, even 4MiB.  Not sure how to ensure that the global gets flushed, though.  Do we have an exit handler somewhere?

third_party/move/tools/move-cli/src/base/test.rs
```diff
@@ -268,6 +269,10 @@ pub fn run_move_unit_tests<W: Write + Send>(
 
     // Compute the coverage map. This will be used by other commands after this.
     if compute_coverage && !no_tests {
+        if *TRACING_ENABLED {
```
rahxephon89, line 272: @brmataptos I add flush before the trace file is used for generating the coverage map.

third_party/move/move-vm/runtime/src/tracing.rs
rahxephon89: Thanks for the review, @brmataptos. As you suggested, I move this to `LOGGING_FILE_WRITER` above.

brmataptos (approved):

third_party/move/move-vm/runtime/src/tracing.rs
```diff
@@ -68,17 +76,13 @@ pub(crate) fn trace(
...
-            function_desc.pretty_string(),
-            pc,
-            instr,
-        )
-        .unwrap();
+        let buf_writer = &mut *LOGGING_FILE_WRITER.lock().unwrap();
+        buf_writer
+            .write_fmt(format_args!("{},{}\n", function_desc.pretty_string(), pc,))
+            .unwrap();
+        if *SINGLE_STEP_FLUSHING {
```
vineethk, line 83: Looks like this is always enabled if tracing in the move vm is enabled.

If so, there does not seem to be any difference in using a buffered write: we end up flushing after write.

Was there any measurable performance increase after adding buffered-writes-with-frequent-flushes vs. using plain unbuffered writes (of the same amount of data)?

third_party/move/tools/move-coverage/src/coverage_map.rs
```diff
@@ -277,7 +277,7 @@ impl TraceMap {
...
```
vineethk, line 280: Same as above: note that this is a dummy (or better yet, call it `"dummy_exec_id"` in both places).

vineethk (changes requested):
Since this is a performance improvement PR, could you provide some numbers on the improvements this PR brings? Eg., it took `x` time to run blah before this PR, and it now takes `y` time.

third_party/move/move-vm/runtime/src/tracing.rs
rahxephon89: It is only enabled in testing

vineethk: Curious: do you have some numbers for the performance gain by enabling write buffering?

rahxephon89:
> Since this is a performance improvement PR, could you provide some numbers on the improvements this PR brings? Eg., it took `x` time to run blah before this PR, and it now takes `y` time.

Updated in the description of this PR.

third_party/move/tools/move-cli/src/sandbox/commands/test.rs
```diff
@@ -46,6 +46,10 @@ pub const TEST_ARGS_FILENAME: &str = "args.txt";
 /// enabled in the move VM.
 const MOVE_VM_TRACING_ENV_VAR_NAME: &str = "MOVE_VM_TRACE";
 
+/// Name of the environment variable we need to set in order to flush
+/// the trace.
```
vineethk, line 50:
```suggestion
/// after every trace call.
```

vineethk (approved):
I have a minor suggestion in the review. lgtm otherwise.

PR #9950, opened 2023-09-06
[aptos-release-builder] Deprecate Single Step Proposal

runtian-zhou:
### Description

Deprecate the Single Step Governance proposal option for the release builder.

### Test Plan

Existing test should cover this.


rahxephon89 (approved):

sherry-x (approved):

movekevin (commented):
Can we kill single step in the governance cli as well? For both proposing and resolving

PR #9951, opened 2023-09-06
[forge/smoke tests] Replace raw config and yaml usage with OverrideNodeConfig

bchocho:
### Description

Introduce OverrideNodeConfig. Its purpose is to allow test writers to use NodeConfig structs, while still outputting only a minimal yaml config. This unifies forge and smoke tests and solves the pain points in the previous approach of each:
* In smoke tests, the entire config was being outputted as the yaml config. Config optimizer treats explicit yaml configs as overrides, meaning some config optimizations (e.g., that rely on the node role) are not applied.
* In forge tests, the raw yaml was being used to change configs. This is awkward and error prone (e.g., if a config gets renamed).

### Test Plan

Run existing forge and smoke tests.


config/src/config/node_config.rs
```diff
@@ -286,4 +459,114 @@ mod test {
...
+        .unwrap();
+        assert_eq!(diff_yaml, expected_yaml);
+    }
+
+    #[test]
+    fn test_override_node_config_with_enum() {
+        let mut override_config =
+            OverrideNodeConfig::new(NodeConfig::default(), NodeConfig::default());
+        let config = override_config.config_mut();
+        config.base.waypoint = WaypointConfig::FromFile(PathBuf::from("test"));
```
JoshLind, line 493: nit: empty line after this to match the formatting of the other tests? 😄 (I actually found it pretty helpful for readability).

config/src/config/node_config.rs
```diff
@@ -286,4 +459,114 @@ mod test {
...
+        assert_eq!(diff_yaml, expected_yaml);
+    }
+
+    #[test]
+    fn test_override_node_config_with_empty_base_vector() {
+        let mut override_config =
+            OverrideNodeConfig::new(NodeConfig::default(), NodeConfig::default());
+        let config = override_config.config_mut();
+        config.full_node_networks.push(Default::default());
+        config.full_node_networks.push(Default::default());
```
JoshLind, line 512: nit: empty line here to match `test_override_node_config_with_non_empty_base_vector` below? 😄

config/src/config/node_config.rs
```diff
@@ -286,4 +459,114 @@ mod test {
...
+        let diff_yaml = override_config.get_yaml().unwrap();
+        let default_node_config = serde_yaml::to_value(&NetworkConfig::default()).unwrap();
+        let mut expected_yaml: serde_yaml::Value = serde_yaml::Value::Null;
+        expected_yaml["full_node_networks"] =
+            serde_yaml::Value::Sequence(vec![default_node_config.clone(), default_node_config]);
+        // TODO: why don't the actual Values match, only matches with as_str?
```
JoshLind, line 518: nit: it might be worth creating a helper function for this, e.g., `assert_equal_yaml()`? That way the"ugliness" and TODO is only in one place 😄

JoshLind (approved):
Awesome, thanks @bchocho 😄 LGTM! Mostly nits.

config/src/config/node_config.rs
```diff
@@ -244,9 +245,181 @@ pub fn merge_node_config(
...
+/// A utility struct for managing a node config that is overriding a base config, and outputting a
+/// yaml representation of it. In most cases, the base config will be the default config, and the
+/// output will be a minimal yaml diff that an aptos node can read.
+///
+/// In rare cases you may want to explicitly write a yaml value that is the same as the default
+/// config (e.g., to avoid the config being optimized by ConfigOptimizer). To do this, change the
+/// base config to a different value before calling get_yaml().
+#[derive(Deserialize, Serialize, PartialEq, Debug, Clone)]
+pub struct OverrideNodeConfig {
+    config: NodeConfig,
```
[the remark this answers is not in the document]
bchocho: Unfortunately `override` is a reserved word. I can call it `override_config` and `base_config`

testsuite/forge-cli/src/main.rs
```diff
@@ -608,23 +663,18 @@ fn run_consensus_only_three_region_simulation() -> ForgeConfig {
             helm_values["chain"]["epoch_duration_secs"] = (24 * 3600).into();
         }))
         .with_node_helm_config_fn(Arc::new(|helm_values| {
```
igor-aptos, line 665: should we provide similar function to what is with_init_config in Smoke tests, to make this less verbose?

i.e. this gets changed with:

```
 .with_init_config(Arc::new(|conf| {
            update_mempool_config_practically_non_expiring(conf.mempool);
            conf.consensus.max_sending_block_txns = 5000;
            conf.consensus.max_receiving_block_txns = 30000;
            conf.consensus.max_sending_block_bytes = 3 * 1024 * 1024;
            update_state_sync_config_execute_transactions(conf.state_sync);
        }))
```

and with_init_config then first creates default NodeConfig, passes it here, and then calls 
helm_values["validator"]["config"] = override_yaml_from_node_config(config);

This removes also the ..Default::default() bit , which is annoying.

igor-aptos (commented):
if our configs are dynamic in the node, do we see an issue here, where we want to set a config to a default value and make sure node doesn't adjust it - and this change making that impossible?

config/src/config/override_node_config.rs
```diff
@@ -0,0 +1,299 @@
+// Copyright © Aptos Foundation
+
+use crate::config::{Error, NodeConfig, PersistableConfig};
+use anyhow::bail;
+use serde::{Deserialize, Serialize};
+use std::path::Path;
+
+/// Diff a config yaml with a base config yaml. Returns None if there is no diff.
+fn diff_override_config_yaml(
```
sitalkedia, line 9: This function can be made a bit cleaner and reduce code duplication by writing as 

```
fn diff_override_config_yaml(
    override_config: serde_yaml::Value,
    base_config: serde_yaml::Value,
) -> anyhow::Result<Option<serde_yaml::Value>> {
    match (override_config, base_config) {
        (serde_yaml::Value::Mapping(override_mapping), serde_yaml::Value::Mapping(base_mapping)) => {
            let mut overrides = serde_yaml::Mapping::new();
            for (override_key, override_value) in override_mapping {
                if let Some(base_value) = base_mapping.get(&override_key) {
                    if let Some(diff_value) = diff_override_config_yaml(override_value, base_value.clone())? {
                        overrides.insert(override_key, diff_value);
                    }
                } else {
                    overrides.insert(override_key, override_value);
                }
            }
            if overrides.is_empty() {
                Ok(None)
            } else {
                Ok(Some(serde_yaml::Value::Mapping(overrides)))
            }
        },
        (serde_yaml::Value::Null, serde_yaml::Value::Null) => Ok(None),
        (serde_yaml::Value::Bool(override_value), serde_yaml::Value::Bool(base_value)) => {
            if override_value == base_value {
                Ok(None)
            } else {
                Ok(Some(override_config))
            }
        },
        (serde_yaml::Value::Number(override_value), serde_yaml::Value::Number(base_value)) => {
            if override_value == base_value {
                Ok(None)
            } else {
                Ok(Some(override_config))
            }
        },
        (serde_yaml::Value::String(override_value), serde_yaml::Value::String(base_value)) => {
            if override_value == base_value {
                Ok(None)
            } else {
                Ok(Some(override_config))
            }
        },
        (serde_yaml::Value::Sequence(override_value), serde_yaml::Value::Sequence(base_value)) => {
            if override_value == base_value {
                Ok(None)
            } else {
                Ok(Some(override_config))
            }
        },
        _ => bail!("base does not match override: {:?}, {:?}", override_config, base_config),
    }
}

```

sitalkedia: You can go more fancier and reduce the duplication further with use of macro as well..


```
macro_rules! compare_and_return {
    ($override_value:expr, $base_value:expr, $override_config:expr, $type_name:literal) => {{
        if $override_value == $base_value {
            Ok(None)
        } else {
            bail!(
                "base does not match override: {}({}), {:?}",
                $type_name,
                $override_value,
                $base_value
            );
        }
    }};
}

fn diff_override_config_yaml(
    override_config: serde_yaml::Value,
    base_config: serde_yaml::Value,
) -> anyhow::Result<Option<serde_yaml::Value>> {
    match (&override_config, &base_config) {
        (
            serde_yaml::Value::Mapping(override_mapping),
            serde_yaml::Value::Mapping(base_mapping),
        ) => {
            let mut overrides = serde_yaml::Mapping::new();
            for (override_key, override_value) in override_mapping {
                if let Some(base_value) = base_mapping.get(override_key) {
                    if let Some(diff_value) =
                        diff_override_config_yaml(override_value.clone(), base_value.clone())?
                    {
                        overrides.insert(override_key.clone(), diff_value);
                    }
                } else {
                    overrides.insert(override_key.clone(), override_value.clone());
                }
            }
            if overrides.is_empty() {
                Ok(None)
            } else {
                Ok(Some(serde_yaml::Value::Mapping(overrides)))
            }
        }
        (serde_yaml::Value::Null, serde_yaml::Value::Null) => Ok(None),
        (serde_yaml::Value::Bool(override_value), serde_yaml::Value::Bool(base_value)) => {
            compare_and_return!(override_value, base_value, override_config, "Bool")
        }
        (serde_yaml::Value::Number(override_value), serde_yaml::Value::Number(base_value)) => {
            compare_and_return!(override_value, base_value, override_config, "Number")
        }
        (serde_yaml::Value::String(override_value), serde_yaml::Value::String(base_value)) => {
            compare_and_return!(override_value, base_value, override_config, "String")
        }
        (serde_yaml::Value::Sequence(override_value), serde_yaml::Value::Sequence(base_value)) => {
            compare_and_return!(override_value, base_value, override_config, "Sequence")
        }
        _ => bail!(
            "base does not match override: {:?}, {:?}",
            override_config,
            base_config
        ),
    }
}

```

config/src/config/override_node_config.rs
```diff
@@ -0,0 +1,299 @@
...
+
+    pub fn override_config_mut(&mut self) -> &mut NodeConfig {
+        &mut self.override_config
+    }
+
+    pub fn base_config(&self) -> &NodeConfig {
+        &self.base_config
+    }
+
+    pub fn base_config_mut(&mut self) -> &mut NodeConfig {
```
sitalkedia, line 142: Seems this is used for testing purpose only, if so, mark this test-only ?

bchocho:
@igor-aptos if we really want to hard-code a config (make sure it is written in the yaml), the base_config can be changed to another value.

testsuite/forge-cli/src/main.rs
bchocho: I agree this is less verbose. But it does completely remove the possibility of hard-coding already default values. I suppose we could include both options, but then it gets a bit confusing (it matters which function is called first)

config/src/config/override_node_config.rs
bchocho: I think that changes the behavior of when override is a Mapping but the base is not. But let me see if I can still simplify it.

config/src/config/override_node_config.rs
bchocho: It's currently only used for testing, but it's not designed only for testing -- if someone needs to write the default value into a yaml they could use this.

testsuite/forge-cli/src/main.rs
igor-aptos: If really needed, we can have function provide two arguments: (config, base_config), and most uses will have _ for the second one.

bchocho: Hmm, as it is now, we'd need two versions:
with_init_validator_config
with_init_fullnode_config
and also some tests change other helm_values, e.g., helm_values["chain"] so we'd still need a way to do that.

igor-aptos: It would be good to have Chain as a struct as well (so that fields are typed and stuff).
and having 3 functions, for each of the validator, fullnode and chain configs - sounds good?

bchocho: Oh nevermind, chain is part of genesis. The only other node helm values that are touched are in `multiregion_benchmark_test`. I think I'll just separate those into its own part of the builder and we could generalize them if people actually need them later.

bchocho:
@igor-aptos could you take another look? forge and smoke tests both now take in (override_config, base_config), although they technically don't share the same interfaces. (That would probably be another PR, for another day)

bchocho:
Also this PR contains #10009 so it might be easier reviewing that first (it's small)

config/src/config/override_node_config.rs
```diff
@@ -0,0 +1,277 @@
+// Copyright © Aptos Foundation
```
JoshLind, line 1: nit: we need to update the lint for this, but technically it should be:
```
```

JoshLind (approved):
Looks reasonable to me (reapproving since Github somehow wiped the last one 😄)

igor-aptos (approved):

PR #9952, opened 2023-09-06
[State Sync] Update latency monitor to track all advertised versions.

JoshLind:
### Description
This PR updates the latency monitor to track all advertised versions, even if the versions were only advertised *after* they were synced by the node.

Currently, if a version is advertised after it was synced by the node, it is not tracked for latency purposes, and this results in several missing metrics. Instead, this PR updates the code such that if we've already synced an advertised version, we'll still track the version latencies and the calculated "seen to synced" latency will instead be set to 0.

### Test Plan
Existing test infrastructure.


bchocho (approved):
LGTM

sitalkedia (approved):

PR #9953, opened 2023-09-06
bounded math utilities for simpler aggregator formulas

igor-aptos:

### DIFF
--- a/aptos-move/aptos-aggregator/src/aggregator_change_set.rs
+++ b/aptos-move/aptos-aggregator/src/aggregator_change_set.rs
@@ -1,9 +1,11 @@
+// Copyright © Aptos Foundation
+
 use crate::{
     aggregator_extension::{
-        validate_history, AggregatorID, AggregatorState, DeltaHistory, DeltaValue,
+        validate_history, AggregatorID, AggregatorState, DeltaHistory,
         SpeculativeStartValue,
     },
-    delta_change_set::{abort_error, addition, subtraction},
+    bounded_math::{abort_error, BoundedMathResult, ok_overflow, ok_underflow, addition_deltavalue},
 };
 use aptos_state_view::StateView;
 use move_binary_format::errors::{PartialVMError, PartialVMResult};
@@ -35,7 +37,7 @@ impl AggregatorChange {
 
     /// Returns the result of applying the AggregatorState to the base value
     /// Returns error if postcondition is not satisfied.
-    pub fn apply_aggregator_change_to(&self, base: u128) -> PartialVMResult<u128> {
+    pub fn apply_aggregator_change_to(&self, base: u128) -> BoundedMathResult<u128> {
         match self.state {
             AggregatorState::Data { value } => Ok(value),
             AggregatorState::Delta { delta, .. } => {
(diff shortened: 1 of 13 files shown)


aptos-move/aptos-aggregator/src/aggregator_change_set.rs
```diff
@@ -85,179 +87,71 @@ impl AggregatorChange {
...
-                        let new_min_overflow = match (
-                            prev_history.min_overflow_positive_delta,
-                            history.min_overflow_positive_delta,
-                        ) {
-                            (
-                                Some(prev_min_overflow_positive_delta),
-                                Some(min_overflow_positive_delta),
-                            ) => {
-                                // We know that previous_speculative_value + prev_delta + min_overflow_positive_delta is
-                                // greater than max_value. Otherwise, valiate_history function should have returned an error.
```
igor-aptos, line 131: we are not calling validate_history above any more. so now we need to check min_overflow_positive_delta + prev_delta >= 0, and fail merging otherwise

aptos-move/aptos-aggregator/src/aggregator_change_set.rs
```diff
@@ -85,179 +87,71 @@ impl AggregatorChange {
...
                                     prev_delta,
                                     self.max_value,
-                                )
-                                .ok()
-                            },
-                            (None, None) => None,
+                                ))
+                            )?;
+
+                            match (adjusted_min_overflow_positive_delta, prev_history.min_overflow_positive_delta) {
```
igor-aptos, line 104: we should be able to simplify merging with this

aptos-move/aptos-aggregator/src/aggregator_change_set.rs
```diff
@@ -85,179 +87,71 @@ impl AggregatorChange {
...
+                            )?;
+
+                            match (adjusted_max_underflow_negative_delta, prev_history.max_underflow_negative_delta) {
+                                (Some(a), Some(b)) => Some(u128::min(a, b)),
+                                (a, b) => a.or(b),
+                            }
                         };
 
                         // new_max_achieved = max(prev_max_achieved, max_achieved + prev_delta)
-                        let new_max_achieved = match prev_delta {
```
igor-aptos, line 213: we need to fail on overflow

aptos-move/aptos-aggregator/src/aggregator_extension.rs
```diff
@@ -324,201 +309,120 @@ impl Aggregator {
     }
 
     /// Implements logic for adding to an aggregator.
+    /// Returns error if transaction is in invalid state, and should be re-executed.
+    /// Returns true if addition succeeded, and false if it would overflow.
     pub fn try_add(
         &mut self,
         resolver: &dyn AggregatorResolver,
         input: u128,
-    ) -> PartialVMResult<()> {
```
igor-aptos, line 331: we need to distinguish try_add needing to return false, from any other error - which we should propagate and abort the transaction

aptos-move/aptos-aggregator/src/aggregator_extension.rs
```diff
@@ -324,201 +309,120 @@ impl Aggregator {
...
-                        }
-                    },
-                };
-                self.state = AggregatorState::Delta {
-                    speculative_start_value,
-                    delta: new_delta,
-                    history,
-                };
-                self.get_mut_history().unwrap().record_success(new_delta);
+                let cur_value = expect_ok(addition_deltavalue(speculative_start_value.get_value(), delta, self.max_value))?;
```
igor-aptos, line 345: we should be able to simplify this by computing current_value first.

also I think it would be useful to distinguish things that should never happen (which expect_ok here does) and we are not fataling purely because we are in move, from the errors that mean speculative execution is incorrect and needs to be re-executed.

i.e. here, unless we have a bug in the code, we should always be able to add delta to speculative start value

igor-aptos (commented):
@vusirikala

aptos-move/aptos-aggregator/src/aggregator_extension.rs
```diff
@@ -247,18 +238,12 @@ pub fn validate_history(
         if history.min_overflow_positive_delta.is_some()
             && base_value <= max_value - history.min_overflow_positive_delta.unwrap()
         {
-            return Err(abort_error(
-                format!("Overflow was expected when setting the aggreagator start value to {}. Previous speculative start value = {:?}, Min overflow delta = {}, Max value = {}", base_value, speculative_start_value, history.min_overflow_positive_delta.unwrap(), max_value),
-                EEXPECTED_OVERFLOW,
-            ));
+            return Err(BoundedMathError::Overflow);
```
vusirikala, line 241: When validating history, this is a scenario where the original execution went well, but an overflow is supposed to occur. I distinguish these EXECPECTED_OVERFLOW, EXPECTED_UNDERFLOW errors from the original ADD_OVERFLOW, SUB_UNDERFLOW errors. Distinguishing the error codes might help us when debugging. Do we wanna merge those error codes?

aptos-move/aptos-aggregator/src/aggregator_extension.rs
vusirikala: So, we are returning Ok(false) when we know that re-execution the transaction again would also fail. We are returning Err(..) when the current execution failed, but re-execution could possibly succeed. Is that right?

aptos-move/aptos-aggregator/src/aggregator_extension.rs
igor-aptos: yeah, I don't know what's better. 

but I can revert change to validate_history, as that doesn't affect anything else anyways

aptos-move/aptos-aggregator/src/bounded_math.rs
```diff
@@ -0,0 +1,141 @@
...
+
+pub fn ok_underflow<T>(value: BoundedMathResult<T>) -> BoundedMathResult<Option<T>> {
+    match value {
+        Ok(value) => Ok(Some(value)),
+        Err(BoundedMathError::Overflow) => Err(BoundedMathError::Overflow),
+        Err(BoundedMathError::Underflow) => Ok(None),
+    }
+}
+
+pub fn expect_ok<T>(value: BoundedMathResult<T>) -> PartialVMResult<T> {
```
vusirikala, line 58: We are mapping even overflow and underflow errors to ECODE_INVARIANT_BROKEN? overflow and underflow errors are not bugs in the code as specified in the below error message right?

aptos-move/aptos-aggregator/src/aggregator_extension.rs
igor-aptos: we are returning Ok(false) when try_add call should return false - i.e. called +1 when value=limit.  No re-execution needed (unless start_value changes)

we return Err(..) when try_add detected irreconcilable state - meaning that current execution is not valid - even if start_value doesn't change.

aptos-move/aptos-aggregator/src/aggregator_extension.rs
igor-aptos: updated

aptos-move/aptos-aggregator/src/aggregator_extension.rs
```diff
@@ -296,7 +287,7 @@ impl Aggregator {
     }
 
     #[allow(dead_code)]
-    fn get_value(&self) -> PartialVMResult<u128> {
+    fn get_value(&self) -> BoundedMathResult<u128> {
```
vusirikala, line 290: Just trying to understand good design patterns. What's the benefit of distinguishing BoundedMathResult from PartialVMResult here?

aptos-move/aptos-aggregator/src/bounded_math.rs
igor-aptos: we are mapping them when expect_ok is explicitly called. 

so the caller decides if it code invariant, or regular case. 

for example:
```
                    let overflow_value = expect_ok(ok_overflow(addition_deltavalue(
                        input,
                        delta,
                        self.max_value,
                    )))?;
```

this will make overflow_value be None on Overflow (due to ok_overflow call), but raise ECODE_INVARIANT_BROKEN on underflow (due to expect_ok call around it)

aptos-move/aptos-aggregator/src/aggregator_extension.rs
```diff
@@ -324,201 +315,120 @@ impl Aggregator {
...
-                self.state = AggregatorState::Delta {
-                    speculative_start_value,
-                    delta: new_delta,
-                    history,
-                };
-                self.get_mut_history().unwrap().record_success(new_delta);
+                let cur_value = expect_ok(addition_deltavalue(speculative_start_value.get_value(), delta, self.max_value))?;
+
+                if addition(cur_value, input, self.max_value).is_err() {
+                    let overflow_value = expect_ok(ok_overflow(addition_deltavalue(input, delta, self.max_value)))?;
```
vusirikala, line 354: Why `expect_ok` on top of `ok_underflow` here? Having an overflow here is not an invariant error.

aptos-move/aptos-aggregator/src/aggregator_extension.rs
igor-aptos: not sure which result we should be returning here. 


but at the high level - PartiaulVMResult is uncostrained - anything error.

BoundedMathResult basically restricts errors to be only overflow and underflow.
If we have a numeric operation (like addition), that can only fail in those two ways, returning BoundedMathResult allows caller to be able to handle errors (i.e. it knows those are the only two outcomes, and can handle each the way it needs) - i.e. enabled things like expect_ok(ok_overflow())

aptos-move/aptos-aggregator/src/aggregator_extension.rs
igor-aptos: in this if, we are in the case that try_add exceeds limit (i.e. the is_err() check above)

and are now computing at what delta that happens, and so are computing  overflow_value = input + delta.

even if delta is negative, this cannot be < 0, and so we want to throw CODE_INVARIANT if that somehow ended up being < 0. 
if it is > limit, then we just don't want to record it, so we want overflow_value to be None in that case.

That's why we do ok_overflow - to transform overflow to None, and then expect_ok to raise CODE_INVARIANT.

aptos-move/aptos-aggregator/src/bounded_math.rs
```diff
@@ -0,0 +1,152 @@
...
+/// When updating the aggregator start value (due to read operations
+/// or at the end of the transaction), we realize that mistakenly raised
+/// an underflow in one of the previus try_sub operation.
+pub(crate) const EEXPECTED_UNDERFLOW: u64 = 0x02_0004;
+
+pub(crate) const ECODE_INVARIANT_BROKEN: u64 = 0x02_0005;
+
+#[derive(Debug)]
+pub enum BoundedMathError {
+    Overflow,
```
vusirikala, line 29: We have 5 error codes Bound_underflow, Bound_overflow, Expected_underflow, Expected_overflow, Code_invariant_Broken. We have only two variants for BoundedMathError?

igor-aptos: correct.

addition and subtraction operations can only fail with overflow and underflow. Caller of those functions decides what is expected, what isn't, and what error with which context should be propagated to the caller.

Issue #6907, opened 2023-03-03
[Bug] Compiler failing bytecode verifier with inline functions 


gerben-stavenga:
# 🐛 Bug

Compiling an function calling an inline function results in an internal compiler error

## To reproduce

at this commit 

https://github.com/aptos-labs/aptos-core/tree/f6070b81e0c791eec38227b5c0227e84907f0639

compile the move-stdlib with 

cargo run -p aptos -- move compile 

**Code snippet to reproduce**
```rust
# Your code goes here
# Please make sure it does not require any external dependencies
```

**Stack trace/error message**
```
// Paste the output here
```
 for_each_ref(v, |elem| push_back(&mut result, f(elem)));
      │         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ ICE failed bytecode verifier: VMError {
      major_status: STLOC_TYPE_MISMATCH_ERROR,
      sub_status: None,
      message: None,
      exec_state: None,
      location: Module(
          ModuleId {
              address: 0000000000000000000000000000000000000000000000000000000000000001,
              name: Identifier(
                  "vector",
              ),
          },
      ),
      indices: [
          (
              FunctionDefinition,
              20,
          ),
      ],
      offsets: [
          (
              FunctionDefinitionIndex(20),
              9,
          ),
      ],
  }

## Expected Behavior

A normal compile error or working compile

brmataptos:
@gerben-stavenga can you provide a specific test case or at least a branch illustrating the problem?  It's hard to figure out what is going on with a dangling commit that's not even in the repo.

"This commit does not belong to any branch on this repository, and may belong to a fork outside of the repository."

https://github.com/aptos-labs/aptos-core/tree/f6070b81e0c791eec38227b5c0227e84907f0639

brmataptos:
Gerben's example is in this dangling commit: https://github.com/aptos-labs/aptos-core/blob/f6070b81e0c791eec38227b5c0227e84907f0639/aptos-move/framework/move-stdlib/sources/vector.move

I've added a test based on this code here in my branch brm-issue-6907-new:
https://github.com/brmataptos/aptos-core/tree/brm-issue-6907-new

Not sure it's the same erorr, but there's a compiler ICE when you run:

cargo test -p bytecode-verifier-transactional-tests -- --skip prover gerbens_test --nocapture

brmataptos:
[...]

Issue #9459, opened 2023-08-03
[Bug] Passing mutable referace to option::is_none function via inline function


K-Pomian:
# 🐛 Bug

The bug occurs when a mutable reference of Option instance is passed to an inline function, that requires reference of Option instance, and then it is passed to `option::is_none` function that's inside an assertion. Passing a mutable reference instead of a normal reference works when an inline function is not used and the assertion is made directly inside an entry function.

## To reproduce

This does not work

```
module 0x1337::reproduce {
    use std::option::{Self, Option};
    #[test_only]
    use aptos_framework::account;

    const ESpaceAlreadyMarked: u64 = 0;

    public entry fun init(_account: &signer) {
        let space: Option<u8> = option::none();
        check_if_space_is_open(&mut space);
    }

    inline fun check_if_space_is_open(space: &Option<u8>) {
        // TODO: Ensure given space is not already marked. If it is, abort with code:
        //          ESpaceAlreadyMarked
        assert!(
            option::is_none(space),
            ESpaceAlreadyMarked
        );
    }

    #[test]
    fun test() {
        let account = account::create_account_for_test(@0x1337);
        init(&account);
    }
}
```

This works

```
module 0x1337::reproduce {
    use std::option::{Self, Option};
    #[test_only]
    use aptos_framework::account;

    const ESpaceAlreadyMarked: u64 = 0;

    public entry fun init(_account: &signer) {
        let space: Option<u8> = option::none();
        assert!(
            option::is_none(&mut space),
            ESpaceAlreadyMarked
        );
    }

    #[test]
    fun test() {
        let account = account::create_account_for_test(@0x1337);
        init(&account);
    }
}
```

**Stack trace/error message**

```
bug: BYTECODE VERIFICATION FAILED
   ┌─ /home/kpomian/reproduce/sources/reproduce.move:17:13
   │
17 │             option::is_none(space),
   │             ^^^^^^^^^^^^^^^^^^^^^^ ICE failed bytecode verifier: VMError {
    major_status: CALL_TYPE_MISMATCH_ERROR,
    sub_status: None,
    message: None,
    exec_state: None,
    location: Module(
        ModuleId {
            address: 0000000000000000000000000000000000000000000000000000000000001337,
            name: Identifier(
                "reproduce",
            ),
        },
    ),
    indices: [
        (
            FunctionDefinition,
            0,
        ),
    ],
    offsets: [
        (
            FunctionDefinitionIndex(0),
            3,
        ),
[...]

PR #9954, opened 2023-09-07
Fix Issue 6907: A verifier failure related to inline functions

brmataptos:
Fix Issue 6907: A verifier failure related to inline functions

### Description

- provide 2 test cases related to Gerben's bug:
  - gerbens_test.move: ICE in hlir/translate.rs if lambda is provided to unknown or non-inline
  - gerbens_test1.move: byte verifier error when (&mut vector) actual is passed to (&vector) param
- fix both errors:
  - generate a useful diagnostic error instead of the ICE
  - retain formal parameter types when inlining, but also
    - substitute all type parameters appropriately in them

### Test Plan
Added test cases to illustrate bug.  Any collateral damage should show up in existing test cases.

### Fixes
https://github.com/aptos-labs/aptos-core/issues/6907

### DIFF
--- a/third_party/move/move-bytecode-verifier/transactional-tests/tests/type_safety/bug_9459_working.move
+++ b/third_party/move/move-bytecode-verifier/transactional-tests/tests/type_safety/bug_9459_working.move
@@ -1,0 +1,22 @@
+//# publish
+module 0x1337::reproduce {
+    use std::option::{Self, Option};
+    #[test_only]
+    use aptos_framework::account;
+
+    const ESpaceAlreadyMarked: u64 = 0;
+
+    public entry fun init(_account: &signer) {
+        let space: Option<u8> = option::none();
+        assert!(
+            option::is_none(&mut space),
+            ESpaceAlreadyMarked
+        );
+    }
+
+    #[test]
+    fun test() {
+        let account = account::create_account_for_test(@0x1337);
+        init(&account);
+    }
+}
--- a/third_party/move/move-compiler/src/inlining/translate.rs
+++ b/third_party/move/move-compiler/src/inlining/translate.rs
@@ -231,7 +231,8 @@ struct SubstitutionVisitor<'l, 'r> {
 
 impl<'l, 'r> Visitor for SubstitutionVisitor<'l, 'r> {
     fn type_(&mut self, ty: &mut Type) -> VisitorContinuation {
-        visit_type(&self.type_arguments, ty)
+        visit_type(&self.type_arguments, ty);
+        VisitorContinuation::Descend
     }
 
     fn exp(&mut self, ex: &mut Exp) -> VisitorContinuation {
@@ -642,6 +643,7 @@ impl<'l> Inliner<'l> {
                 .zip(mcall.type_arguments.iter())
                 .map(|(p, t)| (p.id, t.clone()))
                 .collect();
+
             let mut inliner_visitor = OuterVisitor { inliner: self };
             let mut inlined_args = mcall.arguments.clone();
             Dispatcher::new(&mut inliner_visitor).exp(&mut inlined_args);
@@ -653,16 +655,24 @@ impl<'l> Inliner<'l> {
             let mut param_dispatcher = Dispatcher::new(&mut param_visitor);
             let fix_types = |(var, mut spanned_type): (Var, Type)| {
                 param_dispatcher.type_(&mut spanned_type);
+                self.infer_abilities(&mut spanned_type);
                 (var, spanned_type)
             };
-            let mapped_params = fdef
+            let mapped_params: Vec<_> = fdef
                 .signature
                 .parameters
                 .iter()
                 .cloned()
                 .map(fix_types)
-                .zip(get_args_from_exp(&inlined_args));
-            let (decls_for_let, bindings) = self.process_parameters(call_loc, mapped_params);
+                .zip(get_args_from_exp(&inlined_args))
+                .collect();
+            let (decls_for_let, bindings) =
+                self.process_parameters(call_loc, mapped_params.into_iter());
+
+            // Expand Type formal params in result type
+            let mut result_type = fdef.signature.return_type.clone();
+            param_dispatcher.type_(&mut result_type);
+            self.infer_abilities(&mut result_type);
 
             // Expand the body in its own independent visitor
             self.inline_stack.push_front(global_name); // for cycle detection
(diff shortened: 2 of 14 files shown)


runtian-zhou (approved):

brmataptos:
This also fixes https://github.com/aptos-labs/aptos-core/issues/9459.

third_party/move/move-compiler/src/inlining/translate.rs
```diff
@@ -101,7 +102,7 @@ impl<'l> Inliner<'l> {
                 if !fdef.inline {
                     let mut visitor = OuterVisitor { inliner };
                     Dispatcher::new(&mut visitor).function(fdef);
-                    //Self::eprint_fdef(&format!("fun {} ", _name), fdef);
+                    //DEBUG Self::eprint_fdef(&format!("fun {} ", _name), fdef);
```
rahxephon89, line 105: Do we need this?

third_party/move/move-compiler/src/inlining/translate.rs
```diff
@@ -725,9 +735,12 @@ impl<'l> Inliner<'l> {
             },
         };
 
+        let spanned_lvalues = sp(loc, lvalues);
+        let lvalue_ty = lvalues_expected_types(&spanned_lvalues); // BUGBUG - added
```
rahxephon89, line 739: What does this comment mean?

third_party/move/move-bytecode-verifier/transactional-tests/tests/type_safety/bug_9459_working.move
rahxephon89, line 1: Does your PR have anything to do with this test case?

brmataptos: It also fixes this test case.

third_party/move/move-compiler/src/inlining/translate.rs
brmataptos: Not the new comment.  I'll leave the old one, though.

third_party/move/move-compiler/src/inlining/translate.rs
brmataptos: Oops, was marking my place during debugging.

rahxephon89 (approved):

brmataptos:
Please take another look.  I added some a change to transactional-test-runner/src/framework.rs to replace the temporary filename in error messages with "TEMPFILE" to allow expected compiler errors (as happens with gerbens_test.move).

HOWEVER: running tests again I found that test third_party/move/move-compiler/transactional-tests/tests/inlining/options.move is now failing (see the output shown in the last commit), and I still have to debug this.
I have a handle on it (need to handle types better in the visitor, perhaps why TypedVisitor exists), but it's late.

brmataptos:
Ok, @rahxephon89 and @runtian-zhou please take another look.  I fixed the tests, albeit at the cost of some ugliness in inliner/translate that works but I can't completely explain.  Basically, making sure the right types are applied to each side of the lets representing parameter assignments ensures that "let x: &v = y &mut v;" gets a freeze inserted in a later stage of the compiler.  Exp structs have types, but apparently wrapping in Annotate exprs which apply another type seems to help.  There are probably redundant type conversions in there, but hopefully all the tests pass now.

third_party/move/move-compiler/src/inlining/visitor.rs
```diff
@@ -533,26 +257,27 @@ impl<'l, V: TypedVisitor> TypedDispatcher<'l, V> {
     }
 
     pub fn sequence(&mut self, seq: &mut Sequence) {
-        let mut scoped = false;
+        let mut scope_cnt = 0;
```
rahxephon89, line 260: Is this for consistency with the other one at line 249?

third_party/move/move-compiler/src/inlining/visitor.rs
```diff
@@ -43,285 +43,6 @@ impl<'l, V: Visitor> Dispatcher<'l, V> {
...
-    fn ty(&mut self, _t: &mut Type) -> VisitorContinuation {
-        VisitorContinuation::Descend
-    }
-    fn enter_scope(&mut self) {}
-    fn var_decl(&mut self, _ty: &mut Type, _var: &mut Var) {}
-    fn var_use(&mut self, _ty: &mut Type, _var: &mut Var) {}
-    fn exit_scope(&mut self) {}
-}
-
-pub struct TypedDispatcher<'l, V: TypedVisitor> {
```
rahxephon89, line 315: Same for `TypedDispatcher`

third_party/move/move-compiler/src/inlining/visitor.rs
```diff
@@ -43,285 +43,6 @@ impl<'l, V: Visitor> Dispatcher<'l, V> {
...
-                    self.type_(ty);
-                    self.lvalue(slv, declared);
-                }
-            },
-            LValue_::Ignore => {},
-        }
-    }
-}
-
-pub trait TypedVisitor {
```
rahxephon89, line 302: Could you explain more on what `TypedVisitor` was used for and why it is removed in this PR?

rahxephon89 (changes requested):

third_party/move/move-compiler/src/inlining/visitor.rs
brmataptos: The difference shown is confusing, since the TypedDispatcher is gone.  The Dispatcher code had the scoped_cnt.  The TypedDispatcher had the simpler scoped predicate, which I assume probably doesn't work right in some cases of nested structures.  This diff is showing the diff between end of TypedDispatcher (old) with the end of Dispatcher (new).

third_party/move/move-compiler/src/inlining/visitor.rs
brmataptos: I removed TypedDispatcher and TypedVisitor, adding the missing type_ visiting functionality to Dispatcher/Visitor so it can be used for both cases.

third_party/move/move-compiler/src/inlining/visitor.rs
brmataptos: As discussed in our meeting, TypedVisitor was apparently added by Meng at some point for SignatureExtractionVisitor, because at that point Visitor didn't handle types within the expression tree, and he apparently didn't want to change it.  There was some unnecessary difference between TypedVisitor/Dispatcher and VisitorDispatcher, so I fixed that and added the type handling to Visitor/Dispatcher  (in this PR) so we only need one of them.

third_party/move/move-compiler/src/inlining/visitor.rs
```diff
@@ -565,7 +290,7 @@ impl<'l, V: TypedVisitor> TypedDispatcher<'l, V> {
     fn lvalue(&mut self, lv: &mut LValue, declared: bool) {
         match &mut lv.value {
             LValue_::Var(var, ty) => {
-                self.type_(ty);
+                self.type_(ty.as_mut());
```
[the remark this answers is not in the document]
brmataptos: No, wasn't needed.  Good catch.

rahxephon89 (approved):

vineethk (approved):

PR #9956, opened 2023-09-07
[TS SDK v2] Add serializable classes that represent all entry function argument types

xbtmatt:
### Description

Built on #9881 

Introduces BCS-serializable classes that represent all of the possible entry function argument types. The types that accept meaningful generics (Option/MoveVectors) also support nested/composable serialization.

Nested deserialization is more complicated, since typescript doesn't have support for type information at runtime. Right now, the current solution is to pass in a specific type to deserialize as an argument, but this doesn't facilitate nested deserialization.

```typescript
// Typescript class => Move representation
U8 => u8
U16 => u16
U32 => u32
U64 => u64
U128 => u128
U256 => u256
MoveOption => Option
MoveString => String
MoveVector => vector
MoveObject => Object
AccountAddress => address
```

`MoveVector` and `MoveOption` both have factory methods that support creating generically typed instances aka `MoveVector<u64>` or `MoveOption<u8>` with primities (numbers, strings, etc)

you can use any nested form of these by passing in Serializable values:

```typescript
const vecOptionU8: MoveVector<MoveOption<U8>> = new MoveVector<MoveOption<U8>>([
   MoveOption.U8(1),
   MoveOption.U8(undefined),
   MoveOption.U8(123)
]);
```

Most names to be determined still, just using these for now.

Examples are in the unit tests and some are even documented in the doc comments already:

```typescript
 * @example
 * // in Move: `vector<u8> [1, 2, 3, 4];`
 * const vecOfU8s = new MoveVector<U8>([new U8(1), new U8(2), new U8(3), new U8(4)]);
 * // in Move: `std::bcs::to_bytes(vector<u8> [1, 2, 3, 4]);`
 * const bcsBytes = vecOfU8s.toUint8Array();
 *
 * // vector<vector<u8>> [ vector<u8> [1], vector<u8> [1, 2, 3, 4], vector<u8> [5, 6, 7, 8] ];
 * const vecOfVecs = new MoveVector<MoveVector<U8>>([
 *   new MoveVector<U8>([new U8(1)]),
 *   MoveVector.U8([1, 2, 3, 4]),
 *   MoveVector.U8([5, 6, 7, 8]),
 * ]);
 * 
 * // vector<Option<u8>> [ std::option::some<u8>(1), std::option::some<u8>(2) ];
 * const vecOfOptionU8s = new MoveVector<MoveOption<U8>>([
 *    MoveOption.U8(1),
 *    MoveOption.U8(2),
 * ]);
 *
 * // vector<String> [ std::string::utf8(b"hello"), std::string::utf8(b"world") ];
 * const vecOfStrings = new MoveVector([new MoveString("hello"), new MoveString("world")]);
 * const vecOfStrings2 = MoveVector.String(["hello", "world"]);
 *
 * // where MySerializableStruct is a class you've made that implements Serializable
 * const vecOfSerializableValues = new MoveVector<MySerializableStruct>([
 *   new MySerializableStruct("hello", "world"),
 *   new MySerializableStruct("foo", "bar"),
 * ]);
```

### Test Plan
```shell
cd ~/aptos-core/ecosystem/typescript/sdk_v2/ && pnpm jest bcs-helper.test.ts
```


ecosystem/typescript/sdk_v2/src/bcs/move-types/structs.ts
```diff
@@ -0,0 +1,401 @@
...
+
+  static deserialize(deserializer: Deserializer): MoveString {
+    return new MoveString(deserializer.deserializeStr());
+  }
+}
+
+export class MoveOption<T extends Serializable> extends Serializable {
+  private vec: Vector<T>;
+
+  public value: T | undefined;
```
0xmaayan, line 212:
```suggestion
  public value?: T
```

ecosystem/typescript/sdk_v2/src/bcs/move-types/structs.ts
```diff
@@ -0,0 +1,401 @@
...
+
+  constructor(value?: T) {
+    super();
+    if (typeof value !== "undefined") {
```
0xmaayan, line 216: if making the suggested change (see above) need to change it here to also check for `null`

ecosystem/typescript/sdk_v2/src/bcs/move-types/structs.ts
```diff
@@ -0,0 +1,401 @@
...
+    if (!this.isSome()) {
+      throw new Error("Called unwrap on a MoveOption with no value");
+    } else {
+      return this.value!;
+    }
+  }
+
+  // Check if the MoveOption has a value.
+  isSome(): boolean {
+    return this.value !== undefined;
```
0xmaayan, line 251: check for `null` also?

ecosystem/typescript/sdk_v2/src/bcs/move-types/structs.ts
```diff
@@ -0,0 +1,401 @@
...
+   * @example
+   * MoveOption.U8(1).isSome() === true;
+   * MoveOption.U8().isSome() === false;
+   * MoveOption.U8(undefined).isSome() === false;
+   * @params value: the value used to fill the MoveOption. If `value` is undefined
+   * the resulting MoveOption's .isSome() method will return false.
+   * @returns a MoveOption<U8> with an inner value `value`
+   */
+  static U8(value?: number): MoveOption<U8> {
+    return new MoveOption<U8>(value !== undefined ? new U8(value) : undefined);
```
0xmaayan, line 274: similarly, check for `null`?

ecosystem/typescript/sdk_v2/tests/unit/bcs-helper.test.ts
```diff
@@ -0,0 +1,588 @@
...
+    );
+    expect(deserializedComplexSerializable.myOptionU64.value).toEqual(complexSerializable.myOptionU64.value);
+    expect(deserializedComplexSerializable.myOptionU64.value).toEqual(undefined);
+    expect(deserializedComplexSerializable.myOptionString.value!.value).toEqual(
+      complexSerializable.myOptionString.value!.value,
+    );
+  });
+
+  it("serializes a rotation capability offer struct correctly", () => {
+    class RotationCapabilityOfferProofChallengeV2 extends Serializable {
```
0xmaayan, line 536: a better place for that "test" might be in the future examples folder?

ecosystem/typescript/sdk_v2/tests/unit/bcs-helper.test.ts
```diff
@@ -0,0 +1,588 @@
...
+      public readonly moduleAddress: AccountAddress = AccountAddress.ONE;
+      public readonly moduleName: string = "account";
+      public readonly structName: string = "RotationCapabilityOfferProofChallengeV2";
+      public readonly functionName: string = "offer_rotation_capability";
+
+      constructor(
+        public readonly chainId: number,
+        public readonly sequenceNumber: number,
```
0xmaayan, line 544: sequenceNumber is defined in the sdk as `string | bigint`, how come here is `number`?

0xmaayan (commented):
lets make sure we check for also `null` when a value is optional. I think in ts having an optional variable `consr value? : T` is undefined, but those values can come in as `null` from the "outside" also

ecosystem/typescript/sdk_v2/src/bcs/move-types/structs.ts
```diff
@@ -0,0 +1,398 @@
...
+ *
+ * It is a BCS-serializable, array-like type that contains an array of values of type `T`,
+ * where `T` is a class that implements `Serializable`.
+ *
+ * The purpose of this class is to facilitate easy construction of BCS-serializable
+ * Move `vector<T>` types.
+ *
+ * @example
+ * // in Move: `vector<u8> [1, 2, 3, 4];`
+ * const vecOfU8s = new Vector<U8>([new U8(1), new U8(2), new U8(3), new U8(4)]);
```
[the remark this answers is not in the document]
xbtmatt: so I tried to do this initially, but realized ultimately it's impossible to infer the type and thus serialize it correctly, because you can't use generic type info at runtime.

So I tried both `new Vector([1, 2, 3, 4], U8)` and a few other things, but ultimately realized that this is perfect for a factory method, which is in there:

`Vector.U8([1, 2, 3, 4])`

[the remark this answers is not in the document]
xbtmatt: I played around with enforcing homogenous types when instantiating a vector so you could get strongly typed vectors and not `Vector<Serializable>` but it resulted in other issues (you have to create a `NoUnion<T>` type and figured I'd think about it later

ecosystem/typescript/sdk_v2/tests/unit/bcs-helper.test.ts
xbtmatt: just copied it over from the existing class, think it was made a long time ago, I'll remove it for now and we can put it in an example

ecosystem/typescript/sdk_v2/src/bcs/move-types/primitives.ts
```diff
@@ -0,0 +1,123 @@
...
+  deserialize(deserializer: Deserializer) {
+    this.value = deserializer.deserializeBool();
+  }
+
+  static deserialize(deserializer: Deserializer): Bool {
+    return new Bool(deserializer.deserializeBool());
+  }
+}
+export class U8 extends Serializable {
+  public value: number;
```
0xjinn, line 29: Curious, what happen if the user create an U8 with a number > than u8 can hold? i.e: values that are > 255. Should we do validation on u8 to u256?

0xjinn (approved):
I think this looks good!

TBH, I would like to see some prefix on all of the move type. For example with prefix letter "m", it would `Mu8`, or prefix with word "move" -> "MoveU8". So I know that these are move types. 

Though probably just me lol. Any thoughts?

ecosystem/typescript/sdk_v2/src/bcs/move-types/primitives.ts
xbtmatt: oh yes, good point! We already have a decorator for that, I will add those

xbtmatt: okay so I tried a bunch of ways- but it messes with the underlynig classes/type a ton, so I'm not going to do it.

`serializeX` and `deserializeX` already run the checks- so the only time you'll not see an error is if you create a `U8` that has an invalid number but never serialize it. If they try to deserialize a value that's too large into a `U8` it will also error. so I feel like we don't need to care about the case where they do `new U8(1500)`

I can add checks for it, but it mostly just added more problems than it was worth...

ecosystem/typescript/sdk_v2/src/bcs/serializer.ts
```diff
@@ -310,8 +310,15 @@ export class Serializer {
   }
 }
 
+export function validateNumberInRange<T extends AnyNumber>(value: T, minValue: T, maxValue: T, message?: string) {
+  const valueBigInt = BigInt(value.toString());
+  if (valueBigInt > BigInt(maxValue.toString()) || valueBigInt < BigInt(minValue.toString())) {
+    throw new Error(message || "Value is out of range");
```
0xmaayan, line 316: something like
```suggestion
    throw new Error(message || `${value} is out of range ${minValue} - ${maxValue}`);
```

ecosystem/typescript/sdk_v2/src/bcs/serializer.ts
```diff
@@ -320,13 +327,11 @@ function checkNumberRange<T extends AnyNumber>(minValue: T, maxValue: T, message
   return (target: unknown, propertyKey: string, descriptor: PropertyDescriptor) => {
     const childFunction = descriptor.value;
     // eslint-disable-next-line no-param-reassign
-    descriptor.value = function deco(value: AnyNumber): Serializer {
-      const valueBigInt = BigInt(value.toString());
-      if (valueBigInt > BigInt(maxValue.toString()) || valueBigInt < BigInt(minValue.toString())) {
-        throw new Error(message || "Value is out of range");
-      }
+    descriptor.value = function deco(value: AnyNumber) {
+      validateNumberInRange(value, minValue, maxValue, message);
```
0xmaayan, line 331: what's `message`?

ecosystem/typescript/sdk_v2/src/bcs/move-types/primitives.ts
```diff
@@ -0,0 +1,137 @@
...
+
+  serialize(serializer: Serializer): void {
+    serializer.serializeBool(this.value);
+  }
+
+  static deserialize(deserializer: Deserializer): Bool {
+    return new Bool(deserializer.deserializeBool());
+  }
+}
+export class U8 extends Serializable {
```
0xmaayan, line 32: nit
```suggestion

export class U8 extends Serializable {
```

ecosystem/typescript/sdk_v2/src/bcs/move-types/primitives.ts
```diff
@@ -0,0 +1,137 @@
...
+  }
+
+  serialize(serializer: Serializer): void {
+    serializer.serializeU32(this.value);
+  }
+
+  static deserialize(deserializer: Deserializer): U32 {
+    return new U32(deserializer.deserializeU32());
+  }
+}
```
0xmaayan, line 84: nit
```suggestion

}
```

ecosystem/typescript/sdk_v2/src/bcs/serializer.ts
xbtmatt: whoops, will take that out, was supposed to be an error message for the function

xbtmatt: it's supposed to be a custom optional error message for the decorator, was from the original checkNumberRange sdk v1 function, but yeah we don't use it right now, remove?

ecosystem/typescript/sdk_v2/src/bcs/move-types/primitives.ts
xbtmatt: surprised these got through the formatter

ecosystem/typescript/sdk_v2/src/bcs/serializer.ts
0xmaayan: I mean if we don't use it....

ecosystem/typescript/sdk_v2/src/bcs/move-types/primitives.ts
```diff
@@ -0,0 +1,139 @@
...
+export class Bool extends Serializable {
+  public readonly value: boolean;
+
+  constructor(value: boolean) {
+    super();
+    this.value = value;
+  }
+
+  serialize(serializer: Serializer): void {
+    serializer.serializeBool(this.value);
```
0xmaayan, line 25: should we also check `typeof value == "boolean"`?

ecosystem/typescript/sdk_v2/src/bcs/serializer.ts
```diff
@@ -310,23 +308,36 @@ export class Serializer {
   }
 }
 
+export function ensureBoolean(value: unknown): asserts value is boolean {
+  if (typeof value !== "boolean") {
+    throw new Error("Value needs to be a boolean");
```
0xmaayan, line 313: nit
```suggestion
    throw new Error(`expected value ${value} to be a boolean);
```

0xmaayan (approved):
lgtm

xbtmatt:
@0xmaayan btw I'm moving the TypeTagParser back into typeTag to avoid an import circular dependency so the linter doesn't get mad. Wasn't sure how to fix it quickly so just moved it back in to keep it simple.

Also fixed the multied_25519.ts linter issues too here (just ignored the warnings since it was due to empty `verifySignature` body not implemented yet

xbtmatt:
> I think this looks good!
> 
> TBH, I would like to see some prefix on all of the move type. For example with prefix letter "m", it would `Mu8`, or prefix with word "move" -> "MoveU8". So I know that these are move types.
> 
> Though probably just me lol. Any thoughts?

No, I agree, just not sure what to do with it yet. So far I left the primitives alone, but at least made sure the structs all have `Move` at the front of their names. We can change the names later, will merge it in for now so Maayan can use it :D

0xmaayan:
> > I think this looks good!
> > TBH, I would like to see some prefix on all of the move type. For example with prefix letter "m", it would `Mu8`, or prefix with word "move" -> "MoveU8". So I know that these are move types.
> > Though probably just me lol. Any thoughts?
> 
> No, I agree, just not sure what to do with it yet. So far I left the primitives alone, but at least made sure the structs all have `Move` at the front of their names. We can change the names later, will merge it in for now so Maayan can use it :D

I think the file path is pretty much self explained, but prefix with `Move` every primitive night be good, `MoveU8`, `MoveU16`

PR #9961, opened 2023-09-07
Remove OpenSSL dep from CLI for MacOS

banool:
## Description
My methodology was the following:
1. Find all the openssl related deps in the entire project: `cargo tree --all-features | grep -i openssl`
2. Determine why we have those deps with `cargo tree -i`, e.g. `cargo tree -i openssl-sys`
3. Update those deps to not use `openssl`.

### otool
I'm using otool to check what dynamically linked libraries the binary depends on.

CLI built from main:
```
$ otool -l target/release/aptos | grep openssl
         name /opt/homebrew/opt/openssl@3/lib/libssl.3.dylib (offset 24)
         name /opt/homebrew/opt/openssl@3/lib/libcrypto.3.dylib (offset 24)
$
```

CLI built from this branch:
```
$ otool -l target/release/aptos | grep openssl
$
```

### Virtualized MacOS
I ran a brand new MacOS virtualized with UTM and confirmed that the CLI works out of the box.

CLI built from main:
```
$ ./aptos-main
dyld[1502]: Library not loaded: /opt/homebrew/opt/openssl@3/lib/libssl.3.dylib
  Referenced from: <09857C3A-17E4-3EE4-9935-578128998CA9> /Users/dport/aptos-main
  Reason: tried: '/opt/homebrew/opt/openssl@3/lib/libssl.3.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/opt/homebrew/opt/openssl@3/lib/libssl.3.dylib' (no such file), '/opt/homebrew/opt/openssl@3/lib/libssl.3.dylib' (no such file), '/usr/local/lib/libssl.3.dylib' (no such file), '/usr/lib/libssl.3.dylib' (no such file, not in dyld cache)
zsh: abort      ./aptos-main
```

CLI built from this branch:
```
$ ./aptos-branch       
Command Line Interface (CLI) for developing and interacting with the Aptos blockchain
 
Usage: aptos-branch <COMMAND>
 
Commands:
  account     Tool for interacting with accounts
  config      Tool for interacting with configuration of the Aptos CLI tool
  genesis     Tool for setting up an Aptos chain Genesis transaction
  governance  Tool for on-chain governance
  info        Show build information about the CLI
  init        Tool to initialize current directory for the aptos tool
  key         Tool for generating, inspecting, and interacting with keys
  move        Tool for Move related operations
  multisig    Tool for interacting with multisig accounts
  node        Tool for operations related to nodes
  stake       Tool for manipulating stake and stake pools
  update      Update the CLI itself
  help        Print this message or the help of the given subcommand(s)
 
Options:
  -h, --help     Print help
  -V, --version  Print version
```

### x86 MacOS
One big question is whether this fix is actually due to removing this dep or just due to the fact that I built from an ARM system rather than an x86 system. To test this, I pulled out my old x86 Mac running on MacOS Monterey and built the CLI there and then copied the built binary my native host OS Mac ARM system.

I confirmed that the CLI from this branch worked on both the MacOS on ARM as well as the x86 machine itself (which was a fresh install of the OS without installing openssl through brew). The CLI built there from main did not, with the classic libbl3 path error. So it seems like this PR is indeed the way to remove the OpenSSL dep from the MacOS build of the CLI.

### Ubuntu testing
I tested all of these combinations with Ubuntu:
1. Build on 20, run on 20
2. Build on 20, run on 22
3. Build on 22, run on 20
4. Build on 22, run on 22

2 and 4 do not work. You can tell from this and the output of `lld` that the OpenSSL dep is still present for Linux builds.

This is further evident from the output of `cargo tree`. Whereas the above command was for `cargo tree` on the default target for my platform, if I use `cargo tree --all-features --target x86_64-unknown-linux-gnu | grep -i openssl` you can see that the OpenSSL dep is still there. I suppose there is some dep that only appears on Linux.

In this PR I remove the OpenSSL dep entirely to fix this problem too: https://github.com/aptos-labs/aptos-core/pull/9963 (though a more tactical fix might be possible).

### Indexer
Since the indexer relied on this cloud-storage dep, I ran those tests myself:
```
cd testsuite
poetry run python indexer_grpc_local.py --verbose start --no-indexer-grpc
```
```
cargo nextest run --features integration-tests --package aptos-indexer-grpc-integration-tests
```
They all passed, though I'm not sure these tests will check the cloud-storage path.

I will investigate how to test this more thoroughly

### Other
Waiting to see if CI confirms that this doesn't affect the node or anything else.


banool:
Enabling auto merge to make the tests run.

banool:
I have a `cargo deny` file incoming that will make sure this doesn't regress.

0xjinn:
> otool -l target/release/aptos | grep openssl

Nice command! TIL

Curious, when I check the dependency tree
```
➜  aptos git:(main) ✗ cargo tree -i openssl    
openssl v0.10.50
└── cloud-storage v0.11.1
    └── aptos-indexer-grpc-utils v1.0.0 (/Users/jinhou/Projects/aptos-core/ecosystem/indexer-grpc/indexer-grpc-utils)
        └── aptos-indexer-grpc-fullnode v1.0.0 (/Users/jinhou/Projects/aptos-core/ecosystem/indexer-grpc/indexer-grpc-fullnode)
            └── aptos-node v1.6.0 (/Users/jinhou/Projects/aptos-core/aptos-node)
                └── aptos v2.1.0 (/Users/jinhou/Projects/aptos-core/crates/aptos)
```

I see some indexer stuff depends on cloud-storage. If we disable openssl and use `rustls-tls` feature flag, will aptos-indexer-grpc-utils get impacted?

banool:
Yeah I need to check that that all still works fine.

banool:
Solves https://github.com/aptos-labs/aptos-core/issues/9981.

banool:
Turns out is not possible to add a cargo deny file right now: https://github.com/EmbarkStudios/cargo-deny/issues/563.

banool:
We can almost add a cargo deny file. If we had either the shadow-rs support for rustls or the cargo deny support checking only normal deps we could do it. But it'll have to come in the future. See more here: https://github.com/aptos-labs/aptos-core/pull/10114.

0xjinn (approved):
Amazing!!!

nit: might be good idea to update changelog. Or I can update it during next release.

0xmaayan (approved):

PR #9962, opened 2023-09-07
[BlockSTM] Value exchange infrastructure on reads

georgemitenkov:
Implementation of value exchange for reading. Note that it
is not connected to resolvers because the move type layout
is not propagated to the BlockSTM yet.

- [ ] add tests?

### DIFF
--- a/aptos-move/block-executor/src/executor.rs
+++ b/aptos-move/block-executor/src/executor.rs
@@ -31,7 +31,7 @@ use std::{
     collections::HashMap,
     marker::PhantomData,
     sync::{
-        atomic::AtomicU32,
+        atomic::AtomicU64,
         mpsc,
         mpsc::{Receiver, Sender},
         Arc,
@@ -423,7 +423,7 @@ where
         scheduler: &Scheduler,
         // TODO: should not need to pass base view.
         base_view: &S,
-        shared_counter: &AtomicU32,
+        shared_counter: &AtomicU64,
         role: CommitRole,
     ) {
         // Make executor for each task. TODO: fast concurrent executor.
@@ -529,7 +529,7 @@ where
         assert!(self.concurrency_level > 1, "Must use sequential execution");
 
         let versioned_cache = MVHashMap::new();
-        let shared_counter = AtomicU32::new(0);
+        let shared_counter = AtomicU64::new(0);
 
         if signature_verified_block.is_empty() {
             return Ok(vec![]);
@@ -635,10 +635,7 @@ where
         for (idx, txn) in signature_verified_block.iter().enumerate() {
             let unsync_view = LatestView::<T, S, X>::new(
                 base_view,
-                ViewState::Unsync(SequentialState {
-                    unsync_map: &data_map,
-                    _counter: &0,
-                }),
+                ViewState::Unsync(SequentialState::new(&data_map, 0)),
                 idx as TxnIndex,
             );
             let res = executor.execute_transaction(&unsync_view, txn, idx as TxnIndex, true);
--- a/aptos-move/block-executor/src/liftings.rs
+++ b/aptos-move/block-executor/src/liftings.rs
@@ -1,6 +1,13 @@
 // Copyright © Aptos Foundation
 // SPDX-License-Identifier: Apache-2.0
 
+//! Implementation of sync and unsync versions of data structures which are used
+//! to process aggregator liftings.
+//!
+//! Both have:
+//!   1. A counter used to generate a new unique (per-block) identifier.
+//!   2. A map from identifiers to lifted Move values.
+
 use dashmap::{mapref::one::Ref, DashMap};
 use move_vm_types::values::Value;
 use std::{
--- a/aptos-move/block-executor/src/liftings.rs
+++ b/aptos-move/block-executor/src/liftings.rs
@@ -1,0 +1,58 @@
+// Copyright © Aptos Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+use dashmap::{mapref::one::Ref, DashMap};
+use move_vm_types::values::Value;
+use std::{
+    collections::HashMap,
+    sync::atomic::{AtomicU64, Ordering},
+};
+
+pub(crate) struct UnsyncLiftings {
+    counter: u64,
+    lifted_values: HashMap<u64, Value>,
+}
+
+impl UnsyncLiftings {
+    pub(crate) fn new(counter: u64) -> Self {
+        Self {
+            counter,
+            lifted_values: HashMap::new(),
+        }
+    }
+
+    pub(crate) fn insert(&mut self, value: Value) -> u64 {
+        let identifier = self.counter;
+        self.lifted_values.insert(identifier, value);
+        self.counter += 1;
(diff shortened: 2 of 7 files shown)


aptos-move/block-executor/src/view.rs
```diff
@@ -219,6 +243,58 @@ impl<'a, T: Transaction, S: TStateView<Key = T::Key>, X: Executable> LatestView<
...
+    /// Same as `get_base_value`, but also tries to lift values.
+    fn get_base_value_with_layout(
+        &self,
+        layout: &MoveTypeLayout,
+        state_key: &T::Key,
+    ) -> anyhow::Result<Option<StateValue>> {
+        // Callback for replacing values with identifiers. Right now it is
+        // implemented as a serialization roundtrip.
+        let process_lifting = |bytes: Vec<u8>| -> anyhow::Result<Vec<u8>> {
+            // TODO: alert on fail?
```
igor-aptos, line 257: error!() is alert.

if it is fine in speculative execution, but not in final, then I think speculative_error!() is the macro. @gelash

aptos-move/block-executor/src/view.rs
```diff
@@ -219,6 +243,58 @@ impl<'a, T: Transaction, S: TStateView<Key = T::Key>, X: Executable> LatestView<
         }
         ret
     }
+
+    #[allow(unused)]
+    /// Same as `get_base_value`, but also tries to lift values.
+    fn get_base_value_with_layout(
```
igor-aptos, line 249: it is hard to review without actual place this will be called from being connected

it seems very confusing to me that calling "with_layout" function variant unconditionally performs the exchange. 

exchange will be gated by the AGGREGATOR_EXECUTION flag, and so if the caller will do:

```
if flags.aggregator_execution_enabled() {
   view. get_base_value_with_layout() 
} else {
   view. get_base_value() 
 }
```

I would rename this function to get_base_value_with_aggregator_v2_exchange()


if on the other hand, flag gating check will be performed inside this function - and get_base_value_with_layout will always be called, then function name is ok.

aptos-move/block-executor/src/liftings.rs
```diff
@@ -0,0 +1,77 @@
...
+
+use dashmap::{mapref::one::Ref, DashMap};
+use move_vm_types::values::Value;
+use std::{
+    collections::HashMap,
+    sync::atomic::{AtomicU64, Ordering},
+};
+
+pub(crate) struct UnsyncLiftings {
+    counter: u64,
```
igor-aptos, line 19: does this counter have a separate namespace from aggregator context generating ids? or do we need an Arc and a shared counter?

it's not only sync part that generates ids during native calls

aptos-move/block-executor/src/executor.rs
```diff
@@ -529,7 +529,7 @@ where
         assert!(self.concurrency_level > 1, "Must use sequential execution");
 
         let versioned_cache = MVHashMap::new();
-        let shared_counter = AtomicU32::new(0);
+        let shared_counter = AtomicU64::new(0);
```
vusirikala, line 532: I thought the id would be 32_bit_counter || [metadata such as shard number]. Are we going to have id = 64_bit_counter instead?

georgemitenkov: I think 32_bit_counter || .. but the metadata is not there yet? Will do it with 32-bits and just cast to u64 for now then👍

aptos-move/block-executor/src/liftings.rs
```diff
@@ -0,0 +1,77 @@
...
+    lifted_values: HashMap<u64, Value>,
```
vusirikala, line 20: Just wondering about the terminology. Why are they called lifted values?

aptos-move/block-executor/src/liftings.rs
vusirikala, line 19: Doesn't the counter have to be AtomicU64 instead of u64?

aptos-move/block-executor/src/liftings.rs
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    pub(crate) fn get(&self, identifier: u64) -> Option<&Value> {
+        self.lifted_values.get(&identifier)
+    }
+}
+
+pub(crate) struct SyncLiftings<'a> {
+    counter: &'a AtomicU64,
+    lifted_values: DashMap<u64, Value>,
```
vusirikala, line 52: Why DashMap instead of Hashmap as used in UnsyncLiftings?

aptos-move/block-executor/src/liftings.rs
vusirikala: I believe the aggregator context should also be using this counter for generating ids. We need a way of sharing the counter via traits.

aptos-move/block-executor/src/view.rs
```diff
@@ -161,8 +174,19 @@ impl<'a, T: Transaction, X: Executable> ParallelState<'a, T, X> {
 }
 
 pub(crate) struct SequentialState<'a, T: Transaction, X: Executable> {
-    pub(crate) unsync_map: &'a UnsyncMap<T::Key, T::Value, X>,
-    pub(crate) _counter: &'a u32,
+    unsync_map: &'a UnsyncMap<T::Key, T::Value, X>,
+    unsync_liftings: RefCell<UnsyncLiftings>,
```
vusirikala, line 178: Why RefCell in SequentialState but not in ParallelState?

aptos-move/block-executor/src/liftings.rs
georgemitenkov: Yep, there is `increment` method which should be used by AggregatorResolver trait

aptos-move/block-executor/src/liftings.rs
georgemitenkov: We called them AggregatorLiftings before, so I used that. Didn't use the first aggregator part, should I?

aptos-move/block-executor/src/view.rs
georgemitenkov: This is just to hide both the counter and the hashmap under the same refcell. Otherwise would have moved it inside, maybe it's actually cleaner to do so

aptos-move/block-executor/src/view.rs
```diff
@@ -48,20 +59,22 @@ pub(crate) enum ReadResult<V> {
 pub(crate) struct ParallelState<'a, T: Transaction, X: Executable> {
     versioned_map: &'a MVHashMap<T::Key, T::Value, X>,
     scheduler: &'a Scheduler,
-    _counter: &'a AtomicU32,
+    liftings: SyncLiftings<'a>,
+    lifted_cache: DashMap<T::Key, StateValue>,
```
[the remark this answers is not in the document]
georgemitenkov: Good point!

georgemitenkov:
@igor-aptos @vusirikala after discussing with @gelash we have a nice plan of handling things using MVHashMaps and not the additional data structures, also it's a good point that it's hard to see how everything is connected so I will rebase on top of the PR which refactors views

PR #9832, opened 2023-08-30
[dag] state sync good case unit test

ibalajiarun:

PR #9831, opened 2023-08-30
[dag][bugfix] reverse nodes before adding to dag store

ibalajiarun:
### Description

Fix a bug where the fetched nodes should be reversed before inserting into the DAG.

PR #9830, opened 2023-08-30
[dag] split fetcher into service and helper struct

ibalajiarun:
### Description

This PR splits the DAGFetcher into service and core components. The service is needed for fetching nodes during normal operation and runs in its own task. The core is used for one time fetching of nodes, for example during state sync.

PR #9724, opened 2023-08-22
[dag] preliminary state sync implementation

ibalajiarun:
### Description

This PR provides the preliminary implementation of state sync for the DAG. The integration part will be in another PR.

### Test Plan

Unit test in following PR

PR #9967, opened 2023-09-07
[dag] dag rebootstrap

ibalajiarun:
### Description

This PR introduces the rebootstrap logic for the DAG. Essentially, when there is a need to state sync the DAG, we abort the existing handlers, return to the bootstrapper and let the bootstrapper state sync the DAG and recreate all the components and start the handlers again. This is to unify the logic such that we use the same bootstraping logic for recovery as well as state sync.

### Test Plan

Existing Tests

### DIFF
--- a/consensus/src/dag/bootstrap.rs
+++ b/consensus/src/dag/bootstrap.rs
@@ -16,7 +16,7 @@ use super::{
 };
 use crate::{
     dag::{adapter::NotifierAdapter, dag_fetcher::DagFetcher},
-    experimental::{buffer_manager::OrderedBlocks, commit_reliable_broadcast::DropGuard},
+    experimental::buffer_manager::OrderedBlocks,
     network::IncomingDAGRequest,
     state_replication::{PayloadClient, StateComputer},
 };
@@ -35,10 +35,6 @@ use aptos_types::{
     epoch_state::EpochState,
     ledger_info::{LedgerInfo, LedgerInfoWithSignatures},
     validator_signer::ValidatorSigner,
-};
-use futures::{
-    stream::{AbortHandle, Abortable},
-    FutureExt, StreamExt,
 };
 use futures_channel::{
     mpsc::{UnboundedReceiver, UnboundedSender},
--- a/consensus/src/dag/bootstrap.rs
+++ b/consensus/src/dag/bootstrap.rs
@@ -24,7 +24,7 @@ use aptos_channels::{
     aptos_channel::{self, Receiver},
     message_queues::QueueStyle,
 };
-use aptos_consensus_types::common::{Author, Round};
+use aptos_consensus_types::common::Author;
 use aptos_crypto::HashValue;
 use aptos_infallible::RwLock;
 use aptos_logger::error;
@@ -84,14 +84,13 @@ impl DagBootstrapper {
 
     fn bootstrap_dag_store(
         &self,
-        initial_round: Round,
         latest_ledger_info: LedgerInfo,
         notifier: Arc<dyn Notifier>,
     ) -> (Arc<RwLock<Dag>>, OrderRule) {
         let dag = Arc::new(RwLock::new(Dag::new(
             self.epoch_state.clone(),
             self.storage.clone(),
-            initial_round,
+            latest_ledger_info.round(),
             DAG_WINDOW,
         )));
 
@@ -177,10 +176,7 @@ impl DagBootstrapper {
         ordered_nodes_tx: UnboundedSender<OrderedBlocks>,
         shutdown_rx: oneshot::Receiver<()>,
     ) -> anyhow::Result<()> {
-        let adapter = Arc::new(NotifierAdapter::new(
-            ordered_nodes_tx,
-            self.storage.clone(),
-        ));
+        let adapter = Arc::new(NotifierAdapter::new(ordered_nodes_tx, self.storage.clone()));
 
         let sync_manager = DagStateSynchronizer::new(
             self.epoch_state.clone(),
@@ -204,7 +200,7 @@ impl DagBootstrapper {
 
             // TODO: fix
             let (dag_store, order_rule) =
-                self.bootstrap_dag_store(0, ledger_info.ledger_info().clone(), adapter.clone());
+                self.bootstrap_dag_store(ledger_info.ledger_info().clone(), adapter.clone());
 
             let state_sync_trigger = StateSyncTrigger::new(
                 dag_store.clone(),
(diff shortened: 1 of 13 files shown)


ibalajiarun:
* main
  * **PR #9830** <a href="https://app.graphite.dev/github/pr/aptos-labs/aptos-core/9830" target="_blank">Graphite</a> 
    * **PR #9831** <a href="https://app.graphite.dev/github/pr/aptos-labs/aptos-core/9831" target="_blank">Graphite</a> 
      * **PR #9724** <a href="https://app.graphite.dev/github/pr/aptos-labs/aptos-core/9724" target="_blank">Graphite</a> 
        * **PR #9967** <a href="https://app.graphite.dev/github/pr/aptos-labs/aptos-core/9967" target="_blank">Graphite</a>  👈
        * **PR #9832** <a href="https://app.graphite.dev/github/pr/aptos-labs/aptos-core/9832" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/aptos-labs/aptos-core/9967?utm_source=stack-comment).

consensus/src/dag/bootstrap.rs
```diff
@@ -1,136 +1,303 @@
...
+                    df_handle.abort();
+                    let _ = df_handle.await;
+
+                    let dag_fetcher = DagFetcher::new(self.epoch_state.clone(), self.dag_network_sender.clone(), self.time_service.clone());
+
+                    if let Err(e) = sync_manager.sync_dag_to(&node, dag_fetcher, dag_store.clone()).await {
+                        error!(error = ?e, "unable to sync");
+                    }
+                },
+                _ = handler.start(&mut dag_rpc_rx) => {}
```
zekun000, line 243: I don't think this works? once it starts, it'll never go out from the inner loop

consensus/src/dag/bootstrap.rs
```diff
@@ -1,136 +1,303 @@
...
+            select! {
+                biased;
+                _ = shutdown_rx.select_next_some() => {
+                    df_handle.abort();
+                    let _ = df_handle.await;
+                    return Ok(());
+                },
+                Some(node) = rebootstrap_notification_rx.recv() => {
+                    df_handle.abort();
+                    let _ = df_handle.await;
```
zekun000, line 235: probably should have a guard for these services

consensus/src/dag/bootstrap.rs
ibalajiarun: i think it works. as long as we have an `.await` within the loop, the future should stop polling and yield to another future in the outer select!. If we get rebootstrap notification, it will complete the future and then we will simply not poll handler again. instead, we will start a new handler.

consensus/src/dag/bootstrap.rs
```diff
@@ -1,136 +1,303 @@
...
-    let fetch_requester = Arc::new(fetch_requester);
 
-    let dag_driver = DagDriver::new(
+        // TODO: fetch the correct block info
+        let ledger_info = LedgerInfoWithSignatures::new(
+            LedgerInfo::new(BlockInfo::empty(), HashValue::zero()),
+            AggregateSignature::empty(),
+        );
+
+        let mut shutdown_rx = shutdown_rx.into_stream();
```
zekun000, line 204: I don't think you need this, just &mut shutdown_rx is enough

zekun000 (commented):
looking at it again, why we need to separate the trigger and sync manager? since we block the handler anyway, why not just check and sync directly inside the handler?

ibalajiarun:
> looking at it again, why we need to separate the trigger and sync manager? since we block the handler anyway, why not just check and sync directly inside the handler?

I think I need to abort the fetch service before starting an ad-hoc fetch for state sync. otherwise, there could be two fetches for same nodes happening concurrently?

consensus/src/dag/dag_handler.rs
```diff
@@ -90,10 +107,19 @@ impl NetworkHandler {
...
                 .map(|r| r.into()),
-            DAGMessage::CertifiedNodeMsg(certified_node_msg) => certified_node_msg
-                .verify(&self.epoch_state.verifier)
-                .and_then(|_| self.dag_driver.process(certified_node_msg.certified_node()))
-                .map(|r| r.into()),
+            DAGMessage::CertifiedNodeMsg(certified_node_msg) => {
+                match certified_node_msg.verify(&self.epoch_state.verifier) {
+                    Ok(_) => match self.state_sync_trigger.check(certified_node_msg).await {
+                        ret @ (NeedsSync(_), None) => return Ok(ret.0),
+                        (Synced, Some(certified_node_msg)) => self
```
zekun000, line 114: the message can be carried in the StateSyncStatus::Synced to avoid the second Option?

ibalajiarun: I need to send the Synced status from process_rpc fn as well, so I either clone the ceritifed_node_msg or use the second option. 

Alternatively, I could have two enums, state sync check return one enum and I can convert it to another one for process_rpc. i thought it was too much.

consensus/src/dag/bootstrap.rs
```diff
@@ -1,133 +1,276 @@
...
+                    let _ = df_handle.await;
+                    return Ok(());
+                },
+                certified_node_msg = handler.run(&mut dag_rpc_rx) => {
+                    df_handle.abort();
+                    let _ = df_handle.await;
+
+                    let dag_fetcher = DagFetcher::new(self.epoch_state.clone(), self.dag_network_sender.clone(), self.time_service.clone());
+
+                    if let Err(e) = sync_manager.sync_dag_to(&certified_node_msg, dag_fetcher, dag_store.clone()).await {
```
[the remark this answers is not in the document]
ibalajiarun: yes, we are. I pass the existing dag store to do some assertion checks on whether to actually state sync.

consensus/src/dag/dag_handler.rs
zekun000: I don't think we need to define two enum, this function can just return a Result<(), CertifiedNodeMessage>?

consensus/src/dag/bootstrap.rs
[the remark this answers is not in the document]
zekun000: hmm, that sounds weird, the check should be done in the check function?

zekun000 (approved):

consensus/src/dag/bootstrap.rs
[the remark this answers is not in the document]
ibalajiarun: yes, i am being paranoid. i check in the check function and i assert in the sync_to function.

consensus/src/dag/bootstrap.rs
ibalajiarun: I need this btw, because the oneshot::Receiver seems moved within the select statement.

sasha8 (commented):
Nice!

sasha8 (approved):

consensus/src/dag/dag_state_sync.rs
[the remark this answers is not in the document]
ibalajiarun: yes, when we introduce the onchain config.

PR #9934, opened 2023-09-05
[State Sync] Update metrics and optimize locks.

JoshLind:
### Description
This PR:
1. Replaces the `Mutex<HashMap<...>>` wrapped subscriptions struct with a `DashMap<...>` for improved [performance](https://docs.rs/dashmap/latest/dashmap/) and reduced lock contention.
2. Adds some missing metrics along the way.

Nothing should logically change in this PR.

### Test Plan
Existing test infrastructure.

PR #9970, opened 2023-09-07
[Data Streaming Service] Support missing data chunks in streams.

JoshLind:
Note: (i) this PR is built on https://github.com/aptos-labs/aptos-core/pull/9934; (ii) 90% of this PR is modifications to unit/integration tests.

### Description
This PR updates the data streaming service to better tolerate missing data along each stream, by automatically filling data gaps within the stream. 

Today, when requests are sent to peers for data chunks, peers may choose to return fewer data items than requested (e.g., if the data is too large to fit into a single chunk). In this case, the data stream has "missing data" and the items in the stream are no longer contiguous. This is problematic for performance because it causes all data chunks after the missing data to be invalid, and thus the stream is terminated and a new stream has to be created. This can be expensive and wasteful.

To avoid this, we update the data streaming service to identify when data responses are not complete (e.g., a data request was made for 1000 transactions, but only 500 transactions were returned). In this case, the data streaming service will automatically request the missing data and block the stream until the missing data can be injected into the stream. This has two benefits: (i) streams no longer need to be terminated and created when data is missing; and (ii) data which may already be fetched in the stream is no longer simply thrown away, avoiding unnecessary refetches.

The PR offers two commits:
1. Update the data streaming service to request missing data chunks.
2. Update the unit and integration tests to verify the new functionality. Most of the code lives here 😄 

Note: I've tested this on a devnet PFN and the node is able to sync 30% faster with only around 2-3% of responses containing missing data.

### Test Plan

### DIFF
--- a/state-sync/state-sync-v2/data-streaming-service/src/data_stream.rs
+++ b/state-sync/state-sync-v2/data-streaming-service/src/data_stream.rs
@@ -16,7 +16,7 @@ use crate::{
     error::Error,
     logging::{LogEntry, LogEvent, LogSchema},
     metrics,
-    metrics::{increment_counter, increment_counter_multiple, start_timer},
+    metrics::{increment_counter, increment_counter_multiple_labels, start_timer},
     stream_engine::{DataStreamEngine, StreamEngine},
     streaming_client::{NotificationFeedback, StreamRequest},
 };
@@ -291,10 +291,7 @@ impl<T: AptosDataClientInterface + Send + Clone + 'static> DataStream<T> {
     ) -> PendingClientResponse {
         // Create a new pending client response
         let pending_client_response = Arc::new(Mutex::new(Box::new(
-            data_notification::PendingClientResponse {
-                client_request: data_client_request.clone(),
-                client_response: None,
-            },
+            data_notification::PendingClientResponse::new(data_client_request.clone()),
         )));
 
         // Calculate the request timeout to use, based on the
@@ -317,7 +314,7 @@ impl<T: AptosDataClientInterface + Send + Clone + 'static> DataStream<T> {
             );
 
             // Update the retry counter and log the request
-            increment_counter_multiple(
+            increment_counter_multiple_labels(
                 &metrics::RETRIED_DATA_REQUESTS,
                 data_client_request.get_label(),
                 &request_timeout_ms.to_string(),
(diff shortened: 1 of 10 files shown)


grao1991 (approved):

state-sync/state-sync-v2/data-streaming-service/src/data_stream.rs
```diff
@@ -412,23 +409,53 @@ impl<T: AptosDataClientInterface + Send + Clone + 'static> DataStream<T> {
...
 
+                // Process the client response
                 match client_response {
                     Ok(client_response) => {
-                        if sanity_check_client_response(client_request, &client_response) {
+                        // Sanity check and process the response
+                        if sanity_check_client_response_type(client_request, &client_response) {
+                            // If the response wasn't enough to satisfy the original request (e.g.,
+                            // it was truncated), missing data should be requested.
+                            let missing_data_requested = self.missing_data_requested(
```
bchocho, line 426: Should the function name be more active like `request_missing_data()`? The current name sounds like a simple check rather than actually requesting things

bchocho: Maybe a nit: is there any reason to request missing data before notifying the client of the data? It seems a bit easier to reason if all the data request logic is in one place instead of surrounding the data notification. E.g., I'm not sure if the error handling is correct if we return an error for the data notification but in the meantime have already requested missing data.

state-sync/state-sync-v2/data-streaming-service/src/stream_engine.rs
```diff
@@ -286,24 +286,43 @@ impl DataStreamEngine for StateStreamEngine {
...
                     request.start_index,
                     request.end_index,
                 )?;
 
-                // Update the local stream notification tracker
-                self.next_stream_index = request.end_index.checked_add(1).ok_or_else(|| {
+                // Identify the last received state index and bound it appropriately
+                let last_received_index = match &client_response_payload {
+                    ResponsePayload::StateValuesWithProof(state_values_with_proof) => {
+                        // Verify that we received at least state value
```
bchocho, line 299: typo nit: at least one

state-sync/state-sync-v2/data-streaming-service/src/stream_engine.rs
```diff
@@ -1774,6 +1875,13 @@ impl SubscriptionStream {
     }
 }
 
+/// Bounds the given number by the specified min and max values, inclusive.
+/// If the number is less than the min, the min is returned. If the number is
+/// greater than the max, the max is returned. Otherwise, the number is returned.
+pub(crate) fn bound_by_range(number: u64, min: u64, max: u64) -> u64 {
```
bchocho, line 1881: Should be equivalent to `number.clamp(min, max)`, in which case this utility isn't needed.

JoshLind: 🤦 Amazing, thanks @bchocho! I wondered if something like this existed, but didn't know it was called `clamp!`. Note: I've updated the implementation to use clamp, but still kept the wrapper method because I think it's slightly more readable. Plus, we reference it directly in a bunch of tests, and it feels a little cleaner to know that the implementation and tests are calling the same thing (as opposed to copy/pasting `clamp` everywhere). Does that make sense?

bchocho: IMO just using clamp makes sense. But I don't have a strong preference.

Readability is debatable :) I didn't know about it until recently either, but chatgpt tells me that C++, Python, and Javascript all have built-in `clamp` functions. (Java does not, which is probably why I didn't know about it)

state-sync/state-sync-v2/data-streaming-service/src/data_stream.rs
JoshLind: > Should the function name be more active like request_missing_data()? 

Yeah, that make sense 😄 Updated it to `request_missing_data` and updated the method's comment.

> is there any reason to request missing data before notifying the client of the data?

Aah, yes! I did it to avoid having to clone the response. If you look at the current implementation, we just use references in `request_missing_data` and then `send_data_notification_to_client` takes the actual response. When I flipped it around, it requires me to clone 😢

Let me know what you think. I agree that it would be cleaner to send it first. To be fair, the clone probably isn't that bad 😄

state-sync/state-sync-v2/data-streaming-service/src/stream_engine.rs
JoshLind: Gotcha! I'm inclined to leave it as-is (for now) and then maybe clean things up later 😄

bchocho (approved):
LGTM.

PR #9971, opened 2023-09-07
Modify TokenV2 to add concurrent way of minting tokens using AggregatorV2

igor-aptos:
(recreated from https://github.com/igor-aptos/aptos-core/pull/5)

### DIFF
--- a/aptos-move/framework/aptos-token-objects/sources/collection.move
+++ b/aptos-move/framework/aptos-token-objects/sources/collection.move
@@ -18,11 +18,13 @@
 /// * Add aggregator support when added to framework
 module aptos_token_objects::collection {
     use std::error;
+    use std::features;
     use std::option::{Self, Option};
     use std::signer;
     use std::string::{Self, String};
+    use aptos_framework::aggregator_v2::{Self, Aggregator, AggregatorSnapshot};
     use aptos_framework::event;
-    use aptos_framework::object::{Self, ConstructorRef, Object};
+    use aptos_framework::object::{Self, ConstructorRef, ExtendRef, Object};
 
     use aptos_token_objects::royalty::{Self, Royalty};
 
@@ -40,10 +42,16 @@ module aptos_token_objects::collection {
     const EDESCRIPTION_TOO_LONG: u64 = 5;
     /// The max supply must be positive
     const EMAX_SUPPLY_CANNOT_BE_ZERO: u64 = 6;
+    /// Concurrent feature flag is not yet enabled, so the function cannot be performed
+    const ECONCURRENT_NOT_ENABLED: u64 = 7;
+    /// Tried upgrading collection to concurrent, but collection is already concurrent
+    const EALREADY_CONCURRENT: u64 = 8;
 
     const MAX_COLLECTION_NAME_LENGTH: u64 = 128;
     const MAX_URI_LENGTH: u64 = 512;
     const MAX_DESCRIPTION_LENGTH: u64 = 2048;
+
+    const MAX_U64: u64 = 18446744073709551615;
 
     #[resource_group_member(group = aptos_framework::object::ObjectGroup)]
     /// Represents the common fields for a collection.
@@ -97,13 +105,36 @@ module aptos_token_objects::collection {
         mint_events: event::EventHandle<MintEvent>,
     }
 
+    #[resource_group_member(group = aptos_framework::object::ObjectGroup)]
+    /// Supply tracker, useful for tracking amount of issued tokens.
+    /// If max_value is not set to U64_MAX, this ensures that a limited number of tokens are minted.
+    struct ConcurrentSupply has key {
+        /// Total minted - total burned
+        current_supply: Aggregator<u64>,
+        total_minted: Aggregator<u64>,
+    }
+
     struct BurnEvent has drop, store {
         index: u64,
         token: address,
     }
 
+    #[event]
+    struct ConcurrentBurnEvent has drop, store {
+        collection_addr: address,
+        index: AggregatorSnapshot<u64>,
+        token: address,
+    }
+
     struct MintEvent has drop, store {
         index: u64,
+        token: address,
+    }
+
+    #[event]
+    struct ConcurrentMintEvent has drop, store {
+        collection_addr: address,
+        index: AggregatorSnapshot<u64>,
         token: address,
     }
 
@@ -125,23 +156,40 @@ module aptos_token_objects::collection {
         let constructor_ref = object::create_named_object(creator, collection_seed);
         let object_signer = object::generate_signer(&constructor_ref);
 
-        let supply = FixedSupply {
-            current_supply: 0,
-            max_supply,
-            total_minted: 0,
-            burn_events: object::new_event_handle(&object_signer),
-            mint_events: object::new_event_handle(&object_signer),
-        };
+        if (features::concurrent_assets_enabled()) {
+            let supply = ConcurrentSupply {
+                current_supply: aggregator_v2::create_aggregator(max_supply),
+                total_minted: aggregator_v2::create_unbounded_aggregator(),
+            };
 
-        create_collection_internal(
-            creator,
-            constructor_ref,
-            description,
-            name,
-            royalty,
-            uri,
-            option::some(supply),
-        )
+            create_collection_internal(
+                creator,
+                constructor_ref,
+                description,
+                name,
+                royalty,
+                uri,
+                option::some(supply),
+            )
+        } else {
+            let supply = FixedSupply {
+                current_supply: 0,
+                max_supply,
+                total_minted: 0,
+                burn_events: object::new_event_handle(&object_signer),
+                mint_events: object::new_event_handle(&object_signer),
+            };
+
+            create_collection_internal(
+                creator,
+                constructor_ref,
+                description,
+                name,
+                royalty,
+                uri,
+                option::some(supply),
+            )
+        }
     }
 
     /// Creates an unlimited collection. This has support for supply tracking but does not limit
@@ -157,22 +205,39 @@ module aptos_token_objects::collection {
         let constructor_ref = object::create_named_object(creator, collection_seed);
         let object_signer = object::generate_signer(&constructor_ref);
 
-        let supply = UnlimitedSupply {
-            current_supply: 0,
-            total_minted: 0,
-            burn_events: object::new_event_handle(&object_signer),
-            mint_events: object::new_event_handle(&object_signer),
-        };
+        if (features::concurrent_assets_enabled()) {
+            let supply = ConcurrentSupply {
+                current_supply: aggregator_v2::create_unbounded_aggregator(),
+                total_minted: aggregator_v2::create_unbounded_aggregator(),
+            };
 
-        create_collection_internal(
-            creator,
-            constructor_ref,
-            description,
-            name,
-            royalty,
-            uri,
-            option::some(supply),
-        )
+            create_collection_internal(
+                creator,
+                constructor_ref,
+                description,
+                name,
+                royalty,
+                uri,
+                option::some(supply),
+            )
+        } else {
+            let supply = UnlimitedSupply {
+                current_supply: 0,
+                total_minted: 0,
+                burn_events: object::new_event_handle(&object_signer),
+                mint_events: object::new_event_handle(&object_signer),
+            };
+
+            create_collection_internal(
+                creator,
+                constructor_ref,
+                description,
+                name,
+                royalty,
+                uri,
+                option::some(supply),
+            )
+        }
     }
 
     /// Creates an untracked collection, or a collection that supports an arbitrary amount of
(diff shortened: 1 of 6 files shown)


aptos-move/framework/aptos-token-objects/sources/collection.move
```diff
@@ -125,23 +152,40 @@ module aptos_token_objects::collection {
...
-            constructor_ref,
-            description,
-            name,
-            royalty,
-            uri,
-            option::some(supply),
-        )
+        if (features::concurrent_token_v2_enabled()) {
+            let supply = ConcurrentSupply {
+                current_supply: aggregator_v2::create_aggregator(max_supply),
```
movekevin, line 157: Not the case currently but what would happen if we allow creators to change the max_supply. Is it possible to change it in the aggregator struct? Any implications on conflicts/concurrency etc.?

igor-aptos: We don't have the API for it, but we can add a native function to support it. 

fixed limit is critical to removing conflicts and allowing concurrency, so that transaction would basically be a barrier. As long as the limit is being changed one-off, that should be fine. Changing limit often - i.e. many times within the same block - would be very inefficient.

igor-aptos:
ping @movekevin, @davidiw , @bowenyang007, @junkil-park

aptos-move/framework/aptos-token-objects/sources/token.move
```diff
@@ -77,25 +97,83 @@ module aptos_token_objects::token {
         creator_address: address,
         collection_name: String,
         description: String,
-        name: String,
+        name_prefix: String,
+        name_with_index_suffix: Option<String>,
```
lightmark, line 101: Can we make this a format string with 1 pair of `{}` for index?
This can give user more flexibility of the name format. Such as `aaron_token_{}`

igor-aptos: they can do that too:
name_prefix = "aaron_token_"
name_suffix= option::some("")

currently , I didn't do {}, because I didn't want to support varargs for now, but might want to in the future.

so we might add format() with multiple arguments in the future, and I want to keep the syntax for that time

aptos-move/framework/aptos-token-objects/sources/collection.move
```diff
@@ -97,16 +105,39 @@ module aptos_token_objects::collection {
...
+
     struct BurnEvent has drop, store {
         index: u64,
         token: address,
     }
 
+    #[event]
+    struct ConcurrentBurnEvent has drop, store {
+        collection_addr: address,
+        index: AggregatorSnapshot<u64>,
```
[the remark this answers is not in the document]
igor-aptos: it depends on how many transactions have minted before you - basically index here needs to be exactly in the order in which transactions are ordered in the block.

and even more convoluted, we can reach the collection limit, then some transactions burn a few tokens, and then first ones coming after that can mint successfully.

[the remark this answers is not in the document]
lightmark: Did I get it wrong? This is `ConcurrentBurnEvent` and index is a deterministic argument passed in from the caller. Why do we need a snapshot here...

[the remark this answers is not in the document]
igor-aptos: oh burn event. you are right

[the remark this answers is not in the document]
lightmark: But you ask a very interesting question: mint, burn, and remint... I don't think we want to support that because I have no idea how we assign the right index for remint.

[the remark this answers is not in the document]
igor-aptos: that's already supported :)

we have two counters, index is based on minting order, and supply limits if the new token can be minted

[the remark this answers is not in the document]
lightmark: My question is if we have 1-10 minted and then burn 6. Could 6 be reminted with the right index?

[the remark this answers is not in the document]
igor-aptos: No, index only increases.

aptos-move/framework/aptos-token-objects/sources/token.move
```diff
@@ -77,25 +97,86 @@ module aptos_token_objects::token {
...
+                string::utf8(b"")
+            } else {
+                aggregator_v2::read_snapshot(&name)
+            };
+
+            let token_concurrent = TokenConcurrentFieldsAppendix {
+                index,
+                name,
+            };
+            move_to(&object_signer, token_concurrent);
```
davidiw, line 156: I feel like this could be much cleaner if we just kept the two paths in the if statement and returned back the tuple of index and name.

davidiw (approved):
seems good from the move side.

aptos-move/framework/aptos-token-objects/sources/token.move
```diff
@@ -37,20 +40,37 @@ module aptos_token_objects::token {
...
         /// storage; the URL length will likely need a maximum any suggestions?
         uri: String,
         /// Emitted upon any mutation of the token.
         mutation_events: event::EventHandle<MutationEvent>,
     }
 
+    #[resource_group_member(group = aptos_framework::object::ObjectGroup)]
+    /// Represents first addition to the common fields for all tokens
+    /// Starts being populated once aggregator_v2_api_enabled is enabled.
+    struct TokenConcurrentFieldsAppendix has key {
```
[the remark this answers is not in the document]
igor-aptos: we can still modify the name! :)

but - does name fall into "TokenIdentifiers"? do you have an even better suggestion? :D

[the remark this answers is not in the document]
davidiw: Hard to parse statement, primarily because you should agree with me.

Maybe ConcurrentTokenIdentifiers :grin:

aptos-move/framework/aptos-token-objects/sources/collection.move
```diff
@@ -254,7 +319,20 @@ module aptos_token_objects::collection {
...
+        let index = increment_concurrent_supply(collection, token);
+        if (option::is_none(&index)) {
+            option::none()
+        } else {
+            option::some(aggregator_v2::read_snapshot<u64>(&option::destroy_some(index)))
+        }
+    }
+
+    /// Called by token on mint to increment supply if there's an appropriate Supply struct.
+    public(friend) fun increment_concurrent_supply(
```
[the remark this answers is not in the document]
igor-aptos: is "public(friend) fun" modifiable in a backward incompatible way or not?

Basically, I need to modify the return type from Option<u64> to Option<AggregatorSnapshot<u64>>. If I can do that with increment_supply, I don't need two functions.
But if I cannot modify increment_supply, then I need to deprecate it (I should mark it as deprecated), and create a new one

aptos-move/framework/aptos-token-objects/sources/token.move
igor-aptos: I have the name creation (lines 134-138) that are in common - so if I merge it all the way from index - I would need to duplicate those lines

I could merge deprecated_index and deprecated_name, but those are one-liners, not sure that would be cleaner, but let me know your preference

lightmark (approved):
lgtm

aptos-move/framework/aptos-token-objects/sources/collection.move
movekevin: Just a hypothetical question for token v2 as we don't have a function to update the max supply either :) Good to know it's possible if that ever happens in the future

aptos-move/framework/aptos-token-objects/sources/collection.move
```diff
@@ -125,23 +156,40 @@ module aptos_token_objects::collection {
...
+            create_collection_internal(
+                creator,
+                constructor_ref,
+                description,
+                name,
+                royalty,
+                uri,
+                option::some(supply),
+            )
+        }
```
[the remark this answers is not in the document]
movekevin: I was thinking this but likely wouldn't work because the return types are different which is not allowed by Move

aptos-move/framework/aptos-token-objects/sources/collection.move
```diff
@@ -157,22 +205,39 @@ module aptos_token_objects::collection {
...
-        create_collection_internal(
-            creator,
-            constructor_ref,
-            description,
-            name,
-            royalty,
-            uri,
-            option::some(supply),
-        )
+        if (features::concurrent_assets_enabled()) {
```
movekevin, line 208: Why not refactor this into a separate function for code reuse between this and create_fixed_collection?

aptos-move/framework/aptos-token-objects/sources/collection.move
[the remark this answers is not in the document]
movekevin: public(friend) fun is treated as a private function so it can be freely changed

aptos-move/framework/aptos-token-objects/sources/collection.move
```diff
@@ -324,6 +427,60 @@ module aptos_token_objects::collection {
...
+                max_supply,
+                total_minted,
+                burn_events,
+                mint_events,
+            } = move_from<FixedSupply>(metadata_object_address);
+
+            event::destroy_handle(burn_events);
+            event::destroy_handle(mint_events);
+
+            let supply = ConcurrentSupply {
```
movekevin, line 449: Next 7-8 lines look the same for both branches. How about moving them to after the if/else?

aptos-move/framework/aptos-token-objects/sources/token.move
```diff
@@ -77,25 +97,86 @@ module aptos_token_objects::token {
...
         let collection_addr = collection::create_collection_address(&creator_address, &collection_name);
         let collection = object::address_to_object<Collection>(collection_addr);
-        let id = collection::increment_supply(&collection, signer::address_of(&object_signer));
+
+        // once this flag is enabled, cleanup code for aggregator_api_enabled = false.
+        let aggregator_api_enabled = features::aggregator_v2_api_enabled();
+        let concurrent_assets_enabled = aggregator_api_enabled && features::concurrent_assets_enabled();
+
+        let (deprecated_index, deprecated_name) = if (aggregator_api_enabled) {
+            let index = if (concurrent_assets_enabled) {
```
movekevin, line 124: This can be simplified by combining increment_concurrent_supply and increment_supply

aptos-move/framework/aptos-token-objects/sources/token.move
```diff
@@ -77,25 +97,86 @@ module aptos_token_objects::token {
         creator_address: address,
         collection_name: String,
         description: String,
-        name: String,
+        name_prefix: String,
+        name_with_index_suffix: Option<String>,
```
movekevin, line 101: Does this have to be Option<>? Why not just allowing passing empty string if there's no suffix? This would eliminate some of the option::none checks below to make the code simpler

aptos-move/framework/aptos-token-objects/sources/token.move
```diff
@@ -77,25 +97,86 @@ module aptos_token_objects::token {
...
+            };
+
+            let deprecated_index = if (concurrent_assets_enabled) {
+                0
+            } else {
+                aggregator_v2::read_snapshot(&index)
+            };
+
+            let deprecated_name = if (concurrent_assets_enabled) {
+                string::utf8(b"")
```
movekevin, line 147: Something is strange here. So once we turn on the flags, name will be gone from all tokens? This can be quite a drastic change. What are the considerations here vs getting people to use create_indexed_token if they need indices in the name (and only setting the deprecated index & name fields to empty for that single case)?

aptos-move/framework/aptos-token-objects/sources/token.move
```diff
@@ -229,9 +330,28 @@ module aptos_token_objects::token {
         borrow(&token).description
     }
 
+    // To be added if/when needed
```
movekevin, line 333: What is this for?

aptos-move/framework/aptos-token-objects/sources/token.move
```diff
@@ -280,13 +424,23 @@ module aptos_token_objects::token {
...
             collection,
-            index,
+            index: deprecated_index,
             description: _,
             name: _,
             uri: _,
             mutation_events,
         } = move_from<Token>(addr);
 
+        let index = if (exists<TokenConcurrentFieldsAppendix>(addr)) {
```
movekevin, line 434: There seems to be a lot of duplication with 

if (exists<TokenConcurrentFieldsAppendix>(addr)) {
  borrow_global or borrow_global_mut
} else {
  ...
}
Can we make these into inline functions?

aptos-move/framework/aptos-token-objects/sources/token.move
igor-aptos: @movekevin Concern about that is that people will permanently need to look at both fields, and it is more likely new fields will stay broken for longer. 

This way, after people update indexer, it is going to be uniform for all. 

Cc @bowenyang007 who suggested this approach being better than the alternative

aptos-move/framework/aptos-token-objects/sources/collection.move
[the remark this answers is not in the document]
igor-aptos: I realized why I did it this way - increment_concurrent_supply cannot be called until AGGREGATOR_V2_API flag is enabled. so I need to keep both functions. I'll add TODO to remove it later

aptos-move/framework/aptos-token-objects/sources/collection.move
igor-aptos: it's different - one calls create_unbounded_aggregator, while other aggregator_v2::create_aggregator(max_supply)

aptos-move/framework/aptos-token-objects/sources/collection.move
igor-aptos: moving what is the same, creation of supply object is different though

aptos-move/framework/aptos-token-objects/sources/token.move
igor-aptos: here, since we are in the aggregator_api_enabled , we can combine. will do.

aptos-move/framework/aptos-token-objects/sources/token.move
igor-aptos: we use Option<> here to differentiate whether to create a numbered token or not. (this is private internal create_common function - external APIs are clear, and numbered token doesn't have an option for the suffix.

aptos-move/framework/aptos-token-objects/sources/token.move
igor-aptos: unfortunatelly I don't see a way to make a function to wrap  that - it would need to be able to return different types (i.e. TokenConcurrentFieldsAppendix vs Token) , which I don't think is possible.

igor-aptos:
This was mistakenly closed automatically, recreated here - https://github.com/aptos-labs/aptos-core/pull/10632

PR #9972, opened 2023-09-07
Update fungible assets to use Aggregators for concurrent minting/burning

igor-aptos:


aptos-move/framework/aptos-framework/sources/fungible_asset.move
```diff
@@ -3,7 +3,9 @@
 module aptos_framework::fungible_asset {
     use aptos_framework::event;
     use aptos_framework::object::{Self, Object, ConstructorRef, DeleteRef};
+    use aptos_framework::aggregator_v2::{Self, Aggregator};
```
movekevin, line 6: nit: Sort imports alphabetically. We'll have a formatter that does this soon. For now you could do manually or use the VSCode plugin

aptos-move/framework/aptos-framework/sources/fungible_asset.move
```diff
@@ -177,10 +184,18 @@ module aptos_framework::fungible_asset {
...
             }
         );
-        move_to(metadata_object_signer, Supply {
-            current: 0,
-            maximum: maximum_supply
-        });
+
+        if (features::concurrent_fungible_asset_enabled()) {
+            move_to(metadata_object_signer, ConcurrentSupply {
+                current: aggregator_v2::create_aggregator(option::get_with_default(&mut maximum_supply, MAX_U128)),
```
movekevin, line 190: Shouldn't we still enforce the maximum the user specifies here? It can be a lot smaller than MAX_U128

aptos-move/framework/aptos-framework/sources/fungible_asset.move
```diff
@@ -208,23 +223,29 @@ module aptos_framework::fungible_asset {
...
     /// Get the maximum supply from the `metadata` object.
-    public fun maximum<T: key>(metadata: Object<T>): Option<u128> acquires Supply {
+    public fun maximum<T: key>(metadata: Object<T>): Option<u128> acquires Supply, ConcurrentSupply {
         let metadata_address = object::object_address(&metadata);
         if (exists<Supply>(metadata_address)) {
             let supply = borrow_global<Supply>(metadata_address);
             supply.maximum
+        } else if (exists<ConcurrentSupply>(metadata_address)) {
+            let supply = borrow_global<ConcurrentSupply>(metadata_address);
+            option::some(aggregator_v2::max_value(&supply.current))
```
movekevin, line 248: Similar to previous comment, this should return the user-defined maximum, not the MAX_U128

aptos-move/framework/move-stdlib/sources/configs/features.move
```diff
@@ -251,6 +251,22 @@ module std::features {
         is_enabled(AGGREGATOR_SNAPSHOTS)
     }
 
+    /// Whether enable MOVE functions to call create_auid method to create AUIDs.
+    /// Lifetime: transient
+    const CONCURRENT_TOKEN_V2_ENABLED: u64 = 31;
+
+    public fun concurrent_token_v2_enabled(): bool acquires Features {
```
movekevin, line 258: I wonder if we can just use a single flag for both? They likely would be enabled in the same proposal anyway

movekevin (commented):
Can you add some brief description + related AIP for this change? :)

aptos-move/framework/aptos-framework/sources/fungible_asset.move
igor-aptos: aggregator enforces the maximum - create_aggregator has max_value argument. I pass there maximum_supply if set, or MAX_U128 if maximum_supply is None

aptos-move/framework/move-stdlib/sources/configs/features.move
igor-aptos: I thought of having more flexibility with both, but we can combine as well.

aptos-move/framework/aptos-framework/sources/fungible_asset.move
igor-aptos: aggregator_v2::max_value is function that gets the value set in the aggregator, not MAX_U128.

though there is one logical change here I missed - if user passed None, we have set MAX_U128 on the aggregator, and would return MAX_U128 here. I can document that - if you think is reasonable, or think of how to return None otherwise

aptos-move/framework/aptos-framework/sources/fungible_asset.move
movekevin: Gotcha is it easy to add unit tests for different cases here?

aptos-move/framework/aptos-framework/sources/fungible_asset.move
movekevin: Yes some documentation would be useful

movekevin (commented):
Can we provide a function to allow migrating from normal supply to concurrent supply. You could potential use ExtendRef here to guard this action (there might be more options as well)

igor-aptos:
yeah, will add a migration function, have it as todo for the token 2 PR as well

aptos-move/framework/aptos-framework/sources/fungible_asset.move
georgemitenkov: @movekevin what is the use case for having `Option<u128>` for max supply? It is either a real MAX_U128 in case of none, or a user-define value also capped at MAX_U128, I do not see why would we ever want to differentiate between the two apart from having an error code and a few more instructions to check. From gas costs perspective this also does not save much

igor-aptos:
ping @movekevin, @davidiw , @bowenyang007, @junkil-park

aptos-move/framework/aptos-framework/sources/fungible_asset.move
```diff
@@ -206,25 +231,44 @@ module aptos_framework::fungible_asset {
...
     /// Get the current supply from the `metadata` object.
-    public fun supply<T: key>(metadata: Object<T>): Option<u128> acquires Supply {
+    public fun supply<T: key>(metadata: Object<T>): Option<u128> acquires Supply, ConcurrentSupply {
         let metadata_address = object::object_address(&metadata);
         if (exists<Supply>(metadata_address)) {
             let supply = borrow_global<Supply>(metadata_address);
             option::some(supply.current)
+        } else if (exists<ConcurrentSupply>(metadata_address)) {
+            let supply = borrow_global<ConcurrentSupply>(metadata_address);
+            option::some(aggregator_v2::read(&supply.current))
```
davidiw, line 252: if we expect this to dominate, re-order.

aptos-move/framework/aptos-framework/sources/fungible_asset.move
```diff
@@ -206,25 +231,44 @@ module aptos_framework::fungible_asset {
...
 
     #[view]
     /// Get the maximum supply from the `metadata` object.
-    public fun maximum<T: key>(metadata: Object<T>): Option<u128> acquires Supply {
+    public fun maximum<T: key>(metadata: Object<T>): Option<u128> acquires Supply, ConcurrentSupply {
         let metadata_address = object::object_address(&metadata);
         if (exists<Supply>(metadata_address)) {
             let supply = borrow_global<Supply>(metadata_address);
             supply.maximum
+        } else if (exists<ConcurrentSupply>(metadata_address)) {
```
davidiw, line 265: same, or at least a todo

davidiw (approved):

aptos-move/framework/aptos-framework/sources/fungible_asset.move
```diff
@@ -52,6 +54,8 @@ module aptos_framework::fungible_asset {
     const ESUPPLY_UNDERFLOW: u64 = 20;
     /// Supply resource is not found for a metadata object.
     const ESUPPLY_NOT_FOUND: u64 = 21;
+    /// Flag for Concurrent Supply not enabled
+    const ECONCURRENT_NOT_ENABLED: u64 = 22;
```
lightmark, line 58:
```suggestion
    const ECONCURRENT_SUPPLY_NOT_ENABLED: u64 = 22;
```

aptos-move/framework/aptos-framework/sources/fungible_asset.move
```diff
@@ -177,10 +188,24 @@ module aptos_framework::fungible_asset {
                 project_uri,
             }
         );
-        move_to(metadata_object_signer, Supply {
-            current: 0,
-            maximum: maximum_supply
-        });
+
+        if (features::concurrent_assets_enabled()) {
+            let unlimited = !option::is_some(&maximum_supply);
```
lightmark, line 193: why not `is_none()`?

aptos-move/framework/aptos-framework/sources/fungible_asset.move
```diff
@@ -206,25 +231,44 @@ module aptos_framework::fungible_asset {
         TransferRef { metadata }
     }
 
+    /// Creates a extend ref that can be used to upgrade to concurrent from the given fungible
+    /// object's constructor ref.
+    /// This can only be called at object creation time as constructor_ref is only available then.
+    public fun generate_extend_ref(constructor_ref: &ConstructorRef): ExtendRef {
+        object::generate_extend_ref(constructor_ref)
+        // let metadata = object::object_from_constructor_ref<Metadata>(constructor_ref);
+        // ExtendRef { metadata }
```
lightmark, line 240: nits

aptos-move/framework/aptos-framework/sources/fungible_asset.move
```diff
@@ -206,25 +231,44 @@ module aptos_framework::fungible_asset {
...
         let metadata_address = object::object_address(&metadata);
         if (exists<Supply>(metadata_address)) {
             let supply = borrow_global<Supply>(metadata_address);
             supply.maximum
+        } else if (exists<ConcurrentSupply>(metadata_address)) {
+            let supply = borrow_global<ConcurrentSupply>(metadata_address);
+            if (supply.unlimited) {
+                option::none()
+            } else {
+                option::some(aggregator_v2::max_value(&supply.current))
```
lightmark, line 270: I wonder Can't AG itself tell whether it's unlimited or not?

aptos-move/framework/aptos-framework/sources/fungible_asset.move
```diff
@@ -535,32 +579,54 @@ module aptos_framework::fungible_asset {
...
+            let supply = borrow_global_mut<Supply>(metadata_address);
+            if (option::is_some(&supply.maximum)) {
+                let max = *option::borrow_mut(&mut supply.maximum);
+                assert!(
+                    max - supply.current >= (amount as u128),
+                    error::out_of_range(EMAX_SUPPLY_EXCEEDED)
+                )
+            };
+            supply.current = supply.current + (amount as u128);
+        } else if (exists<ConcurrentSupply>(metadata_address)) {
```
lightmark, line 596: I guess DW wants short-circuiting here too lol

lightmark (approved):
lgtm in general except for some nits

aptos-move/framework/aptos-framework/sources/fungible_asset.move
igor-aptos: I was wondering, why do we do, is there a reason:

let metadata = object::object_from_constructor_ref<Metadata>(constructor_ref);
        TransferRef { metadata }
       
above, instead of object::generate_transfer_ref(constructor_ref)?

aptos-move/framework/aptos-framework/sources/fungible_asset.move
igor-aptos: width of the field implicitly defines the max value. so internally, unlimited means max_u64.

If we want to be able to distinguish between max_u64 and unlimited asset (i.e. to show in the UI "unlimited" vs "max_u64") - we need this boolean.

but that might be better to do in the UI, not here though

aptos-move/framework/aptos-framework/sources/fungible_asset.move
lightmark: wait, this has to be fixed...
`object::generate_extend_ref(constructor_ref)` return `object::ExtendRef`.
`TransferRef` here is `fungible_asset::TransferRef`.
cc @davidiw ... proof that caps is a better name than refs here cuz ppl will be confused......

igor-aptos: so we have two different things referred to as TransferRef, that mean different things ?  it might have been better to name fungible_asset::TransferRef as AssetTransferRef or something?

but that means that I don't need this function at all, I will just directly call object::generate_extend_ref in the test (and I assume creators of fungible assets can do the same)

lightmark: yeah, this one seems unnecessary to me cuz you could just get it from `object::generate_extend_ref`.
`TransferRef` here means transfer FA between FungibleStore whileas `object::TransferRef` is for transferring objects.
I named it as `TransferCap` initially, but DW thought it is better to be consistent with object module.

aptos-move/framework/aptos-framework/sources/fungible_asset.move
lightmark: > but that might be better to do in the UI, not here though
That's what I meant... since we have an upper bound anyway so in the code the max value is always there. UI can treat max_u64 differently tho.

igor-aptos: It was like that,  and than I added the flag.  But we don't have a flag in token.  So for consistency- I agree, let me remove the flag

aptos-move/framework/aptos-framework/sources/fungible_asset.move
igor-aptos: Updated

aptos-move/framework/aptos-framework/sources/fungible_asset.move
```diff
@@ -535,32 +570,54 @@ module aptos_framework::fungible_asset {
...
+            if (option::is_some(&supply.maximum)) {
+                let max = *option::borrow_mut(&mut supply.maximum);
+                assert!(
+                    max - supply.current >= (amount as u128),
+                    error::out_of_range(EMAX_SUPPLY_EXCEEDED)
+                )
+            };
+            supply.current = supply.current + (amount as u128);
+        } else {
+            assert!(false, error::not_found(ESUPPLY_NOT_FOUND));
```
movekevin, line 594: Use abort instead of assert!(false)

movekevin (approved):

igor-aptos:
this was mistakenly closed automatically - recreated here - https://github.com/aptos-labs/aptos-core/pull/10633

PR #9975, opened 2023-09-08
[NFT Metadata Crawler] Improve metrics for parser invocation and termination

just-in-chang:
### Description
Improves the metrics so there is a clearer comparison for how a parser is terminated


larry-aptos (approved):

GhostWalker562 (approved):
LGTM 🚀

PR #9976, opened 2023-09-08
fix ans publish

0xaptosj:


0xjinn:
Is this the same as https://github.com/aptos-labs/aptos-core/pull/9980?

0xaptosj:
> Is this the same as #9980?

yes, don't mind me, im just testing why this didn't work last night, will close soon

PR #9986, opened 2023-09-08
[aptos-logger] truncate long log messages and field values

ibalajiarun:
### Description

Avoid printing too much.

### Test Plan

Unit test.

### DIFF
--- a/crates/aptos-logger/src/aptos_logger.rs
+++ b/crates/aptos-logger/src/aptos_logger.rs
@@ -44,29 +44,39 @@ const FLUSH_TIMEOUT: Duration = Duration::from_secs(5);
 const FILTER_REFRESH_INTERVAL: Duration =
     Duration::from_secs(5 /* minutes */ * 60 /* seconds */);
 
-struct TruncatedString(String);
+/// Note: To disable length limits, set `RUST_LOG_FIELD_MAX_LEN` to -1.
+const RUST_LOG_FIELD_MAX_LEN_ENV_VAR: &str = "RUST_LOG_FIELD_MAX_LEN";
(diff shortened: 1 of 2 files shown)


crates/aptos-logger/src/aptos_logger.rs
```diff
@@ -1025,4 +1075,50 @@ mod tests {
                 "source_path"
             )));
     }
+
+    #[test]
+    fn test_log_event_truncation() {
```
perryjrandall, line 1080: beautiful testing <3

perryjrandall (approved):

crates/aptos-logger/src/aptos_logger.rs
```diff
@@ -43,6 +44,48 @@ const FLUSH_TIMEOUT: Duration = Duration::from_secs(5);
 const FILTER_REFRESH_INTERVAL: Duration =
     Duration::from_secs(5 /* minutes */ * 60 /* seconds */);
 
+struct TruncatedString(String);
+
+impl TruncatedString {
+    const MAX_LEN: usize = 128;
```
JoshLind, line 50: How did you come to this length? i.e., we have a lot of strings that seem larger than this in Humio already: https://cloud.us.humio.com/telemetry-service-ingest/search?live=false&query=%22%23chain_id%22%3Dmainnet&start=7d&tz=America%2FLos_Angeles.

My feeling would be to 100x it, perhaps? e.g., 10KB? @zekun000, any suggestions/preferences?

sitalkedia: +1, @ibalajiarun - Before shipping this, can you take a look at our humio log to see what % of the logs will be trucated after this change and ensure nothing super critical is being trucated?

sitalkedia (approved):
LG

PR #9989, opened 2023-09-09
[forge] Split duration more usefully, for more reliable stats

igor-aptos:
We have a sweep of tests, with some being in 1 phase and some in 2. Previously, each test would get ~5 minutes, giving it 2.5 minutes per phase.

With this, we make all phases equal, i.e. ~4 minutes, so we get similarly stable results for each phase, as we get for full runs, without needing to extend runtime.

(this is for workload sweep test)

### DIFF
--- a/testsuite/testcases/src/load_vs_perf_benchmark.rs
+++ b/testsuite/testcases/src/load_vs_perf_benchmark.rs
@@ -13,6 +13,11 @@ use rand::SeedableRng;
 use std::{fmt::Debug, time::Duration};
 use tokio::runtime::Runtime;
 
+// add larger warmup, as when we are exceeding the max load,
+// it takes more time to fill mempool.
+const PER_TEST_WARMUP_DURATION_FRACTION: f32 = 0.2;
+const PER_TEST_COOLDOWN_DURATION_FRACTION: f32 = 0.05;
+
 pub struct SingleRunStats {
     name: String,
     stats: TxnStats,
@@ -23,8 +28,8 @@ pub struct SingleRunStats {
 
 #[derive(Debug)]
 pub enum Workloads {
-    TPS(&'static [usize]),
-    TRANSACTIONS(&'static [TransactionWorkload]),
+    TPS(Vec<usize>),
+    TRANSACTIONS(Vec<TransactionWorkload>),
 }
 
 impl Workloads {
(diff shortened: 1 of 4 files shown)


bchocho (approved):
LGTM

testsuite/testcases/src/load_vs_perf_benchmark.rs
```diff
@@ -42,6 +43,17 @@ impl Workloads {
         }
     }
 
+    fn num_phases(&self, index: usize) -> usize {
+        match self {
+            Self::TPS(_) => 1,
+            Self::TRANSACTIONS(workloads) => if workloads[index].is_phased() {
```
JoshLind, line 49: nit: Not an issue with this PR, but I had to look up `is_phased()` to determine why some workloads get 1 phase vs. 2 phases. It makes me wonder if we'd be better off changing this method name, or adding a new method to the workload itself? e.g., `num_phases`, which hides this logic? But, no biggie.

testsuite/testcases/src/load_vs_perf_benchmark.rs
```diff
@@ -13,6 +13,7 @@ use rand::SeedableRng;
 use std::{fmt::Debug, time::Duration};
 use tokio::runtime::Runtime;
 
+
```
JoshLind, line 16: Was this deliberate? 😄

testsuite/testcases/src/load_vs_perf_benchmark.rs
```diff
@@ -67,9 +79,16 @@ impl Workloads {
             Self::TRANSACTIONS(workloads) => workloads[index].configure(request),
         }
     }
+
+    fn split_duration(&self, global_duration: Duration) -> (Duration, Duration) {
+        let total_phases: usize = (0..self.len()).into_iter().map(|index| self.num_phases(index)).sum();
+        let phase_duration = global_duration.div_f32(total_phases as f32 + 0.2 * (self.len() - 1) as f32);
```
JoshLind, line 85: nit: might be worth making `0.2` a const, to make it obvious that it's the duration buffer we're using? 😄 That way, you can also reference it in tests.

JoshLind (commented):
LGTM.

JoshLind (approved):

PR #9994, opened 2023-09-11
[consensus] epoch manager refactoring; dag is coming

ibalajiarun:
### Description

This PR refactors the epoch manager and extracts common implementation logic into separate functions, so it can be used when instantiating the DAG. The functionality remains the same, with the only change that it now groups all prometheus metrics update logic in start_round_manager into its own method.

### DIFF
--- a/consensus/src/epoch_manager.rs
+++ b/consensus/src/epoch_manager.rs
@@ -38,6 +38,7 @@ use crate::{
     },
     network_interface::{ConsensusMsg, ConsensusNetworkClient},
     payload_client::QuorumStoreClient,
+    payload_manager::PayloadManager,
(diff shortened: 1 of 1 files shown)


consensus/src/epoch_manager.rs
```diff
@@ -794,6 +828,15 @@ impl<P: OnChainConfigProvider> EpochManager<P> {
         self.spawn_block_retrieval_task(epoch, block_store);
     }
 
+    fn init_network_sender(&mut self, epoch_state: &EpochState) -> NetworkSender {
```
zekun000, line 831: doesn't need to be mut?

zekun000 (approved):

sitalkedia (approved):

sasha8 (approved):
LGTM

Issue #9580, opened 2023-08-09
[Bug] [Combined] Add clarifying error messages for all VMErrors


movekevin:
@wrwg added the ability to attach an error message to VMErrors so we should go through and add errors messages to all common errors. Community developers have repeatedly run into the errors and could not debug exactly why an error happens, for example:
https://github.com/aptos-labs/aptos-core/issues/8078
https://github.com/aptos-labs/aptos-core/issues/6232

Here's a list of the most common errors that are almost impossible to debug without clear error messages (to be expanded more as more are reported):

- [x] #9999 
- [ ] CONSTRAINT_NOT_SATISFIED
- [x] BACKWARD_INCOMPATIBLE_MODULE_UPDATE
- [ ] #9998 
- [x] RESOURCE_DOES_NOT_EXIST (doesn't say which resource doesn't)
- [x] RESOURCE_ALREADY_EXISTS (doesn't say which)
- [x] ARITHMETIC_ERROR (doesn't say overflow, underflow or division by zero)

brmataptos:
Is it possible we can also add a feature (at least when running outside the blockchain) to specify details of errors?  For example, LINKER_ERROR can say what function is missing.  INCOMPATIBLE_TYPES can tell what the two types are.  etc.  Can we roll that into this bug or should I file a new one?

movekevin:
Great idea! The error messages cannot contain a lot of details such as stack traces. Let’s create a separate ticket for that though as I’m also not sure which layer the extra details should be generated - vm, api, or cli

brmataptos:
Here's an example error that might be made into a test case.  This is resulting from an inlining bug, but perhaps it can be converted into .mvir.  Btw, if you can identify the problem please let me know.

[unknown-invariant-violation-error-bytecode.txt](https://github.com/aptos-labs/aptos-core/files/12562814/unknown-invariant-violation-error-bytecode.txt)

wrwg:
Not sure what you guys mean, because this feature — adding a more detailed error message — is already there.

brmataptos:
So..... this error is very clear and detailed?

```
task 1 'run'. lines 93-93:
Error: Function execution failed with VMError: {
    major_status: UNKNOWN_INVARIANT_VIOLATION_ERROR,
    sub_status: Some(2),
    location: 0xcafe::vectors,
    indices: [],
    offsets: [(FunctionDefinitionIndex(0), 27)],
}
```

wrwg:
[...]

PR #10000, opened 2023-09-11
[move] Add error msgs for a few error types

runtian-zhou:
### Description

Added error messages for a few errors. Fixing #9580 

This PR should cover
- RESOURCE_DOES_NOT_EXIST
- RESOURCE_ALREADY_EXISTS
- ARITHMETIC_ERROR 
- BACKWARD_INCOMPATIBLE_MODULE_UPDATE

### Test Plan
Updated tests to make sure error messages gets generated.

### DIFF
--- a/third_party/move/move-binary-format/src/compatibility.rs
+++ b/third_party/move/move-binary-format/src/compatibility.rs
@@ -192,17 +192,17 @@ impl Compatibility {
         if self.check_struct_and_pub_function_linking && !struct_and_pub_function_linking {
             return Err(PartialVMError::new(
                 StatusCode::BACKWARD_INCOMPATIBLE_MODULE_UPDATE,
-            ));
+            ).with_message("Module Update Failure: Public function/struct signature of new module differs from existing module".to_string()));
         }
         if self.check_struct_layout && !struct_layout {
             return Err(PartialVMError::new(
                 StatusCode::BACKWARD_INCOMPATIBLE_MODULE_UPDATE,
-            ));
+            ).with_message("Module Update Failure: Struct layout of new module differs from existing module".to_string()));
         }
         if self.check_friend_linking && !friend_linking {
             return Err(PartialVMError::new(
                 StatusCode::BACKWARD_INCOMPATIBLE_MODULE_UPDATE,
-            ));
+            ).with_message("Module Update Failure: Friend signature of new module differs from existing module".to_string()));
         }
 
         Ok(())
--- a/third_party/move/move-binary-format/src/compatibility.rs
+++ b/third_party/move/move-binary-format/src/compatibility.rs
@@ -192,17 +192,17 @@ impl Compatibility {
         if self.check_struct_and_pub_function_linking && !struct_and_pub_function_linking {
             return Err(PartialVMError::new(
                 StatusCode::BACKWARD_INCOMPATIBLE_MODULE_UPDATE,
-            ).with_message("Module Update Failure: Public function/struct signature of new module differs from existing module".to_string()));
+            ).with_message(format!("Module Update Failure: Public function/struct signature of new module differs from existing module in {:?}::{}", old_module.address, old_module.name)));
         }
         if self.check_struct_layout && !struct_layout {
             return Err(PartialVMError::new(
                 StatusCode::BACKWARD_INCOMPATIBLE_MODULE_UPDATE,
-            ).with_message("Module Update Failure: Struct layout of new module differs from existing module".to_string()));
+            ).with_message(format!("Module Update Failure: Struct layout of new module differs from existing modul in {:?}::{}", old_module.address, old_module.name)));
         }
         if self.check_friend_linking && !friend_linking {
             return Err(PartialVMError::new(
                 StatusCode::BACKWARD_INCOMPATIBLE_MODULE_UPDATE,
-            ).with_message("Module Update Failure: Friend signature of new module differs from existing module".to_string()));
+            ).with_message(format!("Module Update Failure: Friend signature of new module differs from existing module in {:?}::{}", old_module.address, old_module.name)));
         }
 
         Ok(())
--- a/third_party/move/move-vm/runtime/src/interpreter.rs
+++ b/third_party/move/move-vm/runtime/src/interpreter.rs
@@ -576,7 +576,9 @@ impl Interpreter {
             TypeWithLoader { ty, loader },
             res.is_ok(),
         )?;
-        self.operand_stack.push(res?)?;
+        self.operand_stack.push(res.map_err(|err| {
+            err.with_message(format!("Failed to borrow global resource from {:?}", addr))
+        })?)?;
         Ok(())
     }
 
@@ -620,7 +622,7 @@ impl Interpreter {
                 Err(err) => {
                     let val: Option<&Value> = None;
                     gas_meter.charge_move_from(is_generic, TypeWithLoader { ty, loader }, val)?;
-                    return Err(err);
+                    return Err(err.with_message(format!("Failed to move resource from {:?}", addr)));
                 },
             };
         self.operand_stack.push(resource)?;
@@ -658,7 +660,7 @@ impl Interpreter {
                     &resource,
                     false,
                 )?;
-                Err(err)
+                Err(err.with_message(format!("Failed to move resource into {:?}", addr)))
             },
         }
     }
--- a/third_party/move/move-vm/runtime/src/interpreter.rs
+++ b/third_party/move/move-vm/runtime/src/interpreter.rs
@@ -609,22 +609,23 @@ impl Interpreter {
         addr: AccountAddress,
         ty: &Type,
     ) -> PartialVMResult<()> {
-        let resource =
-            match Self::load_resource(loader, data_store, gas_meter, addr, ty)?.move_from() {
-                Ok(resource) => {
-                    gas_meter.charge_move_from(
-                        is_generic,
-                        TypeWithLoader { ty, loader },
-                        Some(&resource),
-                    )?;
-                    resource
-                },
-                Err(err) => {
-                    let val: Option<&Value> = None;
-                    gas_meter.charge_move_from(is_generic, TypeWithLoader { ty, loader }, val)?;
-                    return Err(err.with_message(format!("Failed to move resource from {:?}", addr)));
-                },
-            };
+        let resource = match Self::load_resource(loader, data_store, gas_meter, addr, ty)?
+            .move_from()
+        {
+            Ok(resource) => {
+                gas_meter.charge_move_from(
+                    is_generic,
+                    TypeWithLoader { ty, loader },
+                    Some(&resource),
+                )?;
+                resource
+            },
+            Err(err) => {
+                let val: Option<&Value> = None;
+                gas_meter.charge_move_from(is_generic, TypeWithLoader { ty, loader }, val)?;
+                return Err(err.with_message(format!("Failed to move resource from {:?}", addr)));
+            },
+        };
         self.operand_stack.push(resource)?;
         Ok(())
     }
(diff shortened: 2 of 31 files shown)


vineethk (approved):

vineethk:
@runtian-zhou It is not immediately clear if this resolves `ARITHMETIC_ERROR` as suggested in the PR description.

runtian-zhou:
I just discovered I missed that commit...

third_party/move/move-vm/types/src/values/values_impl.rs
```diff
@@ -1591,31 +1606,36 @@ impl IntegerValue {
         Ok(match self {
             U8(x) => {
                 if n_bits >= 8 {
-                    return Err(PartialVMError::new(StatusCode::ARITHMETIC_ERROR));
+                    return Err(PartialVMError::new(StatusCode::ARITHMETIC_ERROR)
```
rahxephon89, line 1609: I am wondering whether we can use a static string constant to represent the error string such as "Shift Left overflow"?

runtian-zhou: I don't think you can do that? We only have `&'static str` in rust, according to https://stackoverflow.com/questions/32956050/how-to-repeat-a-static-string-at-compile-time

rahxephon89: Sorry for the confusion. I mean something like this: https://github.com/aptos-labs/aptos-core/blob/e0a42ece1909baa8298b24b8e4a12010825051a3/crates/aptos-inspection-service/src/server/mod.rs#L42C7-L42C7

runtian-zhou: Yea my point is even if you define err msgs as `const` you still need to invoke `.to_string()` which at this point I'm not sure how that differs from having a string literal like what we have for now.

third_party/move/testing-infra/transactional-test-runner/src/vm_test_harness.rs
```diff
@@ -329,37 +335,73 @@ impl<'a> MoveTestAdapter<'a> for SimpleVMTestAdapter<'a> {
...
-            format!("{:?}", e.indices())
-        },
-        offsets = if comparison_mode {
-            // During comparison testing, abstract this data.
-            "redacted".to_string()
-        } else {
-            format!("{:?}", e.offsets())
-        },
-    )
+            major_status = e.major_status(),
```
rahxephon89, line 359: Is it possible to reduce duplicated code by concatenating normal error and the verbose message?

third_party/move/move-vm/types/src/values/values_impl.rs
```diff
@@ -1466,7 +1466,10 @@ impl IntegerValue {
                 return Err(PartialVMError::new(StatusCode::INTERNAL_TYPE_ERROR).with_message(msg));
             },
         };
-        res.ok_or_else(|| PartialVMError::new(StatusCode::ARITHMETIC_ERROR))
+        res.ok_or_else(|| {
+            PartialVMError::new(StatusCode::ARITHMETIC_ERROR)
+                .with_message("Add overflow".to_string())
```
vineethk, line 1471: Consider changing this to "Addition" (and similarly, "Subtraction" below) to match with "Multiplication".

third_party/move/move-vm/types/src/values/values_impl.rs
vineethk: A possibility to consider is a private method that takes a string slice and returns the appropriate `Err` object: this may reduce some code duplication.

I also learnt that Rust already interns duplicate string literals, so there should be no binary size penalty for using duplicate string literals.

vineethk (commented):
Adding a couple of comments due to new files added to PR.

third_party/move/move-binary-format/src/compatibility.rs
```diff
@@ -192,17 +192,17 @@ impl Compatibility {
         if self.check_struct_and_pub_function_linking && !struct_and_pub_function_linking {
             return Err(PartialVMError::new(
                 StatusCode::BACKWARD_INCOMPATIBLE_MODULE_UPDATE,
-            ));
+            ).with_message("Module Update Failure: Public function/struct signature of new module differs from existing module".to_string()));
```
movekevin, line 195: Can we include the name of the mismatching function?

third_party/move/move-binary-format/src/compatibility.rs
```diff
@@ -192,17 +192,17 @@ impl Compatibility {
...
         }
         if self.check_struct_layout && !struct_layout {
             return Err(PartialVMError::new(
                 StatusCode::BACKWARD_INCOMPATIBLE_MODULE_UPDATE,
-            ));
+            ).with_message("Module Update Failure: Struct layout of new module differs from existing module".to_string()));
```
movekevin, line 200: Same here. Can you include the struct’s name?

third_party/move/move-binary-format/src/compatibility.rs
```diff
@@ -192,17 +192,17 @@ impl Compatibility {
...
         }
         if self.check_friend_linking && !friend_linking {
             return Err(PartialVMError::new(
                 StatusCode::BACKWARD_INCOMPATIBLE_MODULE_UPDATE,
-            ));
+            ).with_message("Module Update Failure: Friend signature of new module differs from existing module".to_string()));
```
movekevin, line 205: Same

third_party/move/move-vm/runtime/src/interpreter.rs
```diff
@@ -576,7 +576,9 @@ impl Interpreter {
             TypeWithLoader { ty, loader },
             res.is_ok(),
         )?;
-        self.operand_stack.push(res?)?;
+        self.operand_stack.push(res.map_err(|err| {
+            err.with_message(format!("Failed to borrow global resource from {:?}", addr))
```
movekevin, line 580: Can we include the resource name here as well?

third_party/move/move-vm/runtime/src/interpreter.rs
```diff
@@ -658,7 +661,7 @@ impl Interpreter {
                     &resource,
                     false,
                 )?;
-                Err(err)
+                Err(err.with_message(format!("Failed to move resource into {:?}", addr)))
```
movekevin, line 664: Same

movekevin (commented):
Thanks for doing this! This would be a huge improvement! Do we need this verbose flag or should we just enable this by default?

runtian-zhou:
verbose flag is just for testing. We will get this behavior in prod by default.

third_party/move/testing-infra/transactional-test-runner/src/vm_test_harness.rs
```diff
@@ -329,37 +335,73 @@ impl<'a> MoveTestAdapter<'a> for SimpleVMTestAdapter<'a> {
     }
 }
 
-pub fn format_vm_error(e: &VMError, comparison_mode: bool) -> String {
+pub fn format_vm_error(e: &VMError, comparison_mode: bool, verbose: bool) -> String {
```
brmataptos, line 338: Please rebase on main to get the changes from my PR.

third_party/move/move-vm/types/src/values/values_impl.rs
brmataptos: I suspect that Teng means that putting that literal into a const means less work if you decide to change it.  In this case, you could separate the error generation from the shifting, at the cost of an extra match:
```
 match (match self {
            U8(x) => {
                if n_bits >= 8 {
                    Err(1)
                } else {
                    Ok(IntegerValue::U8(x << n_bits))
                },
           ...
}) {
    OK(value) => Ok(value),
    Error(_) => Err(PartialVMError::new(StatusCode::ARITHMETIC_ERROR)
                        .with_message("Shift Left overflow".to_string())),
}
```

brmataptos (changes requested):
I agree with movekevin@.

third_party/move/move-vm/runtime/src/interpreter.rs
```diff
@@ -607,22 +609,23 @@ impl Interpreter {
...
+                    is_generic,
+                    TypeWithLoader { ty, loader },
+                    Some(&resource),
+                )?;
+                resource
+            },
+            Err(err) => {
+                let val: Option<&Value> = None;
+                gas_meter.charge_move_from(is_generic, TypeWithLoader { ty, loader }, val)?;
+                return Err(err.with_message(format!("Failed to move resource from {:?}", addr)));
```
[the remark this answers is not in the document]
runtian-zhou: The resource name is actually hard because the Type doesn't have struct information right now. We need an extra call into the resolver to get this info. However if we land the loader PR we should be able to get this much easier.

runtian-zhou:
Rebased and updated the comment @brmataptos

third_party/move/move-vm/runtime/src/interpreter.rs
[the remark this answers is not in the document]
movekevin: Isn’t there ty there? I’m not looking at this in my phone so cant click through to check the typing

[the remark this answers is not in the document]
brmataptos: @movekevin, enum Type just has an index into a table for structs, so it's not terribly useful to print.  Assuming the bytecode offset is known, though, the user can look at a listing and see the type.

But, @runtian-zhou:  we could add a method to loader to convert a type to a string for this debugging.  You have loader here.

third_party/move/move-vm/types/src/values/values_impl.rs
```diff
@@ -1589,75 +1604,33 @@ impl IntegerValue {
...
-                    return Err(PartialVMError::new(StatusCode::ARITHMETIC_ERROR));
-                }
-                IntegerValue::U128(x << n_bits)
-            },
+            U8(x) if n_bits < 8 => IntegerValue::U8(x << n_bits),
+            U16(x) if n_bits < 16 => IntegerValue::U16(x << n_bits),
+            U32(x) if n_bits < 32 => IntegerValue::U32(x << n_bits),
+            U64(x) if n_bits < 64 => IntegerValue::U64(x << n_bits),
+            U128(x) if n_bits < 128 => IntegerValue::U128(x << n_bits),
             U256(x) => IntegerValue::U256(x << n_bits),
```
brmataptos, line 1612: What happens if I do (1u256 << 300) ?

third_party/move/move-vm/types/src/values/values_impl.rs
```diff
@@ -1589,75 +1604,33 @@ impl IntegerValue {
...
-                }
-                IntegerValue::U64(x << n_bits)
-            },
-            U128(x) => {
-                if n_bits >= 128 {
-                    return Err(PartialVMError::new(StatusCode::ARITHMETIC_ERROR));
-                }
-                IntegerValue::U128(x << n_bits)
-            },
+            U8(x) if n_bits < 8 => IntegerValue::U8(x << n_bits),
```
brmataptos, line 1607: Nice refactor.

runtian-zhou:
> But, @runtian-zhou: we could add a method to loader to convert a type to a string for this debugging. You have loader here.

Well if we changed the type to the representation in loader refactor we could immediately print that out :D

third_party/move/move-vm/types/src/values/values_impl.rs
runtian-zhou: this would just be zero I guess?

brmataptos:
Not an error?

runtian-zhou:
> Not an error?

Ehhh nvm. The RHS of the operation has to be u8 so it can't be greater than 255 so shift is always safe

brmataptos:
Why do we have code reviews if you're just going to dismiss requests once you get enough rubber-stamps?

You should file a followup issue to add resources to the messages in the future when it is easier.

runtian-zhou:
Sry I thought you forgot to click the review button after leaving your comments. Lemme file an issue for that.

PR #10001, opened 2023-09-11
add MOVE_COMPILER_DEBUG and MOVE_TEST_DEBUG env vars to add more info…

brmataptos:
add MOVE_COMPILER_DEBUG and MOVE_TEST_DEBUG env vars to add more info to outputs for debugging

### Description

VMError outputs are deliberately suppressed in unit and transactional tests to make comparison with "golden" expected outputs easier.  But for debugging an unexpectedly failing test, the full VMError output is useful to see.  By setting the environment variable MOVE_TEST_DEBUG=1, the developer can see the full test output.  If RUST_BACKTRACE=1 is also set, then this should include the VM or Loader stack for better debugging.  At Runtian's suggestion, this is also now available by adding --verbose to the publish command in transactional tests, see https://github.com/aptos-labs/aptos-core/pull/10000/.

To debug the Move Compiler, it is most useful to see the compiler IR between passes.  By setting the envronment variable MOVE_COMPILER_DEBUG=1 when running the compiler (or full build), the developer can see that IR after each pass.  Note that this will be too verbose to be useful unless just a small program is being compiled.

### Test Plan
Testing that without these environment variables set, nothing changes.  Manual testing for useful output with environment variables set.

### DIFF
--- a/third_party/move/move-binary-format/src/errors.rs
+++ b/third_party/move/move-binary-format/src/errors.rs
@@ -288,12 +288,10 @@ impl VMError {
     location: {location_string},
     indices: {indices},
     offsets: {offsets},
-    exec_state: {exec_state:?},
 }}",
                 major_status = self.major_status(),
                 sub_status = self.sub_status(),
                 location_string = location_string,
-                exec_state = self.exec_state(),
                 // TODO maybe include source map info?
                 indices = indices,
                 offsets = offsets,
--- a/third_party/move/move-binary-format/src/errors.rs
+++ b/third_party/move/move-binary-format/src/errors.rs
@@ -234,6 +234,53 @@ impl VMError {
             offsets,
         }))
     }
+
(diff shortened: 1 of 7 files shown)


runtian-zhou (approved):
Looks good to me!

brmataptos:
@wrwg @vineethk Please take a look, this now has exactly what I need to debug some transactional tests:
(1) add --print-bytecode to the //#publish task in the test file
(2) run test with MOVE_COMPILER_DEBUG=1 MOVE_TEST_DEBUG=1 UB=1 cargo test .. testname >& test.out
Then: 
(1) .exp file has detailed Error, Core Dump, and bytecode.
(2) test.out has printed IR from all the compiler passes.

Non-compiler developers will prefer to use the --debug flag on the transactional commands to get more detailed VMError messages.

vineethk (approved):
lgtm.

PR #10002, opened 2023-09-11
re-enable sharded executor unit tests

zjma:
Fixed a unit test setup bug which is probably responsible for the previous flakiness...

### DIFF
--- a/aptos-move/aptos-vm/src/sharded_block_executor/test_utils.rs
+++ b/aptos-move/aptos-vm/src/sharded_block_executor/test_utils.rs
@@ -188,7 +188,7 @@ pub fn sharded_block_executor_with_random_transfers<E: ExecutorClient<FakeDataSt
     let mut rng = OsRng;
     let max_accounts = 200;
     let max_txns = 1000;
-    let num_accounts = rng.gen_range(1, max_accounts);
+    let num_accounts = rng.gen_range(2, max_accounts);
     let mut accounts = Vec::new();
     let mut executor = FakeExecutor::from_head_genesis();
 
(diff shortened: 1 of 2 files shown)


aptos-move/aptos-vm/src/sharded_block_executor/test_utils.rs
```diff
@@ -188,7 +188,7 @@ pub fn sharded_block_executor_with_random_transfers<E: ExecutorClient<FakeDataSt
     let mut rng = OsRng;
     let max_accounts = 200;
     let max_txns = 1000;
-    let num_accounts = rng.gen_range(1, max_accounts);
+    let num_accounts = rng.gen_range(2, max_accounts);
```
sitalkedia, line 191: Why would the tests fail when we have 1 account? Is this something we need to handle in the partitioner?

zjma: because somewhere in the txn generator we requires the transfer to be between 2 different accouts...

msmouse (approved):
in4b blind stamp is asked for

aptos-move/aptos-vm/src/sharded_block_executor/test_utils.rs
zjma: i.e., [here](https://github.com/aptos-labs/aptos-core/blob/0c0aa7fe09954d9d21652e49f68a4cb05b56f09c/aptos-move/aptos-vm/src/sharded_block_executor/test_utils.rs#L206)

sitalkedia (approved):
Thanks for the fix...

PR #10008, opened 2023-09-11
[NFT Metadata Crawler] Trim URIs before making request

just-in-chang:
### Description
Some of the token and collection URIs have a new line character at the end. This change trims the provided URIs before making the request.


GhostWalker562 (approved):
Downstream 🌊

larry-aptos (approved):

PR #10010, opened 2023-09-12
feat: add voting::get_proposal_metadata

0xbe1:


0xbe1:
prefer https://github.com/aptos-labs/aptos-core/pull/10010

PR #10012, opened 2023-09-12
[draft] my forge run

msmouse:


testsuite/forge-cli/src/main.rs
```diff
@@ -1647,12 +1647,26 @@ fn realistic_network_tuned_for_throughput_test() -> ForgeConfig {
...
+            helm_values["validator"]["resources"]["requests"]["cpu"] = 48.into();
+            helm_values["validator"]["resources"]["limits"]["memory"] = "72Gi".into();
+            helm_values["validator"]["resources"]["requests"]["memory"] = "72Gi".into();
+            helm_values["fullnode"]["resources"]["limits"]["cpu"] = 48.into();
+            helm_values["fullnode"]["resources"]["requests"]["cpu"] = 48.into();
+            helm_values["fullnode"]["resources"]["limits"]["memory"] = "72Gi".into();
+            helm_values["fullnode"]["resources"]["requests"]["memory"] = "72Gi".into();
+
+            // higher concurrency level
+            helm_values["validator"]["config"]["execution"]["concurrency_level"] = 32.into();
```
JoshLind, line 1667: Note: state sync is in execution mode, so it probably makes sense to set this for the fullnodes, too?

PR #10017, opened 2023-09-12
Add create_unbounded_aggregator to simplify callers

igor-aptos:


georgemitenkov (approved):

aptos-move/framework/aptos-framework/sources/aggregator_v2/aggregator_v2.move
```diff
@@ -42,6 +46,21 @@ module aptos_framework::aggregator_v2 {
...
     public native fun create_aggregator<Element: copy + drop>(max_value: Element): Aggregator<Element>;
 
+    /// Creates new aggregator, without any 'max_value' on top of the implicit bound restriction
+    /// due to the width of the type (i.e. MAX_U64 for u64, MAX_U128 for u128).
+    ///
+    /// Currently supported types for Element are u64 and u128.
+    /// EAGGREGATOR_ELEMENT_TYPE_NOT_SUPPORTED raised if called with a different type.
+    public fun create_unbounded_aggregator<Element: copy + drop>(): Aggregator<Element> {
+        if (type_info::type_of<Element>() == type_info::type_of<u64>()) {
+            create_aggregator(MAX_U64)
```
igor-aptos, line 56: oops, this doesn't compile, as it doesn't match the type.

I don't know if I can write this correctly in move?

maybe it either needs to be native function, or to have two ? (create_unbounded_aggregator_u64 and create_unbounded_aggregator_u128)

georgemitenkov:
@igor-aptos I though you would try to compile before sending the PR haha

aptos-move/framework/aptos-framework/sources/aggregator_v2/aggregator_v2.move
georgemitenkov: This should return true
```
type_name<u64>() == string::utf8(b"u64")
```

igor-aptos: this compiles:
`        if (type_info::type_of<Element>() == type_info::type_of<u64>()) {`


this doesn't:
            return create_aggregator(MAX_U64)
when return type of the function is : Aggregator<Element>

Compiler could probably infer something, but even in C++/rust not sure that would pass. Not sure if there is a different way to write this.

Issue #7306, opened 2023-03-21
[Bug] The Aptos move compile bytecode is missing metadata data


molefund:
# 🐛 Bug

Problem description:

When executing the command aptos move compile using aptos cli, if you use options -- output dir<OUTPUT_ DIR>, it will cause the compiled. mv file to lack metadata information, making # [view] invalid.

Cli version: 1.0.8 and earlier.

Cause of the problem:

When executing the aptos move compile, the specified OUTPUT_ DIR generates. mv files, but when injecting metadata, only the default build file is detected and updated.

Related code:
rev : "d76270cf5e025a4c9533faf88619b2565fde4f7f"    built_ Inject in the package.rs file_ runtime_ Metadata method

```

fn inject_runtime_metadata(
    package_path: PathBuf,
    pack: &mut CompiledPackage,
    metadata: BTreeMap<ModuleId, RuntimeModuleMetadataV1>,
    bytecode_version: Option<u32>,
) -> anyhow::Result<()> {
    for unit_with_source in pack.root_compiled_units.iter_mut() {
        match &mut unit_with_source.unit {
            CompiledUnit::Module(named_module) => {
                if let Some(module_metadata) = metadata.get(&named_module.module.self_id()) {
                    if !module_metadata.is_empty() {
                        if bytecode_version.unwrap_or(METADATA_V1_MIN_FILE_FORMAT_VERSION)
                            >= METADATA_V1_MIN_FILE_FORMAT_VERSION
                        {
                            let serialized_metadata = bcs::to_bytes(&module_metadata)
                                .expect("BCS for RuntimeModuleMetadata");
                            named_module.module.metadata.push(Metadata {
                                key: APTOS_METADATA_KEY_V1.clone(),
                                value: serialized_metadata,
                            });
                        } else {
                            let serialized_metadata =
                                bcs::to_bytes(&module_metadata.clone().downgrade())
                                    .expect("BCS for RuntimeModuleMetadata");
                            named_module.module.metadata.push(Metadata {
                                key: APTOS_METADATA_KEY.clone(),
                                value: serialized_metadata,
                            });
                        }

                        // FIXME: package_path is move package path. There are no .mv files in the folder when OUTPUT_PATH is different from package_path
                        // Also need to update the .mv file on disk.
                        let path = package_path
                            .join(CompiledPackageLayout::CompiledModules.path())
                            .join(named_module.name.as_str())
                            .with_extension(MOVE_COMPILED_EXTENSION);
[...]

PR #10018, opened 2023-09-13
Fixing issue #7306

vineethk:
### Description

See issue #7306 for more details on the problem being fixed.

When using `aptos move compile --output-dir`, the function `inject_runtime_metadata` was erroneously using the default path for the `.mv` file, instead of the path in the specified directory. We now use the specified path if one is provided, or else use the default.

### Test Plan

I manually tested that we pick the right directory. It is not clear what sort of automated test makes sense in this scenario.


brmataptos (approved):
LGTM

PR #10019, opened 2023-09-13
accomodate that LedgerCommitProgress was not recorded after all ledge…

msmouse:
…r commits

### Description

On start up truncate can fail because of that. 
The PR tries to read the progress out of each individual data set.

### Test Plan

existing tests and verified the new build fixed the crash loop on one of the testnet pfns.


zekun000 (approved):

danielxiangzl (approved):
stamp

PR #10031, opened 2023-09-13
[Aptos-release-builder] Add printing package metadata to the aptos-release-builder

rahxephon89:
### Description

This PR adds a new subcommand `print-package-metadata` to the aptos-release-builder to print out the package metadata given the package name and the account address.

Usage: aptos-release-builder print-package-metadata --endpoint '\<URL\>' --package-address \<ADDRESS\> --package-name <PACKAGE_NAME> [--print-json]

For instance, when executing the command `aptos-release-builder print-package-metadata --endpoint 'https://mainnet.aptoslabs.com/v1' --package-address 0x1 --package-name AptosFramework`, the output looks like this:

```PackageMetadata:
Package name:AptosFramework
Upgrade policy:compatible
Upgrade number:8
Source digest:DB1749CC57545C1CF4F3099837619D27A1397E7802EF926DE6E1576BC0CC7112
Manifest:
[package]
name = "AptosFramework"
version = "1.0.0"

[addresses]
std = "0x1"
aptos_std = "0x1"
aptos_framework = "0x1"
aptos_token = "0x3"
core_resources = "0xA550C18"
vm_reserved = "0x0"

[dependencies]
AptosStdlib = { local = "../aptos-stdlib" }
MoveStdlib = { local = "../move-stdlib" }

Package Dependency:
PackageDep { account: 0000000000000000000000000000000000000000000000000000000000000001, package_name: "AptosStdlib" }
PackageDep { account: 0000000000000000000000000000000000000000000000000000000000000001, package_name: "MoveStdlib" }
extension:MoveOption { value: [] }
Modules:
Module name:system_addresses
Source code:
...
```
Note that the metadata of each module will also be printed out and they are omitted here.

### DIFF
--- a/aptos-move/aptos-release-builder/src/main.rs
+++ b/aptos-move/aptos-release-builder/src/main.rs
@@ -3,6 +3,7 @@
 
 use anyhow::Context;
 use aptos_crypto::{ed25519::Ed25519PrivateKey, ValidCryptoMaterialStringExt};
+use aptos_framework::natives::code::PackageRegistry;
 use aptos_release_builder::{
     components::fetch_config,
     initialize_aptos_core_path,
@@ -61,6 +62,18 @@ pub enum Commands {
         #[clap(short, long)]
         print_gas_schedule: bool,
     },
+    /// Print out package metadata.
+    PrintPackageMetadata {
+        /// Url endpoint for the desired network. e.g: https://fullnode.mainnet.aptoslabs.com/v1.
+        #[clap(short, long)]
+        endpoint: url::Url,
+        /// The address under which the package is published
+        #[clap(long)]
+        package_address: String,
+        /// The name of the package
+        #[clap(long)]
+        package_name: String,
+    },
 }
 
 #[derive(Subcommand, Debug)]
@@ -84,7 +97,7 @@ pub enum InputOptions {
 }
 
 #[tokio::main]
-async fn main() {
+async fn main() -> anyhow::Result<()> {
     let args = Argument::parse();
     initialize_aptos_core_path(args.aptos_core_path.clone());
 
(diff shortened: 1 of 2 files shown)


runtian-zhou:
This looks great! Can you add an example of the output for aptos framework on mainnet?

rahxephon89:
> This looks great! Can you add an example of the output for aptos framework on mainnet?

Added an example output in the description.

runtian-zhou:
Looks great! this matches the source digest in: https://github.com/aptos-foundation/mainnet-proposals/blob/main/sources/v1.6/step_1_upgrade_framework/3-aptos-framework.move

aptos-move/aptos-release-builder/src/main.rs
```diff
@@ -61,6 +62,18 @@ pub enum Commands {
         #[clap(short, long)]
         print_gas_schedule: bool,
     },
+    /// Print out package metadata.
```
perryjrandall, line 65: Can you add some usage example like from your test plan?

aptos-move/aptos-release-builder/src/main.rs
```diff
@@ -210,14 +216,31 @@ async fn main() {
...
+            package_name,
+        } => {
+            let client = aptos_rest_client::Client::new(endpoint);
+            let address = AccountAddress::from_str_strict(&package_address)?;
+            let packages = client
+                .get_account_resource_bcs::<PackageRegistry>(address, "0x1::code::PackageRegistry")
+                .await?;
+            for package in packages.into_inner().packages {
+                if package.name == package_name {
+                    println!("PackageMetadata:\n{}", package);
```
perryjrandall, line 240: Can we add an option to produce json instead of pretty printing (looks like package metadata already supports serialize / deserialize)

perryjrandall (approved):
LFG amazing

aptos-move/aptos-release-builder/src/main.rs
rahxephon89: done

aptos-move/aptos-release-builder/src/main.rs
rahxephon89: done

davidiw (approved):
overall great!

why did we move from unwrap to result for the main function?

rahxephon89:
> overall great!
> 
> why did we move from unwrap to result for the main function?

Yeah, I just wanted to avoid panic if the input is not good and provide better error message but it is probably fine to keep unwrap there.

PR #10035, opened 2023-09-13
[dep] moving testing features out from dependencies

zekun000:
Including features directly in dependencies will expose it unconditionally which is not what we want most of the time.

### DIFF
--- a/types/src/state_store/state_value.rs
+++ b/types/src/state_store/state_value.rs
@@ -156,7 +156,7 @@ impl StateValue {
     }
 }
 
-#[cfg(any(test, feature = "fuzzing"))]
+// #[cfg(any(test, feature = "fuzzing"))]
 impl From<Vec<u8>> for StateValue {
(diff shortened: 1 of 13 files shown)


types/src/state_store/state_value.rs
```diff
@@ -156,7 +156,7 @@ impl StateValue {
     }
 }
 
-#[cfg(any(test, feature = "fuzzing"))]
+// #[cfg(any(test, feature = "fuzzing"))]
 impl From<Vec<u8>> for StateValue {
```
msmouse, line 160: remove

msmouse (approved):

gelash (approved):
So the proper way is to fix in dev dependencies, makes perfect sense, thanks and good to remember.

Do you mind replacing those two lines in the delta-set, we could fix it in this PR?
Could you https://github.com/aptos-labs/aptos-core/pull/10029

ibalajiarun (approved):

PR #10043, opened 2023-09-14
[genesis] support account balances file

lebdron:
### Description
- Specify `CliError` for `aptos_github_client::Error::NotFound` to distinguish generic error from a missing file.
- Allow passing a balances file for non-mainnet deployments
- Use `create_account` instead of `create_accounts` to avoid quadratic complexity https://github.com/aptos-labs/aptos-core/blob/a52efba0f43342f654ed5ed861da2570ea77f218/aptos-move/framework/aptos-framework/sources/genesis.move#L166-L172

### Test Plan
Run `aptos genesis` command with `balances.yaml` present.


lebdron:
Bump.

lebdron:
Bump.

PR #10047, opened 2023-09-14
refactor: replace multiply_then_divide using math64::mul_div

0xbe1:

### DIFF
--- a/aptos-move/framework/aptos-framework/sources/delegation_pool.move
+++ b/aptos-move/framework/aptos-framework/sources/delegation_pool.move
@@ -113,6 +113,7 @@ module aptos_framework::delegation_pool {
     use std::signer;
     use std::vector;
 
+    use aptos_std::math64;
     use aptos_std::pool_u64_unbound::{Self as pool_u64, total_coins};
     use aptos_std::table::{Self, Table};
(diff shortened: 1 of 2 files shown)


aptos-move/framework/aptos-framework/sources/delegation_pool.move
```diff
@@ -1590,13 +1591,9 @@ module aptos_framework::delegation_pool {
         current_delegated_votes.pending_inactive_shares = current_delegated_votes.pending_inactive_shares - shares_to_redeem;
     }
 
+    #[deprecated]
     public fun multiply_then_divide(x: u64, y: u64, z: u64): u64 {
-        let result = (to_u128(x) * to_u128(y)) / to_u128(z);
```
movekevin, line 1594: Let's may be still keep the same code here in case someone is using this function on chain for whatever reasons

0xbe1: lemme call `math64::mul_div` instead and add comments

davidiw (approved):

davidiw:
you may need to update the docs

0xbe1:
@davidiw updated docs

movekevin (approved):

PR #10048, opened 2023-09-14
[move] Add error messages for linker error #9999

aalok-t:
### Description

Add error messages for `LINKER_ERROR`, fixing part of https://github.com/aptos-labs/aptos-core/issues/9580


vineethk:
This PR seems to be something other than the description.

PR #10051, opened 2023-09-14
[Forge] Add a netbench test

bchocho:
### Description

Create netbench tests across a pair of nodes, one colocated and the other emulated on a high-latency (and low bandwidth) network.

For now there is no success criteria, but we should be able to add one based on throughput when needed.

### Test Plan

Run the tests in forge.

### DIFF
--- a/testsuite/forge-cli/src/main.rs
+++ b/testsuite/forge-cli/src/main.rs
@@ -6,8 +6,8 @@
 
 use anyhow::{format_err, Context, Result};
 use aptos_config::config::{
-    BootstrappingMode, ConsensusConfig, ContinuousSyncingMode, MempoolConfig, NodeConfig,
-    StateSyncConfig,
+    BootstrappingMode, ConsensusConfig, ContinuousSyncingMode, MempoolConfig, NetbenchConfig,
+    NodeConfig, StateSyncConfig,
 };
 use aptos_forge::{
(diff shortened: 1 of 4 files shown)


testsuite/testcases/src/multi_region_network_test.rs
```diff
@@ -8,20 +8,13 @@ use aptos_types::PeerId;
 use itertools::{self, EitherOrBoth, Itertools};
 use std::collections::BTreeMap;
 
-/// The link stats are obtained from https://github.com/doitintl/intercloud-throughput/blob/master/results_202202/results.csv
```
ibalajiarun, line 11: can we retain this link?

ibalajiarun (approved):

testsuite/forge-cli/src/main.rs
```diff
@@ -1346,6 +1359,47 @@ fn gather_metrics() -> ForgeConfig {
         .add_network_test(GatherMetrics)
 }
 
+fn netbench_config_100_megabytes_per_s() -> NetbenchConfig {
```
JoshLind, line 1362: nit: `*_per_s` reads odd to me. Why not just complete it? `*_per_sec`? Could just be me though 😵‍💫

JoshLind (approved):
LGTM!

PR #10053, opened 2023-09-14
[Node Configs] Update sanitizer to be read-only and migrate indexer ops.

JoshLind:
### Description
This PR:
1. Updates the `ConfigSanitizer` to take in an immutable `NodeConfig` reference. This should help to make it obvious that the sanitizer is responsible only for sanitizing/checking configs and not modifying them. Modifications should happen in the `ConfigOptimizer` step.
2. Migrates any existing `sanitize()` methods to `optimize()` if they need to modify the configs. To avoid breaking anything, I've kept the behaviour the same and added a comment about the requirements for the optimizers to not change values if specified by the node config file. This can be dealt with in a follow-up PR.

### Test Plan
Existing test infrastructure.

### DIFF
--- a/config/src/config/api_config.rs
+++ b/config/src/config/api_config.rs
@@ -135,7 +135,7 @@ impl ApiConfig {
 
 impl ConfigSanitizer for ApiConfig {
     fn sanitize(
-        node_config: &mut NodeConfig,
+        node_config: &NodeConfig,
         node_type: NodeType,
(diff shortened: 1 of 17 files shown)


banool (approved):
Indexer aspect of this looks good. I agree the old indexer config approach right now is sorta jank but yeah let's leave it for now.

grao1991 (approved):

PR #10064, opened 2023-09-15
Fix Issue 9717 by breaking critical edges in CFG in move-compiler (v1)

brmataptos:
### Description

Fix issue https://github.com/aptos-labs/aptos-core/issues/9717 by breaking critical edges in the CFG in move-compiler (v1).

While issue 9717 mentions "inline", this is a red herring as inlining a vector loop just makes it more likely to have a reference parameter on the stack, which leads to a failure to correctly drop dead references in move-compiler/src/cfgir/liveness/mod.rs.  This happens when the constructed CFG has critical edges (edges from a node with multiple outgoing edges to a node with multiple ingoing edges).  Such a situation is well-known in compiler literature to make it difficult to place instructions precisely based on certain analyses.

Fortunately, this seems to be easily fixed by adding a small pass to add a node on each such edge, so that the drop can be placed appropriately even in the case of a break.

Currently there are still some empty blocks left in the bytecode.  Will clean that up soon.

### Test Plan
Running all move tests should be fairly thorough.


third_party/move/move-compiler/transactional-tests/tests/inlining/bug_9717_looponly.move
```diff
@@ -0,0 +1,138 @@
...
+        let i = 0;
+        let flipsref1 = &flips;
+        while (i < vector::length(flipsref1)) {
+            if (*vector::borrow(flipsref1, i) != 0) {
+                break
+            };
+            i = i + 1;
+        };
+        let _v = copy flips; // this is ok
+        // this will fail, a UNKNOWN_INVARIANT_VIOLATION_ERROR (code 2000)
```
vineethk, line 16:
```suggestion
        // this used to fail with an UNKNOWN_INVARIANT_VIOLATION_ERROR (code 2000)
```

third_party/move/move-compiler/transactional-tests/tests/inlining/bug_9717.move
```diff
@@ -0,0 +1,99 @@
+//#publish --print-bytecode
+module 0xcafe::vectors {
+    use std::vector;
+    public entry fun guess_flips(flips: vector<u8>) {
+        loops_with_break(&flips);
+        let _v = copy flips; // this is ok
+
+        // this will fail, a UNKNOWN_INVARIANT_VIOLATION_ERROR (code 2000)
```
vineethk, line 8:
```suggestion
        // this used to fail with an UNKNOWN_INVARIANT_VIOLATION_ERROR (code 2000)
```

third_party/move/move-compiler/src/cfgir/cfg.rs
```diff
@@ -192,6 +192,72 @@ impl<'a> BlockCFG<'a> {
...
+            },
+        }
+    }
+
+    // Helper inserts an edge, but doesn't update maps.  Caller must should ensure that recompute()
+    // is called at some point.  new_label must be a new unique Label.
+    fn add_edge(&mut self, pred: Label, succ: Label, new_label: Label) {
+        let loc = {
+            let from_block = self.block_mut(pred);
+            let branch_cmd = from_block.back_mut().unwrap();
```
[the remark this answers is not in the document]
brmataptos: Not sure that's any better than the panic that will happen in remap_jump if the command and label don't have a match.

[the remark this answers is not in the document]
vineethk: I guess this is a bit subjective: my reasoning it that is a pre-condition of the call, so perhaps better to check before the call. However, I recognize that it is redundant due to the panic, so I added "consider"; feel free to ignore this comment if it does not resonate with you.

third_party/move/move-compiler/src/cfgir/cfg.rs
```diff
@@ -192,6 +192,72 @@ impl<'a> BlockCFG<'a> {
...
     }
+
+    /// Adds a block on any critical edge: an edge from a node with multiple successors to
+    /// a node with multiple predecessors.  Returns true if any changes occur; in this
+    /// case, caller must call recompute().
+    pub fn remove_critical_edges<Ftype>(&mut self, mut make_label: Ftype) -> bool
+    where
+        Ftype: FnMut() -> Label,
+    {
+        let critical_edges: Vec<_> = self
```
wrwg, line 203: I don't think you need to construct the vector, you just can use the iterator in the for loop below. (Clippy should(?) tell you if I'm right -- did you run `rust_lint`?)

third_party/move/move-compiler/src/cfgir/cfg.rs
```diff
@@ -192,6 +192,72 @@ impl<'a> BlockCFG<'a> {
...
+                            .map_or_else(|| false, |preds| (preds.len() > 1))
+                    })
+                    .map(|lbl2| (*lbl, *lbl2))
+            })
+            .collect();
+        for (pred, succ) in &critical_edges {
+            let new_label = make_label();
+            self.add_edge(*pred, *succ, new_label);
+        }
+        !critical_edges.is_empty()
```
wrwg, line 222: If you don't use vector, can use a bool in the loop to see whether its empty.

third_party/move/move-compiler/src/cfgir/cfg.rs
```diff
@@ -192,6 +192,72 @@ impl<'a> BlockCFG<'a> {
             println!();
         }
     }
+
+    /// Adds a block on any critical edge: an edge from a node with multiple successors to
+    /// a node with multiple predecessors.  Returns true if any changes occur; in this
+    /// case, caller must call recompute().
```
wrwg, line 198: Perhaps clarify recompute means something with the overall graph. I first thought you mean recompute for splitting edges again, which seems to be not the case.

third_party/move/move-compiler/src/cfgir/translate.rs
```diff
@@ -423,7 +432,33 @@ fn function_body(
 
             let (mut cfg, infinite_loop_starts, diags) =
                 BlockCFG::new(start, &mut blocks, &block_info);
+
             context.env.add_diags(diags);
+            if context.env.flags().debug() {
```
wrwg, line 437: That is unusual. The idiomatic way would be to have this compile time protected, as in

```
#[cfg(debug_assertions)]
eprintln!(...
```

But understood that seems to be a result of the env vars you have introduced before for printing error details. 

We should discuss whether we should really use this like this. (Not as part of this PR.) There are crates for logging which might be better suited than an explicit println.

wrwg (commented):
Nice! I actually did not know about the critical edge problem, might be useful to cite something like this: https://nickdesaulniers.github.io/blog/2023/01/27/critical-edge-splitting/

Otherwise only nits.

third_party/move/move-compiler/src/cfgir/cfg.rs
brmataptos: There's a borrow problem with that.  To get the items I need &self, but to split edges I need a &mut self.  Easier to leave it separate.  I've changed slightly it to consume critical_edges in the loop.

third_party/move/move-compiler/src/cfgir/cfg.rs
brmataptos: Added a bool, but see above.

third_party/move/move-compiler/src/cfgir/translate.rs
brmataptos: I'll remove the extra printing before submitting,

brmataptos:
Ok, I've addressed comments and pushed some changes to address problems revealed by some tests.

(1) changes to address some bugs revealed in cfgir/optimize/inline_blocks.rs.  This pass does jump folding if successor block has no other predecessors, but it was broken for certain sequences of blocks (for B->A->C, if label B was visited before A, then things didn't work).

(2) stack overflow in hlir/translate.rs on test deep_exp.move.  I've made some changes to hlir/translate.rs to use an existing mechanism in a few more cases to reduce recursion depth, but it's still not quite enough.  I'll give it another shot and/or figure out how to add a useful warning for excessive stack depth.

wrwg:
> Ok, I've addressed comments and pushed some changes to address problems revealed by some tests.
> 
> (1) changes to address some bugs revealed in cfgir/optimize/inline_blocks.rs. This pass does jump folding if successor block has no other predecessors, but it was broken for certain sequences of blocks (for B->A->C, if label B was visited before A, then things didn't work).
> 
> (2) stack overflow in hlir/translate.rs on test deep_exp.move. I've made some changes to hlir/translate.rs to use an existing mechanism in a few more cases to reduce recursion depth, but it's still not quite enough. I'll give it another shot and/or figure out how to add a useful warning for excessive stack depth.

Let us know when you are ready.

third_party/move/move-compiler/src/cfgir/translate.rs
brmataptos: This kind of thing is really not for typical use, just once the target source code is identified and you want to see everything the compiler is doing with it.  That's why it's an awkward environment variable, MOVE_COMPILER_DEBUG.

I changed my mind, probably won't remove this one, as seeing the state between compiler stages can be very useful if we have more debugging to do there.  I wonder if I can get the compiled function name easily enough to have this be a string to match instead, though.

wrwg: SG, as said this is not for this PR. We should have a general logging model using the [log crate similar as in the prover](https://github.com/aptos-labs/aptos-core/blob/55398e79f7f231dccf6f25cf01a648315193ede4/third_party/move/move-prover/boogie-backend/src/boogie_wrapper.rs#L118C59-L118C59) for this kind of thing, at least in v2. Not caring that much what we do in v1.

wrwg (approved):
Approving module further intended changes.

brmataptos:
> Let us know when you are ready.

Seems to work now.  I played around with avoiding unnecessary let constructs in the case of an actual param which is copy(var), but that didn't work out (warnings about unused variables were too hard to fix up).  Fortunately, the existing fixes seem to lower the stack usage enough to get things to pass.

Just pushed a new version, including a few more tests for unused variables in presence of inlining.

third_party/move/move-compiler/src/cfgir/optimize/inline_blocks.rs
```diff
@@ -88,15 +93,35 @@ fn inline_single_target_blocks(
             // be the target of at least 2 jumps: the jump to the loop and the "continue" jump
             // This is always true as long as we start the count for the start label at 1
             sp!(_, Command_::Jump { target, .. }) if single_jump_targets.contains(target) => {
+                // Note that only the last merged block will be left for cur.
                 remapping.insert(cur, *target);
-                let target_block = working_blocks.remove(target).unwrap();
-                block.pop_back();
-                block.extend(target_block);
+                match working_blocks.remove(target) {
```
vineethk, line 98: Consider rewriting (for reduced code duplication) to:
```
                let target_block = working_blocks.remove(target).unwrap_or_else(|| {
                    finished_blocks.remove(target).unwrap_or_else(|| {
                        panic!(
                            "ICE: Target {} not found in working_blocks or finished_blocks",
                            target
                        )
                    })
                });
                block.pop_back();
                block.extend(target_block);
```

third_party/move/move-compiler/src/cfgir/cfg.rs
brmataptos: clarified.

third_party/move/move-compiler/src/cfgir/translate.rs
brmataptos: ok.

third_party/move/move-compiler/src/cfgir/cfg.rs
```diff
@@ -192,6 +192,72 @@ impl<'a> BlockCFG<'a> {
...
+            .successor_map
+            .iter()
+            .filter(|(_, nexts)| nexts.len() > 1)
+            .flat_map(|(lbl, nexts)| {
+                nexts
+                    .iter()
+                    .filter(|lbl2| {
+                        self.predecessor_map
+                            .get(lbl2)
+                            .map_or_else(|| false, |preds| (preds.len() > 1))
```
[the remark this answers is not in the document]
brmataptos: Actually, clippy seems ok with it.  I seem to recall it being misparsed without it, for some reason.

third_party/move/move-compiler/src/cfgir/optimize/inline_blocks.rs
brmataptos: done.

third_party/move/move-compiler/src/cfgir/optimize/inline_blocks.rs
```diff
@@ -32,12 +32,13 @@ fn optimize_(start: Label, blocks: &mut BasicBlocks) -> bool {
...
 
+// Return a list of labels that have just a single branch to them.
+
 fn find_single_target_labels(start: Label, blocks: &BasicBlocks) -> BTreeSet<Label> {
     use Command_ as C;
     let mut counts = BTreeMap::new();
-    // 'start' block starts as one as it is the entry point of the function. In some sense,
-    // there is an implicit "jump" to this label to begin executing the function
-    counts.insert(start, 1);
+    // 'start' block doesn't count as single entry for these purposes.  Give it 2.
```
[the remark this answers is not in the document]
brmataptos: Yeah, I found 1 wasn't enough before but I think that was due to other problems with the code.  Changed back to 1.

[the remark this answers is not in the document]
brmataptos: Done.

third_party/move/move-compiler/src/cfgir/optimize/inline_blocks.rs
```diff
@@ -67,14 +68,18 @@ fn inline_single_target_blocks(
 ) -> bool {
     //cleanup of needless_collect would result in mut and non mut borrows, and compilation warning.
     let labels_vec = blocks.keys().cloned().collect::<Vec<_>>();
-    let mut labels = labels_vec.into_iter();
-    let mut next = labels.next();
 
+    // all blocks
```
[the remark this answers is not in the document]
brmataptos: I fleshed it out to a longer comment, which may still confuse you.

third_party/move/move-compiler/src/cfgir/optimize/inline_blocks.rs
```diff
@@ -32,12 +32,13 @@ fn optimize_(start: Label, blocks: &mut BasicBlocks) -> bool {
     inline_single_target_blocks(&single_target_labels, start, blocks)
 }
 
+// Return a list of labels that have just a single branch to them.
+
```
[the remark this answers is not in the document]
brmataptos: ok

third_party/move/move-compiler/src/cfgir/optimize/inline_blocks.rs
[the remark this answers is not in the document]
vineethk: No longer confused (the confusion was less about the code and more about why that comment existed) :)

third_party/move/move-compiler/src/cfgir/optimize/inline_blocks.rs
```diff
@@ -131,9 +146,18 @@ fn remap_to_last_target(
...
-        remapping.entry(*label).or_insert(*label);
+        let mut label2 = *label;
+        while let Some(target) = remapping.get(&label2) {
+            // proceed until label2 has no mapping, or is self-mapping.
+            if label2 == *target {
+                break;
+            }
+            label2 = *target;
+        }
+        remapping.entry(*label).or_insert(label2);
```
vineethk, line 160: I don't think this closes transitive chains, as suggested in the comments.

Consider the provided example: lab1 -> lab2 -> lab3, when we reach this line for label == lab1, remapping already has an entry for lab1, so the or_insert wont happen, leaving the mapping as is.

third_party/move/move-compiler/src/cfgir/optimize/inline_blocks.rs
```diff
@@ -131,9 +146,18 @@ fn remap_to_last_target(
     if remapping.is_empty() {
         return;
     }
-    // populate remapping for non changed blocks
+    // populate remapping for non changed blocks, and also
```
vineethk, line 149: Adding self mapping to every label that is not actually remapped seem wasteful. It looks like this may have been done so that "all labels are present in the remapping". Instead, `unwrap_or_` with the default may have sufficed (that is, when looking up a label in the remapping, absence indicates "use key as the value", and we get the same semantics), likely reducing the size of the mapping.

This is not a behavior introduced by this PR, so I don't expect this change to be made in this PR. I just wanted to bring this to attention, and see if what I say makes sense.

vineethk (changes requested):

third_party/move/move-compiler/src/cfgir/optimize/inline_blocks.rs
brmataptos: I think this was done to simplify `remap_labels()` (defined in cfgir/ast.rs).  I'll try to rewrite that without making it too ugly.

third_party/move/move-compiler/src/cfgir/optimize/inline_blocks.rs
brmataptos: Thanks, good eye.   I've rewritten it more carefully, and use a missing entry to represent identity.

brmataptos (commented):
Thanks for comments.  Please take another look, @vineethk .

vineethk (approved):
lgtm!

PR #10067, opened 2023-09-15
Fix API spec reference for fee payer address and signer

ari-anchor:
### Description
The bad reference in the API spec was breaking the go code generation. This way it no longer breaks.

### Test Plan

Ran go code generation


gregnazario (approved):
good catch, @banool can you followup on checking how this got in, I don't remember what we did to prevent the `allOf` in the spec generation

banool:
Assuming the test passes that confirms the spec matches the generated spec, we can land this.

banool (approved):

PR #10070, opened 2023-09-15
[e2e-move-tests][harness] Cache module building across tests

igor-aptos:
We have multiple tests using the same move code in data folders, causing them to be rebuilt over and over. Adding to harness methods that cache building, so it can be reused.

Haven't changed the default methods to do the caching, as caller can decide to modify the files, and then rebuilding is needed - so it is up to the caller to decide if it wants caching or not.

This speeds up e2e-move-tests on my machine from 80s to 50s:
RUST_BACKTRACE=1 RUST_MIN_STACK=104857600 cargo test --profile ci --package e2e-move-tests --lib

Before:
test result: ok. 113 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 79.58s

After:
test result: ok. 113 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 54.51s

### DIFF
--- a/aptos-move/e2e-move-tests/src/harness.rs
+++ b/aptos-move/e2e-move-tests/src/harness.rs
@@ -37,15 +37,23 @@ use move_core_types::{
     value::MoveValue,
 };
 use move_package::package_hooks::register_package_hooks;
+use once_cell::sync::Lazy;
 use project_root::get_project_root;
 use rand::{
     rngs::{OsRng, StdRng},
     Rng, SeedableRng,
 };
 use serde::{de::DeserializeOwned, Serialize};
-use std::{collections::BTreeMap, path::Path};
+use std::{
+    collections::{BTreeMap, HashMap},
+    path::{Path, PathBuf},
+    sync::{Arc, Mutex},
+};
 
 const DEFAULT_GAS_UNIT_PRICE: u64 = 100;
+
+static CACHED_BUILT_PACKAGES: Lazy<Mutex<HashMap<PathBuf, Arc<anyhow::Result<BuiltPackage>>>>> =
+    Lazy::new(|| Mutex::new(HashMap::new()));
 
 /// A simple test harness for defining Move e2e tests.
 ///
@@ -330,19 +338,16 @@ impl MoveHarness {
         output.gas_used()
     }
 
-    /// Creates a transaction which publishes the Move Package found at the given path on behalf
+    /// Creates a transaction which publishes the passed already-built Move Package on behalf
     /// of the given account.
     ///
     /// The passed function allows to manipulate the generated metadata for testing purposes.
-    pub fn create_publish_package(
+    pub fn create_publish_built_package(
         &mut self,
         account: &Account,
-        path: &Path,
-        options: Option<BuildOptions>,
+        package: &BuiltPackage,
         mut patch_metadata: impl FnMut(&mut PackageMetadata),
     ) -> SignedTransaction {
-        let package = BuiltPackage::build(path.to_owned(), options.unwrap_or_default())
-            .expect("building package must succeed");
         let code = package.extract_code();
         let mut metadata = package
             .extract_metadata()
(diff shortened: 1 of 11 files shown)


aptos-move/e2e-move-tests/src/harness.rs
```diff
@@ -334,15 +342,12 @@ impl MoveHarness {
     /// of the given account.
     ///
     /// The passed function allows to manipulate the generated metadata for testing purposes.
-    pub fn create_publish_package(
+    pub fn create_publish_built_package(
```
vusirikala, line 345: The comment needs to be changed.

aptos-move/e2e-move-tests/src/harness.rs
```diff
@@ -357,6 +362,58 @@ impl MoveHarness {
...
+
+            cache
+                .entry(path.to_owned())
+                .or_insert_with(|| {
+                    Arc::new(BuiltPackage::build(
+                        path.to_owned(),
+                        BuildOptions::default(),
+                    ))
+                })
+                .clone()
```
vusirikala, line 398: Here, does the `clone` actually clone the package data, or does the `clone` simply creates a new pointer for the package data?

aptos-move/e2e-move-tests/src/harness.rs
```diff
@@ -357,6 +362,58 @@ impl MoveHarness {
...
+        };
+        let package_ref = package_arc
+            .as_ref()
+            .as_ref()
+            .expect("building package must succeed");
```
vusirikala, line 403: Why `.as_ref()` twice?

aptos-move/e2e-move-tests/src/harness.rs
```diff
@@ -357,6 +362,58 @@ impl MoveHarness {
...
+    ) -> SignedTransaction {
+        let package_arc = {
+            let mut cache = CACHED_BUILT_PACKAGES.lock().unwrap();
+
+            cache
+                .entry(path.to_owned())
+                .or_insert_with(|| {
+                    Arc::new(BuiltPackage::build(
+                        path.to_owned(),
+                        BuildOptions::default(),
```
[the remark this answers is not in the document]
igor-aptos: I am leaving the signature of create_publish_package as it was before.
 
 For cases where it's built without options - we can cache it. If someone has custom build options - that means generally that build cannot be cached and reused - so leaving those cases to directly build it.

aptos-move/e2e-move-tests/src/harness.rs
igor-aptos: it clones the Arc

aptos-move/e2e-move-tests/src/harness.rs
igor-aptos: as_ref for the arc, and then as_ref to the Option

aptos-move/e2e-move-tests/src/harness.rs
```diff
@@ -357,6 +362,58 @@ impl MoveHarness {
...
+        self.create_publish_built_package(account, package_ref, patch_metadata)
+    }
+
+    /// Runs transaction which publishes the Move Package.
+    pub fn publish_package_cache_building(
+        &mut self,
+        account: &Account,
+        path: &Path,
+    ) -> TransactionStatus {
+        let txn = self.create_publish_package_cache_building(account, path, |_| {});
```
vusirikala, line 413: What's the use of `patch_metadata`? We always set it to the empty function anyways right?

igor-aptos: I assume some move tests might use it, since it is there.

aptos-move/e2e-move-tests/src/harness.rs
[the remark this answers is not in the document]
vusirikala: I see. So, we are actually saving making multiple calls to `BuiltPackage::build` function. Just curious, what packages are being built for each e2e-move-test?

vusirikala (approved):

aptos-move/e2e-move-tests/src/harness.rs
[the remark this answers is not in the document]
igor-aptos: all the test modules from https://github.com/aptos-labs/aptos-core/tree/main/aptos-move/e2e-move-tests/src/tests

aptos-move/e2e-move-tests/src/harness.rs
georgemitenkov: maybe Arc::clone to avoid confusion

georgemitenkov (approved):

vgao1996 (approved):
Love the optimization! Although I'd like to point out there's a more fundamental issue here.

Right now, if you have a bunch of packages that have a common dependency, and you spawn up a thread for each package to build it, then all of these threads will compete for the ownership of that common dependency, until the whole build is over, keeping everyone else waiting. IMO this is a terrible situation as we are stuck with a single core.

There are two potential solutions to this problem
1. Make the Move package system support some sort of workspace setup, and build all packages as a single CFG
2. Come up with a better file lock implementation that allows us to express 2 kinds of ownerships: exclusive and shared. Exclusive is when you are compiling the package itself, and when you are done, you should change the type of ownership to shared so other people can use it as a dependency.

PR #10069, opened 2023-09-15
Add aggregator_execution_enabled feature flag, and fallback logic

igor-aptos:
recreating as #9996 got closed as it was against Satya's PR.

PR #10075, opened 2023-09-15
[aggregators_v2] Refactor and improve aggregator tests

igor-aptos:
with api enabled and execution disabled, all tests now pass on top of #10069

### DIFF
--- a/aptos-move/e2e-move-tests/src/aggregator_v2.rs
+++ b/aptos-move/e2e-move-tests/src/aggregator_v2.rs
@@ -6,189 +6,411 @@ use aptos_language_e2e_tests::{account::Account, executor::FakeExecutor};
 use aptos_types::{
     account_address::AccountAddress, on_chain_config::FeatureFlag, transaction::SignedTransaction,
 };
-use move_core_types::language_storage::TypeTag;
+use move_core_types::{
+    ident_str,
+    language_storage::{StructTag, TypeTag},
+};
 use std::path::PathBuf;
 
-pub fn initialize(path: PathBuf) -> (MoveHarness, Account) {
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum ExecutorMode {
+    Sequential,
+    // Runs sequential, then parallel, and compares outputs.
+    Both,
+}
+
+pub fn initialize(
+    path: PathBuf,
+    mode: ExecutorMode,
+    aggregator_execution_enabled: bool,
+) -> (MoveHarness, Account) {
     // Aggregator tests should use parallel execution.
-    let executor = FakeExecutor::from_head_genesis().set_parallel();
+    let executor = FakeExecutor::from_head_genesis();
+    let executor = match mode {
+        ExecutorMode::Sequential => executor.set_not_parallel(),
+        // TODO Poorly named function, to rename.
+        ExecutorMode::Both => executor.set_parallel(),
+    };
 
     let mut harness = MoveHarness::new_with_executor(executor);
-    harness.enable_features(vec![FeatureFlag::AGGREGATOR_API], vec![]);
+    if aggregator_execution_enabled {
+        harness.enable_features(
+            vec![
+                FeatureFlag::AGGREGATOR_API,
+                FeatureFlag::AGGREGATOR_EXECUTION,
+            ],
+            vec![],
+        );
+    } else {
+        harness.enable_features(vec![FeatureFlag::AGGREGATOR_API], vec![
+            FeatureFlag::AGGREGATOR_EXECUTION,
+        ]);
+    }
     let account = harness.new_account_at(AccountAddress::ONE);
     assert_success!(harness.publish_package_cache_building(&account, &path));
-    assert_success!(harness.run_entry_function(
-        &account,
-        str::parse("0x1::aggregator_v2_test::initialize").unwrap(),
-        vec![],
-        vec![],
-    ));
     (harness, account)
+}
+
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum UseType {
+    UseResourceType = 0,
+    UseTableType = 1,
+    UseResourceGroupType = 2,
+}
+
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum ElementType {
+    U64,
+    U128,
+    String,
+}
+
+impl ElementType {
+    fn get_type_tag(&self) -> TypeTag {
+        match self {
+            ElementType::U64 => TypeTag::U64,
+            ElementType::U128 => TypeTag::U128,
+            ElementType::String => TypeTag::Struct(Box::new(StructTag {
+                address: AccountAddress::ONE,
+                module: ident_str!("string").to_owned(),
+                name: ident_str!("String").to_owned(),
+                type_params: vec![],
+            })),
+        }
+    }
+
+    fn value_to_bcs(&self, value: u128) -> Vec<u8> {
+        match self {
+            ElementType::U64 => bcs::to_bytes(&(value as u64)),
+            ElementType::U128 => bcs::to_bytes(&value),
+            ElementType::String => bcs::to_bytes(&value.to_string()),
+        }
+        .unwrap()
+    }
+}
+
+#[derive(Debug)]
+pub struct AggLocation<'a> {
+    account: &'a Account,
+    element_type: ElementType,
+    use_type: UseType,
+    index: u64,
+}
+
+impl<'a> AggLocation<'a> {
+    pub fn new(
+        account: &'a Account,
+        element_type: ElementType,
+        use_type: UseType,
+        index: u64,
+    ) -> AggLocation {
+        AggLocation {
+            account,
+            use_type,
+            index,
+            element_type,
+        }
+    }
+}
+
+fn create_entry_agg_func_no_arg(
+    harness: &mut MoveHarness,
+    name: &str,
+    agg_loc: &AggLocation,
+) -> SignedTransaction {
+    harness.create_entry_function(
+        agg_loc.account,
+        str::parse(name).unwrap(),
+        vec![agg_loc.element_type.get_type_tag()],
+        vec![
+            bcs::to_bytes(&(agg_loc.use_type as u32)).unwrap(),
+            bcs::to_bytes(&agg_loc.index).unwrap(),
+        ],
+    )
+}
+
+fn create_entry_agg_func_with_arg(
+    harness: &mut MoveHarness,
+    name: &str,
+    agg_loc: &AggLocation,
+    argument: u128,
+) -> SignedTransaction {
+    harness.create_entry_function(
+        agg_loc.account,
+        str::parse(name).unwrap(),
+        vec![agg_loc.element_type.get_type_tag()],
+        vec![
+            bcs::to_bytes(&(agg_loc.use_type as u32)).unwrap(),
+            bcs::to_bytes(&agg_loc.index).unwrap(),
+            agg_loc.element_type.value_to_bcs(argument),
+        ],
+    )
+}
+
+fn create_entry_agg_func_with_two_args(
+    harness: &mut MoveHarness,
+    name: &str,
+    agg_loc: &AggLocation,
+    argument_1: u128,
+    argument_2: u128,
+) -> SignedTransaction {
+    harness.create_entry_function(
+        agg_loc.account,
+        str::parse(name).unwrap(),
+        vec![agg_loc.element_type.get_type_tag()],
+        vec![
+            bcs::to_bytes(&(agg_loc.use_type as u32)).unwrap(),
+            bcs::to_bytes(&agg_loc.index).unwrap(),
+            agg_loc.element_type.value_to_bcs(argument_1),
+            agg_loc.element_type.value_to_bcs(argument_2),
+        ],
+    )
+}
+
+pub fn init(
+    harness: &mut MoveHarness,
+    account: &Account,
+    use_type: UseType,
+    element_type: ElementType,
+    aggregator: bool,
+) -> SignedTransaction {
+    harness.create_entry_function(
+        account,
+        str::parse(
+            if aggregator {
+                "0x1::aggregator_v2_test::init_aggregator"
+            } else {
+                "0x1::aggregator_v2_test::init_snapshot"
+            },
+        )
+        .unwrap(),
+        vec![element_type.get_type_tag()],
+        vec![bcs::to_bytes(&(use_type as u32)).unwrap()],
+    )
 }
 
 pub fn check(
     harness: &mut MoveHarness,
-    account: &Account,
-    index: u64,
+    agg_loc: &AggLocation,
     expected: u128,
 ) -> SignedTransaction {
-    harness.create_entry_function(
-        account,
-        str::parse("0x1::aggregator_v2_test::check").unwrap(),
-        vec![],
-        vec![
-            bcs::to_bytes(&index).unwrap(),
-            bcs::to_bytes(&expected).unwrap(),
-        ],
-    )
+    create_entry_agg_func_with_arg(harness, "0x1::aggregator_v2_test::check", agg_loc, expected)
 }
 
-pub fn new(
+pub fn check_snapshot(
     harness: &mut MoveHarness,
-    account: &Account,
-    index: u64,
-    limit: u128,
+    snap_loc: &AggLocation,
+    expected: u128,
 ) -> SignedTransaction {
-    harness.create_entry_function(
-        account,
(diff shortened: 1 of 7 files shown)


aptos-move/e2e-move-tests/src/aggregator_v2.rs
```diff
@@ -6,17 +6,31 @@ use aptos_language_e2e_tests::{account::Account, executor::FakeExecutor};
 use aptos_types::{
     account_address::AccountAddress, on_chain_config::FeatureFlag, transaction::SignedTransaction,
 };
-use move_core_types::language_storage::TypeTag;
-use std::path::PathBuf;
+use std::path::Path;
```
vusirikala, line 9: Why is `PathBuf` changed to `Path`?

aptos-move/e2e-move-tests/src/aggregator_v2.rs
```diff
@@ -26,20 +40,31 @@ pub fn initialize(path: PathBuf) -> (MoveHarness, Account) {
...
 }
 
+fn create_2_arg(
+    harness: &mut MoveHarness,
+    account: &Account,
+    name: &str,
+    index: u64,
+    second: u128,
+) -> SignedTransaction {
+    harness.create_entry_function(account, str::parse(name).unwrap(), vec![], vec![
```
vusirikala, line 50: Generally speaking, should we leave the `TypeTag` as empty for all the tests? In the native functions, we explicitly read the type tag and behave according to it.

aptos-move/e2e-move-tests/src/tests/aggregator_v2.data/pack/sources/aggregator_v2_test.move
```diff
@@ -2,27 +2,28 @@ module 0x1::aggregator_v2_test {
...
     /// determinictic integer value, for testing purposes.
-    struct AggregatorStore has key, store {
-        aggregators_u128: Table<u64, Aggregator<u128>>,
-        aggregators_u64: Table<u64, Aggregator<u64>>,
-        aggregator_snapshots_u128: Table<u64, AggregatorSnapshot<u128>>,
-        aggregator_snapshots_u64: Table<u64, AggregatorSnapshot<u64>>,
+    /// We put multiple (10) aggregators/snapshots into same resource, to be
+    /// able to test multiple aggregators/snapshots being inside same resource.
+    struct AggregatorStore<Element: store> has key, store {
+        data: Table<u64, vector<Option<Element>>>,
```
vusirikala, line 17: Why `Option`?

aptos-move/e2e-move-tests/src/tests/aggregator_v2.data/pack/sources/aggregator_v2_test.move
```diff
@@ -51,120 +52,171 @@ module 0x1::aggregator_v2_test {
...
+        let data = &borrow_global<AggregatorStore<Element>>(addr).data;
+        let outer = i / 10;
+        let inner = i % 10;
+        let vector_data = table::borrow(data, outer);
+        let option_data = vector::borrow(vector_data, inner);
+        let value = option::borrow(option_data);
+
+        f(value)
     }
+    inline fun for_element_mut<Element: store, R>(account: &signer, i: u64, f: |&mut Element|R): R acquires AggregatorStore {
```
vusirikala, line 85: nit: new line between functions.

aptos-move/e2e-move-tests/src/tests/aggregator_v2.data/pack/sources/aggregator_v2_test.move
```diff
@@ -51,120 +52,171 @@ module 0x1::aggregator_v2_test {
...
+        if (!table::contains(data, outer)) {
+            table::add(data, outer, vector::empty());
+        };
 
-    /// Checks that the ith aggregator has expected value. Useful to inject into
-    /// transaction block to verify successful and correct execution.
-    public entry fun check(account: &signer, i: u64, expected: u128) acquires AggregatorStore {
+        let vector_data = table::borrow_mut(data, outer);
+        if (vector::length(vector_data) == inner) {
+            vector::push_back(vector_data, option::none());
```
vusirikala, line 66: Not entirely sure the motivation behind this way of organizing the data in AggregatorStore. 
Why not create an aggregator/snapshot and insert at ith index of the data?

aptos-move/e2e-move-tests/src/tests/aggregator_v2.data/pack/sources/aggregator_v2_test.move
```diff
@@ -51,120 +52,171 @@ module 0x1::aggregator_v2_test {
...
 
     public entry fun new(account: &signer, i: u64, limit: u128) acquires AggregatorStore {
-        let addr = signer::address_of(account);
-        let aggregators = &mut borrow_global_mut<AggregatorStore>(addr).aggregators_u128;
-        let aggregator = aggregator_v2::create_aggregator(limit);
-        table::add(aggregators, i, aggregator);
+        insert<Aggregator<u128>>(account, i, aggregator_v2::create_aggregator(limit));
+    }
+
+    public fun call_read(account: &signer, i: u64): u128 acquires AggregatorStore {
```
vusirikala, line 101: `read` instead of `call_read`?

aptos-move/e2e-move-tests/src/tests/aggregator_v2.data/pack/sources/aggregator_v2_test.move
```diff
@@ -51,120 +52,171 @@ module 0x1::aggregator_v2_test {
...
-        let addr = signer::address_of(account);
-        let aggregators = &mut borrow_global_mut<AggregatorStore>(addr).aggregators_u128;
-        let aggregator = table::borrow_mut(aggregators, i);
-        aggregator_v2::try_add(aggregator, value);
+        for_element_mut<Aggregator<u128>, bool>(account, i, |aggregator| aggregator_v2::try_add(aggregator, value));
+    }
+
+    public entry fun add(account: &signer, i: u64, value: u128) acquires AggregatorStore {
+        for_element_mut<Aggregator<u128>, bool>(account, i, |aggregator| {
+            aggregator_v2::add(aggregator, value);
```
vusirikala, line 111: We don't have an `add` function in aggregator_v2 right?

aptos-move/e2e-move-tests/src/tests/aggregator_v2.data/pack/sources/aggregator_v2_test.move
```diff
@@ -51,120 +52,171 @@ module 0x1::aggregator_v2_test {
...
+            true
```
vusirikala, line 112: The `add` function always returns true? Shouldn't the output depend on the output of `try_add`?

aptos-move/e2e-move-tests/src/tests/aggregator_v2.data/pack/sources/aggregator_v2_test.move
```diff
@@ -51,120 +52,171 @@ module 0x1::aggregator_v2_test {
...
 
-    public entry fun try_sub_add(account: &signer, i: u64, a: u128, b: u128) acquires AggregatorStore {
-        let addr = signer::address_of(account);
-        let aggregators = &mut borrow_global_mut<AggregatorStore>(addr).aggregators_u128;
-        let aggregator = table::borrow_mut(aggregators, i);
-        aggregator_v2::try_sub(aggregator, a);
-        aggregator_v2::try_add(aggregator, b);
+    public entry fun sub(account: &signer, i: u64, value: u128) acquires AggregatorStore {
+        for_element_mut<Aggregator<u128>, bool>(account, i, |aggregator| {
+            aggregator_v2::sub(aggregator, value);
```
vusirikala, line 122: Same here. We don't have `sub` in aggregator v2 right?

aptos-move/e2e-move-tests/src/tests/aggregator_v2.data/pack/sources/aggregator_v2_test.move
```diff
@@ -51,120 +52,171 @@ module 0x1::aggregator_v2_test {
...
+            true
+        });
     }
 
     public entry fun materialize(account: &signer, i: u64) acquires AggregatorStore {
-        let addr = signer::address_of(account);
-        let aggregators = &borrow_global<AggregatorStore>(addr).aggregators_u128;
-        let aggregator = table::borrow(aggregators, i);
-        aggregator_v2::read(aggregator);
+        call_read(account, i);
```
vusirikala, line 128: If materialize and call_read are the same, why have two functions?

aptos-move/e2e-move-tests/src/tests/aggregator_v2.rs
```diff
@@ -100,105 +104,214 @@ fn test_aggregators_e2e() {
...
 
-#[test]
-#[should_panic]
-fn test_aggregator_overflow() {
-    let (mut h, acc) = setup();
+        let txns = vec![
+            (0, new(&mut h, &acc, 0, 600)),
+            (0, add(&mut h, &acc, 0, 400)),
+            // Value dropped below zero - abort with EAGGREGATOR_UNDERFLOW.
+            (131074, sub(&mut h, &acc, 0, 500))
```
vusirikala, line 227: Should we call `try_add`, `try_sub` instead? 
Also, `try_sub` would return false here instead of an error code right?

aptos-move/e2e-move-tests/src/tests/aggregator_v2.rs
```diff
@@ -100,105 +104,214 @@ fn test_aggregators_e2e() {
...
+            first_block,
+            second_block,
+            txns,
+        );
+    }
 
-    let txn1 = new(&mut h, &acc, 0, 600);
-    let txn2 = materialize_and_try_sub(&mut h, &acc, 0, 400);
+    #[test]
+    fn test_aggregator_underflow((first_block, second_block) in arb_block_split(3)) {
```
vusirikala, line 218: Why are all the test cases taking `first_block`, `second_block` as input? What are these?

aptos-move/e2e-move-tests/src/tests/aggregator_v2.rs
```diff
@@ -100,105 +104,214 @@ fn test_aggregators_e2e() {
...
-    let txn1 = new(&mut h, &acc, 0, 600);
-    let txn2 = try_add(&mut h, &acc, 0, 400);
-    let txn3 = try_sub(&mut h, &acc, 0, 500);
+proptest! {
+    #![proptest_config(ProptestConfig {
+        // Cases are expensive, few cases is enough.
+        cases: 5,
+        result_cache: prop::test_runner::basic_result_cache,
+        .. ProptestConfig::default()
+    })]
```
vusirikala, line 185: What does this statement do?
What's the difference between the proptests and the regular Rust tests we had earlier?

aptos-move/e2e-move-tests/src/tests/aggregator_v2.data/pack/sources/aggregator_v2_test.move
```diff
@@ -51,120 +52,171 @@ module 0x1::aggregator_v2_test {
...
+        f(value)
+    }
+
 
     public entry fun new(account: &signer, i: u64, limit: u128) acquires AggregatorStore {
-        let addr = signer::address_of(account);
-        let aggregators = &mut borrow_global_mut<AggregatorStore>(addr).aggregators_u128;
-        let aggregator = aggregator_v2::create_aggregator(limit);
-        table::add(aggregators, i, aggregator);
+        insert<Aggregator<u128>>(account, i, aggregator_v2::create_aggregator(limit));
```
vusirikala, line 98: As we treat newly created aggregators and existing aggregators differently, can we have some transactions that creates an aggregator and does add/sub/read operations in the same transaction?

aptos-move/e2e-move-tests/src/tests/aggregator_v2.rs
igor-aptos: proptest allows you to run a parametrized test. 
I.e. to have a test with arguments, and proptest will run it multiple times with different arguments.
Here, I paramtrized how transactions are split into blocks - i.e. whether a whole test is run as a single block, or it is split in two or three blocks, and random positions (first_block and second_block are lenghts of each block)

This config does two things:
- makes every test run 5 times. as tests are expensive, default of 1000 for proptest is not practical here, and 5 gives enough variety
- "caches" the run, so it doesn't call the same test with same arguments twice (default is to allow calling same test with same arguments twice, for some reason)

aptos-move/e2e-move-tests/src/tests/aggregator_v2.rs
igor-aptos: tests you copied are from v1 version. So having them be identical - is starting way to get them to run. It doesn't make much difference for these tests.

will add separately some new tests, where we can use try_add/try_sub

aptos-move/e2e-move-tests/src/tests/aggregator_v2.data/pack/sources/aggregator_v2_test.move
igor-aptos: entry function cannot return result, it needs to be void

igor-aptos:
addressed comments. all tests from the file pass now, that fallback PR has landed

georgemitenkov (changes requested):
Looks good! I think the Move code right now is a bit ugly, testing different things at different places seems cleaner to me

aptos-move/e2e-move-tests/src/tests/aggregator_v2.data/pack/sources/aggregator_v2_test.move
igor-aptos: easier to check correctness

aptos-move/framework/src/natives/aggregator_natives/aggregator_v2.rs
```diff
@@ -255,6 +255,7 @@ fn native_try_add(
         }
     };
 
+    println!("native_try_add(agg(max={}, value={}), input={}) = {}", agg_max_value, agg_value, input, result_value);
```
georgemitenkov, line 258: Are prints intentional?

aptos-move/e2e-move-tests/src/tests/aggregator_v2.data/pack/sources/aggregator_v2_test.move
```diff
@@ -1,170 +1,324 @@
...
+
+            table::borrow_mut(data, outer)
+        } else { // if (use_type == USE_RESOURCE_GROUP_TYPE) {
+            &mut borrow_global_mut<AggregatorInResourceGroup<Agg>>(account_addr).data
+        };
+
+        if (vector::length(vector_data) == i) {
+            vector::push_back(vector_data, option::some(e));
+        } else {
+            let option_data = vector::borrow_mut(vector_data, i);
```
georgemitenkov, line 111: This will give out of bounds if i > length?

aptos-move/e2e-move-tests/src/tests/aggregator_v2.data/pack/sources/aggregator_v2_test.move
```diff
@@ -1,170 +1,324 @@
...
+    const USE_RESOURCE_GROUP_TYPE: u32 = 2;
 
     /// When checking the value of aggregator fails.
     const ENOT_EQUAL: u64 = 17;
 
-    /// Resource to store aggregators. Each aggregator is associated with a
+    const EINVALID_ARG: u64 = 18;
+
+    struct AggregatorInResource<Agg: store> has key, store {
+        data: vector<Option<Agg>>,
```
georgemitenkov, line 20: Why options?

aptos-move/e2e-move-tests/src/tests/aggregator_v2.data/pack/sources/aggregator_v2_test.move
igor-aptos: yes, that's fine for current tests (you need to add them sequentially. we can change later if needed.

aptos-move/e2e-move-tests/src/aggregator_v2.rs
```diff
@@ -26,20 +40,31 @@ pub fn initialize(path: PathBuf) -> (MoveHarness, Account) {
     (harness, account)
 }
 
+fn create_2_arg(
+    harness: &mut MoveHarness,
+    account: &Account,
+    name: &str,
+    index: u64,
```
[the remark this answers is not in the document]
georgemitenkov: I think this is a stale comment, there was an argument called "second" before or similar, ignore

PR #10081, opened 2023-09-17
[Executor] Refactoring and Optimization of the Block STM

ziaptos:
### A series of the commits (so that it's easier to review)

- Removed CommitGuard in the executor - using the standard scopeguard
- Any thread can be a coordinator (The term is not used any more). Attempting to coordinate only after a successful validation. Minor optimizations
- Parallel processing of outputs
- Fixed a potential lost wake-up
- Fixed swapped memory orderings

Additionally, for notifications (e.g. for sharding) this commit ensures that
- if a transaction can be committed after validation it will get committed
- there may not be "fake" abort notifications (due to an earlier skiprest).

### Test Plan
Current tests should cover all the changes.


aptos-move/block-executor/src/executor.rs
```diff
@@ -322,9 +294,6 @@ where
                         !matches!(scheduler_task, SchedulerTask::ExecutionTask(_, _)),
```
gelash, line 294: This is now the only reason we pass scheduler_task. I think the method can return something if it commits the last thing, and we can assert at the caller (we could get rid of it too, but the assert seems somewhat useful)

aptos-move/block-executor/src/executor.rs
```diff
@@ -359,7 +328,7 @@ where
         }
     }
 
-    fn worker_commit_hook(
```
gelash, line 362: is post-commit post processing of the outputs, than the commit, commit seems more natural to refer to synchronization to know something is truly finalized, which has happened already (by the coordinator).

aptos-move/block-executor/src/scheduler.rs
```diff
@@ -1,10 +1,15 @@
+#![allow(dead_code)]
```
gelash, line 1: ?

gelash (approved):

aptos-move/block-executor/src/scheduler.rs
```diff
@@ -21,6 +26,35 @@ const TXN_IDX_MASK: u64 = (1 << 32) - 1;
...
+pub struct ArmedLock {
+    // Last bit:   1 -> unlocked; 0 -> locked
+    // Second bit: 1 -> there's work; 0 -> no work
+    locked: AtomicU64,
+}
+
+impl ArmedLock {
+    pub fn new() -> Self {
+        Self {
+            locked: AtomicU64::new(1), //
```
gelash, line 39: trailing //

aptos-move/block-executor/src/executor.rs
```diff
@@ -263,6 +229,11 @@ where
             scheduler.finish_abort(idx_to_validate, incarnation)
         } else {
             scheduler.finish_validation(idx_to_validate, validation_wave);
+
+            if valid {
```
gelash, line 233: I'd love if we can come up with another precondition for when to arm (e.g. validated index happens to be the next one that needs to be committed). If we do that, we may even be able to just use that as an indicator to do the coordinating duty right after (without extra logic for it there)

aptos-move/block-executor/src/executor.rs
```diff
@@ -424,46 +393,32 @@ where
...
-                    while let Ok(txn_idx) = rx.try_recv() {
-                        self.worker_commit_hook(
-                            txn_idx,
-                            versioned_cache,
-                            last_input_output,
-                            base_view,
-                        );
-                    }
-                },
+            if scheduler.should_coordinate_commits() {
```
gelash, line 407: this would have to be while scheduler.should_coordinate_commits() {}

otherwise, the thread unlocks after someone armed (but couldn't get the lock), they both get stuck on dependencies while executing the next task right after (before coming back to the above loop), and if progress depends on another shard unblocking us, that shard may be waiting for the commit notification.
If this was a while, then one of them will for sure commit anything that can be committed (based on the latest arm)

gelash: @zjma : this is re:resolving the cross-shard coordinating-related deadlock. The PR either solves it already, or should be really close, if you also want to have a look / confirm / join the discussion.

zjma: Took a quick look. i think this solves the sharding issue. Will run some separate experiments to verify.

aptos-move/block-executor/src/executor.rs
ziaptos: I'll rename all *hook functions - they are not really `hooks` anymore. perhaps: prepare_commits/process_commits?

aptos-move/block-executor/src/scheduler.rs
ziaptos: Dead code society

aptos-move/block-executor/src/executor.rs
ziaptos: Cool

aptos-move/block-executor/src/executor.rs
```diff
@@ -438,46 +406,32 @@ where
...
-                    while let Ok(txn_idx) = rx.try_recv() {
-                        self.worker_commit_hook(
-                            txn_idx,
-                            versioned_cache,
-                            last_input_output,
-                            base_view,
-                        );
-                    }
-                },
+            while scheduler.should_coordinate_commits() {
```
gelash, line 420: If you consider it appropriate, maybe we can actually comment here saying that while ensures that any task that can be committed (e.g. due to a validation the worker did right before) gets committed before moving on, or something to that extent.

ziaptos: Yep - a comment is appropriate. Tho, `before moving on` is vague (and possibly incorrect - some other thread might be `moving on`)

aptos-move/block-executor/src/executor.rs
```diff
@@ -271,27 +237,32 @@ where
...
-            let _commit_guard: CommitGuard =
-                CommitGuard::new(post_commit_txs, *worker_idx, txn_idx);
-            // Iterate round robin over workers to do commit_hook.
-            *worker_idx = (*worker_idx + 1) % post_commit_txs.len();
+            let mut txn_fee_state = txn_fee_state.lock();
+            let txn_state_pair = txn_fee_state.deref_mut();
+            let (accumulated_fee_statement, txn_fee_statements) =
+                (&mut txn_state_pair.0, &mut txn_state_pair.1);
+
+            defer! {
```
zekun000, line 263: TIL, didn't know we have this go-style macro too 😄

aptos-move/block-executor/src/executor.rs
```diff
@@ -553,69 +549,48 @@ where
...
+        ));
 
-        let mut roles: Vec<CommitRole> = vec![];
-        let mut senders: Vec<Sender<u32>> = Vec::with_capacity(self.concurrency_level - 1);
-        for _ in 0..(self.concurrency_level - 1) {
-            let (tx, rx) = mpsc::channel();
-            roles.push(CommitRole::Worker(rx));
-            senders.push(tx);
+        let final_results = ExplicitSyncWrapper::new(Vec::with_capacity(num_txns));
+        // Initialize for [] access, takes less than <1ms (on Apple M1) for 50k block.
```
gelash, line 561: let's get rid of this comment and re-check this is fine. The number's coming from me but I quickly timed with a test run, and tests likely use mock types.

aptos-move/block-executor/src/executor.rs
```diff
@@ -32,50 +33,38 @@ use aptos_vm_logging::{clear_speculative_txn_logs, init_speculative_logs};
...
+    marker::Sync,
+    ops::DerefMut,
+    sync::{atomic::AtomicU32, Arc},
 };
 
-struct CommitGuard<'a> {
-    post_commit_txs: &'a Vec<Sender<u32>>,
-    worker_idx: usize,
-    txn_idx: u32,
+struct ExplicitSyncWrapper<T> {
```
gelash, line 44: Let's put this in its own file. Can be in block executor for now I think.

danielxiangzl (commented):
Did we benchmark the performance change?

aptos-move/block-executor/src/executor.rs
```diff
@@ -412,7 +385,10 @@ where
             // Must contain committed value as we set the base value above.
             delta_writes.push((k, WriteOp::Modification(serialize(&committed_delta).into())));
         }
+
         last_input_output.record_delta_writes(txn_idx, delta_writes);
+
+        // TODO(zi) Make sure this here is perfectly reentrant, and doesn't introduce regression.
```
igor-aptos, line 391: not sure what is comment is exactly worried about, and whether you plan to address it before landing this PR?

aptos-move/block-executor/src/executor.rs
```diff
@@ -271,27 +228,35 @@ where
...
+            if valid {
+                scheduler.queueing_commits_arm();
+            }
+
             SchedulerTask::NoTask
         }
     }
 
-    fn coordinator_commit_hook(
+    fn queue_ready_commits(
```
igor-aptos, line 240: for the naming here, this does do some work, and with aggregators_v2 will do even more work (validate is speculation was correct, and if not, re-execute transaction inline, to produce the final output)

maybe calling it prepare_and_queue_ready_commits ? (maybe even prepare_and_queue_for_materialization_ready_commits  :) )

aptos-move/block-executor/src/executor.rs
```diff
@@ -367,12 +339,13 @@ where
         }
     }
 
-    fn worker_commit_hook(
+    fn commit_epilogue(
```
igor-aptos, line 342: epilogue means something else in VM, and this doesn't refer to that - so it is a big confusing.

maybe calling it commit_materialization / commit_materialize_results / commit_finalization / commit_compute_final_results ?

aptos-move/block-executor/src/explicit_sync_wrapper.rs
```diff
@@ -0,0 +1,39 @@
+// Copyright © Aptos Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+use std::cell::UnsafeCell;
+
+// Parallel algorithms often guarantee a sequential use of certain
+// data structures, or parts of the data-structures (like elements of
+// a vector).  The rust compiler can not profe the safety of even
```
igor-aptos, line 8: typo - prove

aptos-move/block-executor/src/executor.rs
zekun000: there's a hidden assumption that this `commit_epilogue` can be executed out of order, we probably want to call it out explicitly

aptos-move/block-executor/src/executor.rs
```diff
@@ -624,12 +581,10 @@ where
             drop(versioned_cache);
         });
 
-        match maybe_err {
+        let maybe_error = shared_commit_state.into_inner();
```
zekun000, line 584: explicit tuple unpack is probably more readable, let (_, _, maybe_error) = ...

aptos-move/block-executor/src/explicit_sync_wrapper.rs
```diff
@@ -0,0 +1,39 @@
...
+// slightly complex parallel algorithms.
+
+/// ExplicitSyncWrapper is meant to be used in parallel algorithms
+/// where we can prove that there will be no concurrent access to the
+/// underlying object (or its elements).  Use with caution - only when
+/// the safety can be proven.
+pub struct ExplicitSyncWrapper<T> {
```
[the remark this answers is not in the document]
igor-aptos: Yeah, I wanted to complain about the name, but couldn't find the better one. On the first read "Sync" suggest being tread safe, and it's the opposite. 

I think the naming means "it is for objects that will have "ExplicitSync" done externally before calling them". maybe ExternalSync is cleaner.
Are we doing wrapper just for the name? UnsafeCell sounds like a better name than ExplicitSyncWrapper :)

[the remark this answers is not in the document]
zekun000: I think we're doing the wrappers to hide the unsafe functions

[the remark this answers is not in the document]
ziaptos: Yep - I don't really like the name either. Perhaps ConcurrentAccessUnsafeCell? Or something similar - open to suggestions.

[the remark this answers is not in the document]
igor-aptos: I mean - they don't become safe, caller makes them safe :)

or does the compiler/linter complain about unsafe functions elsewhere?

aptos-move/block-executor/src/executor.rs
```diff
@@ -199,15 +164,7 @@ where
...
-        if last_input_output
-            .record(idx_to_execute, sync_view.take_reads(), result)
-            .is_err()
-        {
-            // When there is module publishing r/w intersection, can early halt BlockSTM to
-            // fallback to sequential execution.
-            scheduler.halt();
-            return SchedulerTask::NoTask;
-        }
+        last_input_output.record(idx_to_execute, sync_view.take_reads(), result);
```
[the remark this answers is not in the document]
ziaptos: This was an optimization (halting sightly earlier when falling-back to serial execution).  The tiny optimization on the slow path is now removed compromising for an optimization of the fast path.

aptos-move/block-executor/src/executor.rs
ziaptos: @igor-aptos I hope we'll not bloat this function with complicated functionality, as this is the one executed serially.

About the name, how about `prepare_and_queue_commit_ready_txns`

aptos-move/block-executor/src/explicit_sync_wrapper.rs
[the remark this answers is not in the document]
ziaptos: Yes - the user is expected to use them in a safe manner. 

I still don't have the best choice for the name tho. (pressing F1, F1...)

aptos-move/block-executor/src/executor.rs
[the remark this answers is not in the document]
ziaptos: There was also a comment regarding this optimization at the end for the queue function - it's now removed, as it's not relevant anymore.

ziaptos (commented):
Addressed all but the naming of the ExplicitSyncWrapper struct.  Could use help here. I'd go with UnsafeSyncWrapper, or UnsafeSyncCell (Don't really like the cell one as much, as the APIs are different from UnsafeCell, and I don't want to have them same as UnsafeCell).

ziaptos:
> Did we benchmark the performance change?

Yes - I shared a document with you. The gist is that the changes are strictly beneficial. BlockSTM benchmarks show up to 15% improvements. End-to-end benchmarks show ~6% for 50k blocks.

aptos-move/block-executor/src/executor.rs
ziaptos: That was a not to myself - was not suppose to stay here. Thanks for catching it.

zekun000 (approved):
name can be updated later

aptos-move/block-executor/src/executor.rs
```diff
@@ -264,6 +274,7 @@ where
             }
 
             if last_input_output.execution_error(txn_idx).is_some() {
+                let mut maybe_error = maybe_error.lock();
```
gelash, line 275: Are we now overwriting the errors? The flow isn't very clear?
Also can we call halt twice? what will happen in that case?

gelash (changes requested):
I like the simpler solution better. Also don't like Mutex inside SyncWrapper

gelash:
Let's update the description as well to mention parallel processing of outputs, as well as the fact that whatever can be committed is now guaranteed to be committed (useful for sharding notifications).

It's also getting increasingly difficult to review based on a long list of commits.

aptos-move/block-executor/src/executor.rs
ziaptos: The variable might. Tho the final result will be the same. The early exit can only halt with Abort (not skip rest)

gelash: Ok, so I missed return (let's do else as you suggested), so higher abort can't overwrite, should be fine now but let me think some more.

aptos-move/block-executor/src/executor.rs
```diff
@@ -199,13 +165,11 @@ where
...
-            .is_err()
-        {
-            // When there is module publishing r/w intersection, can early halt BlockSTM to
-            // fallback to sequential execution.
-            scheduler.halt();
+        if !last_input_output.record(idx_to_execute, sync_view.take_reads(), result) {
+            if scheduler.halt() {
+                let mut maybe_error = maybe_error.lock();
+                *maybe_error = Some(Error::ModulePathReadWrite);
+            }
```
gelash, line 172: let's do else

gelash (approved):

gelash:
- Removed CommitGuard in the executor - using the standard scopeguard
- Any thread can be a coordinator (The term is not used any more). Attempting to coordinate only after a successful validation. Minor optimizations
- Run 1 test at the time (as each test is multithreaded).
- Parallel processing of outputs

Additionally, for notifications (e.g. for sharding) this commit ensures that
- if a transaction can be committed after validation it will get committed
- there may not be "fake" abort notifications (due to an earlier skiprest).

aptos-move/block-executor/src/explicit_sync_wrapper.rs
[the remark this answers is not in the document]
igor-aptos: oh I replied to Zekun, not you @ziaptos , I see order of messages is different now. I think ConcurrentAccessUnsafeCell is better, and we can change later if we have an even better one :)

feel free to change that in a separate PR

aptos-move/block-executor/src/executor.rs
```diff
@@ -250,27 +215,34 @@ where
...
-        post_commit_txs: &Vec<Sender<u32>>,
-        worker_idx: &mut usize,
         scheduler_task: &mut SchedulerTask,
         last_input_output: &TxnLastInputOutput<T, E::Output, E::Error>,
-        accumulated_fee_statement: &mut FeeStatement,
-        txn_fee_statements: &mut Vec<FeeStatement>,
+        shared_commit_state: &ExplicitSyncWrapper<(
+            FeeStatement,
+            Vec<FeeStatement>,
+            Mutex<Option<Error<E::Error>>>,
```
gelash, line 236: we shouldn't need the Mutex anymore?

aptos-move/block-executor/src/executor.rs
```diff
@@ -250,27 +218,34 @@ where
...
         while let Some(txn_idx) = scheduler.try_commit() {
-            // Create a CommitGuard to ensure Coordinator sends the committed txn index to Worker.
-            let _commit_guard: CommitGuard =
-                CommitGuard::new(post_commit_txs, *worker_idx, txn_idx);
-            // Iterate round robin over workers to do commit_hook.
-            *worker_idx = (*worker_idx + 1) % post_commit_txs.len();
+            let (accumulated_fee_statement, txn_fee_statements, maybe_error) =
+                shared_commit_state.get_mut();
+
+            defer! {
```
gelash, line 246: let's put this in the beginning like the guard?

aptos-move/block-executor/src/executor.rs
```diff
@@ -417,46 +405,44 @@ where
...
+                    self.maybe_block_gas_limit,
+                    scheduler,
+                    &mut scheduler_task,
+                    last_input_output,
+                    shared_commit_state,
+                );
+                scheduler.queueing_commits_mark_done();
+            }
+
+            {
```
gelash, line 436: why these brackets? what's scoped here? let's remove if we can, extra indentation doesn't help reviewing, for instance.

aptos-move/block-executor/src/scheduler.rs
```diff
@@ -779,6 +829,9 @@ impl Scheduler {
     /// Return false when the execution is halted.
     fn suspend(&self, txn_idx: TxnIndex, dep_condvar: DependencyCondvar) -> bool {
         let mut status = self.txn_status[txn_idx as usize].0.write();
+        if self.done_marker.load(Ordering::Acquire) {
```
gelash, line 832: I checked, they already synchronize on the status.
halt acquires the status and sets to ExecutionHalted, and hence if suspend happens after that it will return false.
if suspend happens before, halt will observe Suspended & wake it up.

So we shouldn't need this check.

aptos-move/block-executor/src/scheduler.rs
```diff
@@ -262,20 +295,48 @@ impl Scheduler {
...
     /// If successful, returns Some(TxnIndex), the index of committed transaction.
     /// The current implementation has one dedicated thread to try_commit.
     /// Should not be called after the last transaction is committed.
     pub fn try_commit(&self) -> Option<TxnIndex> {
         let mut commit_state_mutex = self.commit_state.lock();
         let commit_state = commit_state_mutex.deref_mut();
-        let (commit_idx, commit_wave) = (&mut commit_state.0, &mut commit_state.1);
+        let (commit_idx, commit_wave) = commit_state;
+
+        if *commit_idx == self.num_txns {
```
gelash, line 337: The comment says "should not be called after the last transaction is committed".
What change required doing this?

aptos-move/block-executor/src/executor.rs
```diff
@@ -199,16 +165,18 @@ where
...
-        {
-            // When there is module publishing r/w intersection, can early halt BlockSTM to
-            // fallback to sequential execution.
-            scheduler.halt();
-            return SchedulerTask::NoTask;
-        }
+        let _ignore = last_input_output.record(idx_to_execute, sync_view.take_reads(), result);
         scheduler.finish_execution(idx_to_execute, incarnation, updates_outside)
+
+        // if !last_input_output.record(idx_to_execute, sync_view.take_reads(), result) {
```
gelash, line 171: This only changes behavior when there are actual Module R/W conflicts, and after you tried it, the tests that hang were module r/w tests. So maybe here it was actually halting properly, but the other part has some problem?

aptos-move/block-executor/src/scheduler.rs
```diff
@@ -558,22 +606,24 @@ impl Scheduler {
     /// It will set the done_marker to be true, resolve all pending dependencies.
     ///
     /// Currently there are 4 scenarios to early halt the BlockSTM execution.
-    /// 1. There is a module publishing txn that has read/write intersection with any txns even during speculative execution.
+    /// 1. There is a module publishing txn that has read/write intersection with any txns even during speculative execution.t
```
gelash, line 609: extra t at the end.

aptos-move/block-executor/src/executor.rs
```diff
@@ -417,46 +405,44 @@ where
...
-                        scheduler,
-                        post_commit_txs,
-                        &mut worker_idx,
-                        &mut scheduler_task,
+            // Priorotize committing validated transactions
+            while scheduler.should_coordinate_commits() {
+                self.prepare_and_queue_commit_ready_txns(
+                    self.maybe_block_gas_limit,
+                    scheduler,
+                    &mut scheduler_task,
```
gelash, line 429: We used to assign Done to this, and you removed it, so we shouldn't need &mut. I would also probably not even pass it anymore.

aptos-move/block-executor/src/executor.rs
```diff
@@ -250,27 +218,34 @@ where
             scheduler.finish_abort(idx_to_validate, incarnation)
         } else {
             scheduler.finish_validation(idx_to_validate, validation_wave);
```
gelash, line 220: finish validation internally has the check for updating the wave. We should return true if it updated the wave, and use that for arming? If it didn't there is no new information and reason to arm.

Cargo.toml
```diff
@@ -718,6 +719,7 @@ inherits = "release"
 debug = true
 overflow-checks = true
 debug-assertions = true
+panic = "abort"
```
gelash, line 722: let's not forget to remove this later.

aptos-move/block-executor/src/scheduler.rs
```diff
@@ -20,6 +20,48 @@ const TXN_IDX_MASK: u64 = (1 << 32) - 1;
 
 pub type Wave = u32;
 
+#[derive(Debug)]
+pub struct ArmedLock {
+    // Last bit:   1 -> unlocked; 0 -> locked
+    // Second bit: 1 -> there's work; 0 -> no work
+    // locked: AtomicU64,
```
gelash, line 27: let's bring back the Atomic, right?

PR #10086, opened 2023-09-18
Update delegated-staking.md

rtmtree:
[update] turn "1M APT" into "the minimum stake amount" in Delegation Page

### Description

It's better to use a term "the minimum stake amount" instead of a fixed number "1M APT" in case we are updating it to other number in the future.


davidiw:
We intend to keep this as is, since it is current and maintained. Thanks for the suggestion.

PR #10087, opened 2023-09-18
Update system-integrators-guide.md

rtmtree:
The old url file doesn't exist anymore. must be changed to the new one.

### Description

`transfer-two-by-two.py` doesn't not exist, instead changed into `transfer_two_by_two.py`
This is an update of document referencing to it.


davidiw (approved):

davidiw:
thanks

PR #10091, opened 2023-09-18
[indexer-grpc] Overhaul configuration, error handling

banool:
## Note for the reviewer
The debate between large and small PRs rages on. The people I'm working with lately prefer large PRs so here we go. 

Looking for review from Josh for the aptos-node stuff, Greg for the CLI stuff, and Larry / Bowen for all the other stuff.

If this PR is too unwieldy I can split up the stuff that is expressly blocking https://github.com/aptos-labs/aptos-core/pull/10101 and the other general improvements.

## Description
This PR improves how we configure the txn stream services in a variety of ways following these principles:
- Add a `validate` function to RunnableConfig.
- Use enums rather than decoupled options.
- Use the types we want directly in the config, e.g. Url rather than String.
- Prevent configs that never work.
- Use Result rather than panicking unless the error is truly unrecoverable / should never happen.
- When in a function that returns anyhow::Result, use anyhow::Context rather than map_err.
- Leverage Default rather than handling it repeatedly throughout the code with unwrap_or. Do the same thing in the constructors.
- Use newtypes to improve validation further.

I also update how we build the node config for local testing. Previously configuration was split awkwardly between two functions in a way that meant it was impossible to get the "fixed" config without running the node. Now you can, which is neato because I want that for the CLI so I can set necessary flags.

This PR is one step towards improving the quality of the indexer code. I'm tracking issues here: https://www.notion.so/aptoslabs/Indexer-Notes-5e0eb119e58d4069bd4df7c1c5effea9.

I have updated the configs in aptos-core accordingly. I'll go change internal-ops later.

## Test Plan
Local testnet:
```
cargo run -p aptos -- node run-local-testnet
```

```
cd ecosystem/indexer-grpc/indexer-grpc-integration-tests
cargo test
```

CI, in particular forge and the integration tests for indexer-grpc.


ecosystem/indexer-grpc/indexer-grpc-cache-worker/src/worker.rs
```diff
@@ -343,4 +362,8 @@ async fn process_streaming_response(
             },
         }
     }
+
+    // It is expected that we get to this point, the upstream server disconnects
+    // clients after 5 minutes.
+    Ok(())
```
larry-aptos, line 368: Yea the upstream may get disconnected; but it's most likely be indexer node update.

ecosystem/indexer-grpc/indexer-grpc-cache-worker/src/worker.rs
```diff
@@ -79,14 +87,15 @@ impl Worker {
     ///    * If metadata is not present and cache is not empty, crash.
     ///    * If metadata is present, start from file store version.
     /// 4. Process the streaming response.
-    pub async fn run(&mut self) {
+    // TODO: Use the ! return type when it is stable.
```
larry-aptos, line 90: 👍 TIL

larry-aptos (approved):
LGTM from indexer side.

One note for the auth token: we can remove the auth token check later. it is the alternative before api gateway is out. 

now the gateway is ready to launch and we can hide the actual serivce from the internet; and all traffic can be routed from api gateway.

banool:
Note, it is expected that `Docker Indexer gRPC test / test-indexer-grpc-docker-compose (pull_request)` fails on this PR because of how it pulls the images. Once we land this to main it will pass.

bowenyang007:
Who are you working with who prefers larger RPs lol

bowenyang007 (approved):

aptos-node/src/lib.rs
```diff
@@ -253,46 +255,22 @@ where
     let aptos_root_key_path = test_dir.join("mint.key");
 
     // If there's already a config, use it. Otherwise create a test one.
-    let config = if validator_config_path.exists() {
+    let config = if let Some(config) = config {
```
JoshLind, line 258: nit: This looks odd. Why not move it out to the call site? 😄

banool: Because this function does a bunch of magic overrides on the config for local testing. A larger refactor is probably due, eg separating all that into its own function.

PR #10096, opened 2023-09-18
Worklow to disable aggregator snapshots in v1.7

igor-aptos:


msmouse (approved):

aptos-move/aptos-release-builder/data/release.yaml
```diff
@@ -2,86 +2,13 @@
 remote_endpoint: ~
 name: "v1.7"
 proposals:
-  - name: enable_module_event
+  - name: diable_aggregator_snapshots
```
movekevin, line 5: disable

msmouse (changes requested):
won't land

PR #10098, opened 2023-09-18
[replay-verify] increase timeout for mainnet replay

rahxephon89:
Increase timeout for mainnet replay

### DIFF
--- a/.github/workflows/replay-verify.yaml
+++ b/.github/workflows/replay-verify.yaml
@@ -83,7 +83,7 @@ jobs:
       BACKUP_CONFIG_TEMPLATE_PATH: terraform/helm/fullnode/files/backup/s3-public.yaml
       # workflow config
       RUNS_ON: "high-perf-docker-with-local-ssd"
-      TIMEOUT_MINUTES: 240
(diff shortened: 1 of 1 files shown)


msmouse (approved):

.github/workflows/replay-verify.yaml
```diff
@@ -83,7 +83,7 @@ jobs:
       BACKUP_CONFIG_TEMPLATE_PATH: terraform/helm/fullnode/files/backup/s3-public.yaml
       # workflow config
       RUNS_ON: "high-perf-docker-with-local-ssd"
-      TIMEOUT_MINUTES: 240
+      TIMEOUT_MINUTES: 300
```
perryjrandall, line 86: Suggest we bump this a bit higher in case we creep up again before we lower this

perryjrandall (approved):
sheepit

.github/workflows/replay-verify.yaml
rahxephon89: Added one hour so that we can verify the theory more quickly

PR #10101, opened 2023-09-18
[CLI] Add txn stream to local testnet

banool:
### Description
This PR adds a txn stream to the local testnet. It also makes a variety of other changes to improve the local testnet:
- Enable the faucet by default.
- Use tokio tasks rather than raw futures.
- Add a more systematic way to run one service after another.
- Add a more systematic way to do healthchecking after startup to ensure that everything is running.
- Write logs from components that use tracing into specific directories in the testnet directory.
- Add a new ready check server that exposes a unified endpoint for checking overall readiness of all local testnet components.

### Test Plan
Build a tools image locally:
```
docker/builder/docker-bake-rust-all.sh tools
```
I used the remote docker builder for this. I tagged the image as devnet so I could use it as the base network for the CLI E2E tests.

Run the CLI E2E tests where we use the local tools image:
```
cd crates/aptos/e2e
poetry install
poetry run python main.py -d --base-network devnet --test-cli-path ~/a/core/target/debug/aptos --image-repo-with-project aptos-core
```
```
023-09-20 19:58:29,559 - INFO - All tests passed!
```

See that the txn stream comes up:
```
cargo run -p aptos -- node run-local-testnet --force-restart --assume-yes
```
```
        Log file: "/Users/dport/.aptos/testnet/validator.log"
        Test dir: "/Users/dport/.aptos/testnet"
        Aptos root key path: "/Users/dport/.aptos/testnet/mint.key"
        Waypoint: 0:f87e3e33347cfde3b61de0b531e6eeb3d5bbdb3204e21b87d3c4b58fe3c7e105
        ChainId: 4
        Indexer gRPC node stream endpoint: 0.0.0.0:50051

Readiness endpoint: http://0.0.0.0:8090/

Node API is starting, please wait...
Faucet is starting, please wait...
Transaction stream is starting, please wait...

Node API is running. Endpoint: http://0.0.0.0:8080/
Transaction stream is running. Endpoint: http://0.0.0.0:50051/
Faucet is running. Endpoint: http://0.0.0.0:8081/

All services are running, you can now use the local testnet!
```
```
grpcurl -plaintext -d '{ "starting_version": 0 }' 127.0.0.1:50051 aptos.indexer.v1.RawData/GetTransactions
```


crates/aptos/src/node/local_testnet/logging.rs
```diff
@@ -0,0 +1,102 @@
+// Copyright © Aptos Foundation
```
banool, line 1: All of this isn't strictly necessary, we could just write all our tracing output to a single file, but separating it by service makes it easier to understand.

alnoki:
@banool will this change CLI flags, e.g. `--with-local-testnet` since it is now default?

I believe that would bump to v3.0.0 if so

banool:
So it doesn't actually introduce backwards incompatible changes because I keep the `--with-faucet` flag, precisely for this reason. It doesn't do anything, but using it won't cause anything to break.

0xmaayan:
```
Add a more systematic way to do healthchecking after startup to ensure that everything is running.
Add a new ready check server that exposes a unified endpoint for checking overall readiness of all local testnet components.
```
whats the difference between these two?

crates/aptos/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ All notable changes to the Aptos CLI will be captured in this file. This project
 - Added an option `--print-metadata` to the command `aptos move download` to print out the metadata of the package to be downloaded.
   - Example: `aptos move download  --account 0x1 --package AptosFramework --url https://mainnet.aptoslabs.com/v1 --print-metadata`
 
+### Updated
+- The `--with-faucet` flag has been removed from `aptos node run-local-testnet`, we now run a faucet by default. To disable the faucet use the `--no-faucet` flag.
```
0xmaayan, line 11: what happens if we still use `--with-faucet`? will it explode?

banool:
> whats the difference between these two?

The first one does checks at startup and if those checks fail, the CLI exits. The second provides a server that the user can query to see if the servers are all set up. They both use the same functions under the hood.

crates/aptos/CHANGELOG.md
banool: Nope, it's actually still there, I just hid it. Otherwise yeah things would blow up.

ecosystem/indexer-grpc/indexer-grpc-fullnode/src/runtime.rs
```diff
@@ -58,9 +62,25 @@ pub fn bootstrap(
...
+            .build()
+            .expect("Failed to build reflection service");
+
+        let reflection_service_clone = reflection_service.clone();
+
+        let tonic_server = Server::builder()
             .http2_keepalive_interval(Some(std::time::Duration::from_secs(60)))
-            .http2_keepalive_timeout(Some(std::time::Duration::from_secs(5)));
+            .http2_keepalive_timeout(Some(std::time::Duration::from_secs(5)))
+            .add_service(reflection_service_clone);
```
bowenyang007, line 83: I'm confused lol, why are we need this now and how come we didn't need it before?

bowenyang007 (approved):

ecosystem/indexer-grpc/indexer-grpc-fullnode/src/runtime.rs
banool: This is the new service that Larry just added recently so we didn't really have anything before. Without this you can't use grpcurl without "importing" all the proto files. Just a nice UX win, we have this on some of our other grpc servers.

0xmaayan (approved):

larry-aptos (approved):

larry-aptos (approved):

crates/aptos/src/node/local_testnet/mod.rs
```diff
@@ -0,0 +1,426 @@
...
+    /// Running a faucet alongside the node allows you to create and fund accounts
+    /// for testing.
+    #[clap(long)]
+    no_faucet: bool,
+
+    /// This does nothing, we already run a faucet by default. We only keep this here
+    /// for backwards compatibility with tests. We will remove this once the commit
+    /// that added --no-faucet makes its way to the testnet branch.
+    #[clap(long, hide = true)]
+    with_faucet: bool,
```
gregnazario, line 104: We will need to broadly share this before you fully remove this flag, so I suspect we will need this to be a while before we remove it

crates/aptos-faucet/core/src/server/run.rs
```diff
@@ -212,11 +211,6 @@ impl RunConfig {
             }));
         }
 
-        println!(
-            "Faucet is running. Faucet endpoint: http://{}:{}",
-            self.server_config.listen_address, self.server_config.listen_port
-        );
```
gregnazario, line 218: So, this no longer prints out for if you're running a faucet on your own?

crates/aptos/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ All notable changes to the Aptos CLI will be captured in this file. This project
...
+- When using `aptos node run-local-testnet` we now expose a transaction stream. Learn more about the transaction stream service here: https://aptos.dev/indexer/txn-stream/. Opt out of this with `--no-txn-stream`.
```
gregnazario, line 12: Please note that this is a breaking change, and note the new default ports that will be also used.

Say if someone was using this for CI, and then it starts failing because there's an extra port running.

crates/aptos/src/node/local_testnet/logging.rs
gregnazario: Why not use the already existing logging framework?

crates/aptos/src/node/local_testnet/mod.rs
```diff
@@ -0,0 +1,426 @@
...
+    #[clap(long)]
+    do_not_delegate: bool,
+
+    /// Do not run a transaction stream service alongside the node.
+    ///
+    /// Note: In reality this is not the same as running a Transaction Stream Service,
+    /// it is just using the stream from the node, but in practice this distinction
+    /// shouldn't matter.
+    #[clap(long)]
+    no_txn_stream: bool,
```
gregnazario, line 116: How much overhead does this create?

Mostly concerned since we've had to tune the local testnet several times, it can be a lot of overhead on local machines.

crates/aptos/src/node/local_testnet/ready_server.rs
```diff
@@ -0,0 +1,73 @@
...
+    EndpointExt, IntoResponse, Route, Server,
+};
+use serde::Serialize;
+use std::net::{Ipv4Addr, SocketAddrV4};
+
+#[derive(Debug, Clone, Parser)]
+pub struct ReadyServerConfig {
+    #[clap(long, default_value_t = 8090)]
+    pub ready_server_listen_port: u16,
+}
```
gregnazario, line 22: So, there's now yet another service just to check if the other services are running?

Could we shove this into an existing port?

So, now it looks like the local testnet node opens probably like 10 ports, between the node, the faucet, the grpc endpoint, the API, and the ready server.

crates/aptos/src/node/local_testnet/ready_server.rs
```diff
@@ -0,0 +1,73 @@
...
+    web::{Data, Json},
+    EndpointExt, IntoResponse, Route, Server,
+};
+use serde::Serialize;
+use std::net::{Ipv4Addr, SocketAddrV4};
+
+#[derive(Debug, Clone, Parser)]
+pub struct ReadyServerConfig {
+    #[clap(long, default_value_t = 8090)]
+    pub ready_server_listen_port: u16,
```
gregnazario, line 21: This needs a doc comment about what the ready server is

crates/aptos-faucet/core/src/server/run.rs
banool: It does, I just moved this into the healthchecking logic of the CLI.

crates/aptos/CHANGELOG.md
banool: Good point. Odds are low this is breaking for anyone but since it could be, I'll mention it.

crates/aptos/src/node/local_testnet/logging.rs
banool: The txn stream services don't use the existing logger bc we didn't want to bring in all those deps, so it uses tracing instead.

Besides that, so do you know if aptos_logger supports writing to different files on a per Tokio worker basis?

crates/aptos/src/node/local_testnet/mod.rs
banool: I ran some tests and I found the CPU overhead to be negligible. If you don't connect to the stream then this doesn't really do anything. I think maybe it enables some extra tracking in storage for tables (which will be moved out as some point).

crates/aptos/src/node/local_testnet/ready_server.rs
banool: Why shove this into an existing service when it doesn't belong there? I don't think opening a bunch of ports is a big deal.

PR #10104, opened 2023-09-18
[replay-verify] Fix typo 

rahxephon89:
### Test Plan

To test this, goto https://github.com/aptos-labs/aptos-core/actions/workflows/replay-verify.yaml and trigger the workflow from this branch with `mainnet`.


perryjrandall (approved):
In order to test this

Go go https://github.com/aptos-labs/aptos-core/actions/workflows/replay-verify.yaml

Trigger workflow from this branch with value mainnet

Please update your test plan so that future people can also test / modify!

msmouse (approved):

PR #10106, opened 2023-09-18
[dag] state sync refactor and notifier reinit

ibalajiarun:
### Description

This PR splits the notifier into OrderedNotifier and ProofNotifier. 

- Cleans up some code to make sure ordered notifier adapter is re-instantiated after each state sync, so that the parent block info is accurate. To do so, it moves the epoch change notification logic from state sync manager to state sync notifier.
- Also, cleans up some StateSyncStatus logic and makes sure that DAG doesn't reinit when an epoch change is required.
- NetworkSender implements ProofNotifier. This makes a clean abstraction and keeps all epoch manager/buffer manager related components (including messages) outside of the DAG.

### DIFF
--- a/consensus/src/dag/bootstrap.rs
+++ b/consensus/src/dag/bootstrap.rs
@@ -4,7 +4,7 @@ use super::{
     adapter::Notifier,
     anchor_election::RoundRobinAnchorElection,
     dag_driver::DagDriver,
-    dag_fetcher::{DagFetcherService, FetchRequestHandler},
+    dag_fetcher::{DagFetcher, DagFetcherService, FetchRequestHandler},
     dag_handler::NetworkHandler,
     dag_network::TDAGNetworkSender,
     dag_state_sync::{DagStateSynchronizer, StateSyncTrigger, DAG_WINDOW},
@@ -15,7 +15,7 @@ use super::{
     types::{CertifiedNodeMessage, DAGMessage},
 };
 use crate::{
-    dag::{adapter::NotifierAdapter, dag_fetcher::DagFetcher},
+    dag::{adapter::NotifierAdapter, dag_state_sync::StateSyncStatus},
     experimental::buffer_manager::OrderedBlocks,
     network::IncomingDAGRequest,
     state_replication::{PayloadClient, StateComputer},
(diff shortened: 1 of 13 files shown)


ibalajiarun:
* main
  * **PR #10106** <a href="https://app.graphite.dev/github/pr/aptos-labs/aptos-core/10106" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/aptos-labs/aptos-core/10106?utm_source=stack-comment).

consensus/src/dag/bootstrap.rs
```diff
@@ -210,15 +217,27 @@ impl DagBootstrapper {
...
-                        error!(error = ?e, "unable to sync");
+                            if let Err(e) = sync_manager.sync_dag_to(&certified_node_msg, dag_fetcher, dag_store.clone()).await {
+                                error!(error = ?e, "unable to sync");
+                            }
+                        },
+                        StateSyncStatus::EpochEnds => {
+                            // Wait for epoch manager to signal shutdown
+                            _ = shutdown_rx.await;
+                            return;
+                        },
```
ibalajiarun, line 236: Not sure this is correct, because this peer could have received a old ledger info and it may already be in the new epoch.

zekun000 (approved):

consensus/src/dag/bootstrap.rs
zekun000: the invariant we should hold is that the dag instance only receives message from the same epoch. if it's already in the new epoch, it shouldn't receive message from previous epoch.

zekun000: sending multiple/stale epoch change proof is okay I think, epoch manager should handle that

sasha8 (approved):

consensus/src/dag/bootstrap.rs
ibalajiarun: We don't check the ledger info is from the same epoch actually. I need to fix that.

ibalajiarun: I added the check to make sure its same epoch.

PR #10107, opened 2023-09-19
module metadata cache

msmouse:
### Description

3% gain for this comand (32k TPS -> 33k TPS)
```
cargo run -p aptos-executor-benchmark --profile performance -- --block-size 50000  --split-ledger-db  --skip-index-and-usage --use-sharded-state-merkle-db --execution-threads 48 --connected-tx-grps 5000 --shuffle-connected-txns --generate-then-execute run-executor  --data-dir /home/msmouse/work/data/temp/$DB --checkpoint-dir /home/msmouse/work/data/temp/ck --blocks 50 --main-signer-accounts $SIGNERS
```

### DIFF
--- a/aptos-move/framework/src/module_metadata.rs
+++ b/aptos-move/framework/src/module_metadata.rs
@@ -23,7 +23,7 @@ use move_core_types::{
 };
 use move_vm_runtime::move_vm::MoveVM;
 use serde::{Deserialize, Serialize};
-use std::{collections::BTreeMap, env};
+use std::{collections::BTreeMap, env, sync::Arc};
 use thiserror::Error;
 
 /// The minimal file format version from which the V1 metadata is supported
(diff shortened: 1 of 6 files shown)


zekun000 (approved):

grao1991 (approved):

aptos-move/framework/src/module_metadata.rs
```diff
@@ -148,31 +149,64 @@ impl KnownAttribute {
     }
 }
 
+const METADATA_CACHE_SIZE: usize = 1024;
```
sitalkedia, line 152: How did you come up with this magic number?

sitalkedia (approved):

aptos-move/framework/src/module_metadata.rs
msmouse: the epilogue pulls in a pretty large chunk of the framework, and it's 70-ish modules. 1024 would be plenty of modules for a good hit rate.

aptos-move/framework/src/module_metadata.rs
```diff
@@ -148,31 +149,64 @@ impl KnownAttribute {
...
+
+thread_local! {
+    static V1_METADATA_CACHE: RefCell<LruCache<Vec<u8>, Option<Arc<RuntimeModuleMetadataV1>>>> = RefCell::new(LruCache::new(METADATA_CACHE_SIZE));
```
wrwg, line 155: As discussed on #9320, this is likely not ideal (since the key can be kb-ish), but better than we have right now! Thanks for doing this as a temporary fix.

wrwg (approved):

aptos-move/framework/src/module_metadata.rs
msmouse: Yeah it would be nice if modules can live across blocks and carry the metadata themselves.

PR #10110, opened 2023-09-19
[SDK V2] Implement memoize and memoizeAsync high order functions

0xmaayan:
### Description
Implementation of a `memoize` and `memoizeAsync` high order functions that later can be used to cache function response.

- `memoize` - a high order function to memoize a pure function response (for example, class methods)
```
authKey(): HexString {
 const memoizedAuthKey = memoize(() => {
     const pubKey = new Ed25519PublicKey(this.signingKey.publicKey);
     const authKey = AuthenticationKey.fromEd25519PublicKey(pubKey);
     return authKey.derivedAddress();
    }, `authKey`);
  return memoizedAuthKey();
}
```
- `memoizeAsync` - a high order function to memoize an async function response (for example, api calls)
```
export async function getModules(args: {
  aptosConfig: AptosConfig;
  accountAddress: HexInput;
}): Promise<MoveModuleBytecode[]> {
  const memoizedGetModules = memoizeAsync(
    async (args: {
      aptosConfig: AptosConfig;
      accountAddress: HexInput;
      options?: PaginationArgs & LedgerVersion;
    }): Promise<MoveModuleBytecode[]> => {
      const { aptosConfig, accountAddress, options } = args;
      const data = await paginateWithCursor<{}, MoveModuleBytecode[]>({
        url: aptosConfig.getRequestUrl(AptosApiType.FULLNODE),
        endpoint: `accounts/${AccountAddress.fromHexInput({ input: accountAddress }).toString()}/modules`,
        params: { ledger_version: options?.ledgerVersion, start: options?.start, limit: options?.limit ?? 1000 },
        originMethod: "getModules",
        overrides: { ...aptosConfig.clientConfig },
      });
      return data;
    },
    `getAccountModules_${args.accountAddress}`,
    600000, // 10 Minutes
  );

  return memoizedGetModules(args);
}
```

Note: current sdk version uses decorators. Decided to use high order function for some reasons:
1. Decorators is not that common in the JS world but high order functions are everywhere (filter, map, reduce, etc...).
2. Decorators can only be applied on class methods and not pure functions which doesn't conform with our new sdk design.
3. For now the memoize functions are pretty simple, can extend them as we need and think is right.

### Test Plan
```
npx jest -- tests/unit/memoize.test.ts              
 PASS  tests/unit/memoize.test.ts
  memoize
    memoizeAsync
      ✓ it does not execute function again before TTL has passed (103 ms)
      ✓ it executes function again after TTL has passed (204 ms)
      ✓ it does not execute function again when TTL is not provided (105 ms)
      ✓ it returns the expected response based on the provided cache key (206 ms)
    memoize
      ✓ it does not execute function again before TTL has passed (1 ms)
      ✓ it executes function again after TTL has passed
      ✓ it does not execute function again when TTL is not provided (1 ms)
      ✓ it returns the expected response based on the provided cache key
```

### DIFF
--- a/ecosystem/typescript/sdk_v2/src/utils/memoize.ts
+++ b/ecosystem/typescript/sdk_v2/src/utils/memoize.ts
@@ -1,0 +1,64 @@
+/**
+ * The global cache Map
+ */
+const cache = new Map<string, { value: any; timestamp: number }>();
+
+/**
+ * A memoize high order function to cache async function response
+ *
+ * @param func An async function to cache the result of
+ * @param key The provided cache key
+ * @param ttlMs time-to-live in milliseconds for cached data
+ * @returns
+ */
+export function memoizeAsync<T>(
+  func: (...args: any[]) => Promise<T>,
+  key: string,
+  ttlMs?: number,
+): (...args: any[]) => Promise<T> {
+  return async (...args: any[]) => {
+    // Check if the cached result exists and is within TTL
+    if (cache.has(key)) {
+      const { value, timestamp } = cache.get(key)!;
+      if (ttlMs === undefined || Date.now() - timestamp <= ttlMs) {
+        return value;
+      }
+    }
+
+    // If not cached or TTL expired, compute the result
+    const result = await func(...args);
+
+    // Cache the result with a timestamp
+    cache.set(key, { value: result, timestamp: Date.now() });
+
+    return result;
+  };
+}
+
+/**
+ * A memoize high order function to cache function response
+ *
+ * @param func A function to cache the result of
+ * @param key The provided cache key
+ * @param ttlMs time-to-live in milliseconds for cached data
+ * @returns
+ */
+export function memoize<T>(func: (...args: any[]) => T, key: string, ttlMs?: number): (...args: any[]) => T {
+  return (...args: any[]) => {
+    // Check if the cached result exists and is within TTL
+    if (cache.has(key)) {
+      const { value, timestamp } = cache.get(key)!;
+      if (ttlMs === undefined || Date.now() - timestamp <= ttlMs) {
+        return value;
+      }
+    }
+
+    // If not cached or TTL expired, compute the result
+    const result = func(...args);
+
+    // Cache the result with a timestamp
+    cache.set(key, { value: result, timestamp: Date.now() });
+
+    return result;
+  };
+}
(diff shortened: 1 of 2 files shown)


ecosystem/typescript/sdk_v2/src/utils/memoize.ts
```diff
@@ -0,0 +1,64 @@
...
+ * @returns
+ */
+export function memoizeAsync<T>(
+  func: (...args: any[]) => Promise<T>,
+  key: string,
+  ttlMs?: number,
+): (...args: any[]) => Promise<T> {
+  return async (...args: any[]) => {
+    // Check if the cached result exists and is within TTL
+    if (cache.has(key)) {
```
0xjinn, line 21: Noob q - Is this global cache for the whole sdk? 

If I have two functions
```
function getA(value: string): string
function getB(value: string): string
```

Is it possible there might be collision if I am passing the same value e.g `test` to both function that are cached?

0xjinn:
May just me (Who came from different language like php/java), but I like decorator haha

ecosystem/typescript/sdk_v2/src/utils/memoize.ts
0xmaayan: well yes, but it shouldn't happen as we decide what the cache key will be, i.e `getAccountModules_${args.accountAddress}`. This is different than what it is today as the decorator assumes the cache key should be the first function argument.

0xmaayan:
> May just me (Who came from different language like php/java), but I like decorator haha

These are the reason why decided/needed to go with high order functions
```
Decorators is not that common in the JS world but high order functions are everywhere (filter, map, reduce, etc...).
Decorators can only be applied on class methods and not pure functions which doesn't conform with our new sdk design.
For now the memoize functions are pretty simple, can extend them as we need and think is right.
```

ecosystem/typescript/sdk_v2/src/utils/memoize.ts
```diff
@@ -0,0 +1,64 @@
+/**
+ * The global cache Map
+ */
+const cache = new Map<string, { value: any; timestamp: number }>();
```
gregnazario, line 4: This could grow very large right?

We may want to think about things like cache configuration management by the user as well as if there is a way to turn it off.

gregnazario (approved):

ecosystem/typescript/sdk_v2/src/utils/memoize.ts
0xmaayan: agree. we can introduce `clear` method (clear all cache / clear by key / etc).

banool (approved):

PR #10117, opened 2023-09-19
[Rust] Remove deprecated rust-toolchain file.

JoshLind:
### Description
This PR replaces the `rust-toolchain` file with `rust-toolchain.toml`. `rust-toolchain` has been [deprecated](https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file) and replaced with the `TOML` equivalent. This is also useful for dependabot, which cannot handle the old format.

Nothing should change in this PR -- we should still be using the same rust version 😄 

### Test Plan
Existing test infrastructure.


ibalajiarun (approved):

bchocho (approved):

gedigi (approved):

rust-toolchain.toml
```diff
@@ -0,0 +1,3 @@
+[toolchain]
+channel = "1.71.1"
+components = [ "cargo", "clippy", "rustc", "rust-docs", "rust-std" ]
```
msmouse, line 3: no rustfmt?

JoshLind: No, because we rely on nightly for `rustfmt` 😄

JoshLind: Let me add a comment about that...

PR #10120, opened 2023-09-19
[draft][gas] introduce dummy gas parameters

vgao1996:
This introduces dummy gas parameters that can be used to model abstract costs of the test-only natives, allowing them to be used in the gas calibration tool.

### DIFF
--- a/aptos-move/framework/src/natives/cryptography/bls12381.rs
+++ b/aptos-move/framework/src/natives/cryptography/bls12381.rs
@@ -609,10 +609,12 @@ pub fn native_bls12381_verify_signature_share(
 
 #[cfg(feature = "testing")]
 pub fn native_generate_keys(
-    _context: &mut SafeNativeContext,
+    context: &mut SafeNativeContext,
     _ty_args: Vec<Type>,
     mut _arguments: VecDeque<Value>,
 ) -> SafeNativeResult<SmallVec<[Value; 1]>> {
+    context.charge(BLS12381_TEST_ONLY_GENERATE_KEYS)?;
+
     let key_pair = KeyPair::<PrivateKey, PublicKey>::generate(&mut OsRng);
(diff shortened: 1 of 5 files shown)


alinush (approved):
Interesting. So this sets up things so that the test-only natives charge dummy gas parameters and the gas calibration tool can account for that properly (somehow)?

vgao1996:
Precisely. The calibration tool will simply treat the dummy parameters the same way as the regular ones and try to find suitable values to them, which we can simply throw away.

aptos-move/framework/src/natives/cryptography/bls12381.rs
```diff
@@ -622,23 +624,31 @@ pub fn native_generate_keys(
...
     _ty_args: Vec<Type>,
     mut arguments: VecDeque<Value>,
 ) -> SafeNativeResult<SmallVec<[Value; 1]>> {
     let msg = safely_pop_arg!(arguments, Vec<u8>);
     let sk_bytes = safely_pop_arg!(arguments, Vec<u8>);
+
+    context.charge(
+        BLS12381_TEST_ONLY_SIGN_BASE
+            + BLS12381_TEST_ONLY_SIGN_PER_BYTE_MSG * NumBytes::new(msg.len() as u64),
+    )?;
```
vgao1996, line 637: @alinush does this formula look correct to you? 

Later we'll need to do the same thing for other test only natives.

alinush: Yes, that looks good. The message is first hashed (linear gas formula) and then some costant time operations are performed on that hash.

alinush: For `t`-out-of-`n` MultiEd25519 though... there are `t` signatures being computed though. So you will need to account for that.

PR #10007, opened 2023-09-11
dag epoch manager integration wip

ibalajiarun:

PR #10122, opened 2023-09-19
[dag] support for highest committed round and latest ledger info on bootstrap

ibalajiarun:
### Description

This PR changes bootstrapper so that the correct latest ledger info is passed to all the DAG components. If the ledger info in storage ends the epoch, then a genesis ledger info is created and given to the different DAG components. Similarly, for the parent block info for the adapter. 

Also, extracts the highest committed round from dag store and isolates it with Arc<RwLock<>>

### DIFF
--- a/consensus/src/dag/adapter.rs
+++ b/consensus/src/dag/adapter.rs
@@ -15,6 +15,7 @@ use aptos_consensus_types::{
     block::Block,
     common::{Author, Payload, Round},
     executed_block::ExecutedBlock,
+    quorum_cert::QuorumCert,
 };
 use aptos_crypto::HashValue;
 use aptos_executor_types::StateComputeResult;
@@ -24,6 +25,7 @@ use aptos_storage_interface::{DbReader, Order};
 use aptos_types::{
     account_config::{new_block_event_key, NewBlockEvent},
     aggregate_signature::AggregateSignature,
+    block_info::BlockInfo,
     epoch_change::EpochChangeProof,
     epoch_state::EpochState,
     ledger_info::{LedgerInfo, LedgerInfoWithSignatures},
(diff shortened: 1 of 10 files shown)


ibalajiarun:
* main
  * **PR #10106** <a href="https://app.graphite.dev/github/pr/aptos-labs/aptos-core/10106" target="_blank">Graphite</a> 
    * **PR #10122** <a href="https://app.graphite.dev/github/pr/aptos-labs/aptos-core/10122" target="_blank">Graphite</a>  👈
      * **PR #10007** <a href="https://app.graphite.dev/github/pr/aptos-labs/aptos-core/10007" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/aptos-labs/aptos-core/10122?utm_source=stack-comment).

consensus/src/dag/adapter.rs
```diff
@@ -47,41 +48,56 @@ pub trait ProofNotifier: Send + Sync {
...
+            ledger_info_from_storage,
+        )
+    }
+}
+
 pub struct OrderedNotifierAdapter {
     executor_channel: UnboundedSender<OrderedBlocks>,
     storage: Arc<dyn DAGStorage>,
     parent_block_info: Arc<RwLock<BlockInfo>>,
+    highest_committed_anchor_round: Arc<RwLock<Round>>,
```
zekun000, line 86: i think we can just use atomicu64 instead?

sasha8: But you still need the Arc because it is shared with state_sync_trigger, right?

sasha8 (approved):

consensus/src/dag/adapter.rs
ibalajiarun: Okay, we needed to cache the ledger info as well, so I introduced `LedgerInfoProvider` and a trait. New commit, lmk.

zekun000 (approved):

PR #10128, opened 2023-09-19
Implementing VersionedAggregators and AggregatorChange for aggregator V2

igor-aptos:

### DIFF
--- a/aptos-move/aptos-aggregator/src/aggregator_change_set.rs
+++ b/aptos-move/aptos-aggregator/src/aggregator_change_set.rs
@@ -2,331 +2,447 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use crate::{
-    bounded_math::{code_invariant_error, expect_ok, BoundedMath, SignedU128},
-    delta_math::{merge_data_and_delta, merge_two_deltas, DeltaHistory},
-    types::{AggregatorID, SnapshotToStringFormula},
+    bounded_math::code_invariant_error,
+    delta_change_set::DeltaOp,
+    types::{AggregatorID, AggregatorValue, SnapshotToStringFormula},
 };
 use move_binary_format::errors::PartialVMResult;
 
 #[derive(Clone, Debug, Eq, PartialEq)]
-pub enum AggregatorChange {
-    Data {
-        value: u128,
-    },
-    StringData {
-        value: Vec<u8>,
-    },
+pub enum AggregatorApplyChange {
     AggregatorDelta {
-        delta: SignedU128,
-        max_value: u128,
-        history: DeltaHistory,
+        delta: DeltaOp,
     },
-    /// Value of delta is:
-    /// formula(value of base_aggregator at the beginning of the transaction + delta)
+    /// Value is:
+    /// (value of base_aggregator at the beginning of the transaction + delta)
     SnapshotDelta {
-        delta: SignedU128,
         base_aggregator: AggregatorID,
+        delta: DeltaOp,
+    },
+    /// Value is:
+    /// formula(value of base_snapshot at the end of the transaction)
+    SnapshotDerived {
+        base_snapshot: AggregatorID,
         formula: SnapshotToStringFormula,
     },
 }
 
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub enum AggregatorChange {
+    Create(AggregatorValue),
+    Apply(AggregatorApplyChange),
+}
+
+// Contains information on top of which value should AggregatorApplyChange be applied.
+pub enum ApplyBase {
+    // Apply on top of the value end the end of the previous transaction
+    // (basically value at the start of the transaction.
+    // all changes in this transaction are captured in the Apply itself)
+    Previous(AggregatorID),
+    // Apply on top of the value at the end of the current transaction
+    // I.e. if this transaction changes the aggregator under wrapped ID,
+    // that apply needs to be applied first, before the current one is applied.
+    Current(AggregatorID),
+}
+
+impl AggregatorApplyChange {
+    pub fn get_apply_base_id_option(&self) -> Option<ApplyBase> {
+        use AggregatorApplyChange::*;
+
+        match self {
+            AggregatorDelta { .. } => None,
+            SnapshotDelta {
+                base_aggregator, ..
+            } => Some(ApplyBase::Previous(*base_aggregator)),
+            SnapshotDerived { base_snapshot, .. } => Some(ApplyBase::Current(*base_snapshot)),
+        }
+    }
+
+    pub fn get_apply_base_id(&self, self_id: &AggregatorID) -> ApplyBase {
+        self.get_apply_base_id_option()
+            .unwrap_or(ApplyBase::Previous(*self_id))
+    }
+
+    pub fn apply_to_base(&self, base_value: AggregatorValue) -> PartialVMResult<AggregatorValue> {
+        use AggregatorApplyChange::*;
+
+        Ok(match self {
+            AggregatorDelta { delta } => {
+                AggregatorValue::Aggregator(delta.apply_to(base_value.into_aggregator_value()?)?)
+            },
+            SnapshotDelta { delta, .. } => {
+                AggregatorValue::Snapshot(delta.apply_to(base_value.into_aggregator_value()?)?)
+            },
+            SnapshotDerived { formula, .. } => {
+                AggregatorValue::Derived(formula.apply(base_value.into_snapshot_value()?))
+            },
+        })
+    }
+}
+
 impl AggregatorChange {
-    /// Returns the result of applying the AggregatorState to the base value
-    /// Returns error if the aggregator behavior based on the
-    /// speculative_base_value isn't consistent with the provided base value.
-    pub fn apply_aggregator_change_to(&self, base: u128) -> PartialVMResult<u128> {
+    // When squashing a new change on top of the old one, sometimes we need to know the change
+    // from a different AggregatorID to be able to merge them together.
+    // In particular SnapshotDelta represents a change from the aggregator at the beginning of the transaction,
+    // and squashing changes where the aggregator will be at the beginning of the transaction.
+    // For example, let’s say we have two change sets that we need to squash:
+    // change1: agg1 -> Delta(+3)
+    // change2: agg1 -> Delta(+6), snap1 -> (base=agg1, Delta(+2))
+    // the correct squashing of snapshot depends on the change for the base aggregator. I.e. the correct output would be:
+    // agg1 -> Delta(+9), snap(base=agg1, Delta(+5))
+    pub fn get_merge_dependent_id(&self) -> Option<AggregatorID> {
+        use AggregatorApplyChange::*;
+        use AggregatorChange::*;
+
         match self {
-            AggregatorChange::Data { value } => Ok(*value),
-            AggregatorChange::StringData { .. } => Err(code_invariant_error(
-                "There should be no base value for StringData",
-            )),
-            AggregatorChange::AggregatorDelta {
-                delta,
-                history,
-                max_value,
-            } => merge_data_and_delta(base, delta, history, *max_value),
-            AggregatorChange::SnapshotDelta { delta, .. } => {
-                // History validation should already be done before this call.
-                expect_ok(BoundedMath::new(u128::MAX).unsigned_add_delta(base, delta))
-            },
+            // Only SnapshotDelta merging logic depends on current aggregator change
+            Apply(SnapshotDelta {
+                base_aggregator, ..
+            }) => Some(*base_aggregator),
+            Create(_) | Apply(AggregatorDelta { .. } | SnapshotDerived { .. }) => None,
         }
     }
 
-    /// Applies self on top of previous change, merging them together. Note
-    /// that the strict ordering here is crucial for catching overflows correctly.
-    pub fn merge_with_previous_aggregator_change(
-        &mut self,
-        previous_change: &AggregatorChange,
-    ) -> PartialVMResult<()> {
-        match self {
-            // If the current state is Data, then merging with previous state won't change anything.
-            AggregatorChange::Data { .. } | AggregatorChange::StringData { .. } => Ok(()),
-            AggregatorChange::AggregatorDelta {
-                delta,
-                history,
-                max_value,
-            } => match previous_change {
-                AggregatorChange::Data { value: prev_value } => {
-                    let new_data = merge_data_and_delta(*prev_value, delta, history, *max_value)?;
-                    *self = AggregatorChange::Data { value: new_data };
-                    Ok(())
-                },
-                AggregatorChange::StringData { .. } => Err(code_invariant_error(
-                    "Base value cannot be string for an aggregator",
-                )),
-                AggregatorChange::AggregatorDelta {
-                    delta: prev_delta,
-                    max_value: prev_max_value,
-                    history: prev_history,
-                } => {
-                    if max_value != prev_max_value {
-                        return Err(code_invariant_error(
-                            "Max value of aggregator cannot change",
-                        ));
-                    }
-                    let (new_delta, new_history) =
-                        merge_two_deltas(prev_delta, prev_history, delta, history, *max_value)?;
-                    *self = AggregatorChange::AggregatorDelta {
-                        delta: new_delta,
-                        history: new_history,
-                        max_value: *max_value,
-                    };
-                    Ok(())
-                },
-                AggregatorChange::SnapshotDelta { .. } => Err(code_invariant_error(
-                    "SnapshotDelta cannot come before AggregatorDelta",
-                )),
-            },
-            AggregatorChange::SnapshotDelta { .. } => {
-                unimplemented!()
-            },
-        }
-    }
-
-    /// Applies next aggregator change on top of self, merging two changes together. This is a reverse
-    /// of `merge_with_previous_aggregator_change`.
-    pub fn merge_with_next_aggregator_change(
-        &mut self,
+    /// Applies next AggregatorChange on top of the previous state.
+    /// prev_change is AggregatorChange for the same AggregatorID
+    /// prev_dependent_change is AggregatorChange for the get_merge_dependent_id()
+    pub fn merge_two_changes(
+        prev_change: Option<&AggregatorChange>,
+        prev_dependent_change: Option<&AggregatorChange>,
         next_change: &AggregatorChange,
-    ) -> PartialVMResult<()> {
-        // Now self follows the other delta.
-        let mut prev_change = next_change.clone();
-        std::mem::swap(self, &mut prev_change);
+    ) -> PartialVMResult<AggregatorChange> {
+        use AggregatorApplyChange::*;
+        use AggregatorChange::*;
+        use AggregatorValue::*;
 
-        // Perform the merge.
-        self.merge_with_previous_aggregator_change(&prev_change)?;
-        Ok(())
+        // There are only few valid cases for merging:
+        // - next_change being AggregatorDelta, and prev_change being Aggregator Create or Delta
+        // - next_change being SnapshotDelta, and prev_dependent_change being Aggregator Create or Delta
+        // everything else is invalid for various reasons
+        match (&prev_change, &prev_dependent_change, next_change) {
+            (None, None, v) => Ok(v.clone()),
+            (_ , _, Create(_)) => Err(code_invariant_error(
+                "Trying to merge Create with an older change. Create should always be the first change.",
+            )),
+
+            // Aggregators:
+            (Some(Create(Aggregator(prev_value))), None, Apply(AggregatorDelta { delta: next_delta })) => {
+                let new_data = next_delta.apply_to(*prev_value)?;
+                Ok(Create(Aggregator(new_data)))
+            },
+            (Some(Apply(AggregatorDelta { delta: prev_delta })), None, Apply(AggregatorDelta { delta: next_delta })) => {
+                let new_delta = DeltaOp::create_merged_delta(prev_delta, next_delta)?;
+                Ok(Apply(AggregatorDelta { delta: new_delta }))
+            },
+
+            // Snapshots:
+            (Some(Create(Snapshot(_) | Derived(_)) | Apply(SnapshotDelta {..} | SnapshotDerived { .. })), _, _) => Err(code_invariant_error(
+                "Snapshots are immutable, previous change cannot be any of the snapshots type",
+            )),
+            (_, Some(_), Apply(AggregatorDelta { .. } | SnapshotDerived { .. })) =>
+                unreachable!("Only SnapshotDelta should have merge dependent changes"),
+            (_, _, Apply(SnapshotDerived { .. })) => Err(code_invariant_error(
+                "Trying to merge SnapshotDerived with an older change. Snapshots are immutable, should only ever have one change.",
+            )),
+            (Some(_), _, Apply(SnapshotDelta { .. })) => Err(code_invariant_error(
+                "Trying to merge Snapshot (delta or derived) with an older change on the same ID. Snapshots are immutable, should only ever have one change - that creates them",
+            )),
+            (None, Some(Create(Aggregator(prev_value))), Apply(SnapshotDelta { delta: next_delta, .. })) => {
+                let new_data = next_delta.apply_to(*prev_value)?;
+                Ok(Create(Snapshot(new_data)))
+            }
+            (None, Some(Apply(AggregatorDelta { delta: prev_delta })), Apply(SnapshotDelta { delta: next_delta, base_aggregator })) => {
+                let new_delta = DeltaOp::create_merged_delta(prev_delta, next_delta)?;
+                Ok(Apply(SnapshotDelta { delta: new_delta, base_aggregator: *base_aggregator }))
+            }
+            (None, Some(Create(Snapshot(_) | Derived(_)) | Apply(SnapshotDelta {..} | SnapshotDerived { .. })), Apply(SnapshotDelta { .. })) => Err(code_invariant_error(
+                "Trying to merge SnapshotDelta with dependent change of wrong type",
+            )),
+        }
     }
 }
 
 #[cfg(test)]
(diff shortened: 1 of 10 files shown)


aptos-move/mvhashmap/src/versioned_aggregators.rs
```diff
@@ -33,11 +35,11 @@ enum AggregatorEntry {
     // of the same transaction, the delta may also be kept in the entry. This is useful
     // if speculative execution aborts and the entry is marked as estimate, as the delta
     // may be used to avoid waiting on the Estimate entry. More in comments below.
-    Value(u128, Option<DeltaOp>),
-    Delta(DeltaOp),
+    Value(AggregatorValue, Option<AggregatorApplyChange>),
+    // Applies the change on top of the previous entry.
```
georgemitenkov, line 39: Can be merged with the next sentence?

aptos-move/mvhashmap/src/versioned_aggregators.rs
```diff
@@ -33,11 +35,11 @@ enum AggregatorEntry {
...
     // Applies the delta on top of a 'snapshot' of an aggregator with a given ID (at
     // a corresponding, i.e. one less than where the given entry is stored, version).
-    Snapshot(AggregatorID, DeltaOp),
+    Apply(AggregatorApplyChange),
```
georgemitenkov, line 42: nit: Application?

aptos-move/mvhashmap/src/versioned_aggregators.rs
```diff
@@ -15,13 +19,11 @@ use std::{
...
 // operation to not wait for the corresponding dependency.
-#[derive(Clone, Copy, Debug, PartialEq)]
+#[derive(Clone, Debug, PartialEq)]
 enum EstimatedEntry {
     NoBypass,
-    // If applicable, can bypass the Estimate by considering a Delta instead.
-    DeltaBypass(DeltaOp),
-    // If applicable, can bypass the Estimate by considering a Snapshot instead.
-    SnapshotBypass(AggregatorID, DeltaOp),
+    // If applicable, can bypass the Estimate by considering a Delta Change instead.
```
georgemitenkov, line 25: `a change`?

aptos-move/mvhashmap/src/versioned_aggregators.rs
```diff
@@ -175,33 +177,33 @@ impl VersionedValue {
...
-    // Gate the estimate bypass logic by whether that functionality is enabled
-    // (read_estimate_deltas flag) for the aggregator.
-    fn applicable_bypass(&self, bypass: &EstimatedEntry) -> EstimatedEntry {
-        if self.read_estimate_deltas {
-            *bypass
-        } else {
-            EstimatedEntry::NoBypass
-        }
-    }
+    // // Gate the estimate bypass logic by whether that functionality is enabled
```
georgemitenkov, line 198: Why comment out?

aptos-move/mvhashmap/src/versioned_aggregators.rs
```diff
@@ -128,18 +130,18 @@ impl VersionedValue {
...
-                    (Estimate(DeltaBypass(delta_l)), Delta(delta_r) | Value(_, Some(delta_r))) => {
-                        *delta_l == *delta_r
+                    (Estimate(Bypass(apply_l)), Apply(apply_r) | Value(_, Some(apply_r))) => match (apply_l, apply_r) {
+                        (AggregatorDelta { .. }, AggregatorDelta { .. })
+                        | (SnapshotDelta { .. }, SnapshotDelta { .. })
+                        | (SnapshotDerived { .. }, SnapshotDerived { .. }) => {
+                            println!("Checking {:?} and {:?}, match: {}", apply_l, apply_r, *apply_l == *apply_r);
+                            *apply_l == *apply_r
+                        },
+                        _ => unreachable!(
```
georgemitenkov, line 140: Checking, now that we merged everything into the same enum, is this guaranteed? I think so because IDs are always re-generated for re-executions and are atomic

aptos-move/mvhashmap/src/versioned_aggregators.rs
```diff
@@ -211,35 +213,30 @@ impl VersionedValue {
...
     ) -> Result<u128, MVAggregatorsError> {
+        use AggregatorApplyChange::*;
         use AggregatorEntry::*;
         use EstimatedEntry::*;
 
         let mut accumulator = delta;
         while let Some((idx, entry)) = iter.next_back() {
-            let delta = match &**entry {
-                Value(v, _) => {
+            let delta = match (&**entry, self.read_estimate_deltas) {
```
georgemitenkov, line 222: I think having a match on self.read_estimate_deltas for bypass is cleaner?

aptos-move/mvhashmap/src/versioned_aggregators.rs
```diff
@@ -253,43 +250,64 @@ impl VersionedValue {
 
         // Finally, resolve if needed with the base value.
         self.base_value
+            .clone()
```
georgemitenkov, line 253: we can clone inside and_then so we do not clone snapshots and string bytes?

aptos-move/framework/src/natives/aggregator_natives/context.rs
```diff
@@ -63,13 +64,58 @@ impl<'a> NativeAggregatorContext<'a> {
...
+                } => Some(AggregatorChange::Create(AggregatorValue::Snapshot(value))),
+                AggregatorSnapshotState::Create {
+                    value: SnapshotValue::String(value),
+                } => Some(AggregatorChange::Create(AggregatorValue::Derived(value))),
+                AggregatorSnapshotState::Delta {
+                    base_aggregator,
+                    delta,
+                } => {
+                    let delta_op = aggregators.get(&AggregatorVersionedID::V2(base_aggregator)).map_or_else(
+                        || DeltaOp::new(delta, u128::MAX, DeltaHistory::new()),
```
georgemitenkov, line 88: why?

igor-aptos: In the context (AggregatorSnapshotState), we don't keep history/speculative value in the snapshot, only in the aggregator - so that it is not duplicated - and it is guaranteed to be validated fully (more context here - https://aptos-org.slack.com/archives/C057VPCD7QX/p1694555982793999 ) 

In the WriteSet, in order not to need to go to the aggregator to check history (but not use it's delta), we are extracting the DeltaOp and putting it there. 

DeltaOp::new(delta, u128::MAX, DeltaHistory::new()) is basically an empty delta op where checking history always succeeds.

aptos-move/mvhashmap/src/versioned_aggregators.rs
igor-aptos: Why? :) It's much longer, and it goes into AggregatorChange::Apply as well as AggregatorApplyChange. 

maybe if there is a better word instead

aptos-move/mvhashmap/src/versioned_aggregators.rs
igor-aptos: yes, IDs cannot ever be reused

aptos-move/mvhashmap/src/versioned_aggregators.rs
igor-aptos: can you explain?

aptos-move/mvhashmap/src/versioned_aggregators.rs
igor-aptos: because I needed it in the match together with types that are not estimates, and so couldn't use this function

aptos-move/mvhashmap/src/versioned_aggregators.rs
igor-aptos: those should never happen, but sure :)

aptos-move/aptos-aggregator/src/aggregator_change_set.rs
```diff
@@ -2,331 +2,388 @@
...
     },
 }
 
-impl AggregatorChange {
-    /// Returns the result of applying the AggregatorState to the base value
-    /// Returns error if the aggregator behavior based on the
-    /// speculative_base_value isn't consistent with the provided base value.
-    pub fn apply_aggregator_change_to(&self, base: u128) -> PartialVMResult<u128> {
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub enum AggregatorChange {
```
vusirikala, line 31: Trying to understand the design pattern here. Why is it better to divide AggregatorChange into AggregatorValue and AggregatorApplyChange, instead of having it as one piece as before?

aptos-move/aptos-aggregator/src/aggregator_change_set.rs
```diff
@@ -2,331 +2,388 @@
...
+    Create(AggregatorValue),
+    Apply(AggregatorApplyChange),
+}
+
+pub enum ApplyBase {
+    Previous(AggregatorID),
```
vusirikala, line 37: A comment on where/how `ApplyBase` is used would be helpful.

aptos-move/aptos-aggregator/src/aggregator_change_set.rs
```diff
@@ -2,331 +2,388 @@
...
+    Current(AggregatorID),
+}
+
+impl AggregatorApplyChange {
+    pub fn get_apply_base_id_option(&self) -> Option<ApplyBase> {
```
vusirikala, line 42: Kinda hard to understand what `get_apply_base_id_option` means.

aptos-move/aptos-aggregator/src/aggregator_change_set.rs
```diff
@@ -2,331 +2,388 @@
...
-    /// formula(value of base_aggregator at the beginning of the transaction + delta)
+    /// Value is:
+    /// (value of base_aggregator at the beginning of the transaction + delta)
     SnapshotDelta {
-        delta: SignedU128,
         base_aggregator: AggregatorID,
+        delta: DeltaOp,
+    },
+    /// Value is:
+    /// formula(value of base_snapshot at the end of the transaction)
```
vusirikala, line 23: Why is it the value of base_snapshot at the end of the transaction? It should be the value at the point where the snapshot is created right?

vusirikala: Ah... I now get it. This AggregatorID points to a snapshot which doesn't change after it is created. Confused because we are using AggregatorID for both aggregators and snapshots.

aptos-move/aptos-aggregator/src/aggregator_change_set.rs
igor-aptos: afterwards we handle Create and Apply very differently. And for each there is a lot of commonality for each.

For example, I can then add methods to AggregatorApplyChange - get_apply_base_id and apply_to_base  - and then treat all Apply's the same.  
I can utilize it directly in versioned_aggregators - and you can see how simpler things are there, i.e. we have:
Bypass(AggregatorApplyChange) (instead of 3 different bypasses)
similarly, we have only two:
    Value(AggregatorValue, Option<AggregatorApplyChange>),
    Apply(AggregatorApplyChange),
instead of 6 I would need to create otherwise.

aptos-move/aptos-aggregator/src/aggregator_change_set.rs
```diff
@@ -2,331 +2,388 @@
...
     }
 
-    /// Applies next aggregator change on top of self, merging two changes together. This is a reverse
-    /// of `merge_with_previous_aggregator_change`.
-    pub fn merge_with_next_aggregator_change(
-        &mut self,
+    /// Applies next aggregator change on top of self, merging two changes together.
+    pub fn merge_two_changes(
+        prev_change: Option<&AggregatorChange>,
+        prev_dependent_change: Option<&AggregatorChange>,
```
vusirikala, line 93: What is prev_change and prev_dependent_change? Having a comment would help.

igor-aptos: will add a comment. 

prev_change is change from the same AggregatorID as next_change
prev_dependent_change is change from the different AggregatorID - i.e. aggregator that snapshot depends on

aptos-move/aptos-aggregator/src/aggregator_change_set.rs
```diff
@@ -2,331 +2,388 @@
...
-                unimplemented!()
+            SnapshotDerived { formula, .. } => {
+                AggregatorValue::Derived(formula.apply(base_value.into_snapshot_value()?))
             },
+        })
+    }
+}
+
+impl AggregatorChange {
+    pub fn get_merge_dependent_id(&self) -> Option<AggregatorID> {
```
vusirikala, line 77: What does merge_dependent_id mean?

igor-aptos: for the above and this, more context here :

https://aptos-org.slack.com/archives/C057VPCD7QX/p1695160530418949

will add a comment in code for that

aptos-move/mvhashmap/src/versioned_aggregators.rs
```diff
@@ -89,9 +91,8 @@ impl VersionedValue {
         match self.versioned_map.entry(txn_idx) {
             Entry::Occupied(mut o) => {
                 let bypass = match &**o.get() {
-                    Value(_, maybe_delta) => maybe_delta.map_or(NoBypass, DeltaBypass),
-                    Delta(delta) => DeltaBypass(*delta),
-                    Snapshot(id, delta) => SnapshotBypass(*id, *delta),
+                    Value(_, maybe_apply) => maybe_apply.clone().map_or(NoBypass, Bypass),
```
vusirikala, line 94: What does this statement mean?

aptos-move/mvhashmap/src/versioned_aggregators.rs
```diff
@@ -15,13 +19,11 @@ use std::{
...
-#[derive(Clone, Copy, Debug, PartialEq)]
+#[derive(Clone, Debug, PartialEq)]
 enum EstimatedEntry {
     NoBypass,
-    // If applicable, can bypass the Estimate by considering a Delta instead.
-    DeltaBypass(DeltaOp),
-    // If applicable, can bypass the Estimate by considering a Snapshot instead.
-    SnapshotBypass(AggregatorID, DeltaOp),
+    // If applicable, can bypass the Estimate by considering a apply change instead.
+    Bypass(AggregatorApplyChange),
```
vusirikala, line 26: I don't have a strong opinion here. I am just wondering how much efficiency this is brining. 
Suppose `tx5` resulted in an error. Instead of using an Estimate to mark the aggregator in tx5, we are storing some change. Now, tx6 doesn't have to wait for tx5 and can use change. But later when tx5 re-runs, tx6 have re-run again as well. So, what's the benefit?

aptos-move/mvhashmap/src/versioned_aggregators.rs
```diff
@@ -175,34 +177,24 @@ impl VersionedValue {
...
+    ) -> Result<AggregatorValue, MVAggregatorsError> {
         use AggregatorEntry::*;
 
         self.versioned_map
             .range(0..next_idx_to_commit)
             .next_back()
-            .map_or(
-                self.base_value.ok_or(MVAggregatorsError::NotFound),
+            .map_or_else(
+                || self.base_value.clone().ok_or(MVAggregatorsError::NotFound),
```
vusirikala, line 187: Why clone?

aptos-move/mvhashmap/src/versioned_aggregators.rs
igor-aptos: map_or(default, func)

so it takes maybe_apply, if it is one, returns NoBypass, if it is |a| returns Bypass(a)

basically 

                    Value(_, maybe_apply) => maybe_apply.clone().map_or(NoBypass, |v| Bypass(v)),

aptos-move/mvhashmap/src/versioned_aggregators.rs
igor-aptos: if delta produced is the same, we don't need to re-execute.

So it's an optimization in case delta is always the same, i.e. independent from the rest of the computation.


as soon as a transaction produces a different delta, bypass is disabled for that aggregator - and we wait - as it is as you say - we will need to re-execute later.

aptos-move/mvhashmap/src/versioned_aggregators.rs
igor-aptos: I cannot consume self.base_value, I need to clone it to return it?

gelash (approved):
A few comments that's I'm curious to know the answer for, but def looks great!

aptos-move/mvhashmap/src/versioned_aggregators.rs
```diff
@@ -15,13 +19,11 @@ use std::{
 // When an AggregatorEntry (see below) is transformed to an Estimate, internally we store
 // a potential bypass (based on the previously stored entry), which may allow a read
 // operation to not wait for the corresponding dependency.
-#[derive(Clone, Copy, Debug, PartialEq)]
+#[derive(Clone, Debug, PartialEq)]
 enum EstimatedEntry {
```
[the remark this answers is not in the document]
igor-aptos: it's a bit more understandable in code this way, I am going to leave it as is in this PR, and might create and send separate PR just for that , to be able to easily compare side by side

aptos-move/aptos-aggregator/src/aggregator_change_set.rs
```diff
@@ -2,331 +2,388 @@
...
+        use AggregatorApplyChange::*;
+        use AggregatorChange::*;
+        use AggregatorValue::*;
+
+        // There are only few valid cases for merging:
+        // - next_change being AggregatorDelta, and prev_change being Aggregator Create or Delta
+        // - next_change being SnapshotDelta, and prev_dependent_change being Aggregator Create or Delta
+        // everything else is invalid for various reasons
+        match (&prev_change, &prev_dependent_change, next_change) {
+            (None, None, _) => unreachable!("We should be only merging, if there is something to merge"),
```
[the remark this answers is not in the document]
igor-aptos: very nice catch!

unit test caught this as well

aptos-move/mvhashmap/src/versioned_aggregators.rs
```diff
@@ -253,43 +240,64 @@ impl VersionedValue {
...
             .ok_or(MVAggregatorsError::NotFound)
-            .and_then(|base_value| {
-                accumulator
-                    .apply_to(base_value)
-                    .map_err(|_| MVAggregatorsError::DeltaApplicationFailure)
+            .and_then(|base_value| match base_value {
+                AggregatorValue::Aggregator(v) => accumulator
+                    .apply_to(*v)
+                    .map_err(|_| MVAggregatorsError::DeltaApplicationFailure),
+                _ => Err(MVAggregatorsError::DeltaApplicationFailure),
```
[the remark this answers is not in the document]
igor-aptos: isn't this possible if transaction that creates an aggregator gets invalidated?

aptos-move/mvhashmap/src/versioned_aggregators.rs
```diff
@@ -253,43 +240,64 @@ impl VersionedValue {
...
+                        Some(apply_base) => {
+                            let (base_id, end_index) = match apply_base {
+                                ApplyBase::Previous(id) => (id, *idx),
+                                ApplyBase::Current(id) => (id, *idx + 1),
+                            };
+
+                            Ok(VersionedRead::DependentApply(
+                                base_id,
+                                end_index,
+                                apply.clone(),
```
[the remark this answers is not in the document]
igor-aptos: something to think about ,agreed

igor-aptos:
addressed all comments and added bunch of test for corner cases

aptos-move/aptos-aggregator/src/aggregator_change_set.rs
vusirikala: nit: "value end the end of the"

aptos-move/aptos-aggregator/src/aggregator_change_set.rs
```diff
@@ -2,331 +2,447 @@
...
+        use AggregatorApplyChange::*;
+
+        Ok(match self {
+            AggregatorDelta { delta } => {
+                AggregatorValue::Aggregator(delta.apply_to(base_value.into_aggregator_value()?)?)
             },
-            AggregatorChange::SnapshotDelta { .. } => {
-                unimplemented!()
+            SnapshotDelta { delta, .. } => {
+                AggregatorValue::Snapshot(delta.apply_to(base_value.into_aggregator_value()?)?)
```
vusirikala, line 74: I'm not sure if this is a big deal. The semantics of SnapshotDelta and SnapshotDerived is that they can only be applied on base_aggregator. In this function, we are ignoring base_aggregator and applying SnapshotDelta and SnapshotDerived on any base value. Is that fine?

aptos-move/aptos-aggregator/src/aggregator_change_set.rs
```diff
@@ -2,331 +2,447 @@
...
             },
+            SnapshotDerived { formula, .. } => {
+                AggregatorValue::Derived(formula.apply(base_value.into_snapshot_value()?))
```
vusirikala, line 77: nit: `formula.apply_to` to be consistent with `delta.apply_to`?

aptos-move/aptos-aggregator/src/aggregator_change_set.rs
igor-aptos: these two functions go in pair - get_apply_base_id and apply_to_base

only caller knows how to fetch a value - it first needs to call get_apply_base_id to know which value to fetch, and then apply_to_base to apply it to the given value. it's on the caller to fetch correctly. 

I don't see a way to do this better/differently

aptos-move/aptos-aggregator/src/aggregator_change_set.rs
```diff
@@ -2,331 +2,447 @@
...
+    // When squashing a new change on top of the old one, sometimes we need to know the change
+    // from a different AggregatorID to be able to merge them together.
+    // In particular SnapshotDelta represents a change from the aggregator at the beginning of the transaction,
+    // and squashing changes where the aggregator will be at the beginning of the transaction.
+    // For example, let’s say we have two change sets that we need to squash:
+    // change1: agg1 -> Delta(+3)
+    // change2: agg1 -> Delta(+6), snap1 -> (base=agg1, Delta(+2))
+    // the correct squashing of snapshot depends on the change for the base aggregator. I.e. the correct output would be:
+    // agg1 -> Delta(+9), snap(base=agg1, Delta(+5))
+    pub fn get_merge_dependent_id(&self) -> Option<AggregatorID> {
```
vusirikala, line 93: Shouldn't this function be part of `impl AggregatorApplyChange` instead?

aptos-move/aptos-aggregator/src/aggregator_change_set.rs
```diff
@@ -2,331 +2,447 @@
...
-                        delta: new_delta,
-                        history: new_history,
-                        max_value: *max_value,
-                    };
-                    Ok(())
-                },
-                AggregatorChange::SnapshotDelta { .. } => Err(code_invariant_error(
-                    "SnapshotDelta cannot come before AggregatorDelta",
-                )),
+    pub fn get_apply_base_id(&self, self_id: &AggregatorID) -> ApplyBase {
```
vusirikala, line 61: It's just 1 line code. Does this have to be a separate function?

aptos-move/aptos-aggregator/src/aggregator_change_set.rs
vusirikala: In the above example, 
change1: agg1 -> Delta(+3)
change2: agg1 -> Delta(+6), snap1 -> (base=agg1, Delta(+2))

Here,    change1(agg2) is prev_dependent_change.   change2(snap1) is prev_change.  What is next_change then?
I didn't understand why "merge_two_changes" function takes 3 changes as input.

aptos-move/aptos-aggregator/src/aggregator_change_set.rs
```diff
@@ -2,331 +2,447 @@
...
+    ) -> PartialVMResult<AggregatorChange> {
+        use AggregatorApplyChange::*;
+        use AggregatorChange::*;
+        use AggregatorValue::*;
+
+        // There are only few valid cases for merging:
+        // - next_change being AggregatorDelta, and prev_change being Aggregator Create or Delta
+        // - next_change being SnapshotDelta, and prev_dependent_change being Aggregator Create or Delta
+        // everything else is invalid for various reasons
+        match (&prev_change, &prev_dependent_change, next_change) {
```
vusirikala, line 122: Merging 3 changes in "merge_two_changes" function is confusing.
And I didn't understand why we need to merge 3 changes at a time. In the above example,
change1: agg1 -> Delta(+3)
change2: agg1 -> Delta(+6), snap1 -> (base=agg1, Delta(+2))
We can merge [ change1(agg1), change2(agg1) ] and merge [ change1(agg1), change2(snap1) ] separately right?

aptos-move/aptos-aggregator/src/aggregator_change_set.rs
igor-aptos: it is one line, but contains logic - i.e. ApplyBase::Previous(*self_id) - that is relevant for this file. 

and this gets called elsewhere , so encapsulating it here makes sense to me

aptos-move/aptos-aggregator/src/aggregator_change_set.rs
igor-aptos: will do

aptos-move/aptos-aggregator/src/aggregator_change_set.rs
igor-aptos: we are merging AggregatorChange in change_set.rs. I don't want to have there a match on whether a change is a value or not. and it also corresponds to merge_two_changes function from AggregatorChange - so keeping it there makes sense logically

igor-aptos:
@vusirikala addressed all your comments, does it look good now?

PR #10139, opened 2023-09-20
[dag] add payload filter to dag driver

zekun000:


consensus/src/dag/dag_driver.rs
```diff
@@ -139,14 +139,31 @@ impl DagDriver {
 
     pub async fn enter_new_round(&mut self, new_round: Round, strong_links: Vec<NodeCertificate>) {
         debug!("entering new round {}", new_round);
-        // TODO: support pulling payload
+        let payload_filter = {
+            let dag_reader = self.dag.read();
+            let highest_commit_round = dag_reader.highest_committed_anchor_round();
```
ibalajiarun, line 144: we should go as far as highest_commit_round - DAG_WINDOW and select uncommitted payloads as well?

consensus/src/dag/dag_driver.rs
```diff
@@ -139,14 +139,31 @@ impl DagDriver {
...
+            if strong_links.is_empty() {
+                PayloadFilter::Empty
+            } else {
+                PayloadFilter::from(
+                    &dag_reader
+                        .reachable(
+                            strong_links.iter().map(|node| node.metadata()),
```
ibalajiarun, line 151: I think we need to turn off QS proof broadcast step, otherwise same round might have duplicates.

consensus/src/dag/dag_driver.rs
zekun000: good point, maybe we should just reach lowest round for now (with gc it should be approximate the highest_commit - window?)

consensus/src/dag/dag_driver.rs
zekun000: yeah we should make that configurable, but probably not in this pr

consensus/src/dag/dag_driver.rs
ibalajiarun: sounds good.

consensus/src/dag/dag_driver.rs
```diff
@@ -139,14 +140,31 @@ impl DagDriver {
...
+                            Some(highest_commit_round.saturating_sub(DAG_WINDOW as u64)),
```
sasha8, line 153: What if proofs in nodes outside of the window are not committed yet?

consensus/src/dag/dag_driver.rs
sasha8: Actually, it is a trade-off. Maybe we do want to help nodes that are outside of the anchor's causal history?

consensus/src/dag/dag_driver.rs
zekun000: they're not going to be committed ever by the rule right? (if you mean the nodes with round < highest_committed_round - WINDOW)

sasha8: Yes, what I mean is that QS can give it the DAG because they were not committed yet (decoupled execution).

zekun000: that's not possible right? the highest_commit round is what already goes to storage, not the highest ordered round

sasha8: I am confused. I think the window has to be according to the highest-ordered round. Otherwise, it is not deterministic.

sasha8: Sorry, I think you are good. I was confused.

zekun000: ah yes, we haven't done that for order rule, but here we're just being more relaxed? since commit round <= ordered round

sasha8: It is tricky. 
With the current rule, we might filter out proof that we need to resubmit (because their nodes are below the GC line).
But if we do "ordered round - window", we might include stuff that we are already excited but not yet committed.

zekun000: it's fine to filter out proof that need to resubmit since the window is moving, it's just delaying those proof for a little bit?

sasha8: What about recording?
Also, they can expire.

sasha8 (approved):

ibalajiarun (approved):

PR #10148, opened 2023-09-20
Port over `@aptos-labs/aptos-cli` node wrapper package code

0xmaayan:
### Description
Port over `@aptos-labs/aptos-cli` package core code from https://github.com/aptos-labs/cad-aptos-cli.
Also added CLI operations like start a local node and compile a move package - it is not final and can use some more work but first i'd like to move the core code from an outside repo into this repo. 
This is the first step to improve the process of running SDK examples and dev exp.

### DIFF
--- a/ecosystem/typescript/aptos-cli/bin/aptos
+++ b/ecosystem/typescript/aptos-cli/bin/aptos
@@ -1,0 +1,69 @@
+#!/usr/bin/env node
+
+const { spawn } = require("child_process");
+const fs = require("fs");
+const os = require("os");
+
+const PNAME = "aptos-cli";
+const VERSION = "2.0.2";
+
+// Get all arguments passed to the script
+const getOS = () => {
+  const platform = os.platform();
+  switch (platform) {
+    case "darwin":
+      return "MacOSX";
+    case "linux":
+      return "Ubuntu";
+    case "win32":
+      return "Windows";
+    default:
+      throw new Error(`Unsupported OS ${platform}`);
+  }
+};
+
+const main = async () => {
+  const os = getOS();
+  const binaryName = `${PNAME}-${VERSION}-${os}`;
+
+  // If binary does not exist, download it
+  if (!fs.existsSync(`${__dirname}/${binaryName}`)) {
+    await new Promise((resolve, reject) => {
+      const url = `https://github.com/aptos-labs/aptos-core/releases/download/${PNAME}-v${VERSION}/${PNAME}-${VERSION}-${os}-x86_64.zip`;
+      console.log("Downloading aptos CLI...");
+      let downloadProcess;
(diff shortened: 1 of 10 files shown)


ecosystem/typescript/aptos-cli/bin/aptos
```diff
@@ -0,0 +1,69 @@
+#!/usr/bin/env node
+
+const { spawn } = require("child_process");
+const fs = require("fs");
+const os = require("os");
+
+const PNAME = "aptos-cli";
+const VERSION = "2.0.2";
```
davidiw, line 8: feels like we could just host something on aptos-dev that says the latest version.

davidiw (commented):
Cool stuff :)

I would be curious if we want to support this script or to tell the user to just install the CLI into a global path, like the Python approach. How fragile is this script?

ecosystem/typescript/aptos-cli/bin/aptos
banool: If we wanted latest you can just use the GH API to look it up, though I think this is by design. This way we can verify each new CLI version works. You'd think it's not necessary but 2.0.3 doesn't work due to this OpenSSL stuff, so it's probably good to be careful.

0xmaayan:
> Cool stuff :)
> 
> I would be curious if we want to support this script or to tell the user to just install the CLI into a global path, like the Python approach. How fragile is this script?

create-apt-dapp uses this script with the support of downloading and installing the CLI for the user in the current development env. Imo, I look at this package as a node wrapper for the aptos cli and as a better dev exp for dapps/node builders that we include the cli in the project node_modules bin for them. Also, aptos ts sdk can easily add this package as a dependency and have the cli without the end user intervention. tbh, the web app/node builders community expecting for the tools to make their life easier and automated (and not to install things in their local path) like many other packages out there work (react, vue, tsup, jest, etc).

The script has been tested on MacOS and Windows machines and once the script installs the cli we can expect it to be available in the project node_modules (i.e can safely run `npx aptos`).

> feels like we could just host something on aptos-dev that says the latest version.

we currently use this fixed version because of some weird openSSL issues. I believe we have a [PR](https://github.com/aptos-labs/aptos-core/pull/9961) that should help with it.


With that being said, we can always support a local path argument and if not provided, we use this script.

ecosystem/typescript/aptos-cli/bin/aptos
0xmaayan: @banool does https://github.com/aptos-labs/aptos-core/pull/9961 should help with it?

banool: Yeah this particular issue is fixed now. My point is though that using the latest CLI version by default might be slightly risky business. We could take it for a spin though.

PR #10150, opened 2023-09-21
[Aptos Data Client] Small cleanups and refactors.

JoshLind:
### Description
This PR offers two small cleanups and refactors to the Aptos Data Client:
1. Replace `Arc<RwLock<T>>` with `Arc<ArcSwap<T>>` for the global summary cache. This should [reduce](https://docs.rs/arc-swap/latest/arc_swap/) lock contention.
2. Remove the global `RwLock` around `PeerStates` and use `DashMap` and `DashSet` internally, to reduce lock contention.

Nothing else should change 😄 

### Test Plan
Existing test infrastructure.

### DIFF
--- a/state-sync/aptos-data-client/src/client.rs
+++ b/state-sync/aptos-data-client/src/client.rs
@@ -46,9 +46,10 @@ use aptos_types::{
     state_store::state_value::StateValueChunkWithProof,
     transaction::{TransactionListWithProof, TransactionOutputListWithProof, Version},
 };
+use arc_swap::ArcSwap;
 use async_trait::async_trait;
 use rand::prelude::SliceRandom;
-use std::{fmt, sync::Arc, time::Duration};
+use std::{fmt, ops::Deref, sync::Arc, time::Duration};
 use tokio::runtime::Handle;
 
 // Useful constants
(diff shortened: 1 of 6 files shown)


state-sync/aptos-data-client/src/client.rs
```diff
@@ -100,17 +101,21 @@ impl AptosDataClient {
         storage_service_client: StorageServiceClient<NetworkClient<StorageServiceMessage>>,
         runtime: Option<Handle>,
     ) -> (Self, DataSummaryPoller) {
+        // Wrap the configs in an Arc (to be shared across components)
+        let base_config = Arc::new(base_config);
```
grao1991, line 105: why not &?

JoshLind: Aah, because the structs are cloned during program execution, so `Arc` makes the code simpler and avoids me having to add explicit lifetimes to them (e.g., `'static`).

grao1991: As a c++ engineer in general I dislike Arc  (std::shared_ptr in c++) :) 
It creates unnecessary overhead, unclear ownership, potentially unexpected memory consumption, and destruction at undesirable time.

grao1991 (approved):

PR #10157, opened 2023-09-21
[TSSDKv2][1/n] add Ed25519 classes

0xjinn:
### Description

This PR adds the following classes

1. PublicKey
2. PrivateKey
3. Signature

This version matches more closely to our original design from figma https://www.figma.com/file/ATd5LOKTY9JN7P85tYsDsk/SDK-V2?type=whiteboard&node-id=0-1&t=8JK0OM5tsix5DnVX-0

It also matches more closely with the `Python SDK` as well

Update:

Added asymmetric_crypto which contains Private and Public key, Signature base abstracted classes
Update test to include additional Serialize and Deserialize method
Make sure crypto concrete classes all extend from base asymmetric_crypto class
Implement method from base classes

TODO:
We will probably need to update the interface in Deserializable to implement static instead of instance method deserialize

### Test Plan

jest test


0xmaayan (commented):
Looks much better and cleaner!

ecosystem/typescript/sdk_v2/src/crypto/ed25519.ts
```diff
@@ -0,0 +1,180 @@
...
+    serializer.serializeBytes(this.key.toUint8Array());
+  }
+
+  static deserialize(deserializer: Deserializer): PublicKey {
+    const value = deserializer.deserializeBytes();
+    return new PublicKey({ hexInput: value });
+  }
+}
+
+export class PrivateKey {
```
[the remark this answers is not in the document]
0xjinn: Yep can do!

ecosystem/typescript/sdk_v2/tests/unit/ed25519.test.ts
```diff
@@ -0,0 +1,112 @@
...
+    // Verify with correct signed message
+    expect(pubKey.verifySignature({ message: messageHex, signature: expectedSignedMessage })).toBe(true);
+
+    // Verify with incorrect signed message
+    const incorrectSignedMessage =
+      "0xc5de9e40ac00b371cd83b1c197fa5b665b7449b33cd3cdd305bb78222e06a671a49625ab9aea8a039d4bb70e275768084d62b094bc1b31964f2357b7c1af7e0a";
+    expect(pubKey.verifySignature({ message: messageHex, signature: incorrectSignedMessage })).toBe(false);
+  });
+
+  it("should serialize and deserialize correctly", () => {
```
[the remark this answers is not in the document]
0xjinn: Hmm I think for this one, it might be better to keep it as a 1 round trip.

Data1 -> Serialize -> Deserialize -> Data2
Compare(Data1, Data2)

ecosystem/typescript/sdk_v2/tests/unit/helper.ts
```diff
@@ -0,0 +1,11 @@
+// Copyright © Aptos Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+export const ed25519 = {
```
0xmaayan, line 4: this comment is a nit and can be addressed in a later PR. 

The idea was to have a helper file with an Account that we can later use for all tests
```
export const account = {
  privateKey: "0xc5338cd251c22daa8c9c9cc94f498cc8a5c7e1d2e75287a5dda91096fe64efa5",
  publicKey: "0xde19e5d1880cac87d57484ce9ed2e84cf0f9599f12e7cc3a52e4e7657a763f2c",
}
```
don't think all those properties should be under a ed25519. but as mentioned before we can handle it later when we have a complete crypto/account classes

0xmaayan (approved):
noice!

ecosystem/typescript/sdk_v2/tests/unit/ed25519.test.ts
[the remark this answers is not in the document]
gregnazario: We should do both.  Serialize, deserialize ensures that your code is compatible with itself.

But, it's highly possible that you can serialize and deserialize something into a form that's different than is backwards compatible.  E.g.

1 -> 2 -> 1

But you really expect

1 -> 3
3 -> 1

ecosystem/typescript/sdk_v2/tests/unit/helper.ts
gregnazario: This is a little weird, unless we're trying to use these things across multiple places, it probably can be centralized to this test.

gregnazario:
@0xjinn can we look into making an interface that's consistent across all of these?  That may make it easier to add future schemes

ecosystem/typescript/sdk_v2/tests/unit/helper.ts
0xmaayan: just to clarify on my comment, the end goal thought is to use something similar to https://github.com/alchemyplatform/alchemy-sdk-js/blob/main/test/test-util.ts where we create an account, this account will submit some transactions and then we tests everything against this account

0xjinn: Yeah I think this make sense. Once we have the Account class setup, we can export a static Account object for testing only.

ecosystem/typescript/sdk_v2/src/crypto/ed25519.ts
```diff
@@ -0,0 +1,190 @@
...
+import { Hex } from "../core/hex";
+import { HexInput } from "../types";
+
+export class PublicKey {
+  // Correct length of the public key in bytes (Uint8Array)
+  static readonly LENGTH: number = 32;
+
+  // Hex value of the public key
+  private readonly key: Hex;
+
```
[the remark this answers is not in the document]
0xjinn: So the TS Nacl library is a bit different than the one from Python. Nacl in TS doesn't really have a public key, or `verifyKey` in python. 

```
    export interface SignKeyPair {
        publicKey: Uint8Array;
        secretKey: Uint8Array;
    }
```

It has a signkeyPair, which just holds the `Uint8Array` of both public and private key. I am currently holding this actual `SignKeyPair` as an object inside `PrivateKey` class.

ecosystem/typescript/sdk_v2/src/crypto/ed25519.ts
```diff
@@ -0,0 +1,190 @@
+// Copyright © Aptos Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+import nacl from "tweetnacl";
+import { Deserializer } from "../bcs/deserializer";
+import { Serializer } from "../bcs/serializer";
+import { Hex } from "../core/hex";
+import { HexInput } from "../types";
+
+export class PublicKey {
```
xbtmatt, line 10:
```suggestion
export class PublicKey extends Serializable {
```

I don't think it will work right now due to a conflict with the `toUint8Array` function (which I will change to match bcs.move and the rust/python sdk), but writing this here as a reminder that with these Serializable/Deserializable classes we should explicitly `extends Serializable` for the helper function that serializes and returns the bytes

0xjinn:
Update:

1. Added asymmetric_crypto which contains `Private and Public key, Signature base abstracted classes`
2. Update test to include additional Serialize and Deserialize method
3. Make sure crypto concrete classes all extend from base asymmetric_crypto class
4. Implement method from base classes

TODO:
We will probably need to update the interface in `Deserializable` to implement static instead of instance method `deserialize` cc @xbtmatt

ecosystem/typescript/sdk_v2/src/crypto/ed25519.ts
```diff
@@ -0,0 +1,190 @@
...
+  }
+
+  /**
+   * Get the private key as a hex string with the 0x prefix.
+   *
+   * @returns string representation of the private key
+   */
+  toString(): string {
+    return Hex.fromHexInput({ hexInput: this.signingKeyPair.secretKey.slice(0, 32) }).toString();
+  }
```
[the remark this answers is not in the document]
0xjinn: Sounds good!

Fixed

ecosystem/typescript/sdk_v2/tests/unit/ed25519.test.ts
```diff
@@ -0,0 +1,139 @@
...
+  it("should create the instance correctly without error", () => {
+    // Create from string
+    const hexStr = "0x0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef";
+    const publicKey = new PublicKey({ hexInput: hexStr });
+    expect(publicKey).toBeInstanceOf(PublicKey);
+    expect(publicKey.toString()).toEqual(hexStr);
+
+    // Create from Uint8Array
+    const hexUint8Array = new Uint8Array(PublicKey.LENGTH);
+    const publicKey2 = new PublicKey({ hexInput: hexUint8Array });
```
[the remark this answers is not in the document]
0xjinn: Yeah, if user create a public key directly, its not validated. But eventually when user is creating an `Account`, the key pair will be validated.

ecosystem/typescript/sdk_v2/tests/unit/ed25519.test.ts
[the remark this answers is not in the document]
0xjinn: Added additional tests based on above comments.

ecosystem/typescript/sdk_v2/src/crypto/asymmetric_crypto.ts
```diff
@@ -0,0 +1,29 @@
...
+
+import { Deserializable, Deserializer, Serializable, Serializer } from "../bcs";
+import { HexInput } from "../types";
+
+export abstract class PublicKey implements Serializable, Deserializable<PublicKey> {
+  // Verify the given message with the public key and signature.
+  abstract verifySignature(args: { data: HexInput; signature: Signature }): boolean;
+
+  // TODO: This should be a static method.
+  abstract deserialize(deserializer: Deserializer): PublicKey;
```
0xjinn, line 12: I am marking these TODO right now. cc @xbtmatt

ecosystem/typescript/sdk_v2/src/crypto/ed25519.ts
```diff
@@ -0,0 +1,212 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+import nacl from "tweetnacl";
+import { Deserializer } from "../bcs/deserializer";
+import { Serializer } from "../bcs/serializer";
+import { Hex } from "../core/hex";
+import { HexInput } from "../types";
+import * as asymmetric_crypto from "./asymmetric_crypto";
+
+export class PublicKey extends asymmetric_crypto.PublicKey {
```
0xmaayan, line 11: this is a weird js usage. I assume you had naming issue by doing `import {PublicKey, PrivateKey, Signature} from "./asymmetric_crypto";` . if we have an abstract class for each, lets change the derived class name then?
```
export class Ed25519PublicKey
```

0xjinn: I was trying to mimic closely with Python SDK lol

But yeah, `Ed25519PublicKey` works

xbtmatt:
Is the `asymmetric_crypto.ts` file just for mirroring the python SDK? To consolidate the key classes + signature class into one file?

0xjinn:
> Is the `asymmetric_crypto.ts` file just for mirroring the python SDK? To consolidate the key classes + signature class into one file?

@xbtmatt  Yep, basically a file to includes all of the crypto abstract classes. Saw python using the name `asymmetric_crypto` - Thought it would be better to mirror the file name

ecosystem/typescript/sdk_v2/src/crypto/asymmetric_crypto.ts
```diff
@@ -0,0 +1,29 @@
...
+  abstract serialize(serializer: Serializer): void;
+}
+
+export abstract class PrivateKey implements Serializable, Deserializable<PrivateKey> {
+  // Sign the given message with the private key.
+  abstract sign(args: { message: HexInput }): Signature;
```
gregnazario, line 18: Future, add signTransaction

ecosystem/typescript/sdk_v2/src/crypto/asymmetric_crypto.ts
```diff
@@ -0,0 +1,29 @@
+// Copyright © Aptos Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+import { Deserializable, Deserializer, Serializable, Serializer } from "../bcs";
```
0xmaayan, line 4: not completely sure what this file is for? is it for later when we might have different schemas, each schema would extend those classes? is it always going to use bcs to serialize/deserialize?

gregnazario (approved):

ecosystem/typescript/sdk_v2/src/crypto/asymmetric_crypto.ts
0xjinn: Yep, that is the idea. Each PublicKey and PrivateKey will extend from base classes and required to implement the Serialize/Deserialize methods, including (sign for PrivateKey, and verify for PublicKey)

> is it always going to use bcs to serialize/deserialize

Hmm this I am not sure. But if new serialization format is needed, we can modify the serializer/deserializer to accommodate that.

PR #10169, opened 2023-09-21
[GHA] Stop running unnecessary tests on release branch pushes.

JoshLind:
### Description
This PR:
1. Avoids running the PFN and state sync performance tests on release branch pushes. These runs are not needed.
2. Replaces the continuous syncing mode for fast sync in the dev docs.

### Test Plan
Existing test infrastructure.

### DIFF
--- a/.github/workflows/forge-pfn.yaml
+++ b/.github/workflows/forge-pfn.yaml
@@ -23,10 +23,6 @@ on:
   pull_request:
     paths:
       - ".github/workflows/forge-pfn.yaml"
-      - "testsuite/find_latest_image.py"
-  push:
-    branches:
-      - aptos-release-v* # The aptos release branches
 
 env:
   AWS_ACCOUNT_NUM: ${{ secrets.ENV_ECR_AWS_ACCOUNT_NUM }}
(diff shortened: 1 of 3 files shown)


sionescu (approved):

.github/workflows/forge-pfn.yaml
```diff
@@ -23,10 +23,6 @@ on:
   pull_request:
     paths:
       - ".github/workflows/forge-pfn.yaml"
-      - "testsuite/find_latest_image.py"
-  push:
-    branches:
-      - aptos-release-v* # The aptos release branches
```
rustielin, line 29: should we also stop running all other forge-stable/unstable, etc tests on release branch pushes, and instead have them triggered ad-hoc ?

JoshLind: - I think only the stable runs are currently running on the release branch pushes. 😄 None of the other ones should be.
- I agree with you that they should likely all be removed, but not everyone agrees. We'll need to handle that aspect differently.

sherry-x (approved):

grao1991 (approved):

PR #10172, opened 2023-09-21
[TS SDK] Add indexer `events` query to also support events v2

0xmaayan:
### Description
Add `getEvents()` function in `IndexerClient` for a better events query support (includes events v2).

note: we might want to structure a better dev exp when filtering graphql/indexer queries - probably will be part of sdk v2
### Test Plan
```
✓ gets events (1196 ms)
✓ gets events by type (1164 ms)
✓ gets events by type and account address (1154 ms)
```


lightmark:
Would we also add get_module_events?

0xmaayan:
> Would we also add get_module_events?

wdym?

gregnazario (approved):

davidiw (approved):

lightmark:
I mean a helper function that fill the 0 field for the user.

lightmark (approved):

PR #10174, opened 2023-09-21
[State Sync] Increase various latency buckets.

JoshLind:
### Description
This PR increases the range of various latency buckets used by state sync.

### Test Plan
Existing test infrastructure.

### DIFF
--- a/state-sync/state-sync-driver/src/metrics.rs
+++ b/state-sync/state-sync-driver/src/metrics.rs
@@ -143,12 +143,19 @@ pub static STORAGE_SYNCHRONIZER_GAUGES: Lazy<IntGaugeVec> = Lazy::new(|| {
     .unwrap()
 });
 
+// Latency buckets for storage synchronizer operations
+const STORAGE_SYNCHRONIZER_LATENCY_BUCKETS_SECS: &[f64] = &[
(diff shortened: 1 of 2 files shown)


state-sync/state-sync-driver/src/metrics.rs
```diff
@@ -143,14 +143,20 @@ pub static STORAGE_SYNCHRONIZER_GAUGES: Lazy<IntGaugeVec> = Lazy::new(|| {
...
 /// Counter for tracking storage synchronizer latencies
 pub static STORAGE_SYNCHRONIZER_LATENCIES: Lazy<HistogramVec> = Lazy::new(|| {
-    register_histogram_vec!(
+    let histogram_opts = histogram_opts!(
         "aptos_state_sync_storage_synchronizer_latencies",
         "Counters related to the storage synchronizer latencies",
-        &["label"]
-    )
-    .unwrap()
+        STORAGE_SYNCHRONIZER_LATENCY_BUCKETS_SECS.to_vec()
```
bchocho, line 157: IMO it's a bit simpler to add `STORAGE_SYNCHRONIZER_LATENCY_BUCKETS_SECS.to_vec()` as the next argument to the `register_historgram_vec!` macro

bchocho (approved):
nit below

state-sync/state-sync-driver/src/metrics.rs
JoshLind: Updated -- I don't feel super strongly 😃

grao1991 (approved):

Issue #9410, opened 2023-08-01
[Bug] Turn off suppression of warnings on unknown attributes in library code when it's safe


brmataptos:
# 🐛 Bug

I'm soon going to [deploy a fix](https://github.com/aptos-labs/aptos-core/pull/9229) to warn on unknown attributes in Move code, but will include a suppression for library code until libraries can be updated.  Once fixed libraries are deployed everywhere we should remove that suppression (search for this issue # in  `third-party/move/move-compiler/src/expansion/translate.rs`).

brmataptos:
This was merged.

PR #10175, opened 2023-09-21
Fix issue 9410 by removing special-case hack to suppress unknown attribute warnings on aptos_stdlib

brmataptos:
### Description

[PR 9229](https://github.com/aptos-labs/aptos-core/pull/9229) introduced warnings on unknown attributes, but suppressed them for aptos_stdlib to avoid user consternation about problems in the Aptos codebase.  It also removed those attributes from the codebase, so now that it is released it becomes safe to Fix i[ssue 9410](https://github.com/aptos-labs/aptos-core/issues/9410) and remove the special-case for the aptos_stdlib.

### Test Plan
Existing test cases, new test outputs.


vineethk (approved):
lgtm.

wrwg (approved):

PR #4649, opened 2022-09-29
Adding configurabele TCP buffers and minor hafixes

Markuze:
- [Network] set socket RX/TX Buffers

Adding configurable TCP buffers for inbound and outbound TCP connections.
This change will improve networking performance, especially for long RTT connections.

### Test Plan

- - -
This change is [Reviewable](https://reviewable.io/reviews/aptos-labs/aptos-core/4649)

PR #10176, opened 2023-09-21
[Networking] set TCP buffer overrides to None by default

bchocho:
### Description

Tests show that overriding the socket options for send and recv buffer actually hurts performance (even with default linux kernel settings). Logically, this rolls back #4649 but keeping the configs and code around in case this or other socket options are revisited.

### Test Plan

New results:
* `net_bench_two_region_env`: 800 KB/s -> 4 MB/s
* `pfn_performance_with_realistic_env`: 4K -> 6K TPS

### DIFF
--- a/config/src/config/network_config.rs
+++ b/config/src/config/network_config.rs
@@ -53,10 +53,6 @@ pub const MAX_MESSAGE_SIZE: usize = 64 * 1024 * 1024; /* 64 MiB */
 pub const CONNECTION_BACKOFF_BASE: u64 = 2;
 pub const IP_BYTE_BUCKET_RATE: usize = 102400 /* 100 KiB */;
 pub const IP_BYTE_BUCKET_SIZE: usize = IP_BYTE_BUCKET_RATE;
-pub const INBOUND_TCP_RX_BUFFER_SIZE: u32 = 3 * 1024 * 1024; // 3MB ~6MB/s with 500ms latency
-pub const INBOUND_TCP_TX_BUFFER_SIZE: u32 = 512 * 1024; // 1MB use a bigger spoon
-pub const OUTBOUND_TCP_RX_BUFFER_SIZE: u32 = 3 * 1024 * 1024; // 3MB ~6MB/s with 500ms latency
-pub const OUTBOUND_TCP_TX_BUFFER_SIZE: u32 = 1024 * 1024; // 1MB use a bigger spoon
 
 #[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
 #[serde(default, deny_unknown_fields)]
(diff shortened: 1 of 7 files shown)


bchocho:
Forge result links:
* [net_bench_two_region_env](https://github.com/aptos-labs/aptos-core/actions/runs/6270456200): 800 KB/s -> 4 MB/s
* [pfn_performance_with_realistic_env](https://github.com/aptos-labs/aptos-core/actions/runs/6270152596): 4K -> 6K TPS

testsuite/forge-cli/src/main.rs
```diff
@@ -2062,13 +2062,7 @@ fn pfn_performance(
...
 ) -> ForgeConfig {
     // Determine the minimum expected TPS
-    let min_expected_tps = if add_cpu_chaos {
-        3000
-    } else if add_network_emulation {
-        4000
-    } else {
-        4500
-    };
+    let min_expected_tps = 4500;
```
bchocho, line 2065: Keeping this pretty conservative for now. I've seen a range of 5-6.2K TPS.

sitalkedia (approved):

config/src/config/network_config.rs
```diff
@@ -157,10 +149,6 @@ impl NetworkConfig {
             inbound_rate_limit_config: None,
             outbound_rate_limit_config: None,
             max_message_size: MAX_MESSAGE_SIZE,
-            inbound_rx_buffer_size_bytes: Some(INBOUND_TCP_RX_BUFFER_SIZE),
```
JoshLind, line 160: Silly question, but, instead of removing support for this entirely, is there a way to just set the default to not override? e.g., I'm just wondering if there's ever a scenario where we want to set these things?

brianolson: re-adding tcp socket option setting would have the annoying work of figuring out where it goes, so leaving this on but disabled (it already had Option if-None-do-nothing) makes sense to me

bchocho: I honestly don't think we want to mess with this and so might as well clean up the configs now. For this to be effective the user would have to bump up the linux configs, otherwise it has a detrimental effect. 

If you feel strongly I'll leave the config but add a comment on the ways it shouldn't be used.

brianolson: Leave it in or at least leave a comment `// TODO: if needed, set connection socket options here`; _someone_ is going to want to tinker with this again, even if Linux has pretty good defaults

bchocho: Thanks @JoshLind @brianolson . Makes sense, I guess there are other socket options that someone might want to play with in the future as well. I've changed the PR to just set the defaults as `None` and put up a big comment to proceed with caution.

PR #10182, opened 2023-09-22
[DO NOT LAND] CoreResourceGroup

msmouse:

### DIFF
--- a/aptos-move/framework/aptos-framework/sources/account.move
+++ b/aptos-move/framework/aptos-framework/sources/account.move
@@ -6,6 +6,7 @@ module aptos_framework::account {
     use std::signer;
     use std::vector;
     use aptos_framework::chain_id;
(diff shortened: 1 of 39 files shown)


aptos-move/framework/aptos-framework/sources/account.move
```diff
@@ -24,6 +25,7 @@ module aptos_framework::account {
     friend aptos_framework::transaction_validation;
 
     /// Resource representing an account.
+    #[resource_group_member(group=CoreResourceGroup)]
```
movekevin, line 28: This would really break things as making an existing standalone resource part of a group means you wouldn't be able to find the existing values due to how the resolver works

msmouse: yeah testing on forge with fresh genesis

sorry didn't intend to publish

PR #10196, opened 2023-09-22
Get Drand Move example to compile

mstraka100:
### Description

Currently the `drand` example code in `move-examples` compiles using the Rust test harness invoked by `cargo test` but does not compile with `aptos move compile` using the Aptos CLI. This PR fixes this issue. In addition, the package names of the Drand and VeiledCoin examples have been changed to upper camelcase to make it easier to use them as dependencies in other Move packages.

### Test Plan
This PR was tested running `aptos move compile --named-addresses drand=0x8` and  `aptos move test --named-addresses drand=0x8` in `aptos-move/move-examples/drand`, in addition to `RUST_MIN_STACK=10000000000 cargo test -- drand` in `aptos-move/move-examples`.


aptos-move/move-examples/drand/diff.txt
alinush, line 1: Uhm, what's up with this diff file?

aptos-move/move-examples/drand/sources/lottery_test.move
```diff
@@ -119,9 +125,6 @@ module drand::lottery_test {
         //
         // Send a TXN with `drand_signed_bytes` to close the lottery and determine the winner
         //
-        let winner_addr = option::extract(&mut lottery::close_lottery(drand_signed_bytes));
-
-        debug::print(&string::utf8(b"The winner is: "));
-        debug::print(&winner_addr)
+        lottery::close_lottery(drand_signed_bytes);
```
alinush, line 128: Why remove this? We need this when testing.

aptos-move/move-examples/drand/diff.txt
mstraka100: Adding a package with spaces in the middle as a dependency is a pain. 

```[dependencies]
   Veiled Coin = ``<etc>`
   
   doesn't work because of the space.

alinush (changes requested):
Add back the debug printing in the test & remove the diff.txt file?

aptos-move/move-examples/drand/sources/lottery_test.move
mstraka100: Entry functions can't return values so we can't access the winner from `close_lottery`. We could make another function to return it but this would require refactoring the code to store the winner in a lottery instance.

aptos-move/move-examples/drand/diff.txt
mstraka100: Oops, misread this. `Diff.txt` has been removed

aptos-move/move-examples/drand/sources/lottery_test.move
mstraka100: (This is why the code previously compiled with `cargo test` but not `move compile` - for whatever reason the Rust testing harness doesn't check for this)

mstraka100: I just updated the contract with a public function that returns the current lottery winner

alinush: A layer of indirection would have worked too: `public close_lottery_internal()` could return the winner and the `entry` function just calls that and ignores the return value.

That's simpler, but this is okay too.

alinush (approved):

vgao1996 (approved):

zjma (approved):

PR #10200, opened 2023-09-23
[python-sdk] improving usage around keys

davidiw:
* Made a common asymmetric crypto key class
* Made it possible to rotate to multi keys

### DIFF
--- a/ecosystem/python/sdk/aptos_sdk/asymmetric_crypto.py
+++ b/ecosystem/python/sdk/aptos_sdk/asymmetric_crypto.py
@@ -1,0 +1,32 @@
+# Copyright © Aptos Foundation
+# SPDX-License-Identifier: Apache-2.0
+
+from __future__ import annotations
+
+from typing_extensions import Protocol
+
+from .bcs import Deserializable, Serializable
+
+
+class PrivateKey(Deserializable, Serializable, Protocol):
+    def public_key(self) -> PublicKey:
+        ...
+
+    def sign(self, data: bytes) -> Signature:
+        ...
+
+
+class PublicKey(Deserializable, Serializable, Protocol):
+    def to_crypto_bytes(self) -> bytes:
+        """
+        A long time ago, someone decided that we should have both bcs and a special representation
+        for MultiEd25519, so we use this to let keys self-define a special encoding.
+        """
+        ...
+
+    def verify(self, data: bytes, signature: Signature) -> bool:
+        ...
+
+
+class Signature(Deserializable, Serializable, Protocol):
+    ...
--- a/ecosystem/python/sdk/examples/rotate_key.py
+++ b/ecosystem/python/sdk/examples/rotate_key.py
@@ -1,11 +1,13 @@
 import asyncio
+from typing import List
 
+import aptos_sdk.asymmetric_crypto as asymmetric_crypto
+import aptos_sdk.ed25519 as ed25519
 from aptos_sdk.account import Account, RotationProofChallenge
 from aptos_sdk.account_address import AccountAddress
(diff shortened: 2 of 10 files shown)


ecosystem/python/sdk/aptos_sdk/asymmetric_crypto.py
```diff
@@ -0,0 +1,32 @@
...
+        for MultiEd25519, so we use this to let keys self-define a special encoding.
+        """
+        ...
+
+    def verify(self, data: bytes, signature: Signature) -> bool:
+        ...
+
+
+class Signature(Deserializable, Serializable, Protocol):
+    ...
```
gregnazario, line 32: Were you going to implement this?

ecosystem/python/sdk/examples/rotate_key.py
```diff
@@ -122,13 +167,26 @@ async def main():
...
+    )
+    signed_transaction = await rest_client.create_bcs_signed_transaction(alice, payload)
+    tx_hash = await rest_client.submit_bcs_transaction(signed_transaction)
+    await rest_client.wait_for_transaction(tx_hash)
+
+    alice_new_account_info = await rest_client.account(alice.address())
+    auth_key = alice_new_account_info["authentication_key"]
+    print(f"Rotation to MultiPublicKey complete, new authkey: {auth_key}")
+
     await rest_client.close()
```
gregnazario, line 191: Can you add a test for using the multied25519?

gregnazario (approved):

ecosystem/python/sdk/aptos_sdk/asymmetric_crypto.py
davidiw: mypy Protocol is used to define interfaces, so that I can get away with compatible yet distinct types.

ecosystem/python/sdk/examples/rotate_key.py
davidiw: there are some in the code base and unfortunately the rest of the code isn't implemented... maybe I can slap another PR on top of this for richer functionality...

0xjinn:
Are these "asymmetric crypto key class" means to solve the problem for adapting multiple different schemes (ed25519, secp256k1, etc)? I am currently working on the crypto stuff on ts sdkv2. Maybe I will follow this design to support multiple different scheme as well :)

davidiw:
> Are these "asymmetric crypto key class" means to solve the problem for adapting multiple different schemes (ed25519, secp256k1, etc)? I am currently working on the crypto stuff on ts sdkv2. Maybe I will follow this design to support multiple different scheme as well :)

Yes, this is how you have typed inheritance in Python :)

ecosystem/python/sdk/aptos_sdk/asymmetric_crypto.py
```diff
@@ -0,0 +1,32 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+from __future__ import annotations
+
+from typing_extensions import Protocol
+
+from .bcs import Deserializable, Serializable
+
+
+class PrivateKey(Deserializable, Serializable, Protocol):
```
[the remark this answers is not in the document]
davidiw: how do you read a private key from disk? I mean you could use to / from, you could also theoretically have a BCS encoded file on disk representing an account :shrug:

0xjinn (approved):

ecosystem/python/sdk/examples/rotate_key.py
davidiw: gonna punt on this now... using multied is a pain because the account abstraction isn't great... I think once we improve the area around a Multi Account Authenticator TA we can revisit.

PR #10207, opened 2023-09-25
[crypto] add secp256k1 support and some other fun stuff

davidiw:
* Add aptos-crypto support for secp256k1 -- thanks @mstraka100 for his due diligence on r1, so I could get the malleability check in mine as well :grin: 
* Add authenticator support for secp256k1 -- this needs an end-to-end test, but that's quite trivial at this point in time...
* Clean up some legacy authenticator code that just makes it painful to add new authenticators

I think we likely still need to rework transaction code that is all over the code base in 3 different places and just make everyone conform to using something like the SDK.... or push the core of the SDK into types and then make the SDK just aware of the API.

Anyway to land this, I'll still need e2e test and some gatekeeping... I'd probably be fine landing the crypto and updates to the core authenticator code if folks can review that faster than me getting the last bits done.


api/src/tests/transactions_test.rs
```diff
@@ -337,8 +337,8 @@ async fn test_multi_ed25519_signed_transaction() {
     context.commit_block(&vec![create_account_txn]).await;
 
     let raw_txn = factory
-        .mint(auth_key.derived_address(), 1000)
```
movekevin, line 340: lol

api/types/src/transaction.rs
```diff
@@ -1087,25 +1091,98 @@ impl TryFrom<MultiEd25519Signature> for AccountAuthenticator {
...
 ///   1. A single Ed25519 key account, one private key
 ///   2. A k-of-n multi-Ed25519 key account, multiple private keys, such that k-of-n must sign a transaction.
+///   3. A single Secp256k1 key account, one private key
 #[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize, Union)]
 #[serde(tag = "type", rename_all = "snake_case")]
 #[oai(one_of, discriminator_name = "type", rename_all = "snake_case")]
 pub enum AccountSignature {
     Ed25519Signature(Ed25519Signature),
     MultiEd25519Signature(MultiEd25519Signature),
+    Secp256k1Signature(Secp256k1Signature),
```
movekevin, line 1177: Makes sense to add here. The only part that is slightly strange is when a multied25519 account can be authenticated by Secp256k1 keys. Would there be any possible problems with this?

davidiw: No, multied25519, as it is, cannot. we need a new general authenticator that supports arbitrary public keys, which the current one does not.

davidiw: also note, this is just the annoying json representation stuff... the core code is in types/src/transaction/authenticator.rs

movekevin: Hmmm then fee payer would not work with this either?

davidiw: https://github.com/aptos-labs/aptos-core/blob/main/types/src/transaction/authenticator.rs#L61

note, that the feepayer takes in AccountAuthenticator - which can be the Secp256k1
the multied25519 takes in a ... weird multied25519public key

I think the right approach would have been effectively making the multi-keys that use many account authenticators instead of shoving this into the crypto code...

movekevin: Makes sense

movekevin (approved):
Code looks good to me. Just need more tests:
1. E2E API tests that cover valid/invalid signatures. May be one for fee payer using secp256k1 as well
2. (Here or followup) Some light TS or Python SDK examples that use secp256k1 signature

crates/aptos-crypto/src/secp256k1.rs
```diff
@@ -0,0 +1,300 @@
...
+
+impl std::fmt::Debug for Signature {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "secp256k1::Signature({})", self)
+    }
+}
+
+impl std::fmt::Display for Signature {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", hex::encode(&self.to_bytes()[..]))
```
gregnazario, line 249: `&self.to_bytes()[..]` could be `self.to_bytes().as_slice()`

gregnazario (approved):

gregnazario:
@davidiw did you generate the OpenAPI spec as well?

api/src/tests/transactions_test.rs
```diff
@@ -337,8 +337,8 @@ async fn test_multi_ed25519_signed_transaction() {
...
-        .sender(auth_key.derived_address())
+        .mint(auth_key.account_address(), 1000)
```
alinush, line 340: What's the difference?

api/types/src/transaction.rs
```diff
@@ -869,6 +870,7 @@ pub enum TransactionSignature {
     MultiEd25519Signature(MultiEd25519Signature),
     MultiAgentSignature(MultiAgentSignature),
     FeePayerSignature(FeePayerSignature),
+    Secp256k1Signature(Secp256k1Signature),
```
alinush, line 873: Can you please accurately name this `Secp256k1Ecdsa` signature? It's currently incomplete and confusing. Plus, we could have all sorts of other signatures over the `secp256k1` curve later on.

davidiw: what other signature types are there?

api/types/src/transaction.rs
```diff
@@ -1087,25 +1091,98 @@ impl TryFrom<MultiEd25519Signature> for AccountAuthenticator {
...
+    }
+}
+
 /// Account signature scheme
 ///
 /// The account signature scheme allows you to have two types of accounts:
 ///
 ///   1. A single Ed25519 key account, one private key
 ///   2. A k-of-n multi-Ed25519 key account, multiple private keys, such that k-of-n must sign a transaction.
+///   3. A single Secp256k1 key account, one private key
```
alinush, line 1170: `Secp256k1Ecdsa`

api/src/tests/transactions_test.rs
davidiw: I found this name confusing, hard to find, and implies something that isn't accurate.

crates/aptos-crypto/src/secp256k1.rs
alinush, line 1: Rename to `secp256k1_ecdsa.rs`!

crates/aptos-crypto/src/secp256k1.rs
```diff
@@ -0,0 +1,300 @@
...
+
+impl TryFrom<&[u8]> for Signature {
+    type Error = CryptoMaterialError;
+
+    fn try_from(bytes: &[u8]) -> std::result::Result<Signature, CryptoMaterialError> {
+        match libsecp256k1::Signature::parse_standard_slice(bytes) {
+            Ok(signature) => {
+                // Prevent malleability attacks, low order only. The library only signs in low
+                // order, so this was done intentionally.
+                if signature.s.is_high() {
```
alinush, line 230: This needs to be done at signature verification time, not deserialization time.

crates/aptos-crypto/src/secp256k1.rs
```diff
@@ -0,0 +1,300 @@
...
+    pub fn to_bytes(&self) -> Vec<u8> {
+        self.0.serialize().to_vec()
+    }
+
+    fn verify(
+        &self,
+        message: &libsecp256k1::Message,
+        public_key: &libsecp256k1::PublicKey,
+    ) -> Result<()> {
+        if libsecp256k1::verify(message, &self.0, public_key) {
```
alinush, line 206: High `s` check **MUST** be done here.

davidiw: There is no way for a high s to get here if we block it at the deserialization time.

crates/aptos-crypto/src/secp256k1.rs
```diff
@@ -0,0 +1,300 @@
...
+}
+
+impl PrivateKey {
+    /// Serialize the private key into a byte vector
+    pub fn to_bytes(&self) -> Vec<u8> {
+        self.0.serialize().to_vec()
+    }
+
+    fn sign(&self, message: &libsecp256k1::Message) -> Signature {
+        let (signature, _recovery_id) = libsecp256k1::sign(message, &self.0);
```
alinush, line 47: You never "normalize" the signature to have low `s`, do you? I thought you said you do.

This must also be done here.

davidiw: The library does that.

alinush: Oh. `libsecp256k1` does it...

alinush: Can you add a comment at the top of this function saying the signature is "normalized" to have "low" `s`, please? Otherwise, some of us will get a heart-attack every time we look at this function.

types/src/transaction/authenticator.rs
```diff
@@ -379,18 +398,29 @@ impl fmt::Display for Scheme {
     }
 }
 
+/// An `AccountAuthenticator` is an an abstraction of a signature scheme. It must know:
+/// (1) How to check its signature against a message and public key
+/// (2) How to convert its public key into an `AuthenticationKeyPreimage` structured as
+/// (public_key | signaure_scheme_id).
```
alinush, line 404: Typo: signa**t**ure_scheme_id

types/src/transaction/authenticator.rs
```diff
@@ -422,6 +453,14 @@ impl AccountAuthenticator {
         }
     }
 
+    /// Create a single-signature ed25519 authenticator
```
alinush, line 456: Comment is wrong here: should be `secp256k1[_ecdsa]` :)

alinush (dismissed):
Did a pass: most important ask is the malleability check should be done at signature verification time :)

davidiw:
> Did a pass: most important ask is the malleability check should be done at signature verification time :)

Appreciate your time and due diligence.

api/types/src/transaction.rs
zjma: EC-Schnorr, for example. `Secp256k1` sounds more a curve to me, rather than a signature scheme.

crates/aptos-crypto/src/unit_tests/secp256k1_test.rs
```diff
@@ -0,0 +1,105 @@
...
+    assert_eq!(signature, signature_deserialized);
+
+    let private_key_bytes = key_pair.private_key.to_bytes();
+    let private_key_deserialized = secp256k1::PrivateKey::try_from(&private_key_bytes[..]).unwrap();
+    assert_eq!(key_pair.private_key, private_key_deserialized);
+
+    let public_key_bytes = key_pair.public_key.to_bytes();
+    let public_key_deserialized = secp256k1::PublicKey::try_from(&public_key_bytes[..]).unwrap();
+    assert_eq!(key_pair.public_key, public_key_deserialized);
+}
```
zjma, line 77: what about the (de)serialization failures ;)

crates/aptos-crypto/src/unit_tests/secp256k1_test.rs
```diff
@@ -0,0 +1,105 @@
...
+    let signature_deserialized = secp256k1::Signature::try_from(&signature_bytes[..]).unwrap();
+    assert_eq!(signature, signature_deserialized);
+
+    let mut high_signature = signature.clone();
+    high_signature.0.s = -high_signature.0.s;
+    assert!(high_signature.0.s.is_high());
+    assert!(high_signature.0.s != signature.0.s);
+    assert!(high_signature
+        .verify_arbitrary_msg(message, &key_pair.public_key)
+        .is_ok());
```
zjma, line 101: we do really need these 3 assertions? they look more like UTs for libsecp256k1?

hariria:
Instead of making a `TransactionAuthenticator` scoped to just `secp256k1` could we make a `SingleSigner` enum variant for TransactionAuthenticator that takes in a `sender` with type `AccountAuthenticator`? That way we can reuse it for other signature schemes in the future

davidiw:
> Instead of making a `TransactionAuthenticator` scoped to just `secp256k1` could we make a `SingleSigner` enum variant for TransactionAuthenticator that takes in a `sender` with type `AccountAuthenticator`? That way we can reuse it for other signature schemes in the future

Unfortunately that isn't really feasible without a breaking change. We could start one off for secp256k1, but I'm not convinced it is the right problem to solve at this point in time. I do think if we had done this properly, your insight is apt. Unfortunately the original designer of this API got a little too fancy with the multi-ed25519 and other aspects :(

crates/aptos-crypto/src/unit_tests/secp256k1_test.rs
davidiw: Yes, they were left behind when I had a failing test.

crates/aptos-crypto/src/unit_tests/secp256k1_test.rs
davidiw: not sure how interesting that test is ... but added.

davidiw:
@alinush , it is finished, please have a review in the AM. I will intend to catch the devnet train.

api/doc/spec.json
```diff
@@ -11922,20 +11922,24 @@
       },
       "AccountSignature": {
         "type": "object",
-        "description": "Account signature scheme\n\nThe account signature scheme allows you to have two types of accounts:\n\n1. A single Ed25519 key account, one private key\n2. A k-of-n multi-Ed25519 key account, multiple private keys, such that k-of-n must sign a transaction.",
+        "description": "Account signature scheme\n\nThe account signature scheme allows you to have two types of accounts:\n\n1. A single Ed25519 key account, one private key\n2. A k-of-n multi-Ed25519 key account, multiple private keys, such that k-of-n must sign a transaction.\n3. A single Secp256k1Ecdsa key account, one private key",
```
brmataptos, line 11925: "two types" ?

brmataptos (commented):
small typo

PR #10186, opened 2023-09-22
[TSSDKv2][2/n] add Multi Ed25519 classes

0xjinn:
### Description

Note: This is stack on top of #10157

This PR adds the following classes

1. MultiEd25519PublicKey
2. MultiEd25519Signature

This version follows the same design as Python SDK

### Test Plan

jest test

PR #10210, opened 2023-09-25
[TSSDKv2][3/n] Account and AuthenticationKey Classes

0xjinn:
### Description

Note: This is the `3rd` stack of the whole work. Previous stack here -> #10186

This PR adds the Account and AuthenticationKey classes. Most of the thing stays the same as v1, but with some modification.

1. Account now holds the actual `PublicKey` and `PrivateKey` objects
2. `Verify` and `Sign` relies on the method the corresponding key
3. Using object as param

### Test Plan

Run all tests and make sure they all passes

### DIFF
--- a/ecosystem/typescript/sdk_v2/src/core/account.ts
+++ b/ecosystem/typescript/sdk_v2/src/core/account.ts
@@ -1,0 +1,150 @@
+// Copyright © Aptos Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+import nacl from "tweetnacl";
+import * as bip39 from "@scure/bip39";
+import { AccountAddress } from "./account_address";
+import { Hex } from "./hex";
+import { bytesToHex } from "@noble/hashes/utils";
+import { HexInput } from "../types";
+import { PrivateKey, PublicKey, Signature } from "../crypto/ed25519";
+import { derivePath } from "../utils/hd-key";
+import { AuthenticationKey } from "../crypto/authentication_key";
+
+/**
+ * Class for creating and managing account on Aptos network
+ *
+ * Use this class to create accounts, sign transactions, and more.
+ * Note: Creating an account instance does not create the account onchain.
+ */
+export class Account {
+  /**
+   * A private key and public key, associated with the given account
+   */
+  readonly publicKey: PublicKey;
+  readonly privateKey: PrivateKey;
+
+  /**
+   * Account address associated with the account
+   */
+  readonly accountAddress: AccountAddress;
+
+  /**
+   * constructor for Account
+   *
+   * @param args.privateKey PrivateKey - private key of the account
+   * @param args.address AccountAddress - address of the account
+   *
+   * This method is private because it should only be called by the factory static methods.
+   * @returns Account
+   */
+  private constructor(args: { privateKey: PrivateKey; address: AccountAddress }) {
+    const { privateKey, address } = args;
+
+    // Derive the public key from the private key
+    const keyPair = nacl.sign.keyPair.fromSeed(privateKey.toUint8Array().slice(0, 32));
+    this.publicKey = new PublicKey({ hexInput: keyPair.publicKey });
+
+    this.privateKey = privateKey;
+    this.accountAddress = address;
+  }
+
+  /**
+   * Generate a new account with random private key and address
+   *
+   * @returns Account
+   */
+  static generate(): Account {
+    const keyPair = nacl.sign.keyPair();
+    const privateKey = new PrivateKey({ value: keyPair.secretKey.slice(0, 32) });
+    const address = new AccountAddress({ data: Account.authKey({ publicKey: keyPair.publicKey }).toUint8Array() });
+    return new Account({ privateKey, address });
+  }
+
+  /**
+   * Creates new account with provided private key
+   *
+   * @param args.privateKey Hex - private key of the account
+   * @returns Account
+   */
+  static fromPrivateKey(args: { privateKey: HexInput }): Account {
+    const privatekeyHex = Hex.fromHexInput({ hexInput: args.privateKey });
+    const keyPair = nacl.sign.keyPair.fromSeed(privatekeyHex.toUint8Array().slice(0, 32));
+    const privateKey = new PrivateKey({ value: keyPair.secretKey.slice(0, 32) });
+    const address = new AccountAddress({ data: Account.authKey({ publicKey: keyPair.publicKey }).toUint8Array() });
+    return new Account({ privateKey, address });
+  }
+
+  /**
+   * Creates new account with provided private key and address
+   * This is intended to be used for account that has it's key rotated
+   *
+   * @param args.privateKey Hex - private key of the account
+   * @param args.address AccountAddress - address of the account
+   * @returns Account
+   */
+  static fromPrivateKeyAndAddress(args: { privateKey: HexInput; address: AccountAddress }): Account {
+    const privatekeyHex = Hex.fromHexInput({ hexInput: args.privateKey });
+    const signingKey = nacl.sign.keyPair.fromSeed(privatekeyHex.toUint8Array().slice(0, 32));
+    const privateKey = new PrivateKey({ value: signingKey.secretKey.slice(0, 32) });
+    return new Account({ privateKey, address: args.address });
+  }
+
+  /**
+   * Creates new account with bip44 path and mnemonics,
+   * @param path. (e.g. m/44'/637'/0'/0'/0')
+   * Detailed description: {@link https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki}
+   * @param mnemonics.
+   * @returns AptosAccount
+   */
+  static fromDerivationPath(args: { path: string; mnemonic: string }): Account {
+    const { path, mnemonic } = args;
+    if (!Account.isValidPath({ path })) {
+      throw new Error("Invalid derivation path");
+    }
+
+    const normalizeMnemonics = mnemonic
+      .trim()
+      .split(/\s+/)
+      .map((part) => part.toLowerCase())
+      .join(" ");
+
+    const { key } = derivePath(path, bytesToHex(bip39.mnemonicToSeedSync(normalizeMnemonics)));
+
+    const signingKey = nacl.sign.keyPair.fromSeed(key.slice(0, 32));
+    const privateKey = new PrivateKey({ value: signingKey.secretKey.slice(0, 32) });
+    const address = new AccountAddress({ data: Account.authKey({ publicKey: signingKey.publicKey }).toUint8Array() });
+
+    return new Account({ privateKey, address });
+  }
+
+  /**
+   * Check's if the derive path is valid
+   */
+  static isValidPath(args: { path: string }): boolean {
+    return /^m\/44'\/637'\/[0-9]+'\/[0-9]+'\/[0-9]+'+$/.test(args.path);
+  }
+
+  /**
+   * This key enables account owners to rotate their private key(s)
+   * associated with the account without changing the address that hosts their account.
+   * See here for more info: {@link https://aptos.dev/concepts/accounts#single-signer-authentication}
+   * @returns Authentication key for the associated account
+   */
+  static authKey(args: { publicKey: HexInput }): Hex {
+    const publicKey = new PublicKey({ hexInput: args.publicKey });
+    const authKey = AuthenticationKey.fromPublicKey({ publicKey });
+    return authKey.data;
+  }
+
+  sign(data: HexInput): Signature {
+    const signature = this.privateKey.sign({ message: data });
+    return signature;
+  }
+
+  verifySignature(message: HexInput, signature: HexInput): boolean {
+    const rawMessage = Hex.fromHexInput({ hexInput: message }).toUint8Array();
+    const rawSignature = Hex.fromHexInput({ hexInput: signature }).toUint8Array();
+    return this.publicKey.verifySignature({ data: rawMessage, signature: rawSignature });
+  }
+}
--- a/ecosystem/typescript/sdk_v2/src/core/account.ts
+++ b/ecosystem/typescript/sdk_v2/src/core/account.ts
@@ -7,9 +7,10 @@ import { AccountAddress } from "./account_address";
 import { Hex } from "./hex";
 import { bytesToHex } from "@noble/hashes/utils";
 import { HexInput } from "../types";
-import { PrivateKey, PublicKey, Signature } from "../crypto/ed25519";
+import { PrivateKey, PublicKey, Signature } from "../crypto/asymmetric_crypto";
 import { derivePath } from "../utils/hd-key";
 import { AuthenticationKey } from "../crypto/authentication_key";
+import { Ed25519PrivateKey, Ed25519PublicKey, Ed25519Signature } from "../crypto/ed25519";
 
 /**
  * Class for creating and managing account on Aptos network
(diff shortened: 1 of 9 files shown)


ecosystem/typescript/sdk_v2/src/core/account.ts
```diff
@@ -0,0 +1,150 @@
...
+  private constructor(args: { privateKey: PrivateKey; address: AccountAddress }) {
+    const { privateKey, address } = args;
+
+    // Derive the public key from the private key
+    const keyPair = nacl.sign.keyPair.fromSeed(privateKey.toUint8Array().slice(0, 32));
+    this.publicKey = new PublicKey({ hexInput: keyPair.publicKey });
+
+    this.privateKey = privateKey;
+    this.accountAddress = address;
+  }
```
gregnazario, line 50: This is tied to an Ed25519 key, it should probably take in some interface supporting multiple types of keys.  Otherwise, it doesn't support MultiEd25519, secpk etc.

gregnazario (commented):
Can we see if we can provide some sort of interfacing that will support multiple signing schemes on the account?  That way we can support immediately Ed25519, MultiED25519, and extend to other schemes as well.

We should also look into multiple ways to derive account addresses for things like objects

ecosystem/typescript/sdk_v2/src/core/account.ts
```diff
@@ -0,0 +1,150 @@
...
+  /**
+   * constructor for Account
+   *
+   * @param args.privateKey PrivateKey - private key of the account
+   * @param args.address AccountAddress - address of the account
+   *
+   * This method is private because it should only be called by the factory static methods.
+   * @returns Account
+   */
+  private constructor(args: { privateKey: PrivateKey; address: AccountAddress }) {
```
0xmaayan, line 41: TIL

ecosystem/typescript/sdk_v2/src/core/account.ts
```diff
@@ -0,0 +1,150 @@
...
+   */
+  static generate(): Account {
+    const keyPair = nacl.sign.keyPair();
+    const privateKey = new PrivateKey({ value: keyPair.secretKey.slice(0, 32) });
+    const address = new AccountAddress({ data: Account.authKey({ publicKey: keyPair.publicKey }).toUint8Array() });
+    return new Account({ privateKey, address });
+  }
+
+  /**
+   * Creates new account with provided private key
```
0xmaayan, line 65: maybe?
```suggestion
   * Retrieves an account with provided private key
```
as we dont really create a new account

ecosystem/typescript/sdk_v2/src/core/account.ts
```diff
@@ -0,0 +1,150 @@
...
+  static fromPrivateKey(args: { privateKey: HexInput }): Account {
+    const privatekeyHex = Hex.fromHexInput({ hexInput: args.privateKey });
+    const keyPair = nacl.sign.keyPair.fromSeed(privatekeyHex.toUint8Array().slice(0, 32));
+    const privateKey = new PrivateKey({ value: keyPair.secretKey.slice(0, 32) });
+    const address = new AccountAddress({ data: Account.authKey({ publicKey: keyPair.publicKey }).toUint8Array() });
+    return new Account({ privateKey, address });
+  }
+
+  /**
+   * Creates new account with provided private key and address
```
0xmaayan, line 79: same

ecosystem/typescript/sdk_v2/src/core/account.ts
```diff
@@ -0,0 +1,150 @@
...
+   * See here for more info: {@link https://aptos.dev/concepts/accounts#single-signer-authentication}
+   * @returns Authentication key for the associated account
+   */
+  static authKey(args: { publicKey: HexInput }): Hex {
+    const publicKey = new PublicKey({ hexInput: args.publicKey });
+    const authKey = AuthenticationKey.fromPublicKey({ publicKey });
+    return authKey.data;
+  }
+
+  sign(data: HexInput): Signature {
```
0xmaayan, line 140: can we add comments for those two functions

ecosystem/typescript/sdk_v2/src/crypto/authentication_key.ts
```diff
@@ -0,0 +1,102 @@
...
+import { sha3_256 as sha3Hash } from "@noble/hashes/sha3";
+import { AccountAddress, Hex } from "../core";
+import { HexInput } from "../types";
+import { MultiPublicKey } from "./multi_ed25519";
+import { PublicKey } from "./ed25519";
+
+/**
+ * Each account stores an authentication key. Authentication key enables account owners to rotate
+ * their private key(s) associated with the account without changing the address that hosts their account.
+ * @see {@link * https://aptos.dev/concepts/accounts | Account Basics}
```
0xmaayan, line 13:
```suggestion
 * @see {@link https://aptos.dev/concepts/accounts | Account Basics}
```

ecosystem/typescript/sdk_v2/src/core/account.ts
0xjinn: Yeah I see there is an active new scheme `secp256k1` being added now. Will try to sync up more closely with what @davidiw had on `asymmetric_keys` - https://github.com/aptos-labs/aptos-core/pull/10200

ecosystem/typescript/sdk_v2/src/core/account.ts
0xjinn: Hmm, "Retrieves" feels weird to me. I can add a note to indicate - This does not actually creates an account on-chain.

0xmaayan: yea.. well nothing really creates account on chain lol
Maybe something like derive an account from a private key?

0xjinn: `Derive an account` - I like it!

ecosystem/typescript/sdk_v2/src/core/account.ts
0xjinn: Updated to args to be an abstract class of `PrivateKey`

ecosystem/typescript/sdk_v2/src/core/account.ts
```diff
@@ -0,0 +1,168 @@
...
+ *
+ * Use this class to create accounts, sign transactions, and more.
+ * Note: Creating an account instance does not create the account onchain.
+ */
+export class Account {
+  /**
+   * A private key and public key, associated with the given account
+   */
+  readonly publicKey: PublicKey;
+  readonly privateKey: PrivateKey;
```
gregnazario, line 26: Note the private key doesn't apply to the MultiEd25519.

This is a little weird where we'd probably want the ability to have an account that is either:

1. A single key account -> which can sign directly there with the private key
2. A non-signable multi-sig account -> which you can verify, but not sign, and won't have a private key.
3. Future will have passkeys or other types of accounts, which also may not have all the private information directly.

Additionally, for generated docs, let's try to put the comments above each item individually.

Let's roll with this for right now, but we need to clean this up probably mid next week to ensure there's a way to handle signatures with non single private key accounts.

ecosystem/typescript/sdk_v2/src/core/account.ts
```diff
@@ -0,0 +1,168 @@
...
+   * TODO: This constructor uses the nacl library directly, which only works with ed25519 keys.
+   * Need to update this to use the new crypto library if new schemes are added.
+   *
+   * @param args.privateKey PrivateKey - private key of the account
+   * @param args.address AccountAddress - address of the account
+   *
+   * This method is private because it should only be called by the factory static methods.
+   * @returns Account
+   */
+  private constructor(args: { privateKey: PrivateKey; address: AccountAddress }) {
```
0xjinn, line 45: I am thinking of modifying this to 

`private constructor(args: { privateKey: PrivateKey; publicKey: PublicKey; address: AccountAddress })`

So we don't have to worry about schemes. All properties will just get assigned directly.

gregnazario: We should probably add a PrivateKey -> PublicKey function on the PrivateKey.  The great thing about asymmetrical key cryptography, is the public key should be derivable from the private key.

You will want to add an optional publicKey though for MultiEd25519

ecosystem/typescript/sdk_v2/src/core/account.ts
```diff
@@ -0,0 +1,168 @@
...
+    const { privateKey, address } = args;
+
+    // Derive the public key from the private key
+    const keyPair = nacl.sign.keyPair.fromSeed(privateKey.toUint8Array().slice(0, 32));
+    this.publicKey = new Ed25519PublicKey({ hexInput: keyPair.publicKey });
```
gregnazario, line 50: Yeah, this is still tied to Ed25519. Can we add the derive function?

ecosystem/typescript/sdk_v2/src/core/account.ts
```diff
@@ -0,0 +1,168 @@
...
+   * Derives an account with random private key and address
+   *
+   * @returns Account
+   */
+  static generate(): Account {
+    const keyPair = nacl.sign.keyPair();
+    const privateKey = new Ed25519PrivateKey({ value: keyPair.secretKey.slice(0, 32) });
+    const address = new AccountAddress({ data: Account.authKey({ publicKey: keyPair.publicKey }).toUint8Array() });
+    return new Account({ privateKey, address });
+  }
```
gregnazario, line 66: We should probably change this to either `generate_ed25519` or we can take in some sort of option that provides a scheme.

ecosystem/typescript/sdk_v2/src/core/account.ts
```diff
@@ -0,0 +1,168 @@
...
+   * @param args.privateKey Hex - private key of the account
+   * @param args.address AccountAddress - address of the account
+   * @returns Account
+   */
+  static fromPrivateKeyAndAddress(args: { privateKey: HexInput; address: AccountAddress }): Account {
+    const privatekeyHex = Hex.fromHexInput({ hexInput: args.privateKey });
+    const signingKey = nacl.sign.keyPair.fromSeed(privatekeyHex.toUint8Array().slice(0, 32));
+    const privateKey = new Ed25519PrivateKey({ value: signingKey.secretKey.slice(0, 32) });
+    return new Account({ privateKey, address: args.address });
+  }
```
gregnazario, line 95: These can probably be a single one with an optional address (unless we had direct feedback that it was confusing).

Additionally, this would be prime for the derive public key, and all.

ecosystem/typescript/sdk_v2/src/core/account.ts
```diff
@@ -0,0 +1,168 @@
...
+
+  /**
+   * Derives an account with bip44 path and mnemonics,
+   *
+   * @param path. (e.g. m/44'/637'/0'/0'/0')
+   * Detailed description: {@link https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki}
+   * @param mnemonics.
+   * @returns AptosAccount
+   */
```
gregnazario, line 104: Should mention this is for ed25519, or we rename it somehow.

ecosystem/typescript/sdk_v2/src/crypto/authentication_key.ts
```diff
@@ -0,0 +1,104 @@
...
+ */
+export class AuthenticationKey {
+  // Length of AuthenticationKey in bytes(Uint8Array)
+  static readonly LENGTH: number = 32;
+
+  // Scheme identifier for MultiEd25519 signatures used to derive authentication keys for MultiEd25519 public keys
+  static readonly MULTI_ED25519_SCHEME: number = 1;
+
+  // Scheme identifier for Ed25519 signatures used to derive authentication key for MultiEd25519 public key
+  static readonly ED25519_SCHEME: number = 0;
```
gregnazario, line 27: Note that a full list of schemes is here: https://github.com/aptos-labs/aptos-core/blob/c2c8ce30f39c6bf1e69b4db01d47c64e6c9a3cfd/types/src/transaction/authenticator.rs#L349-L365

I'd probably put them in numerical order

ecosystem/typescript/sdk_v2/src/crypto/authentication_key.ts
```diff
@@ -0,0 +1,104 @@
...
+  static fromPublicKey(args: { publicKey: PublicKey }): AuthenticationKey {
+    const { publicKey } = args;
+    const pubKeyBytes = publicKey.toUint8Array();
+
+    const bytes = new Uint8Array(pubKeyBytes.length + 1);
+    bytes.set(pubKeyBytes);
+    bytes.set([AuthenticationKey.ED25519_SCHEME], pubKeyBytes.length);
+
+    const hash = sha3Hash.create();
+    hash.update(bytes);
```
gregnazario, line 90: Nit, could reduce duplicate code to a separate function with a different scheme, since all auth keys are derived the same way.

gregnazario (approved):
It's still pretty tied to single Ed25519, but I would say to unblock the rest of testing, we'll get it through like this.

We should follow up and fix it to be less specific

ecosystem/typescript/sdk_v2/tests/unit/authentication_key.test.ts
```diff
@@ -0,0 +1,49 @@
...
+    });
+
+    const authKey = AuthenticationKey.fromMultiPublicKey({ multiPublicKey: pubKeyMultiSig });
+    expect(authKey).toBeInstanceOf(AuthenticationKey);
+    expect(authKey.data.toString()).toEqual("0xa81cfac3df59920593ff417b45fc347ead3d88f8e25112c0488d34d7c9eb20af");
+  });
+
+  it("should derive an AccountAddress from AuthenticationKey with same string", () => {
+    const authKey = new AuthenticationKey({ data: ed25519.authKey });
+    const accountAddress = authKey.derivedAddress();
```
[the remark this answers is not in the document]
davidiw: It is accountAddress though... my reading was the derivedAddress was due to it being a portion of the auth key. Mind you, it isn't saying that public key is at that account address, it is just the value computed as an account address.

ecosystem/typescript/sdk_v2/src/core/account.ts
0xjinn: Yeah, this was the part where I spent the most time thinking on how to optimize the design, in order to be less confusing and complex.

At the end, I feel like if we want to accommodate all the scenarios, this class might start to get a bit complicate very quickly.

ecosystem/typescript/sdk_v2/src/core/account.ts
0xjinn: If this is multisig account, then we probably don't have private key and cannot derive the public.

Maybe I will just have a optional privateKey in this case.

`private constructor(args: { privateKey?: PrivateKey; publicKey: PublicKey; address: AccountAddress })`

ecosystem/typescript/sdk_v2/src/core/account.ts
0xjinn: Good one! I think taking in a scheme make sense. And probably default to ed25519 if not provided.

ecosystem/typescript/sdk_v2/src/crypto/authentication_key.ts
0xjinn: I will add a scheme enum to TS SDK.

0xjinn:
> It's still pretty tied to single Ed25519, but I would say to unblock the rest of testing, we'll get it through like this.
> 
> We should follow up and fix it to be less specific

Yeah, I believe many of the Phase 1 stuff are depending on the Account class. Lets get this in. I will have another PR to patch and fix all of the remaining comments.

gregnazario:
This looks like an unrelated failure.  I'm ninja landing this

PR #10214, opened 2023-09-25
[CLI] Fix faucet component of local testnet without --force-restart

banool:
### Description
In 4e8bbcc5a109f57af22aeb7ca8f0af70b675ce4e I made it possible to provide a config directly to `setup_test_environment_and_start_node`. Unfortunately I didn't handle this properly in the caller, in that I made it pass in a config unconditionally, even if one already existed on disk. 

Back then I said an overhaul was warranted but no one had time. Well in this PR I do a mini overhaul. Mostly this is just splitting up the big mega functions we had into their constituent pieces. This allows the caller the flexibility to load up a config, customize it, and then run a node with it.

### Test Plan
Previously these two commands one after the other:
```
cargo run -p aptos -- node run-local-testnet --assume-yes --force-restart
cargo run -p aptos -- node run-local-testnet --assume-yes
```

Would result in this error:
```
Faucet stopped unexpectedly Err(
    Error {
        context: "Failed to build Funder",
        source: Error {
            context: "Failed to build MintFunder",
            source: Error {
                context: "Failed to make MintFunder use delegated account",
                source: Error {
                    context: "Failed to create new account",
                    source: AptosTapError {
                        message: "API error Error(VmError): Invalid transaction: Type: Validation Code: INVALID_AUTH_KEY",
                        error_code: TransactionFailed,
                        rejection_reasons: [],
                        txn_hashes: [],
                    },
                },
            },
        },
    },
)
```

Now it does not.


0xmaayan (approved):
noice

banool:
Okay I'm not happy with this, I'm gonna try something else.

banool:
Alright refactored a bit, this is better.

banool:
I added another separate commit to this PR that addresses the feedback from Greg on https://github.com/aptos-labs/aptos-core/pull/10101.

crates/aptos/CHANGELOG.md
```diff
@@ -9,7 +9,7 @@ All notable changes to the Aptos CLI will be captured in this file. This project
 
 ### Updated
 - The `--with-faucet` flag has been removed from `aptos node run-local-testnet`, we now run a faucet by default. To disable the faucet use the `--no-faucet` flag.
-- When using `aptos node run-local-testnet` we now expose a transaction stream. Learn more about the transaction stream service here: https://aptos.dev/indexer/txn-stream/. Opt out of this with `--no-txn-stream`.
+- **Breaking change**: When using `aptos node run-local-testnet` we now expose a transaction stream. Learn more about the transaction stream service here: https://aptos.dev/indexer/txn-stream/. Opt out of this with `--no-txn-stream`. This is marked as a breaking change since the CLI now uses a port (50051 by default) that it didn't used to. If you need this port, you can tell the CLI to use a different port with `--txn-stream-port`.
```
0xjinn, line 12: New port is being used now. Does this break our CLI e2e? (Can't remember if we hard coded the port in e2e)

aptos-node/src/lib.rs
```diff
@@ -318,6 +316,57 @@ where
...
+///   a template when building the final config. If not provided, a default template
+///   will be used. Many overrides are applied on top of this base config.
+/// - `config_path` is similar to `test_config_override_path`, but many of the
+///   overrides that are applied when using `test_config_override_path` are not
+///   applied when using `config_path`. Read the code for more info.
+/// - `config` is a complete NodeConfig. No overrides are applied on top of this if
+///    it is provided. If both `config` and `test_dir` are provided, `config` takes
+///    precedence.
+/// - `test_dir` is a directory that contains a config file. Much like `config`, the
+///   config read from this file is used without any overrides.
```
0xjinn, line 332: Trying wrap my head around this...

Not sure if I read the code right, but If both `config_path` and `test_dir` are provided, the config from `test_dir` is ignored? Should we use the one from `test_dir` if provided?

0xjinn (commented):
Shall we plan a new CLI release this week Thursday 9/28?

crates/aptos/CHANGELOG.md
banool: This is an additional new port, all the existing stuff is on the same port.

aptos-node/src/lib.rs
banool: Actually the other way around, if `test_dir` is provided then `config_path` is ignored. This is a pretty awful function, I'm just trying to document it, an enum would be better here.

0xjinn: I see! Yeah probably a good `ops week` item in Q4. Lets get this fix and do a release soon to unblock 😃 

nit: Intuitively, I feel like if `test_dir` is provided, I would assume `test_dir` is used and overrides everything else, including `config`

0xjinn (approved):

aptos-node/src/lib.rs
banool: Yeah I went back and forth on that. In reality this is just not good code and Rust has enums to solve this problem.

PR #9700, opened 2023-08-21
[Spec] Ensures for stake.move

UIZorrot:
### Description

This is the high-level properties and ensures for stake.move module

### Test Plan

aptos move prove -f stake

PR #10215, opened 2023-09-25
[Spec] Fix spec

rahxephon89:
### Description

Fix the spec of the governance and stake module. All high-level requirements in the stake module and the first two high-level requirements in the staking_contract module are proved.

### Test Plan

aptos move prove

### DIFF
--- a/aptos-move/framework/aptos-framework/doc/aptos_governance.md
+++ b/aptos-move/framework/aptos-framework/doc/aptos_governance.md
@@ -2064,79 +2064,6 @@ The same as spec of <code><a href="aptos_governance.md#0x1_aptos_governance_crea
 </code></pre>
 
 
-<code>stake_pool</code> must exist StakePool.
-The delegated voter under the resource StakePool of the stake_pool must be the proposer address.
-Address @aptos_framework must exist GovernanceEvents.
-
-
-<a name="0x1_aptos_governance_CreateProposalAbortsIf"></a>
-
-
-<pre><code><b>schema</b> <a href="aptos_governance.md#0x1_aptos_governance_CreateProposalAbortsIf">CreateProposalAbortsIf</a> {
-    proposer: &<a href="../../aptos-stdlib/../move-stdlib/doc/signer.md#0x1_signer">signer</a>;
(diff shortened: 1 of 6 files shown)


aptos-move/framework/aptos-framework/doc/aptos_governance.md
```diff
@@ -2064,79 +2064,6 @@ The same as spec of <code><a href="aptos_governance.md#0x1_aptos_governance_crea
 </code></pre>
 
 
-<code>stake_pool</code> must exist StakePool.
```
rahxephon89, line 2067: This was introduced in this PR: https://github.com/aptos-labs/aptos-core/pull/9700 by setting `verify` to false for the toggle_features function.

aptos-move/framework/aptos-framework/sources/staking_contract.spec.move
```diff
@@ -385,24 +386,23 @@ spec aptos_framework::staking_contract {
...
-        // verify create_stake_pool()
-        let seed_0 = bcs::to_bytes(staker_address);
-        let seed_1 = concat(concat(concat(seed_0, bcs::to_bytes(operator)), SALT), contract_creation_seed);
-        let resource_addr = account::spec_create_resource_address(staker_address, seed_1);
-        include CreateStakePoolAbortsIf {resource_addr};
-
-        // verify stake::add_stake_with_cap()
-        include StakeAddStakeWithCapAbortsIf{ pool_address: resource_addr };
-
+        //
```
junkil-park, line 389: Is this comment block intended?

rahxephon89: Yes. I need to verify other two high-level properties defined in it so they are commented out because of timeout.

junkil-park (approved):

aptos-move/framework/aptos-framework/sources/staking_contract.spec.move
junkil-park: I see. Could we put a TODO together with this reason for future reference?

aalok-t (approved):

PR #10216, opened 2023-09-25
[TS SDK V2] Add Indexer account queries

0xmaayan:
### Description
Add indexer account queries. Since it is the first PR that adds Indexer queries into sdk v2, this PR includes the structure and design for handling Indexer/graphql queries.

In sdk_v2 we use generation tool to generate types based on graphql schema. Then we use custom types derived from the generated types to provide a better developer experience and to reduce the final package size.

Generation flow:
1. Write gql query in .graphql files - that should help with syntax, format and support future types of operations like subscription, mutation, etc.
2. run `pnpm run indexer-codegen`
3. Generate types and ts consts based on the .graphql queries. 
4. Use the ts consts as the query in the code and the types as the return and query operation types (such as orderBy, in the future filtering)

Note that we have 2 kinds of response types.
1. Generated types for when the sdk queries graphql. In this flow, the response we are getting back includes some top level keys (like `current_token_ownerships_v2`) - this is by the api design. therefore the query type would look like
```
export type GetAccountCoinsCountQuery = { 
current_fungible_asset_balances_aggregate: { 
 aggregate?: { 
  count: number } | null 
 } 
};
```
5. Custom response types for when the developer/user calls an sdk api function. In this flow, the function executes the actual graphql query (from step 1) and to support good dev exp, the return function type would be a more friendly type
```
export type GetAccountCoinsCountQueryResponse = GetAccountCoinsCountQuery["current_fungible_asset_balances_aggregate"]
```
That way, the end user only gets the actual data and adds support to `orderBy` the type keys (with IDE autocomplete support) by the values supported in Hasura Indexer

Screenshot 2023-09-25 at 3 46 11 PM

Screenshot 2023-09-25 at 3 46 19 PM

--------

We only export the actual types we use in the SDK (mentioned above) to keep the package build size small.

-------- 
NOTE: indexer api design is expected to change in the future and we would probably end up doing some refactoring.

### Test Plan
tested locally, but since indexer doesnt have a local development env, we need to submit some transaction during tests and thats can only be done once we have transaction submission support in sdk v2


ecosystem/typescript/sdk_v2/src/types/index.ts
```diff
@@ -83,6 +85,7 @@ export type AptosRequest = {
   params?: Record<string, string | AnyNumber | boolean | undefined>;
   originMethod?: string;
   overrides?: ClientConfig;
+  isIndexer?: boolean;
```
gregnazario, line 88: We may want an enum type rather than `isIndexer` to allow for future adaptability

ecosystem/typescript/sdk_v2/src/internal/queries/getAccountTransactionAcount.ts
gregnazario, line 1: File is misspelled

ecosystem/typescript/sdk_v2/src/internal/queries/getAccountTokensAcount.ts
gregnazario, line 1: Also, misspelled

ecosystem/typescript/sdk_v2/src/internal/queries/getAccountOwnedTokensByTokenData.ts
gregnazario, line 1: Curious since I'm not a TS developer natively.

Are the file names usually `PascalCase` or `camelCase` to match the constant which is `PascalCase`?

ecosystem/typescript/sdk_v2/src/internal/account.ts
```diff
@@ -121,3 +147,268 @@ export async function getResource(args: {
...
+  const { data } = await post<{}, GetAccountTokensCountQuery>({
+    url: aptosConfig.getRequestUrl(AptosApiType.INDEXER),
+    body: {
+      query: GetAccountTokensCount,
+      variables: { where_condition: whereCondition },
+    },
+    originMethod: "getAccountTokensCount",
+    overrides: { WITH_CREDENTIALS: false, ...aptosConfig.clientConfig },
+    isIndexer: true,
+  });
```
gregnazario, line 173: Let's make this into a helper function, since it looks like you're doing it slightly differently per each call

```
queryIndexer<ReturnType>(
{
  query: Query,
  variables: {},
  name: "QueryName" (Is this required?  What if it differs?  What purpose does it serve?)
  overrides?: {}
}
)

```

We may want to do similarly with the get and post queries that we have for the Fullnode APIs, just to remove duplicate code, and make it easier to read.

gregnazario (commented):
Let's see if we can consolidate some of the duplicate code around querying the indexer, it'll make it easier to review, and less error prone.

I would also use an enum or a different type to identify the destination or source of the API, since indexer is what we have now, but it could be something else in the future.

ecosystem/typescript/sdk_v2/src/api/account.ts
```diff
@@ -140,4 +166,136 @@ export class Account {
...
   }
+
+  /**
+   * Queries the count of tokens owned by an account
+   *
+   * @param accountAddress The account address
+   * @returns an object { count : number }
+   */
+  async getAccountTokensCount(args: { accountAddress: HexInput }): Promise<GetAccountTokensCountQueryResponse> {
+    const count = await getAccountTokensCount({ aptosConfig: this.config, ...args });
```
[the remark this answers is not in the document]
0xmaayan: yes.. this is because `await getAccountTokensCount` is imported from a file path `internal/account` and `async getAccountTokensCount` is a member class. if we were to do `await this.getAccountTokensCount` then we would end up with a loop =)

ecosystem/typescript/sdk_v2/src/types/index.ts
0xmaayan: I agree it is a bit weird. At the end, the question whether `is indexer` depends on the url we use to query. I was trying to define `isAptosIndexer()` method in `aptos_config` and then call it in the `client/core.ts` file, but ended up with cycle dependency. 
What enum have you had in mind?

ecosystem/typescript/sdk_v2/src/internal/queries/getAccountOwnedTokensByTokenData.ts
0xmaayan: yep, usually `camelCase`.
In react/vue/and others we mostly see components/pages files as `PascalCase`. 
JS community is wild

ecosystem/typescript/sdk_v2/src/internal/account.ts
0xmaayan: we should probably have it in the sdk for users to be able to query whatever query they want, might as well add it here

banool (commented):
> As we decided to step away from generation tools, in sdk v2 we write the graphql queries and the query response types by ourselves.

My understanding was we only decided to move away from generation of clients, not types. Writing these response types by hand seems like it's going to become a maintenance nightmare. What's wrong with codegen for this part?

0xmaayan:
> > As we decided to step away from generation tools, in sdk v2 we write the graphql queries and the query response types by ourselves.
> 
> My understanding was we only decided to move away from generation of clients, not types. Writing these response types by hand seems like it's going to become a maintenance nightmare. What's wrong with codegen for this part?

I was self debating on whether we should generate the types while developing. Decided to go with custom types because
1. dev exp. the types we currently provide in the sdk ([for example](https://github.com/aptos-labs/aptos-core/blob/main/ecosystem/typescript/sdk/src/indexer/generated/types.ts#L3697)) are just confusing.
2. the return type and the options types the end user sees when calling an indexer function ([for example](https://github.com/aptos-labs/aptos-core/blob/main/ecosystem/typescript/sdk/src/providers/indexer.ts#L367-L369C14)) that don't mean much. 
3. the final package size, exporting all generated types as is increases the size in about 2MB.

But I do agree we should use the help of the generation tool for developing because writing it ourselves is too much and once we have all of our APIs port over to graphql it would become a mess.

I think I found a way to generate the types to use while developing and types to be exported for the end user. Let me update this PR.

ecosystem/typescript/sdk_v2/src/types/index.ts
heliuchuan: Yeah seems less than ideal to add a param just to represent if it's the indexer even though that information is somewhat implicitly encoded in the url.

I'm guessing what Greg is suggesting is to pass in the `AptosApiType` and core.ts could gate on that to figure out how to handle the response?  It still would be semi-redundant but more adaptable than `isIndexer`

It would be nice if `aptosRequest` could derive the url and know how to handle the response via AptosApiType but you'd have to pass in the config I'm guessing which seems meh.  Not aware of the feasibility of that given the dependency graph.

I'll also suggest maybe instead adding an optional param to `AptosRequest` that is something like
`responseHandler?: (response: any) => AptosResponse<Req, Res>`
as a custom response handler that the query indexer can pass in?

This would allow for more adaptability to handle responses but idk if that bloats `AptosRequest` much.

just thought, I'm new here :)

ecosystem/typescript/sdk_v2/tests/e2e/api/general.test.ts
```diff
@@ -20,7 +20,7 @@ describe("general api", () => {
   test("it fetches block data by block height", async () => {
     const config = new AptosConfig({ network: Network.LOCAL });
     const aptos = new Aptos(config);
-    const blockHeight = 100;
+    const blockHeight = 1;
```
heliuchuan, line 23: is this supposed to be here?  just checking because I don't have enough context to understand changing 100 -> 1

heliuchuan (commented):
mostly nits with suggested changes

ecosystem/typescript/sdk_v2/src/api/account.ts
```diff
@@ -143,4 +169,136 @@ export class Account {
...
+    };
+  }): Promise<GetAccountOwnedTokensFromCollectionResponse> {
+    const tokens = await getAccountOwnedTokensFromCollectionAddress({ aptosConfig: this.config, ...args });
+    return tokens;
+  }
+
+  /**
+   * Queries for all collections that an account has tokens for.
+   *
+   * This query returns all tokens (v1 and v2 standards) an account owns, including NFTs, fungible, soulbound, etc.
```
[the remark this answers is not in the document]
0xmaayan: yes, you can query for all collections with token v1 or v2 - can see it on [hasura playground](https://cloud.hasura.io/public/graphiql?endpoint=https%3A%2F%2Findexer-devnet.staging.gcp.aptosdev.com%2Fv1%2Fgraphql&query=query+getAccountCollectionsWithOwnedTokens+%7B%0A++current_collection_ownership_v2_view%28%0A++++where%3A+%7Bowner_address%3A+%7B_eq%3A+%220xf3888a4ff79a592bf70e8aef4c751c612c69d182d06f9c91de49f1b40bf97e5c%22%7D%2C+current_collection%3A+%7Btoken_standard%3A+%7B_eq%3A+%22v2%22%7D%7D%7D%0A++%29+%7B%0A++++current_collection+%7B%0A++++++collection_id%0A++++++collection_name%0A++++++creator_address%0A++++++current_supply%0A++++++description%0A++++++last_transaction_timestamp%0A++++++last_transaction_version%0A++++++mutable_description%0A++++++max_supply%0A++++++mutable_uri%0A++++++table_handle_v1%0A++++++token_standard%0A++++++total_minted_v2%0A++++++uri%0A++++%7D%0A++++collection_id%0A++++collection_name%0A++++collection_uri%0A++++creator_address%0A++++distinct_tokens%0A++++last_transaction_version%0A++++owner_address%0A++++single_token_uri%0A++%7D%0A%7D%0A)

ecosystem/typescript/sdk_v2/tests/e2e/api/general.test.ts
0xmaayan: yes it is just a test for the get block height query. block height = 100 might be to high for example when you start a local node, block height hasn't reach that number yet. so changing to 1. can read more here https://aptos.dev/reference/glossary/#version

0xmaayan:
@gregnazario @heliuchuan agree with both of about the `isIndexer`. I changed it so that the downstream aptosRequest would accept the aptosConfig as a param and added a function to check whether it is an indexer query. This aptosConfig param would also help us when we start support LogService in the sdk.

heliuchuan (approved):
stamping, i'll let dport/greg give ya the final approval

ecosystem/typescript/sdk_v2/src/client/core.ts
```diff
@@ -67,10 +72,19 @@ export async function aptosRequest<Req, Res>(options: AptosRequest): Promise<Apt
     url: fullEndpoint,
   };
 
+  // to support both fullnode and indexer responses,
+  // check if it is an indexer query, and adjust response.data
+  if (aptosConfig.isIndexerRequest(url)) {
+    // errors from indexer
+    if ((result.data as any).errors) {
+      throw new AptosApiError(options, result, response.data.errors[0].message ?? "Generic Error");
```
gregnazario, line 80: We should probably find a better way than generic error, it's very confusing.  Let's follow up in a future PR

gregnazario (approved):
I may come back and do some refactoring, but otherwise it looks good

ecosystem/typescript/sdk_v2/src/client/core.ts
0xmaayan: you mean the "Generic Error"? tbh, I don't think it should even get to there - this is for the very rare case where the response doesnt have a message prop, which means something weird happened lol

PR #10217, opened 2023-09-25
[Aptos Data Client] Improve peer selection for data retrieval

JoshLind:
Note: (i) this PR is built on https://github.com/aptos-labs/aptos-core/pull/10180; and (ii) most of the PR is test changes.

### Description
This PR improves peer selection in the Aptos Data Client. Specifically, it adds the following logic:
1. Update peer selection for optimistic fetch to select the lowest latency peer when sending the request.
2. Update peer selection for subscription syncing to select the lowest latency peer when creating the subscription stream.
3. Update peer selection for generic data fetches (e.g., transactions and output chunks) to randomly select peers weighted by latency. This helps to prioritize lower latency peers while load balancing. In some cases (e.g., when we have a lot of peers to choose from) peer filtering is also applied, to help avoid picking the slowest peers.

Today, 1, 2 and 3 are currently selecting peers randomly.

This PR offers two commits:
1. Add the new peer selection logic.
2. Update the existing tests and add tests for the new functionality.

### Test Plan
New and existing tests.


config/src/config/state_sync_config.rs
```diff
@@ -332,8 +339,10 @@ impl Default for AptosDataClientConfig {
             max_subscription_lag_secs: 30, // 30 seconds
             max_transaction_chunk_size: MAX_TRANSACTION_CHUNK_SIZE,
             max_transaction_output_chunk_size: MAX_TRANSACTION_OUTPUT_CHUNK_SIZE,
-            optimistic_fetch_timeout_ms: 5000,        // 5 seconds
-            response_timeout_ms: 10_000,              // 10 seconds
+            min_peer_ratio_for_latency_filtering: 5, // Only filter if we have at least 5 potential peers per request
+            min_peers_for_latency_filtering: 10, // Only filter if we have at least 10 total peers
```
bchocho, line 343: How many peers do our mainnet PFNs have?

bchocho (approved):
LGTM. I wonder if there are any metrics that would help debug things if say, state sync is slow for a particular PFN.

config/src/config/state_sync_config.rs
JoshLind: They are currently configured to 4 😄 (That's been the default for a while). But, we should probably revisit that now that things are changing.

bchocho: Somewhat orthogonal to this PR, but I'm concerned that with our 7 peer forge networks, we don't really exercise the numbers that mainnet would see.

perryjrandall (approved):

PR #10218, opened 2023-09-25
[Network] Reduce lock contention in the peers and metadata struct.

JoshLind:
Note: this PR is built on: https://github.com/aptos-labs/aptos-core/pull/10217

### Description
This PR offers two modifications to the peers and metadata struct (to help reduce read lock contention). Each modification is offered in its own commit:
1. Replace `Arc<RwLock<PeerSet>>>` with `Arc<ArcSwap<PeerSet>>>` in the `trusted_peers` set. Given that the trusted peers will be read frequently, but changed infrequently (e.g., during epoch changes), it makes sense to optimize these for reads.
2. To reduce lock contention on `peers_and_metadata`, we add a cached copy of the data to service all reads. The cached copy is updated every time `peers_and_metadata` is updated (which happens semi-frequently). The benefit of this approach is that reads do not need to wait for any locks. This is crucial because we expect `peers_and_metadata` to be read very frequently (e.g., with the new peer monitoring service optimizations).

### Test Plan
New and existing tests.

### DIFF
--- a/network/framework/src/connectivity_manager/mod.rs
+++ b/network/framework/src/connectivity_manager/mod.rs
@@ -39,7 +39,6 @@ use aptos_config::{
     network_id::NetworkContext,
 };
 use aptos_crypto::x25519;
-use aptos_infallible::RwLock;
 use aptos_logger::prelude::*;
 use aptos_netcore::transport::ConnectionOrigin;
 use aptos_num_variants::NumVariants;
@@ -59,7 +58,7 @@ use serde::Serialize;
 use std::{
     cmp::{min, Ordering},
     collections::{hash_map::Entry, HashMap, HashSet},
-    fmt, mem,
+    fmt,
     sync::Arc,
     time::{Duration, SystemTime},
 };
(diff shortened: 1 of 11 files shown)


bchocho (approved):
LGTM

network/framework/src/connectivity_manager/mod.rs
```diff
@@ -810,13 +807,16 @@ where
...
-            // the write lock.
-            if let Some(trusted_peers) = self.get_trusted_peers() {
-                let _old_eligible = {
-                    let mut trusted_peers = trusted_peers.write();
-                    mem::replace(&mut *trusted_peers, new_eligible)
-                };
+            // Swap in the new eligible peers set
+            if let Err(error) = self
+                .peers_and_metadata
+                .set_trusted_peers(&self.network_context.network_id(), new_eligible)
```
brianolson, line 813: I like `set_trusted_peers()` a lot better than returning a mutable container

brianolson (approved):
the local cache inside PeersAndMetadata probably does a lot of the good without the API changes I'm planning for the network/framework2 code.

perryjrandall (approved):
beautiful tests <3

PR #10226, opened 2023-09-26
[Stake] Allows the delegation pool owner to update the operator commission percentage

junkil-park:
### Description

This PR adds the function to allow the delegation pool owner to update the operator commission percentage.

This implements the AIP-50: https://github.com/aptos-foundation/AIPs/pull/250

### Test Plan
aptos move test

### DIFF
--- a/aptos-move/framework/aptos-framework/sources/delegation_pool.move
+++ b/aptos-move/framework/aptos-framework/sources/delegation_pool.move
@@ -191,6 +191,16 @@ module aptos_framework::delegation_pool {
     /// Chaning beneficiaries for operators is not supported.
     const EOPERATOR_BENEFICIARY_CHANGE_NOT_SUPPORTED: u64 = 19;
 
+    /// Commission percentage increase is too large.
+    const ETOO_LARGE_COMMISSION_INCREASE: u64 = 20;
+
+    /// Commission percentage change is too late in this lockup period, and should be done at least a quarter (1/4) of the lockup duration before the lockup cycle ends.
+    const ETOO_LATE_COMMISSION_CHANGE: u64 = 21;
+
+    /// Changing operator commission rate in delegation pool is not supported.
+    const ECOMMISSION_RATE_CHANGE_NOT_SUPPORTED: u64 = 22;
+
+
     const MAX_U64: u64 = 18446744073709551615;
 
     /// Maximum operator percentage fee(of double digit precision): 22.85% is represented as 2285
@@ -214,6 +224,9 @@ module aptos_framework::delegation_pool {
 
     /// Scaling factor of shares pools used within the delegation pool
     const SHARES_SCALING_FACTOR: u64 = 10000000000000000;
+
+    /// Maximum commission percentage increase per lockup cycle. 10% is represented as 1000.
+    const MAX_COMMISSION_INCREASE: u64 = 1000;
 
     /// Capability that represents ownership over privileged operations on the underlying stake pool.
     struct DelegationPoolOwnership has key, store {
@@ -307,6 +320,11 @@ module aptos_framework::delegation_pool {
         beneficiary_for_operator: address,
     }
 
+    struct NextCommissionPercentage has key {
+        commission_percentage_next_lockup_cycle: u64,
+        effective_after_secs: u64,
+    }
+
     struct AddStakeEvent has drop, store {
         pool_address: address,
         delegator_address: address,
@@ -375,6 +393,13 @@ module aptos_framework::delegation_pool {
         new_beneficiary: address,
     }
 
+    #[event]
+    struct CommissionPercentageChange has drop, store {
+        pool_address: address,
+        owner: address,
+        commission_percentage_next_lockup_cycle: u64,
+    }
+
     #[view]
     /// Return whether supplied address `addr` is owner of a delegation pool.
     public fun owner_cap_exists(addr: address): bool {
@@ -408,10 +433,32 @@ module aptos_framework::delegation_pool {
     }
 
     #[view]
+    /// Return whether the commission percentage for the next lockup cycle is effective.
+    public fun is_next_commission_percentage_effective(pool_address: address): bool acquires NextCommissionPercentage {
+        exists<NextCommissionPercentage>(pool_address) &&
+            timestamp::now_seconds() >= borrow_global<NextCommissionPercentage>(pool_address).effective_after_secs
+    }
+
+    #[view]
     /// Return the operator commission percentage set on the delegation pool `pool_address`.
-    public fun operator_commission_percentage(pool_address: address): u64 acquires DelegationPool {
+    public fun operator_commission_percentage(pool_address: address): u64 acquires DelegationPool, NextCommissionPercentage {
         assert_delegation_pool_exists(pool_address);
-        borrow_global<DelegationPool>(pool_address).operator_commission_percentage
+        if (is_next_commission_percentage_effective(pool_address)) {
+            operator_commission_percentage_next_lockup_cycle(pool_address)
+        } else {
+            borrow_global<DelegationPool>(pool_address).operator_commission_percentage
+        }
+    }
+
+    #[view]
+    /// Return the operator commission percentage for the next lockup cycle.
+    public fun operator_commission_percentage_next_lockup_cycle(pool_address: address): u64 acquires DelegationPool, NextCommissionPercentage {
+        assert_delegation_pool_exists(pool_address);
+        if (exists<NextCommissionPercentage>(pool_address)) {
+            borrow_global<NextCommissionPercentage>(pool_address).commission_percentage_next_lockup_cycle
+        } else {
+            borrow_global<DelegationPool>(pool_address).operator_commission_percentage
+        }
     }
 
     #[view]
@@ -560,7 +607,7 @@ module aptos_framework::delegation_pool {
     #[view]
     /// Return the total voting power of a delegator in a delegation pool. This function syncs DelegationPool to the
     /// latest state.
-    public fun calculate_and_update_voter_total_voting_power(pool_address: address, voter: address): u64 acquires DelegationPool, GovernanceRecords, BeneficiaryForOperator {
+    public fun calculate_and_update_voter_total_voting_power(pool_address: address, voter: address): u64 acquires DelegationPool, GovernanceRecords, BeneficiaryForOperator, NextCommissionPercentage {
         assert_partial_governance_voting_enabled(pool_address);
         // Delegation pool need to be synced to explain rewards(which could change the coin amount) and
         // commission(which could cause share transfer).
@@ -574,7 +621,7 @@ module aptos_framework::delegation_pool {
     #[view]
     /// Return the remaining voting power of a delegator in a delegation pool on a proposal. This function syncs DelegationPool to the
     /// latest state.
-    public fun calculate_and_update_remaining_voting_power(pool_address: address, voter_address: address, proposal_id: u64): u64 acquires DelegationPool, GovernanceRecords, BeneficiaryForOperator {
+    public fun calculate_and_update_remaining_voting_power(pool_address: address, voter_address: address, proposal_id: u64): u64 acquires DelegationPool, GovernanceRecords, BeneficiaryForOperator, NextCommissionPercentage {
         assert_partial_governance_voting_enabled(pool_address);
         // If the whole stake pool has no voting power(e.g. it has already voted before partial
         // governance voting flag is enabled), the delegator also has no voting power.
@@ -607,6 +654,13 @@ module aptos_framework::delegation_pool {
         account::create_resource_address(&owner, seed)
     }
 
+    #[view]
+    /// Return the minimum remaining time in seconds for commission change, which is one fourth of the lockup duration.
+    public fun min_remaining_secs_for_commission_change(): u64 {
+        let config = staking_config::get();
+        staking_config::get_recurring_lockup_duration(&config) / 4
+    }
+
     /// Initialize a delegation pool of custom fixed `operator_commission_percentage`.
     /// A resource account is created from `owner` signer and its supplied `delegation_pool_creation_seed`
     /// to host the delegation pool resource and own the underlying stake pool.
@@ -615,7 +669,7 @@ module aptos_framework::delegation_pool {
         owner: &signer,
         operator_commission_percentage: u64,
         delegation_pool_creation_seed: vector<u8>,
-    ) acquires DelegationPool, GovernanceRecords, BeneficiaryForOperator {
+    ) acquires DelegationPool, GovernanceRecords, BeneficiaryForOperator, NextCommissionPercentage {
         assert!(features::delegation_pools_enabled(), error::invalid_state(EDELEGATION_POOLS_DISABLED));
         let owner_address = signer::address_of(owner);
         assert!(!owner_cap_exists(owner_address), error::already_exists(EOWNER_CAP_ALREADY_EXISTS));
@@ -676,7 +730,7 @@ module aptos_framework::delegation_pool {
     /// THe existing voter will be replaced. The function is permissionless.
     public entry fun enable_partial_governance_voting(
         pool_address: address,
-    ) acquires DelegationPool, GovernanceRecords, BeneficiaryForOperator {
+    ) acquires DelegationPool, GovernanceRecords, BeneficiaryForOperator, NextCommissionPercentage {
         assert!(features::partial_governance_voting_enabled(), error::invalid_state(EDISABLED_FUNCTION));
         assert!(features::delegation_pool_partial_governance_voting_enabled(), error::invalid_state(EDISABLED_FUNCTION));
         assert_delegation_pool_exists(pool_address);
@@ -705,7 +759,7 @@ module aptos_framework::delegation_pool {
     /// 2. The delegation pool's lockup period ends after the voting period of the proposal.
     /// 3. The voter still has spare voting power on this proposal.
     /// 4. The delegation pool never votes on the proposal before enabling partial governance voting.
-    public entry fun vote(voter: &signer, pool_address: address, proposal_id: u64, voting_power: u64, should_pass: bool) acquires DelegationPool, GovernanceRecords, BeneficiaryForOperator {
+    public entry fun vote(voter: &signer, pool_address: address, proposal_id: u64, voting_power: u64, should_pass: bool) acquires DelegationPool, GovernanceRecords, BeneficiaryForOperator, NextCommissionPercentage {
         assert_partial_governance_voting_enabled(pool_address);
         // synchronize delegation and stake pools before any user operation.
         synchronize_delegation_pool(pool_address);
@@ -748,7 +802,7 @@ module aptos_framework::delegation_pool {
         metadata_location: vector<u8>,
         metadata_hash: vector<u8>,
         is_multi_step_proposal: bool,
-    ) acquires DelegationPool, GovernanceRecords, BeneficiaryForOperator {
+    ) acquires DelegationPool, GovernanceRecords, BeneficiaryForOperator, NextCommissionPercentage {
         assert_partial_governance_voting_enabled(pool_address);
 
         // synchronize delegation and stake pools before any user operation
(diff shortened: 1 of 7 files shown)


aptos-move/framework/aptos-framework/sources/delegation_pool.move
```diff
@@ -183,6 +183,12 @@ module aptos_framework::delegation_pool {
     /// The stake pool has already voted on the proposal before enabling partial governance voting on this delegation pool.
     const EALREADY_VOTED_BEFORE_ENABLE_PARTIAL_VOTING: u64 = 17;
 
+    /// Commission percentage increase is too large.
+    const ETOO_LARGE_COMMISSION_INCREASE: u64 = 18;
+
+    /// Commission percentage change is too late in this lockup period, and should be done at least 7 days before the lockup cycle ends.
+    const ETOO_LATE_COMMISSION_CHANGE: u64 = 18;
```
movekevin, line 190: Should be 19 not 18

junkil-park: Oh, good catch! I just fixed it.

aptos-move/framework/aptos-framework/sources/delegation_pool.move
```diff
@@ -386,6 +409,17 @@ module aptos_framework::delegation_pool {
         borrow_global<DelegationPool>(pool_address).operator_commission_percentage
     }
 
+    #[view]
+    /// Return the operator commission percentage for the next lockup cycle.
+    public fun operator_commission_percentage_next_lockup_cycle(pool_address: address): u64 acquires DelegationPool, CommissionPercentage {
+        if(exists<CommissionPercentage>(pool_address)) {
```
movekevin, line 415: nit: Space after if

aptos-move/framework/aptos-framework/sources/delegation_pool.move
```diff
@@ -207,6 +213,12 @@ module aptos_framework::delegation_pool {
     /// Scaling factor of shares pools used within the delegation pool
     const SHARES_SCALING_FACTOR: u64 = 10000000000000000;
 
+    /// Maximum commission percentage increase per lockup cycle. 10% is represented as 1000.
+    const MAX_COMMISSION_INCREASE: u64 = 1000;
+
+    /// Minimum remaining time in seconds for commission change, which is 7 days where there are 604800 seconds.
+    const MIN_REMAINING_SECS_FOR_COMMISSION_CHANGE: u64 = 604800;
```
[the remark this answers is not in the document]
junkil-park: I see that it needs to different in the testnet and the mainnet. We may have the following options:
(1) `MIN_REMAINING_SECS_FOR_COMMISSION_CHANGE = recurring_lockup_duration_secs / 4`
(2) `MIN_REMAINING_SECS_FOR_COMMISSION_CHANGE = if (mainnet) { ... } else { ... }`
(3) The pool owner has permission to update `MIN_REMAINING_SECS_FOR_COMMISSION_CHANGE`
I think (3) is not secure. How about going with the option (1)?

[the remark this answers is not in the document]
movekevin: Option 1 seems pretty clean! - having the min remaining time required being a fraction of the total lockup.

aptos-move/framework/aptos-framework/sources/delegation_pool.move
```diff
@@ -1498,6 +1569,11 @@ module aptos_framework::delegation_pool {
                 pool.observed_lockup_cycle,
                 pool_u64::create_with_scaling_factor(SHARES_SCALING_FACTOR)
             );
+
+            // update the commission percentage for the new lockup cycle
+            if (exists<CommissionPercentage>(pool_address)) {
+                pool.operator_commission_percentage = borrow_global<CommissionPercentage>(pool_address).commission_percentage_next_lockup_cycle;
```
movekevin, line 1575: There's a slight problem here - if synchronize_delegation_pool is not called, the commission percentage would not be updated. This might not be too big of a deal as user operations such as staking would call synchronize_delegation_pool. However, the UI would still display the old commission rate before any actual user operations. Can you think a bit more on whether there are some options we can use to address this?

aptos-move/framework/aptos-framework/sources/delegation_pool.move
```diff
@@ -1498,6 +1569,11 @@ module aptos_framework::delegation_pool {
                 pool.observed_lockup_cycle,
                 pool_u64::create_with_scaling_factor(SHARES_SCALING_FACTOR)
             );
+
+            // update the commission percentage for the new lockup cycle
+            if (exists<CommissionPercentage>(pool_address)) {
```
alexfilip2, line 1574: Let's say that the commission has been updated during lockup cycle _N_ on the stake pool. After lockup cycle _N+1_ starts and until the 1st `synchronize_delegation_pool`, rewards are produced and will be shared with the operator using the old commission fee.
Additionally, the new commission fee takes effect when a new _observed_ lockup cycle is detected. not an actual lockup cycle (within the stake pool). To address these issues:

- owner calls `update_commission_percentage`
- owner `unlock`s some small amount of APT and waits for it to be inactivated - this ensures the observed lockup cycle advances together with the actual lockup cycle
- during the first epoch of the new lockup, owner calls `synchronize_delegation_pool` to ensure that all rewards produced starting this new lockup cycle are shared using the new commission

alexfilip2: Ensuring that the new commission is applied exactly when the next lockup starts cannot be achieved at the contract level because: rewards are produced continuously by the stake pool and summed up, but there is no callback that allows to snapshot state at the beginning of the lockup cycle.

aptos-move/framework/aptos-framework/sources/delegation_pool.move
```diff
@@ -956,19 +990,56 @@ module aptos_framework::delegation_pool {
...
+        // with the old commission percentage
+        synchronize_delegation_pool(pool_address);
+
+        let delegation_pool = borrow_global<DelegationPool>(pool_address);
+        let pool_signer = account::create_signer_with_capability(&delegation_pool.stake_pool_signer_cap);
+        if (exists<CommissionPercentage>(pool_address)) {
+            let commission_percentage = borrow_global_mut<CommissionPercentage>(pool_address);
+            commission_percentage.commission_percentage_next_lockup_cycle = new_commission_percentage;
+        } else {
+            move_to(&pool_signer, CommissionPercentage {commission_percentage_next_lockup_cycle: new_commission_percentage});
```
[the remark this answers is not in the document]
junkil-park: Done.

aptos-move/framework/aptos-framework/sources/delegation_pool.move
junkil-park: Done.

aptos-move/framework/aptos-framework/sources/delegation_pool.move
```diff
@@ -956,19 +994,58 @@ module aptos_framework::delegation_pool {
...
+            error::invalid_argument(ETOO_LARGE_COMMISSION_INCREASE)
+        );
+        assert!(
+            stake::get_remaining_lockup_secs(pool_address) >= min_remaining_secs_for_commission_change(),
+            error::invalid_state(ETOO_LATE_COMMISSION_CHANGE)
+        );
+
+        // synchronize delegation and stake pools before any user operation
+        // ensure the operator is paid its uncommitted commission rewards
+        // with the old commission percentage
```
alexfilip2, line 1023: +
// ensure the observed lockup cycle is up-to-date so that the new commission doesn't take effect this observed lockup, but rather the next one

aptos-move/framework/aptos-framework/sources/delegation_pool.move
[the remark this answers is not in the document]
junkil-park: Done!

aptos-move/framework/aptos-framework/sources/delegation_pool.move
```diff
@@ -956,19 +994,58 @@ module aptos_framework::delegation_pool {
...
     }
 
+    /// Allows an owner to update the commission percentage for the operator of the underlying stake pool.
+    public entry fun update_commission_percentage(
+        owner: &signer,
+        new_commission_percentage: u64
+    ) acquires DelegationPoolOwnership, DelegationPool, GovernanceRecords, CommissionPercentage {
+        assert!(new_commission_percentage <= MAX_FEE, error::invalid_argument(EINVALID_COMMISSION_PERCENTAGE));
+        let pool_address = get_owned_pool_address(signer::address_of(owner));
+        assert!(
```
alexfilip2, line 1012: Observation: UX changes whether `synchronize_delegation_pool` is called before or after this assert.
Scenario:
1. delegation pool has commission 7% since creation
2. owner sets commission to 7% + 10% at lockup cycle N
3. lockup N ends
4. no operation happened on the delegation pool yet in lockup N+1
6. owner wants to set commission to 7% + 10% + 10% to apply from lockup N+2 which is a valid commission update

- assert then `synchronize_delegation_pool`: tx fails with `ETOO_LARGE_COMMISSION_INCREASE`, but owner can call `synchronize_delegation_pool` separately which makes commission 17%, then try again and now 27% is accepted
- `synchronize_delegation_pool` then assert: commission is firstly updated to 17% and then 27% is accepted directly

aptos-move/framework/aptos-framework/sources/delegation_pool.move
alexfilip2: However, a snapshotting mechanism could be implemented at the stake pool level when the lockup cycle is renewed.
This would include the stake amounts on the validator at that time (or any other state that may be relevant to other modules). Unfortunately, multiple past snapshots should be stored, not just the last one, as the delegation pool might be idle for more than 1 lockup period.
This doesn't affect the integrity of the staking module, but it is not consistent with its overall implementation.

aptos-move/framework/aptos-framework/sources/delegation_pool.move
```diff
@@ -1571,6 +1667,16 @@ module aptos_framework::delegation_pool {
                 pool.observed_lockup_cycle,
                 pool_u64::create_with_scaling_factor(SHARES_SCALING_FACTOR)
             );
+        };
+
+        if (exists<NextCommissionPercentage>(pool_address)) {
+            let next_commission_percentage = borrow_global<NextCommissionPercentage>(pool_address);
+            if (timestamp::now_seconds() >= next_commission_percentage.effective_after_secs) {
+                // update the commission percentage for the new lockup cycle
+                pool.operator_commission_percentage = borrow_global<NextCommissionPercentage>(
```
alexfilip2, line 1676: use `next_commission_percentage` here

junkil-park: Updated. Couldn't use `operator_commission_percentage_next_lockup_cycle(pool_address)` which causes a dangling reference error because `pool` is already mutably borrowed.

aptos-move/framework/aptos-framework/sources/delegation_pool.move
junkil-park: I updated `operator_commission_percentage` to return the new commission rate if it's available:
```
    public fun operator_commission_percentage(pool_address: address): u64 acquires DelegationPool, NextCommissionPercentage {
        assert_delegation_pool_exists(pool_address);
        if (is_next_commission_percentage_effective(pool_address)) {
            operator_commission_percentage_next_lockup_cycle(pool_address)
        } else {
            borrow_global<DelegationPool>(pool_address).operator_commission_percentage
        }
    }
```
This provides a better UX for those who are adding a new stake to the pool.

aptos-move/framework/aptos-framework/sources/delegation_pool.move
junkil-park: > owner unlocks some small amount of APT and waits for it to be inactivated - this ensures the observed lockup cycle advances together with the actual lockup cycle

As the alternative that we discussed offline, we are using `effective_after_secs` (
https://github.com/aptos-labs/aptos-core/pull/10226/files#diff-072b3b0506ff0035c65fe82ead317e994c034d27f31ce21414d31a6675a7c40aR325)

aptos-move/framework/aptos-framework/sources/delegation_pool.move
```diff
@@ -1027,11 +1080,55 @@ module aptos_framework::delegation_pool {
...
+            stake::get_remaining_lockup_secs(pool_address) >= min_remaining_secs_for_commission_change(),
+            error::invalid_state(ETOO_LATE_COMMISSION_CHANGE)
+        );
+
+        // synchronize delegation and stake pools before any user operation
+        // ensure the operator is paid its uncommitted commission rewards
+        // with the old commission percentage
+        synchronize_delegation_pool(pool_address);
+
+        let delegation_pool = borrow_global<DelegationPool>(pool_address);
```
alexfilip2, line 1107: would move this resource read to the else branch

aptos-move/framework/aptos-framework/sources/delegation_pool.move
```diff
@@ -1027,11 +1080,55 @@ module aptos_framework::delegation_pool {
...
+        } else {
+            move_to(&pool_signer, NextCommissionPercentage {
+                commission_percentage_next_lockup_cycle: new_commission_percentage,
+                effective_after_secs: stake::get_lockup_secs(pool_address),
+            });
+        };
+
+        event::emit(CommissionPercentageChange {
+            pool_address,
+            operator: stake::get_operator(pool_address),
```
alexfilip2, line 1122: would log the pool owner instead of the operator, as that's the account initiating the commission change

aptos-move/framework/aptos-framework/sources/delegation_pool.move
junkil-park: > ensure the observed lockup cycle is up-to-date

@alexfilip2 , could you elaborate more on this ^

aptos-move/framework/aptos-framework/sources/delegation_pool.move
junkil-park: Now, there is no difference between the two options of different order because `operator_commission_percentage` returns the "effective" commission percentage.

aptos-move/framework/aptos-framework/sources/delegation_pool.move
```diff
@@ -191,6 +191,16 @@ module aptos_framework::delegation_pool {
     /// Chaning beneficiaries for operators is not supported.
     const EOPERATOR_BENEFICIARY_CHANGE_NOT_SUPPORTED: u64 = 19;
 
+    /// Commission percentage increase is too large.
+    const ETOO_LARGE_COMMISSION_INCREASE: u64 = 20;
+
+    /// Commission percentage change is too late in this lockup period, and should be done at least 7 days before the lockup cycle ends.
```
alexfilip2, line 197: update comment for 1/4 lockup duration instead of 7 days prior to lockup's end time

aptos-move/framework/aptos-framework/sources/delegation_pool.move
junkil-park: @alexfilip2 , instead of `operator` or in addition to `opreator`?

aptos-move/framework/aptos-framework/sources/delegation_pool.move
```diff
@@ -407,11 +432,32 @@ module aptos_framework::delegation_pool {
...
+        }
+    }
+
+    #[view]
+    /// Return the operator commission percentage for the next lockup cycle.
+    public fun operator_commission_percentage_next_lockup_cycle(pool_address: address): u64 acquires DelegationPool, NextCommissionPercentage {
+        if (exists<NextCommissionPercentage>(pool_address)) {
+            borrow_global<NextCommissionPercentage>(pool_address).commission_percentage_next_lockup_cycle
+        } else {
+            operator_commission_percentage(pool_address)
```
alexfilip2, line 459: safer and more explicit to use `borrow_global<DelegationPool>(pool_address).operator_commission_percentage` instead, as this looks like a recursion even it cannot happen due to the branching conditions

junkil-park: Done changing it. Good point!

aptos-move/framework/aptos-framework/sources/delegation_pool.move
junkil-park: Good catch! Done.

aptos-move/framework/aptos-framework/sources/delegation_pool.move
junkil-park: Done!

aptos-move/framework/aptos-framework/sources/delegation_pool.move
alexfilip2: This comment is outdated now that the next commission becomes effective based on the lockup cycle's ending time.
However, it is still important that `synchronize_delegation_pool` is called firstly in order to make sure any pending commission change is applied, then `NextCommissionPercentage` is overridden with a new value.

aptos-move/framework/aptos-framework/sources/delegation_pool.move
alexfilip2: _instead of operator_, logging the current operator at the time of the commission update doesn't give us any unique information. Operator/beneficiary can change too by the time this new commission value becomes effective and that would be tracked from other events.

aptos-move/framework/aptos-framework/sources/delegation_pool.move
```diff
@@ -407,11 +432,32 @@ module aptos_framework::delegation_pool {
...
+            operator_commission_percentage_next_lockup_cycle(pool_address)
+        } else {
+            borrow_global<DelegationPool>(pool_address).operator_commission_percentage
+        }
+    }
+
+    #[view]
+    /// Return the operator commission percentage for the next lockup cycle.
+    public fun operator_commission_percentage_next_lockup_cycle(pool_address: address): u64 acquires DelegationPool, NextCommissionPercentage {
+        if (exists<NextCommissionPercentage>(pool_address)) {
```
alexfilip2, line 456: `assert_delegation_pool_exists(pool_address);`

aptos-move/framework/aptos-framework/sources/delegation_pool.move
junkil-park: Replaced `operator` with `owner`.

aptos-move/framework/aptos-framework/sources/delegation_pool.move
junkil-park: Added!

movekevin (approved):
Can you also update documentation for this and make it clear that after the delegation pool owner changes the commission rate, at the end of that lockup cycle, they manually need to call synchronize to make sure the new commission rate is effective?

junkil-park:
> Can you also update documentation for this and make it clear that after the delegation pool owner changes the commission rate, at the end of that lockup cycle, they manually need to call synchronize to make sure the new commission rate is effective?

@movekevin , by "documentation", do you mean the delegation pool operation page in `aptos.dev`, right (https://aptos.dev/nodes/validator-node/operator/delegation-pool-operations)?

Do you think we need update the AIP-50 description as well (https://aptos.dev/nodes/validator-node/operator/delegation-pool-operations)?

aptos-move/framework/aptos-framework/sources/delegation_pool.move
```diff
@@ -191,6 +191,16 @@ module aptos_framework::delegation_pool {
     /// Chaning beneficiaries for operators is not supported.
     const EOPERATOR_BENEFICIARY_CHANGE_NOT_SUPPORTED: u64 = 19;
 
+    /// Commission percentage increase is too large.
+    const ETOO_LARGE_COMMISSION_INCREASE: u64 = 20;
+
+    /// Commission percentage change is too late in this lockup period, and should be done at least a quarter (1/4) of the lockup duration before the lockup cycle ends.
+    const ETOO_LATE_COMMISSION_CHANGE: u64 = 21;
+
+    /// Chaning operator commission rate in delegation pool is not supported.
```
lightmark, line 200: typo

lightmark (approved):

aptos-move/framework/aptos-framework/sources/delegation_pool.move
junkil-park: fixed.

PR #10251, opened 2023-09-26
[ts-sdk-v2] Add .npmrc to ensure pnpm lockfile is consistent

gregnazario:
### Description
I kept getting differing lockfiles, because pnpm is configured differently for me than whoever has been submitting commits.  This fixes that issue.

### Test Plan
Ran `pnpm install` and my lockfile didn't change format version


banool (approved):

0xjinn (approved):
Nice fix! 

I kept getting different lock file as well

PR #10254, opened 2023-09-26
[CLI] Add support for running standard indexer processors to the local testnet

banool:
### Stack
- Previous in stack: https://github.com/aptos-labs/aptos-core/pull/10271
- Next in stack: https://github.com/aptos-labs/aptos-core/pull/10310

### Description
This PR adds support for running indexer processors to the local testnet. The processors you run is configurable with `--processors`, which uses the new `ProcessorName` enum.

Note: I will update the dependency from that branch to a specific rev later.

### Test Plan
```
cargo run -p aptos -- node run-local-testnet --assume-yes --force-restart --with-indexer-api
```
```
Readiness endpoint: http://0.0.0.0:8070/

Postgres is starting, please wait...
Transaction stream is starting, please wait...
Node API is starting, please wait...
Faucet is starting, please wait...

Completed generating configuration:
        Log file: "/Users/dport/.aptos/testnet/validator.log"
        Test dir: "/Users/dport/.aptos/testnet"
        Aptos root key path: "/Users/dport/.aptos/testnet/mint.key"
        Waypoint: 0:2cc2b83a3435b6b38f5f8ef84afe96aef2458c524445d3d72cd55b0459e627c5
        ChainId: 4
        REST API endpoint: http://0.0.0.0:8080
        Metrics endpoint: http://0.0.0.0:9101/metrics
        Aptosnet fullnode network endpoint: /ip4/0.0.0.0/tcp/6181
        Indexer gRPC node stream endpoint: 0.0.0.0:50051

Aptos is running, press ctrl-c to exit

Node API is ready. Endpoint: http://0.0.0.0:8080/
Postgres is ready. Endpoint: postgres://postgres@127.0.0.1:5433/local_testnet
Transaction stream is ready. Endpoint: http://0.0.0.0:50051/
Faucet is ready. Endpoint: http://127.0.0.1:8081/

Applying post startup steps...

Setup is complete, you can now use the local testnet!
```

```
local_testnet-# \x
Expanded display is on.
local_testnet-# select * from events limit 1;
ERROR:  syntax error at or near "select"
LINE 2: select * from events limit 1
        ^
local_testnet=# select * from events limit 1;
-[ RECORD 1 ]------------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
sequence_number          | 3
creation_number          | 3
account_address          | 0x0000000000000000000000000000000000000000000000000000000000000001
transaction_version      | 4
[...]

PR #10255, opened 2023-09-26
[CLI] Add support for running an indexer API to the local testnet

banool:
### Stack
- Previous in stack: https://github.com/aptos-labs/aptos-core/pull/10310
- Next in stack: https://github.com/aptos-labs/aptos-core/pull/10311

### Description
This adds support for attaching an indexer API (Hasura) to the DB populated by the processors (support for which is added in https://github.com/aptos-labs/aptos-core/pull/10254).

As part of this I add the notion of shutdown hooks. These are best effort attempts to clean up state that doesn't get cleaned up automatically as part of process shutdown (the perils of Docker).

### Test Plan
```
cargo run --release -p aptos -- node run-local-testnet --assume-yes --force-restart --with-indexer-api
```
```
Readiness endpoint: http://0.0.0.0:8070/

Postgres is starting, please wait...
Faucet is starting, please wait...
Indexer API is starting, please wait...
Node API is starting, please wait...
Transaction stream is starting, please wait...

        Log file: "/Users/dport/.aptos/testnet/validator.log"
        Test dir: "/Users/dport/.aptos/testnet"
        Aptos root key path: "/Users/dport/.aptos/testnet/mint.key"
        Waypoint: 0:67f3a0e93406c0cb8c48f86843606a04dad2b84ce45084a6d3789d16cfd8eefc
        ChainId: 4
        Indexer gRPC node stream endpoint: 0.0.0.0:50051

Node API is ready. Endpoint: http://0.0.0.0:8080/
Postgres is ready. Endpoint: postgres://postgres@127.0.0.1:5433/local_testnet
Transaction stream is ready. Endpoint: http://0.0.0.0:50051/
Indexer API is ready. Endpoint: http://127.0.0.1:8090/
Faucet is ready. Endpoint: http://127.0.0.1:8081/

Applying post startup steps...

Setup is complete, you can now use the local testnet!
```

This is what you get when visiting the indexer API link:
Screenshot 2023-09-29 at 10-02-42 API Explorer Hasura

On my system the local testnet uses about 10-20% CPU.

### DIFF
--- a/crates/aptos/src/node/local_testnet/mod.rs
+++ b/crates/aptos/src/node/local_testnet/mod.rs
@@ -45,7 +45,7 @@ use std::{
     pin::Pin,
 };
 use tokio::task::JoinHandle;
-use tracing::info;
+use tracing::{info, warn};
 use tracing_subscriber::fmt::MakeWriter;
 
 const TESTNET_FOLDER: &str = "testnet";
@@ -340,6 +340,17 @@ impl CliCommand<()> for RunLocalTestnet {
         } else {
             eprintln!("\nOne of the futures exited unexpectedly, running shutdown steps...");
         }
+
+        // At this point replace the ctrl-c handler so the user can kill the CLI
+        // instantly if they send the signal twice.
+        tokio::spawn(async move {
+            tokio::signal::ctrl_c()
+                .await
+                .expect("Failed to register ctrl-c hook");
+            warn!("Received ctrl-c twice and exited immediately");
+            eprintln!();
+            std::process::exit(1);
+        });
 
         // Run shutdown steps, if any.
         for shutdown_step in shutdown_steps {
--- a/crates/aptos/src/node/local_testnet/mod.rs
+++ b/crates/aptos/src/node/local_testnet/mod.rs
@@ -91,6 +91,10 @@ pub struct RunLocalTestnet {
 
     #[clap(flatten)]
     prompt_options: PromptOptions,
+
+    /// By default, tracing output goes to files. With this set, it goes to stdout.
+    #[clap(long, hide = true)]
+    log_to_stdout: bool,
 }
 
 impl RunLocalTestnet {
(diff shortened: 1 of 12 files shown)


crates/aptos/src/node/local_testnet/mod.rs
```diff
@@ -248,6 +268,13 @@ impl CliCommand<()> for RunLocalTestnet {
             health_checkers.clone(),
         )?));
 
+        // Collect steps to run on shutdown. We run these in reverse.
+        let shutdown_steps: Vec<Box<dyn ShutdownStep>> = managers
+            .iter()
+            .flat_map(|m| m.get_shutdown_steps())
+            .rev()
+            .collect();
```
0xjinn, line 276: nit - Is it possible to put this block into `get_shutdown_steps()`, so we can just call and get the steps in the correct order? Would be much cleaner I think

crates/aptos/src/node/local_testnet/mod.rs
```diff
@@ -288,21 +315,47 @@ impl CliCommand<()> for RunLocalTestnet {
...
+        // Create a task that listens for ctrl-c. We want to intercept it so we can run
+        // the shutdown steps before properly exiting. This is of course best effort,
+        // see `ShutdownStep` for more info. In particular, to speak to how "best effort"
+        // this really is, to make sure ctrl-c happens more or less instantly, we only
+        // register this handler after all the services have started.
+        tasks.push(tokio::spawn(async move {
+            tokio::signal::ctrl_c()
+                .await
+                .expect("Failed to register ctrl-c hook");
+        }));
```
0xjinn, line 327: Just making sure - So this register the ctrl-c hook and stops here, until hook is triggered then we continue forward to execute shutdown_steps? If I just close the terminal, I wonder if that count as `ctrl-c`?

crates/aptos/src/node/local_testnet/mod.rs
banool: To clarify, this reverses the order of all the shutdown steps. Doing what you suggest would only effect the order of shutdown steps within the return value of each ServiceManager.

Also the reversing behaviour isn't that important, it's just a little more logical.

0xjinn: Ahh thats right, please ignore me lol

crates/aptos/src/node/local_testnet/mod.rs
banool: It doesn't stop exactly at this point, it stops at the select_all afterwards.

If someone closes the terminal then the CLI will receive a SIGHUP. We don't handle that right now so no cleanup will run. We could, but I'm choosing to only use things that are cross platform for now. Ctrl+C is meaningful on both Unix and Windows but SIGHUP is not.

There are other cases where there's nothing we can do to run the shutdown steps, like if the CLI gets SIGKILLed. So these shutdown steps are just best effort. It's not a huge deal if they don't run, the CLI will handle it on next run if the shutdown steps didn't happen.

I'll include something in the docs explaining how to clean up if the CLI exited uncleanly.

0xjinn (approved):
The general flow of starting and shutting down the local indexer api looks good to me.

Again, will let the indexer team to double check if there anything missing.

PR #10257, opened 2023-09-26
[Consensus] Add support for delaying QC to aggregate more votes

sitalkedia:
### Description

On a high level, this change supports delaying the QC aggregation by certain time period to allow more votes to be aggregated. This is useful in case certain nodes are slower/higher network latency and it gives them a chance to cast their vote and be up to date with the chain with the trade off of slight delay. In practice however, this delay is helping the performance of the network as otherwise nodes fall behind and the cost of remotely fetching the block is much higher than this delay we introduce. 

Following 3 knobs are introduces 

1. `max_delay_after_round_start_ms` -     Maximum Delay for a QC to be aggregated after round start (in milliseconds). This assumes that we have enough voting power to form a QC. If we don't have enough voting power, we will wait until we have enough voting power to form a QC.

2. `aggregated_voting_power_pct_to_wait ` -   Percentage of aggregated voting power to wait for before aggregating a QC. For example, if this is set to 95% then, a QC is formed as soon as we have 95% of the voting power aggregated without any additional waiting.

3. `pct_delay_after_qc_aggregated` -      This knob control what is the % of the time (as compared to time between round start and time when we have enough voting power to form a QC) we wait after we have enough voting power to form a QC. In a sense, this knobs controls how much slower we are willing to make consensus to wait for more votes.

In summary, let's denote the time we have enough voting power (2f + 1) to form a QC as T1 and the time we have aggregated `aggregated_voting_power_pct_to_wait` as T2. Then, we wait for `min((T1 + `pct_delay_after_qc_aggregated` * T1 / 100), `max_delay_after_round_start_ms`, T2)` before forming a QC.

Please note that there is a `NoDelay` aggregator, which is the default currently and so this change is a no-op. We plan to test this feature in previewnet before making the delay based aggregator default. 

### Test Plan

Tested this in previewnet like setup and this helps us achieve 10% higher TPS on a 100 node cluster.


consensus/src/round_manager.rs
```diff
@@ -145,6 +146,7 @@ pub enum VerifiedEvent {
     // network messages
     ProposalMsg(Box<ProposalMsg>),
     VerifiedProposalMsg(Box<Block>),
+    VerifiedDelayedQcMsg(Box<DelayedQcMsg>),
```
zekun000, line 149: I don't think we should add it to VerifiedEvent, it probably can be in its own channel with its own (simpler) type, it's confusing to see this under // network messages

consensus/src/round_manager.rs
```diff
@@ -435,6 +437,27 @@ impl RoundManager {
         self.process_verified_proposal(proposal).await
     }
 
+    pub async fn process_delayed_qc_msg(&mut self, msg: DelayedQcMsg) -> anyhow::Result<()> {
+        if msg.vote.vote_data().proposed().round() != self.round_state.current_round() {
```
zekun000, line 441: nit: just use ensure!

zekun000 (commented):
the changes looks reasonable, but the change is more than I expect.

I was expecting we just have something similar to the current timeout trigger, and the process that just call aggregate_now, and adding a vote has to satisfy `aggregated_voting_power_pct_to_wait` to aggregate directly. I guess adding the ability to delay % accordingly added the complexity

consensus/src/pending_votes.rs
```diff
@@ -257,6 +276,46 @@ impl PendingVotes {
...
+    ) -> VoteReceptionResult {
+        match li_with_sig.aggregate_signatures(validator_verifier) {
+            Ok(ledger_info_with_sig) => VoteReceptionResult::NewQuorumCertificate(Arc::new(
+                QuorumCert::new(vote_data.clone(), ledger_info_with_sig),
+            )),
+            Err(e) => VoteReceptionResult::ErrorAggregatingSignature(e),
+        }
+    }
+
+    pub fn process_delayed_qc(
```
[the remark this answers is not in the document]
sitalkedia: Hmm you are right but it might be better to recheck the voting power again before aggregating the QC.

config/src/config/consensus_config.rs
```diff
@@ -65,6 +65,40 @@ pub struct ConsensusConfig {
...
+    // })
+    Delayed(DelayedQcAggregatorConfig),
+}
+
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
+pub struct DelayedQcAggregatorConfig {
+    // Maximum Delay for a QC to be aggregated after round start (in milliseconds). This assumes that
+    // we have enough voting power to form a QC. If we don't have enough voting power, we will wait
+    // until we have enough voting power to form a QC.
+    pub max_delay_after_round_start_ms: u64,
```
zekun000, line 89: nit: probably worth having a config sanitizer ensuring this is smaller than round timeout.

zekun000 (approved):

igor-aptos (approved):
looks good to me.

for the unnecessary function - I'm fine either way,  no strong opinions

config/src/config/consensus_config.rs
sitalkedia: Let me look into this as a follow up.

PR #10258, opened 2023-09-26
fix build failure in e2e-testsuite

### LINKED ISSUE #10093: [Bug] e2e-testsuite build failure due to commit 33a82b2642
# 🐛 Bug

On the main branch, the `language-e2e-testsuite` build failed.
```
error[E0277]: the trait bound `AccountAddress: Arbitrary` is not satisfied
   --> aptos-move/framework/cached-packages/src/aptos_framework_sdk_builder.rs:860:9
    |
860 |         new_voter: AccountAddress,
    |         ^^^^^^^^^ the trait `Arbitrary` is not implemented for `AccountAddress`
    |
```
I run the `git bisect` and locate the faulty commit is 
`* 33a82b2642 (HEAD) [dep] moving testing features out from dependencies`

so-schen:
close #10093


aptos-move/e2e-testsuite/Cargo.toml
```diff
@@ -41,4 +41,8 @@ proptest = { workspace = true }
 serde = { workspace = true }
 
 [features]
-default = ["aptos-cached-packages/fuzzing"]
+default = [
+    "aptos-cached-packages/fuzzing",
```
wrwg, line 45: We don't want to turn those features on by default -- fuzzing code would end up in production. 

Unit tests are run as part of CI and they pass. See e.g. here for the CI action: https://github.com/aptos-labs/aptos-core/blob/7f9464c556709788f0f56b98a0f882702f968c18/.github/actions/rust-unit-tests/action.yaml#L41

We also run fuzzing regularily, @gedigi can say more about this. So there must be some difference in how you are trying to run things, which is not the defauilt way.

so-schen: this crate is a e2e test suite, why it will will be included in production? I guess the production build should not depend on this crate, right?

wrwg: Ah ok -- I misread the location of the toml. Can you rebase and we run CI again?

banool:
Relevant PR that would help make it easier to debug this btw: https://github.com/aptos-labs/aptos-core/pull/10314.

aptos-move/e2e-testsuite/Cargo.toml
so-schen: done, please check

wrwg (approved):

brmataptos (approved):

PR #10263, opened 2023-09-27
Make InMemoryStateCalculatorV2 work with state sync chunks.

grao1991:


execution/executor-types/src/executed_chunk.rs
```diff
@@ -89,34 +75,24 @@ impl ExecutedChunk {
...
-            for (idx, ((_, txn_data), expected_txn_info)) in
-                itertools::zip_eq(self.to_commit.iter(), transaction_infos.iter()).enumerate()
-            {
-                if &txn_data.txn_info != expected_txn_info {
+        let start_version =
+            self.result_view.txn_accumulator().version() + 1 - self.to_commit.len() as u64;
+        zip_eq(self.to_commit.iter_mut(), transaction_infos.iter())
+            .enumerate()
+            .try_for_each(|(idx, (txn_to_commit, expected_txn_info))| {
+                if expected_txn_info.state_checkpoint_hash().is_some() && txn_to_commit.transaction_info().state_checkpoint_hash().is_none() {
```
msmouse, line 84:
```
// On replaying, we generate state checkpoints only periodically, for the last state checkpoint of each chunk; A mismatch in the SMT will be detected at that occasion too. Here we just copy in the state root from the TxnInfo in the proof.
```

types/src/transaction/mod.rs
```diff
@@ -1345,6 +1345,25 @@ impl TransactionToCommit {
...
+        let transaction_info_with_state_checkpoint_hash = TransactionInfo::new(
+            transaction_info.transaction_hash(),
+            transaction_info.state_change_hash(),
+            transaction_info.event_root_hash(),
+            Some(state_checkpoint_hash),
+            transaction_info.gas_used(),
+            transaction_info.status().clone(),
+        );
+
+        self.transaction_info = transaction_info_with_state_checkpoint_hash;
```
msmouse, line 1362: why not update in place?

execution/executor/src/components/in_memory_state_calculator_v2.rs
```diff
@@ -110,56 +92,104 @@ impl InMemoryStateCalculatorV2 {
...
+                &proof_reader,
+            )?
+        } else {
+            base.base.clone()
         };
 
+        let first_version = base.current_version.map_or(0, |v| v + 1);
+        let mut latest_checkpoint_version = base.base_version;
+        let mut state_checkpoint_hashes = Vec::with_capacity(num_txns);
+        state_checkpoint_hashes.resize(num_txns, None);
```
msmouse, line 144: `vec![None; num_txns]` should work? https://play.rust-lang.org/?version=stable&mode=debug&edition=2021

grao1991: Done

types/src/transaction/mod.rs
grao1991: Want to avoid exposing a mutable API directly from TransactionInfo, to limit the place to mutate it.

execution/executor-types/src/executed_chunk.rs
grao1991: Done

types/src/transaction/mod.rs
msmouse: Then a `TransactionInfo::fill_state_checkpoint_hash(mut self, hash: HashValue) -> Self` would be neat (then you can https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=aaba6e74c2baed5286d6a3431fd275b5). up to you.

grao1991: Just realized I need to update the hash before calculating the accumulator.. Now this function is not needed.

execution/executor/src/components/apply_chunk_output.rs
```diff
@@ -145,56 +141,39 @@ impl ApplyChunkOutput {
     pub fn apply_chunk(
         chunk_output: ChunkOutput,
         base_view: &ExecutedTrees,
+        state_checkpoint_hashes: Option<Vec<Option<HashValue>>>,
         append_state_checkpoint_to_block: Option<HashValue>,
     ) -> Result<(ExecutedChunk, Vec<Transaction>, Vec<Transaction>)> {
-        let ChunkOutput {
```
msmouse, line 150: @grao1991  promises to remove the entire calculator v1 NEXT WEEK :trollface:

msmouse (approved):

zekun000 (approved):

PR #10270, opened 2023-09-27
[marketplace / framework] Fix divide by zero and marketplace royalty edge conditions

gregnazario:
### Description
Adds an error message for a divide by zero error in the aptos stdlib and handles divide by 0 in royalties in the example marketplace contract.

Additionally, adds checks for royalty 100% or 0 denominator in the marketplace, and ensures that listings go through rather than get stuck with insufficient balance errors.

Choices in the marketplace were:
1. Take royalty and commission from full amount, then make one taken before the other (I chose this, where commission is taken second)
2. Take the commission first, then take the royalty percentage from the remaining amount
3. Take the royalty first then take the commission percentage / amount from the remaining amount

### Test Plan
See unit tests

### DIFF
--- a/aptos-move/framework/aptos-stdlib/sources/math128.move
+++ b/aptos-move/framework/aptos-stdlib/sources/math128.move
@@ -6,9 +6,8 @@ module aptos_std::math128 {
     use aptos_std::fixed_point64::FixedPoint64;
     use aptos_std::fixed_point64;
 
-    /// Abort value when an invalid argument is provided.
+    /// Cannot log2 the value 0
     const EINVALID_ARG_FLOOR_LOG2: u64 = 1;
-    const EDIVISION_BY_ZERO: u64 = 2;
 
     /// Return the largest of two numbers.
     public fun max(a: u128, b: u128): u128 {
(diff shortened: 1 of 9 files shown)


movekevin (approved):
Nice fix! Can you add a simple unit test for the marketplace change? (All zero royalty)

gregnazario:
@movekevin this interfacing doesn't work, because you have to copy the divide by zero error code everywhere due to the inline... I will probably just remove the framework changes, since that causes other issues with it.

aptos-move/framework/aptos-stdlib/sources/math128.move
```diff
@@ -29,8 +28,9 @@ module aptos_std::math128 {
         }
     }
 
-    /// Returns a * b / c going through u128 to prevent intermediate overflow
+    /// Returns a * b / c going through u256 to prevent intermediate overflow
     public inline fun mul_div(a: u128, b: u128, c: u128): u128 {
+        assert!(c != 0, std::error::invalid_argument(4));
```
lightmark, line 33: what's 4 here?

lightmark (approved):

aptos-move/framework/aptos-stdlib/sources/math128.move
lightmark: I thought the best practice is to make 4 a constant of this file?

gregnazario: I guess I should fix the comment, that I thought I had.

You cannot have a constant, because it doesn't copy between inlines.  So, the best we can do is to have a common number, in this case 4 is used for fixed_point32 when a divide by 0 happens

PR #10252, opened 2023-09-26
[CLI] Restructure local testnet code

banool:
### Stack
- Previous in stack: https://github.com/aptos-labs/aptos-core/pull/10214
- Next in stack: https://github.com/aptos-labs/aptos-core/pull/10271

### Description
When the local testnet was just a node and a faucet, it was somewhat acceptable to just have everything jammed into one file. With testnets becoming more fully featured, this would've become a big mess.

In this PR I introduce the ServiceManager trait. Using this trait, I separate each service into its own file, in which have:
- Configuration necessary for that service.
- Pre-run and post-healthy steps.
- Code for running the service itself.
- Logic for structuring service startup order, leveraging HealthCheckers.

### Test Plan
Build a tools image locally:
```
DOCKER_DEFAULT_PLATFORM=linux/amd64 docker/builder/docker-bake-rust-all.sh tools
```
I used the remote docker builder for this.

Tag the image as devnet so I could use it as the base network for the CLI E2E tests:
```
docker tag aptos-core/tools:from-local aptos-core/tools:devnet
```

Run the CLI E2E tests where we use the local tools image:
```
cd crates/aptos/e2e
poetry install
DOCKER_DEFAULT_PLATFORM=linux/amd64  poetry run python main.py -d --base-network devnet --test-cli-path ~/a/core/target/debug/aptos --image-repo-with-project aptos-core --no-pull-always
```
```
2023-09-27 13:56:24,157 - INFO - All tests passed!
```

This is what it looks like when you run the local testnet now:
```
$ cargo run -p aptos -- node run-local-testnet --assume-yes --force-restart
Readiness endpoint: http://0.0.0.0:8070/

Faucet is starting, please wait...
Node API is starting, please wait...
Transaction stream is starting, please wait...

Completed generating configuration:
        Log file: "/Users/dport/.aptos/testnet/validator.log"
        Test dir: "/Users/dport/.aptos/testnet"
        Aptos root key path: "/Users/dport/.aptos/testnet/mint.key"
        Waypoint: 0:db2e0a937d667a20f2cd4b16a51f37d2d8d25e4e87301cbab2e89e33b41399be
        ChainId: 4
        REST API endpoint: http://0.0.0.0:8080
        Metrics endpoint: http://0.0.0.0:9101/metrics
        Aptosnet fullnode network endpoint: /ip4/0.0.0.0/tcp/6181
        Indexer gRPC node stream endpoint: 0.0.0.0:50051

Aptos is running, press ctrl-c to exit

[...]

PR #10271, opened 2023-09-27
[CLI] Add support for running postgres in a container to the local testnet + rest of PRs in stack

banool:
## Stack
- Previous in stack: https://github.com/aptos-labs/aptos-core/pull/10252
- Next in stack: https://github.com/aptos-labs/aptos-core/pull/10254

## Description
This PR adds support for:
- Spinning up postgres in a container.
- Using postgres already running in the host system.

This is required to run indexer processors.

I still need to double check that this works on a fresh system, I have a suspicion that if postgres isn't installed things will fail with some error to the effect of "postgres shared library missing".

I also need to make sure I'm not reintroducing the OpenSSL dep.

This depends on https://github.com/aptos-labs/aptos-indexer-processors/pull/146.

Note: To avoid using undesired diesel features in the CLI I had to move the feature declarations from the workspace level to the crate level. Crate level feature declarations are additive, they don't replace the features used at the workspace level.

### Postgres in a container
```
cargo run -p aptos -- node run-local-testnet --assume-yes --force-restart --with-indexer-api --postgres-user dport
```
```
Node API is ready. Endpoint: http://0.0.0.0:8080/
Postgres is ready. Endpoint: postgres://dport@127.0.0.1:5433/local_testnet
Transaction stream is ready. Endpoint: http://0.0.0.0:50051/
Faucet is ready. Endpoint: http://127.0.0.1:8081/
```
```
psql postgres://dport@127.0.0.1:5433/local_testnet
```
```
psql (14.9 (Homebrew))
Type "help" for help.

local_testnet=# \dn
List of schemas
  Name  | Owner
--------+-------
 public | dport
(1 row)
```

Once the CLI shuts down so does the postgres container.

### Host postgres
```
cargo run -p aptos -- node run-local-testnet --assume-yes --force-restart --with-indexer-api --postgres-user dport --use-host-postgres
```
```
Postgres is ready. Endpoint: postgres://dport@127.0.0.1:5432/local_testnet
```
```
psql postgres://dport@127.0.0.1:5433/local_testnet
```
```
psql (14.9 (Homebrew))
Type "help" for help.

local_testnet=# \dn
List of schemas
  Name  | Owner
--------+-------
 public | dport
(1 row)
```

## Further Testing
Now that I've rebased all the other PRs in the stack into this PR, I will do some additional testing. I'm using this invocation:
```
cargo run -p aptos -- node run-local-testnet --assume-yes --force-restart --with-indexer-api
```

Does it work on:
- ARM MacOS: Yes!!
- x86 Ubuntu 22: Yes!!
- x86 Windows (host system, NT, MSVC): Not working, but due to an unrelated error that occurs on main too: https://gist.github.com/banool/135ce49c0662832446778c6268b68d62.
- x86 Windows (Ubuntu 22 WSL): Yes!!
- CLI running inside container from x86 Linux tools image running on ARM MacOS: Yes!!

For the last of these, I ran it like this:
```
DOCKER_DEFAULT_PLATFORM=linux/amd64 docker run -it -v /var/run/docker.sock:/var/run/docker.sock --network host aptos-core/tools:from-local aptos node run-local-testnet --with-indexer-api
```

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -24,6 +24,7 @@
 !aptos-move/framework/
 !aptos-move/move-examples/hello_blockchain/
 !crates/aptos/src/move_tool/*.bpl
+!crates/aptos/src/node/local_testnet/hasura_metadata.json
 !crates/aptos-faucet/doc/
 !api/doc/
 !crates/indexer/migrations/**/*.sql
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -258,6 +258,8 @@ dependencies = [
  "clap_complete",
  "codespan-reporting",
  "dashmap",
+ "diesel",
+ "diesel-async",
  "dirs",
  "futures",
  "hex",
@@ -4324,7 +4326,7 @@ dependencies = [
  "parking",
  "polling",
  "slab",
- "socket2",
+ "socket2 0.4.9",
  "waker-fn",
  "winapi 0.3.9",
 ]
@@ -5786,7 +5788,7 @@ dependencies = [
  "openssl-probe",
  "openssl-sys",
  "schannel",
- "socket2",
+ "socket2 0.4.9",
  "winapi 0.3.9",
 ]
 
@@ -6009,9 +6011,9 @@ checksum = "850878694b7933ca4c9569d30a34b55031b9b139ee1fc7b94a527c4ef960d690"
 
 [[package]]
 name = "diesel"
-version = "2.1.0"
+version = "2.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f7a532c1f99a0f596f6960a60d1e119e91582b24b39e2d83a190e61262c3ef0c"
+checksum = "53c8a2cb22327206568569e5a45bb5a2c946455efdd76e24d15b7e82171af95e"
 dependencies = [
  "bigdecimal",
  "bitflags 2.4.0",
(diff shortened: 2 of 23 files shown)


gregnazario (approved):
lgtm, comments on CLI commands may be a little long.

keep in mind that the first line will show in `-h` and the rest of the lines will show in `--help`, I don't know what the cutoff is.

banool:
Turns out diesel relies on pq-sys, which wraps libpq. It isn't acceptable to add this as a dep for the CLI, we have no way to ensure users have it installed and even if we did, it feels like an unnecessary dep for users who aren't using the local testnet / this feature of the local testnet.

I'm investigating alternatives here: https://aptos-org.slack.com/archives/C04PRP1K1FZ/p1696019475810529. Ultimately we'd have to stop using diesel...

banool:
I switched to tokio-postgres. This solves it for this PR but the processors still use the mainline diesel postgres library, which uses libpg internally. To address this, I will make use of this branch for now in that PR: https://github.com/aptos-labs/aptos-indexer-processors/pull/146. We'll discuss whether we want to actually adopt this PR or if I have to use an alternative, like implementing diesel::Connection on https://github.com/sfackler/rust-postgres and using that.

davidiw (commented):
what happens if you don't have docker installed?

banool:
It tells you you don't have Docker installed and fails fast: https://github.com/aptos-labs/aptos-core/pull/10271/files#diff-860697684f8880833202701e67e2738c3cabf0655aaf9e53ea163877119ae885R155.

0xjinn (approved):

larry-aptos (approved):

banool:
See the further testing section, I just confirmed that the new features work on all major systems we support. It doesn't work on Windows (NT) but that has nothing to do with this PR (I'll chase that issue up separately).

banool:
Landing this is currently blocked: https://aptos-org.slack.com/archives/C03N83P7QUC/p1691767293160939. I need to update the version of serde we use but doing so breaks some kind of old codegen / format gen we have.

banool:
This change to Bollard means we can keep using the version of serde we're using now: https://github.com/fussybeaver/bollard/pull/343.

banool:
I ran this test plan again and it all still works nicely with the Bollard change: https://github.com/aptos-labs/aptos-core/pull/10333. We should be good to go if all the other tests pass.

banool:
I tested that the Docker stuff still works with this fork of Bollard, let's land.

PR #10274, opened 2023-09-27
[Sharded-Execution-GRPC] Add GRPC communication for sharded execution.

manudhundi:
### Description
    In this commit we replace the existing socket based communication with GRPC.
    Here we get the basic GRPC reliability.
    
    More reliability and better performance to come in subsequent commits


crates/aptos-protos/proto/aptos/remote_executor/v1/network_msg.proto
```diff
@@ -0,0 +1,20 @@
...
+    string sender_addr = 1;
+    bytes message = 2;
+    string message_type = 3;
+}
+
+message Empty {
+
+}
+
+service RemoteExecution {
```
sitalkedia, line 18: The name should be generic.. something like `NetworkingService`

secure/net/src/grpc_network_service/mod.rs
```diff
@@ -0,0 +1,194 @@
...
+    ) {
+        let request = tonic::Request::new(NetworkMessage {
+            sender_addr: sender_addr.to_string(),
+            message: message.data,
+            message_type: mt.get_type(),
+        });
+        // TODO: Retry with exponential backoff on failure
+        match self.remote_channel.simple_msg_exchange(request).await {
+            Ok(_) => {},
+            Err(e) => {
```
sitalkedia, line 128: Hmm we should just panic in this case, right?

secure/net/src/network_controller/inbound_handler.rs
```diff
@@ -1,34 +1,29 @@
...
     service: String,
-    server: Arc<Mutex<NetworkServer>>,
-    // Used to route incoming messages to correct channel.
+    listen_addr: SocketAddr,
     inbound_handlers: Arc<Mutex<HashMap<MessageType, Sender<Message>>>>,
 }
 
 impl InboundHandler {
-    pub fn new(service: String, listen_addr: SocketAddr, timeout_ms: u64) -> Self {
+    pub fn new(service: String, listen_addr: SocketAddr, _: u64) -> Self {
```
sitalkedia, line 23: remove the timeout if not applicable anymore.. how do we configure the timeout for gRPC btw?

secure/net/src/network_controller/mod.rs
```diff
@@ -144,6 +193,10 @@ mod tests {
         network_controller1.start();
         network_controller2.start();
 
+        // wait for the server to be ready to serve
+        // TODO: We need to pass this test without this sleep
+        thread::sleep(std::time::Duration::from_millis(10));
```
sitalkedia, line 198: See my comment on using a one shot channel to ensure the server is started.. we can probably get rid of the sleep here..

secure/net/src/network_controller/outbound_handler.rs
```diff
@@ -29,84 +30,126 @@ impl OutboundHandler {
...
-                )
-            })
-            .expect("Failed to spawn outbound handler thread");
+        // Moving the handlers out of self is fine because once 'start()' is called we do not intend
+        // to register any more handlers. A reference count like Arc<Mutex> has issues of being
+        // used across sync and async boundaries, and also not the most efficient because we pay
+        // the cost of the mutex when there is no contention
+        let outbound_handlers = mem::take(self.handlers.as_mut());
+
+        // TODO: Consider using multiple tasks for outbound handlers
```
sitalkedia, line 84: What would it take to multi-thread this? I am afraid, without processing messages in multiple threads, this will be a chokepoint in remote execution environment.

secure/net/src/network_controller/outbound_handler.rs
```diff
@@ -29,84 +30,126 @@ impl OutboundHandler {
...
+            {
+                let oper = select.select();
+                index = oper.index();
+                match oper.recv(&outbound_handlers[index].0) {
+                    Ok(m) => {
+                        msg = m;
+                    },
+                    Err(e) => {
+                        // not necessarily an error, just means a sender closed a channel
+                        trace!(
```
sitalkedia, line 122: Why is this not an error? Silently ignoring error here might mean the transport layer doesn't provide reliable message delivery..

sitalkedia (commented):
Great work. Looks good overall % minor nits.

crates/aptos-protos/proto/aptos/remote_executor/v1/network_msg.proto
```diff
@@ -0,0 +1,20 @@
+// Copyright © Aptos Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+syntax = "proto3";
+
+package aptos.remote_executor.v1;
+
+message NetworkMessage {
```
grao1991, line 8: why do we choose to use generic message here instead of individual request message type for each rpc?

crates/aptos-protos/src/pb/aptos.remote_executor.v1.rs
```diff
@@ -0,0 +1,68 @@
+// Copyright © Aptos Foundation
```
grao1991, line 1: Do we have to include generated file into git repo?

crates/aptos-protos/proto/aptos/remote_executor/v1/network_msg.proto
```diff
@@ -0,0 +1,20 @@
...
+    string sender_addr = 1;
```
grao1991, line 9: sender_addr probably should go to side channel

execution/executor-service/src/tests.rs
```diff
@@ -56,10 +56,22 @@ pub fn create_thread_remote_executor_shards(
...
     let num_shards = 8;
-    let (mut controller, executor_client, _executor_services) =
+    let (mut controller, executor_client, mut executor_services) =
         create_thread_remote_executor_shards(num_shards, Some(2));
     controller.start();
     let sharded_block_executor = ShardedBlockExecutor::new(executor_client);
+
+    // wait for the servers to be ready before sending messages
+    // TODO: We need to pass this test without this sleep
+    thread::sleep(std::time::Duration::from_millis(10));
```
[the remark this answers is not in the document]
manudhundi: wait for server to start.. Yes will address

secure/net/src/grpc_network_service/mod.rs
```diff
@@ -0,0 +1,194 @@
...
+    net::SocketAddr,
+    sync::{Arc, Mutex},
+};
+use tokio::{runtime::Runtime, sync::oneshot};
+use tonic::{
+    transport::{Channel, Server},
+    Request, Response, Status,
+};
+
+pub struct RemoteExecutionServerWrapper {
```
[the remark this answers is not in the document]
manudhundi: Ok

secure/net/src/grpc_network_service/mod.rs
```diff
@@ -0,0 +1,194 @@
...
+    pub fn start(
+        self,
+        rt: &Runtime,
+        _service: String,
+        server_addr: SocketAddr,
+        server_shutdown_rx: oneshot::Receiver<()>,
+    ) {
+        rt.spawn(async move {
+            self.start_async(server_addr, server_shutdown_rx).await;
+        });
```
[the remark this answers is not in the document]
manudhundi: Yes, one_shot channel is a good idea.. This is basically my  3 of my TODOs. Waiting for server to start before sending messages. Hence those sleeps in test cases

secure/net/src/grpc_network_service/mod.rs
manudhundi: Ideally we should have retries. There could be temporary failures on server. We can panic for now.

secure/net/src/network_controller/mod.rs
manudhundi: yes one shot channel will help here

secure/net/src/network_controller/outbound_handler.rs
manudhundi: I suppose minimal overhead. Just the logic of splitting the outbound_handlers across the threads/tasks

secure/net/src/network_controller/outbound_handler.rs
manudhundi: This just means one of the senders has closed the send channel. It can happen when the sender_tx channel is dropped after it has done its job.
Also, we should not panic at this level if one the sender dies because many senders subscribe to this outbound task. If a sender_tx channel dies due to an error, then it is application error, not transport layer error.

crates/aptos-protos/proto/aptos/remote_executor/v1/network_msg.proto
manudhundi: Good question. 
Individual request message type for each rpc is the goal. This is just an initial commit to (i) start evaluating grpc (ii) quickly have a reliable communication version of what we have today.

crates/aptos-protos/src/pb/aptos.remote_executor.v1.rs
manudhundi: Yes, I think so. Other grpc generated files have been added to git repo

secure/net/src/network_controller/outbound_handler.rs
sitalkedia: This can mean many things (not necessarily sender dropping).. so I would be hesitant to ignore this and continue.. what is the downside of panic/exiting from this loop when this happens?

secure/net/src/grpc_network_service/mod.rs
sitalkedia: You can add some retry with exponetial back-off I assume..and panic once we exhaust the retries..

crates/aptos-protos/proto/aptos/remote_executor/v1/network_msg.proto
manudhundi: ok

crates/aptos-protos/proto/aptos/remote_executor/v1/network_msg.proto
manudhundi: I will use 'NetworkMessageService'

secure/net/src/grpc_network_service/mod.rs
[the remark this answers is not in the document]
manudhundi: There is no easy way to do this by APIs provided by GRPC or Tonic. Ways to achieve this would be by doing a check/retry from client side.
As discussed going to leave this a TODO for now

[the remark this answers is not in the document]
manudhundi: One way to do this would be implementing a healthcheck service
https://stackoverflow.com/questions/66798278/golang-grpc-how-to-determine-when-the-server-has-started-listening

secure/net/src/grpc_network_service/mod.rs
manudhundi: Adding retries requires an additional copy of msg to send. 
(This can be achieved by existing retry_async() library)

Since we do not want to bite the bullet of the cost of additional copy, leaving retry as a TODO for now

secure/net/src/network_controller/outbound_handler.rs
manudhundi: Ok as discussed, panic'ing on error here

secure/net/src/network_controller/mod.rs
manudhundi: one shot channel will not help as explained above. Closing this comment as duplicate

secure/net/src/network_controller/inbound_handler.rs
manudhundi: Using the timeouts as is. We can set in gRPC when we create the server

secure/net/src/network_controller/outbound_handler.rs
sitalkedia: @manudhundi - As discussed, instead of continuing here, let's either panic or shut down the handler in case of error?

manudhundi: Shutting down the handler. Panic and shutdown would mean the same I suppose as the task just runs this handler

sitalkedia: In case we want to shut down, we should break out of this loop right?

manudhundi: Yes we do break out using the signal from the msg of type "stop_task"

secure/net/src/grpc_network_service/mod.rs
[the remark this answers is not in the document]
sitalkedia: Health check service is a good idea.

secure/net/src/network_controller/outbound_handler.rs
```diff
@@ -29,84 +30,127 @@ impl OutboundHandler {
...
-    pub fn start(&mut self) {
-        let outbound_handlers = self.handlers.clone();
+    pub fn start(&mut self, rt: &Runtime) -> Option<Sender<Message>> {
+        if self.handlers.is_empty() {
+            return None;
+        }
+
+        // Register a signal handler to stop the outbound task
+        let (stop_signal_tx, stop_signal_rx) = unbounded();
+        self.handlers.push((
```
sitalkedia, line 59: Better you do it in the constructor and remove the handler.is_empty() check on line 53.

secure/net/src/network_controller/outbound_handler.rs
```diff
@@ -29,84 +30,127 @@ impl OutboundHandler {
...
+                        return;
+                    },
+                }
+            }
+
+            let remote_addr = &outbound_handlers[index].1;
+            let message_type = &outbound_handlers[index].2;
 
-        retry(fixed_retry_strategy(5, 20), || network_client.write(&msg)).unwrap();
+            if message_type.get_type() == "stop_task" {
```
sitalkedia, line 136: What is the use of this stop message?

manudhundi: It is a clean way of exiting this task. NetworkController::shutdown calls this

secure/net/src/network_controller/outbound_handler.rs
manudhundi: That will require all the handlers registered in the constructor ? Will that not break the nice modularity that we have ? That is 1 network controller/outbound_task per node. And many modules are individually register their handlers

secure/net/src/network_controller/outbound_handler.rs
manudhundi: For 2 reasons:
(1) Runtime going out scope kills the task in a heavy handed way. That is, if the task in middle of something, it will die in between.
(2) Not an expert on tokio runtimes, but i have seen hangs sometime. That is, runtime object is not dropped until the task dies

sitalkedia: > (1) Runtime going out scope kills the task in a heavy handed way. That is, if the task in middle of something, it will die in between.


This shouldn't be a problem right as the run time going out of scope means we are shutting down the controller anyways.. 

Do you see the test hanging without the shutdown support?

sitalkedia (approved):
Stamping to unblock, there are few TODOs like error handling, shutdown mechanism etc that we need to follow up.

manudhundi:
Thanks. Yup, will follow up on the shutdown mechanism and other TODOs once we have some reasonable perf numbers with remote sharding

secure/net/src/grpc_network_service/mod.rs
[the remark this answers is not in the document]
msmouse: Maybe we can just use the reflection service?

msmouse (approved):

execution/executor-service/src/remote_state_view.rs
```diff
@@ -212,6 +211,12 @@ impl RemoteStateValueReceiver {
             .inner
             .into_iter()
             .for_each(|(state_key, state_value)| {
+                // Though we try to insert the state key during the 'pre_fetch_state_values' remote
+                // request, there is a possibility that the state key is not inserted before we
+                // receive and process the response here. To be on the safe side, we insert the
+                // state key here if it is not inserted already. This does not matter because we
+                // do not have versioning of the state view.
+                state_view_lock.insert_state_key(state_key.clone());
```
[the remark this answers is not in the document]
manudhundi: Not needed, I will remove it.. I will rebase on previous change

secure/net/src/grpc_network_service/mod.rs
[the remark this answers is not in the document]
manudhundi: That is a good point, I will check into it

PR #10275, opened 2023-09-27
[TS SDK V2] Add `getGasPriceEstimation` query

0xmaayan:
### Description
This PR adds `getGasPriceEstimation` query to unblock transaction submission.
It also adds the Transaction API class.

### Test Plan
```
transaction api
    ✓ it queries for the network estimated gas price (26 ms)
```

### DIFF
--- a/ecosystem/typescript/sdk_v2/src/internal/account.ts
+++ b/ecosystem/typescript/sdk_v2/src/internal/account.ts
@@ -19,8 +19,7 @@ import {
 import { get } from "../client";
 import { paginateWithCursor } from "../utils/paginate_with_cursor";
 import { AccountAddress } from "../core";
-import { AptosApiType, APTOS_COIN } from "../utils/const";
-import { getGasPriceEstimation } from "./transaction";
+import { AptosApiType } from "../utils/const";
 
 export async function getInfo(args: { aptosConfig: AptosConfig; accountAddress: HexInput }): Promise<AccountData> {
   const { aptosConfig, accountAddress } = args;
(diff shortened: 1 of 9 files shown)


ecosystem/typescript/sdk_v2/src/internal/account.ts
```diff
@@ -121,3 +122,19 @@ export async function getResource(args: {
...
   return data;
 }
+
+export async function estimateAccountMaxGasAmount(args: { aptosConfig: AptosConfig; accountAddress: HexInput }) {
+  const { aptosConfig, accountAddress } = args;
+  // Only APT coin is accepted as gas
+  const typeTag = `0x1::coin::CoinStore<${APTOS_COIN}>`;
+  const [{ gas_estimate: gasUnitPrice }, resources] = await Promise.all([
+    getGasPriceEstimation({ aptosConfig }),
+    getResources({ aptosConfig, accountAddress }),
```
gregnazario, line 132: We should use `getResource()`, rather than `getResources()`, since we know the resource.  Otherwise, if you have > 10k resources, you'll never find it and it'll be expensive.

ecosystem/typescript/sdk_v2/src/internal/account.ts
```diff
@@ -121,3 +122,19 @@ export async function getResource(args: {
...
+  ]);
+  const accountResource = resources.find((r) => r.type === typeTag);
+  if (!accountResource) {
+    throw new Error(`account ${accountAddress} doesnt have a ${typeTag} resource`);
+  }
+  const balance = BigInt((accountResource.data as any).coin.value);
+  return balance / BigInt(gasUnitPrice);
```
gregnazario, line 139: I think I'm confused what this number is supposed to represent. MaxGasAmount is specific to a transaction, and comes out of simulation.  I'm not sure we need this function

Reason for this is, if you want the `Max Max gas amount`, you'll need to retrieve the gas schedule, and ceiling it to the maximum gas value, which last I remember is 2 million gas units.

See: https://github.com/aptos-labs/aptos-core/blob/main/api/src/transactions.rs#L452-L510

gregnazario (approved):
I would remove `estimateAccountMaxGasAmount` for now

heliuchuan (approved):

PR #10278, opened 2023-09-27
[Forge][PFN] remove epoch changes from some tests

bchocho:
### Description

This is in an effort to reduce noise in the PFN tests. Epoch changes are an obvious source of noise, although it's unclear how much it contributes to noise difference between runs.

To still have coverage of epoch changes, the tests without chaos will still do 5 minute epoch changes.

### Test Plan

Run ad-hoc forge run, observe no epoch changes.

### DIFF
--- a/testsuite/forge-cli/src/main.rs
+++ b/testsuite/forge-cli/src/main.rs
@@ -580,12 +580,12 @@ fn single_test_suite(
         "quorum_store_reconfig_enable_test" => quorum_store_reconfig_enable_test(),
         "mainnet_like_simulation_test" => mainnet_like_simulation_test(),
         "multiregion_benchmark_test" => multiregion_benchmark_test(),
-        "pfn_const_tps" => pfn_const_tps(duration, false, false),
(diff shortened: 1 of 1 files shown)


brianolson (approved):

testsuite/forge-cli/src/main.rs
```diff
@@ -2060,18 +2066,24 @@ fn pfn_performance(
...
     // Create the forge config
     ForgeConfig::default()
         .with_initial_validator_count(NonZeroUsize::new(7).unwrap())
         .with_initial_fullnode_count(7)
         .add_network_test(PFNPerformance::new(7, add_cpu_chaos, add_network_emulation))
-        .with_genesis_helm_config_fn(Arc::new(|helm_values| {
-            // Require frequent epoch changes
-            helm_values["chain"]["epoch_duration_secs"] = 300.into();
+        .with_genesis_helm_config_fn(Arc::new(move |helm_values| {
+            // Avoid epoch changes which can introduce noise
```
JoshLind, line 2085: nit: do we need this comment?

JoshLind (approved):
LGTM.

PR #10279, opened 2023-09-27
[forge] Tunings for realistic_network_tuned_for_throughput

msmouse:
### Description

Tunings used for getting some big TPS numbers

### Test Plan

https://github.com/aptos-labs/aptos-core/actions/runs/6489091369

### DIFF
--- a/testsuite/forge-cli/src/main.rs
+++ b/testsuite/forge-cli/src/main.rs
@@ -660,55 +660,7 @@ fn run_consensus_only_realistic_env_max_tps() -> ForgeConfig {
             helm_values["chain"]["epoch_duration_secs"] = (24 * 3600).into();
         }))
         .with_validator_override_node_config_fn(Arc::new(|config, _| {
-            mempool_config_practically_non_expiring(&mut config.mempool);
-            state_sync_config_execute_transactions(&mut config.state_sync);
-
-            config
-                .consensus
-                .max_sending_block_txns_quorum_store_override = 30000;
-            config
-                .consensus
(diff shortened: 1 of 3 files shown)


grao1991 (approved):

testsuite/forge-cli/src/main.rs
```diff
@@ -1693,53 +1697,54 @@ fn realistic_env_max_load_test(
 }
 
 fn realistic_network_tuned_for_throughput_test() -> ForgeConfig {
```
sitalkedia, line 1699: How about adding a new test and leave the current one unchanged?

msmouse: The original runs at 6k iirc, I don't think it's useful or establish any baseline?

msmouse: Or I'm missing how it's used currently? You and @igor-aptos  track it anywhere?

sitalkedia: I thought the original runs at ~8k TPS and is included in the list of stable forge test that runs daily.. @igor-aptos can confirm..

msmouse: Good catch then. Gonna update.

msmouse: Or maybe just comment out the resource ask out and update the numbers in CI? ( I have difficulty naming things. :trollface: 
 )

sitalkedia (approved):

PR #10283, opened 2023-09-27
[Forge][netbench] split into large and small messages for two region test

bchocho:
### Description

We found that the throughput for large and small messages is very different with large latencies. We update the test to run with large messages and then split out a small messages test. The large messages are useful for sanity checking the network setup, the small messages is something that we can hopefully improve upon.

### DIFF
--- a/testsuite/forge-cli/src/main.rs
+++ b/testsuite/forge-cli/src/main.rs
@@ -589,6 +589,7 @@ fn single_test_suite(
         "gather_metrics" => gather_metrics(),
         "net_bench" => net_bench(),
         "net_bench_two_region_env" => net_bench_two_region_env(),
+        "net_bench_two_region_env_small_messages" => net_bench_two_region_env_small_messages(),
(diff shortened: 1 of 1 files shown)


brianolson (approved):

testsuite/forge-cli/src/main.rs
```diff
@@ -589,6 +589,7 @@ fn single_test_suite(
         "gather_metrics" => gather_metrics(),
         "net_bench" => net_bench(),
         "net_bench_two_region_env" => net_bench_two_region_env(),
+        "net_bench_two_region_env_small_messages" => net_bench_two_region_env_small_messages(),
```
JoshLind, line 592: Not for this PR, but at some point we'll really need to split this file up 😆

JoshLind (approved):
LGTM.

testsuite/forge-cli/src/main.rs
bchocho: Yeah, it's a monster...

PR #10287, opened 2023-09-28
[voting] add more view functions to the voting module

0xbe1:
### Description

for better visibility into the voting module

### Test Plan

not needed since they're pretty straightforward view functions


aptos-move/framework/aptos-framework/sources/voting.move
```diff
@@ -493,6 +493,13 @@ module aptos_framework::voting {
         voting_forum.next_proposal_id
     }
 
+    #[view]
+    public fun get_proposer<ProposalType: store>(voting_forum_address: address, proposal_id: u64): address acquires VotingForum {
+        let voting_forum = borrow_global<VotingForum<ProposalType>>(voting_forum_address);
+        let proposal = table::borrow(&voting_forum.proposals, proposal_id);
```
gregnazario, line 499: Minor nit: I would probably put this in an inline function to be used on all of them, though it looks like it's already done like this elsewhere

aptos-move/framework/aptos-framework/sources/voting.move
```diff
@@ -511,6 +518,27 @@ module aptos_framework::voting {
         false
     }
 
+    #[view]
+    public fun get_proposal_metadata<ProposalType: store>(
+        voting_forum_address: address,
+        proposal_id: u64,
+    ): SimpleMap<String, vector<u8>> acquires VotingForum {
```
gregnazario, line 525: Can you test this out?  I'm not sure how a SimpleMap will be represented on the output of a view function

gregnazario (approved):

aptos-move/framework/aptos-framework/sources/voting.move
0xbe1: discussed offline, a SimpleMap output in view function will be like
image

here's the proposal_votes_breakdown definition:
```
    #[view] /// Table of all votes casted on a proposal
    public fun proposal_votes_breakdown(dao_address: address, proposal_id: u64): SimpleMap<address, ProposalVote> acquires DAO {
        assert!(exists<DAO>(dao_address), ERR_PARLIAMENT_DAO_NOT_EXISTS);
        smart_table::to_simple_map(&smart_table::borrow(&borrow_global<DAO>(dao_address).proposals, proposal_id).casted_votes)
    }
```

aptos-move/framework/aptos-framework/sources/voting.move
0xbe1: i can follow up in a separate PR to change all of them

aptos-move/framework/aptos-framework/sources/voting.move
```diff
@@ -614,6 +642,16 @@ module aptos_framework::voting {
         proposal.is_resolved
     }
 
+    #[view]
+    public fun get_resolution_time<ProposalType: store>(
+        voting_forum_address: address,
+        proposal_id: u64,
+    ): u64 acquires VotingForum {
+        let voting_forum = borrow_global<VotingForum<ProposalType>>(voting_forum_address);
```
movekevin, line 650: Can you add a separate function for fetching a proposal by id and replace all places where this is done with that function? These 2 lines are repeated in these view functions and elsewhere in the file also so this would be a nice cleanup :)

movekevin (approved):

aptos-move/framework/aptos-framework/sources/voting.move
0xbe1: @movekevin done

aptos-move/framework/aptos-framework/sources/voting.move
0xbe1: done

aptos-move/framework/aptos-framework/sources/voting.move
```diff
@@ -632,6 +657,14 @@ module aptos_framework::voting {
         timestamp::now_seconds() > proposal.expiration_secs
     }
 
+    inline fun get_proposal<ProposalType: store>(
```
movekevin, line 660: Let's make this a normal fun to reduce bytecode size. Inline functions save some gas (avoid 1 function call) but bloats bytecode size as it's expanded at every call site. With 5+ callsites, the disadvantage starts to make it not worth it :)

movekevin (approved):

aptos-move/framework/aptos-framework/sources/voting.move
0xbe1: @movekevin fixed

PR #10291, opened 2023-09-28
[ts-sdk-v2] Possible static deserialize

gregnazario:
### Description
A possible way of making BCS deserialize a static function instead of instance.  Stacked on top of https://github.com/aptos-labs/aptos-core/pull/10290 and purposely separated due to some non-ideal tradeoffs


gregnazario:
Closing, this doesn't work

PR #10292, opened 2023-09-28
[move-model] fix internal assertion violation in definition analysis

### LINKED ISSUE #9064: [Bug] Internal assert failed in move-model
# 🐛 Bug

see https://github.com/move-language/move/issues/768

aalok-t:
### Description

Fix failing internal assert in move-model #9064. The assert is triggered because of duplicate declarations of a function name during definition analysis. This is fixed by raising an error explaining the duplicate declaration instead of an assertion violation. Close #9064 

### Test Plan
`move-prover/tests/sources/functional/duplicate_function_declarations.move`

### DIFF
--- a/third_party/move/move-model/src/builder/module_builder.rs
+++ b/third_party/move/move-model/src/builder/module_builder.rs
@@ -43,7 +43,7 @@ use move_bytecode_source_map::source_map::SourceMap;
 use move_compiler::{
     compiled_unit::{FunctionInfo, SpecInfo},
     expansion::ast as EA,
-    parser::ast as PA,
+    parser::{ast as PA, ast::FunctionName},
     shared::{unique_map::UniqueMap, Identifier, Name},
 };
 use move_ir_types::{ast::ConstantName, location::Spanned};
@@ -830,15 +830,14 @@ impl<'env, 'translator> ModuleBuilder<'env, 'translator> {
         let mut visited = BTreeMap::new();
         for (idx, (name, _)) in module_def.functions.key_cloned_iter().enumerate() {
             let is_pure = self.propagate_function_impurity(&mut visited, SpecFunId::new(idx));
-            let full_name = self.qualified_by_module_from_name(&name.0);
             if is_pure {
                 // Modify the types of parameters, return values and expressions
                 // of pure Move functions so they no longer have references.
-                self.deref_move_fun_types(full_name.clone(), idx);
+                self.deref_move_fun_types(name, idx);
             }
             self.parent
                 .fun_table
-                .entry(full_name)
+                .entry(self.qualified_by_module_from_name(&name.0))
                 .and_modify(|e| e.is_pure = is_pure);
         }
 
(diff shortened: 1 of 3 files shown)


third_party/move/move-model/src/builder/module_builder.rs
```diff
@@ -1319,16 +1321,34 @@ impl<'env, 'translator> ModuleBuilder<'env, 'translator> {
...
+        name_loc: &Loc,
+        name_value: move_symbol_pool::Symbol,
+    ) {
+        if let Some(entry) = self.parent.spec_fun_table.get_mut(&full_name) {
+            if entry.is_empty() {
+                self.parent.error(
+                    name_loc,
+                    &format!("missing declaration of `{}`", name_value),
+                );
+            } else if entry.len() > 1 {
```
rahxephon89, line 1337: You could try to remove some code duplication by doing something like:
```                
                let error_msg = if entry.is_empty() {
                    "missing"
                } else {
                    "duplicate"
                };
                self.parent.error(
                    name_loc,
                    &format!("{} declaration of `{}` ", error_msg, name_value),
                );
```

rahxephon89 (changes requested):

third_party/move/move-model/src/builder/module_builder.rs
aalok-t: Done, thanks!

third_party/move/move-model/src/builder/module_builder.rs
```diff
@@ -1319,16 +1321,34 @@ impl<'env, 'translator> ModuleBuilder<'env, 'translator> {
...
-                .map(|Parameter(n, ty)| Parameter(*n, ty.skip_reference().clone()))
-                .collect_vec();
-            e[0].result_type = e[0].result_type.skip_reference().clone();
-        });
+    fn deref_move_fun_types(
+        &mut self,
+        full_name: QualifiedSymbol,
+        spec_fun_idx: usize,
+        name_loc: &Loc,
+        name_value: move_symbol_pool::Symbol,
```
rahxephon89, line 1329: Maybe just use full_name in the error message? Then you don't need name_value.

junkil-park (commented):
@aalok-t , could we add the reported test case to `move-prover/tests/sources/`? So, we can see that the case is fixed by your commit?

third_party/move/move-model/src/builder/module_builder.rs
aalok-t: `full_name` (QualifiedSymbol) does not have a value. An alternative is to pass the function to `deref_move_fun_types` but that creates some issues with borrowing.

rahxephon89: How about using `full_name.symbol` to retrieve name?

aalok-t: full_name.symbol would output the symbol, for example "Symbol(65)" instead of the name of the function. There is no way to extract the function name from the QualifiedSymbol AFAIK.

rahxephon89: I see, could you try this: `self.symbol_pool().string(full_name.symbol).as_str()` or something similar?

third_party/move/move-prover/tests/sources/functional/duplicate_function_declarations.move
```diff
@@ -0,0 +1,15 @@
+module 0x42::DuplicateFunction {
+    struct R0 { x: u8 }
+
+    spec R0 {
+        fun double(x: u8): u8 {
```
junkil-park, line 5: Do you think this duplicated function is necessary to reproduce the problem? Otherwise, can we remove it to avoid the error above?

third_party/move/move-model/src/builder/module_builder.rs
aalok-t: I just passed `name: FunctionName` instead and that simplifies the code.

third_party/move/move-prover/tests/sources/functional/duplicate_function_declarations.move
aalok-t: The duplicated function is neccessary to trigger the error.

rahxephon89 (approved):
Thanks!

junkil-park (approved):
LGTM. Thank you for fixing the bug, @aalok-t !

PR #10296, opened 2023-09-28
[ts-sdk] fix async queue promise dequeue and cancellation

maykonmichel:
### Description

The AsyncQueue is utilized by the TransactionWorker and exhibited a bug when initiated with an empty transactions queue. To address this issue, I could just increase the counter *after* setting the resolveMap. Additionally, the cancellation behavior was enhanced to throw an AsyncQueueCancelledError instead of resolving to it. Furthermore, an unusual behavior was identified where the cancellation status persisted indefinitely. In addition to these fixes, several opinionated changes have been made to enhance the overall functionality and maintainability of the code.

### Test Plan

These implemented tests not only validate the existing functionality but also ensure that recent fixes and enhancements are working as expected.

- **Enqueue and Dequeue**: Verify that items can be enqueued and dequeued as expected, maintaining the queue's order.

- **Handle Dequeue Before Queue**: Ensure that the AsyncQueue can handle dequeuing items even before they are enqueued.

- **Cancellation**: Confirm that the cancel method correctly rejects promises with AsyncQueueCancelledError and updates the cancellation status.

- **Cancellation with Empty Queue**: Verify that cancellation does not cause errors when the queue is empty.

- **Check Queue Emptiness**: Validate the isEmpty method's accuracy in determining whether the queue is empty.

- **Remove Cancelled Status**: Check if the cancelled status is reset after a new item is enqueued.

### DIFF
--- a/ecosystem/typescript/sdk/src/transactions/tests/async_queue.test.ts
+++ b/ecosystem/typescript/sdk/src/transactions/tests/async_queue.test.ts
@@ -1,0 +1,88 @@
+import { AsyncQueue, AsyncQueueCancelledError } from "../async_queue";
+
+describe("asyncQueue", () => {
+  it("should enqueue and dequeue items", async () => {
+    const asyncQueue = new AsyncQueue<number>();
+
+    asyncQueue.enqueue(1);
+
+    const item1 = await asyncQueue.dequeue();
+
+    asyncQueue.enqueue(2);
+    asyncQueue.enqueue(3);
+
+    const item2 = await asyncQueue.dequeue();
+    const item3 = await asyncQueue.dequeue();
+
+    expect(item1).toBe(1);
+    expect(item2).toBe(2);
+    expect(item3).toBe(3);
+  });
+
+  it("should handle dequeue before queue", async () => {
+    const asyncQueue = new AsyncQueue<number>();
+
+    const itemPromise1 = asyncQueue.dequeue();
+    const itemPromise2 = asyncQueue.dequeue();
+
+    expect(asyncQueue.pendingDequeueLength()).toBe(2);
+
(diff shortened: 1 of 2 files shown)


ecosystem/typescript/sdk/src/transactions/tests/async_queue.test.ts
```diff
@@ -0,0 +1,80 @@
+import { AsyncQueue, AsyncQueueCancelledError } from "../async_queue";
+
+describe("asyncQueue", () => {
+  it("should enqueue and dequeue items as a normal queue", async () => {
```
0xmaayan, line 4:
```suggestion
  it("should enqueue and dequeue items", async () => {
```

ecosystem/typescript/sdk/src/transactions/tests/async_queue.test.ts
```diff
@@ -0,0 +1,80 @@
...
+    asyncQueue.enqueue(2);
+
+    const item1 = await itemPromise1;
+    const item2 = await itemPromise2;
+
+    expect(item1).toBe(1);
+    expect(item2).toBe(2);
+  });
+
+  it("should handle cancellation", async () => {
```
0xmaayan, line 38: can we also test that the pendingQueue is empty after cancel is called?

0xmaayan (approved):
LGTM. thanks for improving it!

ecosystem/typescript/sdk/src/transactions/tests/async_queue.test.ts
maykonmichel: Sure, great idea!

sallem-aptos (approved):

PR #10303, opened 2023-09-28
Allow using skip_index_and_usage on state sync code path.

grao1991:
### Test Plan
Successfully synced a devnet PFN from genesis.


msmouse (approved):
Tricky and neat.

Add a TODO wrt that truncate-tool needs to understand usage might not be present at all versions.

grao1991:
> Tricky and neat.
> 
> Add a TODO wrt that truncate-tool needs to understand usage might not be present at all versions.

Done

zekun000 (approved):

PR #10307, opened 2023-09-29
[TS-SDK v2] Updating the `Deserializable` interface and making `Serializable` an abstract class

xbtmatt:
### Description

1. I've removed the export from `Deserializable` in order to facilitate using static `deserialize` methods with Deserializable values.

Typescript doesn't let us use `abstract static` methods, nor does it let us use `static` methods with interfaces, so this is the only solution I could come up with that's simple enough. Note that `Deserializable` is really only meant to facilitate the `Deserializer` instance method `deserialize()`, so it's fine if it's purely used as an interface to describe an object as opposed to being explicitly extended.

2. I also changed `Serializable` to a class instead of an interface and added the `bcsToBytes()` function to it, that way whenever a class `extends Serializable` you can just call `classInstance.bcsToBytes()` on the instance to serialize it and get the bcs serialized bytes.

3. I also removed the abstract `deserialize` functions from the public/private key classes, because you can't declare an `abstract static` method, so it wouldn't work anyway. Unfortunately typescript has quite a few limitations in this regard.

4. Also updated some documentation

### Test Plan
`pnpm jest`

### DIFF
--- a/ecosystem/typescript/sdk_v2/src/bcs/serializer.ts
+++ b/ecosystem/typescript/sdk_v2/src/bcs/serializer.ts
@@ -12,8 +12,23 @@ import {
 } from "./consts";
 import { AnyNumber, Uint16, Uint32, Uint8 } from "../types";
 
-export interface Serializable {
-  serialize(serializer: Serializer): void;
+// This class is intended to be used as a base class for all serializable types.
+// It can be used to facilitate composable serialization of a complex type and
+// in general to serialize a type to its BCS representation.
+export abstract class Serializable {
+  abstract serialize(serializer: Serializer): void;
+
+  /**
+   * Serializes a `Serializable` value to its BCS representation.
+   * This function is the Typescript SDK equivalent of `bcs::to_bytes` in Move.
+   * @param value The Serializable value to serialize
+   * @returns the byte buffer BCS representation of the value
+   */
+  bcsToBytes(): Uint8Array {
+    const serializer = new Serializer();
+    this.serialize(serializer);
+    return serializer.toUint8Array();
+  }
 }
 
 export class Serializer {
@@ -235,9 +250,9 @@ export class Serializer {
    *
    * @example
    * // Define the MoveStruct class that implements the Serializable interface
-   * class MoveStruct implements Serializable {
+   * class MoveStruct extends Serializable {
    *     constructor(
-   *         public creatorAddress: AccountAddress, // where AccountAddress implements Serializable
+   *         public creatorAddress: AccountAddress, // where AccountAddress extends Serializable
    *         public collectionName: string,
    *         public tokenName: string
    *     ) {}
--- a/ecosystem/typescript/sdk_v2/src/bcs/serializer.ts
+++ b/ecosystem/typescript/sdk_v2/src/bcs/serializer.ts
@@ -21,8 +21,7 @@ export abstract class Serializable {
   /**
    * Serializes a `Serializable` value to its BCS representation.
    * This function is the Typescript SDK equivalent of `bcs::to_bytes` in Move.
-   * @param value The Serializable value to serialize
-   * @returns the byte buffer BCS representation of the value
+   * @returns the BCS representation of the Serializable instance as a byte buffer
    */
   bcsToBytes(): Uint8Array {
     const serializer = new Serializer();
--- a/ecosystem/typescript/sdk_v2/src/crypto/asymmetric_crypto.ts
+++ b/ecosystem/typescript/sdk_v2/src/crypto/asymmetric_crypto.ts
@@ -28,8 +28,15 @@ export abstract class PublicKey extends Serializable {
   abstract serialize(serializer: Serializer): void;
 }
 
+/**
+ * An abstract representation of a private key.  This is used to sign transactions and
+ * derive the public key associated.
+ */
 export abstract class PrivateKey extends Serializable {
-  // Sign the given message with the private key.
+  /**
+   * Sign a message with the key
+   * @param args
+   */
   abstract sign(args: { message: HexInput }): Signature;
 
   /**
@@ -50,8 +57,14 @@ export abstract class PrivateKey extends Serializable {
   abstract publicKey(): PublicKey;
 }
 
+/**
+ * An abstract representation of a signature.  This is the product of signing a
+ * message and can be used with the PublicKey to verify the signature.
+ */
 export abstract class Signature extends Serializable {
-  // Convert the signature to bytes or Uint8Array.
+  /**
+   * Get the raw signature bytes
+   */
   abstract toUint8Array(): Uint8Array;
 
   /**
--- a/ecosystem/typescript/sdk_v2/src/crypto/asymmetric_crypto.ts
+++ b/ecosystem/typescript/sdk_v2/src/crypto/asymmetric_crypto.ts
@@ -1,14 +1,14 @@
 // Copyright © Aptos Foundation
 // SPDX-License-Identifier: Apache-2.0
 
-import { Deserializable, Deserializer, Serializable, Serializer } from "../bcs";
+import { Serializable, Serializer } from "../bcs";
 import { HexInput } from "../types";
 
 /**
  * An abstract representation of a public key.  All Asymmetric key pairs will use this to
  * verify signatures and for authentication keys.
  */
-export abstract class PublicKey implements Serializable, Deserializable<PublicKey> {
+export abstract class PublicKey extends Serializable {
   /**
    * Verifies that the private key associated with this public key signed the message with the given signature.
    * @param args
@@ -25,21 +25,11 @@ export abstract class PublicKey implements Serializable, Deserializable<PublicKe
    */
   abstract toString(): string;
 
-  // TODO: This should be a static method.
-  abstract deserialize(deserializer: Deserializer): PublicKey;
-
   abstract serialize(serializer: Serializer): void;
 }
 
-/**
- * An abstract representation of a private key.  This is used to sign transactions and
- * derive the public key associated.
- */
-export abstract class PrivateKey implements Serializable, Deserializable<PrivateKey> {
-  /**
-   * Sign a message with the key
-   * @param args
-   */
+export abstract class PrivateKey extends Serializable {
+  // Sign the given message with the private key.
   abstract sign(args: { message: HexInput }): Signature;
 
   /**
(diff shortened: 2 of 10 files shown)


ecosystem/typescript/sdk_v2/src/crypto/asymmetric_crypto.ts
```diff
@@ -63,23 +50,14 @@ export abstract class PrivateKey implements Serializable, Deserializable<Private
...
-/**
- * An abstract representation of a signature.  This is the product of signing a
- * message and can be used with the PublicKey to verify the signature.
- */
-export abstract class Signature implements Serializable, Deserializable<Signature> {
-  /**
-   * Get the raw signature bytes
-   */
+export abstract class Signature extends Serializable {
+  // Convert the signature to bytes or Uint8Array.
```
gregnazario, line 54: Let's get this back in here, as I added doc comments

ecosystem/typescript/sdk_v2/src/crypto/asymmetric_crypto.ts
```diff
@@ -25,21 +25,11 @@ export abstract class PublicKey implements Serializable, Deserializable<PublicKe
...
 
-/**
- * An abstract representation of a private key.  This is used to sign transactions and
- * derive the public key associated.
- */
-export abstract class PrivateKey implements Serializable, Deserializable<PrivateKey> {
-  /**
-   * Sign a message with the key
-   * @param args
-   */
```
gregnazario, line 42: Similarly

gregnazario (approved):
Looks generally good, see if you can go back and fix some comments disappeared in the rebase

ecosystem/typescript/sdk_v2/src/crypto/asymmetric_crypto.ts
```diff
@@ -1,14 +1,14 @@
...
-import { Deserializable, Deserializer, Serializable, Serializer } from "../bcs";
+import { Serializable, Serializer } from "../bcs";
 import { HexInput } from "../types";
 
 /**
  * An abstract representation of a public key.  All Asymmetric key pairs will use this to
  * verify signatures and for authentication keys.
  */
-export abstract class PublicKey implements Serializable, Deserializable<PublicKey> {
+export abstract class PublicKey extends Serializable {
```
0xjinn, line 11: I see! In this case, the implementation of the method `deserialize` won't be enforced for child classes?

I wonder if theres a way to just implement the deserialize method directly in abstract/base class. Then all child classes will auto inherit it

ecosystem/typescript/sdk_v2/src/bcs/serializer.ts
```diff
@@ -12,8 +12,23 @@ import {
...
+// This class is intended to be used as a base class for all serializable types.
+// It can be used to facilitate composable serialization of a complex type and
+// in general to serialize a type to its BCS representation.
+export abstract class Serializable {
+  abstract serialize(serializer: Serializer): void;
+
+  /**
+   * Serializes a `Serializable` value to its BCS representation.
+   * This function is the Typescript SDK equivalent of `bcs::to_bytes` in Move.
+   * @param value The Serializable value to serialize
```
0xmaayan, line 24: might be confusing saying the this function has a `value` param but irl it doesnt accept any argument, can we explain it better?

0xmaayan (approved):

ecosystem/typescript/sdk_v2/src/crypto/asymmetric_crypto.ts
xbtmatt: oops that's my bad, thought I incorporated the new ones when merging

ecosystem/typescript/sdk_v2/src/crypto/asymmetric_crypto.ts
xbtmatt: Yeah I looked into it and there's apparently a lot of hullabaloo about static methods in abstract classes, so the support there, particularly with typescript, is lacking.

I tried a few different methods to achieve a `static deserialize` method interface inherited from the abstract class but it came down to either being really janky or just not possible.

WRT what you're saying though, the problem with implementing the deserialize method in the base class (say, as an abstract factory) is that it must know ahead of time how to create the child classes. I don't know if this would be a problem, it depends on how we plan on implementing these abstract classes later.

For example here is the `TransactionAuthenticator` abstract factory pattern:

```typescript
export abstract class TransactionAuthenticator {
  abstract serialize(serializer: Serializer): void;

  static deserialize(deserializer: Deserializer): TransactionAuthenticator {
    const index = deserializer.deserializeUleb128AsU32();
    switch (index) {
      case 0:
        return TransactionAuthenticatorEd25519.load(deserializer);
      case 1:
        return TransactionAuthenticatorMultiEd25519.load(deserializer);
      case 2:
        return TransactionAuthenticatorMultiAgent.load(deserializer);
      case 3:
        return TransactionAuthenticatorFeePayer.load(deserializer);
      default:
        throw new Error(`Unknown variant index for TransactionAuthenticator: ${index}`);
    }
  }
}
```

But with this, the child classes actually don't even implement deserialize, they specifically have a `load` function they use.

The other part to this is that these are specifically `enum`s in Rust, so that's why they're serialized/deserialized like this.

With the `PublicKey` stuff, it's not an enum in rust, it's merely an abstract class we're using for the SDK, so we wouldn't be serializing an enum value (`0`, `1`, etc) at the beginning of the byte buffer to differentiate between child class implementations.

ecosystem/typescript/sdk_v2/src/bcs/serializer.ts
```diff
@@ -235,9 +250,9 @@ export class Serializer {
    *
    * @example
    * // Define the MoveStruct class that implements the Serializable interface
-   * class MoveStruct implements Serializable {
+   * class MoveStruct extends Serializable {
    *     constructor(
-   *         public creatorAddress: AccountAddress, // where AccountAddress implements Serializable
+   *         public creatorAddress: AccountAddress, // where AccountAddress extends Serializable
```
0xmaayan, line 255: maybe also include `extends Serializable` for our [AccountAddress](https://github.com/aptos-labs/aptos-core/blob/main/ecosystem/typescript/sdk_v2/src/core/account_address.ts#L37) class as part of this PR?

ecosystem/typescript/sdk_v2/src/bcs/serializer.ts
xbtmatt: Yeah I think that was a copy paste error 😅

ecosystem/typescript/sdk_v2/src/bcs/serializer.ts
xbtmatt: Ah yes, done!

ecosystem/typescript/sdk_v2/tests/unit/deserializer.test.ts
```diff
@@ -132,13 +132,15 @@ describe("BCS Deserializer", () => {
 
   it("deserializes a single deserializable class", () => {
     // Define the MoveStruct class that implements the Deserializable interface
-    class MoveStruct implements Serializable {
+    class MoveStruct extends Serializable {
```
0xmaayan, line 135: the whole point with deserializable and serializable is to later create a Move type class (MoveU8, MoveU16, etc) and have each of them to be deserializable and serializable? correct? I wonder how we can achieve it with this structure as in typescript a class can only extends one class.

ecosystem/typescript/sdk_v2/tests/unit/multi_ed25519.test.ts
```diff
@@ -119,12 +119,12 @@ describe("MultiSignature", () => {
   it("should throws exception when creating a bitmap with wrong bits", async () => {
     expect(() => {
       MultiEd25519Signature.createBitmap({ bits: [32] });
-    }).toThrow("Cannot have a signature larger than 31.");
```
0xmaayan, line 122: that was changed to this in https://github.com/aptos-labs/aptos-core/pull/10290/files#diff-fa35ec805433effb2c47fca68711344abbcbc0f5f15786cae18b1db7143e454dR122

ecosystem/typescript/sdk_v2/tests/unit/deserializer.test.ts
xbtmatt: The point of the Serializable class is partially to encourage developers to use a more intuitive structure for BCS-serializable data and also to allow for composable serialization. Plus the `bcsToBytes` function

The deserialization aspect of things is mostly a separate process in the sense that it doesn't require an instance of a Deserializable class type to deserialize a byte buffer, since it just uses the static deserialize function. That's why it can be an internal interface, because it doesn't actually do anything. The point of `Deserializable` is to be an interface to ensure that when you call `deserializer.deserialize(something)` something is actually deserializable

But we don't necessarily need Serializable to be a class, I suppose, if you think that will lead to issues down the road. It could just remain an interface

ecosystem/typescript/sdk_v2/tests/unit/multi_ed25519.test.ts
xbtmatt: argh I thought I changed this, guess I accidentally left it in, my bad

xbtmatt:
So I've played around with this a ton and basically come to the conclusion that neither `Deserializable` nor `Serializable` need to be need to be classes.

And `Deserializable` in particular doesn't even need to be an interface, especially because the support for interfaces with static methods is...not good.

This is what we'd have to do to get it to work:

```typescript
/**
 * An interface that exists primarily to denote that a type is BCS-deserializable.
 */
export interface IDeserializable {
  new(...args: any[]): any;
  /**
   * @param deserializer An instance of Deserializer, where the BCS-serialized bytes to be deserialized are stored.
   * @returns An instance of the type that implements this interface.
   */
  deserialize(deserializer: Deserializer): any;
}

export type StaticImplements<I extends new (...args: any[]) => any, C extends I> = InstanceType<I>;

export class AccountAddress extends Serializable implements StaticImplements<Deserializable, typeof AccountAddress> {
```

which seems like way more effort than it's worth, especially since `Deserializable` was mainly meant to be an indicator of what you can do with the class- not necessarily because we need it for anything.

ecosystem/typescript/sdk_v2/tests/unit/deserializer.test.ts
xbtmatt: @0xmaayan I'm going to keep it simple and just leave `Serializable` as a class for now, but to be clear, `Deserializable` is an internal interface that exists solely to facilitate the more intuitive syntax with `Deserializer` instances; that is:

```typescript
const myClassInstance = deserializer.deserialize(MyClass);
// as opposed to
const myClassInstance = MyClass.deserialize(deserializer);
```

PR #10308, opened 2023-09-29
[ts-sdk-v2] Add MIME types, and convert input types from string to Enum

gregnazario:
### Description
Adds mime types, framework for allowing BCS in transaction submission and in API output.

### Test Plan
`pnpm fmt && pnpm test`


heliuchuan (approved):

ecosystem/typescript/sdk_v2/src/client/core.ts
```diff
@@ -33,7 +33,7 @@ async function request<Req, Res>(
   const headers: Record<string, string | string[] | undefined> = {
     ...overrides?.HEADERS,
     "x-aptos-client": `aptos-ts-sdk/${VERSION}`,
-    "content-type": contentType ?? "application/json",
+    "content-type": contentType ?? MimeType.JSON.valueOf(),
```
0xmaayan, line 36: why we need the `valueOf`?

ecosystem/typescript/sdk_v2/src/client/get.ts
```diff
@@ -54,7 +58,8 @@ export async function get<Req, Res>(options: GetRequestOptions): Promise<AptosRe
       method: "GET",
       originMethod,
       path,
-      contentType,
+      contentType: contentType?.valueOf(),
```
0xmaayan, line 61: also here?

0xmaayan (approved):

ecosystem/typescript/sdk_v2/src/client/core.ts
gregnazario: Does it automatically convert to string?  I don't know how enums work in typescript necessarily

0xmaayan: I think having `MimeTime.JSON` is enough to output the string...?

PR #10309, opened 2023-09-29
[move unit tests] Run extended checker as part of unit tests

### LINKED ISSUE #9251: [Bug] aptos-move unit tests are run without extended_checks

wrwg:
### Summary

Closes #9251

This runs the extended checker as part of Aptos unit tests (either our own Rust integrated tests or from the CLI). It uses the same technique as we already used for native extensions specific to Aptos: a hook is defined where additional, move-model based validations can be run. This is hook is then connected to the extended checker when running Aptos tests.

The implementation also optimizes the construction of the move model: if that one is already needed by abi generation (which is the default), it is not constructed a 2nd time for the extended checker -- both  for the existing build step and the new test step. This should avoid one full additional compilation (source -> bytecode -> model run).

Addendum: Extended checks until now excluded test code, leading to wrong usage of entry functions and attributes marked as test-only. Because fixing this is a breaking change, this PR adds the behavior to check test code via a new CLI option `--check-test-code`. This flag should eventually become default behavior.

### Test Plan

Manual

### DIFF
--- a/aptos-move/framework/src/built_package.rs
+++ b/aptos-move/framework/src/built_package.rs
@@ -72,6 +72,8 @@ pub struct BuildOptions {
     pub compiler_version: Option<CompilerVersion>,
     #[clap(long)]
     pub skip_attribute_checks: bool,
+    #[clap(long)]
+    pub check_test_code: bool,
     #[clap(skip)]
     pub known_attributes: BTreeSet<String>,
 }
@@ -96,6 +98,7 @@ impl Default for BuildOptions {
             bytecode_version: None,
             compiler_version: None,
             skip_attribute_checks: false,
+            check_test_code: false,
             known_attributes: extended_checks::get_all_attribute_names().clone(),
         }
     }
@@ -126,6 +129,7 @@ pub fn build_model(
         generate_abis: false,
         generate_docs: false,
         generate_move_model: false,
+        full_model_generation: false,
         install_dir: None,
         test_mode: false,
         force_recompilation: false,
@@ -160,6 +164,7 @@ impl BuiltPackage {
             generate_abis: options.with_abis,
             generate_docs: false,
             generate_move_model: true,
+            full_model_generation: options.check_test_code,
             install_dir: options.install_dir.clone(),
             test_mode: false,
             force_recompilation: false,
@@ -177,8 +182,7 @@ impl BuiltPackage {
         let (mut package, model_opt) =
             build_config.compile_package_no_exit(&package_path, &mut stderr())?;
 
-        // Build the Move model for extra processing and run extended checks as well derive
-        // runtime metadata
+        // Run extended checks as well derive runtime metadata
         let model = &model_opt.expect("move model");
         let runtime_metadata = extended_checks::run_extended_checks(model);
         if model.diag_count(Severity::Warning) > 0 {
--- a/aptos-move/framework/src/built_package.rs
+++ b/aptos-move/framework/src/built_package.rs
@@ -125,6 +125,7 @@ pub fn build_model(
         architecture: None,
         generate_abis: false,
         generate_docs: false,
+        generate_move_model: false,
         install_dir: None,
         test_mode: false,
         force_recompilation: false,
@@ -158,6 +159,7 @@ impl BuiltPackage {
             architecture: None,
             generate_abis: options.with_abis,
             generate_docs: false,
+            generate_move_model: true,
             install_dir: options.install_dir.clone(),
             test_mode: false,
             force_recompilation: false,
@@ -172,20 +174,12 @@ impl BuiltPackage {
         };
 
         eprintln!("Compiling, may take a little while to download git dependencies...");
-        let mut package = build_config.compile_package_no_exit(&package_path, &mut stderr())?;
+        let (mut package, model_opt) =
+            build_config.compile_package_no_exit(&package_path, &mut stderr())?;
 
         // Build the Move model for extra processing and run extended checks as well derive
         // runtime metadata
-        let model = &build_model(
-            options.dev,
-            package_path.as_path(),
-            options.named_addresses.clone(),
-            None,
-            bytecode_version,
-            compiler_version,
-            skip_attribute_checks,
-            options.known_attributes.clone(),
-        )?;
+        let model = &model_opt.expect("move model");
         let runtime_metadata = extended_checks::run_extended_checks(model);
         if model.diag_count(Severity::Warning) > 0 {
             let mut error_writer = StandardStream::stderr(ColorChoice::Auto);
--- a/third_party/move/tools/move-cli/src/base/test_validation.rs
+++ b/third_party/move/tools/move-cli/src/base/test_validation.rs
@@ -1,0 +1,33 @@
+// Copyright (c) The Diem Core Contributors
+// Copyright (c) The Move Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+//! This module manages validation of the unit tests, in addition to standard compiler
+//! checking.
+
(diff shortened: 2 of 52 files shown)


aptos-move/framework/src/built_package.rs
```diff
@@ -172,20 +174,12 @@ impl BuiltPackage {
         };
 
         eprintln!("Compiling, may take a little while to download git dependencies...");
-        let mut package = build_config.compile_package_no_exit(&package_path, &mut stderr())?;
+        let (mut package, model_opt) =
+            build_config.compile_package_no_exit(&package_path, &mut stderr())?;
 
         // Build the Move model for extra processing and run extended checks as well derive
```
vineethk, line 180: Comment no longer applicable at this location.

vineethk (approved):
Could you briefly comment on the impact this has on unit test time?

lgtm, other than a minor comment.

third_party/move/tools/move-cli/src/base/test_validation.rs
```diff
@@ -0,0 +1,33 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+//! This module manages validation of the unit tests, in addition to standard compiler
+//! checking.
+
+use move_model::model::GlobalEnv;
+use once_cell::sync::Lazy;
+use std::sync::Mutex;
+
+static VALIDATION_HOOK: Lazy<Mutex<Option<Box<dyn Fn(&GlobalEnv) + Send + Sync>>>> =
```
brmataptos, line 12: This changeable global state is pretty horrible.  Can you avoid it?

We already have problems building/testing in parallel, and this just adds another reason why tests might not behave well.

Can't this info be conveyed in a parameter to the tests somehow?

wrwg:
> Could you briefly comment on the impact this has on unit test time?
> 
> lgtm, other than a minor comment.

When running tests from the CLI, extended checks and model building where performed in an extra step. This is now removed so its a zero-sum game here. (But faster in the `move compile` case)

third_party/move/tools/move-cli/src/base/test_validation.rs
wrwg: I know it is horrible, but as said in the PR description, it is replicating an existing hack as [seen here](https://github.com/aptos-labs/aptos-core/blob/a1ae60836f987cacb6ed3f731e585df7c684c67e/third_party/move/tools/move-unit-test/src/extensions.rs#L19). One big problem with extending the UT package taking an external trait is that all configuration options have to be serializable, and so you can't just add a `dyn` to this data. I think this serializability is in turn needed for some functionality of the `move-cli` -- dead code, put not easy to remove. This is what I meant with the larger refactoring which need to be done first to do this here right. So this is a shortcut until we have cycles for the bigger thing.

brmataptos:
Please file and link to new issues to
- fix existing tests and enable flag by default
- refactor test code to get rid of this hack and the one in extensions.rs.  (Mention the two specific statics.)

Otherwise, LGTM.

wrwg:
> Please file and link to new issues to
> 
> * fix existing tests and enable flag by default
> * refactor test code to get rid of this hack and the one in extensions.rs.  (Mention the two specific statics.)
> 
> Otherwise, LGTM.

Filed and linked an issue for [#1](https://github.com/aptos-labs/aptos-core/issues/10335). For #2, we do not file issues for this kind of work. We have been talking about the need for the refactoring many times (it is the consolidation of the separation between aptos-move and third_party/move), its a project and not an issue, and we will track it on the roadmap.

third_party/move/tools/move-cli/src/base/test_validation.rs
brmataptos: Ok for now.

brmataptos (approved):
Thanks for recording noted issues so we don't have to rediscover them someday.

aptos-move/framework/src/built_package.rs
```diff
@@ -172,20 +179,11 @@ impl BuiltPackage {
...
-            options.named_addresses.clone(),
-            None,
-            bytecode_version,
-            compiler_version,
-            skip_attribute_checks,
-            options.known_attributes.clone(),
-        )?;
+        // Run extended checks as well derive runtime metadata
+        let model = &model_opt.expect("move model");
         let runtime_metadata = extended_checks::run_extended_checks(model);
```
movekevin, line 187: Is it possible to do these steps during compilation instead of afterward, for example via visitors?

wrwg: Not in the compiler v1 architecture.

PR #10310, opened 2023-09-29
[CLI] Add support for not JSONifying errors

banool:
### Stack
- Previous in stack: https://github.com/aptos-labs/aptos-core/pull/10254
- Next in stack: https://github.com/aptos-labs/aptos-core/pull/10255

### Description
For some subcommands JSONifying the error output doesn't really make sense. I make it possible to opt out.

### Test Plan
See https://github.com/aptos-labs/aptos-core/pull/10255/.

### DIFF
--- a/crates/aptos/src/common/types.rs
+++ b/crates/aptos/src/common/types.rs
@@ -1246,6 +1246,11 @@ pub trait CliCommand<T: Serialize + Send>: Sized + Send {
     /// Returns a name for logging purposes
     fn command_name(&self) -> &'static str;
 
+    /// Returns whether the error should be JSONifyed.
+    fn jsonify_error_output(&self) -> bool {
+        true
+    }
+
     /// Executes the command, returning a command specific type
     async fn execute(self) -> CliTypedResult<T>;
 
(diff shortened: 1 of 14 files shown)


crates/aptos/src/common/types.rs
```diff
@@ -1246,6 +1246,11 @@ pub trait CliCommand<T: Serialize + Send>: Sized + Send {
     /// Returns a name for logging purposes
     fn command_name(&self) -> &'static str;
 
+    /// Returns whether the error should be JSONifyed.
+    fn jsonify_error_output(&self) -> bool {
+        true
+    }
```
0xjinn, line 1252: Maybe I am missing something. Since this is hardcoded to true, how do I opt out to not jsonify the output?

banool: This is the default trait implementation, any CLI command can make this return false instead. This is what I do in the local testnet subcommand.

0xjinn: neato!

0xjinn (approved):

PR #10315, opened 2023-09-29
[Quorum Store] 33% increase in backlog_txn_limit_count

bchocho:
### Description

The previous `backlog_txn_limit_count` was too conservative. Especially when running through PFNs, which cause duplicate txns and these are included in the backlog count. We should optimize for this case, since most txns should go through PFNs in prod.

Higher values of `backlog_txn_limit_count` actually show some VFNs lagging behind in non-PFN tests (due to VNs going full throttle). We will need to think more on what to do in this case.

### Test Plan

Ran ad-hoc test with PFNs, see an increase in TPS (latest run 7K).

### DIFF
--- a/config/src/config/quorum_store_config.rs
+++ b/config/src/config/quorum_store_config.rs
@@ -26,7 +26,7 @@ impl Default for QuorumStoreBackPressureConfig {
     fn default() -> QuorumStoreBackPressureConfig {
         QuorumStoreBackPressureConfig {
             // QS will be backpressured if the remaining total txns is more than this number
-            backlog_txn_limit_count: MAX_SENDING_BLOCK_TXNS_QUORUM_STORE_OVERRIDE * 3,
+            backlog_txn_limit_count: MAX_SENDING_BLOCK_TXNS_QUORUM_STORE_OVERRIDE * 4,
             // QS will create batches at the max rate until this number is reached
             backlog_per_validator_batch_limit_count: 4,
             decrease_duration_ms: 1000,


igor-aptos (approved):

bchocho:
Here's the ad-hoc run: https://github.com/aptos-labs/aptos-core/actions/runs/6382833672

JoshLind (approved):

bchocho:
Oh interesting, `realistic_env_max_load` regresses the high gas transaction enough to fail, even though the overall transactions are faster, resulting in 10K TPS.

bchocho:
@igor-aptos so we have an issue where this is now causing higher latency for high gas transactions. Want to hear what you think about a couple things: 

1. There is a general regression. Now that we are creating 3.2K block sizes, pipeline backpressure is being triggered 40% of the time. Should the block size be restricted further, as block time is over 1s?
2. There is a p90+ regression. I'm thinking this is now reflecting the behavior around epoch boundaries, which are triggered every 5 mins. Should we avoid triggering epoch so often?

bchocho:
I'm going to land this with a conservative increase of 33%. I'll have to think harder about what to do with VFNs falling behind when VNs go full throttle, which was observed with more aggressive increases.

Here are the latest test results:
* PFN: https://github.com/aptos-labs/aptos-core/actions/runs/6410569429
* non-PFN: https://github.com/aptos-labs/aptos-core/actions/runs/6410568263

PR #10316, opened 2023-09-29
Revert "WarmVmCache"

msmouse:
This reverts commit 3b99bb75aa68c48814559aafbc3a351c64b45b37.

discussion on https://github.com/aptos-labs/aptos-core/pull/10147


vgao1996:
So it looks like we're just doing a partial revert?

msmouse:
@vgao1996  yeah I plan to put it back, this should be the only part that's relevant to the bug

grao1991 (approved):

msmouse:
okay let's fix forward instead

PR #10320, opened 2023-09-29
[Sharded-Execution] Fix a race condition while fetching the state values on a shard from a remote stateview

manudhundi:


msmouse:
why not insert before sending the request, or remove at all?

manudhundi:
> why not insert before sending the request, or remove at all?

(1) Not insert before sending the request because that reduces parallelism. Currently when we insert the keys, we also have job(s) making the remote requests in parallel.
(2) If we don't at all insert during send request, then we will end up making multiple remote calls during get_state_value().

Reason (2) is more critical than reason (1)

sitalkedia:
> > why not insert before sending the request, or remove at all?
> 
> (1) Not insert before sending the request because that reduces parallelism. Currently when we insert the keys, we also have job(s) making the remote requests in parallel. (2) If we don't at all insert during send request, then we will end up making multiple remote calls during get_state_value().
> 
> Reason (2) is more critical than reason (1)

Actually @msmouse has a good point, can we insert the key before sending the fetch request asynchronously? That probably is a cleaner option?

manudhundi:
> > > why not insert before sending the request, or remove at all?
> > 
> > 
> > (1) Not insert before sending the request because that reduces parallelism. Currently when we insert the keys, we also have job(s) making the remote requests in parallel. (2) If we don't at all insert during send request, then we will end up making multiple remote calls during get_state_value().
> > Reason (2) is more critical than reason (1)
> 
> Actually @msmouse has a good point, can we insert the key before sending the fetch request asynchronously? That probably is a cleaner option?

Yes cleaner option indeed. But we are not as parallel as before. That is, we are not leveraging the IO time to do some work like before. But, writing keys to a hashtable in memory is pretty fast, so it will not matter so much.

I will put out a commit as per Alden's proposal

sitalkedia (approved):

sitalkedia (approved):

msmouse (approved):

PR #10328, opened 2023-09-30
reconfiguration with dkg

zjma:
## Context

To generate on-chain randomness, the current validators will together run a DKG protocol right before starting a new epoch. The DKG protocol needs the exact validator set (address + voting power) to be determined for the new epoch, and can take a while. During a DKG, care must be taken for validator set changes and reconfiguration requests.

### Already done
None.

### In this PR
- [x] Add feature flag: `RECONFIGURE_WITH_DKG`.
- [x] A governance proposal should start a slow reconfiguration.
- [x] When the current epoch expires and there is no slow reconfiguration in progress, block prologue should start a slow reconfigure.
- [x] Triggering a slow reconfiguration should fail if another one is in progress.
- [x] The following on-chain configs will have 2 versions: one for the current epoch and one for the next epoch. The switch will happen at the epoch boundary.
  - [x] `ConsensusConfig`
  - [x] `ExecutionConfig`
  - [x] `GasScheduleV2`
  - [x] `Features`
  - [x] `Version`
- [x] When DKG finishes/exceeds its time limit, block prologue should update the "current" version of every on-chain config and emit `NewEpochEvent`.
- [x] Reject any on-chain config change if a slow reconfiguration is in progress.

### Next steps
- [ ] Identify the remaining on-chain configs that need 2 versions.
- [ ] Update prover tests.
- [ ] Adapt `aptos-release-builder` to slow reconfiguration and on-chain config behavior change.
- [ ] Release proposal/migration plan.


aptos-move/framework/aptos-framework/sources/genesis.move
```diff
@@ -365,7 +365,7 @@ module aptos_framework::genesis {
             validator.consensus_pubkey,
             validator.proof_of_possession,
         );
-        stake::update_network_and_fullnode_addresses(
+        stake::force_update_network_and_fullnode_addresses(
```
movekevin, line 368: Why do we need this for genesis?

zjma: some smoke tests set this flag (disable validator set change) which made genesis failed...

zjma:
### Alternatives discussed

We need 2 modes of reconfiguration.

- Fast reconfiguration. This should be used by governance proposals for on-chain config changes (e.g. consensus/execution config changes). High-level breakdown:
  - Update epoch counter and reset epoch timer.
  - Send `NewEpochEvent` to validators.
- Slow reconfiguration with DKG (2-phase). This should be used in epoch expiry handling.
  - (Phase 0)
    - Compute the next validator set, lock it.
    - Notify validators to start DKG.
  - (Phase 1: Once DKG finishes/times out)
    - Update validator set (apply pending join/leave requests, distribute rewards).
    - Update epoch counter and reset epoch timer.
    - Send `NewEpochEvent` to validators.

The key behavior changes to implement.
- [ ] When the current epoch expires, block prologue should trigger a slow reconfiguration.
- [ ] When DKG finishes/times out, block prologue should trigger the phase 1 of slow reconfiguration.
- [ ] A separate slow reconfiguration triggered during DKG should be rejected.
- [ ] A fast reconfiguration invoked during DKG should be executed and cause the in-progress DKG to be abandoned.
- [ ] When starting a new epoch, a validator should do the following.
  - If the DKG output for the new epoch is available, use it for randomness generation.
  - If the DKG output for the new epoch is not available, but the one for the old epoch is, and the validator set doesn't change, use the old one for randomness generation.
  - Otherwise(, it must be an edge case , e.g. DKG timed out or disaster recovery handled by an urgent validator set change + fast reconfiguration), continue the new epoch without randomness.
- [x] regular validator set change requests (join/leave validator set, add/withdraw stakes) during a DKG should either be rejected (implemented in this PR) or queued up.

This PR starts with the last bullet item.

aptos-move/framework/aptos-framework/sources/stake.move
```diff
@@ -427,11 +448,15 @@ module aptos_framework::stake {
...
 
     /// Allow on chain governance to remove validators from the validator set.
+    ///
+    /// This is a system-level validator set change, therefore not affected by user-level locking.
+    /// Caller should ensure a fast reconfiguration is triggered after this.
     public fun remove_validators(
         aptos_framework: &signer,
         validators: &vector<address>,
-    ) acquires ValidatorSet {
+    ) acquires ValidatorSet, ValidatorSetChangeLockStatus {
```
zjma, line 457: @zekun000 @movekevin will this have backward-compatibility issue?

zekun000: I "think" this is fine

aptos-move/framework/aptos-framework/sources/configs/consensus_config.move
```diff
@@ -24,14 +23,13 @@ module aptos_framework::consensus_config {
...
         system_addresses::assert_aptos_framework(account);
         assert!(vector::length(&config) > 0, error::invalid_argument(EINVALID_CONFIG));
+        std::config_for_next_epoch::upsert<ConsensusConfig>(account, ConsensusConfig {config});
+    }
 
-        let config_ref = &mut borrow_global_mut<ConsensusConfig>(@aptos_framework).config;
-        *config_ref = config;
-
-        // Need to trigger reconfiguration so validator nodes can sync on the updated configs.
-        reconfiguration::reconfigure();
```
zjma, line 35: @zekun000 @movekevin it's now the other way around: `terminate_slow_reconfigure()` calls `consensus_config::on_new_epoch()`. The module dependency also changed. Any compatibility issue?

zekun000: the easiest way is probably to run a framework upgrade test to see if it can pass

aptos-move/aptos-gas-schedule-updator/src/lib.rs
```diff
@@ -86,6 +86,11 @@ fn generate_script(gas_schedule: &GasScheduleV2) -> Result<String> {
         "gas_schedule::set_gas_schedule(&framework_signer, gas_schedule_blob);"
     );
 
+    emitln!(
+        writer,
+        "aptos_governance::reconfigure(&framework_signer);"
+    );
+
```
zjma, line 93: In this approach `consensus_config::set` no longer calls `reconfigure`, I think the proposal script should call `reconfigure` explicitly. (Same below for some other configs)

@zekun000 @movekevin @sherry-x could you confirm if these release builder updates are also needed?

aptos-move/framework/aptos-framework/sources/reconfiguration.move
```diff
@@ -151,6 +154,87 @@ module aptos_framework::reconfiguration {
...
+        };
+
+        stake::on_new_epoch();
+        storage_gas::on_reconfig();
+        consensus_config::on_new_epoch(account);
+        execution_config::on_new_epoch(account);
+        gas_schedule::on_new_epoch(account);
+        std::version::on_new_epoch(account);
+        features::on_new_epoch(account);
+        // TODO: complete the list.
```
zjma, line 220: I think the list should contain on-chain states that may be accessed in rust directly, and here are what I have identified. Anything I missed?

@zekun000 @movekevin

aptos-move/framework/aptos-framework/sources/stake.move
movekevin: Nope changing acquires list is not subject to compat checks. Only the function's type arg, params and return types  are

aptos-move/framework/aptos-framework/sources/configs/consensus_config.move
```diff
@@ -24,14 +24,15 @@ module aptos_framework::consensus_config {
     }
 
     /// This can be called by on-chain governance to update on-chain consensus configs.
-    public fun set(account: &signer, config: vector<u8>) acquires ConsensusConfig {
+    public fun set(account: &signer, config: vector<u8>) {
         system_addresses::assert_aptos_framework(account);
         assert!(vector::length(&config) > 0, error::invalid_argument(EINVALID_CONFIG));
+        std::config_for_next_epoch::upsert<ConsensusConfig>(account, ConsensusConfig {config});
+    }
```
zjma, line 31: Today, `consensus_config::set()` calls `reconfiguration::reconfigure()`.
But in our new design, `consensus_config::set(X)` writes `X` to a buffer; then at the end of DKG, reconfiguration calls a new func `consensus_config::on_new_epoch()` to apply `X`.
This means that we can't control `consensus_config::set()` behavior with a feature flag, otherwise compile fails with a circular dependency.

But I think if we also update `aptos-release-builder` accordingly to append a `aptos_governance::reconfigure()` in the proposal script, things sound be fine?

aptos-move/framework/move-stdlib/sources/configs/features.move
```diff
@@ -260,30 +260,67 @@ module std::features {
...
+    /// Apply all the pending feature flag changes. Should only be used at the end of a reconfiguration with DKG.
+    ///
+    /// While the scope is public, it can only be usd in system transactions like `block_prologue` and governance proposals,
+    /// who have permission to set the flag that's checked in `extract()`.
+    public fun on_new_epoch() acquires Features {
+        if (config_for_next_epoch::does_exist<Features>()) {
+            let features = config_for_next_epoch::extract<Features>();
+            *borrow_global_mut<Features>(@std) = features;
+        }
+    }
```
zjma, line 316: `std::features::Features` is a critical on-chain state that i think has to have a pending change buffer (and therefore a `on_new_epoch` function).
What's special is, it's defined in `move-stdlib` package (while `reconfiguration.move` and all the other configs I identified so far are in `move-framework` package). for all other configs' `on_new_epoch()` , we can protect it by setting scope to `public(friend)` and make `reconfiguration.move` a friend.
But for Features, this protection is not available.

The workaround I'm thinking.
- Leave this function `public`.
- Introduce a flag struct `ExtractPermit`.
- Any system txn that may invoke `reconfigure` should put `0x1::ExtractPermit` before the invocation, and remove it after.
- Inside `config_for_next_epoch::extract`, do the work only if the `0x1::ExtractPermit` exists.

zjma: Potential alternative:
- Have a clone of `reconfigure()` that takes a signer: `reconfigure_v2(account: &signer)`.
- Update all reconfigure callers to use `reconfigure_v2`.
- Have `features::on_new_epoch()` requires a signer and assert the signer is 0x0/0x1.

aptos-move/framework/move-stdlib/sources/configs/config_for_next_epoch.move
```diff
@@ -0,0 +1,126 @@
...
+    struct ForNextEpoch<T> has drop, key {
+        payload: T,
+    }
+
+    /// We need to temporarily reject on-chain config changes during DKG.
+    /// `0x0::UpdateLock` or `0x1::UpdateLock`, whichever has the higher `seq_num`, represents whether we should reject.
+    struct UpsertLock has copy, drop, key {
+        seq_num: u64,
+        locked: bool,
+    }
```
zjma, line 22: This comes from the requested behavior: when a slow reconfigure is in progress, on-chain config changes should be rejected.

Ideally we define a struct `T` to represent the lock and put a `0x1::T` when starting DKG and remove it when finishing. But reconfigure can be triggered by either 0x1 (proposal) or 0x0 (epoch expiry), and 0x0 can't put things under 0x1. 2 solutions i can imagine.

1. Define `struct SomeLock { locked: bool }`; run a initialization script to put `0x1::SomeLock`. Later both 0x0 and 0x1 can read/write the `locked` field.
2. (implemented currently in the PR) Have 0x0 and 0x1 maintain a struct under their own address. Each maintain a `seqnum` so we know who is the last write. When reading the lock state, read from the last write.

PR #10329, opened 2023-09-30
[dag] add various counters

zekun000:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -784,9 +784,11 @@ dependencies = [
  "proptest",
  "rand 0.7.3",
  "rayon",
+ "scopeguard",
  "serde",
  "serde_bytes",
  "serde_json",
+ "strum_macros",
  "tempfile",
  "thiserror",
  "tokio",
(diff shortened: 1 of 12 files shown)


msmouse (approved):

consensus/src/counters.rs
```diff
@@ -594,6 +594,26 @@ pub static NUM_TXNS_PER_BLOCK: Lazy<Histogram> = Lazy::new(|| {
     .unwrap()
 });
 
+/// Traces node latency movement throughout the DAG
```
ibalajiarun, line 597: Should we keep this in a separate file? within dag module so its cleaner? is this combined because we have some shared counters?

consensus/src/counters.rs
```diff
@@ -594,6 +594,26 @@ pub static NUM_TXNS_PER_BLOCK: Lazy<Histogram> = Lazy::new(|| {
...
+pub static NODE_TRACING: Lazy<HistogramVec> = Lazy::new(|| {
+    register_histogram_vec!(
```
ibalajiarun, line 599: The default buckets might not be enough for observing timestamps. Speaking of, I am not sure this is the right approach, because timestamps are monotonically increasing, but histogram buckets have a upper bound.

consensus/src/counters.rs
zekun000: I was sharing the round counter initially but probably separate is cleaner

consensus/src/counters.rs
zekun000: this is not for timestamp, it's for the duration between two timestamps

grao1991 (approved):

consensus/src/counters.rs
ibalajiarun: Got it!

PR #10338, opened 2023-10-02
Add CI to check for banned CLI dynamic deps

banool:
### Description
We recently broke the OpenSSL dep and went through great pains to avoid the libpq dep. This CI stops us from inadvertently reintroducing those deps.

`cargo deny` would be good but we’re blocked on features lacking in that crate: https://github.com/EmbarkStudios/cargo-deny/issues/563. So we could just use cargo tree -i <crate> instead. Or build the CLI and use otool (but that is more expensive).

I just use the system cargo here because our rust-setup action only works on Debian systems.

### Test Plan
https://github.com/aptos-labs/aptos-core/actions/runs/6379345050/job/17311640488?pr=10338

### DIFF
--- a/crates/aptos/scripts/check_dynamic_deps.sh
+++ b/crates/aptos/scripts/check_dynamic_deps.sh
@@ -1,0 +1,46 @@
+#!/bin/sh
+
+# This script checks if the CLI depends on external deps that it shouldn't. We run this
+# in CI to make sure we don't accidentally reintroduce deps that would make the CLI
+# unusable on most systems.
+#
+# While it would be more reliable to actually build the CLI and check what libraries it
+# links to, e.g. with otool, it is much cheaper to use cargo tree. As far as I can tell
+# the entire Rust ecosystem makes use of these `x-sys` libraries to depend on external
+# dynamically linked libraries.
+#
+# We can almost use cargo deny but it doesn't support checking specific build paths. We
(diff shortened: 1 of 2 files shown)


crates/aptos/scripts/check_dynamic_deps.sh
```diff
@@ -0,0 +1,46 @@
...
+# dynamically linked libraries.
+#
+# We can almost use cargo deny but it doesn't support checking specific build paths. We
+# don't care if openssl-sys for example is used at build time (which it is, indirectly
+# by shadow-rs), only at run time. See more here:
+# https://github.com/EmbarkStudios/cargo-deny/issues/563
+#
+# It assumes cargo and friends are available.
+#
+# Run this from the root of the repo.
```
0xjinn, line 19: root of the repo as `/aptos-core`?

0xjinn (approved):
noice!! I wonder, shall we we combine this with rust linter script? So locally we have one less script to run and check :D

gedigi:
have you considered using something like `cargo vet` or `cargo deny`?

gregnazario (approved):
Since cargo deny doesn't work, this should be good

banool:
> have you considered using something like cargo vet or cargo deny?

Yeah check out the summary, it doesn't work for this case unfortunately.

PR #10340, opened 2023-10-02
[e2e-move-tests] remove aptos crate (cli tool) dependency

igor-aptos:
e2e-move-tests were depending on aptos crate, just to use two utility structs. With that it basically required everything to be built. 

moving these to aptos-types (let me know if you have a better preference), reduces number of dependencies from ~1400 to ~1000, and significantly reducing build time. 
Also when modifying some code, and re-running, much less things need to be build, and the very-time-consuming "aptos" crate is not built any more

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6183,7 +6183,6 @@ name = "e2e-move-tests"
 version = "0.1.0"
 dependencies = [
  "anyhow",
- "aptos",
  "aptos-block-executor",
  "aptos-cached-packages",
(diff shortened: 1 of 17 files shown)


msmouse (approved):

igor-aptos:
ping, need a second review :)

vgao1996 (approved):

wrwg (approved):
Thanks for doing this!

banool (approved):
Looks good from the CLI perspective

PR #10341, opened 2023-10-02
Move encoding type from aptos to aptos-crypto crate

igor-aptos:
A few other modules depend on aptos crate for EncodingType, aptos-transaction-emitter-lib solely for that class. And aptos create is very heavy and includes everything - building the full node, and any changes require long rebuild.

Moving it out to where it seems most relevant - to the aptos-crypto crate. Not sure if that is OK, or another place is better?

This reduces number of dependencies for aptos-transaction-emitter-lib crate from ~1000 to ~700.


alinush (approved):
LGTM!

igor-aptos:
(updated to remove accidentally leftover commented out lines)

davidiw (approved):

PR #10348, opened 2023-10-02
[TS SDK V2] Transaction Submission Flow

0xmaayan:
### Description
Support for transaction submission in the TS SDK V2.
There is some follow up work, but this is sufficient enough to try and get this merged to unblock others

**TransactionSubmission**
---
The `TransactionSubmission` class is a user facing API class that holds the different operation a user can interact with. It exposes:
- `generateTransaction()` -  a function that generates any transaction by providing the move function, type arguments and arguments and transaction arguments. It creates a transaction payload and generated a raw transaction based on that.
- `signTransaction()` - function to sign a transaction, that function returns an account authenticator for the account who signed the transaction
- `simulateTransaction()` - a function that simulates a given raw transaction
- `submitTransaction()` - a function that submits a transaction to chain. It generated a signed transaction, serializes it and calls the fullnode submission endpoint.

```
const aptos = new Aptos();
const transaction = aptos.generateTransaction()
const authenticator = aptos.signTransaction()
const response = await aptos.submitTransaction()
```

**TransactionBuilder**
---
The `TransactionBuilder` file exposes transaction lifecycle functions that used by the `TransactionSubmission` class.
- `generateTransactionPayload()` - a function that generates a transaction payload based on the provided arguments.
- `generateRawTransaction()` - a function that generates a raw transaction
- `generateTransaction()` - function that generates a transaction by providing a `TransactionPayload` and additional parameters such as transaction options (max gas, expireTimestamp, etc) , fee payer address and additional signers address. This function returns an object represents the transaction data.
NOTE: the raw transaction is serialized. That helps with transfering the transaction from a server A to a server B or from different machines.
```
{
  rawTransaction: Uint8Array;
  secondarySignerAddresses?: Array<AccountAddress>;
  feePayerAddress?: AccountAddress;
};
```
- `generateSignedTransactionForSimulation()` - this function generates a "signed" transaction that can be used for simulation
- `signTransaction()` - a function to sign a transaction that later can be used to submit to chain
- `generateSignedTransaction()` - a function that generates a signed transaction that can later be submitted to chain.

**Following work**
- Tests. the current flow supports:
  -  entry function, script and multi sig payloads. 
  - single signer, fee payer and multi agent transactions
  - ed25519 and secp256k1 schemes

Need to come up with a robust tests for each option. @xbtmatt 

- We now use `ScriptTransactionArgument` type for a script payload. This should be changed once https://github.com/aptos-labs/aptos-core/pull/10450 lands
- The current implementation uses function overload to support a better type interface. We should probably improve it a bit more to support more transaction combinations.
- https://github.com/aptos-labs/aptos-core/pull/10443 just land, need to understand how it affects our implementation


0xmaayan:
After talking with @hardsetting offline, we might be able to achieve a better API interface while using AccountAuthenticator as the return type. Gonna try and implement it

ecosystem/typescript/sdk_v2/src/types/index.ts
```diff
@@ -817,3 +820,48 @@ export enum AuthenticationKeyScheme {
...
+  secondary_signer_addresses?: Array<HexInput>;
+  fee_payer_address?: HexInput;
+} & (EntryFunctionData | ScriptData | EntryFunctionBCSData);
+
+export type TransactionPayload = {
+  type: "entry_function_payload" | "script_payload";
+  payload: TransactionPayloadEntryFunction | TransactionPayloadScript;
+};
+
+export type AptosRawTransaction = {
```
xbtmatt, line 838: I think we should either keep this `RawTransactionWithData` or at least clarify what exactly it is- because it isn't a `RawTransaction`.

If I understand it correctly, the extra data in `RawTransactionWithData`, i.e. the `secondary_signer_addresses` and `fee_payer_address` are only used in the transaction authentication process (signing), i.e. constructing the `TransactionAuthenticator`

They aren't included in the `RawTransaction` itself, so this name is a little misleading.

Not sure how much it matters, but just saying!

Also, fwiw, the struct in Rust is also named `RawTransactionWithData`.

The clearest name would probably be something like `RawTransactionWithAuthenticatorData` or something 😅

ecosystem/typescript/sdk_v2/src/types/index.ts
```diff
@@ -817,3 +820,48 @@ export enum AuthenticationKeyScheme {
...
+export type EntryFunctionData = {
+  function: MoveStructType;
+  type_arguments: Array<MoveType>;
+  arguments: Array<MoveValue>;
+};
+
+export type ScriptData = {
+  code: string;
+  type_arguments: Array<any>; // should be TypeTag
+  arguments: Array<Uint8Array>;
```
xbtmatt, line 853: Side note, only tangentially relevant:

I just realized that `script` arguments are serialized with a type enum before them because for whatever reason (I think it's the bug Jijun was talking about) we don't have a script transaction arg parser, so it doesn't know ahead of time the layout of the transaction arguments.

We can handle it all internally in the typetag stuff, but this `arguments` field will be an `Array<TransactionArgument>` or at least something we serialize differently, whereas in `EntryFunction` the `arguments` fields don't need to have their types prepended in the serialized args, because it's a "fixed" layout of arguments, so to speak.

ecosystem/typescript/sdk_v2/src/transactions/instances/transactionPayload.ts
```diff
@@ -134,21 +136,52 @@ export class EntryFunction {
...
     this.module_name.serialize(serializer);
     this.function_name.serialize(serializer);
     serializer.serializeVector<TypeTag>(this.type_args);
 
     serializer.serializeU32AsUleb128(this.args.length);
-    this.args.forEach((item: Uint8Array) => {
-      serializer.serializeBytes(item);
+    this.args.forEach((item: Serializable) => {
+      const bytes = item.bcsToBytes();
+      serializer.serializeBytes(bytes);
```
xbtmatt, line 184: So BTW, I realized this really sucks in terms of consistent representations, because it means whenever we deserialize a transaction, we have to treat the bytes as the FixedBytes thing.

I'll add more info about this later, I have a solution (I think) but it changes a lot of the interfaces for stuff, although generally I think this is sorely needed, because `Serializable` isn't really an accurate representation of what we're sending in here

ecosystem/typescript/sdk_v2/src/transactions/instances/transactionPayload.ts
```diff
@@ -134,21 +136,52 @@ export class EntryFunction {
...
+   * ```
+   * public(script) fun transfer<CoinType>(from: &signer, to: address, amount: u64,)
+   * ```
+   * @returns EntryFunction
+   */
+  static build(
+    module_name: `${string}::${string}`,
+    function_name: string,
+    type_args: Array<TypeTag>,
+    args: Array<Serializable>,
```
xbtmatt, line 171: In reference to the other comment below, this will eventually be:

```typescript
args: Array<EntryFunctionArgument>,

// where EntryFunctionArgument is:
export interface EntryFunctionArgument {
  /**
   * Serialize an argument to BCS-serialized bytes.
   */
  serialize(serializer: Serializer): void;

  /**
   * Serialize an argument as a type-agnostic, fixed byte sequence. The byte sequence contains
   * the number of the following bytes followed by the BCS-serialized bytes for a typed argument.
   */
  serializeForEntryFunction(serializer: Serializer): void;
}
```

we can talk about it tomorrow or something, it should make things a little more clear in terms of how we represent data for different types of payloads, and make it difficult/impossible to inadvertently use them incorrectly

ecosystem/typescript/sdk_v2/src/transactions/instances/transactionPayload.ts
```diff
@@ -158,12 +191,15 @@ export class EntryFunction {
     const type_args = deserializer.deserializeVector(TypeTag);
 
     const length = deserializer.deserializeUleb128AsU32();
-    const list: Array<Uint8Array> = [];
+    const list: Array<Serializable> = new Array<MoveVector<U8>>();
+
     for (let i = 0; i < length; i += 1) {
-      list.push(deserializer.deserializeBytes());
+      const vec = MoveVector.deserialize(deserializer, U8);
+      list.push(new MoveVector(vec.values.splice(1)));
```
xbtmatt, line 198: The new change is in my other PR, but keep in mind this is actually wrong. That's my bad, heh, but basically there's no way to correctly represent this in terms of being able to `serialize => deserialize => serialize` unless we use some form of `FixedBytes`.

0xmaayan: updated

ecosystem/typescript/sdk_v2/src/api/transaction_submission.ts
```diff
@@ -0,0 +1,171 @@
...
+  constructor(config: AptosConfig) {
+    this.config = config;
+  }
+
+  /**
+   * We are defining function signatures, each with its specific input and output.
+   * These are the possible function signature for `generateTransaction` function.
+   * When we call `generateTransaction` function with the relevant type properties,
+   * Typescript can infer the return type based on the appropriate function overload.
+   */
```
gregnazario, line 37: Wonder if this shows up in the ts-docs, when we eventually use those.

ecosystem/typescript/sdk_v2/src/transactions/types.ts
```diff
@@ -0,0 +1,230 @@
...
+  TransactionPayloadEntryFunction,
+  TransactionPayloadMultisig,
+  TransactionPayloadScript,
+} from "./instances";
+import { TypeTag } from "./typeTag/typeTag";
+
+/**
+ * Type that holds all raw transaction instances Aptos SDK supports
+ */
+export type AnyRawTransactionInstance = RawTransaction | MultiAgentRawTransaction | FeePayerRawTransaction;
```
gregnazario, line 20: What is a TransactionInstance?

ecosystem/typescript/sdk_v2/src/transactions/types.ts
```diff
@@ -0,0 +1,230 @@
...
+/**
+ * Interface of the arguments to generate a multi agent transaction.
+ * Used to provide to `generateTransaction()` method in the transaction builder flow
+ */
+export interface GenerateMultiAgentRawTransactionArgs {
+  aptosConfig: AptosConfig;
+  sender: HexInput;
+  payload: TransactionPayload;
+  secondarySignerAddresses: HexInput[];
+  feePayerAddress?: undefined;
```
gregnazario, line 108: optional undefined?

ecosystem/typescript/sdk_v2/src/transactions/types.ts
```diff
@@ -0,0 +1,230 @@
...
+
+/**
+ * Interface that holds the return data when generating a single signer transaction
+ *
+ * @param rawTransaction a bcs serialized raw transaction
+ */
+export interface SingleSignerTransaction {
+  rawTransaction: Uint8Array;
+  feePayerAddress?: undefined;
+  secondarySignerAddresses?: undefined;
```
gregnazario, line 128: I'm wondering, why optional undefined?

ecosystem/typescript/sdk_v2/src/transactions/types.ts
```diff
@@ -0,0 +1,230 @@
...
+
+/**
+ * Interface that holds the user data input when generating a single signer transaction
+ */
+export interface GenerateSingleSignerRawTransactionInput {
+  sender: HexInput;
+  feePayerAddress?: undefined;
+  secondarySignerAddresses?: undefined;
+  options?: GenerateTransactionOptions;
+  data: EntryFunctionData | ScriptData | MultiSigData;
```
gregnazario, line 199: GenerateTransactionPayloadData

ecosystem/typescript/sdk_v2/src/transactions/types.ts
```diff
@@ -0,0 +1,230 @@
...
+
+/**
+ * Interface that holds the user data input when generating a fee payer transaction
+ */
+export interface GenerateFeePayerRawTransactionInput {
+  sender: HexInput;
+  feePayerAddress: HexInput;
+  secondarySignerAddresses?: HexInput[];
+  options?: GenerateTransactionOptions;
+  data: EntryFunctionData | ScriptData | MultiSigData;
```
gregnazario, line 210: GenerateTransactionPayloadData

ecosystem/typescript/sdk_v2/src/api/transaction_submission.ts
```diff
@@ -0,0 +1,171 @@
...
+   * `
+   * data: {
+   *  function:"0x1::aptos_account::transfer",
+   *  type_arguments:[]
+   *  arguments:[recieverAddress,10]
+   * }
+   * `
+   *
+   * @example
+   * For a singale signer script function
```
gregnazario, line 64:
```suggestion
   * For a single signer script function
```

ecosystem/typescript/sdk_v2/src/api/transaction_submission.ts
```diff
@@ -0,0 +1,171 @@
...
+   * Generates any transaction by passing in the required arguments
+   *
+   * @param args.sender The transaction sender's account address as a HexInput
+   * @param args.data EntryFunctionData | ScriptData | MultiSigData
+   * @param feePayerAddress optional. For a fee payer (aka sponsored) transaction
+   * @param secondarySignerAddresses optional. For a multi agent or fee payer (aka sponsored) transactions
+   * @param args.options optional. GenerateTransactionOptions type
+   *
+   * @example
+   * For a singale signer entry function
```
gregnazario, line 53:
```suggestion
   * For a single signer entry function
```

ecosystem/typescript/sdk_v2/src/transactions/transaction_builder/transaction_builder.ts
```diff
@@ -0,0 +1,512 @@
...
+  // generate script payload
+  if ("bytecode" in args) {
+    const scriptPayload = new TransactionPayloadScript(
+      new Script(hexToBytes(args.bytecode), args.type_arguments, args.arguments),
+    );
+    return scriptPayload;
+  }
+
+  // generate multi sig payload
+  if ("multisigAddress" in args) {
```
gregnazario, line 92: Guards like this are problematic if say these things are applied with multiple in the future, we may want to approach explicitly saying the expected type in the transaction payload data.

ecosystem/typescript/sdk_v2/src/transactions/transaction_builder/transaction_builder.ts
```diff
@@ -0,0 +1,512 @@
...
+  const { aptosConfig, sender, payload, options } = args;
+  const [{ sequence_number: sequenceNumber }, chainId, { gas_estimate: gasEstimate }] = await Promise.all([
+    options?.accountSequenceNumber
+      ? Promise.resolve({ sequence_number: options.accountSequenceNumber })
+      : await getInfo({ aptosConfig, accountAddress: sender }),
+    NetworkToChainId[aptosConfig.network] ?? (await getLedgerInfo({ aptosConfig })).chain_id,
+    options?.gasUnitPrice
+      ? Promise.resolve({ gas_estimate: options.gasUnitPrice })
+      : await getGasPriceEstimation({ aptosConfig }),
+  ]);
```
gregnazario, line 147: Is there a general way to break this up so it's not a 10 line ternary wrapped line.  It's fairly hard to read.

ecosystem/typescript/sdk_v2/src/transactions/transaction_builder/transaction_builder.ts
```diff
@@ -0,0 +1,512 @@
...
+/**
+ * Simluate a transaction before signing and submit to chain
+ * @param args
+ * @returns
+ */
+export function generateSignedTransactionForSimulation(args: SimulateTransactionData): Uint8Array {
+  const { signerPublicKey, transaction, secondarySignersPublicKeys, feePayerPublicKey } = args;
+
+  const deserializer = new Deserializer(transaction.rawTransaction);
+  const desirializedTransaction = RawTransaction.deserialize(deserializer);
```
gregnazario, line 241:
```suggestion
  const deserializedTransaction = RawTransaction.deserialize(deserializer);
```

ecosystem/typescript/sdk_v2/src/transactions/transaction_builder/transaction_builder.ts
```diff
@@ -0,0 +1,512 @@
...
+  transaction: AnyRawTransaction;
+  senderAuthenticator: AccountAuthenticator;
+  secondarySignerAuthenticators?: {
+    feePayerAuthenticator?: AccountAuthenticator;
+    additionalSignersAuthenticators?: Array<AccountAuthenticator>;
+  };
+}): Uint8Array {
+  const { transaction, senderAuthenticator, secondarySignerAuthenticators } = args;
+
+  const transactionToSubmit = derieveTransactionType(transaction);
```
gregnazario, line 373:
```suggestion
  const transactionToSubmit = deriveTransactionType(transaction);
```

ecosystem/typescript/sdk_v2/src/transactions/transaction_builder/transaction_builder.ts
```diff
@@ -0,0 +1,512 @@
...
+    const transactionAuthenticator = new TransactionAuthenticatorSecp256k1(
+      accountAuthenticator.public_key,
+      accountAuthenticator.signature,
+    );
+    // return signed transaction
+    return new SignedTransaction(transactionToSubmit as RawTransaction, transactionAuthenticator).bcsToBytes();
+  }
+
+  throw new Error(
+    `can not generate signed transaction, ${accountAuthenticator} is not a supported account authentication`,
```
gregnazario, line 408:
```suggestion
    `Cannot generate a signed transaction, ${accountAuthenticator} is not a supported account authentication scheme`,
```

ecosystem/typescript/sdk_v2/src/transactions/transaction_builder/transaction_builder.ts
```diff
@@ -0,0 +1,512 @@
...
+/**
+ * Derive the raw transaction type - FeePayerRawTransaction or MultiAgentRawTransaction or RawTransaction
+ *
+ * @param transaction A aptos transaction type
+ *
+ * @returns FeePayerRawTransaction | MultiAgentRawTransaction | RawTransaction
+ */
+export function derieveTransactionType(transaction: AnyRawTransaction): AnyRawTransactionInstance {
+  const deserializer = new Deserializer(transaction.rawTransaction);
+  const desirializedTransaction = RawTransaction.deserialize(deserializer);
```
gregnazario, line 421:
```suggestion
  const deserializedTransaction = RawTransaction.deserialize(deserializer);
```

ecosystem/typescript/sdk_v2/src/transactions/transaction_builder/transaction_builder.ts
```diff
@@ -0,0 +1,512 @@
...
+}
+
+/**
+ * Derive the raw transaction type - FeePayerRawTransaction or MultiAgentRawTransaction or RawTransaction
+ *
+ * @param transaction A aptos transaction type
+ *
+ * @returns FeePayerRawTransaction | MultiAgentRawTransaction | RawTransaction
+ */
+export function derieveTransactionType(transaction: AnyRawTransaction): AnyRawTransactionInstance {
```
gregnazario, line 419:
```suggestion
export function deriveTransactionType(transaction: AnyRawTransaction): AnyRawTransactionInstance {
```

ecosystem/typescript/sdk_v2/src/transactions/transaction_builder/transaction_builder.ts
```diff
@@ -0,0 +1,512 @@
...
+
+export function getSigningMessage(rawTxn: AnyRawTransactionInstance): Uint8Array {
+  const hash = sha3Hash.create();
+
+  if (rawTxn instanceof RawTransaction) {
+    hash.update(RAW_TRANSACTION_SALT);
+  } else if (rawTxn instanceof MultiAgentRawTransaction) {
+    hash.update(RAW_TRANSACTION_WITH_DATA_SALT);
+  } else if (rawTxn instanceof FeePayerRawTransaction) {
+    hash.update(RAW_TRANSACTION_WITH_DATA_SALT);
```
gregnazario, line 498: Can probably combine these if they're the same

ecosystem/typescript/sdk_v2/src/transactions/transaction_builder/transaction_builder.ts
```diff
@@ -0,0 +1,512 @@
...
+    const multiAgentAuthenticator = new TransactionAuthenticatorMultiAgent(
+      senderAuthenticator,
+      transaction.secondary_signer_addresses,
+      additionalSignersAuthenticators ?? [],
+    );
+    return new SignedTransaction(transaction.raw_txn, multiAgentAuthenticator).bcsToBytes();
+  }
+
+  throw new Error(
+    `cann not prepare multi signers transaction to submission, ${typeof transaction} transaction is not supported`,
```
gregnazario, line 483:
```suggestion
    `Cannot prepare multi signers transaction to submission, ${typeof transaction} transaction is not supported`,
```

gregnazario (commented):
I think there will likely need to be some churn on names, as well as spelling typos.

We will need to add tests for submission of transactions for:
1. Fee payer
2. Multi agent
3. Multi-ed25519 
Etc, as the compatibility onchain is very important, as well as I'd like to see what the development experience becomes for multi-agent and fee payer

ecosystem/typescript/sdk_v2/src/transactions/instances/chainId.ts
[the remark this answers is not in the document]
0xmaayan: maybe a bad name. before it was under `types/` - wanted to rename the folder.

ecosystem/typescript/sdk_v2/src/utils/api-endpoints.ts
```diff
@@ -25,3 +25,8 @@ export enum Network {
   LOCAL = "local",
   CUSTOM = "custom",
 }
+
+export const NetworkToChainId: Record<string, number> = {
+  mainnet: 1,
+  testnet: 2,
+};
```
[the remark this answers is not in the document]
0xmaayan: what if they run it against their own local node?

ecosystem/typescript/sdk_v2/src/utils/const.ts
```diff
@@ -7,3 +7,7 @@ export enum AptosApiType {
   INDEXER,
   FAUCET,
 }
+
+export const DEFAULT_MAX_GAS_AMOUNT = 200000;
+// Transaction expire timestamp
+export const DEFAULT_TXN_EXP_SEC_FROM_NOW = 20;
```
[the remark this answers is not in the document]
0xmaayan: I think he added `DEFAULT_TXN_TIMEOUT_SEC`

ecosystem/typescript/sdk_v2/src/api/transaction_submission.ts
0xmaayan: would we?

0xjinn (approved):
Did a quick glance and looks ok. Approving to unblock dependent works

PR #10349, opened 2023-10-02
[spec] specify functions in framework modules

aalok-t:
### Description

1. Add concrete `aborts_if` clauses in the `mint` function in `coin`. Update the specification for all the functions that call `coin::mint`.
2. Add functional specification for the `mint` function in `managed_coin` to ensure that minted coins are deposited to the provided destination address.
3. Add functional specification for the `register` function in `managed_coin` to ensure that the registered account exists after the function call.
4. Add requirement for the `address_from_bytes` function in `util` to ensure that address input bytes are exactly 32 bytes long. This is an abstract specification and the soundness of this abstraction depends on the native function. If length of address input bytes is not 32, the deserialization will fail. See [utils.rs](https://github.com/aptos-labs/aptos-core/blob/main/crates/indexer/src/util.rs).
5. Add functional specification for the `update_global_time` function in `timestamp` to ensure that the wall clock time is updated by consensus.

### DIFF
--- a/aptos-move/framework/aptos-framework/doc/transaction_fee.md
+++ b/aptos-move/framework/aptos-framework/doc/transaction_fee.md
@@ -805,7 +805,7 @@ Only called during genesis.
 <b>requires</b> <b>exists</b>&lt;<a href="transaction_fee.md#0x1_transaction_fee_AptosCoinCapabilities">AptosCoinCapabilities</a>&gt;(@aptos_framework);
 <b>requires</b> <b>exists</b>&lt;CoinInfo&lt;AptosCoin&gt;&gt;(@aptos_framework);
 <b>let</b> amount_to_burn = (burn_percentage * <a href="coin.md#0x1_coin_value">coin::value</a>(<a href="coin.md#0x1_coin">coin</a>)) / 100;
-<b>include</b> amount_to_burn &gt; 0 ==&gt; <a href="coin.md#0x1_coin_AbortsIfAggregator">coin::AbortsIfAggregator</a>&lt;AptosCoin&gt;{ <a href="coin.md#0x1_coin">coin</a>: Coin&lt;AptosCoin&gt;{ value: amount_to_burn } };
+<b>include</b> amount_to_burn &gt; 0 ==&gt; <a href="coin.md#0x1_coin_CoinSubAbortsIf">coin::CoinSubAbortsIf</a>&lt;AptosCoin&gt;{ amount: amount_to_burn };
 <b>ensures</b> <a href="coin.md#0x1_coin">coin</a>.value == <b>old</b>(<a href="coin.md#0x1_coin">coin</a>).value - amount_to_burn;
 </code></pre>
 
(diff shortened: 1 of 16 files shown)


aptos-move/framework/aptos-framework/sources/managed_coin.spec.move
```diff
@@ -68,10 +67,20 @@ spec aptos_framework::managed_coin {
...
+
+        // Capabilities<CoinType> should not exist in the signer address.
         aborts_if !exists<Capabilities<CoinType>>(account_addr);
+
         let coin_store = global<coin::CoinStore<CoinType>>(dst_addr);
+        // Capabilities<CoinType> should not exist in the destination address and `dst_addr` should not be frozen.
         aborts_if !exists<coin::CoinStore<CoinType>>(dst_addr);
         aborts_if coin_store.frozen;
+
+        // Ensure that 'amount' is withdrawn from 'account'.
```
rahxephon89, line 79: Could you double check whether this property should hold for the mint function? @aalok-t

junkil-park: By the way, did this pass the Prover test locally?

junkil-park: Could you try `ensures false;` and see if Prover proves it? If so, it is an inconsistency problem.

rahxephon89: If it is not an inconsistency issue, it may be related to the spec of  `mint` in the `coin` module, which is opaque. @junkil-park

aptos-move/framework/aptos-framework/sources/util.spec.move
```diff
@@ -6,4 +6,11 @@ spec aptos_framework::util {
     }
 
     spec fun spec_from_bytes<T>(bytes: vector<u8>): T;
+
+    spec address_from_bytes(bytes: vector<u8>): address {
+
```
junkil-park, line 11: nit: remove this line?

aptos-move/framework/aptos-framework/sources/coin.spec.move
```diff
@@ -106,6 +106,21 @@ spec aptos_framework::coin {
             coin.value;
     }
 
+    spec schema AbortsIfAggregatorAdd<CoinType> {
```
junkil-park, line 109: This looks good. Can we refactor this a little bit? To follow the convention:
(1) define the schema `AddAbortsIf` in the module `optional_aggregator`.
(2) include `AddAbortsIf` in the spec of `optional_aggregator::add` instead of using the existing `aborts_if` spec
(3) use (include) `AddAbortsIf` in the spec of `coin::mint` as well

aptos-move/framework/aptos-framework/sources/stake.spec.move
```diff
@@ -190,6 +190,7 @@ spec aptos_framework::stake {
     }
 
     spec on_new_epoch {
+        pragma verify = false;
```
aalok-t, line 193: I am unable to identify how the on_new_epoch function calls coin::mint in order to ensure the correct aborts_if condition.

clairetyping: Is it possible to add requires instead of aborts_if? If it does not work, try to change the spec of mint instead of modifying the spec of the caller

aptos-move/framework/aptos-framework/sources/stake.spec.move
```diff
@@ -215,9 +215,15 @@ spec aptos_framework::stake {
 
     spec update_stake_pool {
         // TODO: set because of timeout in CI. The property is proved (inconsistency check also passes)
+        pragma opaque;
```
rahxephon89, line 218: Could you check whether the spec works when removing `pragma verify = false`?

aalok-t: When we remove `pragma verify = false`, the verification fails as update_stake_pool is missing some `aborts_if` condition. Should I add those too?

rahxephon89: Yes

aptos-move/framework/aptos-framework/doc/transaction_fee.md
```diff
@@ -805,7 +805,7 @@ Only called during genesis.
 <b>requires</b> <b>exists</b>&lt;<a href="transaction_fee.md#0x1_transaction_fee_AptosCoinCapabilities">AptosCoinCapabilities</a>&gt;(@aptos_framework);
```
junkil-park, line 805: The linter fails in CI due to the outdated md files: https://github.com/aptos-labs/aptos-core/actions/runs/6580606157/job/17878952753?pr=10349
Please do `cargo build -p aptos-cached-packages` to regenerate the md files before (force-)pushing the commit.

aptos-move/framework/aptos-framework/sources/stake.spec.move
aalok-t: Resolved by adding the appropriate aborts_if clauses.

rahxephon89 (approved):
Look good. Could you solve the comment from @junkil-park before merging it?

junkil-park (approved):

PR #10350, opened 2023-10-02
Use gcs for replay-verify and Make Replay-verify Resumable

areshand:
### Description

1. use gcs instead of s3 for replay and verify
2. add retry at partition for failed command to improve robustness 

### Test Plan
https://github.com/aptos-labs/aptos-core/actions/runs/6471317308

### DIFF
--- a/storage/backup/backup-cli/src/coordinators/replay_verify.rs
+++ b/storage/backup/backup-cli/src/coordinators/replay_verify.rs
@@ -12,14 +12,28 @@ use crate::{
     storage::BackupStorage,
     utils::{GlobalRestoreOptions, RestoreRunMode, TrustedWaypointOpt},
 };
-use anyhow::{bail, ensure, Result};
+use anyhow::Result;
 use aptos_db::backup::restore_handler::RestoreHandler;
 use aptos_executor_types::VerifyExecutionMode;
 use aptos_logger::prelude::*;
 use aptos_types::{on_chain_config::TimedFeatureOverride, transaction::Version};
 use aptos_vm::AptosVM;
 use std::sync::Arc;
+use thiserror::Error;
 
+#[derive(Debug, Error)]
+pub enum ReplayError {
+    #[error("Txn mismatch error")]
+    TxnMismatch,
+    #[error("Other Replay error {0}")]
+    OtherError(String),
+}
+
+impl From<anyhow::Error> for ReplayError {
+    fn from(error: anyhow::Error) -> Self {
+        ReplayError::OtherError(error.to_string())
+    }
+}
 pub struct ReplayVerifyCoordinator {
     storage: Arc<dyn BackupStorage>,
     metadata_cache_opt: MetadataCacheOpt,
@@ -60,9 +74,8 @@ impl ReplayVerifyCoordinator {
         })
     }
 
-    pub async fn run(self) -> Result<()> {
+    pub async fn run(self) -> Result<(), ReplayError> {
         info!("ReplayVerify coordinator started.");
-
         let ret = self.run_impl().await;
 
         if let Err(e) = &ret {
@@ -77,7 +90,7 @@ impl ReplayVerifyCoordinator {
         ret
     }
 
-    async fn run_impl(self) -> Result<()> {
+    async fn run_impl(self) -> Result<(), ReplayError> {
         AptosVM::set_concurrency_level_once(self.replay_concurrency_level);
         AptosVM::set_timed_feature_override(TimedFeatureOverride::Replay);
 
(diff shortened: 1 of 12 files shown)


testsuite/replay_verify.py
```diff
@@ -64,12 +65,26 @@
...
+    [285_000_001, sys.maxsize],
 ]
 
 
+def replay_with_retry(*args):
+    MAX_RETRIES = 5
+    attempt = 0
+    while attempt < MAX_RETRIES:
+        (n, return_code, msg) = replay_verify_partition(*args)
+        if return_code == 0:
```
msmouse, line 79: should distinguish infra error and txn_info mismatch error (don't retry on the latter)

msmouse: (test with one ver removed from the skip list)

areshand: refactor the code to have more error types and only retry on non-txn mismatch error type.

storage/db-tool/src/replay_verify.rs
```diff
@@ -70,19 +74,40 @@ impl Opt {
...
-            self.end_version.unwrap_or(Version::MAX),
-            self.validate_modules,
-            VerifyExecutionMode::verify_except(self.txns_to_skip).set_lazy_quit(self.lazy_quit),
-        )?
-        .run()
-        .await
+        let mut attempt = 0;
+        let ret = Ok(());
+        while attempt < RETRY_ATTEMPT {
+            let ret = ReplayVerifyCoordinator::new(
```
msmouse, line 80: Marking the outer `ret`?  (the method always returns `Ok(())`?)

perryjrandall (approved):
Does it make sense to run this in parallel instead of just cutting over?

Either way lgtm

areshand:
> Does it make sense to run this in parallel instead of just cutting over?
> 
> Either way lgtm

It should run in parallel. Let's wait @msmouse  to give it a look

perryjrandall:
> > Does it make sense to run this in parallel instead of just cutting over?
> > Either way lgtm
> 
> It should run in parallel. Let's wait @msmouse to give it a look

I mean run aws / gcp based job in parallel

https://github.com/aptos-labs/aptos-core/pull/10350/files#diff-3efa580731ca445ce0e8063eabb4668a669b537d270a4d864a2154129c1ea43eL59

You're cutting over from aws -> gcp again, last time we did this it failed

It would be good to be able to compare the two side by side and compare failure rate, but if not necessary we can always decide if revert to aws later

areshand:
> > > Does it make sense to run this in parallel instead of just cutting over?
> > > Either way lgtm
> > 
> > 
> > It should run in parallel. Let's wait @msmouse to give it a look
> 
> I mean run aws / gcp based job in parallel
> 
> https://github.com/aptos-labs/aptos-core/pull/10350/files#diff-3efa580731ca445ce0e8063eabb4668a669b537d270a4d864a2154129c1ea43eL59
> 
> You're cutting over from aws -> gcp again, last time we did this it failed
> 
> It would be good to be able to compare the two side by side and compare failure rate, but if not necessary we can always decide if revert to aws later

Yes, last time we see GCP has a higher failure rate, which should be the same case now. This PR is to increase the success rate to a level that is not affected by transient errors.

storage/backup/backup-cli/src/coordinators/replay_verify.rs
```diff
@@ -158,16 +175,16 @@ impl ReplayVerifyCoordinator {
...
+            None,                            /* epoch_history */
             self.verify_execution_mode.clone(),
             None,
         )
         .run()
         .await?;
 
         if self.verify_execution_mode.seen_error() {
-            bail!("Seen replay errors, check out logs.")
+            Err( ReplayError::TxnMismatch)
```
msmouse, line 187: did lint fail for the space?

testsuite/replay_verify.py
```diff
@@ -110,11 +111,11 @@ def replay_verify_partition(
             "replay-verify",
             *txns_to_skip_args,
             "--concurrent-downloads",
-            "2",
+            "8",
             "--replay-concurrency-level",
             "2",
             "--metadata-cache-dir",
-            f"./{partition_name}/metadata-cache",
+            "./metadata-cache",
```
msmouse, line 118: revert

testsuite/replay_verify.py
```diff
@@ -172,29 +181,29 @@ def main():
...
-
     assert runner_cnt == len(
         runner_mapping
     ), "runner_cnt must match the number of runners in the mapping"
     runner_start = runner_mapping[runner_no][0]
     runner_end = runner_mapping[runner_no][1]
     if runner_no == runner_cnt - 1:
+        LATEST_VERSION = runner_end
+        try:
+            LATEST_VERSION = query_backup_latest_version(BACKUP_CONFIG_TEMPLATE_PATH)
```
msmouse, line 198: let's just die when it fails. LATEST_VERSION=MAX doesn't work quite well.

msmouse (approved):

PR #10351, opened 2023-10-02
[Execution] Move Signature verification to new pipeline phase

sitalkedia:
### Description

- Add a block preparation phase and move the signature verification to the preparation phase. This is to move the signature verification out of the critical execution path - which consumes around 50 ms today for a block size of 25k.
- Remove cloning of transaction block and pass the reference instead to the block executor - this gives us around 2% of performance boost. 

### Test Plan

Tested with the benchmark and this change along with https://github.com/aptos-labs/aptos-core/pull/10388 gives us around 7% performance gain. The TPS 32k to 34k.

### DIFF
--- a/aptos-move/aptos-vm/src/aptos_vm.rs
+++ b/aptos-move/aptos-vm/src/aptos_vm.rs
@@ -27,6 +27,7 @@ use aptos_gas_meter::{AptosGasMeter, StandardGasAlgebra, StandardGasMeter};
 use aptos_gas_schedule::VMGasParameters;
 use aptos_logger::{enabled, prelude::*, Level};
 use aptos_memory_usage_tracker::MemoryTrackedGasMeter;
+use aptos_metrics_core::{exponential_buckets, register_histogram, Histogram};
 use aptos_state_view::StateView;
 use aptos_types::{
     account_config,
@@ -1509,6 +1510,17 @@ impl AptosVM {
     }
 }
 
+pub static APTOS_VM_EXEC_SECONDS: Lazy<Histogram> = Lazy::new(|| {
+    register_histogram!(
+        // metric name
+        "aptos_vm_exec_seconds",
+        // metric description
+        "The time spent in seconds in rayon thread pool in parallel execution",
+        exponential_buckets(/*start=*/ 1e-6, /*factor=*/ 2.0, /*count=*/ 30).unwrap(),
+    )
+    .unwrap()
+});
+
 // Executor external API
 impl VMExecutor for AptosVM {
     /// Execute a block of `transactions`. The output vector will have the exact same length as the
@@ -1521,6 +1533,7 @@ impl VMExecutor for AptosVM {
         state_view: &(impl StateView + Sync),
         maybe_block_gas_limit: Option<u64>,
     ) -> Result<Vec<TransactionOutput>, VMStatus> {
+        let _timer = APTOS_VM_EXEC_SECONDS.start_timer();
         fail_point!("move_adapter::execute_block", |_| {
             Err(VMStatus::error(
                 StatusCode::UNKNOWN_INVARIANT_VIOLATION_ERROR,
--- a/aptos-move/aptos-vm/src/aptos_vm.rs
+++ b/aptos-move/aptos-vm/src/aptos_vm.rs
@@ -39,7 +39,8 @@ use aptos_types::{
         MultisigTransactionPayload, SignatureCheckedTransaction, SignatureVerifiedTransaction,
         SignedTransaction, Transaction,
         Transaction::{
-            BlockMetadataTransaction, GenesisTransaction, StateCheckpoint, UserTransaction,
+            BlockMetadata as BlockMetadataTransaction, GenesisTransaction, StateCheckpoint,
+            UserTransaction,
         },
         TransactionOutput, TransactionPayload, TransactionStatus, VMValidatorResult,
         WriteSetPayload,
--- a/aptos-move/aptos-vm/src/aptos_vm.rs
+++ b/aptos-move/aptos-vm/src/aptos_vm.rs
@@ -3,9 +3,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use crate::{
-    adapter_common::{
-        discard_error_output, discard_error_vm_status, PreprocessedTransaction, VMAdapter,
-    },
+    adapter_common::{discard_error_output, discard_error_vm_status, VMAdapter},
     aptos_vm_impl::{get_transaction_output, AptosVMImpl, AptosVMInternals},
     block_executor::{AptosTransactionOutput, BlockAptosVM},
     counters::*,
@@ -38,7 +36,11 @@ use aptos_types::{
     on_chain_config::{new_epoch_event_key, FeatureFlag, TimedFeatureOverride},
     transaction::{
         EntryFunction, ExecutionError, ExecutionStatus, ModuleBundle, Multisig,
-        MultisigTransactionPayload, SignatureCheckedTransaction, SignedTransaction, Transaction,
+        MultisigTransactionPayload, SignatureCheckedTransaction, SignatureVerifiedTransaction,
+        SignedTransaction, Transaction,
+        Transaction::{
+            BlockMetadataTransaction, GenesisTransaction, StateCheckpoint, UserTransaction,
+        },
         TransactionOutput, TransactionPayload, TransactionStatus, VMValidatorResult,
         WriteSetPayload,
     },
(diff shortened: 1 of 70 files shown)


aptos-move/aptos-vm/src/aptos_vm.rs
```diff
@@ -541,6 +544,110 @@ impl AptosVM {
         RespawnedSession::spawn(self, session_id, resolver, change_set, storage_refund)
     }
 
+    fn execute_signature_verified_transaction(
+        &self,
+        txn: &Transaction,
```
msmouse, line 549: The loss of type safety in the interfaces (everywhere) is kinda annoying. Can we do better? Maybe introduce something like this? 
```
struct TxnWithValidSig<'a>(&Transaction)
```

```
impl SignatureCheckedTransaction {
    fn expect_valid(&self) -> TxnWithValidSig {
         ...
    }
}
```

consensus/src/execution_pipeline.rs
```diff
@@ -10,41 +10,65 @@ use aptos_executor_types::{
...
+            .thread_name(|index| format!("signature-checker-{}", index))
+            .build()
+            .unwrap(),
+    )
+});
 
 pub struct ExecutionPipeline {
-    block_tx: mpsc::UnboundedSender<ExecuteBlockCommand>,
+    prepare_block_tx: mpsc::UnboundedSender<PrepareBlockCommand>,
+    execute_block_tx: mpsc::UnboundedSender<ExecuteBlockCommand>,
```
msmouse, line 34: don't need to hold this on self?

aptos-move/block-executor/src/task.rs
```diff
@@ -28,30 +24,6 @@ pub enum ExecutionStatus<O, E> {
     SkipRest(O),
 }
 
-/// Trait that defines a transaction type that can be executed by the block executor. A transaction
-/// transaction will write to a key value storage as their side effect.
-pub trait Transaction: Sync + Send + Clone + 'static {
```
igor-aptos, line 33: I understand that you've renamed it because you needed the type externally, and need to distinguish it. 
But this is going to be a huge conflict with an aggregator branch.

how about having an alias for this module :
// TODO rename to BlockExecutableTransaction after aggregators get merged
type Transaction = BlockExecutableTransaction;

and undoing all the changes in block executor with regards to that?

types/src/aggregator.rs
```diff
@@ -0,0 +1,28 @@
...
+use move_core_types::account_address::AccountAddress;
+
+#[derive(Debug, Clone, Copy, Hash, Eq, PartialEq, Ord, PartialOrd)]
+pub struct AggregatorHandle(pub AccountAddress);
+
+/// Uniquely identifies each aggregator instance in storage.
+#[derive(Debug, Clone, Hash, Eq, PartialEq, Ord, PartialOrd)]
+pub struct AggregatorID(StateKey);
+
+impl AggregatorID {
```
igor-aptos, line 15: hmm, where do you need AggregatorID outside?

execution/executor/src/components/chunk_output.rs
```diff
@@ -245,6 +245,178 @@ impl ChunkOutput {
 }
 
 pub fn update_counters_for_processed_chunk(
+    transactions: &[SignatureVerifiedTransaction],
```
igor-aptos, line 248: cannot there be some trait that will allow these two functions to be merged?

execution/executor/src/components/chunk_output.rs
```diff
@@ -245,6 +245,178 @@ impl ChunkOutput {
 }
 
 pub fn update_counters_for_processed_chunk(
```
msmouse, line 247: Man, the duplication is unbearable :trollface: 
Can you do some magic here?

types/src/aggregator.rs
igor-aptos: asking as this ID is getting deprecated and refactored, and shouldn't be needed outside

sitalkedia: This is needed for the `BlockExecutorTransaction` interface.

execution/executor/src/components/chunk_output.rs
sitalkedia: Yes, I intended to do this as a follow up, when I move this counter off of critical path. But will add this in this PR.

aptos-move/block-executor/src/task.rs
sitalkedia: Done.

aptos-move/aptos-vm/src/aptos_vm.rs
sitalkedia: Too fancy for me.. I moved this function back to the original call to address your paranoia.. PTAL.

execution/executor/src/chunk_executor.rs
```diff
@@ -207,12 +220,22 @@ impl<V: VMExecutor> ChunkExecutorInner<V> {
             num_txns,
         )?;
 
+        // TODO(skedia) In the chunk executor path, we ideally don't need to verify the signature
+        // as only transactions with verified signatures are committed to the storage.
+        let sig_verified_txns = SIG_VERIFY_POOL.install(|| {
```
sitalkedia, line 225: @msmouse, @JoshLind -  Is there any case for keeping signature verification in the state sync path? If not, I will remove this paranoid check..

types/src/transaction/signature_verified_transaction.rs
```diff
@@ -0,0 +1,73 @@
...
+            SignatureVerifiedTransaction::Invalid(txn) => txn,
+        }
+    }
+
+    pub fn inner(&self) -> &Transaction {
+        match self {
+            SignatureVerifiedTransaction::Valid(txn) => txn,
+            SignatureVerifiedTransaction::Invalid(txn) => txn,
+        }
+    }
```
msmouse, line 34: make it this so that paranoid me feels safer :trollface: 
```
    pub fn expect_valid(&self) -> &Transaction {
        match self {
            SignatureVerifiedTransaction::Valid(txn) => txn,
            _ => panic!(),
        }
    }
```

sitalkedia: Done!

execution/executor/src/chunk_executor.rs
msmouse: I don't think it needs to be checked -- the txns are proven to be part of the chain represented by the ledger info in the proof. The execution mode, without the signature checking, still verifies the VM executes the txns the same with the validator set, and still saves bandwidth transfering the transaction outputs directly -- just one less thing checked.

types/src/transaction/signature_verified_transaction.rs
msmouse: Possible to remove `inner()` to make me feel safeR :trollface:? How do we justify getting the inner txn indifferently not caring if it's signature checked?

sitalkedia: This is mostly being used for some tests.. let me see if it can be doable..

execution/executor/src/chunk_executor.rs
JoshLind: To clarify, these are the signatures on the individual transactions themselves, right? e.g., verifying the transaction is correctly signed by the sender, etc. If so, this sounds okay to me. But, can we clarify where these checks are being done? i.e., I imagine we're still doing them in mempool, and I imagine that the VM needs to check them, too, right? Otherwise, what if we have a malicious block proposal (e.g., where some transaction signatures are invalid) where would this get caught?

sitalkedia: > But, can we clarify where these checks are being done? i.e., I imagine we're still doing them in mempool, and I imagine that the VM needs to check them, too, right? Otherwise, what if we have a malicious block proposal (e.g., where some transaction signatures are invalid) where would this get caught?

Yes, currently its being done twice. Once in mempool and once just before the proposal execution (although in a separate pipeline phase). 

> Likewise, what would happen in the case that someone has managed to temporarily compromise enough validators to execute incorrectly signed transactions. Would fullnodes be able to detect this if we remove these checks?

If more than f validators are compromised, then the BFT guarantees doesn't hold anymore.. not sure if FNs detecting invalid signature at that point is worth it..

JoshLind: Gotcha, thanks! 😄

> If more than f validators are compromised, then the BFT guarantees doesn't hold anymore.. not sure if FNs detecting invalid signature at that point is worth it..

FWIW, I think this point is debatable. One of the big arguments around the need for "verifying read-replicas" (i.e., fullnodes) is that it improves decentralization. Sure, BFT might be compromised in the validator set, but fullnode operators will always be able to detect and call out the divergence. If this no longer becomes the case, how could one detect divergence? I'm not saying don't do it -- it's just worth thinking about the trade-off 😄

sitalkedia: As discussed offline, let's keep the signature verification for now. We can follow up on this.

types/src/transaction/signature_verified_transaction.rs
sitalkedia: Removed the inner function.

sitalkedia: Please note that we still need the `into_inner` API as we store all txns in the chunk output. This is the same behavior we had previously as well.

msmouse (approved):

types/src/transaction/mod.rs
```diff
@@ -736,6 +741,11 @@ impl SignedTransaction {
         Ok(SignatureCheckedTransaction(self))
     }
 
+    pub fn verify_signature(&self) -> Result<()> {
```
zekun000, line 744: it's confusing to have this and `check_signature`, `SignatureCheckedTransaction` and `SignatureVerifiedTransaction` why do we need to keep the invalid case?

types/src/transaction/signature_verified_transaction.rs
```diff
@@ -0,0 +1,102 @@
...
+    type Key = StateKey;
+    type Tag = StructTag;
+    type Value = WriteOp;
+}
+
+pub fn into_signature_verified_block(txns: Vec<Transaction>) -> Vec<SignatureVerifiedTransaction> {
+    txns.into_iter().map(into_signature_verified).collect()
+}
+
+pub fn into_signature_verified(txn: Transaction) -> SignatureVerifiedTransaction {
```
zekun000, line 74: nit: it's more idiomatic to impl From

types/src/transaction/mod.rs
sitalkedia: @zekun000 - We need to keep the invalid case as we want to notify the mempool/state sync to remove these txns in the commit path.

zekun000 (approved):

types/src/transaction/mod.rs
zekun000: I think it's better to guarantee the block only has valid signature txns (like check in quorum store) but I guess that's for later

sitalkedia: Yes, that is true.. it would be much cleaner to pass only valid sig txn to the executor.. let me add a TODO here and we can follow up on that.

PR #10353, opened 2023-10-02
[release-builder] Fix increase lockup

perryjrandall:
The previous PR was converting the private key arg to string incorrectly...

ED25519 key type has a very silly to_string that returns debug output

I suppose this is to prevent dumping the key, but really i think it should just
not have a to_string and instead only have a debug method that does the same
thing, and a very explicit to hex method so that you dont accidentally dump these.

Test Plan: framework upgrade test succeeds, or at least doesnt fail on this
error (there is still some underlying flakiness in the forge test itself)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2996,6 +2996,7 @@ dependencies = [
  "aptos-framework",
  "aptos-gas-schedule-updator",
  "aptos-genesis",
+ "aptos-keygen",
  "aptos-rest-client",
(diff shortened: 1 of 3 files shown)


runtian-zhou (approved):

runtian-zhou:
Let's wait for framework upgrade test to pass to make sure it fix the issue

JoshLind (approved):
Unblocking for forge stability. 😄

PR #10364, opened 2023-10-03
[TS SDK V2] Port over transaction types

0xmaayan:
### Description
This PR ports over the transaction type classes we need to construct a transaction.
No big changes here as all of these classes are pretty much replications of the internal (rust) representations of them.

The only missing part here is the implementation of the `TypeTag` class which @xbtmatt is working on

### DIFF
--- a/ecosystem/typescript/sdk_v2/src/transactions/types/rawTransaction.ts
+++ b/ecosystem/typescript/sdk_v2/src/transactions/types/rawTransaction.ts
@@ -1,0 +1,198 @@
+import { Deserializer, Serializer } from "../../bcs";
+import { AccountAddress } from "../../core";
+import { ChainId } from "./chainId";
+import { TransactionArgument } from "./transactionArguments";
+import { TransactionPayload } from "./transactionPayload";
+
+/**
+ * Representation of a Raw Transaction that can serialized and deserialized
+ */
+export class RawTransaction {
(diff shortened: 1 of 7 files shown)


ecosystem/typescript/sdk_v2/src/transactions/types/rawTransaction.ts
```diff
@@ -0,0 +1,198 @@
...
+  ) {
+    super();
+    this.raw_txn = raw_txn;
+    this.secondary_signer_addresses = secondary_signer_addresses;
+    this.fee_payer_address = fee_payer_address;
+  }
+
+  serialize(serializer: Serializer): void {
+    // enum variant index
+    serializer.serializeU32AsUleb128(1);
```
gregnazario, line 185: Note for these numbers, maybe we make an enum that matches rust for easy identification, and then just use it here for the number

gregnazario (approved):
Excited for this stuff!  May be easier with an enum that matches the rust enum for auditabilty and adding new pieces

ecosystem/typescript/sdk_v2/src/transactions/types/rawTransaction.ts
0xmaayan: good idea!

heliuchuan (approved):

PR #10367, opened 2023-10-03
[TS SDK V2] Port over account and transaction authenticator, signed transaction type

0xmaayan:
### Description
This PR adds the account and transaction authenticator. In addition, adding the SignedTransaction class

As the previous PR, no big changes as all of these classes are pretty much replications of the internal (rust) representations of them.

### DIFF
--- a/ecosystem/typescript/sdk_v2/src/types/index.ts
+++ b/ecosystem/typescript/sdk_v2/src/types/index.ts
@@ -43,6 +43,26 @@ export enum RustEnumTransactionVariants {
 }
 
 /**
+ * Transaction Authenticator enum as they are represented in Rust
+ * {@link https://github.com/aptos-labs/aptos-core/blob/main/types/src/transaction/authenticator.rs#L44}
+ */
+export enum RustEnumTransactionAuthenticatorVariant {
+  TransactionAuthenticatorEd25519 = 0,
+  TransactionAuthenticatorMultiEd25519 = 1,
+  TransactionAuthenticatorMultiAgent = 2,
+  TransactionAuthenticatorFeePayer = 4,
+}
+
+/**
+ * Transaction Authenticator enum as they are represented in Rust
+ * {@link https://github.com/aptos-labs/aptos-core/blob/main/types/src/transaction/authenticator.rs#L414}
+ */
+export enum RustEnumAccountAuthenticatorVariant {
+  AccountAuthenticatorEd25519 = 0,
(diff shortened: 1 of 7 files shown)


ecosystem/typescript/sdk_v2/src/types/index.ts
```diff
@@ -42,6 +42,26 @@ export enum RustEnumTransactionVariants {
...
 
+/**
+ * Transaction Authenticator enum as they are represented in Rust
+ * {@link https://github.com/aptos-labs/aptos-core/blob/main/types/src/transaction/authenticator.rs#L44}
+ */
+export enum RustEnumTransactionAuthenticatorVariant {
+  TransactionAuthenticatorEd25519 = 0,
+  TransactionAuthenticatorMultiEd25519 = 1,
+  TransactionAuthenticatorMultiAgent = 2,
+  TransactionAuthenticatorFeePayer = 4,
```
gregnazario, line 53: Those should be pretty easy to name simply.  Doesn't have to be that descriptive that it refers to rust.

TransactionAuthenticatorType

Then just 
Ed25519
MultiEd25519
MultiAgent
FeePayer

ecosystem/typescript/sdk_v2/src/types/index.ts
```diff
@@ -42,6 +42,26 @@ export enum RustEnumTransactionVariants {
...
+}
+
+/**
+ * Transaction Authenticator enum as they are represented in Rust
+ * {@link https://github.com/aptos-labs/aptos-core/blob/main/types/src/transaction/authenticator.rs#L414}
+ */
+export enum RustEnumAccountAuthenticatorVariant {
```
gregnazario, line 60: Similarly

gregnazario (approved):

ecosystem/typescript/sdk_v2/src/types/index.ts
0xmaayan: I agree, though most of the time people look at it and ask "what are these numbers? where they come from?" it helped me a lot to figure out that these are fixed numbers that come from our rust code, but also the comment on each enum should help with that.

kk, let me change these names in a another PR

xbtmatt: yeah I feel that but I think we can just document that it comes from Rust, at some point we should be able to say "Read the documentation" rather than have to explain these things

xbtmatt (approved):
LGTM, can rename later

PR #10370, opened 2023-10-03
Update the link

Fr33sson:
There was an incorrect link in the second paragraph https://aptos.dev/nodes/validator-node/voter/index


banool:
Hey, it works for me without `/v1`, are you using the latest version of the CLI?

Fr33sson:
> Hey, it works for me without `/v1`, are you using the latest version of the CLI?

Oh. I haven't checked using the CLI. Yes, Yes, I have all the variants working through the CLI as well.: 
https://fullnode.mainnet.aptoslabs.com
https://mainnet.aptoslabs.com
https://fullnode.mainnet.aptoslabs.com/v1
https://mainnet.aptoslabs.com/v1

But in the browser it only works correctly with /v1 at the end. So I thought the API was not working.

v12023-10-05 150242
nov12023-10-05 150557

banool:
Yeah it's by design that it doesn't work with /v1, our API on the node is versioned.

Saying that, we could make / return a more helpful message. I'll open an issue for that but we will probably close this PR.

cc @saharct to close if that also sounds good to you.

PR #10373, opened 2023-10-03
[Draft] Replacing Aggregator V2 IDs with values when committing the transaction 

vusirikala:
### Description

Replacing Aggregator V2 IDs with values when committing the transaction

### DIFF
--- a/aptos-move/aptos-vm-types/src/change_set.rs
+++ b/aptos-move/aptos-vm-types/src/change_set.rs
@@ -221,9 +221,8 @@ impl VMChangeSet {
             .extend(patched_resource_writes.map(|(k, v)| (k, (v, None))))
     }
 
-    /// The events are replaced with patched events, where the aggregator v2
-    /// identifiers are replaced with values,
-    pub(crate) fn replace_events(&mut self, patched_events: impl Iterator<Item = ContractEvent>) {
+    /// The events are set to the input events.
+    pub(crate) fn set_events(&mut self, patched_events: impl Iterator<Item = ContractEvent>) {
         self.events = patched_events
             .map(|event| (event, None))
             .collect::<Vec<_>>();
--- a/aptos-move/aptos-vm-types/src/change_set.rs
+++ b/aptos-move/aptos-vm-types/src/change_set.rs
@@ -221,12 +221,9 @@ impl VMChangeSet {
             .extend(patched_resource_writes.map(|(k, v)| (k, (v, None))))
     }
 
-    /// The events are replaced with patched events, where the aggregator v2 
-    /// identifiers are replaced with values, 
-    pub(crate) fn replace_events(
-        &mut self,
-        patched_events: impl Iterator<Item = ContractEvent>,
-    ) {
+    /// The events are replaced with patched events, where the aggregator v2
+    /// identifiers are replaced with values,
+    pub(crate) fn replace_events(&mut self, patched_events: impl Iterator<Item = ContractEvent>) {
         self.events = patched_events
             .map(|event| (event, None))
             .collect::<Vec<_>>();
--- a/aptos-move/aptos-vm-types/src/change_set.rs
+++ b/aptos-move/aptos-vm-types/src/change_set.rs
@@ -205,12 +205,31 @@ impl VMChangeSet {
     }
 
     // Called by `try_into_transaction_output_with_materialized_writes` only.
-    pub(crate) fn extend_aggregator_write_set(
+    pub(crate) fn extend_aggregator_v1_write_set(
         &mut self,
         additional_aggregator_writes: impl Iterator<Item = (StateKey, WriteOp)>,
     ) {
         self.aggregator_v1_write_set
             .extend(additional_aggregator_writes)
+    }
+
+    pub(crate) fn extend_resource_write_set(
+        &mut self,
+        patched_resource_writes: impl Iterator<Item = (StateKey, WriteOp)>,
+    ) {
+        self.resource_write_set
+            .extend(patched_resource_writes.map(|(k, v)| (k, (v, None))))
+    }
+
+    /// The events are replaced with patched events, where the aggregator v2 
+    /// identifiers are replaced with values, 
+    pub(crate) fn replace_events(
+        &mut self,
+        patched_events: impl Iterator<Item = ContractEvent>,
+    ) {
+        self.events = patched_events
+            .map(|event| (event, None))
+            .collect::<Vec<_>>();
     }
 
     pub fn aggregator_v1_write_set(&self) -> &HashMap<StateKey, WriteOp> {
--- a/aptos-move/aptos-vm-types/src/output.rs
+++ b/aptos-move/aptos-vm-types/src/output.rs
@@ -7,10 +7,10 @@ use aptos_types::{
     fee_statement::FeeStatement,
     state_store::state_key::StateKey,
     transaction::{TransactionOutput, TransactionStatus},
-    write_set::WriteOp,
+    write_set::WriteOp, contract_event::ContractEvent, //contract_event::ContractEvent,
 };
 use move_core_types::vm_status::VMStatus;
-
+use std::collections::HashMap;
 /// Output produced by the VM after executing a transaction.
 ///
 /// **WARNING**: This type should only be used inside the VM. For storage backends,
@@ -53,6 +53,10 @@ impl VMOutput {
 
     pub fn change_set(&self) -> &VMChangeSet {
         &self.change_set
+    }
+
+    pub fn change_set_mut(&mut self) -> &mut VMChangeSet {
+        &mut self.change_set
     }
 
     pub fn gas_used(&self) -> u64 {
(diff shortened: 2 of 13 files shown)


aptos-move/block-executor/src/executor.rs
```diff
@@ -409,13 +413,109 @@ where
...
+        // For each resource that satisfies the following conditions,
+        //     1. Resource is in read set
+        //     2. Resource is not in write set
+        // replace the aggregator v2 identifiers in the resource with corresponding values.
+        // If any of the aggregator v2 identifiers in the resource are part of aggregator_v2_write_set,
+        // then include the resource in the write set.
+        let aggregator_v2_keys = last_input_output.aggregator_v2_keys(txn_idx);
+        let read_set = last_input_output.read_set(txn_idx);
+        // TODO: Gives some error without this `mut`. See if we can avoid `mut` here.
+        if let Some(mut aggregator_v2_keys) = aggregator_v2_keys {
```
igor-aptos, line 457: this is an iterator, so you consume it the first time. 

instead collect it into a HashSet (for fast lookup), and then you'll not need mut

aptos-move/block-executor/src/executor.rs
```diff
@@ -409,13 +413,109 @@ where
...
+            if let Some(read_set) = read_set {
+                for (key, data_read) in read_set.as_ref().data_reads.iter() {
```
igor-aptos, line 459: instead of making data_reads public, can we have a function that returns iterator over (&V, &MoveTypeLayout) - i.e. get_versioned_read_values() or something

aptos-move/block-executor/src/executor.rs
```diff
@@ -409,13 +413,109 @@ where
...
+                        continue;
+                    }
+                    // layout is Some(_) if it contains an aggregator
+                    if let DataRead::Versioned(_version, value, Some(layout)) = data_read {
+                        if let Some(value_bytes) = value.bytes() {
+                            match latest_view.replace_identifiers_with_values(value_bytes, layout) {
+                                Ok((patched_bytes, aggregator_v2_keys_in_resource)) => {
+                                    let mut changed_aggregator = false;
+                                    for aggregator_v2_key in aggregator_v2_keys_in_resource {
+                                        if aggregator_v2_keys.any(|key| key == aggregator_v2_key) {
```
igor-aptos, line 470: contains on a HashSet, instead of any?

types/src/write_set.rs
```diff
@@ -237,6 +239,16 @@ impl TransactionWrite for WriteOp {
...
         }
     }
+
+    fn set_bytes(&mut self, bytes: Bytes) {
+        use WriteOp::*;
+
+        match self {
+            Creation(data) | CreationWithMetadata { data, .. } => *data = bytes,
+            Modification(data) | ModificationWithMetadata { data, .. } => *data = bytes,
+            Deletion | DeletionWithMetadata { .. } => (),
```
igor-aptos, line 249: should we put unreachable here , for safety?

igor-aptos (changes requested):
we should do one pass later to check any copies we are making, and if they are necessary, but we can leave that for now

for correctness - added comments inline

aptos-move/block-executor/src/executor.rs
```diff
@@ -409,13 +413,109 @@ where
...
+        let latest_view = LatestView::new(base_view, ViewState::Sync(parallel_state), txn_idx);
+
+        // For each aggregator v2 in resource write set, replace the identifiers with values.
+        let mut write_set_keys = vec![];
+        let resource_write_set = last_input_output.resource_write_set(txn_idx);
+        let mut patched_resource_write_set = HashMap::new();
+        if let Some(resource_write_set) = resource_write_set {
+            for (key, (write_op, layout)) in resource_write_set.iter() {
+                // layout is Some(_) if it contains an aggregator
+                if layout.is_some() && !write_op.is_deletion() {
```
georgemitenkov, line 432: Cane use `if let` here to avoid unwraps later

aptos-move/block-executor/src/executor.rs
```diff
@@ -409,13 +413,109 @@ where
...
+                    write_set_keys.push(key.clone());
```
georgemitenkov, line 433: Seems like we need this only for tracking, but we have all the info in `patched_resource_write_set` already? You can call contains on that

aptos-move/block-executor/src/executor.rs
```diff
@@ -409,13 +413,109 @@ where
...
+        }
+
+        // For each aggregator v2 in the event, replace aggregator v2 identifier with value.
+        let events = last_input_output.events(txn_idx);
+        let mut patched_events = vec![];
+        for (event, layout) in events {
+            if let Some(layout) = layout {
+                let (_, _, _, event_data) = event.get_event_data();
+                match latest_view
+                    .replace_identifiers_with_values(&Bytes::from(event_data.to_vec()), &layout)
```
georgemitenkov, line 501: event_data.to_vec().into()? or event_data.into()

aptos-move/block-executor/src/executor.rs
```diff
@@ -409,13 +413,109 @@ where
...
+                            // TODO: Is this unreachable?
+                            unreachable!("Data read value must exist");
+                        }
+                    }
+                }
+            }
+        }
+
+        // For each aggregator v2 in the event, replace aggregator v2 identifier with value.
+        let events = last_input_output.events(txn_idx);
```
georgemitenkov, line 495: We clone out events here, no need to reclone it later in the loop?

aptos-move/block-executor/src/executor.rs
```diff
@@ -409,13 +413,109 @@ where
...
+        let mut patched_events = vec![];
+        for (event, layout) in events {
+            if let Some(layout) = layout {
+                let (_, _, _, event_data) = event.get_event_data();
+                match latest_view
+                    .replace_identifiers_with_values(&Bytes::from(event_data.to_vec()), &layout)
+                {
+                    Ok((bytes, _)) => {
+                        let mut patched_event = event.clone();
```
georgemitenkov, line 504: similar to above: if we consume events, no need to do clones etc. I think

aptos-move/block-executor/src/executor.rs
```diff
@@ -409,13 +413,109 @@ where
...
+        // replace the aggregator v2 identifiers in the resource with corresponding values.
+        // If any of the aggregator v2 identifiers in the resource are part of aggregator_v2_write_set,
+        // then include the resource in the write set.
+        let aggregator_v2_keys = last_input_output.aggregator_v2_keys(txn_idx);
+        let read_set = last_input_output.read_set(txn_idx);
+        // TODO: Gives some error without this `mut`. See if we can avoid `mut` here.
+        if let Some(mut aggregator_v2_keys) = aggregator_v2_keys {
+            if let Some(read_set) = read_set {
+                for (key, data_read) in read_set.as_ref().data_reads.iter() {
+                    if write_set_keys.contains(key) {
```
georgemitenkov, line 460: here we can use patched_resource_write_set instead.

aptos-move/block-executor/src/executor.rs
```diff
@@ -409,13 +413,109 @@ where
...
+                        continue;
+                    }
+                    // layout is Some(_) if it contains an aggregator
+                    if let DataRead::Versioned(_version, value, Some(layout)) = data_read {
+                        if let Some(value_bytes) = value.bytes() {
+                            match latest_view.replace_identifiers_with_values(value_bytes, layout) {
+                                Ok((patched_bytes, aggregator_v2_keys_in_resource)) => {
+                                    let mut changed_aggregator = false;
+                                    for aggregator_v2_key in aggregator_v2_keys_in_resource {
```
georgemitenkov, line 469: HashSet has [is_disjoint](https://doc.rust-lang.org/std/collections/struct.HashSet.html#method.is_disjoint) to avoid loops

aptos-move/block-executor/src/executor.rs
```diff
@@ -771,6 +882,7 @@ where
                     }
 
                     // No delta writes are needed for sequential execution.
+                    // TODO: Should we replace this with output.incorporate_materialized_txn_output(..)?
```
georgemitenkov, line 885: I think so :)

aptos-move/block-executor/src/view.rs
```diff
@@ -364,6 +365,33 @@ impl<'a, T: Transaction, S: TStateView<Key = T::Key>, X: Executable> LatestView<
...
+        &self,
+        bytes: &Bytes,
+        layout: &MoveTypeLayout,
+    ) -> anyhow::Result<(Bytes, Vec<T::Identifier>)> {
+        // TODO: Find a way to replace this with Self::IdentifierV2
+        // This call will replace all occurrences of aggregator / snapshot
+        // identifiers with values with the same type layout.
+        let value = Value::simple_deserialize(bytes, layout).ok_or_else(|| {
+            anyhow::anyhow!(
+                "Failed to deserialize resource during value replacement: {:?}",
```
georgemitenkov, line 380: nit: id replacement?

aptos-move/block-executor/src/view.rs
```diff
@@ -364,6 +365,33 @@ impl<'a, T: Transaction, S: TStateView<Key = T::Key>, X: Executable> LatestView<
...
     }
 
+    /// Given a state value, performs deserialization-serialization round-trip
+    /// to replace any aggregator / snapshot values.
+    pub(crate) fn replace_identifiers_with_values(
+        &self,
+        bytes: &Bytes,
+        layout: &MoveTypeLayout,
+    ) -> anyhow::Result<(Bytes, Vec<T::Identifier>)> {
+        // TODO: Find a way to replace this with Self::IdentifierV2
```
georgemitenkov, line 375: Aren't these two the same?

aptos-move/aptos-vm-types/src/change_set.rs
```diff
@@ -205,14 +205,30 @@ impl VMChangeSet {
...
+        &mut self,
+        patched_resource_writes: impl Iterator<Item = (StateKey, WriteOp)>,
+    ) {
+        self.resource_write_set
+            .extend(patched_resource_writes.map(|(k, v)| (k, (v, None))))
+    }
+
+    /// The events are replaced with patched events, where the aggregator v2
+    /// identifiers are replaced with values,
+    pub(crate) fn replace_events(&mut self, patched_events: impl Iterator<Item = ContractEvent>) {
```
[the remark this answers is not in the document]
igor-aptos: it does? 

self.event = new thing

aptos-move/block-executor/src/executor.rs
igor-aptos: if let doesn't yet allow && something else, and adds more nesting into this nested loop. 

I am fine either way

aptos-move/block-executor/src/executor.rs
igor-aptos: we call write_set_keys.contains in the next section, that continues changing patched_resource_write_set, so it's not exactly the same

aptos-move/block-executor/src/executor.rs
igor-aptos: caveat above, not sure if that is doable

aptos-move/block-executor/src/executor.rs
igor-aptos: separate PR, leave TODO in code :)

aptos-move/block-executor/src/executor.rs
georgemitenkov: i see, yes, then a separate hashset is the way

aptos-move/aptos-vm-types/src/change_set.rs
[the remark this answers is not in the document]
georgemitenkov: nvm, I am blind

aptos-move/aptos-vm-types/src/output.rs
```diff
@@ -118,21 +123,57 @@ impl VMOutput {
     /// externally by the caller beforehand.
     pub fn into_transaction_output_with_materialized_deltas(
```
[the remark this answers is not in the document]
vusirikala: Still using it in the sequential execution now. I will replace this function soon in the sequential execution as well.

aptos-move/aptos-vm-types/src/change_set.rs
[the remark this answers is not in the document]
vusirikala: This replaces self.events with patched_events.

aptos-move/block-executor/src/executor.rs
vusirikala: Tried it. into() function is not implemented here.

aptos-move/block-executor/src/view.rs
vusirikala: They are the same. I just thought you don't like repeated code ;)

igor-aptos (approved):

gelash (commented):
I think main important piece is about not cloning the whole resource write set. We do once but that one is hard to avoid atm, but it's expensive, and here we should avoid (or minimize the impact) - see my comment on last_input_output (and executor)

PR #10384, opened 2023-10-03
[TS SDK V2] Rename enum variants and transaction arguments

0xmaayan:
### Description
- remove RustEnum from the enum variants name
- rename TransactionArguments to ScriptTransactionArguments as they are only used in a script trnasaction


xbtmatt (approved):
Great! the variant names are much clearer than before, esp with link to the rust code

heliuchuan (approved):

PR #10386, opened 2023-10-04
Rename common classes to DelayedField preffix

igor-aptos:


aptos-move/aptos-aggregator/src/delayed_field_extension.rs
```diff
@@ -637,7 +637,7 @@ impl AggregatorData {
         BTreeSet<AggregatorVersionedID>,
```
vusirikala, line 637: Does AggregatorerVersionedID have to be renamed as we renamed AggregatorID?

igor-aptos: no, that one is only used for AggregatorV1 vs AggregatorV2 in the extension, as far as I can see. 

we always use DelayedFieldID for Snapshots

vusirikala (approved):

PR #10387, opened 2023-10-04
[TS-SDK v2] Adding TypeTag and fixing small doc/formatting errors

xbtmatt:
Adding the `TypeTag` type that creates generics into the SDK.

This is largely ported over from the existing SDK with just a few minor changes to which error codes are thrown in unit tests and things like constructing `AccountAddress`es.

Also imported `TypeTag` into `transactionPayload.ts`

### Test Plan
```shell
cd ~/aptos-core/ecosystem/typescript/sdk_v2 && pnpm jest type_tag.test.ts
```


ecosystem/typescript/sdk_v2/tests/unit/type_tag.test.ts
```diff
@@ -0,0 +1,291 @@
...
+  TypeTagU64,
+  TypeTagU8,
+  TypeTagVector,
+} from "../../src/transactions/types/typeTag";
+import { Deserializer, Serializer } from "../../src/bcs";
+import { AccountAddress, AddressInvalidReason, ParsingError } from "../../src/core";
+
+const expectedTypeTag = {
+  string: "0x1::some_module::SomeResource",
+  address: "0x1",
```
0xmaayan, line 24: should we also test for other framework addresses?

ecosystem/typescript/sdk_v2/src/transactions/types/typeTag.ts
```diff
@@ -0,0 +1,488 @@
...
+function isGeneric(c: string): boolean {
+  if (c.match(/T\d+/g)) {
+    return true;
+  }
+  return false;
+}
+
+type TokenType = string;
+type TokenValue = string;
+type Token = [TokenType, TokenValue];
```
0xmaayan, line 274: is there a real reason to have these types? looks like TokenType only used in the Token type and TokenValue in only one more place other then the Token type?

0xmaayan (changes requested):
Overall looks good. I think we can split the TypeTag enum variants and the actual TypeTagParser part.
Also, please make sure to run `pnpm run lint` to make sure no linting issues

ecosystem/typescript/sdk_v2/src/client/get.ts
```diff
@@ -71,9 +71,7 @@ export async function get<Req, Res>(options: GetRequestOptions): Promise<AptosRe
   return response;
 }
 
-export async function getAptosFullNode<Req, Res>(
-  options: GetAptosRequestOptions,
-): Promise<AptosResponse<Req, Res>> {
+export async function getAptosFullNode<Req, Res>(options: GetAptosRequestOptions): Promise<AptosResponse<Req, Res>> {
```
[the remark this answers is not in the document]
xbtmatt: I did run `pnpm fmt` and `pnpm lint` btw, the only reason this is changed is because I ran it :p

the linting errors that remain are from the public key interface not being complete (I think)

ecosystem/typescript/sdk_v2/src/transactions/types/typeTag.ts
```diff
@@ -0,0 +1,488 @@
...
+    serializer.serializeU32AsUleb128(7);
+    this.value.serialize(serializer);
+  }
+
+  static load(deserializer: Deserializer): TypeTagStruct {
+    const value = StructTag.deserialize(deserializer);
+    return new TypeTagStruct(value);
+  }
+
+  isStringTypeTag(): boolean {
```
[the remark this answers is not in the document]
0xmaayan: gotcha

ecosystem/typescript/sdk_v2/src/transactions/types/typeTag.ts
xbtmatt: The more I see things like these, the more I realize they're all just aliased types purely for the sake of code readability.

Can remove it if you want, I think in this case it does increase the readability a bit, but if you think it's just more confusing than anything can just remove it

ecosystem/typescript/sdk_v2/tests/unit/type_tag.test.ts
xbtmatt: I think the only difference with the other framework addresses would be parsing on the `AccountAddress` side of things. Other than that there are already tests for making sure things like object/option/string are `0x1`

Unless I'm misunderstanding..?

ecosystem/typescript/sdk_v2/src/transactions/types/typeTag.ts
xbtmatt: but I will change it cuz it's just confusing

ecosystem/typescript/sdk_v2/src/transactions/types/scriptTransactionArguments.ts
```diff
@@ -18,23 +18,23 @@ export abstract class ScriptTransactionArgument extends Serializable {
     // index enum variant
     const index = deserializer.deserializeUleb128AsU32();
     switch (index) {
-      case ScriptTransactionArgumentVariants.ScriptTransactionArgumentU8:
+      case ScriptTransactionArgumentVariants.U8:
```
0xmaayan, line 21: thanks

0xmaayan (approved):
LGTM

ecosystem/typescript/sdk_v2/src/transactions/types/typeTag.ts
```diff
@@ -0,0 +1,252 @@
...
+import { AccountAddress } from "../../core";
+import { Deserializer, Serializable, Serializer } from "../../bcs";
+import { Identifier } from "./identifier";
+import { TypeTagParser } from "./typeTagParser";
+import { TypeTagVariants } from "../../types";
+
+export abstract class TypeTag extends Serializable {
+  abstract serialize(serializer: Serializer): void;
+
+  static deserialize(deserializer: Deserializer): TypeTag {
```
0xjinn, line 16: Looking at this func, maybe I should do the same to PrivateKey and PublicKey classes - To define the deserialize method at the base level.

ecosystem/typescript/sdk_v2/src/transactions/types/typeTag.ts
```diff
@@ -0,0 +1,252 @@
...
+}
+
+export class TypeTagBool extends TypeTag {
+  serialize(serializer: Serializer): void {
+    serializer.serializeU32AsUleb128(TypeTagVariants.Bool);
+  }
+
+  static load(_deserializer: Deserializer): TypeTagBool {
+    return new TypeTagBool();
+  }
```
0xjinn, line 54: Am I missing something? Looks like deserializer is not being use here? (Same as other classes)

0xjinn (approved):

ecosystem/typescript/sdk_v2/src/transactions/types/typeTag.ts
xbtmatt: I think this was just to implement a common interface where `load` takes a `deserializer`, but in this case, it isn't used, because there's no extra data in a serialized `TypeTagBool` other than the initial enum variant-

that is, there's only 1 byte when you serialize a `TypeTag`, and that's the enum variant. deserializing it results in a call from `deserialize` in `TypeTag` and then `load` in the corresponding type. Since there's nothing else to deserialize, `load` ignores the `deserializer` arg!

ecosystem/typescript/sdk_v2/src/transactions/types/typeTag.ts
xbtmatt: This abstract factory type of stuff only makes sense when what you're deserializing is represented as an `enum` on the Rust side of things.

In the case of `PrivateKey` and `PublicKey` classes, those are already a part of the `AccountAuthenticator` factory methods, where `AccountAuthenticator` would be the base class responsible for the `static deserialize(...)` method

heliuchuan (approved):

PR #10390, opened 2023-10-04
Replacing aggregator v2 identifiers with values before transaction commitment

vusirikala:
### Description

Replacing aggregator v2 identifiers with values before transaction commitment. Finishing some of the TODOs and PR comments on the previous PR.

### DIFF
--- a/aptos-move/block-executor/src/captured_reads.rs
+++ b/aptos-move/block-executor/src/captured_reads.rs
@@ -156,7 +156,7 @@ struct GroupRead<T: Transaction> {
 #[derive(Derivative)]
 #[derivative(Default(bound = "", new = "true"))]
 pub(crate) struct CapturedReads<T: Transaction> {
-    pub(crate) data_reads: HashMap<T::Key, DataRead<T::Value>>,
+    data_reads: HashMap<T::Key, DataRead<T::Value>>,
     group_reads: HashMap<T::Key, GroupRead<T>>,
     // Currently, we record paths for triggering module R/W fallback.
     // TODO: implement a general functionality once the fallback is removed.
@@ -184,6 +184,15 @@ enum UpdateResult {
 }
 
 impl<T: Transaction> CapturedReads<T> {
+    // Return an iterator over the captured reads.
+    pub(crate) fn get_read_values_with_delayed_fields(
+        &self,
+    ) -> impl Iterator<Item = (&T::Key, &DataRead<T::Value>)> {
+        self.data_reads
+            .iter()
+            .filter(|(_, v)| matches!(v, DataRead::Versioned(_, _, Some(_))))
+    }
+
     // Given a hashmap entry for a key, incorporate a new DataRead. This checks
     // consistency and ensures that the most comprehensive read is recorded.
     fn update_entry<K, V: TransactionWrite>(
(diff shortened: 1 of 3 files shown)


aptos-move/block-executor/src/executor.rs
```diff
@@ -562,6 +561,30 @@ where
...
+        txn_idx: TxnIndex,
+        versioned_cache: &MVHashMap<T::Key, T::Tag, T::Value, X, T::Identifier>,
+        scheduler: &Scheduler,
+        shared_counter: &AtomicU32,
+        last_input_output: &TxnLastInputOutput<T, E::Output, E::Error>,
+        base_view: &S,
+    ) {
+        let parallel_state = ParallelState::<T, X>::new(versioned_cache, scheduler, shared_counter);
+        let latest_view = LatestView::new(base_view, ViewState::Sync(parallel_state), txn_idx);
+        let patched_resource_write_set = Self::map_delayed_field_id_to_values_in_read_and_write_set(
```
georgemitenkov, line 578: nit: having simpler function names would make it more readable

aptos-move/block-executor/src/view.rs
```diff
@@ -651,11 +653,52 @@ impl<'a, T: Transaction, S: TStateView<Key = T::Key>, X: Executable> TDelayedFie
...
+                let mut counter = state.counter.borrow_mut();
+                let id = (*counter as u64).into();
+                *counter += 1;
+                id
+            },
+        }
+    }
+
+    pub fn into(self) -> HashSet<T::Identifier> {
+        self.delayed_field_keys.borrow().clone()
```
georgemitenkov, line 693: I think you can use into_inner or something like that to take from the cell

georgemitenkov (approved):
Beautiful

aptos-move/block-executor/src/captured_reads.rs
```diff
@@ -184,6 +184,15 @@ enum UpdateResult {
 }
 
 impl<T: Transaction> CapturedReads<T> {
+    // Return an iterator over the captured reads.
+    pub(crate) fn get_read_values_with_delayed_fields(
+        &self,
+    ) -> impl Iterator<Item = (&T::Key, &DataRead<T::Value>)> {
```
igor-aptos, line 190: can we map to pair of (&V, &MoveTypeLayout), instead of leaking DataRead?

that way we can have the sequential code provide the same iterator potentially?

aptos-move/block-executor/src/executor.rs
```diff
@@ -535,6 +533,25 @@ where
...
+    ) {
+        let parallel_state = ParallelState::<T, X>::new(versioned_cache, scheduler, shared_counter);
+        let latest_view = LatestView::new(base_view, ViewState::Sync(parallel_state), txn_idx);
+        let patched_resource_write_set =
+            Self::map_id_to_values_in_read_write_set(txn_idx, last_input_output, &latest_view);
+        let patched_events =
+            Self::map_id_to_values_events(txn_idx, last_input_output, &latest_view);
 
         // Materialize all the aggregator v1 deltas.
         let aggregator_v1_delta_keys = last_input_output.aggregator_v1_delta_keys(txn_idx);
```
igor-aptos, line 557: I know it's unrelated to your change, but can you extract agg_v1 handling into their own function as well?

igor-aptos (approved):

PR #10394, opened 2023-10-04
[docker] Set image hash to multiplatform ones

ibalajiarun:
### Description

Currently, our docker builders can only build linux/amd64 images because we hardcode base-image SHA hashes to linux/amd64 ones. But, what we really should be using are the multi-platform image hashes from the `docker buildx imagetools inspect` tool. This PR updates those hashes.

This will allow us to build images for both macos/arm and linux/amd.

### Test Plan

Wait for CI

### DIFF
--- a/docker/builder/debian-base.Dockerfile
+++ b/docker/builder/debian-base.Dockerfile
@@ -2,8 +2,10 @@
 
 FROM debian AS debian-base
 
+ARG TARGETARCH
+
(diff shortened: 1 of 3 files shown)


gedigi (approved):

JoshLind (approved):
Nice, thanks @ibalajiarun 😄

PR #10397, opened 2023-10-04
[agg_v2] Adding aggregator_v2.move with sequential (fallback) logic

igor-aptos:
Self-contained logic that can be landed on main, and so ecosystem PRs can land independently of the rest of the aggregator changes

once aggregator backend execution logic lands, this will be a fallback logic used until separate aggregator execution flag is turned on.

Copied from aggregators_v2 branch

### DIFF
--- a/aptos-move/aptos-gas-schedule/src/gas_schedule/aptos_framework.rs
+++ b/aptos-move/aptos-gas-schedule/src/gas_schedule/aptos_framework.rs
@@ -201,10 +201,17 @@ crate::gas_schedule::macros::define_gas_parameters!(
         [aggregator_destroy_base: InternalGas, "aggregator.destroy.base", 10000],
         [aggregator_factory_new_aggregator_base: InternalGas, "aggregator_factory.new_aggregator.base", 10000],
 
+        [aggregator_v2_create_aggregator_base: InternalGas, {12.. => "aggregator_v2.create_aggregator.base"}, 10000],
+        [aggregator_v2_try_add_base: InternalGas, {12.. => "aggregator_v2.try_add.base"}, 6000],
+        [aggregator_v2_try_sub_base: InternalGas, {12.. => "aggregator_v2.try_sub.base"}, 6000],
+        [aggregator_v2_read_base: InternalGas, {12.. => "aggregator_v2.read.base"}, 12000],
+        [aggregator_v2_snapshot_base: InternalGas, {12.. => "aggregator_v2.snapshot.base"}, 6000],
+
         [aggregator_v2_create_snapshot_base: InternalGas, {11.. => "aggregator_v2.create_snapshot.base"}, 6000],
         [aggregator_v2_copy_snapshot_base: InternalGas, {11.. => "aggregator_v2.copy_snapshot.base"}, 6000],
-        [aggregator_v2_read_snapshot_base: InternalGas, {11.. => "aggregator_v2.read_snapshot.base"}, 6000],
+        [aggregator_v2_read_snapshot_base: InternalGas, {11.. => "aggregator_v2.read_snapshot.base"}, 12000],
         [aggregator_v2_string_concat_base: InternalGas, {11.. => "aggregator_v2.string_concat.base"}, 6000],
+        [aggregator_v2_string_concat_per_byte: InternalGasPerByte, { 12.. =>"aggregator_v2.string_concat.per_byte" }, 20],
 
         [object_exists_at_base: InternalGas, { 7.. => "object.exists_at.base" }, 5000],
         // These are dummy value, they copied from storage gas in aptos-core/aptos-vm/src/aptos_vm_impl.rs
--- a/aptos-move/framework/aptos-framework/sources/aggregator_v2/aggregator_v2.move
+++ b/aptos-move/framework/aptos-framework/sources/aggregator_v2/aggregator_v2.move
@@ -27,7 +27,7 @@ module aptos_framework::aggregator_v2 {
     /// The generic type supplied to the aggregator is not supported.
     const EUNSUPPORTED_AGGREGATOR_TYPE: u64 = 7;
 
-    /// Arguments passed to concat exceed max limit of 256 bytes (for prefix and suffix together)
+    /// Arguments passed to concat exceed max limit of 256 bytes (for prefix and suffix together).
     const ECONCAT_STRING_LENGTH_TOO_LARGE: u64 = 8;
 
     /// The native aggregator function, that is in the move file, is not yet supported.
@@ -37,7 +37,7 @@ module aptos_framework::aggregator_v2 {
     /// Represents an integer which supports parallel additions and subtractions
     /// across multiple transactions. See the module description for more details.
     ///
-    /// Currently supported types for Element are u64 and u128.
+    /// Currently supported types for IntElement are u64 and u128.
     struct Aggregator<IntElement> has store, drop {
         value: IntElement,
         max_value: IntElement,
@@ -57,14 +57,14 @@ module aptos_framework::aggregator_v2 {
 
     /// Creates new aggregator, with given 'max_value'.
     ///
-    /// Currently supported types for Element are u64 and u128.
+    /// Currently supported types for IntElement are u64 and u128.
     /// EAGGREGATOR_ELEMENT_TYPE_NOT_SUPPORTED raised if called with a different type.
     public native fun create_aggregator<IntElement: copy + drop>(max_value: IntElement): Aggregator<IntElement>;
 
     /// Creates new aggregator, without any 'max_value' on top of the implicit bound restriction
     /// due to the width of the type (i.e. MAX_U64 for u64, MAX_U128 for u128).
     ///
-    /// Currently supported types for Element are u64 and u128.
+    /// Currently supported types for IntElement are u64 and u128.
     /// EAGGREGATOR_ELEMENT_TYPE_NOT_SUPPORTED raised if called with a different type.
     public native fun create_unbounded_aggregator<IntElement: copy + drop>(): Aggregator<IntElement>;
 
@@ -73,7 +73,7 @@ module aptos_framework::aggregator_v2 {
     public native fun try_add<IntElement>(aggregator: &mut Aggregator<IntElement>, value: IntElement): bool;
 
     // Adds `value` to aggregator, uncoditionally.
-    // If addition would exceed the max_value, EAGGREGATOR_OVERFLOW exception will be thrown
+    // If addition would exceed the max_value, EAGGREGATOR_OVERFLOW exception will be thrown.
     public fun add<IntElement>(aggregator: &mut Aggregator<IntElement>, value: IntElement) {
         assert!(try_add(aggregator, value), error::out_of_range(EAGGREGATOR_OVERFLOW));
     }
@@ -82,8 +82,8 @@ module aptos_framework::aggregator_v2 {
     /// If subtraction would result in a negative value, `false` is returned, and aggregator value is left unchanged.
     public native fun try_sub<IntElement>(aggregator: &mut Aggregator<IntElement>, value: IntElement): bool;
 
-    // Adds `value` to aggregator, uncoditionally.
-    // If subtraction would result in a negative value, EAGGREGATOR_UNDERFLOW exception will be thrown
+    // Subtracts `value` to aggregator, uncoditionally.
+    // If subtraction would result in a negative value, EAGGREGATOR_UNDERFLOW exception will be thrown.
     public fun sub<IntElement>(aggregator: &mut Aggregator<IntElement>, value: IntElement) {
         assert!(try_sub(aggregator, value), error::out_of_range(EAGGREGATOR_UNDERFLOW));
     }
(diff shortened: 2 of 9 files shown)


aptos-move/aptos-native-interface/src/context.rs
```diff
@@ -123,4 +123,9 @@ impl<'a, 'b, 'c, 'd> SafeNativeContext<'a, 'b, 'c, 'd> {
     pub fn aggregator_snapshots_enabled(&self) -> bool {
         self.get_feature_flags().is_aggregator_snapshots_enabled()
     }
+
+    /// TODO: decide if we want to have two flags or merge with aggregator_snapshots_enabled.
```
igor-aptos, line 127: we are introducing new native functions that need to be gated. 

since aggregator_snapshots_enabled was never enabled, we can either:
- decide not to enable it before v1.8, and combine aggregator_api_enabled into it. 
- if we want to enable in point release for v1.7, we would need to add a new flag instead.

aptos-move/framework/aptos-framework/sources/aggregator_v2/aggregator_v2.move
```diff
@@ -1,97 +1,160 @@
...
+    /// Currently supported types for Element are u64 and u128.
+    /// EAGGREGATOR_ELEMENT_TYPE_NOT_SUPPORTED raised if called with a different type.
+    public native fun create_aggregator<IntElement: copy + drop>(max_value: IntElement): Aggregator<IntElement>;
+
+    /// Creates new aggregator, without any 'max_value' on top of the implicit bound restriction
+    /// due to the width of the type (i.e. MAX_U64 for u64, MAX_U128 for u128).
+    ///
+    /// Currently supported types for Element are u64 and u128.
+    /// EAGGREGATOR_ELEMENT_TYPE_NOT_SUPPORTED raised if called with a different type.
+    public native fun create_unbounded_aggregator<IntElement: copy + drop>(): Aggregator<IntElement>;
```
vusirikala, line 50: Does this have to be a native function? It can simply call create_aggregator with an appropriate max_value right?

igor-aptos: I don't think it can be done, because the type is generic.

i.e. something like this doesn't compile:

    public fun create_unbounded_aggregator<IntElement: copy + drop>(): Aggregator<IntElement> {
        if (type_info::type_of<IntElement>() == type_info::type_of<u64>()) {
            create_aggregator(MAX_U64)
        } else if (type_info::type_of<IntElement>() == type_info::type_of<u128>()) {
            create_aggregator(MAX_U128)
        } else {
            abort error::invalid_argument(EUNSUPPORTED_AGGREGATOR_TYPE);
        }
    }


this compiles:
`if (type_info::type_of<Element>() == type_info::type_of<u64>()) {`

this doesn't:
`return create_aggregator(MAX_U64)`
when return type of the function is : Aggregator<IntElement>

Compiler could probably infer something, but even in C++/rust not sure that would pass. Not sure if there is a different way to write this.

aptos-move/framework/aptos-framework/sources/aggregator_v2/aggregator_v2.move
```diff
@@ -1,97 +1,160 @@
...
+        let feature = features::get_aggregator_snapshots_feature();
+        features::change_feature_flags(fx, vector[feature], vector[]);
+
+        let snapshot = create_snapshot(42);
+        let snapshot2 = string_concat(std::string::utf8(b"before"), &snapshot, std::string::utf8(b"after"));
+        assert!(read_snapshot(&snapshot2) == std::string::utf8(b"before42after"), 0);
+    }
+
+    #[test(fx = @std)]
+    public fun test_string_concat2(fx: &signer) {
```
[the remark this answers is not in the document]
igor-aptos: there was a bug in how features were handled in the harness, that was fixed. I'll recheck , but I think they pass now

aptos-move/framework/aptos-framework/sources/aggregator_v2/aggregator_v2.move
```diff
@@ -1,97 +1,160 @@
...
+        features::change_feature_flags(fx, vector[feature], vector[]);
+
+        let snapshot = create_snapshot(42);
+        let snapshot2 = copy_snapshot(&snapshot);
+        assert!(read_snapshot(&snapshot) == 42, 0);
+        assert!(read_snapshot(&snapshot2) == 42, 0);
+    }
+
+    #[test(fx = @std)]
+    public fun test_correct_read_string(fx: &signer) {
```
vusirikala, line 101: Do these tests succeed now? They didn't work last time when I tried to merge aggregator snapshots into main.

aptos-move/framework/src/natives/aggregator_natives/aggregator_v2.rs
```diff
@@ -39,94 +50,418 @@ fn is_string_type(context: &SafeNativeContext, type_arg: &Type) -> SafeNativeRes
     Ok(false)
 }
 
+/// Given the list of native function arguments and a type, returns a tuple of its
```
vusirikala, line 53: Nit: We are only giving one native function argument.

aptos-move/framework/aptos-framework/sources/aggregator_v2/aggregator_v2.move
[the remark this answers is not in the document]
igor-aptos: seems like I imagined something, they still don't work.

aptos-move/framework/src/natives/aggregator_natives/aggregator_v2.rs
```diff
@@ -39,94 +50,418 @@ fn is_string_type(context: &SafeNativeContext, type_arg: &Type) -> SafeNativeRes
...
+        ))?)
+    }
+
+    pub fn pop_snapshot_value_by_type(
+        &self,
+        args: &mut VecDeque<Value>,
+    ) -> SafeNativeResult<SnapshotValue> {
+        match self {
+            SnapshotType::U128 => Ok(SnapshotValue::Integer(safely_pop_arg!(args, u128))),
+            SnapshotType::U64 => Ok(SnapshotValue::Integer(safely_pop_arg!(args, u64) as u128)),
```
vusirikala, line 138: Note for the future: SnapshotValue is replaced by DelayedFieldValue right?

aptos-move/framework/src/natives/aggregator_natives/aggregator_v2.rs
```diff
@@ -39,94 +50,418 @@ fn is_string_type(context: &SafeNativeContext, type_arg: &Type) -> SafeNativeRes
...
+    String(Vec<u8>),
+}
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub enum SnapshotToStringFormula {
+    Concat { prefix: Vec<u8>, suffix: Vec<u8> },
+}
+
+impl SnapshotToStringFormula {
+    pub fn apply_to(&self, base: u128) -> Vec<u8> {
```
vusirikala, line 186: What is we are deriving the snapshot from aggregator_u64? Are we still going to treat u64 as u128 when creating string snapshot?

aptos-move/framework/src/natives/aggregator_natives/aggregator_v2.rs
```diff
@@ -39,94 +50,418 @@ fn is_string_type(context: &SafeNativeContext, type_arg: &Type) -> SafeNativeRes
...
+            }),
+        }
+    }
+}
+
+// ================= START TEMPORARY CODE =================
+// TODO: aggregator_v2 branch will introduce these in different places in code
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum SnapshotValue {
```
vusirikala, line 175: We are going to land aggregator_v2 branch on main in a week anyway. What are we benefitting by landing these stuff here in main now, and changing them again after landing aggregator_v2?

aptos-move/framework/src/natives/aggregator_natives/aggregator_v2.rs
igor-aptos: no - SnapshotValue encompasses types that AggregatorSnapshot can take


DelayedFieldValue is list of Delayed Fields (today aggregator, snapshot delta and snapshot derived. tomorrow maybe more).

this code is identical on aggregators_v2 branch - https://github.com/aptos-labs/aptos-core/blob/aggregators_v2/aptos-move/framework/src/natives/aggregator_natives/aggregator_v2.rs#L143

aptos-move/framework/src/natives/aggregator_natives/aggregator_v2.rs
igor-aptos: we always go to u128, and cast to appropriate type when return to move, for code simplicity.

And all this is identical on aggregator_v2 branch - https://github.com/aptos-labs/aptos-core/blob/aggregators_v2/aptos-move/aptos-aggregator/src/types.rs#L322

vusirikala (approved):

aptos-move/aptos-gas-schedule/src/gas_schedule/aptos_framework.rs
```diff
@@ -201,11 +201,17 @@ crate::gas_schedule::macros::define_gas_parameters!(
         [aggregator_destroy_base: InternalGas, "aggregator.destroy.base", 10000],
         [aggregator_factory_new_aggregator_base: InternalGas, "aggregator_factory.new_aggregator.base", 10000],
 
+        [aggregator_v2_snapshot_base: InternalGas, {11.. => "aggregator_v2.snapshot.base"}, 6000],
```
vgao1996, line 204: Just a reminder to fix the gas feature version here

aptos-move/framework/aptos-framework/sources/aggregator_v2/aggregator_v2.move
```diff
@@ -1,26 +1,118 @@
...
+    /// Returns `max_value` exceeding which aggregator overflows.
+    public fun max_value<IntElement: copy + drop>(aggregator: &Aggregator<IntElement>): IntElement {
+        aggregator.max_value
+    }
+
+    /// Creates new aggregator, with given 'max_value'.
+    ///
+    /// Currently supported types for Element are u64 and u128.
+    /// EAGGREGATOR_ELEMENT_TYPE_NOT_SUPPORTED raised if called with a different type.
+    public native fun create_aggregator<IntElement: copy + drop>(max_value: IntElement): Aggregator<IntElement>;
```
zekun000, line 63: if we only support u64 and u128, I think it's better to write constructors explicitly and leave other functions generic
```
  fun create_u64_aggregator(max_value: u64): Aggregator<u64>;
```

igor-aptos: why? 

usage here seems just fine - https://github.com/aptos-labs/aptos-core/pull/9972/files#diff-7026c328622536165825a6f5dcaf13764bd475489ecba30ba4716dfdc4acc1e9

and set of types we support is something we can expand on in the future.

aptos-move/framework/src/natives/aggregator_natives/aggregator_v2.rs
```diff
@@ -175,41 +511,50 @@ fn native_string_concat(
...
-    result.extend(&snapshot_value);
-    result.extend(&after);
-    let move_string_value = Value::struct_(Struct::pack(vec![Value::vector_u8(result)]));
-    let move_snapshot_value = Value::struct_(Struct::pack(vec![move_string_value]));
-    Ok(smallvec![move_snapshot_value])
+    };
+
+    let prefix = string_to_bytes(safely_pop_arg!(args, Struct))?;
+
+    if prefix.len() + suffix.len() > CONCAT_PREFIX_AND_SUFFIX_MAX_LENGTH {
```
georgemitenkov, line 543: I think I asked on some other PR before, this sum is always < max u64

georgemitenkov (approved):

aptos-move/framework/aptos-framework/sources/aggregator_v2/aggregator_v2.move
```diff
@@ -1,97 +1,214 @@
...
+/// operation that also reduced parallelism, and should be avoided as much as possible.
 module aptos_framework::aggregator_v2 {
+    use std::error;
     use std::string::String;
 
+    /// The value of aggregator overflows. Raised by uncoditional add() call
+    const EAGGREGATOR_OVERFLOW: u64 = 1;
+
+    /// The value of aggregator underflows (goes below zero). Raised by uncoditional sub() call
+    const EAGGREGATOR_UNDERFLOW: u64 = 2;
```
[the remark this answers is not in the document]
igor-aptos: I think that was from v1 aggregators, to disambiguate.

though I see only 3 there now:

    const ENOT_SUPPORTED: u64 = 3;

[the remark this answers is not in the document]
igor-aptos: because rest are consistent:


    /// The value of aggregator overflows. Raised by native code.
    const EAGGREGATOR_OVERFLOW: u64 = 1;

    /// The value of aggregator underflows (goes below zero). Raised by native code.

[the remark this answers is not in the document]
georgemitenkov: ENOT_SUPPORTED is never used I think

PR #10398, opened 2023-10-04
[Sharded-Execution] Integrate remote sharded executor with executor benchmark

manudhundi:
### Description
With this we can run the executor benchmark on the co-ordinator node and shards
individually on other nodes.

First we should run the shard service on node(s) by a cmd like below:
cargo run --profile performance -p aptos-executor-service --manifest-path ./execution/executor-service/Cargo.toml -- --shard-id 0 --num-shards 2 --coordinator-address 10.138.0.6:52200 --remote-executor-addresses 10.138.0.7:52201 10.138.0.8:52202 --num-executor-threads 48

Then we can run the Executor benchmark cmd as:
cargo run --profile performance -p aptos-executor-benchmark -- --block-size 50000 --enable-storage-sharding --num-executor-shards 2 --connected-tx-grps 50000 --remote-sharding --remote-executor-addresses 10.138.0.7:52201 10.138.0.8:52202 --coordinator-address 10.138.0.6:52200 --generate-then-execute run-executor --data-dir ~/workspace/db_dirs/db2/ --checkpoint-dir ~/workspace/db_dirs/chk --blocks 25 --main-signer-accounts 100001


aptos-move/aptos-vm/src/sharded_block_executor/sharded_executor_service.rs
```diff
@@ -23,8 +24,29 @@ use aptos_types::{
 use aptos_vm_logging::disable_speculative_logging;
 use futures::{channel::oneshot, executor::block_on};
 use move_core_types::vm_status::VMStatus;
+use once_cell::sync::Lazy;
 use std::sync::Arc;
 
+pub static APTOS_SHARDED_EXECUTOR_EXECUTE_BLOCK_SECONDS: Lazy<Histogram> = Lazy::new(|| {
```
msmouse, line 30: nit: it's sorta our convention to host these counters in a separate file (counters.rs or metrics.rs)
and nit: shall we use `register_histogram_vec` to make our lifes easier in most cases? (both when coding and on the dashboards)
for example https://github.com/aptos-labs/aptos-core/blob/3ba06ec9f1aab0206f7e0d822e3531b87a218fe9/experimental/execution/ptx-executor/src/metrics.rs#L9-L33

execution/executor-service/src/remote_state_view.rs
```diff
@@ -114,25 +123,61 @@ impl RemoteStateViewClient {
...
             });
     }
 
+    fn pre_fetch_state_values(&self, state_keys: Vec<StateKey>, sync_fetch: bool) {
+        let state_view_clone = self.state_view.clone();
+        let thread_pool_clone = self.thread_pool.clone();
+        let kv_tx_clone = self.kv_tx.clone();
+        let shard_id = self.shard_id;
+
+        if sync_fetch {
```
msmouse, line 160: nit: 
```Rust
let fetch = move ||  Self::insert_keys_and_fetch_values(
                state_view_clone,
                thread_pool_clone,
                kv_tx_clone,
                shard_id,
                state_keys,
            );
if sync_fetch {
	fetch()
} else {
    self.thread_pool_spawn(fetch)
}
```

execution/executor-service/src/remote_state_view.rs
```diff
@@ -142,6 +187,27 @@ impl RemoteStateViewClient {
         let request_message = bcs::to_bytes(&request).unwrap();
         sender.send(Message::new(request_message)).unwrap();
     }
+
+    pub fn print_info(&self) {
+        info!(
+            "REMOTE_EXECUTOR: Total approx get prefetched val time is {} s",
```
msmouse, line 193: nit: I think it's desired to do 
```Rust
info!(
    prefetch_value_time = TIMER1.get_sample_sum(),
    kv_response_time = TIME2.get_sample_sum(),
    ...
    "whatever event happened.",
)
```

execution/executor/src/components/chunk_output.rs
```diff
@@ -187,12 +250,21 @@ impl ChunkOutput {
...
+                maybe_block_gas_limit,
+            )?)
+        } else {
+            Ok(V::execute_block_sharded(
+                SHARDED_BLOCK_EXECUTOR.lock().deref(),
+                partitioned_txns,
+                state_view,
+                maybe_block_gas_limit,
+            )?)
+        }
```
msmouse, line 267: can we move this if-else to `aptos_vm::execute_block_sharded`? 
And probably the `Lazy`'s closer to `ShardedBlockexecutor`?

execution/executor-service/src/remote_state_view.rs
grao1991, line 160: The name `sync_fetch` looks weird here. You are not waiting for the response here anyway.

execution/executor-service/src/main.rs
```diff
@@ -1,26 +1,52 @@
...
 
 #[derive(Debug, Parser)]
 struct Args {
-    #[clap(long, default_value_t = 8080)]
-    pub server_port: u16,
-
     #[clap(long, default_value_t = 8)]
     pub num_executor_threads: usize,
+
+    #[clap(long)]
```
grao1991, line 14: need documentation

aptos-move/aptos-vm/src/sharded_block_executor/sharded_executor_service.rs
grao1991: +1 to use single metric, unless you need a different bucket

aptos-move/aptos-vm/src/sharded_block_executor/aggr_overridden_state_view.rs
```diff
@@ -39,6 +39,7 @@ impl<'a, S: StateView + Sync + Send> TStateView for AggregatorOverriddenStateVie
 
     fn get_state_value(&self, state_key: &StateKey) -> Result<Option<StateValue>> {
         if *state_key == *TOTAL_SUPPLY_STATE_KEY {
+            self.base_view.get_state_value(state_key)?;
```
sitalkedia, line 42: Let's add some comment here explaining why its needed.

aptos-move/aptos-vm/src/sharded_block_executor/sharded_executor_service.rs
```diff
@@ -214,20 +236,29 @@ impl<S: StateView + Sync + Send + 'static> ShardedExecutorService<S> {
...
+                    let execute_block_timer =
+                        APTOS_SHARDED_EXECUTOR_EXECUTE_BLOCK_SECONDS.start_timer();
                     let ret = self.execute_block(
                         transactions,
                         state_view.as_ref(),
                         concurrency_level_per_shard,
                         maybe_block_gas_limit,
                     );
                     drop(state_view);
+                    execute_block_timer.stop_and_record();
```
sitalkedia, line 248: nit: prefer dropping the timer here to be consistent with how we do it in rest of the code.

aptos-move/aptos-vm/src/sharded_block_executor/sharded_executor_service.rs
```diff
@@ -214,20 +236,29 @@ impl<S: StateView + Sync + Send + 'static> ShardedExecutorService<S> {
...
+                    let _timer = APTOS_SHARDED_EXECUTOR_RESULT_TX_SECONDS.start_timer();
                     self.coordinator_client.send_execution_result(ret);
                 },
                 ExecutorShardCommand::Stop => {
                     break;
                 },
             }
         }
+        info!(
```
sitalkedia, line 257: These info logs are going to be spammy and also not useful given we already have the counters?

execution/executor-benchmark/src/block_partitioning.rs
```diff
@@ -17,10 +17,16 @@ pub(crate) struct BlockPartitioningStage {
 }
 
 impl BlockPartitioningStage {
-    pub fn new(num_shards: usize, partitioner_config: &dyn PartitionerConfig) -> Self {
-        let maybe_partitioner = if num_shards <= 1 {
+    pub fn new(
+        num_shards: usize,
+        partitioner_config: &dyn PartitionerConfig,
+        force_partitioning: bool,
```
sitalkedia, line 23: Hmm not sure why we need this, can we just use the partitioner when sharding is enabled and it is enabled when we set the number of shards (even if it is set to 1 ?) and set the default value to 0 (which means off)

execution/executor-benchmark/src/main.rs
```diff
@@ -107,6 +108,12 @@ pub struct PipelineOpt {
     #[clap(long, default_value = "1")]
     num_executor_shards: usize,
     #[clap(long)]
+    remote_sharding: bool,
```
sitalkedia, line 111: Can you take a stab at creating a new ShardingOpt struct and move them there? Also, we don't need a separate `remote_sharding` flag, it should be enabled when `remote_executor_addresses` is set ?

execution/executor-service/src/remote_cordinator_client.rs
```diff
@@ -93,7 +106,22 @@ impl CoordinatorClient<RemoteStateViewClient> for RemoteCoordinatorClient {
...
+                    APTOS_REMOTE_EXECUTOR_CMD_RX_SECONDS.get_sample_sum()
+                );
+                info!(
+                    "REMOTE_EXECUTOR: cmd rx bcs deser time is {} s",
+                    APTOS_REMOTE_EXECUTOR_CMD_RX_BCS_DESERIALIZE_SECONDS.get_sample_sum()
+                );
+                info!(
+                    "REMOTE_EXECUTOR: init prefetch time is {} s",
+                    APTOS_REMOTE_EXECUTOR_INIT_PREFETCH_SECONDS.get_sample_sum()
+                );
```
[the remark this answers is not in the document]
manudhundi: Yeah we dont need 3 lines. Updated

execution/executor-service/src/main.rs
```diff
@@ -1,26 +1,52 @@
...
-    let _args = Args::parse();
+    let args = Args::parse();
     aptos_logger::Logger::new().init();
 
-    // let server_addr = SocketAddr::new(IpAddr::V4(Ipv4Addr::LOCALHOST), args.server_port);
-    // let executor_service =
-    //     ProcessExecutorService::new(server_addr, 1000, args.num_executor_threads);
-    // executor_service.run();
+    let (tx, rx) = crossbeam_channel::unbounded();
+    ctrlc::set_handler(move || {
```
sitalkedia, line 33: why do you need this? I am wondering if we can do some of this clean up when this is dropped..

execution/executor-service/src/remote_cordinator_client.rs
```diff
@@ -78,11 +83,19 @@ impl CoordinatorClient<RemoteStateViewClient> for RemoteCoordinatorClient {
...
             Ok(message) => {
+                let _rx_timer = APTOS_REMOTE_EXECUTOR_CMD_RX_SECONDS.start_timer();
+                let bcs_deser_timer =
+                    APTOS_REMOTE_EXECUTOR_CMD_RX_BCS_DESERIALIZE_SECONDS.start_timer();
                 let request: RemoteExecutionRequest = bcs::from_bytes(&message.data).unwrap();
+                bcs_deser_timer.stop_and_record();
+
                 match request {
                     RemoteExecutionRequest::ExecuteBlock(command) => {
+                        let init_prefetch_timer =
```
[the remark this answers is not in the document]
manudhundi: elegant indeed

execution/executor-service/src/metrics.rs
```diff
@@ -0,0 +1,107 @@
...
+    register_histogram!(
+        // metric name
+        "aptos_remote_executor_init_prefetch_seconds",
+        // metric description
+        "The time spent in seconds on initializing the prefetching of remote state values on a shard",
+        exponential_buckets(/*start=*/ 1e-3, /*factor=*/ 2.0, /*count=*/ 20).unwrap(),
+    ).unwrap()
+});
+
+pub static APTOS_REMOTE_EXECUTOR_REMOTE_KV_RESPONSES_PROCESSING_TIME_SECONDS: Lazy<Histogram> =
```
sitalkedia, line 42: `APTOS_REMOTE_EXECUTOR_REMOTE_KV_RESPONSES_PROCESSING_TIME_SECONDS` -> `REMOTE_EXECUTOR_KV_RESPONSES_SECONDS`

execution/executor-service/src/metrics.rs
```diff
@@ -0,0 +1,107 @@
+// Copyright © Aptos Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+use aptos_metrics_core::{
+    exponential_buckets, register_histogram, register_int_counter, Histogram, IntCounter,
+};
+use once_cell::sync::Lazy;
+
+pub static APTOS_REMOTE_EXECUTOR_CMD_RX_SECONDS: Lazy<Histogram> = Lazy::new(|| {
```
sitalkedia, line 9: `APTOS` prefix is redunant in the counter name - only relevant in the metrics name..

execution/executor-service/src/metrics.rs
```diff
@@ -0,0 +1,107 @@
...
+            // metric name
+            "aptos_remote_executor_remote_kv_response_processing_time_seconds",
+            // metric description
+            "The time spent in seconds on processing the remote key value responses on a shard",
+            exponential_buckets(/*start=*/ 1e-3, /*factor=*/ 2.0, /*count=*/ 20).unwrap(),
+        )
+        .unwrap()
+    });
+
+pub static APTOS_REMOTE_EXECUTOR_REMOTE_KV_RESPONSES_DESER_TIME_SECONDS: Lazy<Histogram> =
```
sitalkedia, line 54: `APTOS_REMOTE_EXECUTOR_REMOTE_KV_RESPONSES_DESER_TIME_SECONDS` -> `REMOTE_EXECUTOR_KV_DESER_SECONDS`

execution/executor-service/src/remote_cordinator_client.rs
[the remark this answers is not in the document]
sitalkedia: +1, please remove these log lines..

execution/executor-service/src/remote_state_view.rs
```diff
@@ -114,25 +123,61 @@ impl RemoteStateViewClient {
 
     pub fn init_for_block(&self, state_keys: Vec<StateKey>) {
         *self.state_view.write().unwrap() = RemoteStateView::new();
-        self.pre_fetch_state_values(state_keys);
+        info!(
```
sitalkedia, line 126: nit: remove this log line as this is going to be spammy

execution/executor-service/src/remote_state_view.rs
```diff
@@ -114,25 +123,61 @@ impl RemoteStateViewClient {
...
         });
         state_keys
             .chunks(REMOTE_STATE_KEY_BATCH_SIZE)
             .map(|state_keys_chunk| state_keys_chunk.to_vec())
             .for_each(|state_keys| {
-                let sender = self.kv_tx.clone();
-                let shard_id = self.shard_id;
-                self.thread_pool.spawn(move || {
+                let sender = kv_tx.clone();
+                thread_pool.clone().spawn(move || {
```
sitalkedia, line 148: You probably don't need to clone the thread pool here?

execution/executor-service/src/remote_state_view.rs
```diff
@@ -114,25 +123,61 @@ impl RemoteStateViewClient {
...
                     Self::send_state_value_request(shard_id, sender, state_keys);
                 });
             });
     }
 
+    fn pre_fetch_state_values(&self, state_keys: Vec<StateKey>, sync_fetch: bool) {
```
sitalkedia, line 154: Please add a comment explaining why `sync_fetch` is needed.

execution/executor-service/src/remote_state_view.rs
sitalkedia: We don't really need these log lines.. let's remove them and build a dashboards for these counters instead :)

execution/executor/src/components/chunk_output.rs
```diff
@@ -31,16 +33,78 @@ use aptos_vm::{
...
+    sync::Arc,
+    time::Duration,
+};
+
+static REMOTE_SHARDING: OnceCell<bool> = OnceCell::new();
+static REMOTE_ADDRESSES: OnceCell<Vec<SocketAddr>> = OnceCell::new();
+static COORDINATOR_ADDRESS: OnceCell<SocketAddr> = OnceCell::new();
+
+/// Sets the flag when invoked the first time.
+pub fn set_remote_sharding(enable: bool) {
```
sitalkedia, line 49: This flag is not needed, just set remote sharding to enabled when `remote_addresses` is set.

execution/executor/src/components/chunk_output.rs
```diff
@@ -31,16 +33,78 @@ use aptos_vm::{
...
+    match COORDINATOR_ADDRESS.get() {
+        Some(value) => *value,
+        None => SocketAddr::new(IpAddr::V4(Ipv4Addr::LOCALHOST), COORDINATOR_PORT),
+    }
+}
 
 pub static SHARDED_BLOCK_EXECUTOR: Lazy<
     Arc<Mutex<ShardedBlockExecutor<CachedStateView, LocalExecutorClient<CachedStateView>>>>,
 > = Lazy::new(|| {
+    info!("LOCAL_SHARDED_BLOCK_EXECUTOR obj created");
```
sitalkedia, line 85: nit: `info!("LOCAL_SHARDED_BLOCK_EXECUTOR created");`

execution/executor/src/components/chunk_output.rs
manudhundi: It may not be best to move it there because
(1) execute_block_sharded is a trait. Today it is not implemented by other executors, but I in future we may want to implement them.
(2) Also the first param is a Generic of type &ShardedBlockExecutor<S, E> where E: ExecutorClient<S>, and that E can be either LocalExecutorClient<S> or RemoteExecutorClient<S>.

I think it is better if executors like AptosVM::execute_block_sharded or PtxBlockExecutor::execute_block_sharded don't know about creating the ShardedBlockExecutor obj.

execution/executor-service/src/main.rs
```diff
@@ -1,26 +1,52 @@
...
+
+    #[clap(long, num_args = 1..)]
+    pub remote_executor_addresses: Vec<SocketAddr>,
+
+    #[clap(long)]
+    pub coordinator_address: SocketAddr,
 }
 
 fn main() {
     // TODO (skedia): Uncomment this once the executor service is implemented.
```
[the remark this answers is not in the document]
manudhundi: Yes done

execution/executor-service/src/main.rs
manudhundi: For now this is the main func() that starts the service. How else can we drop this ?

sitalkedia: Hmm can you move the shutdown function call to `drop` method of `ProcessExecutorService` ?

manudhundi: Ok done

execution/executor-service/src/remote_state_view.rs
manudhundi: ok

aptos-move/aptos-vm/src/sharded_block_executor/sharded_executor_service.rs
manudhundi: ok done

execution/executor-benchmark/src/block_partitioning.rs
manudhundi: Yeah, we can do this by setting default to 0 and use that for unsharded. Just wanted to keep the delta in the cmd usage minimal. But I guess nobody specifies num_shards for unsharded execution, so change should be ok.

execution/executor/src/components/chunk_output.rs
sitalkedia: @manudhundi - I think @msmouse 's point is moving the logic of creating a sharded block executor to the `ShardedBlockexecutor` module - we can do it as an helper function (let's say `create_shared_block_executor`). Currently it is scattered in many places which is not very clean.

manudhundi: I am fine with creating a helper function, but it will not do much. The Lazy business is specifically for the benchmark logic. So it will just another helper function in the same file.

execution/executor-service/src/remote_state_view.rs
manudhundi: Done

execution/executor-service/src/remote_state_view.rs
manudhundi: Ok. changed it to 'sync_insert_keys'

aptos-move/aptos-vm/src/sharded_block_executor/sharded_executor_service.rs
manudhundi: Removed

aptos-move/aptos-vm/src/sharded_block_executor/sharded_executor_service.rs
manudhundi: @msmouse we are already using quite a few stats with register_histogram, say like APTOS_EXECUTOR_VM_EXECUTE_BLOCK_SECONDS.

Do you still prefer using `register_histogram_vec` only ?

> pub static APTOS_EXECUTOR_VM_EXECUTE_BLOCK_SECONDS: Lazy<Histogram> = Lazy::new(|| {
    register_histogram!(
        // metric name
        "aptos_executor_vm_execute_block_seconds",
        // metric description
        "The time spent in seconds of vm block execution in Aptos executor",
        exponential_buckets(/*start=*/ 1e-3, /*factor=*/ 2.0, /*count=*/ 20).unwrap(),
    )
    .unwrap()
});

msmouse: yeah.. I still think it's nice to just use https://github.com/aptos-labs/aptos-core/blob/7329c16e7d336f5f1708310d7c09d237429d8bfb/aptos-move/aptos-vm/src/counters.rs#L142
or add a `SHARDED_EXECUTOR_TIMER` to group things up

manudhundi:
Addressed most of the comments. A couple of them pending. Committed this to run some tests on GCP

execution/executor-benchmark/src/main.rs
manudhundi: Done

execution/executor-service/src/remote_state_view.rs
```diff
@@ -114,25 +123,61 @@ impl RemoteStateViewClient {
...
+            "REMOTE_EXECUTOR: Prefetching {} state keys",
+            state_keys.len()
+        );
+        self.pre_fetch_state_values(state_keys, false);
     }
 
-    fn pre_fetch_state_values(&self, state_keys: Vec<StateKey>) {
+    fn insert_keys_and_fetch_values(
+        state_view_clone: Arc<RwLock<RemoteStateView>>,
+        thread_pool: Arc<ThreadPool>,
```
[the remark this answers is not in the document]
manudhundi: Doesn't allow me to take reference because I am spawning a new thread later. Below is the error:

> 166 |     fn pre_fetch_state_values(&self, state_keys: Vec<StateKey>, sync_insert_keys: bool) {
    |                               -----
    |                               |
    |                               `self` is a reference that is only valid in the method body
    |                               let's call the lifetime of this reference `'1`
...
189 |             self.thread_pool.spawn(insert_and_fetch);
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |             |
    |             `self` escapes the method body here
    |             argument requires that `'1` must outlive `'static`

execution/executor/src/components/chunk_output.rs
manudhundi: As discussed moved some of the ShardedBlockExecutor create logic out of this file

execution/executor-service/src/remote_cordinator_client.rs
```diff
@@ -78,11 +80,26 @@ impl CoordinatorClient<RemoteStateViewClient> for RemoteCoordinatorClient {
     fn receive_execute_command(&self) -> ExecutorShardCommand<RemoteStateViewClient> {
         match self.command_rx.recv() {
             Ok(message) => {
-                let request: RemoteExecutionRequest = bcs::from_bytes(&message.data).unwrap();
+                let _rx_timer = REMOTE_EXECUTOR_TIMER
+                    .with_label_values(&[&self.shard_id.to_string(), "cmd_rx"])
+                    .start_timer();
+                let request: RemoteExecutionRequest;
+                {
+                    let _bcs_deser_timer = REMOTE_EXECUTOR_TIMER
```
sitalkedia, line 88: Hmm I don't like this style of dropping the timer - it is confusing... may be less confusing to remove the implicit scope and drop the timer instead..?

sitalkedia (approved):

execution/executor-service/src/remote_cordinator_client.rs
manudhundi: You mean, I should use stop_and_discard() ?

sitalkedia: I meant to do something like this ..

```
let bcs_deser_timer = REMOTE_EXECUTOR_TIMER.with_label_values(&[&self.shard_id.to_string(), "cmd_rx_bcs_deser"])
                        .start_timer();
let request = bcs::from_bytes(&message.data).unwrap();
drop(bcs_deser_timer)
```

sitalkedia: Here and in other places as well..

execution/executor/src/components/chunk_output.rs
msmouse: Let's say we keep the if-else here. Moving all the `Lazy`s and relevant functions over to `ShardedBlockExecutor`'s module should be doable? 
We should be able to `use` only `REMOTE_SHARDED_BLOCK_EXECUTOR` and `SHARDED_BLOCK_EXECUTOR` here relatively easily?

I might also suggest remove most of the generics and just use `Arc<dyn>`s but that might sound too jerky here, so consider not having seen this. :trollface:

manudhundi: Ok moved all the Lazy business out of chunk_output.rs

Will take the liberty to skip the Arc<dyn> because
(i) It involves little larger scope of changing the trait VMExecutor::execute_block_sharded (and associated implementations)
(ii) I somewhat believe we should use Generics instead of Dyn wherever we can because it is faster.
Hope it is ok

msmouse (approved):

PR #10400, opened 2023-10-04
[CLI][minor] Recommend brew install aptos instead of brew upgrade

bowenyang007:
### Description
Tried to follow brew upgrade aptos and ran into issues. Brew install works better
image


banool:
Does brew install actually update too though?

PR #10402, opened 2023-10-04
split AIP to be different proposal

sherry-x:
### Description

Each AIP should it their own proposal

### DIFF
--- a/aptos-move/aptos-release-builder/data/release.yaml
+++ b/aptos-move/aptos-release-builder/data/release.yaml
@@ -23,23 +23,28 @@ proposals:
           enabled:
             - safer_resource_groups
             - safer_metadata
+            - signature_checker_v2_script_fix
       - Framework:
           bytecode_version: 6
           git_hash: ~
-  - name: step_3_enable_signature_checker_v2_fix
+      - Execution:
+          V3:
+            block_gas_limit: ~
+            transaction_shuffler_type: no_shuffling
+            transaction_deduper_type: txn_hash_and_authenticator_v1
+  - name: step_3_enable_block_gas_limit
     metadata:
-      title: "Enable a fix signature checker v2"
-      description: "Fixing a counting bug"
+      title: "Enable Block Gas Limit"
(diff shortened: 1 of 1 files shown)


rahxephon89 (approved):

aptos-move/aptos-release-builder/data/release.yaml
```diff
@@ -23,23 +23,23 @@ proposals:
...
+      title: "Enable Block Gas Limit"
+      description: "Enable Block Gas Limit"
     execution_mode: MultiStep
     update_sequence:
-      - FeatureFlag:
-          enabled:
-            - signature_checker_v2_script_fix
-  - name: step_4_enable_block_gas_limit_and_transaction_shuffling
+      - Execution:
+          V3:
```
bchocho, line 37:
```
V3:
  block_gas_limit: 35000
  transaction_shuffler_type: no_shuffling
  transaction_deduper_type: txn_hash_and_authenticator_v1
```

aptos-move/aptos-release-builder/data/release.yaml
```diff
@@ -48,8 +48,6 @@ proposals:
           V3:
```
bchocho, line 48:
```
V3:
  block_gas_limit: 35000
  transaction_shuffler_type:
    sender_aware_v2: 32
  transaction_deduper_type: txn_hash_and_authenticator_v1
```

bchocho (commented):
We'll have to explicitly set the fields in V3. I think it needs to look like this (but please check the yaml syntax @sitalkedia )

aptos-move/aptos-release-builder/data/release.yaml
sherry-x: but in this way, if the first step got rejected, this one would enable the other AIP too....

bchocho: Yeah unfortunately the way the code is written I don't think there's a way to avoid this

PR #10408, opened 2023-10-05
[Fix] Do not assume we have VersionData at every version.

grao1991:
### Description
We will only have it at the end of chunk/block.


bchocho (approved):
Approved to unblock

sitalkedia (approved):

PR #10409, opened 2023-10-05
[dag] abstract round state to support different behaviors

zekun000:

### DIFF
--- a/consensus/src/dag/dag_driver.rs
+++ b/consensus/src/dag/dag_driver.rs
@@ -19,6 +19,7 @@ use crate::{
             logging::{LogEvent, LogSchema},
             tracing::{observe_node, observe_round, NodeStage, RoundStage},
         },
+        round_state::RoundState,
         types::{CertificateAckState, CertifiedNode, Node, SignatureBuilder},
     },
     payload_manager::PayloadManager,
@@ -54,13 +55,13 @@ pub(crate) struct DagDriver {
     payload_manager: Arc<PayloadManager>,
     payload_client: Arc<dyn PayloadClient>,
     reliable_broadcast: Arc<ReliableBroadcast<DAGMessage, ExponentialBackoff>>,
-    current_round: Round,
     time_service: TimeService,
     rb_abort_handle: Option<(AbortHandle, u64)>,
     storage: Arc<dyn DAGStorage>,
     order_rule: OrderRule,
     fetch_requester: Arc<FetchRequester>,
     ledger_info_provider: Arc<dyn TLedgerInfoProvider>,
+    round_state: RoundState,
 }
 
 impl DagDriver {
@@ -76,15 +77,13 @@ impl DagDriver {
         order_rule: OrderRule,
         fetch_requester: Arc<FetchRequester>,
         ledger_info_provider: Arc<dyn TLedgerInfoProvider>,
+        round_state: RoundState,
     ) -> Self {
         let pending_node = storage
             .get_pending_node()
             .expect("should be able to read dag storage");
-        let highest_round = dag.read().highest_round();
-        let highest_strong_links_round = dag
-            .read()
-            .get_strong_links_for_round(highest_round, &epoch_state.verifier)
-            .map_or_else(|| highest_round.saturating_sub(1), |_| highest_round);
+        let highest_strong_links_round =
+            dag.read().highest_strong_links_round(&epoch_state.verifier);
 
         let mut driver = Self {
             author,
@@ -93,13 +92,13 @@ impl DagDriver {
             payload_manager,
             payload_client,
             reliable_broadcast,
-            current_round: highest_strong_links_round,
             time_service,
             rb_abort_handle: None,
             storage,
             order_rule,
             fetch_requester,
             ledger_info_provider,
+            round_state,
         };
 
         // If we were broadcasting the node for the round already, resume it
@@ -110,7 +109,7 @@ impl DagDriver {
                 LogSchema::new(LogEvent::NewRound).round(node.round()),
                 "Resume round"
             );
-            driver.current_round = node.round();
+            driver.round_state.set_current_round(node.round());
             driver.broadcast_node(node);
         } else {
             // kick start a new round
@@ -119,8 +118,8 @@ impl DagDriver {
         driver
     }
 
-    pub async fn add_node(&mut self, node: CertifiedNode) -> anyhow::Result<()> {
-        let highest_strong_links_round = {
+    async fn add_node(&mut self, node: CertifiedNode) -> anyhow::Result<()> {
+        let (highest_strong_link_round, strong_links) = {
             let mut dag_writer = self.dag.write();
 
             if !dag_writer.all_exists(node.parents_metadata()) {
@@ -134,20 +133,31 @@ impl DagDriver {
                 .prefetch_payload_data(node.payload(), node.metadata().timestamp());
             dag_writer.add_node(node)?;
 
-            let highest_round = dag_writer.highest_round();
-            dag_writer
-                .get_strong_links_for_round(highest_round, &self.epoch_state.verifier)
-                .map_or_else(|| highest_round.saturating_sub(1), |_| highest_round)
+            let highest_strong_links_round =
+                dag_writer.highest_strong_links_round(&self.epoch_state.verifier);
+            (
+                highest_strong_links_round,
+                // unwrap is for round 0
+                dag_writer
+                    .get_strong_links_for_round(
+                        highest_strong_links_round,
+                        &self.epoch_state.verifier,
+                    )
+                    .unwrap_or(vec![]),
+            )
         };
-
-        if self.current_round <= highest_strong_links_round {
-            self.enter_new_round(highest_strong_links_round + 1).await;
-        }
+        self.round_state
+            .check_for_new_round(highest_strong_link_round, strong_links)
+            .await;
         Ok(())
     }
 
     pub async fn enter_new_round(&mut self, new_round: Round) {
+        if self.round_state.current_round() >= new_round {
+            return;
+        }
         debug!(LogSchema::new(LogEvent::NewRound).round(new_round));
+        self.round_state.set_current_round(new_round);
         counters::CURRENT_ROUND.set(new_round as i64);
         let strong_links = self
             .dag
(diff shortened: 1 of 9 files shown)


consensus/src/dag/dag_driver.rs
```diff
@@ -47,20 +47,75 @@ pub enum DagDriverError {
...
+        OptimisticResponsiveRound {
+            current_round: 0,
+            event_sender,
+        }
+    }
+}
+
+#[async_trait]
+impl RoundState for OptimisticResponsiveRound {
+    async fn update(
```
ibalajiarun, line 79: nit: no "update" actually happens here? 

this should be `enter_new_round` right? the below `enter_new_round` function should just be a setter like `set_new_round`?

consensus/src/dag/dag_driver.rs
```diff
@@ -47,20 +47,75 @@ pub enum DagDriverError {
...
+        &mut self,
+        highest_round: Round,
+        strong_links_for_highest_round: Option<Vec<NodeCertificate>>,
+    );
+    fn enter_new_round(&mut self, new_round: Round);
+
+    fn current_round(&self) -> Round;
+}
+
+pub struct OptimisticResponsiveRound {
```
ibalajiarun, line 63: This is actually `PrevalentResponsive` round right?

zekun000: I never understand what that word means lol, we're going to use a different one anyway, will update soon

consensus/src/dag/dag_driver.rs
zekun000: I think this is more `check_for_new_round`, it doesn't necessarily enter

consensus/src/dag/round_state.rs
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    fn current_round(&self) -> Round {
+        self.current_round
+    }
+
+    fn set_current_round(&mut self, round: Round) {
+        assert!(self.current_round < round);
+        self.current_round = round;
+    }
```
sasha8, line 153: Anyway to make this default impl?

consensus/src/dag/round_state.rs
```diff
@@ -0,0 +1,154 @@
...
+                if voting_power == self.epoch_state.verifier.total_voting_power()
+                    || duration_since_start >= self.minimal_wait_time
+                {
+                    self.send_new_round(new_round).await;
+                } else {
+                    // wait until minimal time reaches before sending
+                    let sender = self.event_sender.clone();
+                    let wait_time = self.minimal_wait_time - duration_since_start;
+                    tokio::spawn(async move {
+                        tokio::time::sleep(wait_time).await;
```
sasha8, line 133: What if we get more votes in the meantime?

consensus/src/dag/round_state.rs
```diff
@@ -0,0 +1,154 @@
...
+}
+
+/// More sophisticated strategy to move round forward given 2f+1 strong links
+/// Delay if backpressure is triggered. (TODO)
+/// Move as soon as 3f+1 is ready. (TODO: make it configurable)
+/// Move if minimal wait time is reached. (TODO: make it configurable)
+pub struct AdaptiveResponsiveRound {
+    current_round: Round,
+    epoch_state: Arc<EpochState>,
+    start_time: Instant,
```
sasha8, line 76: The plan is to create new one every round?

consensus/src/dag/dag_driver.rs
```diff
@@ -134,20 +133,31 @@ impl DagDriver {
...
+        self.round_state
+            .check_for_new_round(highest_strong_link_round, strong_links)
+            .await;
         Ok(())
     }
 
     pub async fn enter_new_round(&mut self, new_round: Round) {
+        if self.round_state.current_round() >= new_round {
+            return;
+        }
```
sasha8, line 158: Maybe assert!()?

sasha8 (approved):
Approving, although it is not clear to me how the AdaptiveResponsiveRound would be connected. See comments.

consensus/src/dag/round_state.rs
zekun000: I didn't find a way to do this

consensus/src/dag/round_state.rs
zekun000: Votes are collected when processing the actual new round event in dag driver so it'll be included

consensus/src/dag/round_state.rs
zekun000: yeah i realized it after waking up 🤦

consensus/src/dag/dag_driver.rs
zekun000: I was worried about duplicate event or out of order event

zekun000:
@sasha8 it's connected now, take another look

consensus/src/dag/round_state.rs
```diff
@@ -0,0 +1,180 @@
...
+impl ResponsiveCheck for AdaptiveResponsive {
+    async fn check_for_new_round(
+        &mut self,
+        highest_strong_links_round: Round,
+        strong_links: Vec<NodeCertificate>,
+    ) {
+        let new_round = highest_strong_links_round + 1;
+        if self.event_sent {
+            return;
+        }
```
sasha8, line 137: nit: move the if up.

consensus/src/dag/round_state.rs
```diff
@@ -0,0 +1,180 @@
...
+            },
+            Ordering::Greater => (),
+        }
+    }
+
+    pub fn current_round(&self) -> Round {
+        self.current_round
+    }
+
+    pub fn set_current_round(&mut self, new_round: Round) {
```
sasha8, line 56: Maybe "reset_current_round"?

consensus/src/dag/round_state.rs
```diff
@@ -0,0 +1,170 @@
...
+        if voting_power == self.epoch_state.verifier.total_voting_power()
+            || duration_since_start >= self.minimal_wait_time
+        {
+            let _ = self.event_sender.send(new_round).await;
+            self.event_sent = true;
+        } else {
+            // wait until minimal time reaches before sending
+            let sender = self.event_sender.clone();
+            let wait_time = self.minimal_wait_time.saturating_sub(duration_since_start);
+            tokio::spawn(async move {
```
sasha8, line 158: This can be called f times, right? Basically, in every add_node after the first 2f+1 stake is added.
If yes, then new_round will be sent f times and we are wasting tasks, aren't we?

sasha8 (approved):

sasha8:
> @sasha8 It's connected now, take another look

Nice! Also no code duplication.

consensus/src/dag/round_state.rs
zekun000: I have event_sent boolean to prevent that, but also check in the enter_new_round to deduplicate

sasha8: I see. But now I am back to my previous question here. What if I start a timer but before the timer expires I get all the votes? Then I still have to wait for the timer, right?

zekun000: you see the if branch above right? 🤦

sasha8: Because event_sent is set to true immediately, it prevent  the new added nodes from sending the new_round event.

zekun000: oh I see your point, maybe we should just deal with duplicate events?

sasha8: Maybe I am missing something, but I think it basically means we will always wait for the timeout.

zekun000: I think duplicate events are fine, we can do something fancier like cancelling inflight tasks but probably not worth it

sasha8: Probably. If it is not a performance penalty to spawn all these tasks.

sasha8: How about selecting between two futures? One is a timer and one returns when the voting in the current_round is satisfied. But maybe ugly because it will have to get the DAG.

zekun000: Updated, I think now it's reasonably good

msmouse (approved):
I'm blind.

consensus/src/dag/round_state.rs
```diff
@@ -0,0 +1,179 @@
...
+            .sum_voting_power(strong_links.iter().map(|cert| cert.metadata().author()))
+            .expect("Unable to sum voting power from strong links");
+        // voting power == 3f+1 or pass minimal wait time
+        let duration_since_start = duration_since_epoch().saturating_sub(self.start_time);
+        if voting_power == self.epoch_state.verifier.total_voting_power()
+            || duration_since_start >= self.minimal_wait_time
+        {
+            let _ = self.event_sender.send(new_round).await;
+            if let State::Scheduled(handle) = std::mem::replace(&mut self.state, State::Sent) {
+                handle.abort();
```
ibalajiarun, line 161: nit: abort first and then send? 

dag_driver handles duplicates, so its fine i think.

consensus/src/dag/dag_store.rs
```diff
@@ -323,7 +331,7 @@ impl Dag {
 
         let bitmask = self
             .nodes_by_round
-            .range(lowest_round..=target_round)
+            .range(lowest_round..target_round)
```
ibalajiarun, line 334: what did this break?

ibalajiarun (approved):

consensus/src/dag/dag_store.rs
zekun000: I added a verify function for the length and it rejects everything now 🤦

consensus/src/dag/round_state.rs
sasha8: nit(I guess): the spawn does not set the state to Sent. This means that the new round event will sent twice. Once from the task and once from the first if which will eventually be satisfied.

zekun000: we abort the task if we send the event directly, in race condition yes we still can receive twice but most cases it should be cancelled properly

sasha8: Yes, I see. When the task sends the event, we will advance the current round and later votes will be simply ignored by the RoundState. So there is only a small period for a possible race. Nice.

PR #10412, opened 2023-10-05
[WIP] comparison testing tool

rahxephon89:
### Description

This PR proposes a comparison testing tool for the new compiler. 
[Design doc](https://www.notion.so/aptoslabs/E2E-comparison-testing-for-the-Move-compiler-25d8897c0f8c453bbebf32301475e52b)

### Test Plan

WIP


aptos-move/aptos-e2e-comparison-testing/src/main.rs
```diff
@@ -0,0 +1,115 @@
...
+        #[clap(long, default_value_t = false)]
+        skip_source_code_check: bool,
+        /// Dump the write set of txns
+        #[clap(long, default_value_t = false)]
+        dump_write_set: bool,
+    },
+    /// Execution of txns
+    Execute {
+        /// Path to the data
+        input_path: Option<PathBuf>,
```
runtian-zhou, line 42: Can you try dump a few txns and have a testsuite for running those txns?

rahxephon89: Added example txn data and execution output

aptos-move/aptos-e2e-comparison-testing/src/data_collection.rs
```diff
@@ -0,0 +1,249 @@
...
+                    let current_dir = self.current_dir.clone();
+                    let dump_write_set = self.dump_write_set;
+                    let data_manager = data_manager.clone();
+                    let index = index_writer.clone();
+
+                    let state_view =
+                        DebuggerStateView::new_with_data_reads(self.debugger.clone(), version);
+
+                    let handler = tokio::task::spawn_blocking(move || {
+                        let epoch_result_res =
```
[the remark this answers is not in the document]
rahxephon89: changed to a more meaningful name

aptos-move/aptos-e2e-comparison-testing/src/data_collection.rs
```diff
@@ -0,0 +1,249 @@
...
+            let batch = if cur_version + self.batch_size <= begin + limit {
+                self.batch_size
+            } else {
+                begin + limit - cur_version
+            };
+            let v = self
+                .debugger
+                .get_and_filter_committed_transactions(cur_version, batch, self.filter_condition)
+                .await
+                .unwrap_or_default();
```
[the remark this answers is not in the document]
rahxephon89: added a log when the call failed

aptos-move/aptos-e2e-comparison-testing/src/data_collection.rs
```diff
@@ -0,0 +1,249 @@
...
+                    let compilation_cache = compilation_cache.clone();
+                    let current_dir = self.current_dir.clone();
+                    let dump_write_set = self.dump_write_set;
+                    let data_manager = data_manager.clone();
+                    let index = index_writer.clone();
+
+                    let state_view =
+                        DebuggerStateView::new_with_data_reads(self.debugger.clone(), version);
+
+                    let handler = tokio::task::spawn_blocking(move || {
```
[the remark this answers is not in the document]
rahxephon89: extract some of the code in the closure into a separate function.

aptos-move/aptos-validator-interface/src/lib.rs
```diff
@@ -46,6 +58,23 @@ pub trait AptosValidatorInterface: Sync {
         limit: u64,
     ) -> Result<(Vec<Transaction>, Vec<TransactionInfo>)>;
 
+    async fn get_and_filter_committed_transactions(
```
runtian-zhou, line 61: As suggested previously, I would recommend using `get_committed_transactions` to implement this method here so that you can get automatic support for db instance as well.

runtian-zhou (commented):
Another question is whether the `get_and_filter` the right API to use for data collection? If we want to deal with a large number of transctions will a rust `Stream` be more suitable?

aptos-move/aptos-e2e-comparison-testing/src/execution.rs
```diff
@@ -0,0 +1,374 @@
...
+        cur_version: Version,
+        state: &FakeDataStore,
+        features: &Features,
+        txn_idx: &TxnIndex,
+        compiled_package_cache: &mut HashMap<PackageInfo, CompiledPackage>,
+        compiled_package_cache_v2: &mut HashMap<PackageInfo, CompiledPackage>,
+    ) {
+        let mut res_1_opt = None;
+        let mut res_2_opt = None;
+        if self.execution_mode.is_v1() || self.execution_mode.is_compare() {
```
runtian-zhou, line 230: Can probably use just a match statement here?

aptos-move/aptos-e2e-comparison-testing/src/execution.rs
```diff
@@ -0,0 +1,374 @@
...
+    ) -> Option<Result<(WriteSet, Vec<ContractEvent>), VMStatus>> {
+        let executor = FakeExecutor::no_genesis();
+        let mut executor = executor.set_not_parallel();
+        *executor.data_store_mut() = state.clone();
+        if let Transaction::UserTransaction(signed_trans) = txn {
+            let sender = signed_trans.sender();
+            let payload = signed_trans.payload();
+            if let TransactionPayload::EntryFunction(entry_function) = payload {
+                // always load 0x1 modules
+                load_aptos_packages_to_executor(&mut executor, compiled_package_cache);
```
runtian-zhou, line 287: Was wondering if this would cause inefficiency issues in the long run if we want to run a huge number of txns. Loading up 0x1 might be a bit slow and if we do that for every txn that might be very inefficient.

aptos-move/aptos-e2e-comparison-testing/src/data_collection.rs
```diff
@@ -0,0 +1,262 @@
...
+
+    fn execute_transactions_at_version_with_state_view(
+        txns: Vec<Transaction>,
+        debugger_stateview: &DebuggerStateView,
+    ) -> Result<Vec<TransactionOutput>> {
+        let sig_verified_txns: Vec<SignatureVerifiedTransaction> =
+            txns.into_iter().map(|x| x.into()).collect::<Vec<_>>();
+        // check whether total supply can be retrieved
+        // used for debugging the aggregator panic issue, will be removed later
+        let val = debugger_stateview.get_state_value(TOTAL_SUPPLY_STATE_KEY.deref());
```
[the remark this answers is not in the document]
rahxephon89: This was added to check whether the issue came from the debugger view (panic here) or the aggregator (panic in the aggregator side) but have not discovered anything yet

aptos-move/aptos-e2e-comparison-testing/src/data_collection.rs
```diff
@@ -0,0 +1,262 @@
...
+    debugger: Arc<dyn AptosValidatorInterface + Send>,
+    current_dir: PathBuf,
+    batch_size: u64,
+    dump_write_set: bool,
+    filter_condition: FilterCondition,
+}
+
+impl DataCollection {
+    pub fn new(
+        debugger: Arc<dyn AptosValidatorInterface + Send>,
```
runtian-zhou, line 42: Same. Was wondering if it would make more sense to take a Stream as an input. Maybe @msmouse can provide some suggestion here.

rahxephon89:
> Another question is whether the `get_and_filter` the right API to use for data collection? If we want to deal with a large number of transctions will a rust `Stream` be more suitable?

Is there a code pointer in aptos-core that I can take a look?

aptos-move/aptos-validator-interface/src/lib.rs
```diff
@@ -117,6 +146,7 @@ pub struct DebuggerStateView {
     query_sender:
         Mutex<UnboundedSender<(StateKey, Version, std::sync::mpsc::Sender<Option<Vec<u8>>>)>>,
     version: Version,
+    pub data_read_stake_keys: Option<Arc<Mutex<HashMap<StateKey, StateValue>>>>,
```
runtian-zhou, line 149: stake or state?

runtian-zhou: Also might be better to keep it private and add a function to dump the hashmap directly.

aptos-move/aptos-e2e-comparison-testing/src/data_collection.rs
```diff
@@ -0,0 +1,262 @@
...
+                    .unwrap()
+                    .write_err(&format!("{}:{}", cur_version, batch));
+            }
+            let txns = res_txns.unwrap_or_default();
+            if !txns.is_empty() {
+                let mut txn_execution_ths = vec![];
+                for (version, txn, source_code_data) in txns {
+                    println!("get txn at version:{}", version);
+
+                    let compilation_cache = compilation_cache.clone();
```
runtian-zhou, line 194: Are the clonings here needed? Or rather, why do we need to move those values around?

aptos-move/aptos-e2e-comparison-testing/src/data_collection.rs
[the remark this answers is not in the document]
gelash: @georgemitenkov

aptos-move/aptos-e2e-comparison-testing/src/data_collection.rs
rahxephon89: It is of Arc type so I have to clone the pointer so that it can be shared among threads. Is there a better way to do it?

aptos-move/aptos-e2e-comparison-testing/src/execution.rs
rahxephon89: done

aptos-move/aptos-validator-interface/src/lib.rs
rahxephon89: done

aptos-move/aptos-e2e-comparison-testing/src/execution.rs
rahxephon89: done

aptos-move/aptos-e2e-comparison-testing/src/execution.rs
rahxephon89: I changed the code to directly store binary blobs in the compiled_package_cache so that serialization only happens once when the source code is compiled.

aptos-move/aptos-validator-interface/src/lib.rs
msmouse: well you can do something like
```rust
  .get_commited_txns()?.zip().enumerate().map(|idx, (txn, txn_info)| {
      let version = begin_version + version;
      (versino, txn, txn_info)
})
```

msmouse (commented):
I see you are force- pushing.. sending out partial review comments for now.

aptos-move/aptos-e2e-comparison-testing/src/data_collection.rs
```diff
@@ -0,0 +1,267 @@
...
+            skip_source_code,
+        ))
+    }
+
+    fn execute_transactions_at_version_with_state_view(
+        txns: Vec<Transaction>,
+        debugger_stateview: &DebuggerStateView,
+    ) -> Result<Vec<TransactionOutput>> {
+        let sig_verified_txns: Vec<SignatureVerifiedTransaction> =
+            txns.into_iter().map(|x| x.into()).collect::<Vec<_>>();
```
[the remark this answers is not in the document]
rahxephon89: Is there an API that does not use SignatureVerifiedTransaction?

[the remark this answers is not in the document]
msmouse: Doesn't looks like so, you probably need to add a helper function inside of your tool's module. 😂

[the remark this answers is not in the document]
rahxephon89: I see. Will refactor this in a separate PR then

aptos-move/aptos-e2e-comparison-testing/src/execution.rs
```diff
@@ -0,0 +1,358 @@
...
+        }
+    }
+
+    fn set_enable(features: &mut Features, flag: FeatureFlag) {
+        let val = flag as u64;
+        let byte_index = (val / 8) as usize;
+        let bit_mask = 1 << (val % 8);
+        if byte_index < features.features.len() {
+            features.features[byte_index] |= bit_mask;
+        }
```
msmouse, line 99: deliberately ignoring features that's beyond the existing bytes? why? 

if not, why not just `features.enable(flag)`?

aptos-move/aptos-e2e-comparison-testing/src/data_collection.rs
```diff
@@ -0,0 +1,267 @@
...
+        // check whether total supply can be retrieved
+        // used for debugging the aggregator panic issue, will be removed later
+        let val = debugger_stateview.get_state_value(TOTAL_SUPPLY_STATE_KEY.deref());
```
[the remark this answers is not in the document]
rahxephon89: done

aptos-move/aptos-e2e-comparison-testing/src/execution.rs
```diff
@@ -0,0 +1,358 @@
...
+    ) -> Result<()> {
+        if !txn_idx.package_info.is_compilable() {
+            return Err(anyhow::Error::msg("not compilable"));
+        }
+        let package_info = txn_idx.package_info.clone();
+        let package_dir = self.input_path.join(format!("{}", package_info));
+        if !package_dir.exists() {
+            return Err(anyhow::Error::msg("source code is not available"));
+        }
+        if (self.execution_mode.is_compare() || self.execution_mode.is_v1())
```
msmouse, line 167: nit: we repeated this a few times, how about adding an `execution_mode.needs_v1()`?

aptos-move/aptos-e2e-comparison-testing/src/execution.rs
rahxephon89: I think `features.enable` was added recently. I will use this.

aptos-move/aptos-e2e-comparison-testing/src/execution.rs
rahxephon89: done

aptos-move/aptos-e2e-comparison-testing/src/execution.rs
rahxephon89: done

msmouse (approved):
Nothing further from me (didn't look at the execution and comparison logics)

rahxephon89:
> Nothing further from me (didn't look at the execution and comparison logics)

Thanks, @msmouse! The execution and comparison logic will be further tested once the V2 is ready.

aptos-move/aptos-e2e-comparison-testing/src/execution.rs
runtian-zhou: I think there are two parts of this problem: one is how the compiler reuse existing compilation result. The other is how the VM will be able to cache the Aptos Framework during execution. The latter one is actually a bit harder to solve but just want to flag out this issue. We can look into this problem together later after I've done my loader refactor

PR #10415, opened 2023-10-05
single_node_performance.py: allow overriding data dir

msmouse:

### DIFF
--- a/testsuite/single_node_performance.py
+++ b/testsuite/single_node_performance.py
@@ -163,6 +163,11 @@ else:
 
 HIDE_OUTPUT = os.environ.get("HIDE_OUTPUT")
 
+DATA_DIR = os.environ.get("DATA_DIR")
+if not(DATA_DIR):
+    print("Please set env_var DATA_DIR")
+    exit()
+
 # Run the single node with performance optimizations enabled
 target_directory = "execution/executor-benchmark/src"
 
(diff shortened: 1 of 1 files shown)


grao1991 (approved):

testsuite/single_node_performance.py
```diff
@@ -163,6 +163,11 @@ class RunGroupConfig:
 
 HIDE_OUTPUT = os.environ.get("HIDE_OUTPUT")
 
+DATA_DIR = os.environ.get("DATA_DIR")
+if not(DATA_DIR):
+    print("Please set env_var DATA_DIR")
```
igor-aptos, line 168: do we want to default one, or require it to be set?

igor-aptos (approved):

testsuite/single_node_performance.py
msmouse: For the purpose of pn2 I think it's better to require people to set.. just one less thing to ask each case that doesn't perform well.

igor-aptos: actually, this will break regression test on PRs, you definitely need to have default :)

msmouse: I mean I don't land it on main

msmouse: (if we do we update the CI workflows to put in /tmp)

sitalkedia (approved):

PR #10418, opened 2023-10-05
[Fix] Reduce logs. (Previous fix was wrong.)

grao1991:


msmouse (approved):
Thought you'd given up writing usage at every version across the board.. that doesn't work?

msmouse (approved):
Thought you'd given up writing usage at every version across the board.. that doesn't work?

grao1991:
> Thought you'd given up writing usage at every version across the board.. that doesn't work?

It should, but I don't want to change the code directly. Let's have it verified on perviewnet first, and roll it out through flag.

PR #10419, opened 2023-10-05
[Storage][Pruner] Make pruners run in parallel.

grao1991:
### Description
Also added a new thread pool for background tasks.

### DIFF
--- a/storage/aptosdb/src/pruner/ledger_pruner/mod.rs
+++ b/storage/aptosdb/src/pruner/ledger_pruner/mod.rs
@@ -24,9 +24,11 @@ use crate::{
     },
     EventStore, TransactionStore,
 };
-use anyhow::Result;
+use anyhow::{anyhow, Result};
+use aptos_experimental_runtimes::thread_manager::THREAD_MANAGER;
 use aptos_logger::info;
 use aptos_types::transaction::{AtomicVersion, Version};
+use rayon::prelude::*;
 use std::{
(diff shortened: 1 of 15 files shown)


areshand (approved):

storage/aptosdb/src/pruner/ledger_pruner/mod.rs
```diff
@@ -67,10 +69,13 @@ impl DBPruner for LedgerPruner {
             self.ledger_metadata_pruner
                 .prune(progress, current_batch_target_version)?;
 
-            // NOTE: If necessary, this can be done in parallel.
-            self.sub_pruners
-                .iter()
-                .try_for_each(|pruner| pruner.prune(progress, current_batch_target_version))?;
+            THREAD_MANAGER.get_io_pool().install(|| {
```
msmouse, line 72: wouldn't these interfere with high pri io gets?

grao1991: This is cheaper than the io_pool operations in commit phase, and the IO pool has 64 threads in the default strategy, so I don't worrying about it for now. Unless you have strong opinion to have another low pri thread pool, and set the priority on threads right now.

grao1991: Offline discussed, added a background pool.

msmouse (approved):

PR #10423, opened 2023-10-06
[aptos-vm] enable automatic creation of sponsored accounts

davidiw:
This is the implementation of AIP-52:
https://github.com/aptos-foundation/AIPs/blob/main/aips/aip-52.md

This AIP proposes to allow sponsored transactions, one in which the gas payer is not the primary signer of the transaction, to create accounts for the primary signer if it does not exist. Currently when submitting a sponsored transaction an account must first exist. This is an unnecessary friction for using gas fee payer accounts for new accounts. As it means that the gas fee payer solution must first submit an independent transaction to the blockchain.

I also cleaned up the tests for fee payer since there were a lot, and they were excessively complex.

### DIFF
--- a/aptos-move/e2e-move-tests/src/tests/fee_payer.rs
+++ b/aptos-move/e2e-move-tests/src/tests/fee_payer.rs
@@ -1,216 +1,17 @@
 // Copyright © Aptos Foundation
 // SPDX-License-Identifier: Apache-2.0
 
-use crate::{assert_success, tests::common, MoveHarness};
-use aptos_language_e2e_tests::{account::TransactionBuilder, transaction_status_eq};
-use aptos_types::{
-    account_address::AccountAddress,
-    move_utils::MemberId,
-    on_chain_config::FeatureFlag,
-    transaction::{EntryFunction, Script, TransactionArgument, TransactionStatus},
+use crate::{assert_success, MoveHarness};
+use aptos_cached_packages::aptos_stdlib;
+use aptos_language_e2e_tests::{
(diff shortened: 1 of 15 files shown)


aptos-move/e2e-move-tests/src/tests/fee_payer.rs
```diff
@@ -1,296 +1,106 @@
...
-        output.status(),
-        &TransactionStatus::Discard(StatusCode::UNEXPECTED_ERROR_FROM_KNOWN_MOVE_FUNCTION)
-    ));
+    assert_eq!(alice_start, alice_after);
+    assert!(bob_start > bob_after);
 }
 
 #[test]
-fn test_normal_tx_with_signer_with_fee_payer() {
+fn test_account_not_exist_with_fee_payer_create_account() {
```
movekevin, line 46: I wonder if we should do this test in api instead to make sure everything works well e2e

movekevin (approved):

aptos-move/e2e-move-tests/src/tests/fee_payer.rs
davidiw: From my understanding, this is comprehensive. But I have another AIP coming, so I'll make a new end to end test for another AIP in this space.

bowenyang007 (approved):

aptos-move/e2e-move-tests/src/tests/fee_payer.rs
davidiw: fyi, tested in the next pr

PR #10429, opened 2023-10-06
[State Sync] Improve request/response observability for peers

JoshLind:
### Description
This PR improves the request/response observability for peers in state sync. Specifically, the PR offers several commits:
1. Separate the logging and metric update frequency in the peer monitoring service (so they can occur at different frequencies).
2. Add request/response tracking for peer buckets to the data client. This allows us to: (i) emit metrics showing which state sync requests are being sent to each peer bucket; and (ii) periodically log the request/response breakdowns by peers and buckets (allowing us to see how data is flowing between peers).
3. Update the node inspection service to include the new request/response breakdowns as well as generic data client metadata in state sync. This is useful for local debugging/observability. See the screenshot below to see how this is exposed:
Screenshot 2023-10-06 at 5 27 06 PM

### Test Plan
Existing test infrastructure.


state-sync/aptos-data-client/src/peer_states.rs
```diff
@@ -13,7 +14,17 @@ use aptos_storage_service_types::{
...
+    cmp::min,
+    collections::{BTreeMap, HashSet},
+    sync::Arc,
+    time::Duration,
+};
+
+// Useful constants
+const LOGS_FREQUENCY_SECS: u64 = 120; // 2 minutes
+const METRICS_FREQUENCY_SECS: u64 = 15; // 15 seconds
+const NUM_PEER_BUCKETS_FOR_METRICS: u8 = 5; // To avoid metric explosion, we bucket peers into groups
```
grao1991, line 27: should we do a power of 2 because you use the first byte module this value as bucket_id? I guess 16 is probably a proper choice.

crates/aptos-inspection-service/src/server/peer_information.rs
```diff
@@ -31,86 +38,263 @@ pub fn handle_peer_information_request(
...
+            ));
+        }
+    }
+}
+
+/// Displays connection metadata for each peer
+fn display_peer_connection_metadata(
+    peer_information_output: &mut Vec<String>,
+    all_peers: &Vec<PeerNetworkId>,
+    peers_and_metadata: &Arc<PeersAndMetadata>,
```
grao1991, line 152: nit: &Arc is weird, you probably want to take &PeersAndMetadata here and other display functions.

grao1991 (approved):

bchocho (approved):

state-sync/aptos-data-client/src/peer_states.rs
JoshLind: Hmm... I think 16 is too large (it'll still lead to some explosion). So, I've changed it to 4. 😄

crates/aptos-inspection-service/src/server/peer_information.rs
JoshLind: I'm not sure it really matters 😄 But, I've changed it to `&PeersAndMetadata`

PR #10434, opened 2023-10-07
[dag] garbage collection

zekun000:

### DIFF
--- a/consensus/src/dag/dag_fetcher.rs
+++ b/consensus/src/dag/dag_fetcher.rs
@@ -9,7 +9,7 @@ use crate::dag::{
     types::{CertifiedNode, FetchResponse, Node, RemoteFetchRequest},
 };
 use anyhow::{anyhow, ensure};
-use aptos_consensus_types::common::Author;
+use aptos_consensus_types::common::{Author, Round};
 use aptos_infallible::RwLock;
 use aptos_logger::{debug, error};
 use aptos_time_service::TimeService;
@@ -207,7 +207,7 @@ impl DagFetcherService {
             RemoteFetchRequest::new(
                 node.metadata().epoch(),
                 missing_parents,
-                dag_reader.bitmask(node.round()),
+                dag_reader.bitmask(node.round().saturating_sub(1)),
             )
         };
         self.inner
@@ -258,6 +258,7 @@ impl TDagFetcher for DagFetcher {
             LogSchema::new(LogEvent::FetchNodes),
             start_round = remote_request.start_round(),
             target_round = remote_request.target_round(),
+            lens = remote_request.exists_bitmask().len(),
             missing_nodes = remote_request.exists_bitmask().num_missing(),
         );
         let mut rpc = RpcWithFallback::new(
@@ -269,30 +270,31 @@ impl TDagFetcher for DagFetcher {
             self.time_service.clone(),
         );
 
-        // TODO retry
         while let Some(response) = rpc.next().await {
-            if let Ok(response) = response
+            match response
                 .and_then(FetchResponse::try_from)
                 .and_then(|response| response.verify(&remote_request, &self.epoch_state.verifier))
             {
-                let certified_nodes = response.certified_nodes();
-                // TODO: support chunk response or fallback to state sync
-                {
-                    let mut dag_writer = dag.write();
-                    for node in certified_nodes.into_iter().rev() {
-                        if let Err(e) = dag_writer.add_node(node) {
-                            error!("Failed to add node {}", e);
+                Ok(response) => {
+                    let certified_nodes = response.certified_nodes();
+                    // TODO: support chunk response or fallback to state sync
+                    {
+                        let mut dag_writer = dag.write();
+                        for node in certified_nodes.into_iter().rev() {
+                            if let Err(e) = dag_writer.add_node(node) {
+                                error!("Failed to add node {}", e);
+                            }
                         }
                     }
-                }
 
-                if dag.read().all_exists(remote_request.targets()) {
-                    return Ok(());
-                }
+                    if dag.read().all_exists(remote_request.targets()) {
+                        return Ok(());
+                    }
+                },
+                Err(err) => error!("Fetch rpc failed {}", err),
             }
-            // TODO retry
         }
-        Err(anyhow!("fetch failed"))
+        Err(anyhow!("Fetch with fallback failed"))
     }
 }
 
@@ -300,6 +302,8 @@ impl TDagFetcher for DagFetcher {
 pub enum FetchRequestHandleError {
     #[error("target nodes are missing")]
     TargetsMissing,
+    #[error("garbage collected, request round {0}, lowest round {1}")]
+    GarbageCollected(Round, Round),
 }
 
 pub struct FetchRequestHandler {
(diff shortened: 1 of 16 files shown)


consensus/src/dag/dag_handler.rs
```diff
@@ -150,11 +150,11 @@ impl NetworkHandler {
                     .to_bytes(&response_msg.into_network_message())
                     .map(Bytes::from)
             })
-            .map_err(RpcError::ApplicationError);
+            .map_err(RpcError::ApplicationError)?;
```
ibalajiarun, line 153: Hmm, I actually wanted to send the response back to the sender as well. Otherwise, the remote rpc might keep retrying after timing out. 

I wanted to distinguish between retriable (parent missing) vs non-retriable errors so sender can make a retry decision accordingly.

consensus/src/dag/dag_state_sync.rs
```diff
@@ -17,13 +17,11 @@ use aptos_time_service::TimeService;
...
     epoch_change::EpochChangeProof, epoch_state::EpochState, ledger_info::LedgerInfoWithSignatures,
 };
-use itertools::Itertools;
 use std::sync::Arc;
 
 // TODO: move this to onchain config
-// TODO: temporarily setting DAG_WINDOW to 1 to maintain Shoal safety
-pub const DAG_WINDOW: usize = 1;
-pub const STATE_SYNC_WINDOW_MULTIPLIER: usize = 30;
+pub const DAG_WINDOW: u64 = 5;
```
ibalajiarun, line 23: time to remove? and just use the on chain config value?

consensus/src/dag/dag_store.rs
```diff
@@ -329,32 +331,42 @@ impl Dag {
     pub fn bitmask(&self, target_round: Round) -> DagSnapshotBitmask {
         let lowest_round = self.lowest_incomplete_round();
 
-        let bitmask = self
+        let mut bitmask: Vec<_> = self
             .nodes_by_round
-            .range(lowest_round..target_round)
+            .range(lowest_round..=target_round)
```
ibalajiarun, line 336: lol

consensus/src/dag/tests/order_rule_tests.rs
```diff
@@ -87,24 +84,23 @@ impl OrderedNotifier for TestNotifier {
         &self,
         ordered_nodes: Vec<Arc<CertifiedNode>>,
         _failed_authors: Vec<(Round, Author)>,
-    ) -> anyhow::Result<()> {
-        Ok(self.tx.unbounded_send(ordered_nodes)?)
+    ) {
+        self.tx.unbounded_send(ordered_nodes).unwrap()
```
ibalajiarun, line 88: why unwrap instead of logging an error?

consensus/src/dag/dag_store.rs
```diff
@@ -339,22 +340,26 @@ impl Dag {
...
+        if let Err(e) = self.storage.delete_certified_nodes(digests) {
             error!("Error deleting expired nodes: {:?}", e);
         }
     }
 
+    pub fn commit_callback(&mut self, commit_round: Round) {
+        let new_round = commit_round.saturating_sub(self.window_size);
+        if new_round > self.start_round {
+            self.start_round = new_round;
+            self.prune();
```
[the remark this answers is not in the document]
zekun000: I don't think this is expensive though

consensus/src/dag/order_rule.rs
```diff
@@ -32,19 +32,14 @@ pub struct OrderRule {
...
         storage: Arc<dyn DAGStorage>,
     ) -> Self {
-        let committed_round = if latest_ledger_info.ends_epoch() {
-            0
-        } else {
-            latest_ledger_info.round()
-        };
         let commit_events = storage
-            .get_latest_k_committed_events(DAG_WINDOW as u64)
+            .get_latest_k_committed_events(DAG_WINDOW)
```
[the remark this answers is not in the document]
zekun000: yes probably another pr

consensus/src/dag/dag_store.rs
```diff
@@ -128,15 +114,31 @@ impl Dag {
     }
 
     pub fn add_node(&mut self, node: CertifiedNode) -> anyhow::Result<()> {
+        ensure!(
```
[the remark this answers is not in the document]
zekun000: I'm using this to prune expired epoch nodes during constructor too

consensus/src/dag/dag_handler.rs
zekun000: our rpc doesn't work like this, it just sends back to network layer and logs there. if we want to send back to sender, we need to add an enum for error message

consensus/src/dag/tests/order_rule_tests.rs
zekun000: I was worried about if one failed for some reason and we continue to send and succeed for some reason. It's simpler to hold the invariant imo

consensus/src/dag/dag_handler.rs
ibalajiarun: where does it log? receiver side or sender side?

zekun000: locally

ibalajiarun: It seems like the rpc receiver side is sending back the whole result? https://github.com/aptos-labs/aptos-core/blob/e7f32fbd72bae6d1a9b334e0b27e4aff68d9ec04/network/framework/src/peer/mod.rs#L272

ibalajiarun: nvm, I see the error result getting dropped later.

ibalajiarun (approved):

consensus/src/dag/dag_store.rs
```diff
@@ -128,15 +114,31 @@ impl Dag {
...
+            "round too low {}, lowest in dag {}",
+            round,
+            self.lowest_round()
+        );
+        ensure!(
+            round <= self.highest_round() + 1,
+            "round too high {}, highest in dag {}",
+            round,
+            self.highest_round()
+        );
```
sasha8, line 141: re: line 154 " self.storage.save_certified_node(&node)?;". We can save this storage call when add_node is called from Dag:new()

consensus/src/dag/dag_store.rs
```diff
@@ -128,15 +114,31 @@ impl Dag {
...
-        ensure!(round >= self.lowest_round(), "round too low");
-        ensure!(round <= self.highest_round() + 1, "round too high");
+        ensure!(
+            round >= self.lowest_round(),
+            "round too low {}, lowest in dag {}",
+            round,
+            self.lowest_round()
+        );
+        ensure!(
+            round <= self.highest_round() + 1,
```
sasha8, line 137: Why do we have this here?

consensus/src/dag/dag_store.rs
```diff
@@ -329,32 +331,42 @@ impl Dag {
...
+            .flat_map(|(_, round_ref)| round_ref.iter().flatten())
+            .map(|node_status| *node_status.as_node().metadata().digest())
+            .collect();
+        if let Err(e) = self.storage.delete_certified_nodes(digests) {
             error!("Error deleting expired nodes: {:?}", e);
         }
     }
 
+    pub fn commit_callback(&mut self, commit_round: Round) {
+        let new_round = commit_round.saturating_sub(self.window_size);
```
sasha8, line 363: "new _ound" => "new_start_round"?

consensus/src/dag/dag_fetcher.rs
```diff
@@ -271,24 +272,27 @@ impl TDagFetcher for DagFetcher {
 
         // TODO retry
         while let Some(response) = rpc.next().await {
-            if let Ok(response) = response
+            match response
```
sasha8, line 275: Delete the "TODO retry"?

sasha8 (approved):

consensus/src/dag/dag_store.rs
zekun000: good point

consensus/src/dag/dag_store.rs
zekun000: what do you mean?

PR #10442, opened 2023-10-08
[SDKv2] Add query lookupOriginalAccountAddress

0xjinn:
### Description

Port over the lookupOriginalAccountAddress query

### Test Plan

pnpm fmt & pnpm test

### DIFF
--- a/ecosystem/typescript/sdk_v2/src/internal/account.ts
+++ b/ecosystem/typescript/sdk_v2/src/internal/account.ts
@@ -50,6 +50,7 @@ import {
   GetAccountTokensCount,
   GetAccountTransactionsCount,
 } from "../types/generated/queries";
+import { getTableItem } from "./general";
 
 export async function getInfo(args: { aptosConfig: AptosConfig; accountAddress: HexInput }): Promise<AccountData> {
   const { aptosConfig, accountAddress } = args;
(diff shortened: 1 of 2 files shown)


0xjinn:
* main
  * **PR #10442** <a href="https://app.graphite.dev/github/pr/aptos-labs/aptos-core/10442" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/aptos-labs/aptos-core/10442?utm_source=stack-comment).

davidiw (commented):
should we add some e2e test for key rotation here?

ecosystem/typescript/sdk_v2/src/internal/account.ts
```diff
@@ -148,6 +149,36 @@ export async function getResource(args: {
...
+  const originalAddr = await getTableItem({
+    aptosConfig,
+    handle,
+    data: {
+      key: Hex.fromHexInput({ hexInput: addressOrAuthKey }).toString(),
+      key_type: "address",
+      value_type: "address",
+    },
+    options: options,
+  });
```
gregnazario, line 177: I wonder if we should migrate to using view functions, so we don't have to deal with standardization of the table address.  That'll help with address migration.

gregnazario (commented):
Let's add an E2E test, since this should be able to run without transaction submission at first, then we can add rotation, which will also test it again.

ecosystem/typescript/sdk_v2/src/internal/account.ts
0xjinn: Ahh that is a good call. I think we can migrate this after the sdkv2 alpha release.

0xjinn:
Closing this PR. Moving to the sdk v2 new repo

PR #10443, opened 2023-10-08
[fee-payer] make fee payer address optionally set during signing

davidiw:
sometimes the client would like to send a transaction to a fee paying network without knowing who will actually pay for it... this PR makes the fee payer address optional by allowing signatures on Address::ZERO

AIP forthcoming will update...

I've built this on the updated sponsored transction PR, will rebase and land as appropriate assuming sufficient approvals land.


bowenyang007 (approved):

davidiw:
ugh... rebase hell

kent-white (approved):

PR #10455, opened 2023-10-10
[Stake] Allow to set the beneficiary for operator

junkil-park:
### Description

This PR allows the operator to set the beneficiary address. This PR implements the AIP-51: https://github.com/aptos-foundation/AIPs/pull/252

### Test Plan
aptos move test

### DIFF
--- a/aptos-move/framework/aptos-framework/sources/delegation_pool.move
+++ b/aptos-move/framework/aptos-framework/sources/delegation_pool.move
@@ -119,11 +119,13 @@ module aptos_framework::delegation_pool {
     use aptos_std::smart_table::{Self, SmartTable};
 
     use aptos_framework::account;
+    use aptos_framework::aptos_account;
     use aptos_framework::aptos_coin::AptosCoin;
     use aptos_framework::aptos_governance;
     use aptos_framework::coin;
-    use aptos_framework::event::{Self, EventHandle};
+    use aptos_framework::event::{Self, EventHandle, emit};
     use aptos_framework::stake;
+    use aptos_framework::stake::get_operator;
     use aptos_framework::staking_config;
     use aptos_framework::timestamp;
 
@@ -182,6 +184,12 @@ module aptos_framework::delegation_pool {
 
     /// The stake pool has already voted on the proposal before enabling partial governance voting on this delegation pool.
     const EALREADY_VOTED_BEFORE_ENABLE_PARTIAL_VOTING: u64 = 17;
+
+    /// The account is not the operator of the stake pool.
+    const ENOT_OPERATOR: u64 = 18;
+
+    /// Chaning beneficiaries for operators is not supported.
+    const EOPERATOR_BENEFICIARY_CHANGE_NOT_SUPPORTED: u64 = 19;
 
     const MAX_U64: u64 = 18446744073709551615;
 
@@ -295,6 +303,10 @@ module aptos_framework::delegation_pool {
         delegate_voting_power_events: EventHandle<DelegateVotingPowerEvent>,
     }
 
+    struct BeneficiaryForOperator has key {
+        beneficiary_for_operator: address,
+    }
+
     struct AddStakeEvent has drop, store {
         pool_address: address,
         delegator_address: address,
@@ -327,6 +339,15 @@ module aptos_framework::delegation_pool {
         commission_pending_inactive: u64,
     }
 
+    #[event]
+    struct DistributeCommission has drop, store {
+        pool_address: address,
+        operator: address,
+        beneficiary: address,
+        commission_active: u64,
+        commission_pending_inactive: u64,
+    }
+
     struct VoteEvent has drop, store {
         voter: address,
         proposal_id: u64,
@@ -345,6 +366,13 @@ module aptos_framework::delegation_pool {
         pool_address: address,
         delegator: address,
         voter: address,
+    }
+
+    #[event]
+    struct SetBeneficiaryForOperator has drop, store {
+        operator: address,
+        old_beneficiary: address,
+        new_beneficiary: address,
     }
 
     #[view]
@@ -444,7 +472,7 @@ module aptos_framework::delegation_pool {
     #[view]
     /// Return total stake owned by `delegator_address` within delegation pool `pool_address`
     /// in each of its individual states: (`active`,`inactive`,`pending_inactive`)
-    public fun get_stake(pool_address: address, delegator_address: address): (u64, u64, u64) acquires DelegationPool {
+    public fun get_stake(pool_address: address, delegator_address: address): (u64, u64, u64) acquires DelegationPool, BeneficiaryForOperator {
         assert_delegation_pool_exists(pool_address);
         let pool = borrow_global<DelegationPool>(pool_address);
         let (
@@ -484,7 +512,7 @@ module aptos_framework::delegation_pool {
         // operator rewards are actually used to buy shares which is introducing
         // some imprecision (received stake would be slightly less)
         // but adding rewards onto the existing stake is still a good approximation
-        if (delegator_address == stake::get_operator(pool_address)) {
+        if (delegator_address == beneficiary_for_operator(get_operator(pool_address))) {
             active = active + commission_active;
             // in-flight pending_inactive commission can coexist with already inactive withdrawal
             if (lockup_cycle_ended) {
@@ -532,7 +560,7 @@ module aptos_framework::delegation_pool {
     #[view]
     /// Return the total voting power of a delegator in a delegation pool. This function syncs DelegationPool to the
     /// latest state.
-    public fun calculate_and_update_voter_total_voting_power(pool_address: address, voter: address): u64 acquires DelegationPool, GovernanceRecords {
+    public fun calculate_and_update_voter_total_voting_power(pool_address: address, voter: address): u64 acquires DelegationPool, GovernanceRecords, BeneficiaryForOperator {
         assert_partial_governance_voting_enabled(pool_address);
         // Delegation pool need to be synced to explain rewards(which could change the coin amount) and
         // commission(which could cause share transfer).
@@ -546,7 +574,7 @@ module aptos_framework::delegation_pool {
     #[view]
     /// Return the remaining voting power of a delegator in a delegation pool on a proposal. This function syncs DelegationPool to the
     /// latest state.
-    public fun calculate_and_update_remaining_voting_power(pool_address: address, voter_address: address, proposal_id: u64): u64 acquires DelegationPool, GovernanceRecords {
+    public fun calculate_and_update_remaining_voting_power(pool_address: address, voter_address: address, proposal_id: u64): u64 acquires DelegationPool, GovernanceRecords, BeneficiaryForOperator {
         assert_partial_governance_voting_enabled(pool_address);
         // If the whole stake pool has no voting power(e.g. it has already voted before partial
         // governance voting flag is enabled), the delegator also has no voting power.
@@ -587,7 +615,7 @@ module aptos_framework::delegation_pool {
         owner: &signer,
         operator_commission_percentage: u64,
         delegation_pool_creation_seed: vector<u8>,
-    ) acquires DelegationPool, GovernanceRecords {
+    ) acquires DelegationPool, GovernanceRecords, BeneficiaryForOperator {
         assert!(features::delegation_pools_enabled(), error::invalid_state(EDELEGATION_POOLS_DISABLED));
         let owner_address = signer::address_of(owner);
         assert!(!owner_cap_exists(owner_address), error::already_exists(EOWNER_CAP_ALREADY_EXISTS));
@@ -634,11 +662,21 @@ module aptos_framework::delegation_pool {
         }
     }
 
+    #[view]
+    /// Return the beneficiary address of the operator.
+    public fun beneficiary_for_operator(operator: address): address acquires BeneficiaryForOperator {
+        if (exists<BeneficiaryForOperator>(operator)) {
+            return borrow_global<BeneficiaryForOperator>(operator).beneficiary_for_operator
+        } else {
+            operator
+        }
+    }
+
     /// Enable partial governance voting on a stake pool. The voter of this stake pool will be managed by this module.
     /// THe existing voter will be replaced. The function is permissionless.
     public entry fun enable_partial_governance_voting(
         pool_address: address,
-    ) acquires DelegationPool, GovernanceRecords {
+    ) acquires DelegationPool, GovernanceRecords, BeneficiaryForOperator {
         assert!(features::partial_governance_voting_enabled(), error::invalid_state(EDISABLED_FUNCTION));
         assert!(features::delegation_pool_partial_governance_voting_enabled(), error::invalid_state(EDISABLED_FUNCTION));
         assert_delegation_pool_exists(pool_address);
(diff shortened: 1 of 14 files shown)


aptos-move/framework/aptos-framework/sources/stake.move
```diff
@@ -197,6 +197,11 @@ module aptos_framework::stake {
         validators: vector<IndividualValidatorPerformance>,
     }
 
+    // The beneficiary address for the operator.
+    struct Beneficiary has key {
```
movekevin, line 201: Create a resource group for this so any more resources added to the stake pool in the future can be in the same resource group for storage efficiency

aptos-move/framework/aptos-framework/sources/stake.move
```diff
@@ -197,6 +197,11 @@ module aptos_framework::stake {
...
+        beneficiary: address
```
movekevin, line 202: Name this more specifically such as operator_beneficiary to make it clearer

aptos-move/framework/aptos-framework/sources/stake.move
```diff
@@ -368,6 +373,16 @@ module aptos_framework::stake {
         borrow_global<StakePool>(pool_address).operator_address
     }
 
+    #[view]
+    /// Return the operator of the validator at `pool_address`.
+    public fun get_beneficiary_for_operator(pool_address: address): address acquires StakePool, Beneficiary {
+        assert_stake_pool_exists(pool_address);
+        if(exists<Beneficiary>(pool_address)) {
```
movekevin, line 380: nit: space after if

aptos-move/framework/aptos-framework/sources/stake.move
```diff
@@ -590,6 +605,20 @@ module aptos_framework::stake {
...
     }
 
+    /// Allows to change the beneficiary for the oeprator of the stake pool.
+    public fun set_beneficiary_for_operator(pool_signer: &signer, new_beneficiary: address) acquires Beneficiary {
+        let pool_address = signer::address_of(pool_signer);
+        assert_stake_pool_exists(pool_address);
+        if (!exists<Beneficiary>(pool_address)) {
+            move_to(pool_signer, Beneficiary { beneficiary: new_beneficiary });
+        }
+        else {
```
movekevin, line 615: nit: Put else on the previous line } else

aptos-move/framework/aptos-framework/sources/stake.move
```diff
@@ -590,6 +605,20 @@ module aptos_framework::stake {
...
+            let beneficiary = borrow_global_mut<Beneficiary>(pool_address);
+            beneficiary.beneficiary = new_beneficiary;
+        }
+        // TODO: emit an event
```
movekevin, line 619: You can emit v2 event now

aptos-move/framework/aptos-framework/sources/staking_contract.move
```diff
@@ -574,6 +586,24 @@ module aptos_framework::staking_contract {
...
         );
+
+        // Set the beneficiary address to the new operator by default.
+        set_beneficiary_for_operator_internal(new_operator, staker_address, new_operator);
+    }
+
+    public entry fun set_beneficiary_for_operator(operator: &signer, staker: address, new_beneficiary: address) acquires Store {
+        let operator_addr = signer::address_of(operator);
+        assert_staking_contract_exists(staker, operator_addr);
+        assert_account_is_registered_for_apt(new_beneficiary);
```
movekevin, line 597: Not needed if we do aptos_account::deposit_coins instead of coin::deposit as the former does registration implicitly if needed.

aptos-move/framework/aptos-framework/sources/stake.move
movekevin, line 201: As discussed, we should move this to staking_contract and delegation_pool. It's a bit redundant but keep the abstraction clear as stake.move is not aware of commission.

movekevin, line 201: Potentially put this resource on the operator account as well so the contracts don't need to remember to reset it when operator is changed.

junkil-park: This has been removed from the `stake` module.

aptos-move/framework/aptos-framework/sources/stake.move
junkil-park: How about leaving it as it is? We could potentially use this for other types of accounts too (e.g., staker).

aptos-move/framework/aptos-framework/sources/stake.move
junkil-park: This code is removed from the stake module.

aptos-move/framework/aptos-framework/sources/stake.move
junkil-park: Done using Event V2.

aptos-move/framework/aptos-framework/sources/stake.move
junkil-park: Done moving this resource to `staking_contract` and `delegation_pool` separately.

junkil-park: Done refactoring in this way.

aptos-move/framework/aptos-framework/sources/staking_contract.move
junkil-park: Done.

aptos-move/framework/aptos-framework/sources/stake.move
junkil-park: This function is removed from the module.

aptos-move/framework/aptos-framework/sources/delegation_pool.move
```diff
@@ -956,19 +990,36 @@ module aptos_framework::delegation_pool {
...
         let pool_address = get_owned_pool_address(signer::address_of(owner));
         // synchronize delegation and stake pools before any user operation
         // ensure the old operator is paid its uncommitted commission rewards
         synchronize_delegation_pool(pool_address);
         stake::set_operator(&retrieve_stake_pool_owner(borrow_global<DelegationPool>(pool_address)), new_operator);
     }
 
+    /// Allows an operator to change its beneficiary.
+    public entry fun set_beneficiary_for_operator(operator: &signer, new_beneficiary: address) acquires Beneficiary {
+        let operator_addr = signer::address_of(operator);
```
alexfilip2, line 1003: Storing `delegation_pool::Beneficiary` within the operator account has some advantages:
- decoupling the beneficiary functionality from the delegation pool
- an operator managing multiple pools collects rewards into the same account always
However, what happens to the commission rewards produced under the old beneficiary and not collected yet? At the next `synchronize_delegation_pool` they will be all granted to the new beneficiary.

aptos-move/framework/aptos-framework/sources/delegation_pool.move
```diff
@@ -295,6 +299,10 @@ module aptos_framework::delegation_pool {
         delegate_voting_power_events: EventHandle<DelegateVotingPowerEvent>,
     }
 
+    struct Beneficiary has key {
```
alexfilip2, line 302: Would the following design be an option worth exploring?
- `Beneficiary` is stored on the resource account of the delegation pool
- Operator which is the default/initial beneficiary can set the next beneficiary the first time
- Only the current beneficiary can change themselves to a different account
Usually, the operator account is used for operations on the `stake` module to rotate validator keys in case of infrastructure migrations etc. The entity managing the validator software usually grants access to the operator to their infrastructure/DevOps team. The beneficiary can be a cold wallet owned by the finance department, but in the current design the operator can still change it which is not completely separating:

- who manages operator-oriented operations
- who receives the rewards for restaking/withdrawing them

aptos-move/framework/aptos-framework/sources/delegation_pool.move
```diff
@@ -956,19 +990,36 @@ module aptos_framework::delegation_pool {
...
+    ) acquires DelegationPoolOwnership, DelegationPool, GovernanceRecords, Beneficiary {
         let pool_address = get_owned_pool_address(signer::address_of(owner));
         // synchronize delegation and stake pools before any user operation
         // ensure the old operator is paid its uncommitted commission rewards
         synchronize_delegation_pool(pool_address);
         stake::set_operator(&retrieve_stake_pool_owner(borrow_global<DelegationPool>(pool_address)), new_operator);
     }
 
+    /// Allows an operator to change its beneficiary.
+    public entry fun set_beneficiary_for_operator(operator: &signer, new_beneficiary: address) acquires Beneficiary {
```
movekevin, line 1002: Can you add some documentation here that changing the beneficiary would also mean any existing unpaid commission would be paid to the new beneficiary as well?

aptos-move/framework/aptos-framework/sources/staking_contract.move
```diff
@@ -576,9 +601,26 @@ module aptos_framework::staking_contract {
         );
     }
 
+    /// Allows an operator to change its beneficiary.
+    public entry fun set_beneficiary_for_operator(operator: &signer, new_beneficiary: address) acquires Beneficiary {
+        let operator_addr = signer::address_of(operator);
```
movekevin, line 606: Can you leave a comment to document that we don't need to validate the operator here as this is a resource stored on the operator account and not at the staking contract level?

movekevin (approved):

aptos-move/framework/aptos-framework/sources/delegation_pool.move
movekevin: I left a comment on this as well but I think this makes sense and is needed in case the old beneficiary account is compromised. This gives the operator full control in how to distribute the commissions, they can call synchronize first before changing the beneficiary if they need the unpaid commission to go to the old beneficiary first.

aptos-move/framework/aptos-framework/sources/delegation_pool.move
movekevin: The main complexity this would lead to is dealing with wiping the beneficiary account when changing operators

alexfilip2: yes, changing the operator requires defaulting the beneficiary too (e.g. to this new operator's address)

junkil-park: The permission structure of the current design is: the owner has permission to set an operator, and the operator has permission to set a beneficiary. I believe that the current design satisfies the requirements that @michelle-aptos has. @michelle-aptos, do you have any comments on this?

junkil-park: Also, if the operator has permission to change its beneficiary, it can handle the situation where the beneficiary account is compromised.

aptos-move/framework/aptos-framework/sources/staking_contract.move
junkil-park: Done.

aptos-move/framework/aptos-framework/sources/delegation_pool.move
junkil-park: Done.

aptos-move/framework/aptos-framework/sources/delegation_pool.move
junkil-park: I added a comment to explain it.

aptos-move/framework/aptos-framework/sources/delegation_pool.move
```diff
@@ -3211,13 +3273,74 @@ module aptos_framework::delegation_pool {
         assert_delegation(new_operator_address, pool_address, 26050290, 0, 26050290);
     }
 
+    #[test(aptos_framework = @aptos_framework, operator = @0x123, delegator = @0x010, beneficiary = @0x020)]
+    public entry fun test_set_beneficiary_for_operator(
```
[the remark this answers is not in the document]
junkil-park: Added the scenarios at the end of the test functions.

aptos-move/framework/aptos-framework/sources/delegation_pool.move
michelle-aptos: the operator should only have permission to change the beneficiary, but the beneficiary receives the rewards and manages the restaking/withdrawal.

aptos-move/framework/aptos-framework/sources/delegation_pool.move
```diff
@@ -327,6 +339,15 @@ module aptos_framework::delegation_pool {
         commission_pending_inactive: u64,
     }
 
+    #[event]
+    struct DistributeCommissionEventV2 has drop, store {
```
lightmark, line 343:
```suggestion
    struct DistributeCommission has drop, store {
```
Let's abandon `EventV2` suffix. v2 is more like an internal codename

aptos-move/framework/aptos-framework/sources/delegation_pool.move
```diff
@@ -347,6 +368,13 @@ module aptos_framework::delegation_pool {
         voter: address,
     }
 
+    #[event]
+    struct SetBeneficiaryForOperator has drop, store {
```
lightmark, line 372: Do we have another event? If not, the suffix `ForOperator` seems unnecessary. And is it `of operator` or `for operator`?

aptos-move/framework/aptos-framework/sources/delegation_pool.move
lightmark: agree with the original comment the beneficiary should be controlled by validator...

aptos-move/framework/aptos-framework/sources/delegation_pool.move
```diff
@@ -295,6 +303,10 @@ module aptos_framework::delegation_pool {
         delegate_voting_power_events: EventHandle<DelegateVotingPowerEvent>,
     }
 
+    struct Beneficiary has key {
+        beneficiary: address,
+    }
+
     struct AddStakeEvent has drop, store {
```
lightmark, line 310: how do you want to migrate to event v2? since you add new v2 events but don't migrate existing v1 events...

aptos-move/framework/aptos-framework/sources/delegation_pool.move
lightmark: n00b Q:
> an operator managing multiple pools collects rewards into the same account always

Is this what we want?

lightmark (approved):
Not quite understand the design of putting beneficiary under operators account.
The events naming may need to be fixed.
otherwise looks good. unblock the PR

aptos-move/framework/aptos-framework/sources/delegation_pool.move
junkil-park: This feature is for the operator rather than the pool owner. This feature allows the operator to choose where to receive the operator commission.

aptos-move/framework/aptos-framework/sources/stake.move
junkil-park: I renamed them `BeneficiaryForOperator` and `beneficiary_for_operator` to be more specific.

aptos-move/framework/aptos-framework/sources/delegation_pool.move
junkil-park: > an operator managing multiple pools
This case does not really exist and is not recommended. However, if it does, it is expected that the commissions from multiple pools will go to the same beneficiary.

aptos-move/framework/aptos-framework/sources/delegation_pool.move
junkil-park: Renamed. Thanks!

aptos-move/framework/aptos-framework/sources/delegation_pool.move
junkil-park: This PR is not migrating all the existing V1 events, but a follow-up PR will do.

aptos-move/framework/aptos-framework/sources/delegation_pool.move
junkil-park: "ForOperator" specifies what it's for. Potentially, there can be beneficiaries for other types of accounts such as stakers.

PR #10456, opened 2023-10-10
[Data Streaming Service] Support early termination in lagging subscription streams.

JoshLind:
### Description
This PR updates the data streaming service in state sync to support early termination of lagging subscription streams. This is useful for subscription streams that cause peers to lag behind the rest of the chain (e.g., if the network channel used by the stream is slow or faulty, or if the peer serving the subscription stream is actively malicious). In these cases, the data streaming service will monitor the lag of a subscription stream and terminate the stream prematurely, iff: (i) the stream is lagging for too long; and (ii) the lag has increased since the last check (indicating that the lag is getting worse and the stream is falling behind).

Note: to avoid being overly aggressive, we don't terminate subscriptions streams that lag temporarily, exhibit a "bounded" lag (that doesn't get worse), or lag and then catch up.

The PR offers several commits:
1. Add early termination support for lagging subscription streams.
2. Add unit tests for the new functionality.
3. Add metrics for stream lag.

### Test Plan

### DIFF
--- a/state-sync/data-streaming-service/src/data_stream.rs
+++ b/state-sync/data-streaming-service/src/data_stream.rs
@@ -172,6 +172,24 @@ impl<T: AptosDataClientInterface + Send + Clone + 'static> DataStream<T> {
         self.sent_data_requests.is_some()
     }
 
+    /// Resets the subscription stream lag on the data stream
+    fn reset_subscription_stream_lag(&mut self) {
+        // Reset the subscription stream lag metrics
+        metrics::set_subscription_stream_lag(0);
+
+        // Reset the stream lag
+        self.subscription_stream_lag = None;
+    }
+
+    /// Sets the subscription stream lag on the data stream
+    fn set_subscription_stream_lag(&mut self, subscription_stream_lag: SubscriptionStreamLag) {
+        // Update the subscription stream lag metrics
+        metrics::set_subscription_stream_lag(subscription_stream_lag.version_lag);
+
+        // Set the stream lag
+        self.subscription_stream_lag = Some(subscription_stream_lag)
+    }
+
     /// Initializes the data client requests by sending out the first batch
     pub fn initialize_data_requests(
         &mut self,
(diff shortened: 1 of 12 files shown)


bchocho:
Not a blocker, but wondering if we should add some adversarial FN tests. Would be useful for state sync, and also mempool dissemination.

state-sync/data-streaming-service/src/data_stream.rs
```diff
@@ -483,6 +503,102 @@ impl<T: AptosDataClientInterface + Send + Clone + 'static> DataStream<T> {
...
+                return Err(
+                    aptos_data_client::error::Error::SubscriptionStreamIsLagging(format!(
+                        "The subscription stream is beyond recovery! Current lag: {:?}, last lag: {:?},",
+                        current_stream_lag, subscription_stream_lag.version_lag
+                    )),
+                );
+            }
+
+            // The stream is lagging, but it's not yet beyond recovery
+            self.subscription_stream_lag = Some(subscription_stream_lag);
```
bchocho, line 577: Would it make sense to have metrics on the lag, both when its "beyond recovery" and even when it's not?

bchocho (approved):
LGTM

state-sync/data-streaming-service/src/data_stream.rs
JoshLind: Yeah, I think that's a great idea. We already track some of this today (i.e., `lag = advertised - synced`). But, it would be useful to measure this, too. Let me update it 😄

state-sync/data-streaming-service/src/metrics.rs
```diff
@@ -171,6 +171,15 @@ pub static RECEIVED_RESPONSE_ERROR: Lazy<IntCounterVec> = Lazy::new(|| {
     .unwrap()
 });
 
+/// Counter that keeps track of the subscription stream lag
+pub static SUBSCRIPTION_STREAM_LAG: Lazy<IntGauge> = Lazy::new(|| {
+    register_int_gauge!(
+        "aptos_data_streaming_service_subscription_stream_lag",
+        "Counters related to the subscription stream lag",
```
brianolson, line 178: units? versions? seconds? transactions?

JoshLind: Versions 😄 Updated the comment.

brianolson (approved):

PR #10458, opened 2023-10-10
[Storage] Don't use Rayon for Async Proof fetcher

sitalkedia:
### Description

Looking at the flamegraph from recent executor benchmark, we see that around 5-7% of the CPU is being spent in Rayon's task stealing. The issue is that we spawn too many task (one task per key - ~100k task for a block of size 25k) and these tasks are very small, spawning too many small tasks on rayon thread pool has a huge overhead that more the thread pool is spending most of the CPU cycles in rayon's task stealing. Changing this from Rayon to threadpool (which doesn't have any task stealing) gets rid of the rayon overhead and gives us around 3-4% TPS boost. 

### Test Plan

Ran the executor benchmark and observed 3-4% TPS improvements.

### DIFF
--- a/storage/storage-interface/Cargo.toml
+++ b/storage/storage-interface/Cargo.toml
@@ -34,7 +34,7 @@ parking_lot = { workspace = true }
 rayon = { workspace = true }
 serde = { workspace = true }
 thiserror = { workspace = true }
-threadpool = "1.8.1"
+threadpool = { workspace = true }
 
 [dev-dependencies]
 aptos-types = { workspace = true, features = ["fuzzing"] }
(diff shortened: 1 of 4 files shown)


storage/storage-interface/Cargo.toml
```diff
@@ -34,6 +34,7 @@ parking_lot = { workspace = true }
 rayon = { workspace = true }
 serde = { workspace = true }
 thiserror = { workspace = true }
+threadpool = "1.8.1"
```
grao1991, line 37: move this to the top level Cargo.toml?

grao1991 (approved):

msmouse (approved):

PR #10462, opened 2023-10-10
[types] require that fee payer acknowledge that they are fee payer

davidiw:
if the fee payer doesn't sign off on the fee payer, any signature can be used, which in turn means any signature can be used as the fee payer.

### Test Plan
tests are updated to represent the new protocol....

### DIFF
--- a/types/src/transaction/authenticator.rs
+++ b/types/src/transaction/authenticator.rs
@@ -226,7 +226,7 @@ impl TransactionAuthenticator {
                 // address, so we verify both and only fail if the signature fails for either of
                 // them. The legacy approach is to assume the address of the fee payer is signed
                 // over.
-                let mut to_verify = vec![sender, fee_payer_signer];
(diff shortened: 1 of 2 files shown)


types/src/transaction/authenticator.rs
```diff
@@ -226,7 +226,7 @@ impl TransactionAuthenticator {
                 // address, so we verify both and only fail if the signature fails for either of
                 // them. The legacy approach is to assume the address of the fee payer is signed
                 // over.
-                let mut to_verify = vec![sender, fee_payer_signer];
+                let mut to_verify = vec![sender];
```
lightmark, line 229: you don't need  a vector for this anymore?

davidiw: we append to it in line 232 and take the remaining in line 246.

gregnazario (approved):

movekevin (approved):

PR #10464, opened 2023-10-11
[Execution][Rayon] Tune minimum parallelism to avoid task stealing overhead

sitalkedia:
### Description

- Rayon task stealing has huge CPU overhead because the minimum parallelism is not tuned properly.  This change makes the parallelism adaptive based on the number of tasks. 

- Also limit the max number of threads for most of the thread pools to be 32 - this change actually give us 5-7% TPS improvement. 

### Test Plan

Ran the executor benchmark on `t2d-standard-60` machine and observed 15% TPS improvement. 

Command ran - 

```
cargo run --profile performance -p aptos-executor-benchmark  -- --execution-threads 48 --generate-then-execute --transactions-per-sender 1 --block-size 32000 --enable-storage-sharding run-executor  --main-signer-accounts 100000  --data-dir ~/data/db2 --checkpoint-dir ~/data/chk --blocks 50
```

This gives us a consistent 40k TPS - which is around 15% higher than 34k TPS without this change.


experimental/runtimes/src/thread_manager.rs
```diff
@@ -65,3 +66,10 @@ impl ThreadManagerBuilder {
         }
     }
 }
+
+/// This assumes that we get a peak performance at 48 parallelism - in future if we can scale up
+/// our code for higher parallelism, we can increase this number.
+static OPTIMAL_MAX_PARALLELISM: usize = 48;
+pub fn optimal_min_parallelism(num_tasks: usize, min_threshold: usize) -> usize {
```
msmouse: "parallelism" in "max parallelism" and "min parallelism" obviously mean different things?

experimental/runtimes/src/thread_manager.rs
```diff
@@ -65,3 +66,10 @@ impl ThreadManagerBuilder {
         }
     }
 }
+
+/// This assumes that we get a peak performance at 48 parallelism - in future if we can scale up
+/// our code for higher parallelism, we can increase this number.
+static OPTIMAL_MAX_PARALLELISM: usize = 48;
```
msmouse: Do we want each worker on average have a few tasks rather than one (in the case of our 60 cores experiments)? 
I can see bad cases where we increase concurrent_threads from 48 to 54 and do not see improvements, or one single task being blocked by something else in the system and we have to wait for it.

experimental/runtimes/src/thread_manager.rs
sitalkedia: Do you have a better suggestion for naming :)

experimental/runtimes/src/thread_manager.rs
sitalkedia: Good point, I actually reduced the max number of threads to 32 and increased this to 64, so on an average, there will be 2 tasks per thread. This reduces the task stealing overhead as well as leaves some room for load balancing in the worst case. This configuration performs optimal in my experiment and gives us around 15% performance boost.

grao1991: 2 tasks per thread doesn't sound right, unless the task is super cheap. My intuition is to make one task at 100us to 1ms level.

experimental/runtimes/src/strategies/default.rs
```diff
@@ -14,9 +15,11 @@ pub struct DefaultThreadManager {
...
 impl DefaultThreadManager {
     pub(crate) fn new() -> DefaultThreadManager {
-        let exe_threads = spawn_rayon_thread_pool("exe".into(), Some(num_cpus::get()));
-        let non_exe_threads = spawn_rayon_thread_pool("non_exe".into(), Some(num_cpus::get()));
-        let io_threads = spawn_rayon_thread_pool("io".into(), Some(64));
+        // Do not use more than 32 threads for rayon thread pools as we have seen scalability issues with more threads.
+        // This needs to be revisited once we resolve the scalability issues.
+        let exe_threads = spawn_rayon_thread_pool("exe".into(), Some(min(num_cpus::get(), 32)));
+        let non_exe_threads = spawn_rayon_thread_pool("non_exe".into(), Some(min(num_cpus::get(), 32)));
+        let io_threads = spawn_rayon_thread_pool("io".into(), Some(32));
```
grao1991, line 22: We probably want to leave IO threads to a higher number, when it hits IO you will need more threads to get enough thoughput. also tasks in this pool tent to be larger, so the overhead ratio should be smaller.

experimental/runtimes/src/thread_manager.rs
sitalkedia: > 2 tasks per thread doesn't sound right, unless the task is super cheap.

Just to be clear, task here is stealable task - and each stealable task might contain ~1000s of small tasks. 

> in other word it should be based on the computation cost of the task, not a global number

Why does it need to be based on the computation cost? We know the parallelism in the system and as long as we know the tasks are uniform, the best strategy is to evenly distribute tasks across threads, leaving some room for task stealing but not allow task stealing on a huge scale that is happening today.

experimental/runtimes/src/strategies/default.rs
sitalkedia: I doubt having higher number of these threads are helping even when the DB is large - because even in that case I assume we are CPU bound. Can you help run the large DB benchmark you have in mind and see if we are seeing any regression?

experimental/runtimes/src/thread_manager.rs
msmouse: `optimal_min_len()` to be brutal?

msmouse:
Can you list out the actual benchmark command you've run?  For the purpose of documentation.

experimental/runtimes/src/thread_manager.rs
```diff
@@ -65,3 +66,10 @@ impl ThreadManagerBuilder {
         }
     }
 }
+
+/// This assumes that we get a peak performance at 64 parallelism - in future if we can scale up
+/// our code for higher parallelism, we can increase this number.
+static OPTIMAL_MAX_PARALLELISM: usize = 64;
```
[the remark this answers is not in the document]
sitalkedia: Hmm..sure we can do that but its probably not worth it.. passing the thread pool context (or size) down to each function. Current values are tuned for a maximum of 64 concurrency, which is what we see our single node performance benchmarks peaking at anyways.. we can update these values in future if we find a way to scale beyond 64 threads..

experimental/runtimes/src/thread_manager.rs
grao1991: You don't want a stealable task to be too large so that one thread is doing it while other threads are waiting and cannot steal.
On the other hand you don't want it to be too small so that the stealing benefit is less than the overhead.

experimental/runtimes/src/thread_manager.rs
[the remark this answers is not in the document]
msmouse: How much do we lose if this is 128? 

(indeed we should rely less on these micro parallelism for sure).

experimental/runtimes/src/strategies/default.rs
grao1991: with a very little amount of IO (account creation on 100M db on a 128G RAM machine), the put_stats_and_indices which uses this pool is ~15% to ~20% slower.. I'd expect in real world it would be worse.

sitalkedia: How about we run the large DB benchmark on this change - if that is significantly slower, we can bump this?

experimental/runtimes/src/thread_manager.rs
sitalkedia: Made this 4 * number of threads - trying to find a balance between both of them is tricky for sure.

experimental/runtimes/src/strategies/default.rs
grao1991: If you only look at TPS you probably won't see the difference because the execution is slower today (and because the large DB benchmark is not really large, unless you run it on a smaller RAM machine), but conceptually this is not something I want to put in production. 

BTW I also did a side by side comparison for 32 v.s. 64 on small db for p2p, and I don't really see improvement. Maybe it just implicitly slow down the commit slightly and speed up execution slightly..

grao1991 (approved):

msmouse (approved):

PR #10470, opened 2023-10-11
[Compiler] remove an unnecessary assertion in visibility translation

rahxephon89:
### Description

This PR removes an assertion in visibility translation. This assertion will cause the compiler to panic when the target move code contains `public (script)` functions and there are other errors at or above `NonblockingError` level. It is not necessary because the compiler will output these errors after compilation.

### DIFF
--- a/third_party/move/move-compiler/tests/move_check/deprecated/public_script.exp
+++ b/third_party/move/move-compiler/tests/move_check/deprecated/public_script.exp
@@ -1,0 +1,12 @@
+warning[W00001]: DEPRECATED. will be removed
+  ┌─ tests/move_check/deprecated/public_script.move:2:5
+  │
+2 │     public(script) fun main() {
+  │     ^^^^^^^^^^^^^^ 'public(script)' is deprecated in favor of the 'entry' modifier. Replace with 'public entry'
+
+error[E03001]: address with no value
+  ┌─ tests/move_check/deprecated/public_script.move:4:30
+  │
+4 │         let _addr:address = @Test;
+  │                              ^^^^ address 'Test' is not assigned a value. Try assigning it a value when calling the compiler
+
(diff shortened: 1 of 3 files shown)


vineethk:
Can you please add a test case (perhaps the simplified version of the code that led to your discovery of this bug)?

rahxephon89:
> Can you please add a test case (perhaps the simplified version of the code that led to your discovery of this bug)?

Sure, do you have a recommendation on where to add the test?

vineethk:
> Sure, do you have a recommendation on where to add the test?

Others may have better places to suggest, but `third_party/move/move-compiler/tests/move_check/deprecated` or one of its sibling directories seems reasonable.

rahxephon89:
> > Sure, do you have a recommendation on where to add the test?
> 
> Others may have better places to suggest, but `third_party/move/move-compiler/tests/move_check/deprecated` or one of its sibling directories seems reasonable.

Thanks, I added a test case.

vineethk (approved):
lgtm

third_party/move/move-compiler/tests/move_check/deprecated/public_script.exp
```diff
@@ -0,0 +1,12 @@
...
+  ┌─ tests/move_check/deprecated/public_script.move:2:5
+  │
+2 │     public(script) fun main() {
+  │     ^^^^^^^^^^^^^^ 'public(script)' is deprecated in favor of the 'entry' modifier. Replace with 'public entry'
+
+error[E03001]: address with no value
+  ┌─ tests/move_check/deprecated/public_script.move:3:30
+  │
+3 │         let _addr:address = @Test;
+  │                              ^^^^ address 'Test' is not assigned a value. Try assigning it a value when calling the compiler
```
brmataptos, line 11: Any idea how this test is different than third_party/move/move-compiler/tests/move_check/parser/function_visibility_script.move ?  Why does this one trigger the assertion?

brmataptos (commented):
What is special about this test?

third_party/move/move-compiler/tests/move_check/deprecated/public_script.exp
rahxephon89: This assertion will cause the compiler to panic when the target move code contains public (script) functions and there are other errors at or above NonblockingError level. This example injects an error.

brmataptos: Ideally, you might add a comment to indicate this:
```
// Previously, deprecation plus an error led to a compiler assert.  Make sure that doesn't come back.
```

brmataptos (approved):

PR #10472, opened 2023-10-11
[localnet] surface up the error for docker compose up failure

larry-aptos:
### Description

* Throws exception when docker compose up failes.
* Also we need to build an image with `main` for indexer grpc. 

### Test Plan

* an example looks like this

image


rustielin (approved):

davidiw (approved):

larry-aptos:
Abandoned; this is replaced by the CLI localnet.

PR #10474, opened 2023-10-11
[Forge] Add forge network benchmark tests and clean up some logic.

JoshLind:
### Description
This PR adds a suite of forge network benchmark tests (to help identify network bottlenecks). It also moves some match statements into grouped functions to help reduce the size/complexity of managing many different forge tests in a single function. Given that we expect to continue adding forge tests, it probably makes sense to introduce some type of refactoring now.

### Test Plan
Existing test infrastructure.

### DIFF
--- a/testsuite/forge-cli/src/main.rs
+++ b/testsuite/forge-cli/src/main.rs
@@ -69,6 +69,10 @@ use std::{
 use tokio::{runtime::Runtime, select};
 use url::Url;
 
+// Useful constants
+const KILOBYTE: usize = 1000;
+const MEGABYTE: usize = KILOBYTE * 1000;
+
 #[cfg(unix)]
 #[global_allocator]
 static ALLOC: jemallocator::Jemalloc = jemallocator::Jemalloc;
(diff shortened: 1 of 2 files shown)


bchocho (approved):
LGTM

testsuite/forge-cli/src/main.rs
```diff
@@ -525,32 +529,27 @@ fn single_test_suite(
     duration: Duration,
     test_cmd: &TestCommand,
 ) -> Result<ForgeConfig> {
+    // Check the test name against grouped test suites
+    if let Some(forge_config) = get_land_blocking_test(test_name, duration, test_cmd) {
```
brianolson, line 533: this to me is less clear than the previous way that matched names to setup functions. If we want to have some kind of grouping that was already being done with comments interspersed in the list.

JoshLind: I think this is debatable. 😄 The issue I see is that we're going to keep adding to this list (in fact, I have several sets of network tests to add, as well as PFN tests, etc.). So, unless folks are strict in keeping the list tidy, and ordered (e.g., alphabetically or by functionality?), it is going to get very unwieldy quickly. In fact, I'd argue it already is.

If we see a better approach, I'm happy to follow it, but given that this code is "unowned", it tends to messy very fast. At least with the grouping, it's easier to split things up.

brianolson (approved):

testsuite/forge-cli/src/main.rs
JoshLind: FWIW, I've added a comment to the function. Hopefully someone can propose/implement a cleaner solution 😄

PR #10477, opened 2023-10-11
[Storage] Support db_path_overrides in StorageConfig, to allow people putting different db into different root path.

grao1991:
### Description
This makes it more flexible to support multiple disks. 
If RAID is not a valid option, people can choose to mount different disks to different folders, and load balance shards among them.

Here is an example, DBs that not present in the overrides will stay at the place specified by upper level config.

Screenshot 2023-10-16 at 2 45 09 PM


config/src/config/storage_config.rs
```diff
@@ -113,6 +169,10 @@ pub struct StorageConfig {
     /// since genesis. To recover operation after data loss, or to bootstrap a node in fast sync
     /// mode, the indexer db needs to be copied in from another node.
     pub enable_indexer: bool,
+    /// Fine grained control for db paths of individal databases/shards.
+    /// If not specificed, will use `dir` as default.
+    /// Only allowed when sharding is enabled.
```
areshand, line 174: just curious why sharding enabled is required?

storage/aptosdb/src/lib.rs
```diff
@@ -484,19 +490,39 @@ impl AptosDB {
...
     pub fn open_dbs<P: AsRef<Path> + Clone>(
         db_root_path: P,
+        db_path_overrides: Option<DbPathConfig>,
         rocksdb_configs: RocksdbConfigs,
         readonly: bool,
         max_num_nodes_per_lru_cache_shard: usize,
     ) -> Result<(LedgerDb, StateMerkleDb, StateKvDb)> {
-        let ledger_db = LedgerDb::new(db_root_path.as_ref(), rocksdb_configs, readonly)?;
+        let ledger_db = LedgerDb::new(
+            db_path_overrides
```
areshand, line 499: any chance that we provide some util functions to get these db pathes that other component can use?

config/src/config/storage_config.rs
grao1991: Just to reduce the complexity of implementation.. there will be no need to put data on multiple disks if db is not sharded.

storage/aptosdb/src/lib.rs
grao1991: Rewrote, PTAL

areshand (approved):

msmouse (approved):

PR #10478, opened 2023-10-11
[Docs] Consolidate local development docs and explain new features

banool:
### Description
This PR adds a doc guiding users through using the new local testnet features. In this I rebrand it to a "local development network". I clean up a lot of related docs to reduce duplication and funnel people into the one doc.

### Test Plan
CI.


bowenyang007 (approved):

developer-docs-site/docs/guides/local-development-network.md
```diff
@@ -0,0 +1,346 @@
...
+```
+/var/run/docker.sock
+```
+
+If you're on Mac or Windows, we recommend you use Docker Desktop rather than installing Docker via a package manager (e.g. Homebrew or Choco).
+
+### How do I use the Postgres on my host machine?
+By default when using `--with-indexer-api` the CLI will run a Postgres instance in Docker. If you have Postgres running on your host machine and would like to use that instead, you can do so with the `--use-host-postgres` flag. There are also flags for specifying how it should connect to the host Postgres. Here is an example invocation:
+```bash
+aptos node run-local-testnet --with-indexer-api --use-host-postgres --postgres-user dport
```
gregnazario, line 238: Please switch to the common variable $USER, so it will be easier to translate rather then determining what dport is

gregnazario (approved):

PR #10487, opened 2023-10-12
[gas-profiler] support storage refunds and generate report as a webpage

vgao1996:
## Overview
This is a major enhancement to the gas profiler that introduces support for storage refunds and the capability to generate the gas report as a local webpage.

Here's an example:
Firefox_Screenshot_2023-10-12T12-44-38 230Z

More samples: 
[sample-gas-reports.zip](https://github.com/aptos-labs/aptos-core/files/12882495/sample-gas-reports.zip)


vgao1996:
Here's a break-down of the changes to make it easier to review this PR:
- aptos-gas-profiling/src
  - erased.rs
    - Change to the tree structure to include refunds -- not that important
  - log.rs
    - Added support for storage refunds
    - Two helper functions to check if the log is consistent. Being consistent means all the costs add up.
  - profiler.rs
    - Some small changes to support storage refunds
  - textualize.rs
    - Deleted since it's a burden to maintain the the text-based report when the web-based one is available.
  - report.rs
    - The main file of this PR. This is basically a function that converts the transaction gas log into a json object that can be used by the handlebars templating engine to render a web page.
- aptos-gas-profiling/templates
  - index.html
    - The web template written in the handlebars format

msmouse (approved):

wrwg (approved):
Wow, this pretty cool.

PR #10494, opened 2023-10-12
[State Sync] Migrate to the new storage synchronizer pipeline.

JoshLind:
### Description
This PR updates the storage synchronizer in state sync to migrate from a 2-step synchronization pipeline (i.e., `apply/execute -> commit`), to the 3-step pipeline (i.e., `apply/execute -> ledger_update -> commit`).

### Test Plan
New and existing tests.

### DIFF
--- a/state-sync/state-sync-driver/src/storage_synchronizer.rs
+++ b/state-sync/state-sync-driver/src/storage_synchronizer.rs
@@ -190,12 +190,16 @@ impl<
         metadata_storage: MetadataStorage,
         storage: DbReaderWriter,
         runtime: Option<&Runtime>,
-    ) -> (Self, JoinHandle<()>, JoinHandle<()>) {
+    ) -> (Self, JoinHandle<()>, JoinHandle<()>, JoinHandle<()>) {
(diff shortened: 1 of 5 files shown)


brianolson (approved):

state-sync/state-sync-driver/src/storage_synchronizer.rs
```diff
@@ -207,6 +211,16 @@ impl<
             chunk_executor.clone(),
             error_notification_sender.clone(),
             executor_listener,
+            ledger_updater_notifier,
```
areshand, line 214: one orthogonal question. would this also cover the genesis ledger data since  I assume there is no execution done for version 0?

areshand (approved):

state-sync/state-sync-driver/src/storage_synchronizer.rs
JoshLind: Correct -- there is no execution for the genesis data, so it doesn't even enter this flow 😄 'cc @msmouse

PR #10495, opened 2023-10-12
Mark AggregatorV1 as #[deprecated] in move

igor-aptos:

### DIFF
--- a/aptos-move/framework/aptos-framework/sources/aggregator/aggregator.move
+++ b/aptos-move/framework/aptos-framework/sources/aggregator/aggregator.move
@@ -1,3 +1,5 @@
+/// DEPRECATED in favor of `aptos_framework::aggregator_v2`.
+///
 /// This module provides an interface for aggregators. Aggregators are similar to
 /// unsigned integers and support addition and subtraction (aborting on underflow
 /// or on overflowing a custom upper limit). The difference from integers is that
@@ -23,6 +25,7 @@ module aptos_framework::aggregator {
 
     /// Represents an integer which supports parallel additions and subtractions
     /// across multiple transactions. See the module description for more details.
+    #[deprecated]
     struct Aggregator has store {
         handle: address,
         key: address,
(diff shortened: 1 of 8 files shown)


aptos-move/framework/aptos-framework/sources/coin.move
```diff
@@ -79,6 +79,7 @@ module aptos_framework::coin {
     /// Represents a coin with aggregator as its value. This allows to update
     /// the coin in every transaction avoiding read-modify-write conflicts. Only
     /// used for gas fees distribution by Aptos Framework (0x1).
+    #[deprecated]
```
georgemitenkov, line 82: We use it in `CollectedFeesPerBlock` which has never been used. Should that also be deprecated?

igor-aptos: I can mark that as deprecated too

aptos-move/framework/aptos-framework/sources/coin.move
```diff
@@ -162,6 +163,7 @@ module aptos_framework::coin {
 
     /// Creates a new aggregatable coin with value overflowing on `limit`. Note that this function can
     /// only be called by Aptos Framework (0x1) account for now becuase of `create_aggregator`.
+    #[deprecated]
```
vusirikala, line 166: Dumb question. Does adding the deprecated flag affect the functionality in anyway? Or does it just show a warning to the developer that this function is deprecated when he uses it?

georgemitenkov: I think it's only a compiler warning

brmataptos: Yes, just a compiler warning.

movekevin (commented):
Is there any need to mark these as deprecated if they are privileged functions anyway?

aptos-move/framework/aptos-framework/sources/coin.move
brmataptos: You should probably do that.  While we still suppress warnings on Aptos framework code, that's kind of kludgey.  We also suppress use of deprecated things warnings in code which is itself deprecated, which is less kludgey.

aptos-move/framework/aptos-framework/sources/aggregator/aggregator.move
```diff
@@ -1,3 +1,5 @@
+/// DEPRECATED in favor of `aptos_framework::aggregator_v2`.
```
brmataptos, line 1: If the transition to v2 is not trivial, you may want to point to some useful doc on how to do that.

gelash (approved):

aptos-move/framework/aptos-framework/sources/transaction_fee.move
```diff
@@ -32,6 +32,8 @@ module aptos_framework::transaction_fee {
 
     /// Stores information about the block proposer and the amount of fees
     /// collected when executing the block.
+    /// TODO needs to be modified to AggregatorV2 to be able to be used.
```
georgemitenkov, line 35: Do we need this TODO?

georgemitenkov (approved):

gregnazario:
> Is there any need to mark these as deprecated if they are privileged functions anyway?

Closing this for now.  Please re-open if you decide to resolve the conflicts

PR #10497, opened 2023-10-12
[move_vm] Fix loader dependency check

runtian-zhou:
### Description

The runtime dependency check is a bit stricter than what bytecode verifier permits. Relax the constraint so that runtime check aligns with the static behavior.

### Test Plan

Added a test case. The test would fail without the fix.

### DIFF
--- a/third_party/move/move-vm/types/src/loaded_data/runtime_types.rs
+++ b/third_party/move/move-vm/types/src/loaded_data/runtime_types.rs
@@ -127,8 +127,14 @@ impl StructType {
 
     // Check if the local struct handle is compatible with the defined struct type.
     pub fn check_compatibility(&self, struct_handle: &StructHandle) -> PartialVMResult<()> {
-        if !self.abilities.is_subset(struct_handle.abilities)
-            || self.phantom_ty_args_mask.len() != struct_handle.type_parameters.len()
+        if !self.abilities.is_subset(struct_handle.abilities) {
+            return Err(
+                PartialVMError::new(StatusCode::UNKNOWN_INVARIANT_VIOLATION_ERROR)
+                    .with_message("Ability definition of module mismatch".to_string()),
+            );
+        }
+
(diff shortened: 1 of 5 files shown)


third_party/move/move-vm/types/src/loaded_data/runtime_types.rs
```diff
@@ -122,6 +124,27 @@ impl StructType {
...
+    // Check if the local struct handle is compatible with the defined struct type.
+    pub fn check_compatibility(&self, struct_handle: &StructHandle) -> PartialVMResult<()> {
+        if !self.abilities.is_subset(struct_handle.abilities)
+            || self.phantom_ty_args_mask.len() != struct_handle.type_parameters.len()
+            || !self
+                .phantom_ty_args_mask
+                .iter()
+                .zip(struct_handle.type_parameters.iter())
+                .all(|(defined_is_phantom, local_type_parameter)| {
+                    !local_type_parameter.is_phantom || *defined_is_phantom
```
runtian-zhou, line 137: This logic needs to align with the rules specified in https://github.com/aptos-labs/aptos-core/blob/main/third_party/move/move-bytecode-verifier/src/dependencies.rs

third_party/move/move-vm/types/src/loaded_data/runtime_types.rs
```diff
@@ -122,6 +124,27 @@ impl StructType {
...
+                })
+        {
+            return Err(
+                PartialVMError::new(StatusCode::UNKNOWN_INVARIANT_VIOLATION_ERROR)
+                    .with_message("Ability definition of module mismatch".to_string()),
```
vgao1996, line 142: Would be better if we can have more specialized error messages for each kind of the violations listed in the if condition.

vgao1996 (approved):
Looks correct to me

wrwg (approved):

PR #10502, opened 2023-10-13
[VM] Limit the maximum identifier length

junkil-park:
### Description

This PR limits the maximum identifier length for the deserializer and serializer. The update on the deserializer is feature-gated.

### Test Plan
cargo test


third_party/move/move-binary-format/src/file_format_common.rs
```diff
@@ -59,6 +59,7 @@ pub const BYTECODE_INDEX_MAX: u64 = 65535;
 pub const LOCAL_INDEX_MAX: u64 = 255;
 
 pub const IDENTIFIER_SIZE_MAX: u64 = 65535;
```
vgao1996, line 61: Can you rename this to something like "LEGACY_IDENTIFIER_SIZE_MAX"? 

Would be even better if we could mark it as deprecated. Not sure if this will work though.

third_party/move/move-vm/runtime/src/config.rs
```diff
@@ -12,6 +12,7 @@ pub const DEFAULT_MAX_VALUE_NEST_DEPTH: u64 = 128;
 pub struct VMConfig {
     pub verifier: VerifierConfig,
     pub max_binary_format_version: u32,
+    pub max_identifier_size: u32,
```
vgao1996, line 15: How about we just include the whole `DeserializeConfig` as a field here? This way we can easily add new config options for the deserializer without having to touch the VM.

vgao1996 (changes requested):
Maybe I missed something, but I don't see the new identifier size config being applied. I mean, you are supposed to touch this function, right:
```
fn load_identifier_size(cursor: &mut VersionedCursor) -> BinaryLoaderResult<usize> {
    read_uleb_internal(cursor, IDENTIFIER_SIZE_MAX)
}
```

Should also fix the serializer. This one does not need to be feature gated though. Just make it use the new limit.
```
fn serialize_identifier_size(binary: &mut BinaryData, len: usize) -> Result<()> {
    write_as_uleb128(binary, len as u64, IDENTIFIER_SIZE_MAX)
}
```

vgao1996:
Otherwise, the PR looks good. You are following exactly the standard way to handle this type of fixes.

third_party/move/move-binary-format/src/file_format_common.rs
junkil-park: Done renaming!

I think we can deprecate consts. Perhaps, we should do it after the feature is enabled in every network?

junkil-park:
>Maybe I missed something, but I don't see the new identifier size config being applied. I mean, you are supposed to touch this function, right:

Fixed this. Right, this part of the change was not committed by mistake.

junkil-park:
> Should also fix the serializer. This one does not need to be feature gated though. Just make it use the new limit.

After renaming the consts, `IDENTIFIER_SIZE_MAX` now denotes the new value of 255. So, the serializer will use the new value without updating the code.

third_party/move/move-binary-format/src/file_format_common.rs
```diff
@@ -423,16 +424,22 @@ pub(crate) mod versioned_data {
     use std::io::{Cursor, Read};
     pub struct VersionedBinary<'a> {
         version: u32,
+        max_identifier_size: u64,
         binary: &'a [u8],
     }
 
     pub struct VersionedCursor<'a> {
         version: u32,
+        max_identifier_size: u64,
```
brmataptos, line 433: There's something a bit smelly about having this as a field in a "VersionedCursor", but I suppose it's ok for now.  Next time we extend this we should think about a separate VersionedBinaryConfig struct instead.

brmataptos (approved):

wrwg (approved):

vgao1996 (approved):

PR #10503, opened 2023-10-13
[small][agg_v2] Fix identifier mapping - layout caching, and MVHashMap storing

igor-aptos:

### DIFF
--- a/third_party/move/move-vm/runtime/src/loader/mod.rs
+++ b/third_party/move/move-vm/runtime/src/loader/mod.rs
@@ -1709,15 +1709,21 @@ impl Script {
     }
 }
 
+#[derive(Clone)]
+struct StructLayoutInfoCacheItem {
+    struct_layout: MoveStructLayout,
+    node_count: u64,
+    has_identifier_mappings: bool,
+}
+
 //
 // Cache for data associated to a Struct, used for de/serialization and more
 //
 #[derive(Clone)]
 struct StructInfoCache {
     struct_tag: Option<(StructTag, u64)>,
-    struct_layout: Option<MoveStructLayout>,
+    struct_layout_info: Option<StructLayoutInfoCacheItem>,
     annotated_struct_layout: Option<MoveStructLayout>,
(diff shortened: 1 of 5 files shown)


third_party/move/move-vm/runtime/src/loader/mod.rs
```diff
@@ -1993,68 +2007,62 @@ impl Loader {
...
+                )
             },
             Type::Struct { name, .. } => {
                 *count += 1;
-                MoveTypeLayout::Struct(self.struct_name_to_type_layout(
-                    name,
-                    &[],
-                    count,
-                    has_identifier_mappings,
-                    depth,
```
igor-aptos, line 2046: is it correct to not use depth+1 here? is this only vectorr depth, but not object depth? should it be renamed?

igor-aptos: @runtian-zhou

gelash (approved):
LGTM, but indeed @runtian-zhou please have a look?

third_party/move/move-vm/runtime/src/loader/mod.rs
georgemitenkov: But the depth is incremented when processing each field in a struct?

georgemitenkov (approved):

third_party/move/move-vm/runtime/src/loader/mod.rs
```diff
@@ -1719,25 +1719,30 @@ impl Script {
     }
 }
 
+#[derive(Clone)]
+struct StructLayoutInfoCacheItem {
+    struct_layout: MoveStructLayout,
+    node_count: u64,
+    has_identifier_mappings: bool,
```
vusirikala, line 1726: What are node_count and has_identifier_mappings?

igor-aptos: number of recursive nodes, and whether struct has a field with (aggregator) identifier

Issue #9222, opened 2023-07-21
[Bug] Misleading compiler error when passing the wrong type to a function with generics


movekevin:
│
101 │         if (smart_table::contains(my_simple_map, &key)) {
    │             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Could not infer this type. Try adding an annotation

When passing a simple_map to a smart_table function that requires the generic, the expected error should be argument type mismatch, not "Could not infer this type".

PR #10505, opened 2023-10-13
[compiler v1] Better compiler error messages (#9222)

fEst1ck:
### Description
This improves the compiler error messages when type inference fails by

- removing duplicated error messages for the same uninferred type variable ([uninferred_type_unpack_assign.exp](https://github.com/aptos-labs/aptos-core/compare/main...fEst1ck:aptos-core:better-compiler-err-msg#diff-f1676ba2459a1234f8bb7314ec79fee1ad79868ceace16149c2f5557c1204e3e)).
- telling why we cannot infer; e.g., which generic needs to be provided.
### Test Plan
Existing tests of compiler v1.

### DIFF
--- a/third_party/move/move-compiler/src/typing/expand.rs
+++ b/third_party/move/move-compiler/src/typing/expand.rs
@@ -362,8 +362,10 @@ fn lvalue(context: &mut Context, b: &mut T::LValue) {
             type_(context, ty);
             core::check_non_fun(context, ty.as_ref())
         },
-        L::BorrowUnpack(_, _, _, bts, fields) | L::Unpack(_, _, bts, fields) => {
-            types(context, bts);
+        L::BorrowUnpack(_, mod_id, struct_name, bts, fields) | L::Unpack(mod_id, struct_name, bts, fields) => {
+            for (i, b) in bts.iter_mut().enumerate() {
+                type_struct_ty_param(context, b, i, &TypeName_::ModuleType(*mod_id, *struct_name))
+            }
             for (_, _, (_, (bt, innerb))) in fields.iter_mut() {
                 type_(context, bt);
                 lvalue(context, innerb)
--- a/third_party/move/move-compiler/src/typing/expand.rs
+++ b/third_party/move/move-compiler/src/typing/expand.rs
@@ -5,9 +5,9 @@
 use super::core::{self, Context, Subst, forward_tvar};
 use crate::{
     diag,
-    expansion::ast::Value_,
+    expansion::ast::{Value_, ModuleIdent},
     naming::ast::{BuiltinTypeName_, FunctionSignature, Type, TypeName_, Type_},
-    parser::ast::Ability_,
+    parser::ast::{Ability_, FunctionName},
     typing::ast as T,
 };
 use move_core_types::u256::U256;
@@ -67,6 +67,11 @@ fn type_struct_ty_param(context: &mut Context, ty: &mut Type, i: usize, ty_name:
         },
         _ => type_(context, ty),
     }
+}
+
+fn type_fun_ty_param(context: &mut Context, ty: &mut Type, i: usize, m: &ModuleIdent, n: &FunctionName) {
+   let param_name = context.function_info(m, n).signature.type_parameters[i].user_specified_name;
+   type_msg_(context, ty, &format!("Cannot infer the type parameter {param_name} for generic function {m}::{n}. Try providing a type parameter."));
 }
 
 pub fn type_msg_(context: &mut Context, ty: &mut Type, msg_uninferred: &str) {
@@ -375,7 +380,9 @@ fn lvalue(context: &mut Context, b: &mut T::LValue) {
 }
 
 fn module_call(context: &mut Context, call: &mut T::ModuleCall) {
-    types(context, &mut call.type_arguments);
+    for (i, t) in call.type_arguments.iter_mut().enumerate() {
+        type_fun_ty_param(context, t, i, &call.module, &call.name);
+    }
     exp(context, &mut call.arguments);
     types(context, &mut call.parameter_types)
 }
--- a/third_party/move/move-compiler/src/typing/expand.rs
+++ b/third_party/move/move-compiler/src/typing/expand.rs
@@ -2,7 +2,7 @@
 // Copyright (c) The Move Contributors
 // SPDX-License-Identifier: Apache-2.0
 
-use super::core::{self, Context};
+use super::core::{self, Context, Subst, forward_tvar};
 use crate::{
     diag,
     expansion::ast::Value_,
@@ -64,7 +64,23 @@ pub fn type_(context: &mut Context, ty: &mut Type) {
         Ref(_, b) => type_(context, b),
         Var(tvar) => {
             let ty_tvar = sp(ty.loc, Var(*tvar));
-            let replacement = core::unfold_type(&context.subst, ty_tvar);
+            fn unfold_type_mut(subst: &mut Subst, sp!(loc, t_): Type) -> Type {
+                match t_ {
+                    Type_::Var(i) => {
+                        let last_tvar = forward_tvar(subst, i);
+                        match subst.get(last_tvar) {
+                            Some(sp!(_, Type_::Var(_))) => unreachable!(),
+                            None => {
+                                subst.insert(last_tvar, sp(loc, Type_::UnresolvedError));
+                                sp(loc, Type_::Anything)
+                            },
+                            Some(inner) => inner.clone(),
+                        }
+                    },
+                    x => sp(loc, x),
+                }
+            }
+            let replacement = unfold_type_mut(&mut context.subst, ty_tvar);
             let replacement = match replacement {
                 sp!(_, Var(_)) => panic!("ICE unfold_type_base failed to expand"),
                 sp!(loc, Anything) => {
--- a/third_party/move/move-compiler/src/typing/expand.rs
+++ b/third_party/move/move-compiler/src/typing/expand.rs
@@ -85,7 +85,8 @@ fn type_fun_ty_param(
     type_msg_(context, ty, &format!("Cannot infer the type parameter `{param_name}` for generic function `{m}::{n}`. Try providing a type parameter."));
 }
 
-pub fn type_msg_(context: &mut Context, ty: &mut Type, msg_uninferred: &str) {
+// Try to expand the type of ty, warning with msg_uninferred if type cannot be inferred.
+fn type_msg_(context: &mut Context, ty: &mut Type, msg_uninferred: &str) {
     use Type_::*;
     match &mut ty.value {
         Anything | UnresolvedError | Param(_) | Unit => (),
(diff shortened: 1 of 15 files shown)


third_party/move/move-compiler/src/typing/expand.rs
```diff
@@ -57,43 +57,84 @@ fn types(context: &mut Context, ss: &mut Vec<Type>) {
...
+        _ => type_(context, ty),
+    }
+}
+
+fn type_fun_ty_param(context: &mut Context, ty: &mut Type, i: usize, m: &ModuleIdent, n: &FunctionName) {
+   let param_name = context.function_info(m, n).signature.type_parameters[i].user_specified_name;
+   type_msg_(context, ty, &format!("Cannot infer the type parameter {param_name} for generic function {m}::{n}. Try providing a type parameter."));
+}
+
+pub fn type_msg_(context: &mut Context, ty: &mut Type, msg_uninferred: &str) {
```
brmataptos, line 77: This name isn't not self-explanatory.  Add a comment, perhaps something like:
```
/// Try to expand the type of ty, warning with msg_uninferred if type cannot be inferred.
```
HOWEVER, I think it would be better to change this method to `try_to_infer_type(context, ty): bool` which returns true on success, and have the message be created only if the it returns failure.  This would avoid creating a lot of unneeded warning messages, just in case they might be needed.

brmataptos (commented):
Changing `type_msg_` to a more meaningful name and having it return `bool` to indicate that the type could not be inferred and the caller should issue a warning would be more useful than creating a lot of warning messages that mostly won't be used.

third_party/move/move-compiler/src/typing/expand.rs
```diff
@@ -57,43 +57,84 @@ fn types(context: &mut Context, ss: &mut Vec<Type>) {
     }
 }
 
-pub fn type_(context: &mut Context, ty: &mut Type) {
+// type_ for expanding the `i`-th type param of struct `ty_name`
+fn type_struct_ty_param(context: &mut Context, ty: &mut Type, i: usize, ty_name: &TypeName_) {
+    match ty_name {
+        TypeName_::ModuleType(mod_id, struct_name) => {
+            let param_name = &context.struct_definition(mod_id, struct_name).type_parameters[i].param.user_specified_name.value;
+            let msg = format!("Cannot infer the type parameter {param_name} for generic struct {ty_name}. Try providing a type parameter.");
```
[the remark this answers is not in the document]
vineethk: It's the other way round for me, I am used to reading inline format (from Python f-strings, eg) and find this style easier.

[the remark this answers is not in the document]
brmataptos: Unlike Python, Rust isn't an interpreted language.  So I expect stuff in string literals to be unrelated to variable names.  This recent extension of the format! macro isn't great, in my opinion.  But I guess the compiler will sort it out.

third_party/move/move-compiler/src/typing/expand.rs
fEst1ck: I was halfway transitting, and realized this: there cannot be a `try_to_infer_type(context, ty) -> bool` that lets the caller choose what to do when type inference fails. Consider when we are expanding `S<X, Y>`, and we cannot infer `X`. The caller only gets a `false`, and doesn't know which parameter cannot be inferred, so no way to provide proper diagnostics.

Admittedly, `type_msg_` is a bad name, but I derived that from the original `type_`, which uses a general warning for uninferred types.

Hence I will maintain the current design, and add comments for `type_msg_`.

third_party/move/move-compiler/src/typing/expand.rs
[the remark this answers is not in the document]
fEst1ck: resolved in https://github.com/aptos-labs/aptos-core/pull/10505/commits/1195c84e4dec35ab1ce00fc1f803c42ec1c40ae5

third_party/move/move-compiler/src/typing/expand.rs
```diff
@@ -57,21 +57,65 @@ fn types(context: &mut Context, ss: &mut Vec<Type>) {
...
         Ref(_, b) => type_(context, b),
         Var(tvar) => {
             let ty_tvar = sp(ty.loc, Var(*tvar));
-            let replacement = core::unfold_type(&context.subst, ty_tvar);
+            fn unfold_type_mut(subst: &mut Subst, sp!(loc, t_): Type) -> Type {
+                match t_ {
+                    Type_::Var(i) => {
+                        let last_tvar = forward_tvar(subst, i);
+                        match subst.get(last_tvar) {
+                            Some(sp!(_, Type_::Var(_))) => unreachable!(),
```
brmataptos, line 101: If this is truly unreachable, why do you need to test for it?  The `Some(inner)` clause below encapsulates this case.

third_party/move/move-compiler/src/typing/expand.rs
```diff
@@ -57,21 +57,65 @@ fn types(context: &mut Context, ss: &mut Vec<Type>) {
...
+    i: usize,
+    m: &ModuleIdent,
+    n: &FunctionName,
+) {
+    let param_name = context.function_info(m, n).signature.type_parameters[i].user_specified_name;
+    type_msg_(context, ty, &format!("Cannot infer the type parameter `{param_name}` for generic function `{m}::{n}`. Try providing a type parameter."));
+}
+
+// Try to expand the type of ty, warning with msg_uninferred if type cannot be inferred.
+fn type_msg_(context: &mut Context, ty: &mut Type, msg_uninferred: &str) {
```
brmataptos, line 89: Better to use a lambda parameter for msg_uninferred so you can defer string construction unless it's needed.  Probably need a dyn Fun to avoid generic expansion, though.

third_party/move/move-compiler/src/typing/expand.rs
```diff
@@ -57,21 +57,65 @@ fn types(context: &mut Context, ss: &mut Vec<Type>) {
...
     use Type_::*;
     match &mut ty.value {
         Anything | UnresolvedError | Param(_) | Unit => (),
         Ref(_, b) => type_(context, b),
         Var(tvar) => {
             let ty_tvar = sp(ty.loc, Var(*tvar));
-            let replacement = core::unfold_type(&context.subst, ty_tvar);
+            fn unfold_type_mut(subst: &mut Subst, sp!(loc, t_): Type) -> Type {
```
wrwg, line 96: Please do not nest function definitions deeply inside other definitions, particular if its large like this. Pull to the outer scope. Also why is there `mut` in the function name?

third_party/move/move-compiler/src/typing/expand.rs
wrwg: Use `panic!` with message as on line 114 in this code.

Otherwise its good practice to have this case explicit

third_party/move/move-compiler/src/typing/expand.rs
wrwg, line 89: Better name then `type_msg_` as `type_with_context_msg`? Also the convention in this code to use a trailing `_` in a name is when you try to avoid a keyword, or you are processing the non-spanned version of a type, which is not the case here, so no `_` needed at the end of the name.

third_party/move/move-compiler/src/typing/expand.rs
wrwg, line 96: Why is `core::unfold_type` not doing what you are trying to achieve here? It should unfold the type by its final replacement, does it not?

third_party/move/move-compiler/src/typing/expand.rs
```diff
@@ -57,21 +57,65 @@ fn types(context: &mut Context, ss: &mut Vec<Type>) {
...
+                    },
+                    x => sp(loc, x),
+                }
+            }
+            let replacement = unfold_type_mut(&mut context.subst, ty_tvar);
             let replacement = match replacement {
                 sp!(_, Var(_)) => panic!("ICE unfold_type_base failed to expand"),
                 sp!(loc, Anything) => {
-                    let msg = "Could not infer this type. Try adding an annotation";
                     context
```
fEst1ck, line 115: This is to avoid duplicate error messages for uninferred type variables. In the first time they are resolved to Err(_), and Anything for all following queries.

third_party/move/move-compiler/src/typing/expand.rs
```diff
@@ -57,43 +57,84 @@ fn types(context: &mut Context, ss: &mut Vec<Type>) {
     }
 }
 
-pub fn type_(context: &mut Context, ty: &mut Type) {
+// type_ for expanding the `i`-th type param of struct `ty_name`
+fn type_struct_ty_param(context: &mut Context, ty: &mut Type, i: usize, ty_name: &TypeName_) {
+    match ty_name {
```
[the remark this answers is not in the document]
fEst1ck: done

third_party/move/move-compiler/src/typing/expand.rs
fEst1ck: > If this is truly unreachable, why do you need to test for it? The `Some(inner)` clause below encapsulates this case.

It should be truly unreachable, and this is like a debug assert.

fEst1ck: done

third_party/move/move-compiler/src/typing/expand.rs
fEst1ck: done

third_party/move/move-compiler/src/typing/expand.rs
fEst1ck: done

fEst1ck:
PTLA

third_party/move/move-compiler/src/typing/expand.rs
```diff
@@ -57,43 +57,98 @@ fn types(context: &mut Context, ss: &mut Vec<Type>) {
...
+                        type_struct_ty_param(context, ty, i, &ty_name.value)
+                    }
                 },
                 _ => panic!("ICE impossible. tapply switched to nontapply"),
             }
         },
     }
 }
 
+pub fn type_(context: &mut Context, ty: &mut Type) {
```
brmataptos, line 144: While you're modifying this, can you add a comment to say what this function does?    Perhaps
```
// Try to expand the type of ty, warning with a default message if type cannot be inferred.
```
This whole file is woefully undocumented.  Thanks for figuring it out for us.

third_party/move/move-compiler/src/typing/expand.rs
brmataptos: Please add a comment with this text.

brmataptos (approved):
Please see 2 suggested comments to add.

third_party/move/move-compiler/src/typing/expand.rs
fEst1ck: done

fEst1ck:
PTLA @vineethk  @wrwg

vineethk (approved):
lgtm.

PR #10509, opened 2023-10-13
[trivial][agg v2] Limit string length as safety for now

igor-aptos:
and have it be larger than token name

### DIFF
--- a/aptos-move/framework/src/natives/aggregator_natives/aggregator_v2.rs
+++ b/aptos-move/framework/src/natives/aggregator_natives/aggregator_v2.rs
@@ -46,9 +46,14 @@ pub const EAGGREGATOR_API_NOT_ENABLED: u64 = 0x03_0006;
 /// The generic type supplied to the aggregators is not supported.
 pub const EUNSUPPORTED_AGGREGATOR_TYPE: u64 = 0x03_0007;
 
+/// Arguments passed to concat exceed max limit of 256 bytes (for prefix and suffix together)
+pub const ECONCAT_STRING_LENGTH_TOO_LARGE: u64 = 0x03_0008;
+
 /// The native aggregator function, that is in the move file, is not yet supported.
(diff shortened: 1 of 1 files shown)


aptos-move/framework/src/natives/aggregator_natives/aggregator_v2.rs
```diff
@@ -495,6 +500,12 @@ fn native_string_concat(
 
     let prefix = string_to_bytes(safely_pop_arg!(args, Struct))?;
 
+    if prefix.len() + suffix.len() > CONCAT_PREFIX_AND_SUFFIX_MAX_LENGTH {
```
vusirikala, line 503: Shouldn't we be checking that the length of Prefix + Snapshot + Suffix <= 256?

georgemitenkov: I don't think we can check it because then we read the snapshot value? Unless it is a speculative read

georgemitenkov, line 503: Can `prefix.len() + suffix.len()` overflow?

georgemitenkov (approved):

aptos-move/framework/src/natives/aggregator_natives/aggregator_v2.rs
vusirikala: In that case we can check Prefix + 8 + Suffix <= 256, assuming snapshot length is at most 8 bytes?

igor-aptos: u128 is up to 20 characters.

but why to sum up, it's just a random constant? since I want it to be predictable - having it explicit on prefix and suffix seems cleaner.

vusirikala (approved):

PR #10512, opened 2023-10-14
use crazy machines and no VFN

msmouse:


sitalkedia:
@msmouse - You mean to merge this change in `previewnet` branch and not `main` right?

sitalkedia:
Also, seems like this PR contain two unrelated changes - 1. reverting the VM loader stuff and 2. configuration changes for higher TPS. Possible to split it into two commits so that the previewnet branch history remains cleaner?

msmouse:
@sitalkedia  yes this is not mean to be landed on main, just demonstrating what you need to run the forge tests. previewnet2 branch is not cut yet.

PR #10515, opened 2023-10-14
[telemetry-svc][bugfix] Fix PeerId check for custom event

ibalajiarun:
### Description

This PR fixes PeerId check during custom event Ingestion. #8727 changed how PeerId is represented as string by default making it a backwards-incompatible change. Therefore, instead of comparing strings, this PR parses the peer ID string into a PeerId struct and compares it instead.

### Test Plan

Added new unit tests to catch this incompatibility.

### DIFF
--- a/crates/aptos-telemetry-service/src/custom_event.rs
+++ b/crates/aptos-telemetry-service/src/custom_event.rs
@@ -14,9 +14,10 @@ use crate::{
     },
 };
 use anyhow::anyhow;
+use aptos_types::PeerId;
 use gcp_bigquery_client::model::table_data_insert_all_request::TableDataInsertAllRequest;
 use serde_json::json;
-use std::time::Duration;
+use std::{str::FromStr, time::Duration};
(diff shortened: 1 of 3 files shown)


crates/aptos-telemetry-service/src/custom_event.rs
```diff
@@ -129,3 +141,60 @@ pub(crate) async fn handle_custom_event(
...
+            events: vec![TelemetryEvent {
+                name: "test".into(),
+                params: BTreeMap::new(),
+            }],
+        };
+        assert_ok!(validate_custom_event_body(&claims, &body));
+
+        let body = TelemetryDump {
+            client_id: String::new(),
+            user_id: String::from("1234"),
```
bchocho, line 191: Do you also want to add the test with some leading zeros?

bchocho (approved):
LGTM, one nit

rustielin (approved):

PR #10518, opened 2023-10-15
[fee-payer/python-sdk] Fix Optional Fee payer validation and Python SDK updates around authenticators

davidiw:
add secp256k1 ecdsa demo
clean up authenticator and account code

due to some weird interactions with this PR: https://github.com/aptos-labs/aptos-core/pull/10351

this also removes the feature gating on the updates to sponsored transactions with respect to verifying them conditionally.

all changes thoroughly tested with a couple new end-to-end tests in Python


ecosystem/python/sdk/aptos_sdk/secp256k1_ecdsa.py
```diff
@@ -0,0 +1,224 @@
...
+
+    def public_key(self) -> PublicKey:
+        return PublicKey(self.key.verifying_key)
+
+    @staticmethod
+    def random() -> PrivateKey:
+        return PrivateKey(SigningKey.generate(curve=SECP256k1, hashfunc=hashlib.sha3_256))
+
+    def sign(self, data: bytes) -> Signature:
+        return Signature(self.key.sign(data))
```
0xjinn, line 53: ohh, so we don't have to take the `sha3_256()` of the input data?

xbtmatt: it looks like it's set in the signing key by default as the hash function, see anywhere a PrivateKey instance is made

```python
    return PrivateKey(SigningKey.generate(curve=SECP256k1, hashfunc=hashlib.sha3_256))
```

davidiw: And public key

ecosystem/python/sdk/aptos_sdk/account.py
```diff
@@ -70,10 +77,46 @@ def auth_key(self) -> str:
         """Returns the auth_key for the associated account"""
         return str(AccountAddress.from_key(self.private_key.public_key()))
 
-    def sign(self, data: bytes) -> ed25519.Signature:
+    def sign(self, data: bytes) -> asymmetric_crypto.Signature:
         return self.private_key.sign(data)
 
-    def public_key(self) -> ed25519.PublicKey:
+    def sign_into_simulated_authenticator(self) -> authenticator.Authenticator:
```
xbtmatt, line 83: cc @0xmaayan We should implement something like this for the authenticator classes, that way it's at the class level vs a helper functioncc @0xmaayan We should implement something like this for the authenticator classes, that way it's at the class level vs a helper function

grao1991 (approved):

ecosystem/python/sdk/examples/fee_payer_transfer_coin.py
```diff
@@ -0,0 +1,108 @@
...
+        )
+        raise Exception("This should never be accessed")
+    except ApiError:
+        alice_sequence_number = 0
+
+    try:
+        bob_balance = await rest_client.account_balance(bob.address())
+        raise Exception("This should never be accessed")
+    except ResourceNotFound:
+        bob_balance = None
```
gregnazario, line 53: FYI, this is a perfect example of how `account_balance` might want to instead just return 0 or None rather than having to wrap it like this

ecosystem/python/sdk/aptos_sdk/transactions.py
```diff
@@ -128,30 +175,35 @@ def __init__(
...
+        self.fee_payer = fee_payer
 
-    def verify(self, key: ed25519.PublicKey, signature: ed25519.Signature) -> bool:
-        return key.verify(self.keyed(), signature)
+    def serialize(self, serializer: Serializer):
+        serializer.u8(1)
+        serializer.struct(self.raw_transaction)
+        serializer.sequence(self.secondary_signers, Serializer.struct)
+        fee_payer = (
+            AccountAddress.from_str("0x0") if self.fee_payer is None else self.fee_payer
```
gregnazario, line 204: May be nice to provide these kinds of addresses as constants.

ecosystem/python/sdk/aptos_sdk/secp256k1_ecdsa.py
```diff
@@ -0,0 +1,233 @@
...
+        return self.hex()
+
+    @staticmethod
+    def from_str(value: str) -> PublicKey:
+        if value[0:2] == "0x":
+            value = value[2:]
+        if (
+            len(value) != PublicKey.LENGTH * 2
+            and len(value) != PublicKey.PREFIX_LENGTH * 2
+        ):
```
gregnazario, line 100: Should probably make these a more common parsing function since you're always checking length.

Though, why prefix length and length?  You can provide both lengths?  Shouldn't this be:

```
len(value) != (PublicKey.PREFIX_LENGTH + PublicKey.LENGTH) * 2
```

ecosystem/python/sdk/aptos_sdk/authenticator.py
```diff
@@ -109,14 +175,77 @@ def serialize(self, serializer: Serializer):
...
+class FeePayerAuthenticator:
+    sender: AccountAuthenticator
+    secondary_signers: List[typing.Tuple[AccountAddress, AccountAuthenticator]]
+    fee_payer: typing.Tuple[AccountAddress, AccountAuthenticator]
+
+    def __init__(
+        self,
+        sender: AccountAuthenticator,
+        secondary_signers: List[typing.Tuple[AccountAddress, AccountAuthenticator]],
+        fee_payer: typing.Tuple[AccountAddress, AccountAuthenticator],
```
gregnazario, line 187: Tbh, this kind of tuple would have probably been easier as an object for naming purposes and readability.  Otherwise as a user you have to keep read the code later to figure out which is which.

davidiw: Hmm... maybe, not sure what we'd call it. Should look at what you guys do in the TS version. I suspect you probably follow the Rust implementation and just store more fields. Not sure that's much better. Typing is a huge win here if you use it right.

ecosystem/python/sdk/aptos_sdk/secp256k1_ecdsa.py
davidiw: ahh... LENGTH_WITH_PREFIX is probably a better phrasing

ecosystem/python/sdk/aptos_sdk/transactions.py
davidiw: hmm... zero :)

ecosystem/python/sdk/examples/fee_payer_transfer_coin.py
davidiw: hmm... I had been pondering that, but it gets into the nuances of handling different exceptions, which I didn't want to commit to in the SDK just yet.

ecosystem/python/sdk/examples/fee_payer_transfer_coin.py
```diff
@@ -0,0 +1,108 @@
...
+
+    print("\n=== Initial Data ===")
+    # :!:>section_4
+    try:
+        alice_sequence_number = await rest_client.account_sequence_number(
+            alice.address()
+        )
+        raise Exception("This should never be accessed")
+    except ApiError:
+        alice_sequence_number = 0
```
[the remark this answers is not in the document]
davidiw: yeah, this is an interesting one especially since for sponsored txns, defaulting to 0 is fine

PR #10522, opened 2023-10-16
[Aptos Node] Bump aptos-node version.

JoshLind:
### Description
This PR bumps the `aptos-node` crate version to `1.7` given that this is the latest release.

### Test Plan
Existing test infrastructure.

### DIFF
--- a/aptos-node/Cargo.toml
+++ b/aptos-node/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "aptos-node"
 description = "Aptos node"
-version = "1.6.0"
+version = "1.7.0"
 
 # Workspace inherited keys
 authors = { workspace = true }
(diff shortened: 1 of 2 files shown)


aptos-node/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "aptos-node"
 description = "Aptos node"
-version = "1.6.0"
+version = "1.7.0"
```
ibalajiarun, line 4: We don't bump patch version?

JoshLind: No -- we've typically only done major release versions. But if we can automate and/or test this, minor versions make sense to me 😄

brianolson:
I wish there was a build script or test that ensured the version agreed in
```
aptos-move/aptos-release-builder/data/release.yaml
aptos-node/Cargo.toml
```

JoshLind:
> I wish there was a build script or test that ensured the version agreed in

Yeah, I think someone is going to need to tackle this in a more holistic way (in the future). But, I won't be doing that in this PR 😄

brianolson (approved):

PR #10523, opened 2023-10-16
[forge] adjust thresholds for long-running realistic max, and haproxy (cpu)

bchocho:
### Description

Adjusting thresholds based on behavior after landing #10315


testsuite/forge-cli/src/main.rs
```diff
@@ -1693,23 +1693,26 @@ fn realistic_env_max_load_test(
                     (duration.as_secs() / 10).max(60),
                 )
                 .add_system_metrics_threshold(SystemMetricsThreshold::new(
-                    // Check that we don't use more than 14 CPU cores for 30% of the time.
-                    MetricsThreshold::new(14.0, max_cpu_threshold),
+                    // Check that we don't use more than 16 CPU cores for 30% of the time.
```
grao1991, line 1696: We probably also want to adjust the % of time here.

grao1991 (approved):

testsuite/forge-cli/src/main.rs
bchocho: Why adjust the %? Do you have a suggested %?

testsuite/forge-cli/src/main.rs
```diff
@@ -934,7 +934,7 @@ fn realistic_env_workload_sweep_test() -> ForgeConfig {
         criteria: [
             (3700, 0.35, 0.5, 0.8, 0.65),
             (2800, 0.35, 0.5, 1.2, 1.3),
-            (1800, 0.35, 0.5, 1.5, 2.7),
+            (1800, 0.35, 2.0, 1.5, 2.7),
```
bchocho, line 937: Since QS backpressure is less aggressive, the queue builds up here.

igor-aptos (approved):

PR #10525, opened 2023-10-16
[trivial][agg_v2] Generate IDs starting with random large number

igor-aptos:
to easily detect issues with exchange

### DIFF
--- a/aptos-move/aptos-aggregator/src/tests/types.rs
+++ b/aptos-move/aptos-aggregator/src/tests/types.rs
@@ -29,7 +29,8 @@ impl Default for FakeAggregatorView {
         Self {
             v1_store: HashMap::new(),
             v2_store: HashMap::new(),
-            counter: RefCell::new(0),
+            // Put some recognizable number, to easily spot missed exchanges
+            counter: RefCell::new(87654321),
         }
     }
 }
--- a/aptos-move/aptos-vm/src/natives.rs
+++ b/aptos-move/aptos-vm/src/natives.rs
@@ -48,7 +48,8 @@ struct AptosBlankStorage {
 impl AptosBlankStorage {
     pub fn new() -> Self {
         Self {
-            counter: AtomicU32::new(0),
+            // Put some recognizable number, to easily spot missed exchanges
+            counter: AtomicU32::new(55551111),
         }
     }
 }
--- a/aptos-move/block-executor/src/executor.rs
+++ b/aptos-move/block-executor/src/executor.rs
@@ -42,6 +42,7 @@ use bytes::Bytes;
 use claims::assert_none;
 use move_core_types::value::MoveTypeLayout;
 use num_cpus;
+use rand::{thread_rng, Rng};
 use rayon::ThreadPool;
 use std::{
     cell::RefCell,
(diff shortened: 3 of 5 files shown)


aptos-move/aptos-aggregator/src/tests/types.rs
```diff
@@ -31,7 +31,8 @@ impl Default for FakeAggregatorView {
         Self {
             v1_store: HashMap::new(),
             v2_store: HashMap::new(),
-            counter: RefCell::new(0),
+            // Put some recognizable number, to easily spot missed exchanges
+            counter: RefCell::new(87654321),
```
georgemitenkov, line 35: using something that ends in 0s might be more helpful to track down on which value the exchange was incorrect?

aptos-move/aptos-vm/src/natives.rs
```diff
@@ -54,7 +54,8 @@ struct AptosBlankStorage {
 impl AptosBlankStorage {
     pub fn new() -> Self {
         Self {
-            counter: AtomicU32::new(0),
+            // Put some recognizable number, to easily spot missed exchanges
+            counter: AtomicU32::new(55551111),
```
georgemitenkov, line 58: Same here, just so that we don't have to subtract 11, although doesn't mater too much

aptos-move/block-executor/src/executor.rs
```diff
@@ -943,7 +944,11 @@ where
         assert!(self.concurrency_level > 1, "Must use sequential execution");
 
         let versioned_cache = MVHashMap::new();
-        let shared_counter = AtomicU32::new(0);
+        // IDs are ephemeral. Pick a random prefix, and different each time,
+        // in case exchange is mistakenly not performed - to more easily catch it.
+        // And in a bad case where it happens in prod, to and make sure incorrect
+        // block doesn't get committed, but chain halts.
+        let shared_counter = AtomicU32::new(thread_rng().gen_range(1, 1000) * 1_000_000);
```
georgemitenkov, line 951: nit (optional): it might be nicer to factor it out

georgemitenkov (approved):

aptos-move/aptos-aggregator/src/tests/types.rs
igor-aptos: I am doing that for actual prod (ends with 1000000), for these fake I put something distinct from that

aptos-move/block-executor/src/view.rs
```diff
@@ -979,15 +979,7 @@ impl<'a, T: Transaction, S: TStateView<Key = T::Key>, X: Executable>
...
-        match &self.latest_view.latest_view {
-            ViewState::Sync(state) => (state.counter.fetch_add(1, Ordering::SeqCst) as u64).into(),
-            ViewState::Unsync(state) => {
-                let mut counter = state.counter.borrow_mut();
-                let id = (*counter as u64).into();
-                *counter += 1;
-                id
-            },
-        }
+        self.latest_view.generate_delayed_field_id()
```
vusirikala, line 982: T::Identifier is a DelayedFieldID. 
`generate_delayed_field_id()` generates u32. Does the type casting work?

igor-aptos: on line 944 above is the function:

    fn generate_delayed_field_id(&self) -> Self::IdentifierV2 {

so no casting needed

igor-aptos: I am calling latest_view, not latest_view.latest_view

PR #10526, opened 2023-10-16
[Bug] Fix issue 10468

fEst1ck:
### Description
Move unit tests continue to run in the presence of compiler warnings, but the warnings are shown.
### Test Plan
Existing tests &
```
MOVE_COMPILER_WARN_OF_DEPRECATION_USE=1 WARN_OF_DEPRECATION_USE_IN_APTOS_LIBS=1 cargo test -p aptos-framework --test move_unit_test -- --skip prover
```

### DIFF
--- a/third_party/move/move-compiler/src/diagnostics/mod.rs
+++ b/third_party/move/move-compiler/src/diagnostics/mod.rs
@@ -23,7 +23,7 @@ use move_symbol_pool::Symbol;
 use std::{
     collections::{BTreeMap, HashMap, HashSet},
     iter::FromIterator,
-    ops::Range, intrinsics::unreachable,
+    ops::Range,
 };
 
 //**************************************************************************************************
--- a/third_party/move/move-compiler/src/diagnostics/mod.rs
+++ b/third_party/move/move-compiler/src/diagnostics/mod.rs
@@ -61,6 +61,15 @@ pub fn report_diagnostics(files: &FilesSourceText, diags: Diagnostics) -> ! {
     std::process::exit(1)
 }
 
+// report diagnostics, but do not exit if diags are all warnings
+pub fn report_diagnostics_maynot_exit(files: &FilesSourceText, diags: Diagnostics) {
+    let should_exit = diags
+        .diagnostics
+        .iter()
+        .any(|diag| diag.info.severity() > Severity::Warning);
+    report_diagnostics_impl(files, diags, should_exit);
+}
+
 pub fn report_warnings(files: &FilesSourceText, warnings: Diagnostics) {
     if warnings.is_empty() {
         return;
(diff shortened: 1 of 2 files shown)


third_party/move/move-compiler/src/diagnostics/mod.rs
```diff
@@ -61,6 +61,15 @@ pub fn report_diagnostics(files: &FilesSourceText, diags: Diagnostics) -> ! {
     std::process::exit(1)
 }
 
+// report diagnostics, but do not exit if diags are all warnings
+pub fn report_diagnostics_maynot_exit(files: &FilesSourceText, diags: Diagnostics) {
```
brmataptos, line 65: Naming this `report_diagnostics_exit_on_error` might be better.

brmataptos:
Please test by running aptos-framework tests with environment variables MOVE_COMPILER_WARN_OF_DEPRECATION_USE=1 and WARN_OF_DEPRECATION_USE_IN_APTOS_LIBS=1
set, as this will trigger warnings building aptos libs.

fEst1ck:
> Please test by running aptos-framework tests with environment variables MOVE_COMPILER_WARN_OF_DEPRECATION_USE=1 and WARN_OF_DEPRECATION_USE_IN_APTOS_LIBS=1 set, as this will trigger warnings building aptos libs.

Yes, I did.

brmataptos (approved):

gregnazario (approved):

gregnazario:
Please change the commit message to match the contributing guidelines.

Something like:
[compiler] Allow unit testing with warnings

gregnazario (changes requested):
Can we additionally add a direct unit test for this with the CLI so it doesn't regress?

wrwg (commented):
Nice. As @gregnazario suggested, please add a unit test

brmataptos:
@gregnazario @wrwg this seems hard to write a test for, since the problem only shows up if the warning is in the stdlib (not an explicit dependency).  We'd have to add a lot of mechanism to inject a warning into warning code.  Perhaps Zekun can merge it without a test?

gregnazario:
> @gregnazario @wrwg this seems hard to write a test for, since the problem only shows up if the warning is in the stdlib (not an explicit dependency).  We'd have to add a lot of mechanism to inject a warning into warning code.  Perhaps Zekun can merge it without a test?

Does this not also happen if you have a warning in your test?

brmataptos:
> > @gregnazario @wrwg this seems hard to write a test for, since the problem only shows up if the warning is in the stdlib (not an explicit dependency).  We'd have to add a lot of mechanism to inject a warning into warning code.  Perhaps Zekun can merge it without a test?
> 
> Does this not also happen if you have a warning in your test?

No.  The flow is special for the stdlib.  That's the problem.

gregnazario (approved):

Issue #10529, opened 2023-10-16
[Event v2] Framework Migration 


lightmark:
# 🚀 Feature Request

## Motivation

As module events (event v2) is alive, we have to start the migration procedure.
1. Double Emitting all v1 events in aptos-framework to v2 events
2.  Replace all v1 events in move-examples to v2.
3. Change Consensus code to query v2 events instead.
4. Docs

brmataptos:
You should also deprecate v1 and say how to move to v2.

PR #10532, opened 2023-10-16
[event v2] double emitting in account, coin, fungible asset and object and migrate move-examples

lightmark:
### Description
step 1 of #10529 
leave other modules for grabs.

### Test Plan
ut

### DIFF
--- a/aptos-move/framework/aptos-framework/sources/account.move
+++ b/aptos-move/framework/aptos-framework/sources/account.move
@@ -23,6 +23,13 @@ module aptos_framework::account {
     friend aptos_framework::resource_account;
     friend aptos_framework::transaction_validation;
 
+    #[event]
+    struct KeyRotation has drop, store {
+        account: address,
+        old_authentication_key: vector<u8>,
+        new_authentication_key: vector<u8>,
+    }
+
     /// Resource representing an account.
     struct Account has key, store {
         authentication_key: vector<u8>,
@@ -593,6 +600,11 @@ module aptos_framework::account {
         let new_auth_key = from_bcs::to_address(new_auth_key_vector);
         table::add(address_map, new_auth_key, originating_addr);
 
+        event::emit<KeyRotation>(KeyRotation {
+            account: originating_addr,
+            old_authentication_key: account_resource.authentication_key,
+            new_authentication_key: new_auth_key_vector,
+        });
         event::emit_event<KeyRotationEvent>(
             &mut account_resource.key_rotation_events,
             KeyRotationEvent {
--- a/aptos-move/framework/aptos-framework/sources/aptos_account.move
+++ b/aptos-move/framework/aptos-framework/sources/aptos_account.move
@@ -3,7 +3,7 @@ module aptos_framework::aptos_account {
     use aptos_framework::aptos_coin::AptosCoin;
     use aptos_framework::coin::{Self, Coin};
     use aptos_framework::create_signer::create_signer;
-    use aptos_framework::event::{EventHandle, emit_event};
+    use aptos_framework::event::{EventHandle, emit_event, emit};
     use std::error;
     use std::signer;
     use std::vector;
@@ -32,6 +32,12 @@ module aptos_framework::aptos_account {
 
     /// Event emitted when an account's direct coins transfer config is updated.
     struct DirectCoinTransferConfigUpdatedEvent has drop, store {
+        new_allow_direct_transfers: bool,
+    }
+
+    #[event]
+    struct AllowDirectTransfers has drop, store {
+        account: address,
         new_allow_direct_transfers: bool,
     }
 
@@ -134,6 +140,8 @@ module aptos_framework::aptos_account {
             };
 
             direct_transfer_config.allow_arbitrary_coin_transfers = allow;
+            emit(
+                AllowDirectTransfers { account: addr, new_allow_direct_transfers: allow });
             emit_event(
                 &mut direct_transfer_config.update_coin_transfer_events,
                 DirectCoinTransferConfigUpdatedEvent { new_allow_direct_transfers: allow });
@@ -142,6 +150,8 @@ module aptos_framework::aptos_account {
                 allow_arbitrary_coin_transfers: allow,
                 update_coin_transfer_events: new_event_handle<DirectCoinTransferConfigUpdatedEvent>(account),
             };
+            emit(
+                AllowDirectTransfers { account: addr, new_allow_direct_transfers: allow });
             emit_event(
                 &mut direct_transfer_config.update_coin_transfer_events,
                 DirectCoinTransferConfigUpdatedEvent { new_allow_direct_transfers: allow });
--- a/aptos-move/framework/aptos-framework/sources/coin.move
+++ b/aptos-move/framework/aptos-framework/sources/coin.move
@@ -124,8 +124,20 @@ module aptos_framework::coin {
         amount: u64,
     }
 
+    #[event]
+    struct Deposit<phantom CoinType> has drop, store {
+        account: address,
+        amount: u64,
+    }
+
     /// Event emitted when some amount of a coin is withdrawn from an account.
     struct WithdrawEvent has drop, store {
+        amount: u64,
+    }
+
+    #[event]
+    struct Withdraw<phantom CoinType> has drop, store {
+        account: address,
         amount: u64,
     }
 
@@ -196,7 +208,10 @@ module aptos_framework::coin {
     }
 
     /// Merges `coin` into aggregatable coin (`dst_coin`).
-    public(friend) fun merge_aggregatable_coin<CoinType>(dst_coin: &mut AggregatableCoin<CoinType>, coin: Coin<CoinType>) {
+    public(friend) fun merge_aggregatable_coin<CoinType>(
+        dst_coin: &mut AggregatableCoin<CoinType>,
+        coin: Coin<CoinType>
+    ) {
         spec {
             update supply<CoinType> = supply<CoinType> - coin.value;
         };
@@ -253,8 +268,8 @@ module aptos_framework::coin {
     #[view]
     /// Returns `true` is account_addr has frozen the CoinStore or if it's not registered at all
     public fun is_coin_store_frozen<CoinType>(account_addr: address): bool acquires CoinStore {
-        if(!is_account_registered<CoinType>(account_addr)) {
-          return true
+        if (!is_account_registered<CoinType>(account_addr)) {
+            return true
         };
 
         let coin_store = borrow_global<CoinStore<CoinType>>(account_addr);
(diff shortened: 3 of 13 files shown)


aptos-move/framework/aptos-framework/sources/object.move
```diff
@@ -157,6 +157,7 @@ module aptos_framework::object {
     }
 
     /// Emitted whenever the object's owner field is changed.
+    #[event]
```
lightmark, line 160: note: not sure reusing this struct or create a new one. seems unnecessary to me but a different opinion is welcomed.

aptos-move/move-examples/defi/sources/locked_coins.move
```diff
@@ -38,35 +37,38 @@ module defi::locked_coins {
...
-        cancel_lockup_events: EventHandle<CancelLockupEvent>,
-        claim_events: EventHandle<ClaimEvent>,
-        update_lockup_events: EventHandle<UpdateLockupEvent>,
-        update_withdrawal_address_events: EventHandle<UpdateWithdrawalAddressEvent>,
     }
 
+    #[event]
     /// Event emitted when a lock is canceled.
-    struct CancelLockupEvent has drop, store {
+    struct CancelLockup has drop, store {
```
junkil-park, line 44: Is it going to be the convention to drop the suffix "Event"?

lightmark: My current pattern for aptos-framework is:
1. If the current event struct doesn't need add new field to migrate to v2, I will just add `#[event]` attribute directly to that struct to spare the hassle of creating a new struct.
2. otherwise, create a new struct without `Event` suffix since attribute can tell whether it is an event or not and you cannot share the same event name in the same module.

Does it make sense?

junkil-park: Make sense!
What if the current event v1 struct (for example AbcEvent) needs to add a new field? It will become AbcEventV2 or AbcV2?

lightmark: there is no way for them to add field. I think any new v2 does not need `Event` string in their name.

aptos-move/move-examples/token_objects/ambassador/sources/ambassador.move
```diff
@@ -68,8 +66,10 @@ module ambassador::ambassador {
         ambassador_level: u64,
     }
 
+    #[event]
     /// The ambassador level update event
-    struct LevelUpdateEvent has drop, store {
+    struct LevelUpdate has drop, store {
+        token: address,
```
movekevin, line 72: This can just be the object reference?

lightmark: I updated this in another PR. #10657

aptos-move/framework/aptos-framework/sources/account.move
```diff
@@ -23,6 +23,19 @@ module aptos_framework::account {
...
 
+    #[event]
+    struct KeyRotation has drop, store {
+        account: address,
+        old_authentication_key: vector<u8>,
+        new_authentication_key: vector<u8>,
+    }
+
+    #[event]
+    struct CoinRegister has drop, store {
```
movekevin, line 34: I wonder if we just leave this one out since registration is no longer a required step

aptos-move/framework/aptos-framework/sources/account.move
```diff
@@ -593,6 +606,11 @@ module aptos_framework::account {
         let new_auth_key = from_bcs::to_address(new_auth_key_vector);
         table::add(address_map, new_auth_key, originating_addr);
 
+        event::emit<KeyRotation>(KeyRotation {
```
movekevin, line 609: <KeyRotation> is not necessary

lightmark: why?

aptos-move/framework/aptos-framework/sources/coin.move
```diff
@@ -356,11 +371,11 @@ module aptos_framework::coin {
             !coin_store.frozen,
             error::permission_denied(EFROZEN),
         );
-
         event::emit_event<DepositEvent>(
             &mut coin_store.deposit_events,
             DepositEvent { amount: coin.value },
         );
+        event::emit<Deposit>(Deposit { account: account_addr, amount: coin.value });
```
movekevin, line 378: <Deposit> is not necessary

aptos-move/framework/aptos-framework/sources/fungible_asset.move
```diff
@@ -444,6 +462,7 @@ module aptos_framework::fungible_asset {
         borrow_global_mut<FungibleStore>(store_addr).frozen = frozen;
 
         let events = borrow_global_mut<FungibleAssetEvents>(store_addr);
+        event::emit<Frozen>(Frozen { store: store_addr, frozen });
```
movekevin, line 465: Same here and below

movekevin (approved):

aptos-move/framework/aptos-framework/sources/account.move
lightmark: if nobody uses it, we could omit it. But hard to get feedbacks.
btw, now my question is it seems this PR breaks a lot of tests...

aptos-move/framework/aptos-framework/sources/account.move
movekevin: < KeyRotation >

aptos-move/framework/aptos-framework/sources/coin.move
movekevin: < Deposit >

api/goldens/aptos_api__tests__transactions_test__test_get_transactions_output_user_transaction_with_entry_function_payload.json
```diff
@@ -114,7 +114,7 @@
     "state_change_hash": "",
     "event_root_hash": "",
     "state_checkpoint_hash": null,
-    "gas_used": "6",
+    "gas_used": "8",
```
junkil-park, line 117: A noob question: Event V2 (module event) increase gas usage? If so, what is the cause?

aptos-move/framework/aptos-framework/sources/account.move
junkil-park: > why?

It's because the type can be inferred from the type of the message.

api/goldens/aptos_api__tests__transactions_test__test_get_transactions_output_user_transaction_with_entry_function_payload.json
lightmark: additional i think

aptos-move/framework/aptos-framework/sources/coin.move
```diff
@@ -124,11 +124,23 @@ module aptos_framework::coin {
         amount: u64,
     }
 
+    #[event]
+    struct Deposit has drop, store {
```
junkil-park, line 128: Would `Deposit` and `Withdraw` need the `CoinType` parameter? Otherwise, one can't tell the coin type from the `Deposit` (`Withdraw`) events? (cc: @movekevin )

lightmark: I think you are right.

aptos-move/framework/aptos-framework/sources/account.move
xbtmatt: It breaks because of this? Or something else

I say leave it out if it doesn't break anything, it is less confusing in the long term

lightmark: not this one. you mean keeping it?

aptos-move/framework/aptos-framework/sources/coin.move
```diff
@@ -124,11 +125,25 @@ module aptos_framework::coin {
...
         amount: u64,
     }
 
+    #[event]
+    struct Withdraw has drop, store {
+        account: address,
+        coin_type: TypeInfo,
+        amount: u64,
+    }
+
```
xbtmatt, line 146: Do we have a preference for how we emit events with generics or not? I was wondering this yesterday while using it.

Like how come not:

```rust
struct Withdraw<phantom CoinType> has drop, store {
   account: address,
   amount: u64,
}
```

aptos-move/framework/aptos-framework/sources/account.move
xbtmatt: Yeah, I meant: does this change right here break a lot of tests? Or did you mean in general this double emission PR break a lot of tests? Haha

And then:

I think we should leave this:
```rust
#[event]
struct CoinRegister has drop, store {
 ...
}
```
out, because we don't recommend people use the register flow anymore

aptos-move/framework/aptos-framework/sources/coin.move
lightmark: I am dumb. I think you are absolutely right.

aptos-move/framework/aptos-framework/sources/aptos_account.move
```diff
@@ -35,6 +35,12 @@ module aptos_framework::aptos_account {
         new_allow_direct_transfers: bool,
     }
 
+    #[event]
+    struct DirectCoinTransferConfigUpdated has drop, store {
+        account: address,
+        new_allow_direct_transfers: bool,
```
davidiw, line 41: yucky name here... why not just "allow_direct_transfers"?

aptos-move/framework/aptos-framework/sources/object.move
davidiw: go for it...

davidiw (approved):

aptos-move/framework/aptos-framework/sources/aptos_account.move
```diff
@@ -3,7 +3,7 @@ module aptos_framework::aptos_account {
     use aptos_framework::aptos_coin::AptosCoin;
     use aptos_framework::coin::{Self, Coin};
     use aptos_framework::create_signer::create_signer;
-    use aptos_framework::event::{EventHandle, emit_event};
+    use aptos_framework::event::{EventHandle, emit_event, emit};
```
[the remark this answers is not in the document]
lightmark: ... I don't know why at the first place...

aptos-move/framework/aptos-framework/sources/fungible_asset.move
```diff
@@ -111,13 +111,6 @@ module aptos_framework::fungible_asset {
         frozen: bool,
     }
 
-    #[resource_group_member(group = aptos_framework::object::ObjectGroup)]
-    struct FungibleAssetEvents has key {
```
movekevin, line 115: This would violate the upgrade policy checks

lightmark: I put it at the bottom.

PR #10534, opened 2023-10-16
[AdminService] Add an admin service to support some profiling functionalities on a running node.

grao1991:

### DIFF
--- a/config/src/config/admin_service_config.rs
+++ b/config/src/config/admin_service_config.rs
@@ -1,0 +1,64 @@
+// Copyright © Aptos Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::{
+    config::{
+        config_optimizer::ConfigOptimizer, config_sanitizer::ConfigSanitizer,
+        node_config_loader::NodeType, Error, NodeConfig,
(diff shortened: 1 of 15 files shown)


msmouse (approved):

config/src/config/admin_service_config.rs
```diff
@@ -0,0 +1,64 @@
...
+    utils,
+};
+use aptos_types::chain_id::ChainId;
+use serde::{Deserialize, Serialize};
+use serde_yaml::Value;
+
+#[derive(Clone, Debug, Deserialize, PartialEq, Eq, Serialize)]
+#[serde(default, deny_unknown_fields)]
+pub struct AdminServiceConfig {
+    pub enabled: Option<bool>,
```
zekun000, line 18: why this is Option of bool?

zekun000 (approved):

config/src/config/admin_service_config.rs
grao1991: Want the Optimizer to set some default values based on chainid if user is not explicitly set it, and I'm too lazy to check the yaml file directly.

PR #10537, opened 2023-10-17
[Agg v2] Add MoveTypeLayout to resource group change set

vusirikala:
### Description
Add MoveTypeLayout to resource group change set in VMChangeSet.

### DIFF
--- a/aptos-move/aptos-vm-types/src/change_set.rs
+++ b/aptos-move/aptos-vm-types/src/change_set.rs
@@ -48,7 +48,7 @@ pub struct GroupWrite {
     /// exist in the group. Note: During parallel block execution, due to speculative
     /// reads, this invariant may be violated (and lead to speculation error if observed)
     /// but guaranteed to fail validation and lead to correct re-execution in that case.
-    inner_ops: HashMap<StructTag, WriteOp>,
+    inner_ops: HashMap<StructTag, (WriteOp, Option<Arc<MoveTypeLayout>>)>,
 }
 
 impl GroupWrite {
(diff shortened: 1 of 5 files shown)


aptos-move/aptos-vm/src/move_vm_ext/respawned_session.rs
```diff
@@ -254,7 +254,8 @@ impl<'r> TResourceGroupView for ExecutorViewWithChangeSet<'r> {
         resource_tag: &Self::ResourceTag,
         maybe_layout: Option<&Self::Layout>,
     ) -> anyhow::Result<Option<Bytes>> {
-        if let Some(write_op) = self
+        // TODO: resource_group_write_set also contains a layout. What to do with it?
+        if let Some((write_op, _layout)) = self
```
igor-aptos, line 258: seems like on line 233 we also ignore the layout. 

I assume at this point, it is required to be the same layout. maybe we should just have a function expect_layouts_match(maybe_layout_1, maybe_layout_2)

which we call here and there, and then inside we add a TODO of what exactly to do, but we can start with asserting that both are either None, or both are Some. 

we can later also do something like - randomly for 1% of calls, we check fully. when we run tests, we can locally change to always compare, etc. but there will be only a single place to change it. you can put that function in aptos-move/aptos-aggregator/src/types.rs next to expect_ok, or if you see a better place elsewhere

vusirikala: Why should the layout in resource_group_write_set, and the layout used by read operations be the same? If the current transaction has updated the layout, then the new layout will be present in the resource_group_write_set. Will the read operation use the new layout or the old layout?

aptos-move/aptos-vm-types/src/change_set.rs
```diff
@@ -58,9 +58,9 @@ impl GroupWrite {
     pub fn new(
         mut metadata_op: WriteOp,
         group_size: u64,
-        inner_ops: HashMap<StructTag, WriteOp>,
+        inner_ops: HashMap<StructTag, (WriteOp, Option<Arc<MoveTypeLayout>>)>,
```
ibalajiarun, line 61: I suggest either cherrypicking the sev changes or merging main to aggregator branch so we don't accidentally merge change it back to hasmap.

PR #10538, opened 2023-10-17
[doc] remove useless tags to make the doc render properly

flex-enjoy:
### Description
Some useless `span` tags make the markdown render failed under `Docusaurus v2.4.1`
, need remove them


davidiw (changes requested):
unfortunately these are used for redirects, so we would need to find some other means of placing an anchor.

thanks for the improvement though, it does look better....

PR #10543, opened 2023-10-17
[agg_v2] Fix mistakenly overriding on concurrent set_base_value

igor-aptos:

### DIFF
--- a/aptos-move/mvhashmap/src/versioned_data.rs
+++ b/aptos-move/mvhashmap/src/versioned_data.rs
@@ -10,7 +10,12 @@ use claims::assert_some;
 use crossbeam::utils::CachePadded;
 use dashmap::DashMap;
 use move_core_types::value::MoveTypeLayout;
-use std::{collections::btree_map::BTreeMap, fmt::Debug, hash::Hash, sync::Arc};
+use std::{
+    collections::btree_map::{self, BTreeMap},
+    fmt::Debug,
+    hash::Hash,
+    sync::Arc,
+};
 
 /// Every entry in shared multi-version data-structure has an "estimate" flag
 /// and some content.
@@ -255,25 +260,49 @@ impl<K: Hash + Clone + Debug + Eq, V: TransactionWrite> VersionedData<K, V> {
             .unwrap_or(Err(MVDataError::Uninitialized))
     }
 
-    pub fn provide_base_value(&self, key: K, data: V, maybe_layout: Option<Arc<MoveTypeLayout>>) {
+    pub fn set_base_value(&self, key: K, data: V, maybe_layout: Option<Arc<MoveTypeLayout>>) {
         let mut v = self.values.entry(key).or_default();
         let bytes_len = data.bytes_len();
-        // For base value, incarnation is irrelevant, set to 0.
-        let prev_entry = v.versioned_map.insert(
-            ShiftedTxnIndex::zero(),
-            CachePadded::new(Entry::new_write_from(0, data, maybe_layout)),
-        );
+        // For base value, incarnation is irrelevant, and is always set to 0.
 
-        assert!(prev_entry.map_or(true, |entry| -> bool {
-            if let EntryCell::Write(i, v, _) = &entry.cell {
-                // base value may have already been provided due to a concurrency race,
-                // but it has to be the same as being set.
-                // Assert the length of bytes for efficiency (instead of full equality)
-                *i == 0 && v.bytes_len() == bytes_len
-            } else {
-                true
-            }
-        }));
+        use btree_map::Entry::*;
+        match v.versioned_map.entry(ShiftedTxnIndex::zero()) {
+            Vacant(v) => {
+                v.insert(CachePadded::new(Entry::new_write_from(
+                    0,
+                    data,
+                    maybe_layout,
+                )));
+            },
+            Occupied(o) => {
+                assert!(
+                    if let EntryCell::Write(i, v, layout) = &o.get().cell {
+                        // base value may have already been provided by another transaction
+                        // executed simultaneously and asking for the same resource.
+                        // Value from storage must be identical, but then delayed field
+                        // identifier exchange could've modiefied it.
+                        //
+                        // If maybe_layout is None, they are required to be identical
+                        // If maybe_layout is Some, there might have been an exchange
+                        // Assert the length of bytes for efficiency (instead of full equality)
+                        layout.is_some() == maybe_layout.is_some()
+                            && *i == 0
+                            && (layout.is_some() || v.bytes_len() == bytes_len)
+                    } else {
+                        true
+                    }
+                );
+            },
+        };
+
+        // let prev_entry = v.versioned_map.insert(
+        //     ShiftedTxnIndex::zero(),
+        //     CachePadded::new(Entry::new_write_from(0, data, maybe_layout)),
+        // );
+
+        // assert!(prev_entry.map_or(true, |entry| -> bool {
+
+        // }));
     }
 
     /// Versioned write of data at a given key (and version).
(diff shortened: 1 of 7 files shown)


aptos-move/mvhashmap/src/versioned_group_data.rs
```diff
@@ -77,6 +77,30 @@ impl<T: Hash + Clone + Debug + Eq + Serialize, V: TransactionWrite> Default
...
+        &mut self,
+        shifted_idx: ShiftedTxnIndex,
+        incarnation: Incarnation,
+        // TODO add layout to values
+        values: impl IntoIterator<Item = (T, V)>,
+    ) {
+        match self.idx_to_update.get(&shifted_idx) {
+            Some(previous) => {
+                // base value may have already been provided due to a concurrency race.
+                // If maybe_layout is None, they are required to be identical.
```
georgemitenkov, line 90: The comment is wrong / old / stale?

aptos-move/mvhashmap/src/versioned_group_data.rs
```diff
@@ -77,6 +77,30 @@ impl<T: Hash + Clone + Debug + Eq + Serialize, V: TransactionWrite> Default
...
+                // If maybe_layout is Some, there might have been an exchange
+                // Assert the length of bytes for efficiency (instead of full equality)
+                for (tag, v) in values.into_iter() {
+                    let prev_v = previous
+                        .get(&tag)
+                        .expect("Reading twice from storage must be consistent");
+                    assert!(v.bytes_len() == prev_v.bytes_len());
+                }
+            },
+            None => self.write(shifted_idx, incarnation, values),
```
georgemitenkov, line 100: What is the problem with overwriting concurrently the base values? They should be identical?

aptos-move/mvhashmap/src/versioned_data.rs
```diff
@@ -259,21 +264,41 @@ impl<K: Hash + Clone + Debug + Eq, V: TransactionWrite> VersionedData<K, V> {
...
-                // base value may have already been provided due to a concurrency race,
-                // but it has to be the same as being set.
-                // Assert the length of bytes for efficiency (instead of full equality)
-                *i == 0 && v.bytes_len() == bytes_len
-            } else {
-                true
-            }
-        }));
+        use btree_map::Entry::*;
+        match v.versioned_map.entry(ShiftedTxnIndex::zero()) {
```
georgemitenkov, line 269: Why match over old code? Seems like layout is the only thing that got added?

igor-aptos: we cannot do insert - that overwrites existing value. we need to do insert only if Vacant.

https://aptos-org.slack.com/archives/C057VPCD7QX/p1697520278313169

aptos-move/mvhashmap/src/versioned_group_data.rs
igor-aptos: it is future looking! 
once we add maybe layout (TODO above), that is what check should be doing :D

aptos-move/mvhashmap/src/versioned_group_data.rs
igor-aptos: not if there were aggregators, and they got exchanged with ephemeral IDs

georgemitenkov (approved):

aptos-move/mvhashmap/src/versioned_group_data.rs
```diff
@@ -77,6 +77,30 @@ impl<T: Hash + Clone + Debug + Eq + Serialize, V: TransactionWrite> Default
 }
 
 impl<T: Hash + Clone + Debug + Eq + Serialize, V: TransactionWrite> VersionedGroupValue<T, V> {
+    fn set_base_values(
+        &mut self,
+        shifted_idx: ShiftedTxnIndex,
```
vusirikala, line 82: What does shifted transaction index mean here? Why is transaction index shifted?

igor-aptos: we need to distinguish storage value from all other transactions.

So shifted means 0 is storage, and i>0 means transaction i-1

I forgot the reason why we didn't do Option<TxnIndex> instead though (I mentioned but @gelash had a reason)

aptos-move/mvhashmap/src/versioned_data.rs
```diff
@@ -259,21 +264,41 @@ impl<K: Hash + Clone + Debug + Eq, V: TransactionWrite> VersionedData<K, V> {
...
+            Vacant(v) => {
+                v.insert(CachePadded::new(Entry::new_write_from(
+                    0,
```
vusirikala, line 272: Is it okay to hardcode the incarnation to zero, even though this function might be called with higher incarnations?

aptos-move/mvhashmap/src/versioned_data.rs
```diff
@@ -259,21 +264,41 @@ impl<K: Hash + Clone + Debug + Eq, V: TransactionWrite> VersionedData<K, V> {
...
+                    data,
+                    maybe_layout,
+                )));
+            },
+            Occupied(o) => {
+                assert!(
+                    if let EntryCell::Write(i, v, layout) = &o.get().cell {
+                        // base value may have already been provided due to a concurrency race.
```
vusirikala, line 280: What is concurrency race?

aptos-move/mvhashmap/src/versioned_data.rs
```diff
@@ -259,21 +264,41 @@ impl<K: Hash + Clone + Debug + Eq, V: TransactionWrite> VersionedData<K, V> {
...
+                        // If maybe_layout is None, they are required to be identical.
+                        // If maybe_layout is Some, there might have been an exchange
+                        // Assert the length of bytes for efficiency (instead of full equality)
+                        layout.is_some() == maybe_layout.is_some()
+                            && *i == 0
+                            && (layout.is_some() || v.bytes_len() == bytes_len)
```
vusirikala, line 286: I didn't get it. We check that the length of the data matches only when layout is None. Why is that?

aptos-move/mvhashmap/src/versioned_data.rs
```diff
@@ -259,21 +264,41 @@ impl<K: Hash + Clone + Debug + Eq, V: TransactionWrite> VersionedData<K, V> {
         let mut v = self.values.entry(key).or_default();
```
vusirikala, line 264: I like the name `store_base_value` or `record_based_value` better than `provide_base_value`.

aptos-move/mvhashmap/src/versioned_group_data.rs
```diff
@@ -77,6 +77,30 @@ impl<T: Hash + Clone + Debug + Eq + Serialize, V: TransactionWrite> Default
...
+        incarnation: Incarnation,
```
vusirikala, line 83: The incarnation is given as input to `set_base_values` function here. But the incarnation is not given as input to `provide_base_value` function in versioned_data.rs. Any reason for that?

aptos-move/mvhashmap/src/versioned_group_data.rs
```diff
@@ -77,6 +77,30 @@ impl<T: Hash + Clone + Debug + Eq + Serialize, V: TransactionWrite> Default
...
+        // TODO add layout to values
+        values: impl IntoIterator<Item = (T, V)>,
+    ) {
+        match self.idx_to_update.get(&shifted_idx) {
+            Some(previous) => {
+                // base value may have already been provided due to a concurrency race.
```
vusirikala, line 89: Please also explain here what concurrency race means.

vusirikala (approved):

aptos-move/mvhashmap/src/versioned_data.rs
igor-aptos: "value" of the base value is always from version 0, and we will not re-fetch it, even if incarnation is newer, so fixed incarnation is correct here

aptos-move/mvhashmap/src/versioned_data.rs
igor-aptos: added more comments, and explanation here - https://aptos-org.slack.com/archives/C057VPCD7QX/p1697520278313169

PR #10547, opened 2023-10-17
[forge] adjust thresholds (again) for max load and workload sweep tests

bchocho:
### Description
* max load test: add 5% breach allowance for frequent epoch changes. It was observed that spikes only appear during epoch changes.
* workload sweep: bump up the order->commit. The numbers fluctuate, but experiments show that there is a large increase because the duration of the test increased.


bchocho:
ModifyGlobalResource order->commit is comfortably under 2 with the exact same commit, but with a shorter duration:
* Shorter duration: (two attempts: 1.2, 1.8) https://github.com/aptos-labs/aptos-core/actions/runs/6552483608/attempts/1
* Failed test (2.9): https://github.com/aptos-labs/aptos-core/actions/runs/6543137203/attempts/1#summary-17772043404

igor-aptos:
oooooh, on the short test it doesn't stabilize, it continues increasing. thanks so much for digging into this - this is an opportunity for the future to make things more predictable, but for now increasing the limit sounds good.

igor-aptos (approved):

ibalajiarun (approved):
LGTM

PR #10549, opened 2023-10-17
[move-vm] Fix compatibility issue

runtian-zhou:
### Description

The compatibility rule causes the loader paranoid check to break. Fix this issue.

### Test Plan

Added a replay.

We need to wait to see if this fix replay issues.


runtian-zhou:
https://github.com/aptos-labs/aptos-core/actions/runs/6551227889

wrwg (approved):

vgao1996 (approved):

PR #10552, opened 2023-10-17
[Aptos Data Client] Update peer selection to incorporate validator distance

JoshLind:
### Description
This PR updates the peer selection logic in the Aptos Data Client to leverage "validator distance" when selecting peers to request "new" data (i.e., data that has not yet been advertised).

Previously, we were selecting the lowest latency peer when fetching "new" data. However, it's possible to have low latency peers that are far away from the network (i.e., a large number of hops from the validator set) and thus unlikely to have new data. To avoid this, it makes sense to instead select the peer with the lowest distance from the validators as well as lowest latency. This should help to avoid close but not up-to-date peers.

The PR offers several commits:
1. Update peer selection to select the peer with the lowest distance from the validators and latency.
2. Update and add new unit tests.

### Test Plan
New and existing unit tests.


brianolson (approved):

state-sync/aptos-data-client/src/tests/priority.rs
```diff
@@ -1313,25 +1513,31 @@ fn enumerate_subscription_requests(known_version: u64, known_epoch: u64) -> Vec<
...
-    for peer in peers {
-        // Get the peer's latency
-        let ping_latency = utils::get_peer_ping_latency(mock_network, *peer);
 
-        // Update the lowest latency peer
-        if ping_latency < lowest_latency {
-            lowest_latency = ping_latency;
-            lowest_latency_peer = *peer;
+    // Identify the peer with the lowest distance and latency.
+    // Distance is prioritized over latency.
```
grao1991, line 1526: Distance might not be a good indicator, why don't we just maintain something like `verison` locally for peers?

JoshLind: This logic is for selecting peers for "new" data (i.e., data that has not been advertised yet). We do maintain a `version` for each peer (that's the advertised version), but that version will always lag the true version (e.g., we don't poll every peer all the time, and even if we did, it would always lag by the network latency, etc.). So we have these types of optimistic requests to fetch data before we see it. This is what we're optimizing for here.

What currently happens is we split peers into priority and regular peers (based on trusted peers, etc.), and then when selecting a peer to send an optimistic fetch, we select the priority peer with the lowest latency and distance from the set. If one doesn't exist, we then select the same for a regular peer 😄

grao1991: Sorry I didn't make it clear. What I was trying to say was: it's probably more accurate to infer the true version from the advertised version, than from the distance.

JoshLind: > Sorry I didn't make it clear. What I was trying to say was: it's probably more accurate to infer the true version from the advertised version, than from the distance.

Aah, how do you propose that? 😄 The reason for using this algorithm is because it: (i) optimizes for network throughput (e.g., lower latency -> higher bandwidth and quicker to get any new data); (ii) optimizes for network topology (i.e., the closer the node is to the validators, the sooner it should get the data). Note also that these types of "new" data requests are best effort. State sync will always fallback to fetching advertised data when it sees it.

grao1991: > (i.e., the closer the node is to the validators, the sooner it should get the data)

I'm not sure if I agree with this. Ideally it's probably should be `f(the lag X seconds ago, X, the latency to validator, the lag of the validator, ...)`.. I'd say `if a node has less lag in the past, it's more likely to have less lag now`

JoshLind: Hmm... I think that's something we'd need to explore in the future. It's not trivial to me to detect the lag probability of a peer and use that to predict the selection (although, it sounds promising 😄).

FWIW, the reason for landing this change is because it actually addresses a known situation we saw. A node operator actually deployed a VFN that was misconfigured (and had a distance from the validators of 3, instead of 1). So, optimistic fetches sent to the node were essentially no-ops. Hence the reason for this fix (i.e., if a PFN has VFNs peers that include a badly connected/configured VFN, it'll skip that in the selection process using this 😸 ).

JoshLind: > the closer the node is to the validators, the sooner it should get the data)

Also, if you look at the existing data today, this is overwhelmingly true (95%+) 😄

grao1991: I'm worrying it's statistically true just because for most of the time when the network is healthy, the distance is proportional to the latency. But when network is not healthy, let say I'm a PFN and I connect to 1 VFN and 3 PFN, and the VFN is not healthy (even slightly lagging), I will always choose it over other PFNs, which seems... not ideal.

grao1991 (approved):

state-sync/aptos-data-client/src/tests/priority.rs
JoshLind: Makes sense 😄 Note: the only time a PFN will connect to other PFNs is if seeds have been added which isn't the normal case (except for when bootstrapping testnet). We also have action items for the networking code to prevent seeds from overriding VFN connections (i.e., if I have 3 PFN seeds, I should be able to connect to them in addition to the regular 4 VFNs). We'll be looking into that soon, so hopefully it should give nodes more options.

PR #10554, opened 2023-10-17
[Storage] add epoch 0 ledger info to DB 

areshand:
### Description
without this ledger info, backup service would fail to backup since it cannot establish provenance to genesis

### Test Plan

1. start local node and check db content 
2. updated unit tests


grao1991 (approved):

storage/aptosdb/src/lib.rs
```diff
@@ -1217,9 +1217,30 @@ impl AptosDB {
...
+            .get_latest_ledger_info_option()
+            .map_or(0, |li| li.ledger_info().next_block_epoch());
+        ensure!(genesis_li.ledger_info().epoch() == current_epoch && current_epoch == 0);
+        self.ledger_store
+            .put_ledger_info(genesis_li, &ledger_batch)?;
+
+        ledger_batch.put::<DbMetadataSchema>(
+            &DbMetadataKey::LedgerCommitProgress,
+            &DbMetadataValue::Version(0),
+        )?;
```
msmouse, line 1236: I don't think we should do this? Other things in the ledger doesn't present.

storage/aptosdb/src/lib.rs
```diff
@@ -1217,9 +1217,30 @@ impl AptosDB {
         self.ledger_db.write_set_db().write_schemas(batch)
     }
 
+    pub fn commit_genesis_ledger_info(&self, genesis_li: &LedgerInfoWithSignatures) -> Result<()> {
+        let _timer = OTHER_TIMERS_SECONDS
+            .with_label_values(&["commit_genesis_ledger_info"])
+            .start_timer();
```
msmouse, line 1223: nit: I don't think this is worth measuring..

storage/aptosdb/src/lib.rs
```diff
@@ -2103,6 +2120,14 @@ impl DbWriter for AptosDB {
 
                 self.commit_ledger_info(last_version, new_root_hash, ledger_info_with_sigs)?;
 
+                // Record overall commit progress
+                let batch = SchemaBatch::new();
+                batch.put::<DbMetadataSchema>(
+                    &DbMetadataKey::OverallCommitProgress,
+                    &DbMetadataValue::Version(last_version),
+                )?;
+                self.ledger_db.metadata_db().write_schemas(batch)?;
```
msmouse, line 2129: why? You've already made a separate method to special case the genesis LI?

storage/aptosdb/src/lib.rs
```diff
@@ -1217,9 +1217,30 @@ impl AptosDB {
...
+        ensure!(genesis_li.ledger_info().epoch() == current_epoch && current_epoch == 0);
+        self.ledger_store
+            .put_ledger_info(genesis_li, &ledger_batch)?;
+
+        ledger_batch.put::<DbMetadataSchema>(
+            &DbMetadataKey::LedgerCommitProgress,
+            &DbMetadataValue::Version(0),
+        )?;
+
+        self.ledger_db.metadata_db().write_schemas(ledger_batch)
```
msmouse, line 1238: How does it work with the truncator? If immediately restarted, doe it go away?

aptos-node/src/storage.rs
```diff
@@ -42,10 +46,15 @@ pub(crate) fn bootstrap_db(
...
-                maybe_apply_genesis(&DbReaderWriter::from_arc(temp_db), node_config)?;
-
+                let ledger_info_opt =
+                    maybe_apply_genesis(&DbReaderWriter::from_arc(temp_db), node_config)?;
                 let (db_arc, db_rw) = DbReaderWriter::wrap(fast_sync_db_wrapper);
                 let fast_sync_db = db_arc.get_fast_sync_db();
+                // FastSyncDB requires ledger info at epoch 0 to establish provenance to genesis
+                if let Some(ledger_info) = ledger_info_opt {
+                    fast_sync_db.commit_genesis_ledger_info(&ledger_info)?;
+                }
```
msmouse, line 56: At this point the main DB must have the genesis LI. Shall we just read it out and put it to the fast sync DB unconditionally / if it doesn't present already?

areshand: if the temp_db doesn't apply the genesis, not sure this is a clean start so that we don't apply genesis to the fast_synced_db

storage/aptosdb/src/lib.rs
areshand: should be

storage/aptosdb/src/lib.rs
areshand: overall progress is separate from ledger progress since I may only want to commit ledger progress while the overall progress doesn't change

msmouse: I think the logic is "LI goes in last, together with the overall progress". I don't think it's necessary to change that -- especially not necessary in this PR since it's irrelevant with your introduction of the `commit_genesis_LI()` method?

@grao1991 ?

aptos-node/src/storage.rs
msmouse: It's fine, right? The genesis LI is deterministic. If you are paranoid, write only if it doesn't exist in the fast sync DB.

storage/aptosdb/src/lib.rs
areshand: deleted, since I no longer need to commit ledger progress at 0 separately now

aptos-node/src/storage.rs
areshand: addressed

execution/executor/src/db_bootstrapper.rs
```diff
@@ -47,13 +47,13 @@ pub fn maybe_bootstrap<V: VMExecutor>(
     db: &DbReaderWriter,
     genesis_txn: &Transaction,
     waypoint: Waypoint,
-) -> Result<bool> {
+) -> Result<Option<LedgerInfoWithSignatures>> {
```
msmouse, line 50: Change is no longer relevant. I wanted to say it's okay, but looks like making a clean commit isn't too hard in this case?  Your choice.

msmouse (approved):

PR #10555, opened 2023-10-17
Unify protos, add codegen for all languages, set up for package publishing

banool:
### Description
In this PR we make the following changes:
- We move anything proto related to the protos directory. I do it this way because there doesn't seem to be a good place currently for something that contains code for a bunch of different languages + IDL stuff.
- We delete crate/aptos-protos and use protos/rust instead.
- We set up codegen for TS and Python based on the protos.
- We set up all 3 for package publishing by using proper versioning and adding CHANGELOG, CONTRIBUTING, and README files.

Lots of this is copy pasta of the stuff I did in https://github.com/aptos-labs/aptos-indexer-processors.

### Test Plan
Rust:
```
./scripts/rust_lint.sh
```

Python:
```
cd protos/python
poetry install
poetry run poe generate
```

TS:
```
cd protos/typescript
pnpm install
pnpm prepare
```

I have confirmed that this works with aptos-indexer-processors: https://github.com/aptos-labs/aptos-indexer-processors/pull/163.

### Compatibility Testing
To test that old clients work with this txn stream, first run a local testnet from this code (without the indexer API):
```
cargo run -p aptos -- node run-local-testnet --assume-yes --force-restart
```

Now run a processor built from main in aptos-indexer-processors:
```
cargo run -p processor -- -c ~/config.yaml
```

Using this config:
```
health_check_port: 8084
server_config:
  processor_config:
    type: default_processor
  postgres_connection_string: postgresql://postgres:@localhost:5432/default_processor
  indexer_grpc_data_service_address: http://127.0.0.1:50051
  auth_token: dummy
```

The user_transaction_processor works too.


protos/proto/aptos/transaction/v1/transaction.proto
```diff
@@ -403,6 +403,7 @@ message Signature {
     TYPE_MULTI_ED25519 = 2;
     TYPE_MULTI_AGENT = 3;
     TYPE_FEE_PAYER = 4;
+    TYPE_SECP256K1_ECDSA = 5;
```
banool, line 406: Two things:
1. Are we keeping TYPE_SECP256K1_ECDSA and friends or TYPE_SINGLE_SENDER and friends?
2. Unless we're confident we haven't rolled these out yet, instead of totally deleting one of them we should mark it as reserved.

bowenyang007 (approved):

protos/proto/aptos/transaction/v1/transaction.proto
larry-aptos: Let's do the `reserved`; we unintentionally released the proto change to testnet and mainnet.

banool: Which one are we not using? TYPE_SECP256K1_ECDSA or TYPE_SINGLE_SENDER? I'll mark it as reserved.

larry-aptos: I realized it'll be challenging to handle the oneof removal.. let's remove instead of `reserved`; i expect no clients expect these fields.

banool: Spoke with Maayan, we're getting rid of TYPE_SECP256K1_ECDSA.

larry-aptos (approved):

ecosystem/indexer-grpc/indexer-grpc-fullnode/src/convert.rs
```diff
@@ -585,10 +585,7 @@ fn convert_signature(signature: &Signature) -> transaction::AnySignature {
             r#type: transaction::any_signature::Type::Ed25519 as i32,
             signature: s.0.clone(),
         },
-        Signature::Secp256k1Ecdsa(s) => transaction::AnySignature {
-            r#type: transaction::any_signature::Type::Secp256k1Ecdsa as i32,
-            signature: s.0.clone(),
-        },
+        Signature::Secp256k1Ecdsa(_) => panic!("Secp256k1Ecdsa conversion not supported"),
```
banool, line 588: Please confirm that this is okay.

larry-aptos: looks good. tyty!

bowenyang007: why do we need this? i thought this was removed

larry-aptos: let's remove the proto?

PR #10558, opened 2023-10-18
drop smt in parallel

msmouse:
hen a old SMT drops late (many descendants droped early), a long chain
of `Inner`s drop sequantially in the current implementation. This Makes that
concurrent.

### DIFF
--- a/crates/aptos-drop-helper/src/lib.rs
+++ b/crates/aptos-drop-helper/src/lib.rs
@@ -10,7 +10,7 @@ use aptos_metrics_core::TimerHelper;
 use once_cell::sync::Lazy;
 use std::sync::mpsc::{channel, Receiver, Sender};
 
-pub static DEFAULT_DROP_HELPER: Lazy<DropHelper> = Lazy::new(|| DropHelper::new("default", 128));
+pub static DEFAULT_DROP_HELPER: Lazy<DropHelper> = Lazy::new(|| DropHelper::new("default", 32));
 
 /// A helper to send things to a thread pool for asynchronous dropping.
 ///
--- a/crates/aptos-drop-helper/src/lib.rs
+++ b/crates/aptos-drop-helper/src/lib.rs
@@ -28,7 +28,9 @@ impl DropHelper {
     pub fn new(name: &'static str, max_concurrent_drops: usize) -> Self {
         let (token_tx, token_rx) = channel();
         for _ in 0..max_concurrent_drops {
-            token_tx.send(()).expect("Failed to buffer initial tokens.");
+            token_tx
+                .send(())
+                .expect("DropHelper: Failed to buffer initial tokens.");
         }
         Self {
             name,
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3336,6 +3336,7 @@ name = "aptos-scratchpad"
 version = "0.1.0"
 dependencies = [
  "aptos-crypto",
+ "aptos-drop-helper",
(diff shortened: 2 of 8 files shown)


crates/aptos-drop-helper/src/lib.rs
```diff
@@ -0,0 +1,108 @@
...
+    name: &'static str,
+    token_tx: Sender<()>,
+    token_rx: Mutex<Receiver<()>>,
+}
+
+impl DropHelper {
+    pub fn new(name: &'static str, max_concurrent_drops: usize) -> Self {
+        let (token_tx, token_rx) = channel();
+        for _ in 0..max_concurrent_drops {
+            token_tx.send(()).expect("Failed to buffer initial tokens.");
```
areshand, line 31: nit: "failed to buffer initial tokens for concurrent dropping "

crates/aptos-drop-helper/src/lib.rs
```diff
@@ -0,0 +1,108 @@
...
+    pub fn schedule_drop_with_waiter<V: Send + 'static>(&self, v: V) -> Receiver<()> {
+        let (tx, rx) = channel();
+        self.schedule_drop_impl(v, Some(tx));
+        rx
+    }
+
+    fn schedule_drop_impl<V: Send + 'static>(&self, v: V, notif_sender_opt: Option<Sender<()>>) {
+        let _timer = TIMER.timer_with(&[self.name, "enqueue_drop"]);
+
+        self.token_rx.lock().recv().unwrap();
```
areshand, line 53: if we don't lock, what problem could happen here?

crates/aptos-drop-helper/src/lib.rs
```diff
@@ -0,0 +1,108 @@
...
+pub struct DropHelper {
+    name: &'static str,
+    token_tx: Sender<()>,
+    token_rx: Mutex<Receiver<()>>,
+}
+
+impl DropHelper {
+    pub fn new(name: &'static str, max_concurrent_drops: usize) -> Self {
+        let (token_tx, token_rx) = channel();
+        for _ in 0..max_concurrent_drops {
```
areshand, line 30: Do we need to check how the memory will increase if V is not dropped in time?

crates/aptos-drop-helper/src/lib.rs
msmouse: You can't receive from multiple threads at the same time for a mpsc channel. (can't compile either)

crates/aptos-drop-helper/src/lib.rs
msmouse: That's why we made it bounded?

crates/aptos-drop-helper/src/lib.rs
msmouse: okay... 
I put in an `expect` instead of `unwrap` just to avoid being "nit: expect()"ed, and now I still got nitted and have to make this bullshit 3 lines 😢 :

areshand (approved):

grao1991 (approved):

crates/aptos-drop-helper/src/lib.rs
msmouse: done

crates/aptos-drop-helper/src/lib.rs
msmouse: Yeah but only a limited number (32) of drops can happen in concurrency.

PR #10560, opened 2023-10-18
[Storage Service] Small improvements to storage service.

JoshLind:
### Description
This PR makes several small improvements to the storage service, each in their own commits:
1. Add additional latency metrics to request handling.
2. Replace `RwLock<HashMap<T>>` with `DashMap<T>` in the request moderator.
3. Replace `LRUCache` with a more efficient concurrent cache (to avoid a single `Mutex`).

### Test Plan
Existing test infrastructure.

### DIFF
--- a/state-sync/storage-service/server/src/handler.rs
+++ b/state-sync/storage-service/server/src/handler.rs
@@ -6,8 +6,8 @@ use crate::{
     logging::{LogEntry, LogSchema},
     metrics,
     metrics::{
-        increment_counter, start_timer, LRU_CACHE_HIT, LRU_CACHE_PROBE, OPTIMISTIC_FETCH_ADD,
-        SUBSCRIPTION_ADD, SUBSCRIPTION_FAILURE, SUBSCRIPTION_NEW_STREAM,
+        increment_counter, LRU_CACHE_HIT, LRU_CACHE_PROBE, OPTIMISTIC_FETCH_ADD, SUBSCRIPTION_ADD,
+        SUBSCRIPTION_FAILURE, SUBSCRIPTION_NEW_STREAM,
     },
     moderator::RequestModerator,
     network::ResponseSender,
@@ -121,8 +121,8 @@ impl<T: StorageReaderInterface> Handler<T> {
         request: StorageServiceRequest,
         optimistic_fetch_related: bool,
     ) -> aptos_storage_service_types::Result<StorageServiceResponse> {
-        // Time the request processing (the timer will stop when it's dropped)
-        let _timer = start_timer(
+        // Time the request processing
+        let _timer = metrics::start_timer(
             &metrics::STORAGE_REQUEST_PROCESSING_LATENCY,
             peer_network_id.network_id(),
             request.get_label(),
@@ -371,6 +371,13 @@ impl<T: StorageReaderInterface> Handler<T> {
             );
             return Ok(response.clone());
         }
+
+        // Otherwise, we need to fetch the data from storage. Start the timer.
+        let _timer = metrics::start_timer(
+            &metrics::STORAGE_FETCH_PROCESSING_LATENCY,
+            peer_network_id.network_id(),
+            request.get_label(),
+        );
 
         // Fetch the data response from storage
         let data_response = match &request.data_request {
--- a/state-sync/storage-service/server/src/handler.rs
+++ b/state-sync/storage-service/server/src/handler.rs
@@ -16,7 +16,6 @@ use crate::{
     subscription::{SubscriptionRequest, SubscriptionStreamRequests},
 };
 use aptos_config::{config::StorageServiceConfig, network_id::PeerNetworkId};
-use aptos_infallible::Mutex;
 use aptos_logger::{debug, error, sample, sample::SampleRate, trace, warn};
 use aptos_storage_service_types::{
     requests::{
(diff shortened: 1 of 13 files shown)


grao1991 (commented):
Have you checked if it will indeed improve the performance. Last time I tried moka cache, it didn't work well.

JoshLind:
> Have you checked if it will indeed improve the performance. Last time I tried moka cache, it didn't work well.

Good question! It should do. Internally it uses a `DashMap` which is better than having a single lock. But, I'll run some benchmarks to double verify 😄

JoshLind:
So, I did do some local benchmarking to compare the `Arc` vs. `Cache`. Things definitely don't get worse. They look like they get better, but there's a lot of noise in our experiments, so it's hard to tell. Either way, I'm happy to move forward with this cache and will see how things go in previewnet (where the difference will likely matter 😄).

brianolson:
this is 200 lines of code, and 7000 lines of new dependency `mini-moka`

JoshLind:
> this is 200 lines of code, and 7000 lines of new dependency mini-moka

😄 It's only 200 lines long because I used a dependency. Should I beef up the PR to make it more even, or reimplement the library myself? More seriously though, what's the exact concern? `LRUCache` is at least a few thousand lines.

state-sync/storage-service/server/src/handler.rs
```diff
@@ -400,10 +406,8 @@ impl<T: StorageReaderInterface> Handler<T> {
         let storage_response = StorageServiceResponse::new(data_response, request.use_compression)?;
 
         // Cache the response before returning
-        let _ = self
-            .lru_response_cache
-            .lock()
-            .put(request.clone(), storage_response.clone());
+        self.lru_response_cache
```
bchocho, line 409: This is interesting. I was expecting some kind of [get_with](https://docs.rs/moka/latest/moka/sync/struct.Cache.html#method.get_with) but looks like moka-mini doesn't have that.

I'm wondering if this matters though. For any request, would the response possibly change with additional data? For example-

So a question about the original design here. Could this happen, and is it ok?:
A request comes in with a large end_index, and but there is only data up to end_index - 10 at this point. Then any subsequent requests with the same start_index, end_index will always return only up to end_index - 10.

Then additionally for this change, is it ok if the same start_index, end_index can return different answers, and the first returns end_index - 10 but the second only returns end_index - 20, and all subsequent calls of the same request also return end_index - 20?

JoshLind: > I was expecting some kind of [get_with] but looks like moka-mini doesn't have that.

Yeah, it's kinda annoying, but not a deal breaker (see below). 😄 

The high-level design is essentially that it's okay for two identical requests to race, miss the cache, fetch from storage and insert them into the cache. The work is duplicated, but this is okay because: (i) any requests that are identical will always have identical responses (see below); (ii) in the current `LRUCache` implementation, there's a single `Mutex`, so we don't want to block the entire cache (and all other requests) while waiting to fetch the data for one request; and (iii) even in the new design, it appears we don't have the necessary interfaces to block only the identical requests -- but this is ultimately okay.

> A request comes in with a large end_index, and but there is only data up to end_index - 10 at this point.

This isn't possible -- any request that explicitly requests more data than the node has is invalid and will be dropped by the request moderator.

> is it ok if the same start_index, end_index can return different answers, and the first returns end_index - 10 but the second only returns end_index - 20, and all subsequent calls of the same request also return end_index - 20?

I don't think this is possible. Any identical requests will get the same data. The only edge case is that they get less data than they requested (i.e., they requested 10k items, but the max is 1k, so we truncate). But, even in this case, the truncation rule is static and deterministic (i.e., defined by the configs), so both requests will be truncated identically.

Taking a step back though, even if what you were saying is possible (which I'm confident it's not), this still wouldn't be a big issue because the client-side already has to deal with truncated data, so the clients would identify the missing data pieces and re-request them 😄 

Does this make sense? Have I understood the question?

state-sync/storage-service/server/src/moderator.rs
```diff
@@ -137,8 +136,15 @@ impl RequestModerator {
         peer_network_id: &PeerNetworkId,
         request: &StorageServiceRequest,
     ) -> Result<(), Error> {
+        // Time the request validation
+        let _timer = metrics::start_timer(
```
brianolson, line 140: the timer is counted when it goes out of scope?

brianolson:
I skimmed mini-moka and didn't see any red flags. it has some tests but I wish there were more. I guess it's probably okay.

state-sync/storage-service/server/src/moderator.rs
JoshLind: Correct 😄

brianolson (approved):

JoshLind:
Also, it seems like there's enough push-back on the `mini-moka` change. Personally, I don't have any concerns with it, but I'm happy to just delete the last commit until I have more data to justify it?

state-sync/storage-service/server/src/handler.rs
bchocho: Thanks @JoshLind your explanation makes sense.

Just FYI, in my mempool dissemination branch, I use `moka` (not `mini-moka`) and it provides `get_with*`. But it's also a larger dependency.

bchocho (approved):

PR #10569, opened 2023-10-18
[aptos-debugger] introduce remote-gas-profiler binary

vgao1996:
This adds a new binary `remote-gas-profiler` to the `aptos-debugger` crate. The tool allows you to generate a gas profiling report for a committed transaction.

Here's an example command:
```
cargo run -p aptos-move-debugger --bin remote-gas-profiler -- --version 721512672
 rest https://testnet.aptoslabs.com/v1
```

### DIFF
--- a/aptos-move/aptos-debugger/src/bin/remote-gas-profiler.rs
+++ b/aptos-move/aptos-debugger/src/bin/remote-gas-profiler.rs
@@ -1,0 +1,75 @@
+// Copyright © Aptos Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+use anyhow::{bail, Result};
+use aptos_move_debugger::aptos_debugger::AptosDebugger;
+use aptos_rest_client::Client;
+use aptos_types::transaction::Transaction;
+use aptos_vm::AptosVM;
+use clap::{Parser, Subcommand};
+use std::path::{Path, PathBuf};
+use url::Url;
+
+#[derive(Subcommand)]
+pub enum Target {
+    /// Use full node's rest api as query endpoint.
+    Rest { endpoint: String },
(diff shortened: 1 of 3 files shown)


aptos-move/aptos-debugger/src/bin/remote-gas-profiler.rs
```diff
@@ -0,0 +1,75 @@
...
+pub struct Args {
+    #[clap(subcommand)]
+    target: Target,
+
+    #[clap(long)]
+    version: u64,
+}
+
+#[tokio::main]
+async fn main() -> Result<()> {
```
vgao1996, line 31: @runtian-zhou I tried to add this functionality to the main debugger binary but gave up due to a difference in the execution flow.

The debugger has the ability to execute multiple transactions as a block, while the the gas profiler doesn't. It's possible to add block support but that will require some changes to Aptos VM's interfaces, which I'd not prioritize for now.

grao1991 (approved):

aptos-move/aptos-debugger/src/bin/remote-gas-profiler.rs
vgao1996: > I think you can add directly to the debugger binary? Just cloning the logic there would be fine. With that being said, who will be the user of this binary?

Tried, and it didn't work out due to the logical difference I mentioned in my previous comment.

> With that being said, who will be the user of this binary? Cuz if we expect regular user to use this I would suggest adding this functionality to Aptos binary

For now this is only for core developers like us to debug and study transactions. However you made a good point that this could also be useful for all developers. I'll add CLI support, but at a later time.

runtian-zhou (approved):

PR #10519, opened 2023-10-16
[types] introducing single sender auth of single and multi keys

davidiw:
Introduces a way to encapsulate generic crypto as the sender of a transaction creating an indirection an allow us to have multi keys of differing types.

PR #10576, opened 2023-10-19
[authenticator] add sanity checks to multiauth

alinush:
### Description

A few small fixes for #10519:
 - Make sure that when we create a multiauth address, the # of signatures required is $> 0$.
 - Make sure that when we create a multiauth address, the # of PKs is $\ge$ to the # of signatures required.
 - Make sure that when we verify a multiauth signature, the signature bitmap is not empty
 - Make sure that when we verify a multiauth signature, the signature bitmap does not have out-of-bounds indices

### DIFF
--- a/types/src/transaction/authenticator.rs
+++ b/types/src/transaction/authenticator.rs
@@ -786,6 +786,16 @@ impl MultiKeyAuthenticator {
 
     pub fn verify<T: Serialize + CryptoHash>(&self, message: &T) -> Result<()> {
         ensure!(
+            self.signatures_bitmap.last_set_bit().is_some(),
+            "There were no signatures set in the bitmap."
+        );
+
+        ensure!(
+            (self.signatures_bitmap.last_set_bit().unwrap() as usize) < self.public_keys.len(),
+            "Mismatch in the position of the last signature and the number of PKs, {} >= {}.",
+            self.signatures_bitmap.last_set_bit().unwrap(), self.public_keys.len(),
+        );
+        ensure!(
             self.signatures_bitmap.count_ones() as usize == self.signatures.len(),
             "Mismatch in number of signatures and the number of bits set in the signatures_bitmap, {} != {}.", self.signatures_bitmap.count_ones(), self.signatures.len(),
         );
@@ -820,11 +830,22 @@ pub struct MultiKey {
 }
 
 impl MultiKey {
-    pub fn new(public_keys: Vec<AnyPublicKey>, signatures_required: u8) -> Self {
-        Self {
+    pub fn new(public_keys: Vec<AnyPublicKey>, signatures_required: u8) -> Result<Self> {
+        ensure!(
+            signatures_required > 0,
+            "The number of required signatures is 0."
+        );
+
+        ensure!(
+            public_keys.len() >= signatures_required as usize,
+            "The number of public keys is smaller than the number of required signatures, {} < {}",
+            public_keys.len(), signatures_required
+        );
+
+        Ok(Self {
             public_keys,
             signatures_required,
-        }
+        })
     }
 
     pub fn is_empty(&self) -> bool {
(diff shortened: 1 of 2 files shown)


zjma (approved):

mstraka100 (approved):

davidiw (commented):
thank you.

davidiw (approved):

types/src/transaction/authenticator.rs
```diff
@@ -785,6 +785,17 @@ impl MultiKeyAuthenticator {
     }
 
     pub fn verify<T: Serialize + CryptoHash>(&self, message: &T) -> Result<()> {
+        ensure!(
+            self.signatures_bitmap.last_set_bit().is_some(),
+            "There were no signatures set in the bitmap."
+        );
+
+        ensure!(
+            (self.signatures_bitmap.last_set_bit().unwrap() as usize) < self.public_keys.len(),
```
zekun000, line 794: nit: I think the two checks can be combined
```
last_set_bit().map_or(false, |idx| idx < self.public_keys.len())
```

we don't need to enforce the bitvec buckets, but from conservative perspective I prefer to enforce restrictions as much as possible

types/src/transaction/authenticator.rs
```diff
@@ -820,11 +831,23 @@ pub struct MultiKey {
...
 impl MultiKey {
-    pub fn new(public_keys: Vec<AnyPublicKey>, signatures_required: u8) -> Self {
-        Self {
+    pub fn new(public_keys: Vec<AnyPublicKey>, signatures_required: u8) -> Result<Self> {
+        ensure!(
+            signatures_required > 0,
+            "The number of required signatures is 0."
+        );
+
+        ensure!(
```
zekun000, line 840: if we want to enforce this, it's not sufficient in constructor. we directly deserialize the multikey from network, it needs to be checked either in a verify function or deserialization.

although I think it might be impossible to enforce it, move only takes a hash of authenticator for key rotations, we don't have access to the actual authenticator value

types/src/transaction/authenticator.rs
alinush: If I combine the two checks, the error messages will not be as clear anymore. I'll leave them like this for now.

types/src/transaction/authenticator.rs
alinush: Hm...

It is indeed impossible to enforce, since anyone can send a payment to a non-existing **invalid** multiauth address (e.g., with 0 signatures required).

This is not unique to multiauth addresses though.

Even, say, Ed25519 addresses have the same problem: anyone can send a payment to a non-existing **invalid** Ed25519 address (e.g., the SK associated with the address's underlying PK is unknown).

It will not be possible to check anything about `MultiKey` during deserialization, since that happens behind our backs via `serde`...

So you are right: we need to duplicate some of these checks in `MultiKeyAuthenticator::verify`...

alinush: cc @davidiw 

I would add the following to `MultiKeyAuthenticator::verify`:

```
diff --git a/types/src/transaction/authenticator.rs b/types/src/transaction/authenticator.rs
index 492a5c179f..379adf80cb 100644
--- a/types/src/transaction/authenticator.rs
+++ b/types/src/transaction/authenticator.rs
@@ -785,11 +785,20 @@ impl MultiKeyAuthenticator {
     }

     pub fn verify<T: Serialize + CryptoHash>(&self, message: &T) -> Result<()> {
+        ensure!(
+            self.public_keys.signatures_required() > 0,
+            "The number of required signatures is 0."
+        );
+        ensure!(
+            self.public_keys.len() >= self.public_keys.signatures_required() as usize,
+            "The number of public keys is smaller than the number of required signatures, {} < {}",
+            self.public_keys.len(),
+            signatures_required
+        );
         ensure!(
             self.signatures_bitmap.last_set_bit().is_some(),
             "There were no signatures set in the bitmap."
         );
-
         ensure!(
             (self.signatures_bitmap.last_set_bit().unwrap() as usize) < self.public_keys.len(),
             "Mismatch in the position of the last signature and the number of PKs, {} >= {}.",
```

This would prevent accessing [accidentally-created?] multiauth accounts with either (1) 0 required signatures or (2) less PKs than required signatures.

For (2), the other logic in `MultiKeyAuthenticator::verify` would prevent access anyway. 

But for (1) is there a reason we might want to allow _anyone_ to spend/access a 0-out-of-n multiauth account?

PR #10592, opened 2023-10-19
[agg_v2] Use aggregator v2 fallback when resolver is not "delayed field capable"

igor-aptos:
DelayedFields is an optimization within BlockExecutor.

For resolvers that are not coming from BlockExecutor, they wouldn't know how to handle it - how to do the exchanges or materialize the output. And so add additional function within the TDelayedFieldView - is_delayed_field_optimization_capable - that will tell native code whether it can use delayed fields, or should directly execute operations on move values.

### DIFF
--- a/aptos-move/aptos-vm-types/src/change_set.rs
+++ b/aptos-move/aptos-vm-types/src/change_set.rs
@@ -181,6 +181,8 @@ impl VMChangeSet {
     pub fn try_from_storage_change_set(
         change_set: StorageChangeSet,
         checker: &dyn CheckChangeSet,
+        // Pass in within which resolver context are we creating this change set.
+        // Used to eagerly reject changes created in an incompatible way.
         is_delayed_field_optimization_capable: bool,
     ) -> anyhow::Result<Self, VMStatus> {
         assert!(
@@ -223,8 +225,12 @@ impl VMChangeSet {
     }
 
     pub(crate) fn into_storage_change_set_unchecked(self) -> StorageChangeSet {
-        // It is still invalid to convert VMChangeSet that was created in a mode
-        // that produces DelayedField or ResourceGroup write sets
+        // Converting VMChangeSet into TransactionOutput (i.e. storage change set), can
+        // be done here only if dynamic_change_set_optimizations have not been used/produced
+        // data into the output.
+        // If they (DelayedField or ResourceGroup) have added data into the write set, translation
+        // into output is more complicated, and needs to be done within BlockExecutor context
+        // that knows how to deal with it.
         assert!(self.delayed_field_change_set().is_empty());
         assert!(self.resource_group_write_set().is_empty());
 
--- a/aptos-move/aptos-vm-types/src/change_set.rs
+++ b/aptos-move/aptos-vm-types/src/change_set.rs
@@ -181,7 +181,12 @@ impl VMChangeSet {
     pub fn try_from_storage_change_set(
         change_set: StorageChangeSet,
         checker: &dyn CheckChangeSet,
+        is_delayed_field_optimization_capable: bool,
     ) -> anyhow::Result<Self, VMStatus> {
+        assert!(
+            !is_delayed_field_optimization_capable,
+            "try_from_storage_change_set can only be called in non-is_delayed_field_optimization_capable context, as it doesn't support delayed field changes (type layout) and resource groups");
+
         let (write_set, events) = change_set.into_inner();
 
         // There should be no aggregator writes if we have a change set from
@@ -341,6 +346,12 @@ impl VMChangeSet {
         self.events = patched_events
             .map(|event| (event, None))
             .collect::<Vec<_>>();
+    }
+
+    pub(crate) fn drain_delayed_field_change_set(
+        &mut self,
+    ) -> impl Iterator<Item = (DelayedFieldID, DelayedChange<DelayedFieldID>)> + '_ {
+        std::mem::take(&mut self.delayed_field_change_set).into_iter()
     }
 
     pub fn aggregator_v1_write_set(&self) -> &BTreeMap<StateKey, WriteOp> {
--- a/aptos-move/aptos-vm-types/src/change_set.rs
+++ b/aptos-move/aptos-vm-types/src/change_set.rs
@@ -5,7 +5,7 @@ use crate::check_change_set::CheckChangeSet;
 use aptos_aggregator::{
     delayed_change::DelayedChange,
     delta_change_set::{serialize, DeltaOp},
-    resolver::DelayedFieldResolver,
+    resolver::AggregatorV1Resolver,
     types::{code_invariant_error, DelayedFieldID},
 };
 use aptos_types::{
@@ -218,6 +218,11 @@ impl VMChangeSet {
     }
 
     pub(crate) fn into_storage_change_set_unchecked(self) -> StorageChangeSet {
+        // It is still invalid to convert VMChangeSet that was created in a mode
+        // that produces DelayedField or ResourceGroup write sets
+        assert!(self.delayed_field_change_set().is_empty());
+        assert!(self.resource_group_write_set().is_empty());
+
         let Self {
             resource_write_set,
             resource_group_write_set: _,
@@ -245,11 +250,19 @@ impl VMChangeSet {
     /// - deltas are not materialized.
     /// - resource group writes are not (combined &) converted to resource writes.
     pub fn try_into_storage_change_set(self) -> anyhow::Result<StorageChangeSet, VMStatus> {
-        if !self.aggregator_v1_delta_set.is_empty() || !self.delayed_field_change_set.is_empty() {
+        if !self.aggregator_v1_delta_set.is_empty() {
             return Err(VMStatus::error(
                 StatusCode::DATA_FORMAT_ERROR,
                 err_msg(
-                    "Cannot convert from VMChangeSet with non-materialized deltas to ChangeSet.",
+                    "Cannot convert from VMChangeSet with non-materialized Aggregator V1 deltas to ChangeSet.",
+                ),
+            ));
+        }
+        if !self.delayed_field_change_set.is_empty() {
+            return Err(VMStatus::error(
+                StatusCode::DATA_FORMAT_ERROR,
+                err_msg(
+                    "Cannot convert from VMChangeSet with non-materialized Delayed Field changes to ChangeSet.",
                 ),
             ));
         }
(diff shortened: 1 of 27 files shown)


georgemitenkov (approved):

aptos-move/aptos-vm-types/src/change_set.rs
```diff
@@ -218,6 +223,11 @@ impl VMChangeSet {
     }
 
     pub(crate) fn into_storage_change_set_unchecked(self) -> StorageChangeSet {
+        // It is still invalid to convert VMChangeSet that was created in a mode
+        // that produces DelayedField or ResourceGroup write sets
+        assert!(self.delayed_field_change_set().is_empty());
+        assert!(self.resource_group_write_set().is_empty());
```
vusirikala, line 229: Can you please expand the comment a little more on why it is invalid to convert VMChangeSet when the resource group write set is non-empty.

aptos-move/aptos-vm/src/block_executor/vm_wrapper.rs
```diff
@@ -43,6 +45,13 @@ impl<'a, S: 'a + StateView + Sync> ExecutorTask for AptosExecutorTask<'a, S> {
         txn_idx: TxnIndex,
         materialize_deltas: bool,
     ) -> ExecutionStatus<AptosTransactionOutput, VMStatus> {
+        if txn.is_valid() && executor_with_group_view.is_delayed_field_optimization_capable() {
+            if let Transaction::GenesisTransaction(WriteSetPayload::Direct(_)) = txn.expect_valid()
+            {
+                return ExecutionStatus::DirectWriteSetTransactionNotCapableError;
+            }
+        }
```
vusirikala, line 53: Please add a comment here on why we are returning an error in this case.

aptos-move/block-executor/src/errors.rs
```diff
@@ -4,17 +4,20 @@
...
+    /// TODO: (short-mid term) relax the limitation, and (mid-long term) provide proper multi-versioning
+    /// for code (like data) for the cache.
+    ModulePathReadWrite,
+    DirectWriteSetTransaction,
+}
 
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub enum Error<E> {
-    FallbackToSequential(PanicOr<ModulePathReadWrite>),
+    FallbackToSequential(PanicOr<IntentionalFallbackToSequential>),
```
vusirikala, line 20: As `IntentionalFallbackToSequential` is an enum anyways, should we just add `CodeInvariantError` as a variant to the enum? Is PanicOr serving any additional benefit here?

aptos-move/block-executor/src/errors.rs
```diff
@@ -4,17 +4,20 @@
...
 #[derive(Copy, Clone, Debug, PartialEq, Eq)]
-pub struct ModulePathReadWrite;
+pub enum IntentionalFallbackToSequential {
+    /// The same module access path for module was both read & written during speculative executions.
+    /// This may trigger a race due to the Move-VM loader cache implementation, and mitigation requires
+    /// aborting the parallel execution pipeline and falling back to the sequential execution.
+    /// TODO: (short-mid term) relax the limitation, and (mid-long term) provide proper multi-versioning
+    /// for code (like data) for the cache.
+    ModulePathReadWrite,
+    DirectWriteSetTransaction,
```
vusirikala, line 15: Please also add some documentation for `DirectWriteSetTransaction` variant.

aptos-move/block-executor/src/executor.rs
```diff
@@ -750,6 +757,9 @@ where
                 ExecutionStatus::Abort(_) => {
                     txn_commit_listener.on_execution_aborted(txn_idx);
                 },
+                ExecutionStatus::DirectWriteSetTransactionNotCapableError => {
+                    panic!("Cannot be materializing with DirectWriteSetTransactionNotCapableError");
```
vusirikala, line 761: panic? Shouldn't we fall back to sequential execution?

aptos-move/block-executor/src/executor.rs
```diff
@@ -763,6 +773,9 @@ where
                 final_results[txn_idx as usize] = t;
             },
             ExecutionStatus::Abort(_) => (),
+            ExecutionStatus::DirectWriteSetTransactionNotCapableError => {
+                panic!("Cannot be materializing with DirectWriteSetTransactionNotCapableError");
```
vusirikala, line 777: panic?

aptos-move/block-executor/src/executor.rs
```diff
@@ -1114,17 +1134,25 @@ where
                     // Record the status indicating abort.
                     return Err(Error::UserError(err));
                 },
+                ExecutionStatus::DirectWriteSetTransactionNotCapableError => {
+                    return Err(Error::FallbackToSequential(PanicOr::Or(
+                        IntentionalFallbackToSequential::DirectWriteSetTransaction,
```
vusirikala, line 1139: We are already in sequential execution. What would happen differently when we fallback to sequential execution again here?

aptos-move/block-executor/src/executor.rs
```diff
@@ -1188,15 +1216,28 @@ where
...
         if let Err(Error::FallbackToSequential(e)) = ret {
-            if let PanicOr::Or(ModulePathReadWrite) = e {
-                debug!("[Execution]: Module read & written, sequential fallback");
-            } else {
-                error!("[Execution]: {:?}, sequential fallback", e);
-            }
+            let dynamic_change_set_optimizations_enabled = match e {
+                PanicOr::Or(IntentionalFallbackToSequential::ModulePathReadWrite) => {
+                    debug!("[Execution]: Module read & written, sequential fallback");
+                    true
```
vusirikala, line 1227: What's the idea behind enabling the `dynamic_change_set_optimizations_enabled` flag when ModulePathReadWrite error and disabling the flag for the other two error types?

aptos-move/block-executor/src/executor.rs
```diff
@@ -1206,7 +1247,39 @@ where
...
+                    match e {
+                        PanicOr::Or(IntentionalFallbackToSequential::ModulePathReadWrite) => {
+                            panic!("ModulePathReadWrite shouldn't happen in sequential execution")
+                        },
+                        PanicOr::Or(IntentionalFallbackToSequential::DirectWriteSetTransaction) => {
+                            info!("[Execution]: DirectWriteSetTransaction found, during ModulePathReadWrite sequential fallback");
+                        },
+                        PanicOr::CodeInvariantError(msg) => {
+                            error!("[Execution]: CodeInvariantError({:?}) in sequential with dynamic_change_set_optimizations_enabled, sequential fallback", msg);
+                        },
```
vusirikala, line 1264: Let's hope that these errors don't occur the second time we run the transactions sequentially. In case they ever occur in production, what's the plan to solve those during on-call quickly?

aptos-move/aptos-vm/src/storage_adapter/executor_view_base.rs
```diff
@@ -44,32 +44,39 @@ impl<S: StateView> AsExecutorView<S> for S {
...
     ) -> anyhow::Result<Option<StateValue>> {
         self.base.get_state_value(state_key)
     }
+}
+
+impl<'s, S: StateView> TDelayedFieldView for ExecutorViewBase<'s, S> {
+    type Identifier = DelayedFieldID;
+
+    fn is_aggregator_v2_delayed_fields_enabled(&self) -> bool {
+        false
```
vusirikala, line 63: The flag is hardcoded here? We earlier used to have feature flags AGGREGATOR_V2_API_ENABLED, AGGREGATOR_V2_EXECUTION_ENABLED right?

aptos-move/framework/src/natives/aggregator_natives/aggregator_v2.rs
```diff
@@ -185,12 +185,20 @@ impl SnapshotType {
...
+) -> Option<(&'b dyn DelayedFieldResolver, RefMut<'t, DelayedFieldData>)> {
     let aggregator_context = context.extensions().get::<NativeAggregatorContext>();
-    (
-        aggregator_context.resolver,
-        aggregator_context.delayed_field_data.borrow_mut(),
-    )
+    if aggregator_context
+        .delayed_field_resolver
+        .is_delayed_field_optimization_capable()
+        && context.aggregator_v2_delayed_fields_enabled()
```
vusirikala, line 193: What's the difference between `is_delayed_field_optimization_capable` and `aggregator_v2_delayed_fields_enabled` flags?

aptos-move/block-executor/src/executor.rs
igor-aptos: this means we are committing transaction that raised this error - this should never happen.

instead executor will catch it in execute(), and fallback to sequential

aptos-move/block-executor/src/executor.rs
igor-aptos: we now have two sequential fallbacks - with dynamic_change_set_optimizations_enabled true and false.

aptos-move/block-executor/src/executor.rs
igor-aptos: ModulePathReadWrite happens often, and code should be able to handle it with dynamic_change_set_optimizations_enabled=true (and that we test/exercise more of the code)

But, I agree, I am gong to do this fallback for code_invariant_error as well. I am refactoring this code a bit to make it clear

aptos-move/block-executor/src/executor.rs
igor-aptos: We can disable AGGREGATOR_V2_EXECUTION and corresponding RG flag

aptos-move/framework/src/natives/aggregator_natives/aggregator_v2.rs
igor-aptos: is_delayed_field_optimization_capable means whether resolver we are using is capable of using that feature.

aggregator_v2_delayed_fields_enabled means whether a feature has been enabled.

we need both capable resolver (i.e. either one from BlockExecutor) and feature flag enabled.

for example, executing view functions is outside of BlockExecutor, and even after flag is enabled, resolver will not be capable of handling it.

aptos-move/aptos-vm-types/src/change_set.rs
```diff
@@ -223,8 +225,12 @@ impl VMChangeSet {
...
 
     pub(crate) fn into_storage_change_set_unchecked(self) -> StorageChangeSet {
-        // It is still invalid to convert VMChangeSet that was created in a mode
-        // that produces DelayedField or ResourceGroup write sets
+        // Converting VMChangeSet into TransactionOutput (i.e. storage change set), can
+        // be done here only if dynamic_change_set_optimizations have not been used/produced
+        // data into the output.
+        // If they (DelayedField or ResourceGroup) have added data into the write set, trasnlation
+        // into output is more complicated, and needs to be done within BlockExecutor context
+        // that konws how to deal with it.
```
vusirikala, line 233: nit: konws, trasnlation

vusirikala (approved):

PR #10604, opened 2023-10-20
[gas] make gas meter more resilient

vgao1996:
This makes the gas meter more resilient in response to the recent network outage.
1. The gas meter no longer records any "overcharge"
2. This enforces the invariant that "per-category gas costs must add up to the total cost"


aptos-move/aptos-gas-meter/src/algebra.rs
```diff
@@ -85,15 +91,55 @@ impl GasAlgebra for StandardGasAlgebra {
         self.balance
     }
 
+    fn check_consistency(&self) -> PartialVMResult<()> {
```
vgao1996, line 94: @wrwg I wish this is a property I can prove through formal analysis. Do you know if this is possible in Rust?

aptos-move/aptos-vm/src/aptos_vm.rs
```diff
@@ -548,6 +548,17 @@ impl AptosVM {
...
         }
 
+        if self.0.get_gas_feature_version() >= 12 {
+            if let Err(err) = gas_meter.algebra().check_consistency() {
+                error!(
+                    "[aptos-vm][gas-meter] {}",
+                    err.message()
+                        .unwrap_or("No message found -- this should not happen.")
+                );
+                return Err(err.finish(Location::Undefined).into()); //resilient
```
msmouse, line 558: What does resilient refer to here?

msmouse (approved):

aptos-move/aptos-gas-meter/src/algebra.rs
```diff
@@ -156,10 +208,17 @@ impl GasAlgebra for StandardGasAlgebra {
             },
         );
 
-        self.charge(gas_consumed_internal)?;
+        let (actual, res) = self.charge(gas_consumed_internal);
+        if self.feature_version >= 12 {
+            self.storage_fee_in_internal_units += actual;
+            self.storage_fee_used += amount;
```
msmouse, line 214: We don't care to proportionate the `amount`? 😂

msmouse (approved):

aptos-move/aptos-vm/src/aptos_vm.rs
vgao1996: My bad. Got this deleted.

vgao1996:
@msmouse Just realized I didn't invoke the check correctly. Could you take a look at the updated AptosVM code? Thanks

aptos-move/aptos-gas-meter/src/algebra.rs
vgao1996: Good point. The real amount being charged is one that first gets converted into gas units and then converted back. Do you think it makes more sense for us to record that instead?

msmouse: Yeah this field is not that important -- can't be 100% precise anyway. Just wanted to make the FeeStatement makes a bit more sense when we hit OUT_OF_GAS in the middle of charging it. Again, seems no perfect solution here since the converted value has to be part of the total gas used.

msmouse:
@vgao1996 thought you deliberately checked only before the epilogue. what's your thoughts behind it?

grao1991 (approved):

vgao1996:
Replay (old behavior): https://github.com/aptos-labs/aptos-core/actions/runs/6597752756

vgao1996:
@msmouse yes I deliberately only called the check before epilogue. Don't think it's worth to check it after every charge -- it would slow down the VM further.

PR #10622, opened 2023-10-23
[RestClient] Provide a way to disable TLS SNI in environment variable.

e99243506bigplay:
### Description
Somehow I unable to use Aptos CLI to connect https://fullnode.devnet.aptoslabs.com in China, however after I set to no TLS SNI, everything is worked.

### DIFF
--- a/crates/aptos-rest-client/src/client_builder.rs
+++ b/crates/aptos-rest-client/src/client_builder.rs
@@ -51,8 +51,17 @@ impl ClientBuilder {
             HeaderValue::from_static(X_APTOS_SDK_HEADER_VALUE),
         );
 
+        let mut builder = ReqwestClient::builder();
+
+        match ::std::env::var("APTOS_NO_SNI") {
+            Ok(_) => {
+                builder = builder.tls_sni(false);
+            },
+            Err(_) => ()
+        }
+
         Self {
-            reqwest_builder: ReqwestClient::builder(),
+            reqwest_builder: builder,
             base_url: aptos_base_url.to_url(),
             version_path_base: DEFAULT_VERSION_PATH_BASE.to_string(),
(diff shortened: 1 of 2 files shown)


crates/aptos-rest-client/src/client_builder.rs
```diff
@@ -51,8 +51,17 @@ impl ClientBuilder {
...
         );
 
+        let mut builder = ReqwestClient::builder();
+
+        match ::std::env::var("APTOS_NO_SNI") {
+            Ok(_) => {
+                builder = builder.tls_sni(false);
+            },
+            Err(_) => ()
+        }
```
gregnazario, line 61: Probably would be better to provide as an input rather than just an environment variable.  Add a function similar to `base_url` and `timeout` and you can set SNI to be turned off in the builder way which matches the design pattern currently laid out.

gregnazario (commented):
Looks like a good feature to have!  Can you just add it similar to the other builder pattern items

crates/aptos-rest-client/src/client_builder.rs
e99243506bigplay: At that pattern, How could I apply on Aptos CLI to turn off the SNI? through the RestClient parameter? I could only think this is only the suitable way for turn off the SNI. But yea, I should add an option to switch SNI on this impl.

gregnazario: Yeah, it would be through the RestClient parameter.  Similar to this PR https://github.com/aptos-labs/aptos-core/pull/10631

e99243506bigplay: This time should be able to merge I think. Success to pass the compile now I think.

PR #10623, opened 2023-10-23
[Agg v2] Aggregator unit tests

vusirikala:
### Description

Adding a few aggregator unit tests in view.rs to verify if the read operations are performed correctly.

### DIFF
--- a/aptos-move/block-executor/src/view.rs
+++ b/aptos-move/block-executor/src/view.rs
@@ -1088,7 +1088,7 @@ mod test {
         types::{DelayedFieldValue, DelayedFieldsSpeculativeError, PanicOr, ReadPosition},
     };
     use aptos_mvhashmap::{
-        types::{MVDelayedFieldsError, TxnIndex},
+        types::{MVDelayedFieldsError, StorageVersion, TxnIndex},
         unsync_map::UnsyncMap,
         versioned_delayed_fields::TVersionedDelayedFieldView,
         MVHashMap,
@@ -1952,6 +1952,7 @@ mod test {
                 unsync_map: &unsync_map,
                 counter: &counter,
                 read_set: RefCell::new(HashSet::new()),
+                dynamic_change_set_optimizations_enabled: true,
             }),
             1,
         );
@@ -2075,7 +2076,7 @@ mod test {
         );
 
         //TODO: This is printing Ok(Versioned(Err(StorageVersion), ValueType { bytes: Some(b"!0\0\0\0\0\0\0"), metadata: None }, None))
-        // Is Err(StorageVersion) expected here?
+        // Is ValueType = b"!0\0\0\0\0\0\0" expected here?
         println!(
             "data: {:?}",
             versioned_map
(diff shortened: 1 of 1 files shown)


aptos-move/block-executor/src/view.rs
```diff
@@ -1889,4 +1895,135 @@ mod test {
...
+                &counter
+            )),
+            1,
+        );
+
+        assert_eq!(latest_view.get_resource_state_value(&KeyType::<u32>(1, false), None).unwrap(), None);
+        assert_eq!(latest_view.get_resource_state_value(&KeyType::<u32>(2, false), Some(&layout)).unwrap(), None);
+        assert_eq!(latest_view.get_resource_state_value(&KeyType::<u32>(3, false), None).unwrap(), Some(state_value_3.clone()));
+       
+        //TODO: This is printing Ok(Versioned(Err(StorageVersion), ValueType { bytes: Some(b"!0\0\0\0\0\0\0"), metadata: None }, None))
```
georgemitenkov, line 2003: Having `ValueType { bytes: Some(b"!0\0\0\0\0\0\0"), metadata: None }, None)` seems wrong, here we should have bytes from `state_value_3`?

aptos-move/block-executor/src/view.rs
```diff
@@ -1889,4 +1895,135 @@ mod test {
...
+        let state_value_4 = StateValue::new_legacy(patched_value.simple_serialize(&layout).unwrap().into());
+        assert_eq!(latest_view.get_resource_state_value(&KeyType::<u32>(4, false), Some(&layout)).unwrap(), Some(state_value_4.clone()));
+        
+        let captured_reads = latest_view.take_reads();
+        assert!(captured_reads.validate_data_reads(&versioned_map.data(), 1));
+        let mut read_set_with_delayed_fields = captured_reads.get_read_values_with_delayed_fields();
+        
+        // TODO: This prints
+        // read: (KeyType(4, false), Versioned(Err(StorageVersion), Some(Struct(Runtime([Struct(Runtime([Tagged(IdentifierMapping(Aggregator), U64), U64]))])))))
+        // read: (KeyType(2, false), Versioned(Err(StorageVersion), Some(Struct(Runtime([Struct(Runtime([Tagged(IdentifierMapping(Aggregator), U64), U64]))])))))
```
georgemitenkov, line 2020: Seems like we record the read when we have `Some(layout)`, even though there is nothing in storage? Seems like a bug?

aptos-move/block-executor/src/view.rs
```diff
@@ -1889,4 +1895,135 @@ mod test {
...
+        for read in read_set_with_delayed_fields {
+            println!("read: {:?}", read);
+        }
+        
+        // TODO: This assertion fails.
+        // let data_read = DataRead::Versioned(Ok((1,0)), Arc::new(TransactionWrite::from_state_value(Some(state_value_4))), Some(Arc::new(layout)));
+        // assert!(read_set_with_delayed_fields.any(|x| x == (&KeyType::<u32>(4, false), &data_read)));
```
georgemitenkov, line 2027: What is the difference here exactly? I think the version should be not `Ok((1,0))` but `Err(StorageVersion)` because txn 1 never wrote anything, the value was always in storage.

aptos-move/block-executor/src/view.rs
```diff
@@ -740,7 +737,12 @@ impl<'a, T: Transaction, S: TStateView<Key = T::Key>, X: Executable> LatestView<
                                         patched_state_value
                                     },
                                     (_, maybe_state_value, _) => {maybe_state_value}
-                                }
+                                };
+                                state.unsync_map.write(state_key.clone(),
```
georgemitenkov, line 741: Why move here? We should only write patched values

aptos-move/block-executor/src/view.rs
```diff
@@ -1889,4 +1895,135 @@ mod test {
         );
         assert_eq!(identifiers, identifiers2);
     }
+
+    #[test]
+    fn test_read_operations_sequential() {
+        // create latest_view
```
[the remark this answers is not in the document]
igor-aptos: if it is a general sentence or documentation, that makes sense
for few word comments - I don't think that is the convention. doing it here seems worse :)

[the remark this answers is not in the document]
georgemitenkov: /// is documentation, // is just a comment. I suggest this only because LLVM (which has a very high coding standard) is doing that.

aptos-move/block-executor/src/view.rs
igor-aptos: can you explain why this change? was there any test /invariant that required this change?

vusirikala: For the parallel case, we are writing the fetched resource value to versioned_map irrespective of whether layout is Some(.) or None. So, I changed the sequential case to do the same. Should I revert back?

georgemitenkov: I see, I had this question for one of your tests in parallel - as seems like we add `None` in readset but the state value does not exist in storage. I think we should fix parallel instead, and keep sequential as is @igor-aptos ?

aptos-move/block-executor/src/view.rs
vusirikala: I think the value is null because the version is Err(StorageVersion)? Not sure if this is the expected behaviour or a bug.

aptos-move/block-executor/src/view.rs
igor-aptos: I'll take a look in more depth tomorrow, but just thinking about it - isn't it a very different situation?

In parallel, we use CapturedReads to know which speculative reads were valid or not - and for those we need to capture all reads. Then, we reuse that info - for identity exchange in the end

for sequential execution - without identity exchange - we don't need any read information. so we can store here only if move layout is present - no need to store anything else ?

PR #10625, opened 2023-10-23
[CLI] Make startup more responsive in the error case, use a volume for postgres, improve logging

banool:
### Description
I just did a bit of plane coding and while doing so I found a few local testnet issues:
1. There is a rare race where postgres fails to start a second time when using `--force-restart` both times. I figured out in the end that this issue doesn't occur if you use a volume instead of a bindmount. Given no one would want to look at the actual postgres files (and if they do, there are ways), I think this makes sense. It is more performant too.
2. Previously when something crashed on startup (e.g. due to point 1) the user only saw this after `wait_for_startup` timed out. Now we check for either a service crashing or startup completing, so startup failures are surfaced immediately.
3. I improved logging a bit.
4. I put all the docker stuff in its own file.

I also add some guidance to the docs for a common issue that arises from having an old Docker installation (even after upgrading it).

### Test Plan
I ran the local testnet with and without the `--force-restart` and `--with-indexer-api` flag in a variety of different orders, both in debug and release mode.


developer-docs-site/docs/guides/local-development-network.md
```diff
@@ -230,7 +230,26 @@ Make sure the socket for connecting to Docker is present on your machine in the
...
+```
+sudo ln -s /Users/dport/.docker/run/docker.sock /var/run/docker.sock
+```
+
+Alternatively, run the CLI like this to tell it where the socket is:
+```
+DEFAULT_SOCKET=/Users/dport/.docker/run/docker.sock aptos node run-local-testnet --with-indexer-api
+```
+
+Note: If you're on Mac or Windows, we recommend you use Docker Desktop rather than installing Docker via a package manager (e.g. Homebrew or Choco).
```
gregnazario, line 249: Please move this to the top of Prerequisites.  Where you note them to install Docker Desktop first rather than get all the way to the bottom and realize this.

0xmaayan (approved):

PR #10626, opened 2023-10-23
Delete CLI and SDK release pages

banool:
### Description
As discussed offline, these pages are out of date and don't serve much of a purpose given sites already exist for these packages. We're nuking them.

### Test Plan
CI.


0xjinn (approved):

gregnazario (commented):
For SEO, and linking purposes, can you change the pages to point to the other locations that are used for this information?

banool:
I set redirects in docusaurus, is that not sufficient?

gregnazario (approved):

PR #10627, opened 2023-10-23
Visibility Analysis for compiler v2

vineethk:
Compiler v2 now checks that function calls and const uses respect visibility rules.

### Description
We now do the following checks:
* Private functions cannot be called outside the module
* Public friend functions cannot be be called from non-friend modules.
* Constants can only be used directly within the module.

We perform all the function visibility checks as a pass on the stackless bytecode. However, it is too late to do the constant visibility check at this point, as constants have been converted to their values by then, so this is done at an earlier stage (expression builder when building the move model). @brmataptos : inlining implementation could affect this check - something to keep in mind when that implementation rolls in.

### Not in this PR
* Various checks on validity of a friend declaration.
* Fixing (incorrectly flagged) errors in stackless bytecode generation on successfully typechecked programs (there will be a separate issue on this)
* Remove some deadcode identified (we may discuss some of it in this PR)

### Test Plan
* Added new test cases targeting visibility checks.
* Moved over some tests from compiler v1 that also seemed to target visibility rules.

### DIFF
--- a/third_party/move/move-compiler-v2/tests/checking/typing/constant_internal.exp
+++ b/third_party/move/move-compiler-v2/tests/checking/typing/constant_internal.exp
@@ -1,12 +1,13 @@
-// ---- Model Dump
-module 0x2::X {
-} // end 0x2::X
-module 0x2::M {
-    use 0x2::X::{Self, C}; // resolved as: 0x2::X
-    private fun foo() {
-        0;
-        0;
-        Tuple()
-    }
-    spec fun $foo();
-} // end 0x2::M
+
+Diagnostics:
+error: constant `0x2::X::C` cannot be used here because it is private to the module `0x2::X`
+   ┌─ tests/checking/typing/constant_internal.move:10:9
+   │
+10 │         X::C;
+   │         ^^^^
+
+error: constant `0x2::X::C` cannot be used here because it is private to the module `0x2::X`
+   ┌─ tests/checking/typing/constant_internal.move:11:9
+   │
+11 │         C;
+   │         ^
--- a/third_party/move/move-compiler-v2/tests/testsuite.rs
+++ b/third_party/move/move-compiler-v2/tests/testsuite.rs
@@ -7,7 +7,10 @@ use move_binary_format::{binary_views::BinaryIndexedView, file_format as FF};
 use move_command_line_common::files::FileHash;
 use move_compiler::compiled_unit::CompiledUnit;
 use move_compiler_v2::{
-    pipeline::livevar_analysis_processor::LiveVarAnalysisProcessor, run_file_format_gen, Options,
+    pipeline::{
+        livevar_analysis_processor::LiveVarAnalysisProcessor, visibility_checker::VisibilityChecker,
+    },
+    run_file_format_gen, Options,
 };
 use move_disassembler::disassembler::Disassembler;
 use move_ir_types::location;
@@ -28,13 +31,15 @@ pub const EXP_EXT: &str = "exp";
 #[derive(Default)]
 struct TestConfig {
     /// Whether only type check should be run.
-    check_only: bool,
+    type_check_only: bool,
     /// Whether we should dump the AST after successful type check.
     dump_ast: bool,
     /// A sequence of bytecode processors to run for this test.
     pipeline: FunctionTargetPipeline,
-    /// Whether we should generate file format from resulting bytecode
+    /// Whether we should generate file format from resulting bytecode.
     generate_file_format: bool,
+    /// Whether we should dump annotated targets for each stage of the pipeline.
+    dump_annotated_targets: bool,
 }
 
 fn path_from_crate_root(path: &str) -> String {
@@ -74,25 +79,37 @@ impl TestConfig {
         let mut pipeline = FunctionTargetPipeline::default();
         if path.contains("/checking/") {
             Self {
-                check_only: true,
+                type_check_only: true,
                 dump_ast: true,
                 pipeline,
                 generate_file_format: false,
+                dump_annotated_targets: false,
             }
         } else if path.contains("/bytecode-generator/") {
             Self {
-                check_only: false,
+                type_check_only: false,
                 dump_ast: true,
                 pipeline,
                 generate_file_format: false,
+                dump_annotated_targets: true,
             }
         } else if path.contains("/file-format-generator/") {
             pipeline.add_processor(Box::new(LiveVarAnalysisProcessor {}));
             Self {
-                check_only: false,
+                type_check_only: false,
                 dump_ast: false,
                 pipeline,
                 generate_file_format: true,
+                dump_annotated_targets: true,
+            }
+        } else if path.contains("/visibility-checker/") {
+            pipeline.add_processor(Box::new(VisibilityChecker {}));
+            Self {
+                type_check_only: false,
+                dump_ast: false,
+                pipeline,
+                generate_file_format: false,
+                dump_annotated_targets: false,
             }
         } else {
             panic!(
@@ -124,7 +141,7 @@ impl TestConfig {
             out.push_str(&env.dump_env());
             out.push('\n');
         }
-        if ok && !self.check_only {
+        if ok && !self.type_check_only {
             // Run stackless bytecode generator
             let mut targets = move_compiler_v2::run_bytecode_gen(&env);
             let ok = Self::check_diags(&mut test_output.borrow_mut(), &env);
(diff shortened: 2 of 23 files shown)


third_party/move/move-compiler-v2/src/bytecode_generator.rs
```diff
@@ -1047,7 +1047,7 @@ impl<'env> Generator<'env> {
             _ => self.error(
                 id,
                 "assignment to tuple must be tuple itself or a function call",
-            ),
+            ), // TODO: allow assignment to a tuple from move_to, abort, etc.
```
vineethk, line 1050: I'll create a separate issue for this, but when I tried to add the visibility checks on all type check only tests, it exposed a bug where we don't run stackless bytecode gen (sbcodegen) on those programs, and sbcodegen itself could report errors.

Turns out many of those errors are not valid. Eg, `let () = move_to(...)` should be valid, but is not because of this code flagged.

third_party/move/move-compiler-v2/tests/testsuite.rs
```diff
@@ -74,25 +79,38 @@ impl TestConfig {
...
             Self {
-                check_only: false,
+                type_check_only: false,
                 dump_ast: false,
                 pipeline,
                 generate_file_format: true,
+                dump_annotated_targets: true,
+            }
+        } else if path.contains("/visibility-checker/") {
+            // TODO: merge this branch into "/checking/".
```
vineethk, line 106: After speaking with @wrwg, we may keep this separate, i.e., keep this code as is.

third_party/move/move-model/src/ast.rs
```diff
@@ -384,6 +384,18 @@ pub struct UseDecl {
...
 }
 
+// =================================================================================================
+/// # Friend Declarations
+
+/// Represents a `friend` declaration in the source.
+#[derive(Debug, Clone)]
+pub struct FriendDecl {
+    /// Location covered by this declaration.
+    pub loc: Loc,
```
brmataptos, line 394: Do we use this field anywhere?  I don't see it.

third_party/move/move-model/src/model.rs
```diff
@@ -2112,6 +2098,15 @@ impl<'env> ModuleEnv<'env> {
         &self.data.use_decls
     }
 
+    /// Does this module have a friend declaration for `mod_name`?
+    pub fn has_friend_named(&self, mod_name: &ModuleName) -> bool {
+        // Note: we could optimize this lookup using a hash set if needed.
```
brmataptos, line 2103: The old field `friend_modules: BTreeSet<ModuleId>` seems more suitable for doing this lookup.  Why did you change to a vector of new `FriendDecl` structs?

vineethk: This was actually how I started my original implementation (i.e., using `friend_modules`).

Here is why I changed it: that field was never populated. The field is part of `ModuleData`, which is constructed by visiting the modules in a topologically sorted order (with "used" modules visited first) and mapped to a `ModuleId`. Because friend decls forms the reverse of module usage (you declare a module as a friend if you expect that module to use you), the `ModuleId` for friend modules is not available when constructing `ModuleData`. Thus,  you would need to store the `FriendDecls` anyway, and do a pass over all `ModuleData` after their construction to populate the `friend_modules` (possibly at two locations in the code, `run_move_checker` and `run_spec_checker`).  Instead, I chose to just use the friend decls directly (as is, the `friend_modules` were being cloned by the getter, so we would have wanted to implement a method similar to this).

We could still populate the `friend_modules` (as redundant, but potentially more efficient in some cases, especially if there are other use cases for friend module lookups in the future), and use that underneath this method (hence the comment), as a usecase-driven optimization in the future.

third_party/move/move-compiler-v2/src/bytecode_generator.rs
vineethk: I was thinking the other way round would be better: link this code in the issue (which I have yet to create, along with a bunch of others), and then we can do all the issue management we want.

third_party/move/move-compiler-v2/tests/checking/typing/constant_internal.exp
```diff
@@ -1,12 +1,13 @@
...
-    private fun foo() {
-        0;
-        0;
-        Tuple()
-    }
-    spec fun $foo();
-} // end 0x2::M
+
+Diagnostics:
+error: constant `0x2::X::C` is inaccessible here
```
wrwg, line 3: Align on nomenclature? Either 'inaccessible' or 'not visible'

third_party/move/move-compiler-v2/tests/testsuite.rs
wrwg: Remove the TODO then?

third_party/move/move-model/src/ast.rs
wrwg: Yes please keep this, its a public API which aims at providing all available info.

third_party/move/move-model/src/builder/exp_builder.rs
```diff
@@ -1922,6 +1922,19 @@ impl<'env, 'translator, 'module_translator> ExpTranslator<'env, 'translator, 'mo
             },
         };
         if let Some(entry) = self.parent.parent.const_table.get(&global_var_sym).cloned() {
+            // Check if the constant was declared in a different module (unless this is in a spec).
+            if self.mode != ExpTranslationMode::Spec
+                && global_var_sym.module_name != self.parent.module_name
```
wrwg, line 1927: I think this should be move into translate_constant

third_party/move/move-model/src/builder/module_builder.rs
```diff
@@ -608,6 +614,32 @@ impl<'env, 'translator> ModuleBuilder<'env, 'translator> {
         });
     }
 
+    fn decl_ana_friend_decl(&mut self, friend_mod_id: &EA::ModuleIdent, friend: &EA::Friend) {
+        let address = match friend_mod_id.value.address {
```
wrwg, line 618: There is already a function `self.resolve_address` which you can use here. It will error on NamedUnassigned since it is expected that this has been resolved during expansion.

third_party/move/move-model/src/builder/module_builder.rs
```diff
@@ -608,6 +614,32 @@ impl<'env, 'translator> ModuleBuilder<'env, 'translator> {
...
+                    Address::Numerical(num_address)
+                } else {
+                    symbolic_addr
+                }
+            },
+        };
+        let friend_mod_sym = self
+            .symbol_pool()
+            .make(friend_mod_id.value.module.0.value.as_str());
+        let friend_mod_name = ModuleName::new(address, friend_mod_sym);
```
wrwg, line 635: I get it that this name cannot be resolved here as the module is possibly not yet processed (modules are build in bottom-up dependency order). But we really would like to have `FriendDecl.module_id` instead of `FriendDecl.module_name` because that is the general way how modules are represented in the model.

Achieving this is a bit more work. You would have first build some temporary data structure to keep friend decls (like `BTreeMap<ModuleId, FriendEntry>` in the `ModelBuilder` and than after all modules have been build, patch this against the modules in the env. This can be done e.g. in `run_move_checker` which is the main entry point for v2 compiler model building.

third_party/move/move-model/src/model.rs
```diff
@@ -1303,10 +1304,8 @@ impl GlobalEnv {
         }
 
         let used_modules = self.get_used_modules_from_bytecode(&module);
-        let friend_modules = self.get_friend_modules_from_bytecode(&module);
```
wrwg, line 1306: I think we need to keep this for the case we work from a CompiledModule. In general, we need to be able to construct a model from bytecode only, without access to the expansion AST, and then this is the only info we have. If this conflicts with your FriendDecl, you could have two functions like `get_friends_via_bytecode`  which accesses this field and your new function.

third_party/move/move-model/src/model.rs
wrwg: See my comments above: (a) want to keep the old function for the case model is build from bytecode only (b) have new function for v2 but have the module name resolve to module id.

wrwg (commented):
Looks great so far

third_party/move/move-model/src/builder/module_builder.rs
vineethk: To confirm, we don't need to do this for `run_spec_checker` as well because it is deprecated?

wrwg: run_spec_checker is not deprecated but the entry point for the Move prover, but yes, we don't need to do it there, and only for v2, since run_spec_checker gets this info from bytecode.

vineethk:
I believe I have addressed all comments. Please take another look.

third_party/move/move-model/src/builder/module_builder.rs
```diff
@@ -587,9 +593,9 @@ impl<'env, 'translator> ModuleBuilder<'env, 'translator> {
         let module_name = ModuleName::new(given_addr, module_sym);
         let module_id = self
             .parent
-            .env
-            .find_module(&ModuleName::new(resolved_addr, module_sym))
-            .map(|e| e.get_id());
+            .module_table
```
vineethk, line 596: `find_module` is a linear search, so using the newly added `module_table` for a lookup instead.

third_party/move/move-compiler-v2/src/bytecode_generator.rs
wrwg: Yes, those places were not following this practice, but I agree with Brian that moving forward we should do this.

wrwg (approved):
Great!

rahxephon89 (approved):
looks good to me, thanks!

PR #10637, opened 2023-10-24
[BlockSTM] Modify/Revert Gas Counter Behavior

ziaptos:
### Description
The gas counters (for logging) and log info! was not shown during all types of block_executor exits. Ensure we show it when 

### Test Plan
Given tests


aptos-move/block-executor/src/txn_last_input_output.rs
```diff
@@ -162,13 +162,13 @@ impl<T: Transaction, O: TransactionOutput<Txn = T>, E: Debug + Send + Clone>
...
     /// Does a transaction at txn_idx have SkipRest or Abort status.
-    pub(crate) fn block_skips_rest_at_idx(&self, txn_idx: TxnIndex) -> bool {
+    pub(crate) fn block_truncated_at_idx(&self, txn_idx: TxnIndex) -> bool {
         matches!(
             &self.outputs[txn_idx as usize]
                 .load_full()
                 .expect("[BlockSTM]: Execution output must be recorded after execution")
                 .output_status,
-            ExecutionStatus::SkipRest(_)
+            ExecutionStatus::SkipRest(_) | ExecutionStatus::Abort(_)
```
sitalkedia, line 171: This is not only changing the gas counter behavior but also changing the early abort behavior.. is this expected?

ziaptos: It's actually not changing the early abort behavior, the two conditions were just checked separately in `prepare_and_queue_commit_ready_txns`

sitalkedia: @ziaptos - May be I am missing something but this seems to be changing the early exit behavior because of the check - https://github.com/aptos-labs/aptos-core/blob/08f0aaa2523d8d9da5e539e7a642f99b7d726088/aptos-move/block-executor/src/txn_last_input_output.rs#L176. Previously we early exited when we hit this but now that might not be the case..

ziaptos: @sitalkedia You're right - we do skip that early exit with this PR. Let me send a new PR that's on top the new changes.

But to clarify, we don't want to update the counters/log info when we early exit b/c of the rw conflict, and we do want to log in all other cases?

Also, I assume the counters are used for benchmarking? Do we actually need them here at all? why not just do it after blockSTM is called?

sitalkedia: These counters are not used for benchmarking..I didn't add them.. may be @danielxiangzl has more context.

igor-aptos: These are the main gas counters, which we show on dashboard (and show in executor benchmark) etc, which previous refactor broke.
If we don't fix this - they are going to disappear from the dashboard.


We can move them a level higher if we want - probably in the execute_block() inside executor.rs?


If nothing easy can be done immediately, we can just revert to the previous behavior - of not making logging conditional on halt() output?

ziaptos: I think the logic is fixed now - reverted to the previous behavior. We still need to protect with `if halt()` to guarantee we only do it once tho.

I agree that we can move that to a different function, like you suggested - that actually makes more sense.  Please review regardless :)

zekun000 (approved):

danielxiangzl (approved):
LGTM

PR #10641, opened 2023-10-24
[cleanup] Improve natives, remove extra aggregator location info

georgemitenkov:


aptos-move/framework/src/natives/mod.rs
```diff
@@ -83,12 +83,3 @@ pub fn all_natives(
 
     make_table_from_iter(framework_addr, natives)
 }
-
-/// A temporary hack to patch Table -> table module name as long as it is not upgraded
```
georgemitenkov, line 87: @vgao1996 this is not needed right? I checked and there are no usages

igor-aptos (approved):

davidiw (approved):

PR #10651, opened 2023-10-24
[Agg v2] Add failpoints that raise code invariant error 

vusirikala:
### Description

Adding failpoints that raise code invariant error. This is to test whether the recent changes in block executor properly handle the code invariant error and falls back to sequential execution.

### DIFF
--- a/aptos-move/aptos-aggregator/src/delayed_field_extension.rs
+++ b/aptos-move/aptos-aggregator/src/delayed_field_extension.rs
@@ -12,6 +12,7 @@ use crate::{
         SnapshotValue,
     },
 };
+use fail::fail_point;
 use move_binary_format::errors::PartialVMResult;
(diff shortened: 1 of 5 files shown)


aptos-move/aptos-aggregator/src/delayed_field_extension.rs
```diff
@@ -40,6 +41,12 @@ impl DelayedFieldData {
         input: SignedU128,
         resolver: &dyn DelayedFieldResolver,
     ) -> PartialVMResult<bool> {
+        if resolver.is_delayed_field_optimization_capable() {
+            fail_point!(
```
georgemitenkov, line 45: Seems that this way we end up with a branch in real code? I think you can add a boolean argument to the fail macro to make it execute conditionally:

https://docs.rs/fail/latest/fail/macro.fail_point.html

Not sure though if it is evaluated lazily or not

PR #10658, opened 2023-10-24
[forge] Ability to set onchain config in tests

ibalajiarun:
### Description

This PR enables setting on-chain consensus config in Forge tests as part of the genesis config.

### Test Plan

Test in Forge

### DIFF
--- a/crates/aptos-genesis/src/config.rs
+++ b/crates/aptos-genesis/src/config.rs
@@ -7,6 +7,7 @@ use aptos_types::{
     account_address::{AccountAddress, AccountAddressWithChecks},
     chain_id::ChainId,
     network_address::{DnsName, NetworkAddress, Protocol},
+    on_chain_config::{OnChainConsensusConfig, OnChainExecutionConfig},
     transaction::authenticator::AuthenticationKey,
 };
 use aptos_vm_genesis::{AccountBalance, EmployeePool, Validator, ValidatorWithCommissionRate};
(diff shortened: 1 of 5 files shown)


terraform/helm/genesis/values.yaml
```diff
@@ -31,16 +31,18 @@ chain:
...
 imageTag: testnet
 
 genesis:
   image:
     # -- Image repo to use for tools image for running genesis
-    repo: aptoslabs/tools
+    repo: us-west1-docker.pkg.dev/aptos-global/aptos-internal/tools
     # -- Image tag to use for tools image. If set, overrides `imageTag`
-    tag:
+    tag: 8597cbdff337833d70560b6705ff7eadbcf62111
```
sionescu, line 45: Let's avoid hardcoding a SHA1 here.

ibalajiarun: That's for local testing. I'll undo.

crates/aptos-genesis/src/config.rs
```diff
@@ -103,6 +107,7 @@ impl Default for Layout {
             total_supply: None,
             employee_vesting_start: Some(1663456089),
             employee_vesting_period_duration: Some(5 * 60), // 5 minutes
+            on_chain_consensus_config: OnChainConsensusConfig::default(),
```
sitalkedia, line 110: Can you add support for `OnChainExecutionConfig` as well ? :) that way we can just remove some ugly overrides @msmouse has for large TPS test to run...

ibalajiarun: Sure, done.

sitalkedia (approved):

terraform/helm/genesis/values.yaml
msmouse: Thanks bosses!

crates/aptos-genesis/src/config.rs
msmouse: Thanks bosses!

msmouse (approved):

PR #10661, opened 2023-10-24
Gate AggregatovV2 via DelayedFeilds via node config flag

igor-aptos:
improve testing, to assert same execution between both enabled and disabled mode as well.

### DIFF
--- a/aptos-move/block-executor/src/executor.rs
+++ b/aptos-move/block-executor/src/executor.rs
@@ -734,18 +734,27 @@ where
         let resource_write_set = last_input_output.resource_write_set(txn_idx);
         let delayed_field_keys = last_input_output.delayed_field_keys(txn_idx);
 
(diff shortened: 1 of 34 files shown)


aptos-move/block-executor/src/executor.rs
```diff
@@ -1066,7 +1096,10 @@ where
                     unsync_map: &unsync_map,
                     counter: &counter,
                     read_set: RefCell::new(HashSet::new()),
-                    dynamic_change_set_optimizations_enabled,
```
georgemitenkov, line 1069: Do we still need `dynamic_change_set_optimizations_enabled` flag?

movekevin (approved):

igor-aptos:
turns out we cannot make gas charges work equivalently, so we will have to use flag to gate aggregators. (follow-ups PR https://github.com/aptos-labs/aptos-core/pull/10698 addresses gas charging)

PR #10662, opened 2023-10-24
[agg_v2] As a precaution - verify all DelayedFieldIDs that native code was called with

igor-aptos:
…

### DIFF
--- a/aptos-move/aptos-aggregator/src/tests/types.rs
+++ b/aptos-move/aptos-aggregator/src/tests/types.rs
@@ -6,7 +6,10 @@ use crate::{
     bounded_math::{BoundedMath, SignedU128},
     delta_change_set::serialize,
     resolver::{TAggregatorV1View, TDelayedFieldView},
-    types::{expect_ok, DelayedFieldID, DelayedFieldValue, DelayedFieldsSpeculativeError, PanicOr},
+    types::{
+        code_invariant_error, expect_ok, DelayedFieldID, DelayedFieldValue,
+        DelayedFieldsSpeculativeError, PanicOr,
+    },
 };
 use aptos_types::{
     aggregator::PanicError,
@@ -36,6 +39,7 @@ pub struct FakeAggregatorView {
     // delayed_field_try_add_delta_outcome operate on different state
     v1_store: HashMap<StateKey, StateValue>,
     v2_store: HashMap<DelayedFieldID, DelayedFieldValue>,
+    start_counter: u32,
     counter: RefCell<u32>,
 }
 
@@ -45,6 +49,7 @@ impl Default for FakeAggregatorView {
             v1_store: HashMap::new(),
             v2_store: HashMap::new(),
             // Put some recognizable number, to easily spot missed exchanges
+            start_counter: FAKE_AGGREGATOR_VIEW_GEN_ID_START,
             counter: RefCell::new(FAKE_AGGREGATOR_VIEW_GEN_ID_START),
         }
     }
@@ -111,6 +116,28 @@ impl TDelayedFieldView for FakeAggregatorView {
         let id = Self::Identifier::new(*counter as u64);
         *counter += 1;
         id
+    }
+
+    fn validate_and_convert_delayed_field_id(
+        &self,
+        id: u64,
+    ) -> Result<Self::Identifier, PanicError> {
+        if id < self.start_counter as u64 {
+            return Err(code_invariant_error(format!(
+                "Invalid delayed field id: {}, we've started from {}",
+                id, self.start_counter
+            )));
+        }
+
+        let current = *self.counter.borrow();
+        if id > current as u64 {
+            return Err(code_invariant_error(format!(
+                "Invalid delayed field id: {}, we've only reached to {}",
+                id, current
+            )));
+        }
+
+        Ok(Self::Identifier::new(id))
     }
 
     fn get_reads_needing_exchange(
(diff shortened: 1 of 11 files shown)


georgemitenkov:
@igor-aptos overall looks good, any reason for these extra checks? Or these are "just in case"?

aptos-move/block-executor/src/view.rs
```diff
@@ -997,6 +1001,36 @@ impl<'a, T: Transaction, S: TStateView<Key = T::Key>, X: Executable> TDelayedFie
...
         }
     }
+
+    fn validate_and_convert_delayed_field_id(&self, id: u64) -> Result<Self::Identifier, PanicError> {
+        let start_counter = match &self.latest_view {
+            ViewState::Sync(state) => state.start_counter,
+            ViewState::Unsync(state) => state.start_counter,
+        };
+
+        if id < start_counter as u64 {
```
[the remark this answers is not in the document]
igor-aptos: if we forget to do the exhcnage - this will happen.

i.e. if we were to enable aggregators now , without resource group PR, this would catch it.

in the future, if someone adds a new thing that needs special handling (lke table and resource group needed it), this will catch it.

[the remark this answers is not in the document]
igor-aptos: or if in case gating doesn't work correctly - we think we don't need to do the exchange, but native code assumes those are identifiers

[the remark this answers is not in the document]
igor-aptos: (i.e. these are not checks on whether counter worked correctly, but whether exchange actually happened. and that spans many different files - and so is good to check)

aptos-move/aptos-vm/src/natives.rs
```diff
@@ -22,6 +22,7 @@ use aptos_gas_schedule::{MiscGasParameters, NativeGasParameters, LATEST_GAS_FEAT
 use aptos_native_interface::SafeNativeBuilder;
 #[cfg(feature = "testing")]
 use aptos_table_natives::{TableHandle, TableResolver};
+use aptos_types::aggregator::PanicError;
```
georgemitenkov, line 25: Seems like format is missing?

aptos-move/aptos-vm/src/natives.rs
```diff
@@ -45,20 +46,16 @@ use {
     },
     move_vm_runtime::native_extensions::NativeContextExtensions,
     once_cell::sync::Lazy,
-    std::sync::atomic::{AtomicU32, Ordering},
 };
 
 #[cfg(feature = "testing")]
 struct AptosBlankStorage {
```
georgemitenkov, line 52: `struct AptosBlankStorage { } `--> `struct AptosBlankStorage;`

aptos-move/aptos-vm/src/natives.rs
```diff
@@ -101,7 +98,11 @@ impl TDelayedFieldView for AptosBlankStorage {
     }
 
     fn generate_delayed_field_id(&self) -> Self::Identifier {
-        (self.counter.fetch_add(1, Ordering::SeqCst) as u64).into()
+        unimplemented!()
```
georgemitenkov, line 101: unreachable?

aptos-move/aptos-vm/src/natives.rs
georgemitenkov: Also I guess you need #[cfg(feature = "testing")] here if not merged into the next import

aptos-move/aptos-vm/src/storage_adapter/executor_view_base.rs
```diff
@@ -22,18 +22,12 @@ pub trait AsExecutorView<S> {
 /// Adapter to convert a `StateView` into an `ExecutorView`.
```
georgemitenkov, line 22: This file is gone on main :)

aptos-move/framework/src/natives/aggregator_natives/helpers_v2.rs
```diff
@@ -14,16 +14,20 @@ use move_vm_types::values::{StructRef, Value};
 const VALUE_FIELD_INDEX: usize = 0;
 const LIMIT_FIELD_INDEX: usize = 1;
 
-pub(crate) fn aggregator_value_field_as_id(value: u128) -> PartialVMResult<DelayedFieldID> {
-    u128_to_u64(value).map(DelayedFieldID::new)
+pub(crate) fn aggregator_value_field_as_id(value: u128, resolver: &dyn DelayedFieldResolver) -> PartialVMResult<DelayedFieldID> {
+    let value_u64 = u128_to_u64(value)?;
+    Ok(resolver.validate_and_convert_delayed_field_id(value_u64)?)
```
georgemitenkov, line 19: So here PanicError gets converted to invariant violation, but the state is not recovered. I guess we cannot fall-back from here anyway?

aptos-move/aptos-aggregator/src/tests/types.rs
```diff
@@ -100,4 +102,25 @@ impl TDelayedFieldView for FakeAggregatorView {
         *counter += 1;
         id
     }
+
+    fn validate_and_convert_delayed_field_id(&self, id: u64) -> Result<Self::Identifier, PanicError> {
```
georgemitenkov, line 106: I don't think this is used in tests? e.g. `set_from_aggregator_id` uses a delayed id already

georgemitenkov (approved):

gelash (approved):

aptos-move/framework/src/natives/aggregator_natives/helpers_v2.rs
igor-aptos: no, it gets converted to DELAYED_FIELDS_CODE_INVARIANT_ERROR

PR #10665, opened 2023-10-25
 FrozenSmt not holding oldest ancester 

msmouse:
### Description
Now that ledger update is a seperate stage, the oldest ancester is held                                                                                                                                  from multiple threads, which can result in it not being freed when the                                                                                                                                   pipeline is busy.    

### Test Plan
On certain machines where we used to observe high memory usage, it's now good.
Local benchmark runs didn't observe regression.
Unit tests.

### DIFF
--- a/storage/aptosdb/src/test_helper.rs
+++ b/storage/aptosdb/src/test_helper.rs
@@ -10,7 +10,6 @@ use crate::{
 use aptos_crypto::hash::CryptoHash;
 use aptos_executor_types::ProofReader;
 use aptos_jellyfish_merkle::node_type::{Node, NodeKey};
-use aptos_scratchpad::SparseMerkleTree;
 use aptos_temppath::TempPath;
 use aptos_types::{
     contract_event::ContractEvent,
@@ -109,8 +108,6 @@ pub fn update_in_memory_state(state: &mut StateDelta, txns_to_commit: &[Transact
         if txn_to_commit.is_state_checkpoint() {
             state.current = state
                 .current
-                .clone()
-                .freeze()
                 .batch_update(
                     state
                         .updates_since_base
@@ -118,11 +115,9 @@ pub fn update_in_memory_state(state: &mut StateDelta, txns_to_commit: &[Transact
                         .flatten()
                         .map(|(k, v)| (k.hash(), v.as_ref()))
                         .collect(),
-                    StateStorageUsage::new_untracked(),
                     &ProofReader::new_empty(),
                 )
-                .unwrap()
-                .unfreeze();
+                .unwrap();
             state.current_version = next_version.checked_sub(1);
             state.base = state.current.clone();
             state.base_version = state.current_version;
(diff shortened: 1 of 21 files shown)


storage/scratchpad/src/sparse_merkle/ancestors.rs
```diff
@@ -0,0 +1,87 @@
...
+
+type Result<V> = std::result::Result<V, Error>;
+
+/// Keep the oldest SMTs in a central place so that:
+/// 1. elsewhere, dropping the SMT will be fast since this indirectly hols a ref to every tree in
+///    the entire forest.
+/// 2. this must be invoked somewhere in the critical path so that it can provide some
+///    back pressure to slow things down to prevent from leaking memory
+#[derive(Clone, Debug)]
+pub struct SmtAncestors<V: Clone + Send + Sync + 'static> {
```
grao1991, line 26: nit: remove Clone

grao1991 (approved):

storage/scratchpad/src/sparse_merkle/ancestors.rs
msmouse: Clone is required down the stack from the updater. I'd follow up later on this in favor of time. 😢

storage/scratchpad/src/sparse_merkle/ancestors.rs
```diff
@@ -0,0 +1,87 @@
...
+            oldest.log_generation("evict_ancestor");
+            if !oldest.is_the_only_ref() {
+                sample!(
+                    SampleRate::Duration(Duration::from_secs(1)),
+                    error!(
+                        "Oldest SMT being tracked by SmtAncestors is still referenced elsewhere. Potential memory leak.",
+                    )
+                );
+            } else {
+                self.drop_queue.enqueue_drop(oldest);
```
areshand, line 77: is it supposed to enque when MAX_PENDING_DROPS reaches?

msmouse: enqueue_drop

msmouse: `enqueue_drop()` will block until backlog drops back below that

storage/aptosdb/src/test_helper.rs
```diff
@@ -109,20 +108,16 @@ pub fn update_in_memory_state(state: &mut StateDelta, txns_to_commit: &[Transact
         if txn_to_commit.is_state_checkpoint() {
             state.current = state
                 .current
-                .clone()
-                .freeze()
```
areshand, line 113: why is this freeze and unfreeze removed?

areshand (approved):

storage/aptosdb/src/test_helper.rs
msmouse: just a different way to hack.. (SparseMerkleTree::batch_update() now generates `Usage::Untracked` internally, as is the intention here)

PR #10672, opened 2023-10-25
[move-vm] Remove hashmap

runtian-zhou:
### Description

Use back btreemap to replace hashmap due to potential non-determinism concern. Also use bit vector to replace the inefficient Vec<bool>

### Test Plan
We should run a replay on this.

### DIFF
--- a/third_party/move/move-vm/types/src/loaded_data/runtime_types.rs
+++ b/third_party/move/move-vm/types/src/loaded_data/runtime_types.rs
@@ -13,6 +13,7 @@ use move_core_types::{
     gas_algebra::AbstractMemorySize, identifier::Identifier, language_storage::ModuleId,
     vm_status::StatusCode,
 };
+use smallbitvec::SmallBitVec;
 use std::{cmp::max, collections::BTreeMap, fmt::Debug, sync::Arc};
 
 pub const TYPE_DEPTH_MAX: usize = 256;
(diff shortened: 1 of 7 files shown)


vgao1996:
I'm a bit confused by the commit history -- it seems like you are changing a Vec<bool> into bit vec and then changing it back?

vgao1996:
NVM I confused myself by getting the order wrong. However it seems like you are still including the bit set commits. Is that intended?

third_party/move/move-vm/types/src/loaded_data/runtime_types.rs
```diff
@@ -194,6 +190,43 @@ pub enum Type {
     U256,
 }
 
+// Cache for the ability of struct. They will be ignored when comparing equality or Ord as they are just used for caching purpose.
+#[derive(Derivative)]
+#[derivative(Debug, Clone, Eq, Hash, PartialEq, Ord, PartialOrd)]
```
wrwg, line 195: Seems you needed this for the ignore semantics (shouldn't it have been there already for hashmap?). Not a fan of doing something like this (i.e. cheating on equality :-) ), I think the abilities belong into a different map, but this is for another time.

wrwg (approved):
Was the HashMap actually iterated? In any case, good to switch to ordered map.

vgao1996 (approved):

runtian-zhou:
> Was the HashMap actually iterated? In any case, good to switch to ordered map.

It was iterated. The reaonsing why the current code is fine is a bit more nuanced here: we put the iteration result back to a sorted list so the happy path should be fine. The unhappy path is also not possible because there were only simple serialziation errors there. But it would be great to avoid this footgun in general.

PR #10675, opened 2023-10-25
[api] Add contextual error message to simulation api

runtian-zhou:
### Description

Add an optional message field to MoveAbort errors. Return the full error message by simulation api,

### Test Plan
Abort will now look like following:

```
$ ~/workspace/aptos-core/target/debug/aptos move run --function-id "2a1f6c336636b1243d24cfd4edfd4086057f4a2db069c247ba43433e3bebaad7::todolist::create_list"
Simulation failed with status: Move abort in 0x2a1f6c336636b1243d24cfd4edfd4086057f4a2db069c247ba43433e3bebaad7::todolist: 0x2a
ABORTED {
    code: 42,
    location: 2a1f6c336636b1243d24cfd4edfd4086057f4a2db069c247ba43433e3bebaad7::todolist,
    message: Some(
        "0x2a1f6c336636b1243d24cfd4edfd4086057f4a2db069c247ba43433e3bebaad7::todolist::create_list at offset 9",
    ),
}
```

In the future we can have full nodes to attach full stack trace to the message field.


georgemitenkov (approved):
Nice! I was actually propagating errors in my refactoring but as a `SubmitTransactionError` instead. So we want the simulation be "successful" even on errors? I can update in my PR accordingly then.

runtian-zhou:
> Nice! I was actually propagating errors in my refactoring but as a `SubmitTransactionError` instead. So we want the simulation be "successful" even on errors? I can update in my PR accordingly then.

I don't mind if it returns error or not. To me it's more about for the error case we can attach more error message. Abort is not a concern here cuz we've already explained it.

api/src/transactions.rs
```diff
@@ -1232,7 +1233,22 @@ impl TransactionsApi {
                 let mut user_transactions = Vec::new();
                 for transaction in transactions.into_iter() {
                     match transaction {
-                        Transaction::UserTransaction(user_txn) => user_transactions.push(*user_txn),
+                        Transaction::UserTransaction(user_txn) => {
+                            let mut txn = *user_txn;
+                            match &vm_status {
+                                VMStatus::Error {
```
georgemitenkov, line 1239: You can use `!vm_status.is_success()` to catch `MoveAbort`s as well

davidiw (approved):
@gregnazario fyi

api/src/transactions.rs
gregnazario: Let's try to use whatever @georgemitenkov has refactored it to be to be consistent.

gregnazario (approved):

api/src/transactions.rs
```diff
@@ -1256,7 +1256,13 @@ impl TransactionsApi {
...
                     match transaction {
-                        Transaction::UserTransaction(user_txn) => user_transactions.push(*user_txn),
+                        Transaction::UserTransaction(user_txn) => {
+                            let mut txn = *user_txn;
+                            if let VMStatus::Error { .. } = vm_status {
+                                txn.info.vm_status += format!("\nExecution failed with status: {:#?}", vm_status).as_str();
+                            }
+                            user_transactions.push(txn);
+                        },
                         _ => {
```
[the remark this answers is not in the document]
georgemitenkov: No, that is still under review, I only split the PR and landed the necessary refactoring so far.

PR #10676, opened 2023-10-25
[AdminService] Add several debug APIs to dump consensus data.

grao1991:
### Description

Added 3 endpoints:

`/debug/consensus/consensusdb`: dump everything in consensus db.
`/debug/consensus/quorumstoredb`: dump all quorum store batchs, also a query param `digest` is supported for dumping a single batch.
`/debug/consensus/block`: dump all blocks in consensus, also a query param `block_id` is supported for dumping a single block.

### Test Plan
Manually tested on a local test node.

### DIFF
--- a/crates/aptos-admin-service/src/server/consensus/mod.rs
+++ b/crates/aptos-admin-service/src/server/consensus/mod.rs
@@ -1,0 +1,239 @@
+// Copyright © Aptos Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::server::utils::{reply_with, reply_with_status, spawn_blocking};
+use anyhow::{bail, Error};
+use aptos_consensus::{
+    persistent_liveness_storage::PersistentLivenessStorage,
+    quorum_store::{quorum_store_db::QuorumStoreStorage, types::PersistedValue},
+};
+use aptos_consensus_types::{block::Block, common::Payload};
+use aptos_crypto::HashValue;
+use aptos_logger::info;
+use aptos_types::transaction::SignedTransaction;
+use http::header::{HeaderValue, CONTENT_LENGTH};
+use hyper::{Body, Request, Response, StatusCode};
+use std::{collections::HashMap, sync::Arc};
+
+pub async fn handle_dump_consensus_db_request(
+    _req: Request<Body>,
+    consensus_db: Arc<dyn PersistentLivenessStorage>,
+) -> hyper::Result<Response<Body>> {
+    info!("Dumping consensus db.");
+
+    match spawn_blocking(move || dump_consensus_db(consensus_db.as_ref())).await {
+        Ok(result) => {
+            info!("Finished dumping consensus db.");
+            let headers: Vec<(_, HeaderValue)> =
+                vec![(CONTENT_LENGTH, HeaderValue::from(result.len()))];
+            Ok(reply_with(headers, result))
+        },
+        Err(e) => {
+            info!("Failed to dump consensus db: {e:?}");
+            Ok(reply_with_status(
+                StatusCode::INTERNAL_SERVER_ERROR,
+                e.to_string(),
+            ))
+        },
+    }
+}
+
+pub async fn handle_dump_quorum_store_db_request(
+    req: Request<Body>,
+    quorum_store_db: Arc<dyn QuorumStoreStorage>,
+) -> hyper::Result<Response<Body>> {
+    let query = req.uri().query().unwrap_or("");
+    let query_pairs: HashMap<_, _> = url::form_urlencoded::parse(query.as_bytes()).collect();
+
+    let digest: Option<HashValue> = match query_pairs.get("digest") {
+        Some(val) => match val.parse() {
+            Ok(val) => Some(val),
+            Err(err) => return Ok(reply_with_status(StatusCode::BAD_REQUEST, err.to_string())),
+        },
+        None => None,
+    };
+
+    info!("Dumping quorum store db.");
+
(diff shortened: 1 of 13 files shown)


crates/aptos-admin-service/src/server/consensus/mod.rs
```diff
@@ -0,0 +1,217 @@
...
+    let mut body = String::new();
+
+    let (last_vote, highest_tc, consensus_blocks, consensus_qcs) =
+        consensus_db.consensus_db().get_data()?;
+
+    body.push_str(&format!("Last vote: \n{last_vote:?}\n\n"));
+    body.push_str(&format!("Highest tc: \n{highest_tc:?}\n\n"));
+    body.push_str(&format!("Blocks: \n"));
+    for block in consensus_blocks {
+        body.push_str(&format!(
```
msmouse, line 118: why not `serde_json::to_string()`?

crates/aptos-admin-service/src/server/consensus/mod.rs
```diff
@@ -0,0 +1,217 @@
...
+            block.id(),
+            block.author(),
+            block.epoch(),
+            block.round(),
+            block.parent_id(),
+            block.timestamp_usecs(),
+            block.payload(),
+        ));
+    }
+    body.push_str(&format!("QCs: \n"));
```
msmouse, line 129: and why not output the whole thing as Json (or json per line), to enable processing with jq / more sophisticated scripts

crates/aptos-admin-service/src/server/consensus/mod.rs
```diff
@@ -0,0 +1,217 @@
...
+        body.push_str(&format!("{qc:?}\n\n"));
+    }
+
+    Ok(body)
+}
+
+fn dump_quorum_store_db(quorum_store_db: &dyn QuorumStoreStorage) -> anyhow::Result<String> {
+    let mut body = String::new();
+
+    for (digest, batch) in quorum_store_db.get_all_batches()? {
```
msmouse, line 140: So the entirety of QS DB is expected to always fit in mem? @zekun000 ?

zekun000: we call it upon restart too, so hopefully it fits :) cc @bchocho

crates/aptos-admin-service/src/server/consensus/mod.rs
```diff
@@ -0,0 +1,217 @@
...
+            Ok(reply_with_status(
+                StatusCode::INTERNAL_SERVER_ERROR,
+                e.to_string(),
+            ))
+        },
+    }
+}
+
+fn dump_consensus_db(consensus_db: &dyn PersistentLivenessStorage) -> anyhow::Result<String> {
+    let mut body = String::new();
```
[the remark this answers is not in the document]
grao1991: this is not huge.. I want to be able to return error.

crates/aptos-admin-service/src/server/consensus/mod.rs
bchocho: Yeah, it might not be small, but it should fit in memory 🫢

crates/aptos-admin-service/src/server/consensus/mod.rs
grao1991: my format looks more beautiful

crates/aptos-admin-service/src/server/consensus/mod.rs
grao1991: ditto

msmouse (approved):

crates/aptos-admin-service/src/server/consensus/mod.rs
ibalajiarun: that's a reason we shouldn't (or don't?) store remote batches in the db.

crates/aptos-admin-service/src/server/consensus/mod.rs
```diff
@@ -0,0 +1,239 @@
...
+};
+use aptos_consensus_types::{block::Block, common::Payload};
+use aptos_crypto::HashValue;
+use aptos_logger::info;
+use aptos_types::transaction::SignedTransaction;
+use http::header::{HeaderValue, CONTENT_LENGTH};
+use hyper::{Body, Request, Response, StatusCode};
+use std::{collections::HashMap, sync::Arc};
+
+pub async fn handle_dump_consensus_db_request(
```
ibalajiarun, line 18: This looks very barebones - pretty much dumping the DB (literally). 

Do we have plans to add user-friendly APIs to inspect the consensus DB like we do with Aptos DB via the Aptos API?

grao1991: Any specific API in your mind? I'm under an impression that the consensus db won't be too large (assuming quorum store is enabled), so I didn't add those query by epoch/round/block_id api.

crates/aptos-admin-service/src/server/consensus/mod.rs
sitalkedia: Can you clarify how we plan to consume this? If this is needed for some other tooling (like replay tool), then dumping this as binary file might be useful?

grao1991: I expect only the block?block_id=xxx API will be consumed by program (I left a TODO to support bcs format for that). For other API I only expect people use it in browser so I'm trying to make them human friendly.

sitalkedia (approved):
LGTM

PR #10678, opened 2023-10-26
Limit the maximum identifier length

junkil-park:
### Description
cherry pick the PR [[VM] Limit the maximum identifier length ](https://github.com/aptos-labs/aptos-core/pull/10502)

### Test Plan
cargo test


movekevin (approved):

zi0Black (approved):

davidiw:
Folks, let's not break compatibility with main. I'm reverting this PR.

This need to take a flag that has not yet been allocated. So maybe we get this landed on main first?

PR #10679, opened 2023-10-26
Update release.yaml

junkil-park:
### Description
Update `release.yaml` for v1.8

### DIFF
--- a/aptos-move/aptos-release-builder/data/release.yaml
+++ b/aptos-move/aptos-release-builder/data/release.yaml
@@ -1,75 +1,54 @@
 ---
 remote_endpoint: ~
-name: "v1.7"
+name: "v1.8"
 proposals:
-  - name: step_1_enable_module_event
+  - name: upgrade_framework
     metadata:
-      title: "Enable module event"
-      description: "AIP-44: The new module event framework associates every event stream with a static struct type instead of an EventHandle of current event system."
-      discussion_url: "https://github.com/aptos-foundation/AIPs/issues/200"
-    execution_mode: MultiStep
-    update_sequence:
-      - FeatureFlag:
-          enabled:
-            - module_event
-  - name: step_2_upgrade_framework
-    metadata:
-      title: "Multi-step proposal to upgrade mainnet framework to v1.7"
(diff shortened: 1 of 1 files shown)


aptos-move/aptos-release-builder/data/release.yaml
```diff
@@ -1,75 +1,64 @@
...
+      title: "AIP-55: Generalize Transaction Authentication and Support Arbitrary K-of-N MultiKey Accounts"
+      description: "AIP-55: Generalize Transaction Authentication and Support Arbitrary K-of-N MultiKey Accounts."
+      discussion_url: "https://github.com/aptos-foundation/AIPs/issues/267"
     execution_mode: MultiStep
     update_sequence:
       - FeatureFlag:
           enabled:
-            - bulletproofs_natives
-  - name: step_6_enable_storage_deletion_refund
+            - single_sender_authenticator
```
davidiw, line 54: should we combine this was the secp? since it is the same feature flag.

junkil-park: @runtian-zhou also confirmed that they should be combined if they use the same feature flag. 

Would this be a proper way to combine the information from two steps? https://github.com/aptos-labs/aptos-core/pull/10679/files#diff-f0942a7d2fc980b7d9d90ec193ec4e9214cdb34c3e3c1d59fa7d5507325b1574R15

I am not sure if `discussion_url: https://url1, https://url2` would end up generating a proper link to the discussions. (@movekevin , @perryjrandall ?)

davidiw (approved):

movekevin (commented):
All look correct

aptos-move/aptos-release-builder/data/release.yaml
junkil-park: @kent-white confirmed that two discussion urls in one proposal would not work in UI. So, I added the links to the `description` field and kept only one link (for AIP-55) in the `discussion_url` field.
@davidiw , @movekevin , @perryjrandall , @runtian-zhou , please let me know if you have any comments or concern for this.

perryjrandall (approved):

movekevin (approved):

Issue #9343, opened 2023-07-27
[Bug][compiler v2] Evaluation order of pack expressions


wrwg:
Compiler v1 currently implements order dependent evaluation of pack arguments. That is, in an expression `S{b: e1, a: e2}`, even though if field `a` comes before `b`, the expression `e1` will be evaluated first.

The move-model AST and hence the current v2 compile chain does not support this currently, and always evaluates pack arguments in the order of the fields in the struct.

There is a test case in `move-compiler-v2/transactional-tests/tests/evaluation-order/struct_arguments.move` showing this problem.

PR #10682, opened 2023-10-26
[compiler-v2] Fix evaluation order for pack expressions

fEst1ck:
### Description
Fixes issue #9343 by transforming `pack { f: e, ... }` to `let $f = e; ...; pack { f: $f, ...}`. See
[third_party/move/move-compiler-v2/tests/bytecode-generator/pack_order.move](https://github.com/aptos-labs/aptos-core/pull/10682/files#diff-d3af3789d9ccb449f4d70fdf1c60f021f172de9603f276bb3f5ad44e2a55dd2d)
and
[bytecode-generator/pack_order.exp](https://github.com/aptos-labs/aptos-core/pull/10682/files#diff-296562a9c14c9dda440922f6008b77cc7c2ab0142e5b94f1082452a610c49ec4).

This PR refactors `translate_fields` to avoid deeply nested `if let Some(_) ... { ... } else { error_handling() }`, and specializes it into `translate_exp_fields` and `translate_lvalue_fields` (see https://github.com/aptos-labs/aptos-core/pull/10682#pullrequestreview-1703286926).

### Test Plan
Existing compiler tests, including `move-compiler-v2/transactional-tests/tests/evaluation-order/struct_arguments.move`, plus `pack_order.move`.

### DIFF
--- a/third_party/move/move-compiler-v2/tests/bytecode-generator/fields.exp
+++ b/third_party/move/move-compiler-v2/tests/bytecode-generator/fields.exp
@@ -24,32 +24,14 @@ module 0x42::fields {
     }
     private fun write_local_direct(): fields::S {
         {
-          let x: fields::S = {
-            let .f: u64 = 0;
-            {
-              let .g: fields::T = {
-                let .h: u64 = 0;
-                pack fields::T(.h)
-              };
-              pack fields::S(.f, .g)
-            }
-          };
+          let x: fields::S = pack fields::S(0, pack fields::T(0));
           select fields::T.h(select fields::S.g(x)) = 42;
           x
         }
     }
     private fun write_local_via_ref(): fields::S {
         {
-          let x: fields::S = {
-            let .f: u64 = 0;
-            {
-              let .g: fields::T = {
-                let .h: u64 = 0;
-                pack fields::T(.h)
-              };
-              pack fields::S(.f, .g)
-            }
-          };
+          let x: fields::S = pack fields::S(0, pack fields::T(0));
           {
             let r: &mut fields::S = Borrow(Mutable)(x);
             select fields::T.h(select fields::S.g(r)) = 42;
@@ -138,30 +120,24 @@ fun fields::write_local_direct(): fields::S {
      var $t1: fields::S
      var $t2: fields::S
      var $t3: u64
-     var $t4: u64
-     var $t5: fields::T
-     var $t6: fields::T
-     var $t7: u64
-     var $t8: u64
-     var $t9: u64
-     var $t10: &mut u64
-     var $t11: &mut fields::T
-     var $t12: &mut fields::S
-  0: $t4 := 0
-  1: $t3 := move($t4)
-  2: $t8 := 0
-  3: $t7 := move($t8)
-  4: $t6 := pack fields::T($t7)
-  5: $t5 := move($t6)
-  6: $t2 := pack fields::S($t3, $t5)
-  7: $t1 := move($t2)
-  8: $t9 := 42
-  9: $t12 := borrow_local($t1)
- 10: $t11 := borrow_field<fields::S>.g($t12)
- 11: $t10 := borrow_field<fields::T>.h($t11)
- 12: write_ref($t10, $t9)
- 13: $t0 := move($t1)
- 14: return $t0
+     var $t4: fields::T
+     var $t5: u64
+     var $t6: u64
+     var $t7: &mut u64
+     var $t8: &mut fields::T
+     var $t9: &mut fields::S
+  0: $t3 := 0
+  1: $t5 := 0
+  2: $t4 := pack fields::T($t5)
+  3: $t2 := pack fields::S($t3, $t4)
+  4: $t1 := move($t2)
+  5: $t6 := 42
+  6: $t9 := borrow_local($t1)
+  7: $t8 := borrow_field<fields::S>.g($t9)
+  8: $t7 := borrow_field<fields::T>.h($t8)
+  9: write_ref($t7, $t6)
+ 10: $t0 := move($t1)
+ 11: return $t0
 }
 
 
@@ -171,32 +147,26 @@ fun fields::write_local_via_ref(): fields::S {
      var $t1: fields::S
      var $t2: fields::S
      var $t3: u64
-     var $t4: u64
-     var $t5: fields::T
-     var $t6: fields::T
-     var $t7: u64
+     var $t4: fields::T
+     var $t5: u64
+     var $t6: &mut fields::S
+     var $t7: &mut fields::S
      var $t8: u64
-     var $t9: &mut fields::S
-     var $t10: &mut fields::S
-     var $t11: u64
-     var $t12: &mut u64
-     var $t13: &mut fields::T
-  0: $t4 := 0
-  1: $t3 := move($t4)
-  2: $t8 := 0
-  3: $t7 := move($t8)
-  4: $t6 := pack fields::T($t7)
-  5: $t5 := move($t6)
-  6: $t2 := pack fields::S($t3, $t5)
-  7: $t1 := move($t2)
-  8: $t10 := borrow_local($t1)
-  9: $t9 := move($t10)
- 10: $t11 := 42
- 11: $t13 := borrow_field<fields::S>.g($t9)
- 12: $t12 := borrow_field<fields::T>.h($t13)
- 13: write_ref($t12, $t11)
- 14: $t0 := move($t1)
- 15: return $t0
+     var $t9: &mut u64
+     var $t10: &mut fields::T
+  0: $t3 := 0
+  1: $t5 := 0
+  2: $t4 := pack fields::T($t5)
+  3: $t2 := pack fields::S($t3, $t4)
+  4: $t1 := move($t2)
+  5: $t7 := borrow_local($t1)
+  6: $t6 := move($t7)
+  7: $t8 := 42
+  8: $t10 := borrow_field<fields::S>.g($t6)
+  9: $t9 := borrow_field<fields::T>.h($t10)
+ 10: write_ref($t9, $t8)
+ 11: $t0 := move($t1)
+ 12: return $t0
 }
 
 
--- a/third_party/move/move-compiler-v2/tests/bytecode-generator/fields.exp
+++ b/third_party/move/move-compiler-v2/tests/bytecode-generator/fields.exp
@@ -24,14 +24,32 @@ module 0x42::fields {
     }
     private fun write_local_direct(): fields::S {
         {
-          let x: fields::S = pack fields::S(0, pack fields::T(0));
+          let x: fields::S = {
+            let .f: u64 = 0;
+            {
+              let .g: fields::T = {
+                let .h: u64 = 0;
+                pack fields::T(.h)
+              };
+              pack fields::S(.f, .g)
+            }
+          };
           select fields::T.h(select fields::S.g(x)) = 42;
           x
         }
     }
     private fun write_local_via_ref(): fields::S {
         {
-          let x: fields::S = pack fields::S(0, pack fields::T(0));
+          let x: fields::S = {
+            let .f: u64 = 0;
+            {
+              let .g: fields::T = {
+                let .h: u64 = 0;
+                pack fields::T(.h)
+              };
+              pack fields::S(.f, .g)
+            }
+          };
           {
             let r: &mut fields::S = Borrow(Mutable)(x);
             select fields::T.h(select fields::S.g(r)) = 42;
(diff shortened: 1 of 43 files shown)


third_party/move/move-model/src/builder/exp_builder.rs
```diff
@@ -2739,6 +2740,116 @@ impl<'env, 'translator, 'module_translator> ExpTranslator<'env, 'translator, 'mo
         }
     }
 
+    fn translate_fields_exp(
```
fEst1ck, line 2743: This is adapted from `translate_fields`, but it additionally returns the local bindings introduced in the transformation.

vineethk: I have not fully reviewed this carefully, but on a first glance, this seems like a lot of repeated code from `translate_fields`. Please try refactoring (for example, can the same function be used to do both, using a boolen input to do uncommon stuff).

vineethk (changes requested):

third_party/move/move-compiler-v2/tests/checking/typing/implicit_deref_borrow_field_non_ref_non_local_root.exp
```diff
@@ -4,7 +4,10 @@ module 0x8675309::M {
         f: u64,
     }
     private fun bar(): M::S {
-        pack M::S(0)
+        {
+          let .f: u64 = 0;
```
wrwg, line 8: Can avoid doing the transformation if its not needed? Only the cases were the arguments are out of order need this.

third_party/move/move-compiler/src/diagnostics/mod.rs
```diff
@@ -61,6 +61,15 @@ pub fn report_diagnostics(files: &FilesSourceText, diags: Diagnostics) -> ! {
     std::process::exit(1)
 }
 
+// report diagnostics, but do not exit if diags are all warnings
+pub fn report_diagnostics_exit_on_error(files: &FilesSourceText, diags: Diagnostics) {
```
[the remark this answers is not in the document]
wrwg: `report_diagnostics_and_exit_on_error`? Though the current name is also good with me.

Note that in this corner of the code, 'diagnostics' is spelled out.

third_party/move/move-model/src/builder/exp_builder.rs
```diff
@@ -2652,8 +2675,11 @@ impl<'env, 'translator, 'module_translator> ExpTranslator<'env, 'translator, 'mo
         maccess: &EA::ModuleAccess,
         generics: &Option<Vec<EA::Type>>,
         fields: &EA::Fields<T>,
-        mut field_translator: impl FnMut(&mut Self, &Type, &T) -> S,
-    ) -> Option<(QualifiedInstId<StructId>, Vec<S>)> {
+        init: S,
+        // used to fold results when iterating over the fields
+        // the args are: translator, acc, field_name, exp_idx, def_idx, field_ty, value
+        mut combine: impl FnMut(&mut Self, &mut S, &Symbol, usize, usize, &Type, &T),
```
wrwg, line 2681: This is too complex. A lambda with 6 arguments is not readable. Besides it is not documented what all those things mean. Perhaps put this into a struct (or part of it, like `FieldInfo{field_name,, exp_idx, def_idx, field_ty}`?

third_party/move/move-model/src/builder/exp_builder.rs
```diff
@@ -2673,15 +2699,23 @@ impl<'env, 'translator, 'module_translator> ExpTranslator<'env, 'translator, 'mo
                         .keys()
                         .filter(|s| *s != &self.parent.dummy_field_name()),
                 );
-                let mut args = BTreeMap::new();
-                for (name_loc, name_, (_, value)) in fields.iter() {
+
+                let mut res = init;
+                // exp_idx is the idx of the field expression in the pack/unpack expression
+                // def_idx is the idx of the field as in the struct definition
+                for (name_loc, name_, (exp_idx, value)) in fields.iter() {
```
wrwg, line 2706: Why `name_`? Something special with it?

third_party/move/move-model/src/builder/exp_builder.rs
wrwg: I'm not quite sure why the same as translate_fields? Its actually calling it.

wrwg (changes requested):
This has become quite complex, and I think the reason is the generic `translate_fields` which takes a lambda for dealing with both patterns and expressions. With the new complexity introduced by the lets, this generalization does not make a lot of sense any more. For patterns, we do not need the reording. I'd suggest to specialize this into two functions `translate_exp_fields` and `translate_lvalue_fields`, remove the lambda, and simplify this significantly. There might be a bit of duplication but it will greatly improve readability.

fEst1ck:
The code has been thoroughly refactored now, including the original code for translating unpack expression.

wrwg (commented):
Very nice! Solved the problem and the overall code is in the end less complex than before. Please also address the issues from last round about avoiding unnecessary transformation and temporary names. If you done that, there should be only minimal .exp  changes I assume. One explicit test would be also good.

third_party/move/move-compiler-v2/tests/bytecode-generator/fields.exp
```diff
@@ -15,14 +15,32 @@ module 0x42::fields {
     }
     private fun write_local_direct(): fields::S {
         {
-          let x: fields::S = pack fields::S(0, pack fields::T(0));
+          let x: fields::S = {
+            let .f: u64 = 0;
```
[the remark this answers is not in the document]
brmataptos: [citation needed]
Is this documented anywhere?  If not, @wrwg please do so.

third_party/move/move-model/src/builder/exp_builder.rs
```diff
@@ -2813,112 +2825,235 @@ impl<'env, 'translator, 'module_translator> ExpTranslator<'env, 'translator, 'mo
...
+        // maps i to (x_i, e_i) where i is the expression idx, i.e., idx of the field exp in the pack exp
+        let mut bindings = BTreeMap::new();
+        // maps i to x_i where i is the field definition idx, i.e., idx of the field in the struct definition
+        let mut args = BTreeMap::new();
+        for (_, name, (exp_idx, field_exp)) in fields.iter() {
+            let field_name = self.symbol_pool().make(name);
+            let (_, def_idx, field_ty) = field_decls.get(&field_name).unwrap();
+            let field_ty = field_ty.instantiate(&instantiation);
+            let translated_field_exp = self.translate_exp(field_exp, &field_ty);
+            // starts with . to avoid shadowing
```
brmataptos, line 2979: We need to document somewhere if we're using special symbol names to mean something.  I'm having the same problem with inlining (avoiding shadowing existing variables) and we should avoid potential conflicts with future name-mangling or something.

wrwg:
There is no citation needed here. The leading `$` is an existing convention in the Move Prover for compiler internal names (see e.g. prelude.bpl -- its all over the place). I'm saying here we should use this convention for compiler v2 as well. There is no need this to be documented anywhere, we are in the design phase. Do you have any objections against the `$` convention?

third_party/move/move-compiler-v2/tests/bytecode-generator/fields.exp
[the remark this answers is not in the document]
fEst1ck: > As said in the last round, our convention is to start temps with '$'. This is important to identify because those names should not be shown in error messages. To avoid clash with other '$' vars, you could generated `$S_f`.

Just realized that shadowing is actually not a problem here:
```
let $f = 0;
pack { f: $f + 1 }
```
will still produce
```
let $f = 0;
{
    let $f = $f + 1;
    pack {f : $f}
}
```
correctly.

fEst1ck:
The latest commit minimizes the number of fields transformed. See [bytecode-generator/pack_order.exp](https://github.com/aptos-labs/aptos-core/pull/10682/files#diff-296562a9c14c9dda440922f6008b77cc7c2ab0142e5b94f1082452a610c49ec4).

third_party/move/move-model/src/builder/exp_builder.rs
```diff
@@ -2813,112 +2825,281 @@ impl<'env, 'translator, 'module_translator> ExpTranslator<'env, 'translator, 'mo
...
+                    prev = cur;
+                    in_order_fields = BTreeSet::new();
+                    in_order_fields.insert(i);
+                }
+            }
+        }
+        in_order_fields
+    }
+
+    // given pack<S>{ f_p(1): e_1, ... }, where p is a permutation of the fields
```
brmataptos, line 2981: This comment is confusing.  How does `pack<S>(f_p(1): e_1, ...}` correspond to the parameters to this function?

third_party/move/move-model/src/builder/exp_builder.rs
```diff
@@ -2813,112 +2825,281 @@ impl<'env, 'translator, 'module_translator> ExpTranslator<'env, 'translator, 'mo
...
+            succeed = false;
+        }
+        if succeed {
+            Some(())
+        } else {
             None
         }
     }
 
+    // return the indices of fields that can be left in place during transforming pack exprs
```
brmataptos, line 2942: This seems very convoluted.  I haven't thought about it as much as you, but it seems to me to be simpler to map from exp_idx to def_idx, rather than the other way around.  Given a list of (field, expr) sorted by exp_idx, the tail which is in order are the in_order_fields; the prefix of that list must be evaluated in evaluation order and stored in temporary vars.

In any case, adding a comment before each line to explain what you are computing might be helpful.

third_party/move/move-model/src/builder/exp_builder.rs
```diff
@@ -2813,112 +2825,281 @@ impl<'env, 'translator, 'module_translator> ExpTranslator<'env, 'translator, 'mo
...
+        }
+        let permutation = permutation
+            .into_iter()
+            .sorted_by_key(|(i, _)| *i)
+            .map(|(_, value)| value)
+            .collect_vec();
+        let mut in_order_fields = BTreeSet::new();
+        if !permutation.is_empty() {
+            let mut prev = permutation[0];
+            in_order_fields.insert(0);
```
brmataptos, line 2963: Some comment explaining how this works or citing some standard algorithm would be useful...

third_party/move/move-model/src/builder/exp_builder.rs
```diff
@@ -2813,112 +2825,281 @@ impl<'env, 'translator, 'module_translator> ExpTranslator<'env, 'translator, 'mo
...
+        for (_, name, (exp_idx, field_exp)) in fields.iter() {
+            let field_name = self.symbol_pool().make(name);
+            let (_, def_idx, field_ty) = field_decls.get(&field_name).unwrap();
+            let field_ty = field_ty.instantiate(&instantiation);
+            let translated_field_exp = self.translate_exp(field_exp, &field_ty);
+            if in_order_fields.contains(def_idx) {
+                args.insert(*def_idx, translated_field_exp);
+            } else {
+                // starts with $ for internal generated vars
+                let var_name = self
```
brmataptos, line 3025: Please use some helper functions to build these AST elements.  Any set of k let statements with just 1 output var and <k input vars would be better to put in a separate function.  Maybe even k lines of let statements.

third_party/move/move-compiler-v2/tests/bytecode-generator/pack_order.exp
```diff
@@ -0,0 +1,157 @@
...
+        f1: u8,
+        f2: u8,
+        f3: u8,
+    }
+    private fun pack1(x: u8,y: u8,z: u8): pack_unpack::S {
+        pack pack_unpack::S(x, y, z)
+    }
+    private fun pack2(x: u8,y: u8,z: u8): pack_unpack::S {
+        {
+          let $f1: u8 = x;
```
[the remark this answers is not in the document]
wrwg: This may not work because of scoping, and the current thing is fine. Lets not add unnecessary asks to this.

wrwg (approved):
Great, lets get this in.

third_party/move/move-compiler-v2/tests/bytecode-generator/pack_order.exp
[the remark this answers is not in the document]
wrwg: FWIW, scoping shouldn't be an issue, so this would work, as long as the evaluation order in the tuple is from left to right.

third_party/move/move-model/src/builder/exp_builder.rs
fEst1ck: You are right; sorting by exp_idx is better. The right implementation if we sort by def_idx is slow. The above impl only works if `def_idx`'s are consecutive.

fEst1ck: fixed in 5fac01bb1e2f171acd8e559dc12e443b4cb6a4e3

third_party/move/move-model/src/builder/exp_builder.rs
brmataptos: Actually this is Rust, so you can return a tuple.

Btw, better if your variable names can be related to in the comment above.  "var" = "x" and "arg" = "exp"?

third_party/move/move-model/src/builder/exp_builder.rs
```diff
@@ -2874,6 +2888,120 @@ impl<'env, 'translator, 'module_translator> ExpTranslator<'env, 'translator, 'mo
...
+        &mut self,
+        loc: &Loc,
+        maccess: &EA::ModuleAccess,
+        generics: &Option<Vec<EA::Type>>,
+        fields: &EA::Fields<EA::LValue>,
+        // whether matching against a reference value
+        ref_expected: Option<ReferenceKind>,
+        // pack { f_i : t_i }, should the type of f_i >: t_i or <: t_i
+        expected_order: WideningOrder,
+        // whether the pattern vars are already bound
```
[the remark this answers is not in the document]
brmataptos: Um, those 2 expressions look the same to me.  What is the difference between the 2 examples?
-  let (x, y) = (1, 2)
-  let (x, y) = (1, 2)

third_party/move/move-model/src/builder/exp_builder.rs
fEst1ck: > Btw, better if your variable names can be related to in the comment above. "var" = "x" and "arg" = "exp"?
I added some comments for the vars.

fEst1ck: > Please use some helper functions to build these AST elements. Any set of k let statements with just 1 output var and <k input vars would be better to put in a separate function. Maybe even k lines of let statements.
I refactored the if branch; but for the else branch, I actually prefer the current.

third_party/move/move-model/src/builder/exp_builder.rs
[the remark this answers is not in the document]
brmataptos: Oh, first one isn't a `let`.

brmataptos (approved):
Ok, time to let this get in.

vineethk (approved):

